PR #1514, opened 2015-05-12
Fix #1469 - Adding framework name to the marathon config block on 'About' section.

### LINKED ISSUE #1469: Marathon About page presents a confusing 'Name' which is not the framework name
On the Marathon about page a **Name** field is shown, which is the project BuildInfo.name (see InfoResource.scala and not the --framework_name. It is probably a good idea to either change or add a Framework Name to the Marathon about page. For example, here is a Marathon instance started with `--framework_name` set to `marathon-public`

screen shot 2015-05-05 at 1 30 28 pm

hekaldama:


kolloch:
:100:

kolloch:
There is still work in the UI necessary for fixing #1469, right? Or does the UI pick that up automatically?

hekaldama:
From what I can tell in the code path, it should automagically pick it up. I tried to verify with a test but couldn't figure out how to add one. Will try again now.

PR #1523, opened 2015-05-13
Set a default value for zk_max_versions, if this parameter is omitted.

### LINKED ISSUE #1521: Use a sensible default for maximum number of versions, if zk_max_versions is not defined.
Every change to a group or application creates a new version, that is stored in zk. 
The number of versions can grow infinitely by default, which can lead to read timeouts.
The maximum number of versions can be defined with --zk_max_versions.
If this parameter is not defined, a sensible default should take place.

aquamatthias:
Fixes #1521 

@drexin please review


drexin:
:100:

PR #1527, opened 2015-05-13
Fixes #1460 - Allow configuring resource roles per app

### LINKED ISSUE #1460: Let users specify the Mesos role for each Marathon app
In the [future](https://issues.apache.org/jira/browse/MESOS-1763), frameworks will be able to register with multiple roles.  When this happens, we'll want the ability to tell Marathon which role's resources to use to launch tasks for each app.

Frameworks are currently offered resources associated both with the registration role and the default (`*`) role.  We already allow users to choose what role to register as (via the `--mesos_role` flag).  The current behavior is to ignore the role associated with each resource offer when launching tasks.  A per-app role would already be useful to distinguish between these sets of resources.

Currently, and even more so in the "future" scenario, a `--default_roles` flag could tune the scheduler behavior when inspecting offer roles.  For example given that a Marathon scheduler is registered with role `A`, receiving offers assiciated with roles `A` and `*`, the administrator may want to control the default roles for apps that do not specify per-app roles.

cc @cmaloney @pyronicide

**To summarize the proposal:**

- [ ] Introduce a `--default_app_roles` command line flag to Marathon that defaults to `*` if no value for `--mesos_role` is supplied and `*,A` if `--mesos_role A` is supplied.
- [ ] Introduce a `mesosRoles` (or similarly named) attribute to the App definition to control the role-associated resources to use for launching tasks for that app.  Validation should ensure that the value is either `*` or the role that Marathon used to register with Mesos.

kolloch:
and provide a possibility to configure a global default.

### DIFF
--- a/docs/docs/command-line-flags.md
+++ b/docs/docs/command-line-flags.md
@@ -69,7 +69,11 @@ The core functionality flags can be also set by environment variable `MARATHON_O
 * `--max_tasks_per_offer_cycle` (Optional. Default: 1000): Launch at most this number of tasks per Mesos offer cycle.
   A larger value speeds up launching new tasks.
   Yet, choosing a too large value might overwhelm Mesos/Marathon with processing task updates.
-* `--mesos_role` (Optional. Default: None): Mesos role for this framework.
+* `--mesos_role` (Optional. Default: None): Mesos role for this framework. If set, Marathon receives resource offers
+    for the specified role in addition to resources with the role designation '*'.
+* `--default_accepted_resource_roles` (Optional. Default: all roles): Default for the `"acceptedResourceRoles"`
+    attribute as a comma-separated list of strings. All app definitions which do not specify this attribute explicitly
+    use this value for launching new tasks. Examples: `*`, `production,*`, `production`
 * `--mesos_user` (Optional. Default: current user): Mesos user for
     this framework. _Note: Default is determined by
     [`SystemProperties.get("user.name")`](http://www.scala-lang.org/api/current/index.html#scala.sys.SystemProperties@get\(key:String\):Option[String])._
--- a/docs/docs/rest-api.md
+++ b/docs/docs/rest-api.md
@@ -112,6 +112,9 @@ The full JSON format of an application resource is as follows:
     "constraints": [
         ["attribute", "OPERATOR", "value"]
     ],
+    "acceptedResourceRoles": [
+        "role1", "*"
+    ],
     "labels": {
         "environment": "staging"
     },
@@ -194,6 +197,31 @@ The command that is executed.  This value is wrapped by Mesos via `/bin/sh -c ${
 Valid constraint operators are one of ["UNIQUE", "CLUSTER",
 "GROUP_BY"]. For additional information on using placement constraints see
 the [Constraints doc page]({{ site.baseurl }}/docs/constraints.html).
+
+##### acceptedResourceRoles
+
+Optional. A list of resource roles. Marathon considers only resource offers with roles in this list for launching
+tasks of this app. If you do not specify this, Marathon considers all resource offers with roles that have been
+configured by the `--default_accepted_resource_roles` command line flag. If no `--default_accepted_resource_roles` was
+given on startup, Marathon considers all resource offers.
+
+Example 1: `"acceptedResourceRoles": [ "production", "*" ]` Tasks of this app definition are launched either
+on "production" or "*" resources.
+
+Example 2: `"acceptedResourceRoles": [ "public" ]` Tasks of this app definition are launched only on "public"
+resources.
+
+Background: Mesos can assign roles to certain resource shares. Frameworks which are not explicitly registered for
+a role do not see resources of that role. In this way, you can reserve resources for frameworks. Resources not reserved
+for custom role, are available for all frameworks. Mesos assigns the special role "*" to them.
+
+To register Marathon for a role, you need to specify the `--mesos_role` command line flag on startup.
+If you want to assign all resources of a
+slave to a role, you can use the `--default_role` argument when starting up the slave. If you need a more
+fine-grained configuration, you can use the `--resources' argument to specify resource shares per role. The Mesos master
+needs to be started with `--roles` followed by a comma-separated list of all roles you want to use across your cluster.
+See
+[the Mesos command line documentation](http://mesos.apache.org/documentation/latest/configuration/) for details.
 
 ##### labels
 
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -71,14 +71,14 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
     case None       => Set("*")
   }
 
-  lazy val defaultAcceptedResourceRolesSet = defaultAcceptedResourceRoles.get.get
+  lazy val defaultAcceptedResourceRolesSet = defaultAcceptedResourceRoles.get.getOrElse(expectedResourceRoles)
 
   lazy val defaultAcceptedResourceRoles = opt[String]("default_accepted_resource_roles",
     descr =
       "Default for the defaultAcceptedResourceRoles attribute of all app definitions" +
         " as a comma-separated list of strings." +
         "This defaults to all roles for which this Marathon instance is configured to receive offers.",
-    default = Some(expectedResourceRoles.mkString(",")),
+    default = None,
     validate = validateDefaultAcceptedResourceRoles).map(parseDefaultAcceptedResourceRoles)
 
   private[this] def parseDefaultAcceptedResourceRoles(str: String): Set[String] =
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -61,8 +61,40 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
   def executor: Executor = Executor.dispatch(defaultExecutor())
 
   lazy val mesosRole = opt[String]("mesos_role",
-    descr = "Mesos role for this framework",
+    descr = "Mesos role for this framework. " +
+      "If set, Marathon receives resource offers for the specified role in addition to " +
+      "resources with the role designation '*'.",
     default = None)
+
+  def expectedResourceRoles: Set[String] = mesosRole.get match {
+    case Some(role) => Set(role, "*")
+    case None       => Set("*")
+  }
+
+  lazy val defaultAcceptedResourceRolesSet = defaultAcceptedResourceRoles.get.get
+
+  lazy val defaultAcceptedResourceRoles = opt[String]("default_accepted_resource_roles",
+    descr =
+      "Default for the defaultAcceptedResourceRoles attribute of all app definitions" +
+        " as a comma-separated list of strings." +
+        "This defaults to all roles for which this Marathon instance is configured to receive offers.",
+    default = Some(expectedResourceRoles.mkString(",")),
+    validate = validateDefaultAcceptedResourceRoles).map(parseDefaultAcceptedResourceRoles)
+
+  private[this] def parseDefaultAcceptedResourceRoles(str: String): Set[String] =
+    str.split(',').map(_.trim).toSet
+
+  private[this] def validateDefaultAcceptedResourceRoles(str: String): Boolean = {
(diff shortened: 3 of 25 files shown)


docs/docs/command-line-flags.md
```diff
@@ -69,7 +69,15 @@ The core functionality flags can be also set by environment variable `MARATHON_O
 * `--max_tasks_per_offer_cycle` (Optional. Default: 1000): Launch at most this number of tasks per Mesos offer cycle.
   A larger value speeds up launching new tasks.
   Yet, choosing a too large value might overwhelm Mesos/Marathon with processing task updates.
-* `--mesos_role` (Optional. Default: None): Mesos role for this framework.
+* `--mesos_role` (Optional. Default: None): Mesos role for this framework. If set, Marathon receives resource offers
+    for the specified role in addition to resources with the role designation '*'.
+* `--default_accepted_resource_roles` (Optional. Default: all roles): Default for the `"defaultAcceptedResourceRoles"`
```
aquamatthias: Default for `acceptedResourceRoles`.

docs/docs/command-line-flags.md
```diff
@@ -69,7 +69,15 @@ The core functionality flags can be also set by environment variable `MARATHON_O
...
+    attribute as a comma-separated list of strings. All app definitions which do not specify this attribute explicitly
+    use this value. This defaults to all roles
+    for which this Marathon instance is configured to receive offers (see `--mesos_role`). Changing this configuration
```
aquamatthias: I think it is better to describe _when_ this default  applies and what the consequence is. 
I would omit hints regarding restarts in usage info.

src/main/scala/mesosphere/marathon/tasks/DefaultTaskFactory.scala
```diff
@@ -21,6 +21,12 @@ class DefaultTaskFactory @Inject() (
 
   private[this] val log = LoggerFactory.getLogger(getClass)
 
+  // validate here and not in task builder so that we detect the error on startup
+  require(config.defaultAcceptedResourceRolesSet.nonEmpty, "--default_accepted_resource_roles must not be empty")
```
aquamatthias: can we do the validation directly in the scallop option?

kolloch: ah cool, didn't know that this is possible. Will do that.

docs/docs/command-line-flags.md
kolloch: I asked @pyronicide to help me with the text.

src/main/scala/mesosphere/marathon/tasks/ResourceUtil.scala
```diff
@@ -109,10 +114,10 @@ object ResourceUtil {
...
     */
   def consumeResources(resources: Iterable[Resource], usedResources: Iterable[Resource]): Iterable[Resource] = {
-    val usedResourceMap = usedResources.map(resource => resource.getName -> resource).toMap
+    val usedResourceMap = usedResources.map(resource => ResourceId(resource) -> resource).toMap
 
     resources.flatMap { resource =>
-      usedResourceMap.get(resource.getName) match {
+      usedResourceMap.get(ResourceId(resource)) match {
         case Some(usedResource) =>
           if (resource.getType != usedResource.getType) {
```
aquamatthias: Would it make sense to also add the resource type into the ResourceId, so that we don't have to handle that case specifically?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -8,22 +8,28 @@ import org.apache.log4j.Logger
...
+    def findScalarResourceRole(tpe: String, value: Double): Option[Role] =
+      groupedResources.get(tpe).flatMap {
+        _
+          .filter(resource => acceptedResourceRoles(resource.getRole))
+          .find { resource =>
+            resource.getScalar.getValue >= value
+          }.map(_.getRole)
+      }
 
     val cpuRoleOpt = findScalarResourceRole(Resource.CPUS, app.cpus)
```
aquamatthias: Can we move from val to def here for all opts? The for comprehension below will only evaluate, if all preceding options are defined.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -29,20 +29,31 @@ class TaskBuilder(app: AppDefinition,
...
+    val acceptedResourceRoles: Set[String] = app.acceptedResourceRoles.getOrElse(config.defaultAcceptedResourceRolesSet)
 
-      case _ =>
-        log.info(
-          s"No matching offer for ${app.id} (need cpus=${app.cpus}, mem=${app.mem}, " +
-            s"disk=${app.disk}, ports=${app.hostPorts()}) : " + offer
-        )
-        None
+    if (log.isDebugEnabled) {
+      log.debug(s"acceptedResourceRoles $acceptedResourceRoles")
```
aquamatthias: Just a side note: we should change the logging to something different to not test for isDebugEnabled...

src/main/scala/mesosphere/marathon/tasks/ResourceUtil.scala
kolloch: Hmmm. it's possible but is it what we want? We wouldn't get warnings for this anymore and I think a warning is appropriate. Don't you?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
kolloch: these are very, very cheap, this has nothing to do with the PR but if I can make you happy that easily ;)

kolloch: well the constraint/port matching might not be cheap (even though it is usually in practice). So that might make things better.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -29,20 +29,31 @@ class TaskBuilder(app: AppDefinition,
...
     }
+
+    ResourceMatcher.matchResources(
+      offer, app, taskTracker.get(app.id),
+      acceptedResourceRoles = acceptedResourceRoles) match {
```
aquamatthias: why a named parameter here?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
kolloch: that's right.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
kolloch: because the type of acceptedResourceRoles is fairly generic (Set[String]) and I like to use named parameters to avoid confusion in these cases.

docs/docs/command-line-flags.md
kolloch: jepp.

kolloch:
I added quite some tests now and addressed the comments above.

REST API documentation is still missing. I will fix that next.

aquamatthias:
Initial review done. 
I had some comments, but overall I agree with this approach.
We need to do some real world testing, before we merge it into master.

kolloch:
I added some REST API documentation. I also provided some background information there because I think it is hard to understand otherwise. 

@lloesche helped me to test it. We found one bug which I fixed now (forgot the writing part of the google protocol buffer serialization). Otherwise it looked good.

We used the following test setup:

* Start the Mesos master with `--roles=public`
* Start one Mesos slave with `--ip=127.0.0.1`
* Start another Mesos slave on `--ip=127.0.0.2`, a different work dir and `--default_role=public` 
* Start the Marathon Master with `--mesos_role public --default_accepted_resource_roles \*`

Now we can create a "private" app like this:

```http
PUT /v2/apps/service-private
```
```json
{
  "id": "service-private",
  "cmd": "python -m SimpleHTTPServer $PORT0",
  "cpus": 0.0001,
  "instances": 1,
  "mem": 32,
  "ports": [0]
}
```

and a public app like this

```http
PUT /v2/apps/service-public
```
```json
{
  "id": "service-public",
  "cmd": "python -m SimpleHTTPServer $PORT0",
  "cpus": 0.0001,
  "instances": 1,
  "acceptedResourceRoles": [
    "public"
  ],
  "mem": 32,
  "ports": [ 0]
}
```

Scale up and have fun! :)

docs/docs/rest-api.md
```diff
@@ -195,6 +198,30 @@ Valid constraint operators are one of ["UNIQUE", "CLUSTER",
 "GROUP_BY"]. For additional information on using placement constraints see
 the [Constraints doc page]({{ site.baseurl }}/docs/constraints.html).
 
+##### acceptedResourceRoles
+
+Optional. A list of resource roles. Marathon considers only resource offers with roles in this list for launching
+tasks of this app. If you do not specify this, Marathon considers all resource offers except you override the
```
pyronicide: If you do not specify this, Marathon considers all resource offers that have been configured by the `--default_accepted_resource_roles` command line flag. Note that this defaults to "*" by default.

pyronicide:
I had one minor documentation update. This looks like it'll do exactly what we need.

docs/docs/rest-api.md
kolloch: I will try to rephrase this. Note that it does NOT default to "*" if you specify --mesos_role as well.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -61,9 +61,41 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
...
+  }
+
+  lazy val defaultAcceptedResourceRolesSet = defaultAcceptedResourceRoles.get.get
+
+  lazy val defaultAcceptedResourceRoles = opt[String]("default_accepted_resource_roles",
+    descr =
+      "Default for the defaultAcceptedResourceRoles attribute of all app definitions" +
+        " as a comma-separated list of strings." +
+        "This defaults to all roles for which this Marathon instance is configured to receive offers.",
+    default = Some(expectedResourceRoles.mkString(",")),
```
aquamatthias: This default value leads to an exception while generating the usage with --help.
```
> run --help
[info] Running mesosphere.marathon.Main --help
[2015-05-18 16:11:59,605] INFO Starting Marathon 0.8.2-SNAPSHOT (mesosphere.marathon.Main$:87)
[error] (run-main-1) org.rogach.scallop.exceptions.UnknownOption: Unknown option 'help'
org.rogach.scallop.exceptions.UnknownOption: Unknown option 'help'
	at org.rogach.scallop.Scallop$$anonfun$5.apply(Scallop.scala:125)
	at org.rogach.scallop.Scallop$$anonfun$5.apply(Scallop.scala:125)
	at scala.Option.getOrElse(Option.scala:120)
	at org.rogach.scallop.Scallop.parse(Scallop.scala:124)
	at org.rogach.scallop.Scallop.parse(Scallop.scala:85)
	at org.rogach.scallop.Scallop.org$rogach$scallop$Scallop$$parsed$lzycompute(Scallop.scala:187)
	at org.rogach.scallop.Scallop.org$rogach$scallop$Scallop$$parsed(Scallop.scala:175)
	at org.rogach.scallop.Scallop$$anonfun$get$5.apply(Scallop.scala:540)
	at org.rogach.scallop.Scallop$$anonfun$get$5.apply(Scallop.scala:537)
	at scala.Option.map(Option.scala:145)
	at org.rogach.scallop.Scallop.get(Scallop.scala:537)
	at org.rogach.scallop.ScallopConf$$anon$1$$anonfun$fn$1.apply(ScallopConf.scala:126)
	at org.rogach.scallop.ScallopConf$$anon$1$$anonfun$fn$1.apply(ScallopConf.scala:126)
	at org.rogach.scallop.ScallopOption.get(ScallopOption.scala:25)
	at mesosphere.marathon.MarathonConf$class.expectedResourceRoles(MarathonConf.scala:69)
	at mesosphere.marathon.MarathonApp$AllConf.expectedResourceRoles(Main.scala:76)
	at mesosphere.marathon.MarathonConf$$anonfun$defaultAcceptedResourceRoles$1.apply(MarathonConf.scala:81)
	at mesosphere.marathon.MarathonConf$$anonfun$defaultAcceptedResourceRoles$1.apply(MarathonConf.scala:81)

```

kolloch: Good that you noticed that.

Fixed now. I also added an integration test for that.

kolloch:
I should probably also adjust the JSON schema.

kolloch:
Rebased and added JSON schema extension.

Issue #1185, opened 2015-02-11
Serve json-schema for resources


ConnorDoyle:
See [json-schema.org](http://json-schema.org).

Especially important are:

- [ ] AppDefinition
- [ ] AppUpdate
- [ ] Group
- [ ] DeploymentStep
- [ ] Deployment
- [ ] Timestamp

ConnorDoyle:
Let's think about adding a new `/v2/schemas` collection.  A client could fetch the schema for AppDefinition via `GET /v2/schemas/app`, for example.

ConnorDoyle:
I've written up a json-schema document for the current Marathon app resource to use as a starting point: https://raw.githubusercontent.com/mesosphere/universe/master/repo/meta/schema/marathon-schema.json

drexin:
Cool, thanks! We should also use the schema to verify the output of the rest endpoints in our tests.

sttts:
@drexin I like that idea. This also removes my complaint that the schema is in parallel to the Reader and Writers for the api.

drexin:
Will move this to 0.9.0 for the missing schemas.

jsancio:
We should make sure that we check JSON input for POST /v2/apps and PUT /v2/apps/{id} (when it is a new object) against the schema.

We had a user miss-spelled the `labels` as `lables` and he didn't get any errors. This kind of programming errors are hard to debug. Checking the input JSON against the schema will help us return nicer errors.

ConnorDoyle:
In the interest of keeping things DRY, maybe we should also investigate automatic generation of JSON schemata for arbitrary `play.api.json.Reads` instances.

kolloch:
*nitpick*

@ConnorDoyle I think you do not really propose JSON-schema generation for _arbitrary_ `Reads` because you would have to analyse arbitrary code.

ConnorDoyle:
@kolloch yes, you're right of course.  The implementation could get complex really quickly.  I think the general goal is still valid though: somehow ensure the schema is in sync with parsers / validators.

PR #1529, opened 2015-05-13
add privileged and parameters definition to AppDefinition json schema

tamarrow:
DCOS-1345

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -171,6 +171,7 @@ object Dependencies {
     beanUtils % "compile",
     scallop % "compile",
     playJson % "compile",
+    jsonSchemaValidator % "compile",
 
     // test
     Test.scalatest % "test",
@@ -196,6 +197,7 @@ object Dependency {
     val Hadoop = "2.4.1"
     val Scallop = "0.9.5"
     val PlayJson = "2.3.7"
+    val JsonSchemaValidator = "2.2.6"
 
     // test deps versions
     val Mockito = "1.9.5"
@@ -224,6 +226,7 @@ object Dependency {
   val hadoopCommon = "org.apache.hadoop" % "hadoop-common" % V.Hadoop excludeAll(excludeMortbayJetty, excludeJavaxServlet)
   val beanUtils = "commons-beanutils" % "commons-beanutils" % "1.9.2"
   val scallop = "org.rogach" %% "scallop" % V.Scallop
+  val jsonSchemaValidator = "com.github.fge" % "json-schema-validator" % V.JsonSchemaValidator
 
   object Test {
     val scalatest = "org.scalatest" %% "scalatest" % V.ScalaTest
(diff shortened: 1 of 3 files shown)


ConnorDoyle:
Related: #1185

ConnorDoyle:
Although not necessarily required to accept this change, how do we validate conformance between the JSON schema and the actual JSON formatters used to render response bodies?

There's a Java project to do this called [json-schema-validator](https://github.com/fge/json-schema-validator}.  The README states it's looking for a new maintainer, but maybe that's not a huge problem if it's only used in the tests.

It seems like it wouldn't be too much work to render the JSON for a number of AppDefinition instances and ensure they validate against the in-tree schema.  This could increase confidence / awareness when the API changes, both intentionally and otherwise.

tamarrow:
In the future the best way is to go directly from the source to generate the schema. We could even have the api call generate the schema so we don't save state that could get outdated. And yes, change the tests to use this schema to validate. 

Where is the AppDefinition defined in Marathon? There seems to be a few sources of truth that are used in the codebase.

ConnorDoyle:
The JSON rendering is currently defined in `mesosphere.marathon.api.v2.json.Formats.scala`.

tamarrow:
@ConnorDoyle I added the json-schema-validator project and used it to verify the json schema with sample apps in the AppDefinitionValidator tests. Using it I found some fields we missed!

project/build.scala
```diff
@@ -224,6 +226,7 @@ object Dependency {
   val hadoopCommon = "org.apache.hadoop" % "hadoop-common" % V.Hadoop excludeAll(excludeMortbayJetty, excludeJavaxServlet)
   val beanUtils = "commons-beanutils" % "commons-beanutils" % "1.9.2"
   val scallop = "org.rogach" %% "scallop" % V.Scallop
+  val jsonSchemaValidator = "com.github.fge" % "json-schema-validator" % V.JsonSchemaValidator
```
ConnorDoyle: Verified Apache 2.0 compliant: https://github.com/fge/json-schema-validator/blob/master/LICENSE#L13

ConnorDoyle:
Awesome, thanks @tamarrow!
If it's not too much trouble could we also validate the JSON in https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/state/AppDefinitionTest.scala?

ConnorDoyle:
I already feel a lot more confident in the schema correctness though :sunglasses:

kolloch:
Looks good to me. Thank you.

ConnorDoyle:
@kolloch we were waiting to merge this until additional tests were added.  Please read the outstanding comments before merging.

ConnorDoyle:
The follow-up PR is here: #1547

kolloch:
Hi @ConnorDoyle ,

I actually wanted this as a basis for my schema extensions which is why I merged it. I think it was already an improvement over the prior state.

Anyways, sorry if I caused you any trouble and thanks for the work.

PR #1537, opened 2015-05-18
Update service-discovery-load-balancing.md

hansbogert:
reference to /etc/haproxy_marathon_bridge/marathons is wrong, should be dashes, i.e. /etc/haproxy-marathon-bridge/marathons


kolloch:
Thanks!

PR #1542, opened 2015-05-18
Fix #1204 by adding a search bar to the applications overview

### LINKED ISSUE #1204: Please add a search bar to the applications overview to filter the list of applications
It would be great to be able to filter the list of running applications.

The filter should look in available information in the app, but also tasks within the app for results. However, only display apps in the list (no tasks).

Way:
Regarding to the issue #1204 I've implemented a search bar to filter the list of applications on the applications overview.

This is the result of my "work day" at mesosphere ;)


aldipower:
@kolloch This is in progress.. ;)

src/main/resources/assets/js/components/AppListComponent.jsx
```diff
@@ -19,6 +19,26 @@ var AppListComponent = React.createClass({
     return this.props.collection;
   },
 
+  getCollection: function () {
+    var collection = this.props.collection;
+    var filter = this.props.filter;
+
+    // Apply filter
+    if (this.props.filter) {
+      collection = collection.filter(function (model) {
```
aldipower: Cool, pretty clear. :)

src/main/resources/assets/js/components/AppListFilterComponent.jsx
```diff
@@ -0,0 +1,54 @@
...
+      numFiltered: null
+    };
+  },
+
+  handleChange: function (event) {
+    this.props.onChange(event.target.value);
+  },
+
+  handleClear: function () {
+    this.props.value = "";
```
aldipower: Manipulating the props _inside_ the component isn't allowed. You should set a state here and working with states.
States are intended to use with changing data.

src/main/resources/assets/js/components/AppListFilterComponent.jsx
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    /* jshint trailing:false, quotmark:false, newcap:false */
+    /* jscs:disable disallowTrailingWhitespace, validateQuoteMarks, maximumLineLength */
+    return (
+      <form className="form-inline list-filter">
+        <div className="form-group">
+          <input
+            type="text"
+            value={this.props.value}
```
aldipower: I think it's
```
value={this.state.value}
```
here.

src/main/resources/assets/js/components/AppListFilterComponent.jsx
```diff
@@ -0,0 +1,54 @@
...
+            onChange={this.handleChange}
+            className={filterClassSet}
+            placeholder="Filter list" />
+
+            { this.props.value ? <span className="input-clear" onClick={this.handleClear}>x</span> : null }
```
aldipower:
```
this.state.value
```
Ans we don't use tenery operator by convention, could you move the condition out of the markuk and make a variable?

src/main/resources/assets/js/components/Marathon.jsx
```diff
@@ -461,6 +464,16 @@ var Marathon = React.createClass({
     /* jscs:enable disallowTrailingWhitespace, validateQuoteMarks, maximumLineLength */
   },
 
+  handleFilterChange: function (filterValue) {
+    this.setState({appListFilter: filterValue});
+  },
+
+  handleFilterApply: function (filtered) {
```
aldipower: Tidy up

src/main/resources/assets/js/components/Marathon.jsx
```diff
@@ -37,6 +38,8 @@ var Marathon = React.createClass({
       activeAppView: null,
       activeTabId: tabs[0].id,
       appVersionsFetchState: States.STATE_LOADING,
+      appListFilter: null,
+      appListFilterFiltered: null,
```
aldipower: Maybe here is a better naming necessary? Quite difficult to understand.

aldipower: I've just seen, that's not used anymore, so please remove. :)

aldipower:
In general this is a pretty good approach and could be merged after some small changes.
Use the state instead of the props for value changes inside the AppListFilterComponent.jsx .
Also the form input need some CSSing to match the screens supposed in the issue.
After that it could go in.
I really like it!

src/main/resources/assets/js/components/AppListComponent.jsx
```diff
@@ -19,6 +19,26 @@ var AppListComponent = React.createClass({
...
+        if (model.id && model.id.indexOf(filter) > -1) {
+          return model;
+        }
+      });
+
+      if (this.props.onFilter) {
+        this.props.onFilter(collection);
```
rcorral: This looks great!
Could you add `onFilter` to the `propTypes` definition?

aldipower:
Workday PR. Closing this now, can we reopened if necessary.

PR #1549, opened 2015-05-19
Fixes #937 - TaskFactory.newTask should return an Option of a case class

### LINKED ISSUE #937: Refactor MarathonScheduler.newTask
There is a long-standing TODO in the in-line comments to fix the return value of this method.  It currently returns an anonymous tuple wrapped in an option.

kolloch:
instead of a tuple.

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/TaskFactory.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/TaskFactory.scala
@@ -6,9 +6,15 @@ import org.apache.mesos.Protos.{ TaskInfo, Offer }
 
 /** Create tasks from app definitions and offers. */
 trait TaskFactory {
+  import TaskFactory.CreatedTask
+
   /**
     * Return the corresponding task if and only if the offer matches the app.
     */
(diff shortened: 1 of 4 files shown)


kolloch:
I do not really like the name `TaskFactory.CreatedTask`. If you have a better one, I am open for suggestions.

src/main/scala/mesosphere/marathon/tasks/TaskFactory.scala
```diff
@@ -6,9 +6,15 @@ import org.apache.mesos.Protos.{ TaskInfo, Offer }
...
+
   /**
     * Return the corresponding task if and only if the offer matches the app.
     */
-  def newTask(app: AppDefinition, offer: Offer): Option[(TaskInfo, MarathonTask)]
+  def newTask(app: AppDefinition, offer: Offer): Option[CreatedTask]
+}
+
+object TaskFactory {
+  case class CreatedTask(mesosTask: TaskInfo, marathonTask: MarathonTask)
```
ConnorDoyle: Can't think of any better name.

ConnorDoyle:
Thanks @kolloch

PR #1550, opened 2015-05-19
Fixes #1522 - Include disk resource share in TaskInfo

### LINKED ISSUE #1522: Disk resource quota not communicated to Mesos on task launch
We accept disk resources in our app definitions and also ensure that we only launch tasks for offers with sufficient disk resources. Unfortunately, we do not add the resource consumption to the TaskInfo resource that we communicate to mesos.

It is trivial to add this but we do not want to break things that worked before. Currently, our internal disk limit defaults to "0.0". If the user doesn't explicitly specify a disk limit, this is saved to the AppDefinition. Thus we cannot distinguish between the default value and a customer that set that explicitly to "0.0".

If we now enforce that limit with a new release, we break customer work flows that worked before.

Thus, I would suggest setting our default to the Mesos default. I would also insert a AppDefinition migration which sets the disk resource of AppDefinitions where it is "0.0" also to the Mesos default. If the user specified this limit explicitly, this is not so good but should also not be really harmful.

kolloch:


aquamatthias:
LGTM

PR #1552, opened 2015-05-20
Fixes #1520 - Satisfy port resources from any of the offered ranges

### LINKED ISSUE #1520: Marathon don't match correctly the resources for a defined role
*Using Marathon 0.8.2-RC2 and Mesos 0.22.1*

I started Marathon with  mesos_role setted as "fixed", a role to aggregate apps with fixed ports. I have another Marathon without mesos_role.
```--mesos_role fixed```

The mesos-slave has the following config:
```--resources=ports(fixed):[31999-32000, 8080-8080, 80-80, 8081-8081, 9000-9000];ports(*):[31000-31998]```

When I launch the app on Marathon, I have this log:
```
[2015-05-13 04:01:50,542] INFO All assigned host ports must be present in the resource offer.
Assigned: [PortMapping(8000,31999,10000,tcp), PortMapping(8080,8080,10001,tcp), PortMapping(80,80,10002,tcp)] (mesosphere.marathon.tasks.PortsMatcher:49)
[2015-05-13 04:01:50,542] WARN App ports are not available in the offer. (mesosphere.mesos.ResourceMatcher$:50)
[2015-05-13 04:01:50,543] INFO No matching offer for /bamboo-sd/v1 (need cpus=1.0, mem=1024.0, disk=0.0, ports=Vector(31999, 8080, 80)) : id {
  value: "20150510-200829-1403029514-5050-25155-O3746573"
}
framework_id {
  value: "20150508-182649-1403029514-5050-22921-0000"
}
slave_id {
  value: "20150510-200829-1403029514-5050-25155-S63"
}
hostname: "10.134.100.50"
resources {
  name: "ports"
  type: RANGES
  ranges {
    range {
      begin: 31000
      end: 31499
    }
    range {
      begin: 31503
      end: 31998
    }
  }
  role: "*"
}
resources {
  name: "cpus"
  type: SCALAR
  scalar {
    value: 7.6000000000000005
  }
  role: "*"
}
resources {
  name: "mem"
  type: SCALAR
  scalar {
    value: 14256.0
  }
  role: "*"
}
resources {
  name: "disk"
  type: SCALAR
  scalar {
    value: 25084.0
  }
  role: "*"
}
resources {
  name: "ports"
  type: RANGES
  ranges {
    range {
      begin: 31999
      end: 32000
    }
    range {
      begin: 8080
      end: 8080
    }
    range {
      begin: 80
      end: 80
    }
    range {
      begin: 8081
      end: 8081
    }
    range {
      begin: 9000
      end: 9000
    }
  }
  role: "fixed"
}
 (mesosphere.mesos.TaskBuilder:37)
```
My app needs ports 31999, 8080 and 80, the slave did the offer, but Marathon don't match.

### PR DESCRIPTION
**Pull request by kolloch**
We now also
* always log a reason if we cannot match all ports
* use randomized hosts ports for container port mappings
* make ResourceUtil work with the consumption of multiple roles with
  the same name and role

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -39,7 +39,11 @@ class PortsMatcher(
   /**
     * @return the resulting assigned (host) ports.
     */
-  def ports: Seq[Long] = portRanges.map(_.flatMap(_.ranges.flatMap(_.asScala()))).getOrElse(Nil)
+  def ports: Seq[Long] = for {
+    resource <- portRanges.getOrElse(Nil)
+    range <- resource.ranges
+    port <- range.asScala()
+  } yield port
 
   private[this] def portsWithRoles: Option[Seq[PortWithRole]] = {
     val portMappings: Option[Seq[Container.Docker.PortMapping]] =
@@ -59,11 +63,11 @@ class PortsMatcher(
         mappedPortRanges(mappings)
 
       case (appPorts, None) if app.requirePorts =>
-        findPortsInOfferer(appPorts, failLog = true)
+        findPortsInOffer(appPorts, failLog = true)
 
       case (appPorts, None) if !app.ports.contains(Integer.valueOf(0)) =>
-        // we try to use the user supplied ports if possible, fallback to dynamically assigned ports
-        findPortsInOfferer(appPorts, failLog = false).orElse(randomPorts(appPorts.size))
+        // We try to use the user supplied ports as host ports if possible, fallback to dynamically assigned ports.
+        findPortsInOffer(appPorts, failLog = false).orElse(randomPorts(appPorts.size))
 
       case (appPorts, None) =>
         randomPorts(appPorts.size)
@@ -71,10 +75,9 @@ class PortsMatcher(
   }
 
   /**
-    * Try to satisfy all app.ports exactly (not dynamically) if that's possible to do with one of the offered
-    * ranges.
-    */
-  private[this] def findPortsInOfferer(requiredPorts: Seq[Integer], failLog: Boolean): Option[Seq[PortWithRole]] = {
+   * Try to find supplied ports in offer. Returns `None` if not all ports were found.
+   */
+  private[this] def findPortsInOffer(requiredPorts: Seq[Integer], failLog: Boolean): Option[Seq[PortWithRole]] = {
     hasExpectedSizeOpt(expectedSize = requiredPorts.size) {
       requiredPorts.sorted.iterator.map { port =>
         offeredPortRanges.find(_.scalaRange.contains(port)).map { offeredRange =>
@@ -88,7 +91,7 @@ class PortsMatcher(
   }
 
   /**
-    * Choose ports from the available ports randomly.
+    * Choose random ports from offer.
     */
   private[this] def randomPorts(numberOfPorts: Int): Option[Seq[PortWithRole]] = {
     hasExpectedSizeOpt(expectedSize = numberOfPorts) {
@@ -100,9 +103,8 @@ class PortsMatcher(
   }
 
   /**
-    * Returns Some(ports) if all zero-valued docker host-ports
-    * can be assigned to some port from the resource offer, AND all other
-    * (non-zero-valued) docker host-ports are available in the resource offer.
+    * Try to find all non-zero host ports in offer and use random ports from the offer for dynamic host ports (=0).
+    * Return `None` if not all host ports could be assigned this way.
     */
   private[this] def mappedPortRanges(mappings: Seq[PortMapping]): Option[Seq[PortWithRole]] = {
     hasExpectedSizeOpt(expectedSize = mappings.size) {
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -5,7 +5,6 @@ import mesosphere.marathon.state.Container.Docker.PortMapping
 import mesosphere.mesos.protos
 import mesosphere.mesos.protos.{ RangesResource, Resource }
 import mesosphere.util.Logging
-import org.apache.mesos.Protos
 import org.apache.mesos.Protos.Offer
 
 import scala.annotation.tailrec
@@ -140,17 +139,15 @@ class PortsMatcher(
     if (ports.size == expectedSize) Some(ports) else None
   }
 
-  private[this] lazy val offeredPortsResources: Seq[Protos.Resource] = offer.getResourcesList.asScala
-    .filter(resource => acceptedResourceRoles(resource.getRole))
-    .filter(_.getName == Resource.PORTS)
-    .to[Seq]
-
-  private[this] lazy val offeredPortRanges: Seq[PortRange] =
-    for {
-      resource <- offeredPortsResources
+  private[this] lazy val offeredPortRanges: Seq[PortRange] = {
+    val portRangeIter = for {
+      resource <- offer.getResourcesList.asScala.iterator
+      if acceptedResourceRoles(resource.getRole) && resource.getName == Resource.PORTS
       rangeInResource <- resource.getRanges.getRangeList.asScala
       range = Range.inclusive(rangeInResource.getBegin.toInt, rangeInResource.getEnd.toInt)
     } yield PortRange(resource.getRole, range)
+    portRangeIter.to[Seq]
+  }
 
   private[this] def shuffledAvailablePorts: Iterator[PortWithRole] =
     PortWithRole.randomPortsFromRanges(random)(offeredPortRanges)
@@ -243,15 +240,15 @@ object PortsMatcher {
         (rangeIdx, startPortIdx - startPortIdxOfCurrentRange)
       }
 
-      val shuffled = Random.shuffle(offeredPortRanges).toVector
-      val startPortIdx = Random.nextInt(numberOfOfferedPorts)
+      val shuffled = random.shuffle(offeredPortRanges).toVector
+      val startPortIdx = random.nextInt(numberOfOfferedPorts)
 
       val (rangeIdx, portInRangeIdx) = findStartPort(shuffled, startPortIdx)
       val startRangeOrig = shuffled(rangeIdx)
 
       val startRange = startRangeOrig.withoutNPorts(portInRangeIdx)
-      val afterStartRange = shuffled.slice(rangeIdx + 1, shuffled.length).flatMap(_.portsWithRolesIterator)
-      val beforeStartRange = shuffled.slice(0, rangeIdx).flatMap(_.portsWithRolesIterator)
+      val afterStartRange = shuffled.slice(rangeIdx + 1, shuffled.length).iterator.flatMap(_.portsWithRolesIterator)
+      val beforeStartRange = shuffled.slice(0, rangeIdx).iterator.flatMap(_.portsWithRolesIterator)
       val endRange = startRangeOrig.firstNPorts(portInRangeIdx)
 
       startRange ++ afterStartRange ++ beforeStartRange ++ endRange
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -75,12 +75,12 @@ class PortsMatcher(
   }
 
   /**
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala [hunk 1]
```diff
@@ -1,44 +1,47 @@
...
+
+  /**
+    * @return true if and only if the port requirements could be fulfilled by the given offer.
+    */
+  def matches: Boolean = portRanges.isDefined
+
+  /**
+    * @return the resulting assigned (host) ports.
+    */
+  def ports: Seq[Long] = portRanges.map(_.flatMap(_.ranges.flatMap(_.asScala()))).getOrElse(Nil)
```
aquamatthias: Probably the shortest expression possible, which is good, but also very hard to understand (especially with asScala I think of a java sequence).
Can you make that more readable?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +50,213 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
+      val startRange = startRangeOrig.withoutNPorts(portInRangeIdx)
+      val afterStartRange = shuffled.slice(rangeIdx + 1, shuffled.length).flatMap(_.portsWithRolesIterator)
+      val beforeStartRange = shuffled.slice(0, rangeIdx).flatMap(_.portsWithRolesIterator)
+      val endRange = startRangeOrig.firstNPorts(portInRangeIdx)
+
+      startRange ++ afterStartRange ++ beforeStartRange ++ endRange
+    }
+  }
+
+  case class PortRange(role: String, range: Protos.Value.Range) {
```
aquamatthias: Would it be better to make `range` a scala range with an apply in the PortRange object, that accepts the protobuf?

kolloch: yes, definitely! Thanks for the suggestion.

kolloch:
Rebased and included suggested improvements.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
+      *
+      * The implementation idea:
+      *
+      * * Randomize the order of the offered ranges.
+      * * Now treat the ports contained in the ranges as one long sequence of ports.
+      * * We randomly choose an index where we want to start assigning dynamic ports in that sequence. When
+      *   we hit the last offered port with wrap around and start offering the ports at the beginning
+      *   of the sequence up to (excluding) the port index we started at.
+      */
+    def randomPortsFromRanges(random: Random = Random)(offeredPortRanges: Seq[PortRange]): Iterator[PortWithRole] = {
```
aquamatthias: You specify a random instance, but it is not used (used is always the scala Random object). Also this could be a good fit for an implicit parameter.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
-        val maxOffset = (range.getEnd - range.getBegin - app.ports.length + 2).toInt
-        val firstPort = range.getBegin + Random.nextInt(maxOffset)
-        val rangeProto = protos.Range(firstPort, firstPort + app.ports.length - 1)
-        return Some(
-          RangesResource(Resource.PORTS, Seq(rangeProto), role)
-        )
-      }
+  private[this] def randomPorts(numberOfPorts: Int): Option[Seq[PortWithRole]] = {
+    hasExpectedSizeOpt(expectedSize = numberOfPorts) {
+      shuffledAvailablePorts.take(numberOfPorts).to[Seq]
```
aquamatthias: This will convert all offeredPorts into an Iterable[PortWithRole] and than only a certain number of ports is taken. Can we use a Stream for this?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
+            process(Some(nextPort.toRange), next.tail)
+        }
+      }
+      process(ranges.headOption.map(_.toRange), ranges.tail)
+
+      builder.result()
+    }
+
+    /**
+      * We want to avoid reusing the same dynamic ports. Thus we shuffle the ports to make it less likely.
```
aquamatthias: That is a lot of code for a simple task, which also generates a lot of objects. (Consider using a Stream which knows all the ports that have been already taken and randomly chooses the next one)

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
-          val role = offeredPortRanges.find(_.scalaRange.contains(pm.hostPort)) match {
-            case Some(PortRange(role, _)) => role
+          offeredPortRanges.find(_.range.contains(pm.hostPort)) match {
+            case Some(PortRange(role, _)) =>
+              Some(PortWithRole(role, pm.hostPort))
             case None =>
-              throw PortResourceException(
-                s"Cannot find range with port ${pm.hostPort}"
-              )
+              log.info(s"Cannot find range with host port ${pm.hostPort} for app [${app.id}]")
```
aquamatthias: Could we short circuit the computation here?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
         case PortMapping(containerPort, hostPort, servicePort, protocol) if hostPort == 0 =>
-          if (!availablePorts.hasNext) throw PortResourceException(
-            s"Insufficient ports in offer for app [${app.id}]"
-          )
-          val portWithRole = availablePorts.next()
-          PortMappingWithRole(
-            portWithRole.role,
-            PortMapping(containerPort, portWithRole.port, servicePort, protocol)
-          )
+          if (!availablePortsWithoutStaticHostPorts.hasNext) {
```
aquamatthias: can we put the two cases into separate def's?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
+              Some(PortWithRole(role, pm.hostPort))
             case None =>
-              throw PortResourceException(
-                s"Cannot find range with port ${pm.hostPort}"
-              )
+              log.info(s"Cannot find range with host port ${pm.hostPort} for app [${app.id}]")
+              None
           }
-          PortMappingWithRole(role, pm)
+      }.takeWhile(_.isDefined).flatten.to[Seq]
```
aquamatthias: Why takeWhile?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +54,215 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
+    }
+  }
+
+  private[this] def hasExpectedSizeOpt[T](expectedSize: Int)(ports: Seq[T]): Option[Seq[T]] = {
+    if (ports.size == expectedSize) Some(ports) else None
+  }
+
+  private[this] lazy val offeredPortsResources: Seq[Protos.Resource] = offer.getResourcesList.asScala
```
aquamatthias: It is only used in offeredPortRanges. Can this be a local val to that function?

aquamatthias:
@kolloch some helper functions are difficult to read/understand. I made some comments and suggestions for improvement (please read as pure advise).

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
kolloch: It is already lazy. It is an Iterator (which should have less overhead than a Stream).

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
kolloch: Good catch! I don't like the implicit parameter suggestion, though.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
kolloch: Short-lived objects are usually no problem for the GC.

I don't understand your suggestion about streams.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
kolloch: to shurt-circuit the computation ;)

kolloch:
I rebased the PR and made some improvements. The complexity of the random port assignment remains, unfortunately.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +53,224 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
-      if (range.getEnd - range.getBegin + 1 >= app.ports.length) {
-        val maxOffset = (range.getEnd - range.getBegin - app.ports.length + 2).toInt
-        val firstPort = range.getBegin + Random.nextInt(maxOffset)
-        val rangeProto = protos.Range(firstPort, firstPort + app.ports.length - 1)
-        return Some(
-          RangesResource(Resource.PORTS, Seq(rangeProto), role)
-        )
-      }
+  private[this] def randomPorts(numberOfPorts: Int): Option[Seq[PortWithRole]] = {
+    takeEnoughPortsOrNone(expectedSize = numberOfPorts) {
```
aquamatthias: More a style question: I would have expected, that takeEnoughPortsOrNone gets a function parameter (curly braces is usually used for passing a function), but it is a normal parameter. Do you have an opinion on that?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -47,145 +53,224 @@ class PortsMatcher(app: AppDefinition, offer: Offer, acceptedResourceRoles: Set[
...
+      *
+      * The implementation idea:
+      *
+      * * Randomize the order of the offered ranges.
+      * * Now treat the ports contained in the ranges as one long sequence of ports.
+      * * We randomly choose an index where we want to start assigning dynamic ports in that sequence. When
+      *   we hit the last offered port with wrap around and start offering the ports at the beginning
+      *   of the sequence up to (excluding) the port index we started at.
+      */
+    def lazyRandomPortsFromRanges(rand: Random = Random)(offeredPortRanges: Seq[PortRange]): Iterator[PortWithRole] = {
```
aquamatthias: Why 2 parameter lists?

aquamatthias:
@kolloch 
This PR is a huge improvement with respect to the current state :+1: 
I have added 2 minor styling questions - let's discuss shortly.
Other than that, I will merge that right away.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
kolloch: Summarizing the discussion. I see the following alternatives:

* `({...})` instead of `{...}`
* Making the parameter unnecessarily a function
* calculating the parameter before the method invocation and pass per variable name

I don't like anything better than the current choice.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
kolloch: First parameter list is conceptually for an initialization state. You could wire the rand via dependency injection and just directly use the function with the second parameter list. The second parameter list is for parameters that are specific to the invocation.

kolloch:
Improved the laziness test.

aquamatthias:
Cool :+1:

PR #1561, opened 2015-05-22
Fixes #1560 - Versioning recent changes in the API

### LINKED ISSUE #1560: Versioning for recent additions in docs
Currently, we are documenting new features without mentioning at which release we introduce the feature.

Then, after a Marathon release we are updating our information.

This is problematic for (at least) two reasons:

* If we improve the existing documentation, the improvements get only published to the gh-pages branch after the next release and thus the public web page.

* Users who do not use the most recent release, do not see clearly if a feature is supported or not.

Looking through all recent documentation changes and add versioning information. Then we can publish the updated documentation immediately.

In future, we version all additions when adding them and update our documentation more frequently.

kolloch:
such that we can publish the documentation even before publishing
the associated Marathon releases.


aquamatthias:
Great addition!

PR #1575, opened 2015-05-27
Fixes #1545 - Make it easy to run integration tests

### LINKED ISSUE #1545: Make it easy to run integration/scale tests via Docker
At Mesosphere we run the integration tests via docker on our CI system. We would like to allow everyone with a docker installation to run these tests easily on their development machine.

What is more, we reuse the regular Dockerfile for integration testing which results in duplicate/non-optimal builds. Thus we would like to provide a tailored Dockerfile basis for integration testing and a simple way to run integration tests/scaling tests with this file.

kolloch:
via a script and docker.

Adjust test settings:

* to be sensitive to whether sbt formatting is enabled
* to include durations
* to print stack traces in error summary

Make scaling test slightly more agressive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
+bin/run-tests-config.sh
 .idea
 *.iml
+docker-volumes
 target
 .reviewboardrc
 dependency-reduced-pom.xml
--- a/Dockerfile.build-base
+++ b/Dockerfile.build-base
@@ -4,7 +4,6 @@ RUN apt-get update && \
     apt-get upgrade -y && \
     apt-get install --no-install-recommends -y \
     default-jdk \
-    scala \
     curl
 
 RUN curl -SsL -O http://dl.bintray.com/sbt/debian/sbt-0.13.5.deb && \
@@ -12,10 +11,3 @@ RUN curl -SsL -O http://dl.bintray.com/sbt/debian/sbt-0.13.5.deb && \
 
 COPY . /marathon
 WORKDIR /marathon
-
-RUN sbt -Dsbt.log.format=false assembly && \
-    mv $(find target -name 'marathon-assembly-*.jar' | sort | tail -1) ./ && \
-    rm -rf target/* ~/.sbt ~/.ivy2 && \
-    mv marathon-assembly-*.jar target
-
-ENTRYPOINT ["./bin/start"]
--- a/bin/run-tests.sh
+++ b/bin/run-tests.sh
@@ -1,0 +1,132 @@
+#!/bin/bash
+
+#
+# Run all tests including integration tests via docker.
+#
+# The script places some files into target/docker-build-volumes.
+# If you keep it around between builds your can greatly speed up resolving artifacts
+# but on the other hand you do not test if all artifacts are still resolvable.
+#
+# You can configure this script by copying run-tests-config-template.sh to run-tests-config.sh
+# and adjusting it to your needs.
+#
+# Usage:
+#  ./run-tests.sh $0 [<unique build id>]
+#
+#  The build id is used as part of the docker image version/container names.
+#
+
+
+# be verbose and print all commands, better for debugging
+# set -x
+
+PROJECT_DIR=`dirname $0`/..
+PROJECT_DIR=$(cd "$PROJECT_DIR"; pwd)
+
+if [ -r "$PROJECT_DIR/bin/run-tests-config.sh" ]; then
+    . "$PROJECT_DIR/bin/run-tests-config.sh"
+fi
+
+function fatal {
+    echo "======== FATAL ERROR ===================================================================================" >&2
+    echo "$*" >&2
+    echo "========================================================================================================" >&2
+    exit 2
+}
+
+BUILD_ID=${1-test}
+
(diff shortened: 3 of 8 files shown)


kolloch:
I already enabled downloading the test scripts from this PR to TeamCity.

So this patch kind off builds itself.

.gitignore
```diff
@@ -1,5 +1,6 @@
 .idea
 *.iml
+docker-volumes
```
aquamatthias: Why not put this under target?

Dockerfile.build-base
```diff
@@ -0,0 +1,13 @@
+FROM mesosphere/mesos:0.22.1-1.0.ubuntu1404
```
aquamatthias: build-base sounds like, it will be reused/refined somewhere?

bin/run-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# be verbose and print all commands, better for debugging
+set -x
+
+PROJECT_DIR=`dirname $0`/..
+PROJECT_DIR=$(cd "$PROJECT_DIR"; pwd)
+
+if [ -z "$1" ]; then
+    echo "Usage:" >&2
+    echo "$0 <unique build id without crazy characters or spaces>" >&2
```
aquamatthias: Would it make sense to create a buildId, if it is omitted?
As I started it the first time, my question was: where do I get the buildId....

bin/run-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+if ! docker build -t marathon-buildbase:$BUILD_ID -f "$PROJECT_DIR/Dockerfile.build-base" "$PROJECT_DIR"; then
+    fatal "Build for the buildbase failed" >&2
+fi
+
+docker run \
+    --rm=true \
+    --name marathon-itests-$BUILD_ID \
+    -v "$BUILD_VOLUME_DIR/sbt:/root/.sbt" \
+    -v "$BUILD_VOLUME_DIR/ivy2:/root/.ivy2" \
+    -v "$BUILD_VOLUME_DIR/targets/main:/marathon/target" \
```
aquamatthias: For local tests: can we change that in a way, that only changes need to get recompiled?

bin/run-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if ! docker build -t marathon-buildbase:$BUILD_ID -f "$PROJECT_DIR/Dockerfile.build-base" "$PROJECT_DIR"; then
+    fatal "Build for the buildbase failed" >&2
+fi
+
+docker run \
+    --rm=true \
+    --name marathon-itests-$BUILD_ID \
+    -v "$BUILD_VOLUME_DIR/sbt:/root/.sbt" \
+    -v "$BUILD_VOLUME_DIR/ivy2:/root/.ivy2" \
```
aquamatthias: Can we make it configurable, so that I can use ~/.ivy2 and ~/.sbt

project/build.scala
```diff
@@ -47,12 +47,32 @@ object MarathonBuild extends Build {
...
+   * U - unformatted mode
+   * I - show reminder of failed and canceled tests without stack traces
+   * T - show reminder of failed and canceled tests with short stack traces
+   * G - show reminder of failed and canceled tests with full stack traces
+   * K - exclude TestCanceled events from reminder
+   *
+   * http://scalatest.org/user_guide/using_the_runner
+   */
+  lazy val formattingTestArg = Tests.Argument(if (withFormatting) "-eDST" else "-eWUDST")
+
```
aquamatthias: Great comment!

bin/run-tests.sh
```diff
@@ -0,0 +1,63 @@
...
+
+cleanup
+trap cleanup EXIT
+
+if ! docker build -t marathon-buildbase:$BUILD_ID -f "$PROJECT_DIR/Dockerfile.build-base" "$PROJECT_DIR"; then
+    fatal "Build for the buildbase failed" >&2
+fi
+
+docker run \
+    --rm=true \
```
bobrik: `--rm` works without args

.gitignore
kolloch: Because I wanted it not to be gone after an sbt clean.

We talked about it and I leave this the default but allow configuration.

Dockerfile.build-base
kolloch: Currently, no. It would potentially make sense to use this for building the assembly of the official image as well but as far as I know two phase builds are not supported on dockerhub,

bin/run-tests.sh
kolloch: I tried to document it and we just use "test" as default.

bin/run-tests.sh
kolloch: I made everything configurable in my local version.

bin/run-tests.sh
kolloch: I made everything configurable in my local version.

project/build.scala
kolloch: Thanks!

kolloch:
Made stuff configurable.

aquamatthias:
This is cool!

PR #1582, opened 2015-05-28
Move from State abstraction to direct zk access with scala based Futures

### LINKED ISSUE #1539: JVM-only library for storing state (in Zookeeper)
### Proposal

We should implement storing the state in Zookeeper without relying on any native libraries and without relying on the Mesos release cycle to fix issues. 

We can either just provide another implementation for the existing [State](https://github.com/apache/mesos/blob/master/src/java/src/org/apache/mesos/state/State.java)  interface or provide an improved interface with a similar purpose. Whatever we choose as an interface in our code, we must keep the current data format or provide a migration.

### Problem

Currently, Marathon uses the [org.apache.mesos.state.State](https://github.com/apache/mesos/blob/master/src/java/src/org/apache/mesos/state/State.java) abstraction to persist data. It uses the provided ZookeeperState implementation which uses native code.

This implementation used to cause crashes (#834) repeatedly. These crashes have been solved for Mesos 0.22.1 but at the same time made more likely for older versions. 

It also leads to comments like these https://github.com/mesosphere/marathon/blob/e94d68503ab728b7e92e72ee6484aea364b0b327/src/main/scala/mesosphere/marathon/state/MarathonStore.scala#L124-L130 because the native code doesn't correctly create `ExecutionExceptions` with proper causes.

We also think that the timeout of the ZookeeperState-abstraction doesn't always work. We have seen hanging Zookeeper requests in production.

Mesos is a nice resource scheduler but it is not good framework for holding state.

### Affected/Related Issues

#834 should be solved for older Mesos-Library-Versions as well because they are not used anymore.

When implementing this, we should probably also allow authentication (see #1336).

This issue potentially obsoletes "#706 Add a config flag for selecting the Mesos state implementation" if we use another interface for state abstraction.

### Outlook

In the future, we potentially want to move away from Zookeeper to a more scalable solution for state keeping. In the mean time we have to provide something that works reliably across Mesos versions.

aquamatthias:
@kolloch Please review.

Discussion points:
-package of Store abstraction
-Future timeouts 

Fixes #1539


src/main/proto/marathon.proto
```diff
@@ -174,3 +174,12 @@ message TaskFailure {
   required string version = 6;
   required string timestamp = 7;
 }
+
+// Describes a state entry, a versioned (via a UUID) key/value pair.
+// Copied from libmesos (src/messages/state.pro) to ensure
+// compatibility with ZooKeeperState from libmesos.
```
kolloch: Nice comment! in mesos.proto they use `/**... */` though.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,127 @@
...
+}
+object ZKData {
+  def apply(bytes: Array[Byte]): ZKData = {
+    val proto = Protos.ZKStoreEntry.parseFrom(bytes)
+    new ZKData(proto.getName, UUIDUtil.uuid(proto.getUuid.toByteArray), proto.getValue.toByteArray)
+  }
+}
+
+object ZKStore {
+  implicit class Twitter2Scala[T](val twitterF: TWFuture[T]) extends AnyVal {
```
kolloch: I am somewhat an implicit conversion hater.

If I want to understand .asScala above, I need to check out the code and use my IDE. Otherwise I'll have a hard time finding the relevant code.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,127 @@
...
+    mapFutureResult(node.exists(), s"Can not delete entity $key") { data =>
+      node.delete(data.stat.getVersion).map(n => ZKEntity(n, ZKData(key, UUID.randomUUID())))
+    }
+  }
+
+  override def allIds(): Future[Seq[ID]] = {
+    root.getChildren().map(_.children.map(_.name)).asScala
+  }
+
+  private def mapFutureResult[A, B](value: TWFuture[A], errorMessage: String)(fn: A => TWFuture[B]): Future[B] = {
```
kolloch: maybe make errorMessage a function? It should only be evaluated in the error case.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,127 @@
...
+  /**
+    * This will store a previously fetched entity.
+    * The entity will be either created or updated, depending on the read state.
+    * @return Some value, if the store operation is successful otherwise None
+    */
+  override def save(entity: PersistentEntity): Future[ZKEntity] = {
+    val zk = zkEntity(entity)
+    def update(nodeData: ZNode.Data): Future[ZKEntity] = {
+      val future: TWFuture[Data] = zk.node.setData(zk.data.toProto.toByteArray, nodeData.stat.getVersion)
+      mapFutureResult(future, s"Can not update entity $entity") { data =>
```
kolloch: I would map to a scala future as early as possible an then use the well known methods on the scala future.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,127 @@
...
+    zk.read.map(update).getOrElse(create())
+  }
+
+  /**
+    * Delete an entry with given identifier.
+    */
+  override def delete(key: ID): Future[ZKEntity] = {
+    val node = root(key)
+    require(node.parent == root, "Nested paths are not supported!")
+    mapFutureResult(node.exists(), s"Can not delete entity $key") { data =>
```
kolloch: why do an exists first? Can't we just catch errors of the delete? Saves round-trips for us.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  private[this] val log = Logger.getLogger(getClass)
+
+  val root = createNode(rootNode)
+
+  /**
+    * Fetch data and return entity.
+    * The entity is returned also if it is not found in zk, since it is needed for the store operation.
+    */
+  override def load(key: ID): Future[ZKEntity] = {
```
kolloch: Does it cause a lot of changes if we change that to Future[Option[ZKEntity]]?

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
```diff
@@ -0,0 +1,45 @@
...
+import mesosphere.util.state.{ PersistentEntity, PersistentStore }
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+/**
+  * Simple in memory store implementation.
+  * This is intended only for tests - do not use in production!
+  * @param ec the execution context to use.
+  */
+class InMemoryStore(implicit val ec: ExecutionContext = ThreadPoolContext.context) extends PersistentStore {
```
kolloch: I would be nice if InMemory store also checked the version on updates.

src/main/scala/mesosphere/marathon/state/PersistenceStore.scala
```diff
@@ -4,16 +4,17 @@ import scala.concurrent.Future
 
 trait PersistenceStore[T] {
 
+  type Read = () => T //returns deserialized T value.
```
kolloch: I think I would understand the intention better if this was called `Deserialize`

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
aquamatthias: Added that ability.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
aquamatthias: Understand the point. I think/hope that explicit implicit makes the code more readable.

aquamatthias:
@kolloch 

-ZKStore: moved from TwitterFuture to ScalaFuture early. Subsequent handling relies on standard types.
-PersistentStore.load will return an Option[Entity] and not an Entity.
 that implies, that create and update has to be handled explicitly
 I added the functionality and rewrote the code, that relied on the implicit entity creation.
-EntityStore.{store, modify} will return no Option, but either succeed or fail
 Rewrote the functionality, that handled the Option case (which generally was considerer an error)
-MarathonStore: clean up of the existing functionality

There is the ability to switch the backend store via the hidden flag --internal_store_backend.
I used this for testing backward compatibility - not sure, if we should remove it.

I removed the zookeeper authentication flags/authentication.
The PR is already too big. Can add this functionality, once this is landed.

src/main/scala/mesosphere/util/state/mesos/MesosStateStore.scala
```diff
@@ -0,0 +1,49 @@
...
+
+class MesosStateStore(state: State, timeout: Duration) extends PersistentStore {
+
+  implicit val timeoutDuration = Timeout(timeout)
+  implicit val ec = ThreadPoolContext.context
+  import mesosphere.util.BackToTheFuture.futureToFuture
+
+  override def load(key: ID): Future[Option[PersistentEntity]] = {
+    futureToFuture(state.fetch(key)).map { variable =>
+      if (variable.value().length > 0) Some(MesosStateEntity(key, variable)) else None
```
kolloch: Maybe `.nonEmpty` is better? (Also see below)

Maybe this also deserves an explanation.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  override def allIds(): Future[Seq[ID]] = {
+    root.getChildren().asScala
+      .map(_.children.map(_.name))
+      .recover(exceptionTransform("Can not list all identifiers"))
+  }
+
+  private def exceptionTransform[T](errorMessage: String): PartialFunction[Throwable, T] = {
+    case ex: KeeperException => throw new StoreCommandFailedException(errorMessage, ex)
+    case ex: Throwable       => throw ex
```
kolloch: `NonFatal(ex)` better?

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+    * Fetch data and return entity.
+    * The entity is returned also if it is not found in zk, since it is needed for the store operation.
+    */
+  override def load(key: ID): Future[Option[ZKEntity]] = {
+    val node = root(key)
+    require(node.parent == root, "Nested paths are not supported!")
+    node.getData().asScala
+      .map{ data => Some(ZKEntity(node, ZKData(data.bytes), Some(data.stat.getVersion))) }
```
kolloch: Being picky. Space before `{` as below?

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  val root = createPath(rootNode)
+
+  /**
+    * Fetch data and return entity.
+    * The entity is returned also if it is not found in zk, since it is needed for the store operation.
+    */
+  override def load(key: ID): Future[Option[ZKEntity]] = {
+    val node = root(key)
+    require(node.parent == root, "Nested paths are not supported!")
```
kolloch: have the node path in the error message

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+    node.getData().asScala
+      .map{ data => Some(ZKEntity(node, ZKData(data.bytes), Some(data.stat.getVersion))) }
+      .recover { case ex: NoNodeException => None }
+      .recover(exceptionTransform(s"Could not load key $key"))
+  }
+
+  override def create(key: ID, content: Array[Byte]): Future[ZKEntity] = {
+    val node = root(key)
+    require(node.parent == root, "Nested paths are not supported!")
```
kolloch: node path in the error message

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  private def zkEntity(entity: PersistentEntity): ZKEntity = {
+    entity match {
+      case zk: ZKEntity => zk
+      case _            => throw new IllegalArgumentException("Can not handle this kind of entity")
+    }
+  }
+
+  private def createPath(path: ZNode): ZNode = {
```
kolloch: This uses blocking which is fine (in my opinion) in initialization but shouldn't be used elsewhere.

Thus I would name this method `createPathBlocking` or something like this.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  private def exceptionTransform[T](errorMessage: String): PartialFunction[Throwable, T] = {
+    case ex: KeeperException => throw new StoreCommandFailedException(errorMessage, ex)
+    case ex: Throwable       => throw ex
+  }
+
+  private def zkEntity(entity: PersistentEntity): ZKEntity = {
+    entity match {
+      case zk: ZKEntity => zk
+      case _            => throw new IllegalArgumentException("Can not handle this kind of entity")
```
kolloch: Add type of the argument to the error message.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+      zk.node.setData(zk.data.toProto.toByteArray, version).asScala
+        .map { data => zk.copy(version = Some(data.stat.getVersion)) }
+        .recover(exceptionTransform(s"Can not update entity $entity"))
+    }
+    def create(): Future[ZKEntity] = {
+      zk.node.create(zk.data.toProto.toByteArray).asScala
+        .flatMap { _.getData().map{ data => zk.copy(version = Some(data.stat.getVersion)) }.asScala }
+        .recover(exceptionTransform(s"Can not update entity $entity"))
+    }
+    zk.version.map(update).getOrElse(create())
```
kolloch: The decision whether to save(= update) or to create (= insert) is now actually handled in `MarathonStore.modify`. Thus this code can go. Or did I misunderstand something?

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -0,0 +1,149 @@
...
+      .map(n => ZKEntity(n, data, Some(0))) //first version after create is 0
+      .recover(exceptionTransform(s"Can not create entity $key"))
+  }
+
+  /**
+    * This will store a previously fetched entity.
+    * The entity will be either created or updated, depending on the read state.
+    * @return Some value, if the store operation is successful otherwise None
+    */
+  override def save(entity: PersistentEntity): Future[ZKEntity] = {
```
kolloch: Maybe rename to `update`, see comment below.

kolloch:
Review still in progress. Looks nice so far! Notes for myself:

* Looked at ZKStore
* Looked at InMemoryStore
* Ran integration tests with mesos_zk, they failed, investigate.

What does native in `NativeConnector` mean?

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -153,7 +152,7 @@ class MarathonSchedulerActor(
       val origSender = sender()
       withLockFor(appId) {
         val promise = Promise[Unit]()
-        val tasksToKill = taskIds.flatMap(taskTracker.fetchTask(appId, _)).toSet
+        val tasksToKill = taskIds.flatMap(taskTracker.fetchTask(appId, _))
```
kolloch: Just a note to myself that this is OK ;)

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -416,7 +415,7 @@ class SchedulerActions(
   def stopApp(driver: SchedulerDriver, app: AppDefinition): Future[_] = {
     appRepository.expunge(app.id).map { successes =>
       if (!successes.forall(_ == true)) {
```
kolloch: It is not your code but WTF?? they are always all true but errors are not handled correctly

(Not part of your PR)...

src/main/scala/mesosphere/marathon/ZookeeperConf.scala
```diff
@@ -12,23 +13,19 @@ trait ZookeeperConf extends ScallopConf {
   private val zkNode = """[^/]+"""
   private val zkURLPattern = s"""^zk://(?:$userAndPass@)?($hostAndPort(?:,$hostAndPort)*)(/$zkNode(?:/$zkNode)*)$$""".r
 
-  @Deprecated
-  lazy val zooKeeperHostString = opt[String]("zk_hosts",
-    descr = "[DEPRECATED use zk] The list of ZooKeeper servers for storing state",
-    default = Some("localhost:2181"))
-
-  @Deprecated
-  lazy val zooKeeperPath = opt[String]("zk_state",
```
kolloch: You should remove this from the Docs as well (see rest-api.md and developing-vm.md)

kolloch: And add a comment in changelog.md just in case.

src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
```diff
@@ -49,7 +50,7 @@ class HttpEventModule(httpEventConfiguration: HttpEventConfiguration) extends Ab
   @Named(HttpEventModule.SubscribersKeeperActor)
   def provideSubscribersKeeperActor(conf: HttpEventConfiguration,
                                     system: ActorSystem,
-                                    store: MarathonStore[EventSubscribers]): ActorRef = {
+                                    store: EntityStore[EventSubscribers]): ActorRef = {
```
kolloch: I like it :)

src/main/scala/mesosphere/marathon/event/http/SubscribersKeeperActor.scala
```diff
@@ -2,43 +2,38 @@ package mesosphere.marathon.event.http
...
-}
 import mesosphere.marathon.event.http.SubscribersKeeperActor._
-import mesosphere.marathon.state.MarathonStore
+import mesosphere.marathon.event.{ MarathonSubscriptionEvent, Subscribe, Unsubscribe }
+import mesosphere.marathon.state.EntityStore
+
 import scala.concurrent.Future
 
-class SubscribersKeeperActor(val store: MarathonStore[EventSubscribers]) extends Actor with ActorLogging {
+class SubscribersKeeperActor(val store: EntityStore[EventSubscribers]) extends Actor with ActorLogging {
```
kolloch: Your changes made me think that MarathonStore can actually be package private. What do you think?

src/main/scala/mesosphere/marathon/state/EntityStore.scala
```diff
@@ -0,0 +1,20 @@
+package mesosphere.marathon.state
+
+import scala.concurrent.Future
+
+trait EntityStore[T] {
+
+  type Deserialize = () => T //returns deserialized T value.
+  type Update = Deserialize => T //Update function Gets an Read and returns the (modified) T
+
+  def fetch(key: String): Future[Option[T]]
```
kolloch: Some docs?

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -1,139 +1,75 @@
...
 class MarathonStore[S <: MarathonState[_, S]](
-  conf: MarathonConf,
-  state: State,
-  registry: MetricRegistry,
-  newState: () => S,
-  prefix: String = "app:")(
-    implicit val timeout: BackToTheFuture.Timeout = BackToTheFuture.Timeout(Duration(conf.marathonStoreTimeout(),
-      MILLISECONDS)))
-    extends PersistenceStore[S] {
+    val store: PersistentStore,
```
kolloch: Thank you for not extending PersistenceStore!!!

kolloch: But why is store a val?

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -1,139 +1,75 @@
...
-          case e: InvalidProtocolBufferException =>
-            if (variable.value.nonEmpty) {
-              log.error(s"Failed to read $key, could not deserialize data (${variable.value.length} bytes).", e)
-            }
-            None
-        }
-      case None =>
-        throw new StorageException(s"Failed to read $key, does not exist, should have been created automatically.")
+    store.load(prefix + key).map {
+      _.map{ entity =>
```
kolloch: This is so much better than before!

format nitpicking, space after `map`?

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -1,139 +1,75 @@
...
+        bytesRead.update(entity.bytes.length)
+        stateFromBytes(entity.bytes)
+      }
```
kolloch: A recover that adds context and rethrows an exception would be nice.

For example, if serialization fails, it would be very nice to know for what kind of entity and for which ID.

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -1,139 +1,75 @@
...
-        log.warn(s"exception while calling $getClass.names() for prefix $prefix", e)
-        Iterator.empty
     }
   }
 
   private def stateFromBytes(bytes: Array[Byte]): S = {
     newState().mergeFromProto(bytes)
   }
+
+  private def withLock[T](key: String)(future: => Future[T]): Future[T] = {
```
kolloch: Really nice that you have extracted that.

Optimally, LockManager would return a LockManger object containing this method, though.

src/main/scala/mesosphere/util/state/mesos/MesosStateStore.scala
```diff
@@ -0,0 +1,49 @@
...
+    case _                       => throw new IllegalArgumentException("Can not handle this kind of entity")
+  }
+
+  override def delete(key: ID): Future[PersistentEntity] = {
+    futureToFuture(state.fetch(key)).flatMap { variable =>
+      futureToFuture(state.expunge(variable)).map(_ => MesosStateEntity(key, variable))
+    }
+  }
+
+  override def allIds(): Future[Seq[ID]] = futureToFuture(state.names()).map(_.asScala.toSeq)
```
kolloch: Doesn't that through an NPE if the Zookeeper is empty currently? See the code you removed from `MarathonStore.names()` which is used in the Migration code. Or am I missing something?

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -1,139 +1,75 @@
...
-      case None => throw new StorageException(s"Failed to read $key")
-    }
-
-    res onComplete { _ =>
-      lock.release()
-    }
-
-    res
+  def expunge(key: String): Future[Boolean] = withLock(key) {
+    store.delete(prefix + key) map { _ => true }
```
kolloch: I know it is ugly and I do not really want to burden you with correcting all this in this PR but to keep the old semantics: 

Don't you not at least need a `recover { case NonFatal(e) => log something; false }`?

But better actually distinguish between errors and missing nodes here.

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -0,0 +1,71 @@
...
+    */
+  def save(entity: PersistentEntity): Future[PersistentEntity]
+
+  /**
+    * Expunge the entity with given id.
+    * @param key the key identifier
+    * @return either the entity or a failure.
+    *         In case of a storage specific failure, a StoreCommandFailedException is thrown.
+    */
+  def delete(key: ID): Future[PersistentEntity]
```
kolloch: I guess this should return a `Future[Option[PersistentEntity]]` to distinguish deletes from noops (because the entity is already gone).

Also see the caller `expunge` which would benefit from that.

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -0,0 +1,71 @@
...
+    * @param entity the entity to store
+    * @return either the entity or a failure.
+    *         In case of a storage specific failure, a StoreCommandFailedException is thrown.
+    */
+  def save(entity: PersistentEntity): Future[PersistentEntity]
+
+  /**
+    * Expunge the entity with given id.
+    * @param key the key identifier
+    * @return either the entity or a failure.
```
kolloch: the entity prior to the delete

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -0,0 +1,71 @@
...
+  /**
+    * Fetch entity with given key identifier. If the item does not exists, None is returned.
+    * @param key the identifier of this id
+    * @return either the entity or None.
+    *         In case of a storage specific failure, a StoreCommandFailedException is thrown.
+    */
+  def load(key: ID): Future[Option[PersistentEntity]]
+
+  /**
+    * Create a new entity with given id and content.
```
kolloch: Document: Results in error if the entity already exists.

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -0,0 +1,71 @@
...
+    */
+  def create(key: ID, content: Array[Byte]): Future[PersistentEntity]
+
+  /**
+    * Store the given entity.
+    * @param entity the entity to store
+    * @return either the entity or a failure.
+    *         In case of a storage specific failure, a StoreCommandFailedException is thrown.
+    */
+  def save(entity: PersistentEntity): Future[PersistentEntity]
```
kolloch: Document: Optimistic locking. Save will fail, if the given entity doesn't correspond to current version.

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
kolloch: I added a note to `PersistentStore.delete` to return a `Future[Option[...]]` so you can map it to true false "correctly".

src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
```diff
@@ -0,0 +1,58 @@
...
+import scala.concurrent.{ Future, Await, ExecutionContext }
+
+/**
+  * Utility class for keeping track of a framework ID in Mesos state.
+  *
+  * @param state State implementation
+  * @param key The key to store the framework ID under
+  */
+
+class FrameworkIdUtil(val state: PersistentStore, val key: String = "frameworkId") {
```
kolloch: Somehow I do not find the deleted FrameworkIdUtil in the changes, strange. 

Why do you use `val` here? And wouldn't it be to use `MarathonStore` instead of `PersistentStore`?

src/main/scala/mesosphere/util/state/PersistentStore.scala
kolloch: Maybe call `update` instead of `save` to make it clear that we deal with existing entities?

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -159,4 +159,12 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
     descr = "Mesos Authentication Secret",
     noshort = true
   )
+
+  //Internal settings, that are not intended for external use
+  lazy val internalStoreBackend = opt[String]("internal_store_backend",
+    descr = "The backend storage system to use. One of zk, mesos_zk",
+    hidden = true,
+    validate = Set("zk", "mesos_zk").contains,
```
kolloch: It would be occasionally useful to also allow `in_memory`. For example, to test whether persistence is our bottleneck.

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  var entities = Map.empty[ID, InMemoryEntity]
```
kolloch: It would be nice to allow concurrency (for testing).

src/test/scala/mesosphere/util/state/zk/ZKStoreTest.scala
```diff
@@ -0,0 +1,124 @@
...
+import mesosphere.marathon.integration.setup.{ IntegrationFunSuite, IntegrationTestConfig, ProcessKeeper }
+import mesosphere.util.state.PersistentEntity
+import org.apache.commons.io.FileUtils
+import org.apache.zookeeper.ZooDefs.Ids
+import org.scalatest._
+import org.scalatest.concurrent.ScalaFutures._
+
+import scala.collection.JavaConverters._
+
+class ZKStoreTest extends IntegrationFunSuite with BeforeAndAfterAll with BeforeAndAfter with Matchers {
```
kolloch: Nice Test!

It would be great if you created an `AbstractStoreTest` or something similar and then instantiated with your three different store implementations.

kolloch: Of course, some of the tests are backend specific but many are general.

kolloch:
Now, I have looked at everything at least once. It's really an improvement to what we have right now and I can't wait to merge the final PR.

Here some comments in addition the inline comments.

It would be cool, if we could run the integration tests against all three possible storage backends on teamcity.

It would also be nice to test the compatibility of the storage formats for both Zookeeper backends.

Does the Mesos Zookeeper state abstraction contain compatibility code for older formats? Then we would need to support older formats, too, or document that we do not.

src/main/scala/mesosphere/util/state/mesos/MesosStateStore.scala
aquamatthias: I thought you were an implicit hater ... :)
I replace that with a small helper function.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
aquamatthias: Good catch - it is better.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
aquamatthias: Removed the create version.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
aquamatthias: Agree, it should be CRUD: Create/Update/Delete

src/main/scala/mesosphere/marathon/event/http/SubscribersKeeperActor.scala
aquamatthias: The guice binder (MarathonModule) needs access to this class.
It would need more actions to accomplish this, so I will leave that in current state.

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
aquamatthias: Mmmh - I removed the val and it still compiles, so I don't think it is needed.

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
aquamatthias: Changed it. Is there a way to force this via scalariform?

kolloch: I don't know. Do you? :)

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
aquamatthias: Like the idea. Still think, there is certain functionality missing in the standard Future companion, perhaps we can collect such functionality into such one.

src/main/scala/mesosphere/util/state/mesos/MesosStateStore.scala
aquamatthias: Thanks for remembering - that one is really ugly...

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
aquamatthias: I don't like the option return type here. The future already has the notion of success/failure, which is somehow translated to true/false or would be some/none in this case. Either Boolean as also Option say: you have to handle the success of the operation yourself. I will make it Future[Unit], which signals: delete was ok or not. What do you think?

src/main/scala/mesosphere/util/state/PersistentStore.scala
aquamatthias: See my comment on expunge.

src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
aquamatthias: This class lives in mesos-utils and depend on the State abstraction.
I basically copied this class and made the needed adaptions.
Yes: it should also use MarathonStore - ok I will change this also.

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
aquamatthias: Now use concurrent.TrieMap.

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
kolloch: I don't agree.

You typically want to handle the case "the node does not exist" differently from "I couldn't connect to zookeeper".

A compromise would be to define an Exception which is only thrown in the "the node does not exist case".

src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
kolloch: Ah, ok, thanks!

I know this PR gets large.

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
kolloch: :+1:

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
aquamatthias: The question is, do we need an exists method somewhere. We could put that on the PersistentEntity if needed. The notion of delete(entity) = Success(None) reminds me on Some(null)...

aquamatthias: Since there is no type to wrap into an option, I will take Future[Boolean]
Success: The entity was deleted(true) or is not existent(false) => in any case it is gone
Failure: The entity could not get deleted

kolloch: Sounds good. The only thing that is missing then is the ability to take a look at the old instance (since we retrieve it anyway). But since we do not need that...

Besides: `Success(None)` is nothing like `Some(null)` but may make a lot of sense. `Future[Option[...]]` is common and useful.

src/main/scala/mesosphere/marathon/ZookeeperConf.scala
aquamatthias: Done. Thanks for the hint.

src/main/scala/mesosphere/marathon/state/EntityStore.scala
aquamatthias: Added class docs.

src/main/scala/mesosphere/util/state/PersistentStore.scala
aquamatthias: Added a note on that.

aquamatthias:
Incorporated the suggested changes - they were very helpful, Thanks!

1) I investigated some time for Future Timeouts - it probably make sense to add that before release
2) I ran the integration tests against the 3 possible storage backends. All were green. I am not 100% sure, if we should add that to TeamCity. The build time would get trippled and I hope we use only one option in the future.
3) Some scale testing is still missing - will start to investigate .
4) What do you think to the package question?

kolloch:
Cool, nearly there.

Besides my comments, I would really like to see some compatibility testing (if the storage format is compatible).

aquamatthias:
Cool.
Just added tests for reading,writing to/from zkStore <-> mesosState. See ZkStoreTest.

aquamatthias:
Rebased and squashed.

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+    * Get the content bytes of this entity.
+    */
+  def bytes: Array[Byte]
+
+  /**
+    * Mutate this entity with given bytes.
+    */
+  def mutate(bytes: Array[Byte]): PersistentEntity
```
kolloch: different name and description: This method does not actually modify the instance itself but returns a new instance.

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+    * The identifier of this entity.
+    */
+  def id: String
+
+  /**
+    * Get the content bytes of this entity.
+    */
+  def bytes: Array[Byte]
```
kolloch: Maybe use `IndexedSeq[Byte]` instead of `Array[Byte]` to avoid exporting mutable methods.

http://stackoverflow.com/questions/3350421/what-is-most-efficient-way-to-do-immutable-byte-arrays-in-scala

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
```diff
@@ -0,0 +1,53 @@
...
+    entities.get(key) match {
+      case Some(value) => Future.successful(entities.remove(key).isDefined)
+      case None        => Future.successful(false)
+    }
+  }
+
+  override def allIds(): Future[Seq[ID]] = Future.successful(entities.keySet.toSeq)
+}
+
+case class InMemoryEntity(id: String, version: Int)(val bytes: Array[Byte] = Array.empty) extends PersistentEntity {
```
kolloch: This syntax confused me at least: I didn't know that equality/hash is restricted to the first parameter class
of a case class.

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
```diff
@@ -0,0 +1,53 @@
...
+  override def mutate(bytes: Array[Byte]): PersistentEntity = copy()(bytes)
+  def incrementVersion: InMemoryEntity = copy(version = version + 1)(bytes)
```
kolloch: I like `withIncrementedVersion` better.

src/main/scala/mesosphere/util/state/PersistentStore.scala
aquamatthias: Renamed the method to withNewContent.

src/main/scala/mesosphere/util/state/PersistentStore.scala
aquamatthias: Scala uses a WrappedArray by default, so that should be not to expensive.

src/main/scala/mesosphere/util/state/memory/InMemoryStore.scala
aquamatthias: I reverted the change to a simple case class again.

aquamatthias:
Incorporated your suggestions.

kolloch:
I restarted the test but the PR seems to need a rebase anyways.

aquamatthias:
Rebased.

aquamatthias:
Rebased and squashed.

changelog.md
```diff
@@ -36,6 +36,16 @@ The new Marathon now provides an
 where you receive all events conveniently as
 [Server Sent Events](http://www.w3schools.com/html/html5_serversentevents.asp).
 
+#### Abstraction for Persistent Storage added with Zookeeper access directly in the JVM
```
kolloch: Capitalization not consistent with other head lines.

changelog.md
```diff
@@ -106,12 +116,17 @@ parameters.
 WARNING: If you use the `http_endpoints` parameter with multiple space
 separated URLs, you will need to migrate to the comma-separated format.
 
-#### Do not delay task launches anymore as a result of failed health checks
-
+####
```
kolloch: Line break by accident?

changelog.md
```diff
@@ -106,12 +116,17 @@ parameters.
...
 Marathon uses an exponential back off strategy to delay further task launches after task failures. This should
 prevent keeping the cluster busy with task launches which are set up to fail anyway. The delay was also increased
 when health checks failed leading to delayed recovery.
 
+#### Removed deprecated command line arguments `zk_hosts` and `zk_state`
+
+The command line arguments `zk_hosts` and `zk_state` were deprecated for some time and got removed in this version.
+Use the `--zk` command line argument to define the zookeeper connection string. 
+
 Since health checks typically (depending on configuration) take a while to determine that a task is unhealthy,
```
kolloch: You interrupted the section with your new section.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -228,6 +227,16 @@ class MarathonSchedulerService @Inject() (
     driver = None
   }
 
+  def isLeader: Boolean = leader.get()
```
kolloch: Nope. you readded this by accident, I guess.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -345,7 +354,7 @@ class MarathonSchedulerService @Inject() (
     timer.schedule(
       new TimerTask {
         def run() {
-          if (leader.get()) {
```
kolloch: nope, keep `leader.get()`

src/main/scala/mesosphere/util/LockManager.scala
```diff
@@ -3,8 +3,38 @@ package mesosphere.util
...
+
+  /**
+    * Execute the future with a lock on the given key.
+    * @param key the key to use for locking.
+    * @param future the future to execute, once the lock is available.
+    * @param ec the execution context, the future gets executed on.
+    * @tparam T the resulting type of the future.
+    * @return A new future, that gets executed, once the lock is available.
+    */
+  def executeSequential[T](key: String)(future: => Future[T])(implicit ec: ExecutionContext): Future[T]
```
kolloch: sequentially?

kolloch:
Some merge errors, otherwise fine.

changelog.md
aquamatthias: Persistent Storage -> persistent storage

changelog.md
aquamatthias: Yes. reverted.

changelog.md
aquamatthias: Moved after the section.

kolloch:
:+1: !!! LGTM !!!

Waiting for tests.

PR #1590, opened 2015-06-01
Fixes #1569 - Split http_endpoints parameter

### LINKED ISSUE #1569: http_endpoints not being split on ","
Corresponding with the [configuration shown in the documentation](https://mesosphere.github.io/marathon/docs/event-bus.html), I expected all 3 of my endpoints to be hit.  Instead marathon is failing to split on the commas and just sends only 1 request.

$ cat /etc/marathon/conf/http_endpoints:

    http://mesos-primary3a:8000/v1/eventBus,http://mesos-primary1a:8000/v1/eventBus,http://mesos-primary2a:8000/v1/eventBus

Resulting HTTP log:

    method=POST
    url=/v1/eventBus,http://mesos-primary1a:8000/v1/eventBus,http://mesos-primary2a:8000/v1/eventBus
    remoteAddr=172.1.2.180:33178
    referer=

src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala:

```scala
trait HttpEventConfiguration extends ScallopConf {

  lazy val httpEventEndpoints = opt[List[String]]("http_endpoints",
    descr = "The URLs of the event endpoints master",
    required = false,
    noshort = true)
}
```

It doesn't look like the value is being split appropriately here.

gkleiman:
This change is inspired on code from MarathonConf. I'm still a scala newbie, so don't be afraid to bash me ;-).

### DIFF
--- a/src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
+++ b/src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
@@ -19,10 +19,13 @@ import mesosphere.marathon.MarathonConf
 
 trait HttpEventConfiguration extends ScallopConf {
 
-  lazy val httpEventEndpoints = opt[List[String]]("http_endpoints",
-    descr = "The URLs of the event endpoints master",
+  lazy val httpEventEndpoints = opt[String]("http_endpoints",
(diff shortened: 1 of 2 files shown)


kolloch:
:100: Very nice!

PR #1597, opened 2015-06-03
GroupsResource use play json

### LINKED ISSUE #1259: Reject null bodies in REST API
When sending an empty body to an endpoint, instead of throwing a parse exception, Jackson returns `null` which breaks some of our endpoints.

aquamatthias:
Add tests for GroupUpdate/Group Json rendering.
Add Formatter for GroupUpdate.
Fixes #1259 
@kolloch Please review


src/test/scala/mesosphere/marathon/api/v2/json/FormatsTest.scala
```diff
@@ -0,0 +1,110 @@
...
+import mesosphere.marathon.api.v2.GroupUpdate
+import mesosphere.marathon.state.{ Timestamp, AppDefinition, Group }
+import play.api.libs.json.{ JsNull, JsValue, Format, Json }
+import mesosphere.marathon.state.PathId._
+import org.scalatest.Matchers._
+
+class FormatsTest extends MarathonSpec {
+  import Formats._
+
+  test("Can read GroupUpdate json") {
```
kolloch: You extended the tests majorly. That's great!

But why did you choose a more general test name? It really only tests the GroupFormat, does it?

I actually do like more specific test suits that don't get too long.

kolloch:
Looks nice!

500 is not really the best error code for parse errors, though:

```http
PUT /v2/groups/test HTTP/1.1
Accept: application/json
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 0
Content-Type: application/json
Host: localhost:8080
User-Agent: HTTPie/0.9.2



HTTP/1.1 500 Internal Server Error
Cache-Control: no-cache, no-store, must-revalidate
Content-Type: application/json
Expires: 0
Pragma: no-cache
Server: Jetty(8.1.15.v20140411)
Transfer-Encoding: chunked

{
    "message": "No content to map due to end-of-input\n at [Source: [B@57202b60; line: 1, column: 1]"
}
```

src/test/scala/mesosphere/marathon/api/v2/json/FormatsTest.scala
aquamatthias: Old habit: If I test functionality in class XXX, than I use the name XXXTest. That is also aligned in most IDE's (IntelliJ: Cmd+Shift T).
I am aware of the fact, that the tests for all formats is not exhaustive...

aquamatthias:
Map this kind of exception to an HTTP 400 code.

src/test/scala/mesosphere/marathon/api/v2/json/FormatsTest.scala
aquamatthias: This format is defined in the DeploymentFormats trait. I renamed the test accordingly.

aquamatthias:
Rename FormatsTests -> DeploymentFormatsTest

Issue #1540, opened 2015-05-18
A marathon instance should never proxy to itself


kolloch:
For convenience, non-leader Marathon instances proxy all REST API requests to the current leader. This is implemented in `LeaderProxyFilter`.

Occasionally, it seems to happen that a Marathon instance proxies requests to itself. We should prevent that.

kolloch:
This might or might not relate to #828.

kolloch:
Sometimes the Marathon host name isn't set correctly. This might also lead to this behaviour. To quote David Kesler from the mailing list:

It looks like the problem is that the instances are all registering themselves as 'localhost.localdomain' rather than their correct externally available hostnames.  Thus when you visit a non-master instance in your browser, it tries to proxy the request to 'localhost' which, since it's not the master, just keeps proxying the request (to itself).

You can verify that this is the issue by visiting the other members of the cluster.  One of them should work properly.  If this is the case, you can set the hostname marathon registers with via the '--hostname' parameter when starting marathon.

Issue #828, opened 2014-11-25
rewrite proxy filter


drexin:
Currently the proxy filter has several issues:

- it changes headers
- it uses blocking IO without timeouts (see #1541)
- it handles requests as if the current instance was the master, when it does not know the real master (see #1540)

kolloch:
Case in point, an unresponsive Marathon instance in our test cluster showed a lot of these stack traces in a thread dump:

```
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: "qtp1656012274-16757" prio=10 tid=0x00007f6c24029800 nid=0x2
223 runnable [0x00007f6d5e6e4000]
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: java.lang.Thread.State: RUNNABLE
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at java.net.SocketInputStream.socketRead0(Native Method)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at java.net.SocketInputStream.read(SocketInputStream.java:152)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at java.net.SocketInputStream.read(SocketInputStream.java:122)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at java.io.BufferedInputStream.fill(BufferedInputStream.java:235)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at java.io.BufferedInputStream.read1(BufferedInputStream.java:275)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at java.io.BufferedInputStream.read(BufferedInputStream.java:334)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: - locked <0x00000000e1880b18> (a java.io.BufferedInputStream)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at sun.net.www.http.HttpClient.parseHTTPHeader(HttpClient.java:687)
Mar 23 18:34:22 srv4.hw.ca1.mesosphere.com marathon[9935]: at sun.net.www.http.HttpClient.parseHTTP(HttpClient.java:633)
[...]

PR #1598, opened 2015-06-03
Fixes #1541, #1540 - LeaderProxyFilter, do not proxy to self, timeouts

### LINKED ISSUE #1541: Proxying Marathon requests should use timeouts
For convenience, non-leader Marathon instances proxy all REST API requests to the current leader. This is implemented in `LeaderProxyFilter`.

The current implementation uses blocking IO without timeouts. We should at least introduce timeouts.

(This was proposed in #828 but I would like to split the issue into something more specific.)

kolloch:
also:

* prevent excessive calling of MarathonSchedulerService.getLeaderData
  since that results in synchronous Zookeeper calls (now we call
  it once per request instead of four times)
* provide a metric for getLeaderData
* wait for consistent leadership data if necessary

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonLeaderInfo.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonLeaderInfo.scala
@@ -1,0 +1,33 @@
+package mesosphere.marathon
+
+import java.util.concurrent.atomic.AtomicBoolean
+import javax.inject.{ Named, Inject }
+
+import com.codahale.metrics.MetricRegistry
+import com.twitter.common.zookeeper.Candidate
+import mesosphere.marathon.api.LeaderInfo
+import mesosphere.util.TimerUtils
+import mesosphere.util.TimerUtils.ScalaTimer
+
+class MarathonLeaderInfo @Inject() (
+    @Named(ModuleNames.NAMED_CANDIDATE) candidate: Option[Candidate],
+    @Named(ModuleNames.NAMED_LEADER_ATOMIC_BOOLEAN) leader: AtomicBoolean,
+    metrics: MarathonLeaderInfoMetrics) extends LeaderInfo {
+
+  /** Query whether we are the current leader. This should be cheap. */
+  override def elected: Boolean = leader.get()
+
+  override def currentLeaderHostPort(): Option[String] = metrics.getLeaderDataTimer {
+    candidate.flatMap { c =>
+      Option(c.getLeaderData.orNull()).map { data =>
+        new String(data, "UTF-8")
+      }
+    }
+  }
+}
+
+class MarathonLeaderInfoMetrics @Inject() (metrics: MetricRegistry) {
+  val getLeaderDataTimer: ScalaTimer =
+    TimerUtils.timer(metrics, getClass, "current-leader-host-port")
+}
+
--- a/src/main/scala/mesosphere/marathon/MarathonModule.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonModule.scala
@@ -15,6 +15,7 @@ import com.google.inject.name.Names
 import com.twitter.common.base.Supplier
 import com.twitter.common.zookeeper.{ Candidate, CandidateImpl, Group => ZGroup, ZooKeeperClient }
 import mesosphere.chaos.http.HttpConf
+import mesosphere.marathon.api.LeaderInfo
 import mesosphere.marathon.event.EventModule
 import mesosphere.marathon.event.http.HttpEventStreamActor
 import mesosphere.marathon.health.{ HealthCheckManager, MarathonHealthCheckManager }
@@ -32,6 +33,8 @@ import scala.util.control.NonFatal
 
 object ModuleNames {
   final val NAMED_CANDIDATE = "CANDIDATE"
+  final val NAMED_HOST_PORT = "HOST_PORT"
+
   final val NAMED_LEADER_ATOMIC_BOOLEAN = "LEADER_ATOMIC_BOOLEAN"
   final val NAMED_SERVER_SET_PATH = "SERVER_SET_PATH"
   final val NAMED_SERIALIZE_GROUP_UPDATES = "SERIALIZE_GROUP_UPDATES"
@@ -48,14 +51,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
     bind(classOf[MarathonConf]).toInstance(conf)
     bind(classOf[HttpConf]).toInstance(http)
     bind(classOf[ZooKeeperClient]).toInstance(zk)
+    bind(classOf[LeaderProxyConf]).toInstance(conf)
 
     // needs to be eager to break circular dependencies
     bind(classOf[SchedulerCallbacks]).to(classOf[SchedulerCallbacksServiceAdapter]).asEagerSingleton()
 
     bind(classOf[MarathonSchedulerDriverHolder]).in(Scopes.SINGLETON)
     bind(classOf[SchedulerDriverFactory]).to(classOf[MesosSchedulerDriverFactory]).in(Scopes.SINGLETON)
+    bind(classOf[MarathonLeaderInfoMetrics]).in(Scopes.SINGLETON)
     bind(classOf[MarathonScheduler]).in(Scopes.SINGLETON)
     bind(classOf[MarathonSchedulerService]).in(Scopes.SINGLETON)
+    bind(classOf[LeaderInfo]).to(classOf[MarathonLeaderInfo]).in(Scopes.SINGLETON)
     bind(classOf[TaskTracker]).in(Scopes.SINGLETON)
     bind(classOf[TaskQueue]).in(Scopes.SINGLETON)
     bind(classOf[TaskFactory]).to(classOf[DefaultTaskFactory]).in(Scopes.SINGLETON)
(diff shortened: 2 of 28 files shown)


kolloch:
I thought about some more possible improvements. So I reassign to myself.

kolloch:
Ok, this is hopefully it.

kolloch:
If you review this, please also think about whether we should keep https://github.com/mesosphere/marathon/issues/828 or not.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -1,137 +1,263 @@
...
+    val leaderConnection: HttpURLConnection = url.openConnection().asInstanceOf[HttpURLConnection]
+
+    leaderConnection.setConnectTimeout(leaderProxyConf.leaderProxyConnectionTimeout())
+    leaderConnection.setReadTimeout(leaderProxyConf.leaderProxyReadTimeout())
+
+    // getHeaderNames() and getHeaders() are known to return null, see:
+    //http://docs.oracle.com/javaee/6/api/javax/servlet/http/HttpServletRequest.html#getHeaders(java.lang.String)
+    val names = Option(request.getHeaderNames).map(_.asScala).getOrElse(Nil)
+    for {
+      name <- names if !name.equalsIgnoreCase("host") && !name.equalsIgnoreCase("connection")
```
aquamatthias: Host is usually preserved when proxying. Is there any reason for this?

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -1,137 +1,263 @@
...
-              proxy.getErrorStream().close()
-          }
-          finally {
-            responseOutputStream.close()
-          }
+  private[this] val viaValue: String = s"1.1 $myHostPort"
+
+  override def forward(url: URL, request: HttpServletRequest, response: HttpServletResponse): Unit = {
+
+    val viaOpt = Option(request.getHeaders(HEADER_VIA)).map(_.asScala.toVector)
```
aquamatthias: That check really was missing - thanks!

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -1,137 +1,263 @@
...
-          val names = Option(request.getHeaderNames).map(_.asScala).getOrElse(Nil)
+      false
+    }
 
-          for (name <- names) {
-            val values = request.getHeaders(name)
-            if (values != null) {
-              proxy.setRequestProperty(name, values.asScala.mkString(","))
-            }
+    (rawRequest, rawResponse) match {
```
aquamatthias: There is a lot of nested if/else statements. Can you define some helper methods and make the main control flow more obvious?

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -143,20 +149,23 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
...
-      val candidate = new CandidateImpl(new ZGroup(zk, ZooDefs.Ids.OPEN_ACL_UNSAFE,
-        conf.zooKeeperLeaderPath),
+      log.info("Registering in Zookeeper with hostPort:" + hostPort)
+      val candidate = new CandidateImpl(new ZGroup(zk, ZooDefs.Ids.OPEN_ACL_UNSAFE, conf.zooKeeperLeaderPath),
         new Supplier[Array[Byte]] {
           def get(): Array[Byte] = {
-            //host:port
-            "%s:%d".format(conf.hostname(),
-              http.httpPort()).getBytes
+            hostPort.getBytes
```
aquamatthias: This uses platform specific encoding - define encoding.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -143,20 +149,23 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
...
   @Singleton
-  def provideCandidate(zk: ZooKeeperClient): Option[Candidate] = {
+  def provideCandidate(zk: ZooKeeperClient, @Named(ModuleNames.NAMED_HOST_PORT) hostPort: String): Option[Candidate] = {
     if (conf.highlyAvailable()) {
-      log.info("Registering in Zookeeper with hostname:"
-        + conf.hostname())
-      val candidate = new CandidateImpl(new ZGroup(zk, ZooDefs.Ids.OPEN_ACL_UNSAFE,
-        conf.zooKeeperLeaderPath),
+      log.info("Registering in Zookeeper with hostPort:" + hostPort)
+      val candidate = new CandidateImpl(new ZGroup(zk, ZooDefs.Ids.OPEN_ACL_UNSAFE, conf.zooKeeperLeaderPath),
```
aquamatthias: Side note: call parentheses either remove from the calling side or add to the caller side.

src/main/scala/mesosphere/marathon/MarathonLeaderInfo.scala
```diff
@@ -0,0 +1,33 @@
...
+    @Named(ModuleNames.NAMED_LEADER_ATOMIC_BOOLEAN) leader: AtomicBoolean,
+    metrics: MarathonLeaderInfoMetrics) extends LeaderInfo {
+
+  /** Query whether we are the current leader. This should be cheap. */
+  override def elected: Boolean = leader.get()
+
+  override def currentLeaderHostPort(): Option[String] = metrics.getLeaderDataTimer {
+    candidate.flatMap { c =>
+      Option(c.getLeaderData.orNull()).map { data =>
+        new String(data)
```
aquamatthias: This method uses system dependent encoding and should be specified.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -1,137 +1,263 @@
...
+    log.debug("request properties {}", leaderConnection.getRequestProperties.asScala)
+
+    method match {
+      case "GET" | "HEAD" | "DELETE" =>
+        leaderConnection.setDoOutput(false)
+      case _ =>
+        leaderConnection.setDoOutput(true)
+
+        val proxyOutputStream = leaderConnection.getOutputStream
+        try {
```
aquamatthias: There is logic in the IO package for that.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -1,137 +1,263 @@
...
+    finally {
+      Try(response.getOutputStream.close())
+      Try(leaderConnection.getInputStream.close())
+      Try(leaderConnection.getErrorStream.close())
     }
   }
 
-  def destroy() {
-    //NO-OP
+  private[this] def copy(inputOrNull: InputStream, outputOrNull: OutputStream): Unit = {
```
aquamatthias: There is an IO trait with such functionality. Please use/adapt existing functionality.

src/test/scala/mesosphere/marathon/api/LeaderProxyFilterTest.scala
```diff
@@ -0,0 +1,191 @@
...
+    verifyNoMoreInteractions(leaderInfo, forwarder, request, response, chain)
+    leaderInfo = null
+    forwarder = null
+    filter = null
+    request = null
+    response = null
+    chain = null
+  }
+
+  test("we are leader") {
```
aquamatthias: Really like the tests!

aquamatthias:
Looks good!
Points that crossed my mind while reading:
-for GET/HEAD/DELETE we could also think about redirecting - most http clients are capable of this out of the box.
-is there a chance, to deny proxy requests when a certain limit of outstanding requests is reached?
-in the case of non leader: is the UI served directly or from the leader?

src/main/scala/mesosphere/marathon/MarathonLeaderInfo.scala
kolloch: If I change that, customer systems might break. Should I change that nevertheless?

src/main/scala/mesosphere/marathon/MarathonModule.scala
kolloch: See above

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
kolloch: Well, it's reverse proxying. So I guess it shouldn't be preserved.

See for example:
http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#proxypreservehost

It should work either way.

kolloch: Nginx also redefines host and connection by default.

http://nginx.com/resources/admin-guide/reverse-proxy/

bobrik: Most of the time you use `proxy_set_header Host $host;` if upstream depends on host header somehow.

kolloch:
Some answers:

* Redirects: I would stick with proxying instead of redirecting for now. Redirecting might cause issues with some clients and can cause problems with incorrect firewalling.
* Max number of concurrent requests: The code uses blocking IO and thus there is natural back-pressure. In combination with timeouts that should be fine.
* UI: Yes, the UI is directly served from the contacted instance.

src/main/scala/mesosphere/marathon/MarathonLeaderInfo.scala
aquamatthias: I would specify that, because It could prevent problems.

kolloch:
Ok, I updated the code to include an explicit encoding for the candidate info.

aquamatthias:
We need to handle EventStreams specifically: the current leader can change over time. The connections should be closed and reestablished.

aquamatthias:
See #1625 as reference to the missing stream handling.

PR #1623, opened 2015-06-08
Fixes #1572 - Improve metrics name generation

### LINKED ISSUE #1572: Remove `$$EnhancerByGuice$$...` from class names in metrics
When you use `--enable_metrics` with Marathon you get many useful metrics. Unfortunately, the names of the metrics partially contain random parts so that different metric files are not directly comparable.

See https://gist.github.com/bobrik/bb8b852eb1156624a3b8 for an example.

The `$$EnhancerByGuice$$...` should be removed from the class name in the metrics.

gkleiman:
Fixes #1572 - Improve metrics name generation

Updated the metric name generation code in order to remove the Guice marker
("$$EnhancerByGuice$$") from the names.

Added to the name of the metrics generated by the classes using the StateMetrics
trait a prefix, as well as a comma between the class and method name.

For example:

mesosphere.marathon.tasks.TaskTrackerwrite-request-time -> service.mesosphere.marathon.tasks.TaskTracker.write-request-time


src/test/scala/mesosphere/marathon/metrics/MetricsTest.scala
```diff
@@ -0,0 +1,66 @@
...
+
+    metricsSpy.name("prefix", classOf[FooBar], "method1")
+    metricsSpy.name("prefix", classOf[FooBar], "method2")
+    metricsSpy.name("prefix", classOf[MetricsTest], "method1")
+
+    verify(metricsSpy, times(1)).stripGuiceMarksFromClassName(classOf[FooBar])
+    verify(metricsSpy, times(2)).stripGuiceMarksFromClassName(any())
+  }
+
+  test("Metrics#name should use a comma to join the class name and the method name") {
```
kolloch: s/comma/dot/

gkleiman:
I created the following list of metric names (after having applied my changes):

https://gist.github.com/gkleiman/48424cbe40dc9596b2d1

kolloch:
Looks great! Some more comments:

* `TimerUtils` should be removed.
* "Added to the name of the metrics generated by the classes using the StateMetrics
trait a prefix, as well as a comma between the class and method name." I think you can improve the wording here ;)
* One could provide a more convenient API for the named metrics. I am not sure if it is better. Let's talk ;)

Maybe it's silly but you actually test your timer/histogram/gauge implementations in the test.

gkleiman:
Thanks for your comments, I incorporated them in this new version. =)

PR #1627, opened 2015-06-09
CommandInfo protobuf generation fixed for args.

### LINKED ISSUE #1564: args[] does not work
Any job defined by args[] fails to run on mesos with:

F0522 15:37:32.258134 13237 executor.cpp:138] Check failed: task.command().has_value() Executable of task snoozer.5df34f57-00c2-11e5-a2fd-00163e01175a is not specified!

This appears to be a failure to setValue in the Task protobuf in src/main/scala/mesos/TaskBuilder.scala around 192

Example args data:

  "args": [
    "/bin/sleep",
    "90"
  ],

marathon 0.8.1
mesos 0.21.1

aquamatthias:
Fixes #1564

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -152,14 +152,15 @@ case class AppDefinition(
         r => r.getName -> (r.getScalar.getValue: JDouble)
       }.toMap
 
-    val commandOption =
-      if (proto.getCmd.hasValue && proto.getCmd.getValue.nonEmpty)
-        Some(proto.getCmd.getValue)
+    val argsOption =
+      if (proto.getCmd.getArgumentsCount > 0)
+        Some(proto.getCmd.getArgumentsList.asScala.to[Seq])
       else None
 
-    val argsOption =
-      if (commandOption.isEmpty && proto.getCmd.getArgumentsCount != 0)
-        Some(proto.getCmd.getArgumentsList.asScala.to[Seq])
+    //Precondition: either args or command is defined
+    val commandOption =
+      if (argsOption.isEmpty && proto.getCmd.hasValue && proto.getCmd.getValue.nonEmpty)
+        Some(proto.getCmd.getValue)
       else None
 
     val containerOption =
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -197,8 +197,8 @@ object TaskBuilder {
     app.args.foreach { argv =>
       builder.setShell(false)
       builder.addAllArguments(argv.asJava)
-      //mesos containerizer expects the cmd and arguments
-      app.container.getOrElse(builder.setValue(argv.head))
+      //mesos command executor expects cmd and arguments
+      if (app.container.isEmpty) builder.setValue(argv.head)
     }
 
     if (app.uris != null) {
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -52,7 +52,7 @@ class TaskBuilder(app: AppDefinition,
   }
 
   private def build(offer: Offer, cpuRole: String, memRole: String, diskRole: String,
-                    portsResources: Seq[RangesResource]) = {
+                    portsResources: Seq[RangesResource]): Some[(TaskInfo, Seq[Long])] = {
 
     val executor: Executor = if (app.executor == "") {
       config.executor
@@ -99,7 +99,7 @@ class TaskBuilder(app: AppDefinition,
         val portMappings = c.docker.map { d =>
           d.portMappings.map { pms =>
             pms zip ports map {
-              case (mapping, port) => {
+              case (mapping, port) =>
                 // Use case: containerPort = 0 and hostPort = 0
                 //
                 // For apps that have their own service registry and require p2p communication,
@@ -114,7 +114,6 @@ class TaskBuilder(app: AppDefinition,
                 else {
                   mapping.copy(hostPort = port.toInt)
                 }
-              }
             }
           }
         }
@@ -126,7 +125,7 @@ class TaskBuilder(app: AppDefinition,
             }
           )
         }
-        containerWithPortMappings.toMesos
+        containerWithPortMappings.toMesos()
       }
 
     executor match {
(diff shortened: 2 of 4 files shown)


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -152,8 +152,11 @@ case class AppDefinition(
         r => r.getName -> (r.getScalar.getValue: JDouble)
       }.toMap
 
+    //The command is executed inside a shell, but the args are not.
+    //Since both attributes set the proto.cmd, we use this flag to
+    //distinguish if the command was set or the args array.
     val commandOption =
-      if (proto.getCmd.hasValue && proto.getCmd.getValue.nonEmpty)
+      if (proto.getCmd.getShell && proto.getCmd.hasValue && proto.getCmd.getValue.nonEmpty)
```
kolloch: It's is not clear what your comment relates to. I guess you should refer to `proto.getCmd.getShell`.

kolloch:
LGTM! except the comment wording

ConnorDoyle:
I think based on the Mesos docs it might not be quite this simple: https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L280

Following the docs, at least, Marathon should probabaly allow `cmd` and `args` to be supplied together, but only when using the Mesos containerizer.

1. Docker containerizer
  - If the command value is specified, overwrite the ENTRYPOINT with `bin/sh -c` and pass the command following `docker run ...`.
  - If args are supplied, leave the entrypoint unchanged and pass the args following `docker run ...`
1. Mesos containerizer
  - If the command value is supplied, and `shell` is set to `true`, exec `bin/sh -c '<value>'`.
  - If shell is set to `false`, `execlp(<value> <args[0]> .. <args[n-1]);`

The previous protobuf marshaling code works for the Docker containerizer.  In Mesos version `0.20.0`, the `Command.Value` field was changed from required to optional.  I'm wondering whether this should also be considered a bug in the Mesos command executor, or if this is the desired behavior.

cc @tnachen for a 2nd opinion

src/test/scala/mesosphere/marathon/state/AppDefinitionTest.scala
```diff
@@ -90,6 +87,41 @@ class AppDefinitionTest extends MarathonSpec with Matchers {
     assert(!proto2.hasAcceptedResourceRoles)
   }
 
+  test("CMD to proto and nback again") {
```
ConnorDoyle: Tiny typo

aquamatthias:
Hey @ConnorDoyle 
Thanks for the feedback. 
The problem described in the issue arises with cmd/args and the CommandExecutor.
The protobuf expects a command according to the posix execvp syscall, that means
that the cmd.args holds the command and all parameter, where cmd.value is only the executable.
(e.g. ["sleep", "90"] ->  {cmd.value="sleep", cmd.args=["sleep", "90"]}
That behavior is implemented in this fix.

If I understand your comment correctly, you suggest to not allow args without cmd and than adapt the protobuf to include cmd as the first arg?

I did not change the implicit decision, that args are executed without a shell, where the cmd is executed with a shell. But that is also questionable. What is your opinion?

kolloch:
Just to provide some more explicit context to the discussion. As you have implicitly acknowledged, our AppDefinitionValidator does not allow both cmd/args to be specified add the same time: `(cmd ^ args) || (!(cmd || args) && container)` 

It would actually be great if Mesos respected the shell parameter even for docker containers with args. Are you suggesting something similar? We could then reexport this parameter to the user. That would solve some user problems by providing command line substitution on the args.

kolloch:
My question would be: Does this PR break anything that worked before this PR? Or does this PR establish behavior that prevents improving our semantics in the future?

If not, I would go ahead.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -197,6 +196,7 @@ object TaskBuilder {
     // args take precedence over command, if supplied
     app.args.foreach { argv =>
       builder.setShell(false)
+      builder.setValue(argv.head)
```
ConnorDoyle: Because of https://github.com/apache/mesos/blob/0.22.1/src/docker/docker.cpp#L451 -- we should only do this if this task uses the Mesos containerizer.

ConnorDoyle: If this was tested already and it somehow worked, I'd be interested in that too.  At the very least, we should verify that both the Mesos containerizer + command executor and the Docker containerizer both still work with both `cmd` and `args`.

aquamatthias: It did not work for the Mesos containerizer, but probably for the docker containerizer. It looks like the protobuf fields (cmd, args) have a different format with a different containerizer. We could deal with that easily, but we use the protobufs also for persisting the AppDefinition (and that needs to be synchronous (write/read)).

kolloch: I created a Mesos JIRA ticket to fix the documentation (at least):

https://issues.apache.org/jira/browse/MESOS-2859

aquamatthias:
Implemented Change:

The ServiceDefinitionProto <-> AppDefinition mapping now works vice versa with or without container, with args or cmd.

1. MesosExecutor: 
   - `cmd`: __(not changed)__ is executed in a shell  
   - `args`: is executed without shell (no cmd allowed). In that case the `CommandInfo.value` is set to the first arg in the args array

1. DockerContainerizer: 
  - `cmd`: __(not changed)__ overrides the command with `/bin/sh` and sets the command
  - `args`: __(not changed)__ adds the arguments to the docker run command

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -198,6 +197,8 @@ object TaskBuilder {
     app.args.foreach { argv =>
       builder.setShell(false)
       builder.addAllArguments(argv.asJava)
+      //mesos containerizer expects the cmd and arguments
+      app.container.getOrElse(builder.setValue(argv.head))
```
kolloch: I don't like that you use `.getOrElse` for a side effect. I would like `if(app.container.isEmpty)` better. 

It is also hard for me to tell what the special case is. Is docker the special case or the command executor? What would you pass to a custom executor? Or another container executor?

But maybe that's to theoretical and we should focus on fixing this ;)

kolloch:
Thanks, looks good!

PR #1628, opened 2015-06-09
Fixes #1556 - Allow proxying to leader with HTTPS

### LINKED ISSUE #1556: Proxying doesn't work for HTTPS
I have a cluster of three Marathon instances running HTTPS only (Port 8443) using key store and --disable_http.

In this mode, the Marathon UI & API across the entire cluster seems to be broken. I see the following errors in logs every time an API call is done or the UI makes /apps or /deployments AJAX calls to the backend:

[2015-05-20 20:02:12,654] INFO Proxying request to leader at __marathon3.abc.com:8080__ (mesosphere.marathon.api.LeaderProxyFilter:62)
[2015-05-20 20:02:12,655] WARN Exception while proxying (mesosphere.marathon.api.LeaderProxyFilter:128)
java.net.ConnectException: Connection refused
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:339)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:200)

The reason for this is because the proxy is happening to marathon-leader i.e, marathon3.abc.com:__8080__ and NOT __8443__ which would clearly break as none of our nodes are listening on 8080 after using the --disable_http flag.

Looking into the code, the LeaderProxyFilter.scala is using the latest data from zookeeper. On doing some digging into data in my zookeeper cluster I found the leader stored in there points to 8080, NOT 8443:

----
root@zookeeper2:/a-zookeeper/bin# ./zkCli.sh -server localhost:2181
Connecting to localhost:2181
Welcome to ZooKeeper!
JLine support is enabled

WATCHER::
WatchedEvent state:SyncConnected type:None path:null
[zk: localhost:2181(CONNECTED) 0] ls /marathon/leader 
[member_0000000010, member_0000000009, member_0000000008]
[zk: localhost:2181(CONNECTED) 1] get /marathon/leader/member_0000000010
__marathon3.abc.com:8080__

----

This seems to be a serious bug and breaks the entire system in SSL/TLS only mode. I can't think of any work around to get this working without compromising security.

When done right, Marathon should be proxying to 8443 or whatever port the current leader is serving https traffic on. I hope we can get this fix in as a part of #828. Adding @kolloch for context.

Thanks!

kolloch:
so it works if http is disabled.

It assumes that all Marathon instances are similarly configured.


src/main/scala/mesosphere/marathon/io/SSLContextUtil.scala
```diff
@@ -0,0 +1,35 @@
...
+      ts.load(in, passwordOpt.map(_.toCharArray).orNull)
+    }
+
+    // initialize a new TMF with the ts we just loaded
+    val tmf = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm)
+    tmf.init(ts)
+
+    // acquire X509 trust manager from factory
+    val context = SSLContext.getInstance("TLS")
+    context.init(null, tmf.getTrustManagers, null)
```
aquamatthias: I had to look up the JavaDoc to understand the nulls here. Can you add a short sentence on that?

src/test/scala/mesosphere/marathon/integration/setup/ForwarderService.scala
```diff
@@ -12,6 +15,7 @@ import mesosphere.marathon.api._
 import mesosphere.marathon.{ LeaderProxyConf, ModuleNames }
 import org.rogach.scallop.ScallopConf
 import org.slf4j.LoggerFactory
+import mesosphere.marathon.io.{ SSLContextUtil, IO }
```
aquamatthias: Minor: most of the added imports are unused.

src/test/scala/mesosphere/marathon/integration/ForwardToLeaderIntegrationTest.scala
```diff
@@ -34,16 +38,54 @@ class ForwardToLeaderIntegrationTest extends IntegrationFunSuite with BeforeAndA
...
+      "--https_address", "localhost",
+      "--https_port", ports.head.toString))
+
+    val pingURL = new URL(s"https://localhost:${ports.head}/ping")
+    val connection = SSLContextTestUtil.sslConnection(pingURL)
+    val response = IO.using(connection.getInputStream)(IO.copyInputStreamToString)
+    assert(response == "pong\n")
+  }
+
+  test("forwarding HTTPS ping") {
```
aquamatthias: Nice tests!

aquamatthias:
Just 2 minor comments. Looks good!

kolloch:
Thanks for the comments! Rebased and followed up on your comments.

PR #1646, opened 2015-06-11
Fixes #1625 - Close stream handle when the current instance loses

### LINKED ISSUE #1625: SSE Stream handle should close the connection, when the leadership is lost
If marathon is setup in a quorum, only the leader should send events.
If one leader abdicates with currently connected stream handlers, those stream handlers should get notified and the stream should be closed. 
The following connection will be established to the current elected leader.

kolloch:
leadership

Also:

* Refactor Leadership management into EventStream messages
* Report the number of open event streams via metrics

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
@@ -16,7 +16,7 @@ import com.twitter.common.zookeeper.Candidate.Leader
 import com.twitter.common.zookeeper.Group.JoinException
 import mesosphere.marathon.MarathonSchedulerActor._
 import mesosphere.marathon.Protos.MarathonTask
-import mesosphere.marathon.event.LocalLeadershipEvent
+import mesosphere.marathon.event.{ EventModule, LocalLeadershipEvent }
 import mesosphere.marathon.health.HealthCheckManager
 import mesosphere.marathon.state.{ AppDefinition, AppRepository, Migration, PathId, Timestamp }
 import mesosphere.marathon.tasks.TaskTracker
@@ -49,7 +49,7 @@ class MarathonSchedulerService @Inject() (
     system: ActorSystem,
     migration: Migration,
     @Named("schedulerActor") schedulerActor: ActorRef,
-    eventStream: EventStream) extends AbstractExecutionThreadService with Leader {
+    @Named(EventModule.busName) eventStream: EventStream) extends AbstractExecutionThreadService with Leader {
 
   import mesosphere.util.ThreadPoolContext.context
 
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
@@ -6,6 +6,7 @@ import java.util.{ Timer, TimerTask }
 import javax.inject.{ Inject, Named }
 
 import akka.actor.{ ActorRef, ActorSystem }
+import akka.event.EventStream
 import akka.pattern.{ after, ask }
 import akka.util.Timeout
 import com.google.common.util.concurrent.AbstractExecutionThreadService
@@ -15,6 +16,7 @@ import com.twitter.common.zookeeper.Candidate.Leader
 import com.twitter.common.zookeeper.Group.JoinException
 import mesosphere.marathon.MarathonSchedulerActor._
 import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.event.LocalLeadershipEvent
 import mesosphere.marathon.health.HealthCheckManager
 import mesosphere.marathon.state.{ AppDefinition, AppRepository, Migration, PathId, Timestamp }
 import mesosphere.marathon.tasks.TaskTracker
@@ -46,7 +48,8 @@ class MarathonSchedulerService @Inject() (
     driverFactory: SchedulerDriverFactory,
     system: ActorSystem,
     migration: Migration,
-    @Named("schedulerActor") schedulerActor: ActorRef) extends AbstractExecutionThreadService with Leader {
+    @Named("schedulerActor") schedulerActor: ActorRef,
+    eventStream: EventStream) extends AbstractExecutionThreadService with Leader {
 
   import mesosphere.util.ThreadPoolContext.context
 
@@ -273,7 +276,7 @@ class MarathonSchedulerService @Inject() (
   private def defeatLeadership(): Unit = {
     log.info("Defeat leadership")
 
-    schedulerActor ! Suspend(LostLeadershipException("Leadership was defeated"))
+    eventStream.publish(LocalLeadershipEvent.Standby)
 
     timer.cancel()
     timer = newTimer()
@@ -291,7 +294,7 @@ class MarathonSchedulerService @Inject() (
     leader.set(true)
     runDriver(abdicateOption)
 
-    schedulerActor ! Start
+    eventStream.publish(LocalLeadershipEvent.ElectedAsLeader)
 
     // Start the timer
     schedulePeriodicOperations()
--- a/src/main/scala/mesosphere/marathon/event/Events.scala
+++ b/src/main/scala/mesosphere/marathon/event/Events.scala
@@ -40,6 +40,14 @@ object EventModule {
   final val busName = "events"
 }
 
+/** Local leadership events. They are not delivered via the event endpoints. */
+sealed trait LocalLeadershipEvent
+
+object LocalLeadershipEvent {
+  case object ElectedAsLeader
+  case object Standby
+}
+
 sealed trait MarathonEvent {
   val eventType: String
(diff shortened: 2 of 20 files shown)


kolloch:
rebased

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -291,7 +294,7 @@ class MarathonSchedulerService @Inject() (
     leader.set(true)
     runDriver(abdicateOption)
 
-    schedulerActor ! Start
+    eventStream.publish(LeadershipStateEvent.ElectedAsLeader)
```
aquamatthias: This is cool for logic, that gets executed sometimes in the future after the leader was elected. I could think of other logic (think of migration etc), that has to be executed after election but before the driver starts. I know this is not part of this PR, but would be really useful.

src/main/scala/mesosphere/marathon/event/Events.scala
```diff
@@ -40,6 +40,14 @@ object EventModule {
   final val busName = "events"
 }
 
+/** Local leadership events. They are not delivered via the event endpoints. */
+sealed trait LeadershipStateEvent
+
+object LeadershipStateEvent {
+  case object ElectedAsLeader
```
aquamatthias: I could imagine, that this kind of event is also helpful for external services. 
For that it should have something like the IP Address.
As SSE Listener this is would probably be the last message, that is received.

src/main/scala/mesosphere/marathon/event/Events.scala
```diff
@@ -40,6 +40,14 @@ object EventModule {
...
+  case object Standby
```
aquamatthias: Same comment as for ElectedAsLeader.

src/main/scala/mesosphere/marathon/event/Events.scala
```diff
@@ -40,6 +40,14 @@ object EventModule {
   final val busName = "events"
 }
 
+/** Local leadership events. They are not delivered via the event endpoints. */
+sealed trait LeadershipStateEvent
```
aquamatthias: I wonder, if we should make the concept of "Local Events" more visible (like exception hierarchy for checked, runtime exceptions).
See my comments below, where the question is: do we need "Local Events" for this.

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
```diff
@@ -16,49 +22,117 @@ trait HttpEventStreamHandle {
   def close(): Unit
 }
 
+class HttpEventStreamActorMetrics @Inject() (metrics: Metrics) {
+  val numberOfStreams: AtomicIntGauge =
+    metrics.gauge(metrics.name("api", getClass, "number-of-streams"), new AtomicIntGauge)
```
aquamatthias: I would like to have a more specific method like metrics.name -> metrics.api or something like that. So that we don't deal with strings here.

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
```diff
@@ -16,49 +22,117 @@ trait HttpEventStreamHandle {
...
+  private[this] val active: Receive = behaviour(acceptingNewConnections)
+  private[this] val standby: Receive = behaviour(rejectingNewConnections)
+
+  /**
+    * Helper method to create behaviours.
+    * The behaviours only differ in how they deal with new connections.
+    */
+  private[this] def behaviour(newConnectionBehaviour: Receive): Receive = {
+    // Prevent fatal warning about deriving type Any as type parameter
+    def orElse(receives: Receive*): Receive = receives.reduceLeft { _.orElse[Any, Unit](_) }
```
aquamatthias: Minor: For me the simpler version would be easier to understand: Seq(handleLeadership, cleanupHandlerActors, newConnectionBehaviour, warnAboutUnknownMessages).reduceLeft { _.orElse[Any, Unit](_) }

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
```diff
@@ -16,49 +22,117 @@ trait HttpEventStreamHandle {
...
-      eventStream,
-      maxOutstandingMessages), handle.id)
-    context.watch(actor)
-    clients += handle -> actor
+  // behaviour components
+
+  /** Immediately close new connections. */
+  private[this] def rejectingNewConnections: Receive = {
+    case HttpEventStreamConnectionOpen(handle) =>
+      log.warn("Ignoring open connection request. Closing handle.")
```
aquamatthias: The LeaderProxyFilter should prevent that from happening, correct?

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
```diff
@@ -16,49 +22,117 @@ trait HttpEventStreamHandle {
...
+        s"Current nr of listeners: ${streamHandleActors.size}")
     }
   }
 
-  def removeActor(actor: ActorRef): Unit = {
-    clients.find(_._2 == actor).foreach {
+  private[this] def removeHandlerActor(actor: ActorRef): Unit = {
+    streamHandleActors.find(_._2 == actor).foreach {
       case (handle, ref) =>
         log.error(s"Actor terminated unexpectedly: $handle")
```
aquamatthias: This is no longer unexpected, since it is used when this actor becomes Standby. Either change the logic in the standby case to have the logic here as exception (that was my original intend) or change the log statement.

aquamatthias:
This looks generally very good. 
I have some questions regarding local events and some minor comments.  
Can you also look into the tests? ( I restarted it, but it keeps failing )

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
kolloch: Yes, I agree. (It would be useful and it is not part of this PR).

You can create an issue for that.

src/main/scala/mesosphere/marathon/event/Events.scala
kolloch: I thought about that but I think it is something different that might be useful on its own.

These are events from the viewpoint of the Marathon instance processing these. They are not global events. You could turn them into global events but it gets more complicated.

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
kolloch: What about moving `"service"` and `"api"` as constants to an object `mesosphere.marathon.MetricPrefixes`?

I would not like to introduce an "api" method in Metrics since I would like to keep it general.

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
kolloch: Ok, I'll change that.

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
kolloch: There is always a race condition.

"Doppelt hält besser ;)"

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
kolloch: Good point!

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
aquamatthias: I agree.

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
aquamatthias: I would not omit the code - like double checks.

kolloch:
changed according to your suggestions or our discussion

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamActor.scala
aquamatthias: You introduced the MetricsPrefixes but did not use it here...

aquamatthias:
The one "api" slipped through.
Also the integration tests fail constantly on my local machine.

kolloch:
I used the supplied constants for the "api" string and I fixed the guice wiring.

In a separate commit, I also made ProcessKeeper.startProcess fail immediately if the started process unexpectedly exits.

kolloch:
Integration tests still fail.

kolloch:
I changed the leadership subscription logic, too.

aquamatthias:
Cool!

Issue #555, opened 2014-09-05
Constraints not satisfied when scaling down


normanjoyner:
When scaling down, it seems that the constraints are not kept. For example, I have a region attribute which is set to either ```us-east-1b``` or ```us-east-1c```.  When running 8 containers using the ```GROUP_BY``` constraint, there are 4 containers satisfying each attribute. When I scale down to 4 containers, there are not always 2 containers satisfying each attribute.

This was tested against the master branch.

PR #1650, opened 2015-06-12
Fixes #1397 update AppDefinition in group when killing tasks w/ scale

### LINKED ISSUE #1397: Kill tasks with scaling does not update group
When killing tasks of an app with scaling, the app definition in the group is not being updated leading to scaling to the old number afterwards not being possible because no change is detected. Scaling to a different number of instances works.

meichstedt:
Fixes #1397 When killing tasks via the UI with scale, update the AppDefinition via GroupManager so that, effectively, a deployment is initiated and a new version created. The tasks that shall be killed will be passed over to the ScaleApplication DeploymentStep and, if still running, be killed before any other task that might be killed in order to satisfy the scaleTo instance value.


kolloch:
Hi @meichstedt,

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -88,10 +88,33 @@ class AppTasksResource @Inject() (service: MarathonSchedulerService,
                 @PathParam("taskId") id: String,
                 @QueryParam("scale") scale: Boolean = false): Response = {
     val pathId = appId.toRootPath
+
+    import mesosphere.util.ThreadPoolContext.context
+
     if (taskTracker.contains(pathId)) {
       val tasks = taskTracker.get(pathId)
       tasks.find(_.getId == id).fold(unknownTask(id)) { task =>
-        service.killTasks(pathId, Seq(task), scale)
```
kolloch: This is fairy similar to the `deleteMany` call above -- so you would need to adjust that two, I guess.

I do not like that we now have more non-API logic in here than before. Maybe introduce a new service? It might even be something specific as the `TaskKillService` or something like that this contains this logic and is also used by the `deleteMany` call.

We can look at other groupManager.update calls later...

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -88,10 +88,33 @@ class AppTasksResource @Inject() (service: MarathonSchedulerService,
...
+        if (scale) {
+          service.getApp(pathId) match {
```
kolloch: You probably copied this from other similar code but it is not ideal.

The `groupManager.updateApp` below passes the old app to the lambda function that you pass in.

The `updateApp` call itself is serialized and concurrent updates of the app definition are prevented. Your code contains a race condition. You update your app based on the result of the getApp call. If you have updates to the same app in-between, your code might discard them.

src/main/scala/mesosphere/marathon/upgrade/DeploymentActor.scala
```diff
@@ -115,12 +115,19 @@ class DeploymentActor(
     promise.future
   }
 
-  def scaleApp(app: AppDefinition, scaleTo: Int): Future[Unit] = {
+  def scaleApp(app: AppDefinition, scaleTo: Int, toKill: Option[Set[MarathonTask]]): Future[Unit] = {
     val runningTasks = taskTracker.get(app.id)
-    if (scaleTo == runningTasks.size) {
-      Future.successful(())
+    val runningTasksToKill = DeploymentDeathRow(runningTasks, toKill.getOrElse(Set.empty), scaleTo).tasksToBeKilled
```
kolloch: This could be a simple function call.

I'd actually like to move as much logic out of this function in the DeploymentActor as possible into a pure function outside.

It could just return the Set of tasks to kill (might be empty or you model that with an Option[Set]) and an Option[Int]  scaleToOpt.

src/main/scala/mesosphere/marathon/upgrade/DeploymentDeathRow.scala
```diff
@@ -0,0 +1,18 @@
+package mesosphere.marathon.upgrade
+
+import mesosphere.marathon.Protos.MarathonTask
+
+private[upgrade] case class DeploymentDeathRow(runningTasks: Set[MarathonTask],
```
kolloch: If you expand this as suggested above, you could call it "ScalingCalculator" or something like this.

src/main/scala/mesosphere/marathon/upgrade/DeploymentDeathRow.scala
```diff
@@ -0,0 +1,18 @@
...
+                                               sentencedToDeath: Set[MarathonTask],
+                                               scaleTo: Int) {
+
+  val tasksToBeKilled: Seq[MarathonTask] = {
+    val (sentencedAndRunning, notSentencedAndRunning) = runningTasks partition sentencedToDeath
+    val keepAliveCount = runningTasks.size - math.max(sentencedAndRunning.size, runningTasks.size - scaleTo)
```
kolloch: Wouldn't it be easier to calculate the `killCount` and use `take` instead of `dropRight` below?

src/main/scala/mesosphere/marathon/upgrade/DeploymentDeathRow.scala
```diff
@@ -0,0 +1,18 @@
...
+    // is getStartedAt a good value to sortBy?
+    val ordered = sentencedAndRunning.toSeq ++ notSentencedAndRunning.toSeq.sortBy(_.getStartedAt)
```
kolloch: prioritize unhealthy tasks here?

kolloch: Now you kill the oldest tasks first. Before we killed the newest ones first.

I am not sure what is better. Maybe don't sort at all ;)

src/test/scala/mesosphere/marathon/upgrade/DeploymentActorTest.scala
```diff
@@ -86,9 +85,9 @@ class DeploymentActorTest
     // the AppDefinition is never used, so it does not mater which one we return
     when(repo.store(any())).thenReturn(Future.successful(AppDefinition()))
 
-    when(driver.killTask(TaskID(task1_2.getId))).thenAnswer(new Answer[Status] {
+    when(driver.killTask(TaskID(task1_1.getId))).thenAnswer(new Answer[Status] {
       def answer(invocation: InvocationOnMock): Status = {
-        system.eventStream.publish(MesosStatusUpdateEvent("", "task1_2", "TASK_KILLED", "", app1.id, "", Nil, app1New.version.toString))
+        system.eventStream.publish(MesosStatusUpdateEvent("", "task1_1", "TASK_KILLED", "", app1.id, "", Nil, app1New.version.toString))
```
kolloch: Maybe try to preserve the old behavior here

src/test/scala/mesosphere/marathon/upgrade/DeploymentDeathRowTest.scala
```diff
@@ -0,0 +1,71 @@
+package mesosphere.marathon.upgrade
+
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.state.Timestamp
+import mesosphere.marathon.tasks.MarathonTasks
+import org.scalatest.FunSuite
+
+class DeploymentDeathRowTest extends FunSuite {
```
kolloch: It is nice that your separated some of the logic to be testable in isolation.

But you also need a more high-level test that tests the changes in the DeploymentActor.

kolloch:
Pretty involved starting tasks, so kudos for getting so far. :+1: 

I would like toe see some things to be changed, though.

meichstedt:
@kolloch thanks for the review! Please see the applied changes and enhanced tests.

kolloch:
Hi @meichstedt, I start reviewing this now. It would be great if you squashed this into one commit (without changes). http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Maybe you could also document that we want PRs with one commit and how to squash them in our contribution documentation?

Thanks!

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -88,15 +81,46 @@ class AppTasksResource @Inject() (service: MarathonSchedulerService,
...
+        }
+        result(future)
+      }
+      else {
+        // setting scale = false in order to make the differing behaviour explicit
+        service.killTasks(pathId, toKill, scale = false)
+      }
+
+      if (toKill.size == 1) {
+        ok(Map("task" -> toKill.head))
```
kolloch: I fear that your generalization does not work here because it might result in `deleteMany` returning a result with a `task` field if only on task was found. Also `deleteOne` might return an empty `tasks` list instead of `"task": null`.

Unfortunately, I cannot think of an elegant solution to this right now.

kolloch: Maybe, you let the method return a `Set[MarathonTask]` instead and handle the generation of the response in the `deleteMany`/`deleteOne` methods?

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -88,15 +81,46 @@ class AppTasksResource @Inject() (service: MarathonSchedulerService,
...
+              pathId, updateAppFunc, Timestamp.now(), force = false, toKill = toKill))
+            deploymentResult(deploymentPlan)
+
+          case None => unknownGroup(pathId.parent)
+        }
+        result(future)
+      }
+      else {
+        // setting scale = false in order to make the differing behaviour explicit
+        service.killTasks(pathId, toKill, scale = false)
```
kolloch: this method should never be called directly with `scale = true`, correct? Can we make that more explicit somehow? Maybe remove that possibility completely? It is still called in `TasksResource` which should probably be changed as well. This would be easier if you moved this method out of the `api` package.

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
```diff
@@ -0,0 +1,9 @@
+package mesosphere.marathon.upgrade
+
+import mesosphere.marathon.Protos.MarathonTask
+
+class ScalingProposition(private val candidatesToKill: Seq[MarathonTask],
+                         private val numberOfTasksToStart: Int) {
+  val tasksToKill = if (candidatesToKill.nonEmpty) Some(candidatesToKill) else None
```
kolloch: I think that I would use a case class and directly calculate the `Option`s in `scalingProposition` method.

src/main/scala/mesosphere/marathon/upgrade/ScalingCalculator.scala
```diff
@@ -0,0 +1,19 @@
+package mesosphere.marathon.upgrade
+
+import mesosphere.marathon.Protos.MarathonTask
+
+object ScalingCalculator {
+
+  def scalingProposition(runningTasks: Set[MarathonTask],
```
kolloch: You could move this to the companion object of ScalingProposition, maybe even call it `apply`.

src/main/scala/mesosphere/marathon/upgrade/DeploymentActor.scala
```diff
@@ -115,12 +115,15 @@ class DeploymentActor(
     promise.future
   }
 
-  def scaleApp(app: AppDefinition, scaleTo: Int): Future[Unit] = {
+  def scaleApp(app: AppDefinition, scaleTo: Int, toKill: Option[Set[MarathonTask]]): Future[Unit] = {
     val runningTasks = taskTracker.get(app.id)
-    if (scaleTo == runningTasks.size) {
-      Future.successful(())
+    val proposition = ScalingCalculator.scalingProposition(runningTasks, toKill, scaleTo)
```
kolloch: If `ScalingCalculator` was a case class, you could destructure it directly here into tasksToKill/tasksToStart.

src/main/scala/mesosphere/marathon/upgrade/DeploymentActor.scala
```diff
@@ -137,9 +140,8 @@ class DeploymentActor(
       )
       promise.future
     }
-    else {
-      killTasks(app.id, runningTasks.toSeq.sortBy(_.getStartedAt).drop(scaleTo))
-    }
+
+    killTasksIfNeeded.flatMap(_ => startTasksIfNeeded)
```
kolloch: Nice method structure now!

src/test/scala/mesosphere/marathon/upgrade/DeploymentPlanTest.scala
```diff
@@ -304,4 +305,36 @@ class DeploymentPlanTest extends MarathonSpec with Matchers with GivenWhenThen {
...
+    val targetGroup = Group(
+      id = "/test".toPath,
+      apps = Set(newApp),
+      groups = Set(
+        Group("/test/some".toPath, Set(newApp))
+      )
+    )
+
+    val taskToKill = MarathonTask.getDefaultInstance
+    val plan = DeploymentPlan(original = originalGroup, target = targetGroup, resolveArtifacts = Seq.empty, version = Timestamp.now(), toKill = Map(aId -> Set(taskToKill)))
```
kolloch: this line is a bit long :)

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
```diff
@@ -0,0 +1,82 @@
...
+  test("Determine tasks to kill when there are no sentenced") {
+    val runningTasks = Set.empty[MarathonTask]
+    val sentencedToDeath = Some(Set.empty[MarathonTask])
+    val scaleTo = 5
+    val proposition = ScalingCalculator.scalingProposition(runningTasks, sentencedToDeath, scaleTo)
+    proposition.tasksToKill shouldBe empty
+    proposition.tasksToStart shouldBe Some(5)
+  }
+
+  test("Determine tasks to kill when none are sentenced and no need for scaling") {
```
kolloch: I like the test but there is a need for scaling or what do you mean by that?

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
```diff
@@ -0,0 +1,82 @@
...
+    val runningTasks: Set[MarathonTask] = Set(
+      createTask(1),
+      createTask(2),
+      createTask(3)
+    )
+    val sentencedToDeath = Some(Set.empty[MarathonTask])
+    val scaleTo = 0
+    val proposition = ScalingCalculator.scalingProposition(runningTasks, sentencedToDeath, scaleTo)
+    proposition.tasksToKill shouldBe defined
+    proposition.tasksToKill.get should have length 3
```
kolloch: why not be more specific and compare to `runningTasks`?

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
```diff
@@ -0,0 +1,82 @@
...
+    val sentencedToDeath = Some(Set(
+      task_2,
+      task_3,
+      MarathonTasks.makeTask("alreadyKilled", "", Nil, Nil, version = Timestamp(0L))))
+    val scaleTo = 3
+    val proposition = ScalingCalculator.scalingProposition(runningTasks, sentencedToDeath, scaleTo)
+
+    proposition.tasksToKill shouldBe defined
+    proposition.tasksToKill.get shouldEqual Seq(task_2, task_3)
+    proposition.tasksToStart shouldBe Some(2) // is None
```
kolloch: `// is None` can be removed, correct?

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
```diff
@@ -0,0 +1,82 @@
+package mesosphere.marathon.upgrade
+
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.state.Timestamp
+import mesosphere.marathon.tasks.MarathonTasks
+import org.scalatest.{ Matchers, FunSuite }
+
+class ScalingCalculatorTest extends FunSuite with Matchers {
```
kolloch: I really like that you factored out the logic of which tasks to kill. The tests are easy to understand.

src/test/scala/mesosphere/marathon/upgrade/DeploymentActorTest.scala
```diff
@@ -280,4 +279,78 @@ class DeploymentActorTest
...
+    val task1_3 = MarathonTasks.makeTask("task1_3", "", Nil, Nil, app1.version).toBuilder.setStartedAt(1000).build()
+
+    val plan = DeploymentPlan(original = origGroup, target = targetGroup, toKill = Map(app1.id -> Set(task1_2)))
+
+    when(tracker.get(app1.id)).thenReturn(Set(task1_1, task1_2, task1_3))
+
+    // the AppDefinition is never used, so it does not mater which one we return
+    when(repo.store(any())).thenReturn(Future.successful(AppDefinition()))
+
+    when(driver.killTask(TaskID(task1_1.getId))).thenAnswer(new Answer[Status] {
```
kolloch: you do not need this stub, do you? and the task1_3 stub not either, right?

src/test/scala/mesosphere/marathon/upgrade/DeploymentActorTest.scala
```diff
@@ -280,4 +279,78 @@ class DeploymentActorTest
...
+        )
+      )
+
+      plan.steps.zipWithIndex.foreach {
+        case (step, num) => managerProbe.expectMsg(5.seconds, DeploymentStepInfo(plan, step, num + 1))
+      }
+
+      managerProbe.expectMsg(5.seconds, DeploymentFinished(plan))
+
+      verify(driver, times(1)).killTask(TaskID(task1_2.getId))
```
kolloch: `verifyNoMoreInteractions` (or similar) would be appropriate here.

kolloch:
I finished my review and suggested further changes...

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
meichstedt: I originally wanted to avoid construction of a ScalingProposition with, e.g.
 ScalingProposition(Some(Seq.empty[...]), Some(-100))
because this is semantical nonsense. The class as written prevents the programmer from doing this. I admit, though, that the implementation is somewhat weird because it looks like a case class but is not.

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
meichstedt: There is indeed a need for scaling – I guess this was a c&p typo :|

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
meichstedt: yep

src/test/scala/mesosphere/marathon/upgrade/DeploymentActorTest.scala
meichstedt: Indeed. There are even more flaws:
- the original AppDefinition should have an instance count of 3, since 3 tasks are 'running'
- the new AppDefinition should have an instance count of 2, since we kill 1 task and scale

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
meichstedt: It would definitely be nicer to separate business/presentation logic here. The problem is, that returning a `Set[MarathonTask]` does not allow materializing responses like `unknownGroup` or `unknownTask` without somehow creating/handling exceptions or using a `Try`, respectively. I don't want to introduce a new approach here, as this would be orthogonal to the current usage of `RestResource`.

src/main/scala/mesosphere/marathon/upgrade/ScalingCalculator.scala
meichstedt: It's moved. Didn't call it apply though: it seems strange to have a case class with an `apply` method whose parameters cannot be destructured via pattern matching. Having a convenience method called `propose()`, which turns a set of parameters into the expected case class values, is imho better to understand.

src/test/scala/mesosphere/marathon/upgrade/ScalingCalculatorTest.scala
meichstedt: Good point, fixed that.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
meichstedt: Agreed. I also removed the `scale` flag from the `KillTasks` message that's being sent to the `SchedulerActor`, since this is no longer needed and the underlying implementation was the problem in question.

meichstedt:
Thanks – I squashed the commits and commented on the changes above

kolloch:
TaskKiller and the Resource logic is untested. Or did I miss something?

src/test/scala/mesosphere/marathon/state/TaskKillerTest.scala
```diff
@@ -0,0 +1,74 @@
...
+import mesosphere.marathon.{ MarathonSchedulerService, MarathonSpec }
+import org.scalatest.mock.MockitoSugar
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.{ BeforeAndAfterAll, Matchers }
+import org.mockito.Mockito._
+import org.mockito.Matchers.{ any, anyBoolean }
+
+import scala.concurrent.Future
+
+class TaskKillerTest extends MarathonSpec
```
kolloch: Nice test suite!

src/test/scala/mesosphere/marathon/state/TaskKillerTest.scala
```diff
@@ -0,0 +1,74 @@
...
+    val result = taskKiller.kill(appId, (tasks) => Set.empty[MarathonTask], scale = true)
+    result.futureValue shouldEqual TaskKiller.GroupNotFound(appId.parent)
+  }
+
+  test("KillRequested") {
+    val appId = PathId(List("my", "app"))
+    val tasksToKill = Set(MarathonTask.getDefaultInstance)
+    when(tracker.contains(appId)).thenReturn(true)
+    when(groupManager.group(appId.parent)).thenReturn(Future.successful(Some(Group.emptyWithId(appId.parent))))
+    when(groupManager.updateApp(any[PathId], any(), any[Timestamp](), anyBoolean(), any[Set[MarathonTask]]())).thenReturn(
```
kolloch: Please make you test more specific here. You should test that the correct parameters are passed to updateApp. E.g. by capturing them and then test them. This is really the good case and it should be tested completely.

src/test/scala/mesosphere/marathon/upgrade/ScalingPropositionTest.scala
```diff
@@ -0,0 +1,124 @@
...
+import mesosphere.marathon.tasks.MarathonTasks
+import org.scalatest.{ FunSuite, Matchers }
+
+class ScalingPropositionTest extends FunSuite with Matchers {
+
+  test("propose - empty tasksToKill should lead to ScalingProposition(None, _)") {
+    val running = Set.empty[MarathonTask]
+    val toKill = Some(Set.empty[MarathonTask])
+    val scaleTo = 0
+    val proposition = ScalingProposition.propose(running, toKill, scaleTo)
```
kolloch: No need to change but an alternative for separate variable here might actually be named arguments.

kolloch:
Looks good:

* make that TaskKiller test complete
* Un-ignore the related test in `AppDeployIntegrationTest` as discussed:

```scala
  ignore("kill all tasks of an App with scaling") {
```

kolloch:
The coverage now got significantly up! Great work!

meichstedt:
@aquamatthias I had to merge your changes regarding #555 – can you please have a look into `DeploymentActor.scaleApp` and `ScalingProposition.propose` to make sure this fits your idea of meeting constraints?

meichstedt:
rebased!

src/main/scala/mesosphere/marathon/state/TaskKiller.scala
```diff
@@ -0,0 +1,43 @@
...
+    if (taskTracker.contains(appId)) {
+      val tasks = taskTracker.get(appId)
+      val toKill = findToKill(tasks)
+      if (scale) {
+        def scaleExisting(opt: Option[AppDefinition]): AppDefinition = opt
+          .map(app => app.copy(instances = app.instances - toKill.size))
+          .getOrElse(throw new UnknownAppException(appId))
+
+        val deploymentPlan = groupManager.updateApp(
+          appId, scaleExisting, Timestamp.now(), force = false, toKill = toKill)
```
aquamatthias: You anticipate force=false. I think we should introduce this as parameter.

src/main/scala/mesosphere/marathon/state/TaskKiller.scala
```diff
@@ -0,0 +1,43 @@
...
+        deploymentPlan.map(_ => toKill)
```
aquamatthias: That means, API users don't see the deployment.

src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala
```diff
@@ -18,8 +19,10 @@ sealed trait DeploymentAction {
 // application has not been started before
 final case class StartApplication(app: AppDefinition, scaleTo: Int) extends DeploymentAction
 
-// application is started, but more instances should be started
-final case class ScaleApplication(app: AppDefinition, scaleTo: Int) extends DeploymentAction
+// application is started, but the instance count should be changed
+final case class ScaleApplication(app: AppDefinition,
+                                  scaleTo: Int,
+                                  sentencedToDeath: Option[Set[MarathonTask]] = None) extends DeploymentAction
```
aquamatthias: Option[Set[MarathonTask]]: in your case is None and Set.empty any different?

src/main/scala/mesosphere/marathon/state/TaskKiller.scala
meichstedt: Agreed, I introduced this param (not yet in the API though).

src/main/scala/mesosphere/marathon/state/TaskKiller.scala
meichstedt: Good point. There are now 2 methods, killTasks and killTasksAndScale – the first one returns a `Future[Set[MarathonTask]]`, the other one a `Future[DeploymentPlan]`.

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
```diff
@@ -107,9 +111,16 @@ class TasksResource @Inject() (
...
+        def findToKill(appTasks: Set[MarathonTask]) = tasks.toSet
+        if (scale) {
+          taskKiller.killAndScale(appId, findToKill, force = true)
+        }
+        else {
+          taskKiller.kill(appId, findToKill, force = true)
+        }
     }
 
+    // TODO: does anyone expect a response with all the deployment plans in case of scaling?
```
aquamatthias: Thanks for the TODO and indeed, it is a strange result. Also there should also only be one deployment.

src/main/scala/mesosphere/marathon/state/TaskKiller.scala
```diff
@@ -0,0 +1,50 @@
+package mesosphere.marathon.state
```
aquamatthias: The TaskKiller holds some functionality that is shared in two Resource classes. Does it makes sense to move this class here? 

(I know the package structure we have is far from ideal, but at least it does not deal with state)

src/main/scala/mesosphere/marathon/state/TaskKiller.scala
```diff
@@ -0,0 +1,50 @@
...
+      val toKill = findToKill(tasks)
+      service.killTasks(appId, toKill)
+      Future.successful(toKill)
+    }
+    else {
+      Future.failed(UnknownAppException(appId))
+    }
+  }
+
+  def killAndScale(appId: PathId,
```
aquamatthias: I like that we now have 2 separate functions for very separate functionality.

aquamatthias:
@meichstedt sorry to say, that you need to rebase again.

PR #1651, opened 2015-06-12
Fixes #1637 - fix recommended Mesos versions and

### LINKED ISSUE #1637: Need upgrade instructions for 0.7.x to 0.8.x?
Navigating through the docs, I found https://mesosphere.github.io/marathon/docs/upgrade/06xto070.html but no similar doc for upgrading to 0.8.x. Even if it's extremely simple, it would be great to have a doc that says so.

kolloch:
mention upgrading from 0.7.* to 0.8.* / 0.8.* to 0.9.*


gkleiman:
Thanks!

PR #1658, opened 2015-06-15
Fix all scalastyle reported issues

aquamatthias:
Use latest scalastyle version.
Fix all existing scala style problems.
All indicators will fail the build.


gkleiman:
:+1:  thanks!

PR #1664, opened 2015-06-15
Tests: ProcessKeeper reports total number of processes on process

kolloch:
launch.

This might be useful to detect resource overuse or something like that.


gkleiman:
:thumbsup:

kolloch:
It's a crazy PR and I am really happy that we got it through in this amazing show of team work.

Thank you @gkleiman !

PR #1665, opened 2015-06-15
Define TeamCity env parameters

aquamatthias:


project/build.scala
```diff
@@ -166,6 +167,27 @@ object MarathonBuild extends Build {
...
+   * It is only executed within teamcity and can be ignored otherwise.
+   */
+  lazy val teamCitySetEnvSettings = Seq(
+    onLoad in Global := {
+      sys.env.get("TEAMCITY_VERSION") match {
+        case None => // no-op
+        case Some(teamcityVersion) =>
+          def reportParameter(key: String, value: String): Unit = {
+            println(s"##teamcity[setParameter name='env.SBT_$key' value='$value']")
+            println(s"##teamcity[setParameter name='system.sbt.$key' value='$value']")
```
kolloch: Why this two lines? What do they mean?

I think some comment would be nice.

kolloch:
Just FYI, this effects the following build steps of the current build if I understood that correctly.

https://confluence.jetbrains.com/display/TCD8/Build+Script+Interaction+with+TeamCity

It would be nice if we added that information to the comments.

Otherwise really cool that you took care of that! Looking forward to improved CI!

PR #1666, opened 2015-06-16
MarathonSchedulerActor: Remove some indirect dependencies

kolloch:
The MarathonSchedulerActor constructor did not only get dependencies for itself but also for other actors/SchedulerActions which were create in preStart.


aquamatthias:
A step in the right direction - I still see the spiders web....

PR #1667, opened 2015-06-16
Fixes #1573 - Report number of apps/groups in metrics

### LINKED ISSUE #1573: Report number of apps/groups in metrics
The `/metrics` endpoint has proved valuable in finding scalability issues recently. See #1497 for an example.

Unfortunately, some very basic information information is missing. The current number of

* apps
* groups
* tasks

gkleiman:
Added the number of add/groups to the metrics exposed by Marathon.

Note: With this change, Marathon will load all the groups from ZK each
time the /metrics endpoint is hit. If this becomes a performance
issue, we can cache the metrics and move the refreshing logic to an
actor. I decided to try to avoid premature optimizations.


aquamatthias:
LGTM

PR #1676, opened 2015-06-17
change help text based on Joel's suggestions

lloesche:
Our technical writer enhanced the wording on a lot of the help text.

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -1,11 +1,13 @@
 #!/usr/bin/env python2
 
 """Overview:
-  Servicerouter is a replacement for the haproxy-marathon-bridge.
-  It reads Marathon task information and generates haproxy configuration.
+  The servicerouter is a replacement for the haproxy-marathon-bridge.
+  It reads the Marathon task information and dynamically generates
+  haproxy configuration details.
 
-  Service configuration lives in Marathon via environment variables.
-  The servicerouter just needs to know where to find Marathon.
+  To gather the task information, the servicerouter needs to know where
+  to find Marathon. The service configuration details are stored in Marathon
+  environment variables.
 
   Every service port in Marathon can be configured independently.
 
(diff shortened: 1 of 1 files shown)


bin/servicerouter.py
```diff
@@ -27,85 +29,87 @@
...
-  Only servicerouter instances which are members of the given group will
-  point to the service. Service routers with the group '*' will collect
-      all groups.
+    The group of servicerouter instances that point to the service.
+    Service routers with the group '*' will collect all groups.
 
   HAPROXY_{n}_VHOST
-  HTTP Virtual Host proxy hostname to catch
-  Ex: HAPROXY_0_VHOST = 'marathon.mesosphere.com'
+    The Marathon HTTP Virtual Host proxy hostname to gather.
```
kolloch: I like neither catch nor gather a lot here.

What is meant is that HAProxy will accept requests on this virtual host and forward them to tasks which have this HAPROXY_{n}_VHOST setting, correct?

I would also mention that you can specify multiple host names and extend the example accordingly.

bin/servicerouter.py
```diff
@@ -27,85 +29,87 @@
...
+    Ex: HAPROXY_0_VHOST = 'marathon.mesosphere.com'
 
   HAPROXY_{n}_STICKY
-  Use sticky request routing for the service
-  Ex: HAPROXY_0_STICKY = true
+    Enable sticky request routing for the service.
```
kolloch: I would like a reference to the HAPROXY configuration that results in. Is it done via cookies, IP, connection? How sticky is it ;)

bin/servicerouter.py
```diff
@@ -27,85 +29,87 @@
...
+    Ex: HAPROXY_0_STICKY = true
 
   HAPROXY_{n}_REDIRECT_TO_HTTPS
-  Redirect HTTP traffic to https
-  Ex: HAPROXY_0_REDIRECT_TO_HTTPS = true
+    Redirect HTTP traffic to HTTPS.
+    Ex: HAPROXY_0_REDIRECT_TO_HTTPS = true
```
kolloch: A permanent redirect? Does it support this as well?

https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security

bin/servicerouter.py
kolloch: Maybe, you can also explain the mechanism with the {n} on top of the list for all variables?

bin/servicerouter.py
```diff
@@ -27,85 +29,87 @@
...
-  Bind to the specific address for the service
-  Ex: HAPROXY_0_BIND_ADDR = '10.0.0.42'
+    Bind to the specific address for the service.
+    Ex: HAPROXY_0_BIND_ADDR = '10.0.0.42'
 
   HAPROXY_{n}_MODE
-  Set the connection mode to either tcp or http. Default is tcp.
-  Ex: HAPROXY_0_MODE = 'http'
+    Set the connection mode to either TCP or HTTP. The default is TCP.
+    Ex: HAPROXY_0_MODE = 'http'
```
kolloch: Not a text comment. Just because we had an issue for that in Marathon: Does it also allow HTTPS? I know that you could argue that this is unnecessary but people have requested it for Marathon.

bin/servicerouter.py
```diff
@@ -27,85 +29,87 @@
...
+    all virtual hosts as defined by the HAPROXY_{n}_VHOST variable.
 
   HAPROXY_HTTPS_FRONTEND_HEAD
-    Same as the previous option just for https encrypted connections
-    coming in on *:443. This needs to be modified to include a different
-    certificate.
+    An HTTPS frontend for encrypted connections that binds to port *:443 by
+    default and gathers all virtual hosts as defined by the
+    HAPROXY_{n}_VHOST variable. You must modify this file to
+    include your certificate.
```
kolloch: Why do I need to modify that if I also have the  HAPROXY_{n}_SSL_CERT variables above?

bin/servicerouter.py
kolloch: The {n} part relates to the n-th virtual host?

bin/servicerouter.py
```diff
@@ -27,85 +29,87 @@
...
 
   HAPROXY_FRONTEND_HEAD
-    Defines which address and port to bind to.
+    Defines the address and port to bind to.
 
   HAPROXY_BACKEND_HEAD
-    Includes the type of load balancing (default is roundrobin)
-    and connection mode (tcp or http).
+    Defines the type of load balancing, roundrobin by default,
+    and connection mode, TCP or HTTP.
```
kolloch: what about the sticky setting above?

lloesche:
This PR just fixes the existing wording so it's correct english and consistent writing. I'm sure every single option could be explained in much more detail at some point. Really not a priority in the scope of this PR.

kolloch:
Ok, it is an improvement. Would be nice if some of the things got addressed though.

PR #1690, opened 2015-06-19
Remove the frameworkId, if the scheduler reports an error.

### LINKED ISSUE #1316: Invalidate framework ID if registration triggers an error
Currently when the scheduler receives the `error` callback, we bail out of the program.  One cause for this callback to be called is to attempt to re-register after being marked completed by the master (e.g. for exceeding the configured framework failover timeout.)

A symptom is that once Marathon is marked completed, it can _never_ re-register until the leading Mesos master process is kicked over, or the persisted FrameworkID is manually deleted from ZooKeeper!

Currently, it's pretty hard to disambiguate this error from other causes of framework errors.  I've filed [MESOS-2522](https://issues.apache.org/jira/browse/MESOS-2522) to add a reason field for framework errors to help with this.

aquamatthias:
Add a note, that we should refine to handle only specific messages.
Fixes #1316 
@ConnorDoyle: please review

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -19,6 +19,7 @@ import org.apache.mesos.{ Scheduler, SchedulerDriver }
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{ Await, Future }
+import scala.util.control.NonFatal
 import scala.util.{ Failure, Success }
 
 trait SchedulerCallbacks {
@@ -187,19 +188,36 @@ class MarathonScheduler @Inject() (
 
   override def error(driver: SchedulerDriver, message: String) {
     log.warn("Error: %s".format(message))
-    suicide()
+
+    // Currently, it's pretty hard to disambiguate this error from other causes of framework errors.
+    // Watch MESOS-2522 which will add a reason field for framework errors to help with this.
+    // For now the frameworkId is removed for all messages.
+    val removeFrameworkId = true
+    suicide(removeFrameworkId)
   }
 
-  private def suicide(): Unit = {
-    log.fatal("Committing suicide")
+  /**
+    * Exits the JVM process, optionally deleting Marathon's FrameworkID
+    * from the backing persistence store.
+    *
+    * If `removeFrameworkId` is set, the next Marathon process elected
+    * leader will fail to find a stored FrameworkID and invoke `register`
+    * instead of `reregister`.  This is important because on certain kinds
+    * of framework errors (such as exceeding the framework failover timeout),
+    * the scheduler may never re-register with the saved FrameworkID until
+    * the leading Mesos master process is killed.
+    */
+  private def suicide(removeFrameworkId: Boolean): Unit = {
+    log.fatal(s"Committing suicide!")
 
-    //scalastyle:off magic.number
+    if (removeFrameworkId) Await.ready(frameworkIdUtil.expunge(), config.zkTimeoutDuration)
+
     // Asynchronously call sys.exit() to avoid deadlock due to the JVM shutdown hooks
-    Future {
-      sys.exit(9)
-    } onFailure {
-      case t: Throwable => log.fatal("Exception while committing suicide", t)
+    // scalastyle:off magic.number
+    Future(sys.exit(9)).onFailure {
+      case NonFatal(t) => log.fatal("Exception while committing suicide", t)
     }
+    // scalastyle:on
   }
 
   private def postEvent(status: TaskStatus, task: MarathonTask): Unit = {
--- a/src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
+++ b/src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
@@ -4,7 +4,7 @@ import mesosphere.marathon.state.{ MarathonState, MarathonStore }
 import org.apache.mesos.Protos
 import org.apache.mesos.Protos.FrameworkID
 
-import scala.concurrent.Await
+import scala.concurrent.{ Future, Await }
 import scala.concurrent.duration.Duration
 
 /**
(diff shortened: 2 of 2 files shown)


src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -187,19 +188,22 @@ class MarathonScheduler @Inject() (
...
 
-    //scalastyle:off magic.number
     // Asynchronously call sys.exit() to avoid deadlock due to the JVM shutdown hooks
-    Future {
-      sys.exit(9)
-    } onFailure {
-      case t: Throwable => log.fatal("Exception while committing suicide", t)
-    }
+    // scalastyle:off magic.number
+    Future(sys.exit(9)).onFailure { case NonFatal(t) => log.fatal("Exception while committing suicide", t) }
```
ConnorDoyle: Do newlines cost money or something? :)

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -187,19 +188,22 @@ class MarathonScheduler @Inject() (
 
   override def error(driver: SchedulerDriver, message: String) {
     log.warn("Error: %s".format(message))
-    suicide()
+    suicide(message)
   }
 
-  private def suicide(): Unit = {
-    log.fatal("Committing suicide")
+  private def suicide(message: String): Unit = {
```
ConnorDoyle: Please document this (scaladoc) to make it clear that this method not only shuts down the scheduler but now expunges the framework ID.  In fact, maybe it's better to add a parameter to make that behavior explicit.

aquamatthias:
@ConnorDoyle incorporated your feedback

aquamatthias:
Hey @ConnorDoyle any thoughts on this?

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -187,18 +188,31 @@ class MarathonScheduler @Inject() (
...
+    // For now the frameworkId is removed for all messages.
+    val removeFrameworkId = true
+    suicide(removeFrameworkId)
   }
 
-  private def suicide(): Unit = {
-    log.fatal("Committing suicide")
+  /**
+    * Exit the JVM process.
+    * Optionally deleting the frameworkId. Depending on the cause of this error,
```
ConnorDoyle: Needs rewording for clarity IMO.

How about:

```
/**
  * Exits the JVM process, optionally deleting Marathon's FrameworkID
  * from the backing persistence store.
  *
  * If `removeFrameworkId` is set, the next Marathon process elected
  * leader will fail to find a stored FrameworkID and invoke `register`
  * instead of `reregister`.  This is important because on certain kinds
  * of framework errors (such as exceeding the framework failover timeout),
  * the scheduler may never re-register with the saved FrameworkID until
  * the leading Mesos master process is killed.
  */

File: src/main/scala/mesosphere/marathon/MarathonScheduler.scala [hunk 4]
```diff
@@ -187,18 +188,31 @@ class MarathonScheduler @Inject() (
...
+
+    if (removeFrameworkId) Await.ready(frameworkIdUtil.expunge, config.zkTimeoutDuration)
 
-    //scalastyle:off magic.number
     // Asynchronously call sys.exit() to avoid deadlock due to the JVM shutdown hooks
-    Future {
-      sys.exit(9)
-    } onFailure {
-      case t: Throwable => log.fatal("Exception while committing suicide", t)
+    // scalastyle:off magic.number
```
ConnorDoyle: This turns off the rule for the remainder of the file.  Please use `//scalastyle:on magic.number` following the next line or move the scalastyle directive to the end of the line, e.g.:

```scala
Future(sys.exit(9)).onFailure {  // scalastyle:off magic.number
```

src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
```diff
@@ -18,6 +18,7 @@ class FrameworkIdUtil(mStore: MarathonStore[FrameworkId], timeout: Duration, key
     val frameworkId = FrameworkId(proto.getValue)
     Await.result(mStore.modify(key) { _ => frameworkId }, timeout)
   }
+  def expunge: Future[Boolean] = mStore.expunge(key)
```
ConnorDoyle: The convention is to declare and call nullary methods _with_ the parentheses when there are side effects.  See http://docs.scala-lang.org/style/method-invocation.html.

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -187,18 +188,34 @@ class MarathonScheduler @Inject() (
...
+   * leader will fail to find a stored FrameworkID and invoke `register`
+   * instead of `reregister`.  This is important because on certain kinds
+   * of framework errors (such as exceeding the framework failover timeout),
+   * the scheduler may never re-register with the saved FrameworkID until
+   * the leading Mesos master process is killed.
+   */
+  private def suicide(removeFrameworkId: Boolean): Unit = {
+    log.fatal(s"Committing suicide!")
+
+    if (removeFrameworkId) Await.ready(frameworkIdUtil.expunge, config.zkTimeoutDuration)
```
ConnorDoyle: `.expunge()` (with parens)

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -187,19 +188,36 @@ class MarathonScheduler @Inject() (
...
+    * leader will fail to find a stored FrameworkID and invoke `register`
+    * instead of `reregister`.  This is important because on certain kinds
+    * of framework errors (such as exceeding the framework failover timeout),
+    * the scheduler may never re-register with the saved FrameworkID until
+    * the leading Mesos master process is killed.
+    */
+  private def suicide(removeFrameworkId: Boolean): Unit = {
+    log.fatal(s"Committing suicide!")
+
+    if (removeFrameworkId) Await.ready(frameworkIdUtil.expunge, config.zkTimeoutDuration)
```
ConnorDoyle: `.expunge()` (with parens)

ConnorDoyle:
Fix it, then :shipit:

aquamatthias:
@ConnorDoyle Incorporated changes. Thanks for the feedback.

PR #1697, opened 2015-06-22
Remove the publishing process to s3.

aquamatthias:
It worked from Dario's local machine, was never integrated into TeamCity process and does not make that much sense anyway.


kolloch:
Are you sure that none of our builds use that?

kolloch:
I didn't see references at least.

PR #1698, opened 2015-06-22
Remove the frontend travis build setup

aquamatthias:
That has moved to the separate UI project and should not be part of the service test layer.
Also only test is performed and no assembly and doc, since this increases the build time but gains not much information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,9 @@
 language: scala
-before_install:
-  - git submodule update --init --recursive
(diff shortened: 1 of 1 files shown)


kolloch:
At least the doc target does add information. I added it myself because before that we only got errors about incorrect doc markup when we released a new Marathon version.

aquamatthias:
Ah good point. Added the doc target again.

PR #1718, opened 2015-06-27
Fixes #1716 - Add Proxy headers

### LINKED ISSUE #1716: Add headers to Marathon responses indicating current leader and proxying
We currently pass all our requests through the `LeaderProxyFilter` which proxies to the current leader.

It would be nice to make it more explicit by adding to headers:

* `X-Marathon-Leader` should point to the current leader (as a URL)
* `X-Marathon-ProxiedBy` should include the non-leader that proxied the request to the leaders (as a URL)

kolloch:
X-Marathon-Leader points to current leader and X-Marathon-Via is now also
set in the proxy responses.


aquamatthias:
LGTM

PR #1743, opened 2015-07-02
Use a synchronized GroupManager.upgrade for all changes

### LINKED ISSUE #1709: Concurrent changes to the same AppDefinition will lead to inconsistent state
One customer makes changes to the same AppDefinition concurrently (simultaneous requests every 10-100 millisecs). This will result in completely broken AppDefinitions.

aquamatthias:
Fixes #1709


kolloch:
Please rebase

meichstedt:
In general much more readable than before. Few points I commented:
- there is a new but ignored test in AppsResourceTest. Doesn't run if its active – comment on that?
- It might make sense to test response entities somewhere. as far as I can see, this is seldomly done.
- (this was the same before) not very nice that we need to manually create a DeploymentPlan and schedule that for deployment after the group has been updated, since this is expected to be scheduled during the group update. Create a ticket for that?

PR #1749, opened 2015-07-06
Constraints: add selectTasksToKill based on constraints.

aquamatthias:
Idea: only the groupBy constraint has to be respected -> please prove!

### DIFF
--- a/src/main/scala/mesosphere/mesos/Constraints.scala
+++ b/src/main/scala/mesosphere/mesos/Constraints.scala
@@ -1,5 +1,7 @@
 package mesosphere.mesos
 
+import mesosphere.marathon.state.AppDefinition
+
 import scala.collection.JavaConverters._
 import mesosphere.marathon.Protos.{ MarathonTask, Constraint }
 import mesosphere.marathon.Protos.Constraint.Operator
@@ -82,13 +84,12 @@ object Constraints {
           (value.isEmpty || attr.get.getText.getValue == value) &&
             // All running tasks should have the matching attribute
             matches.size == tasks.size
-        case Operator.GROUP_BY => {
+        case Operator.GROUP_BY =>
           val groupFunc = (task: MarathonTask) =>
             task.getAttributesList.asScala
               .find(_.getName == field)
               .map(_.getText.getValue)
           checkGroupBy(attr.get.getText.getValue, groupFunc)
-        }
         case Operator.LIKE =>
           if (value.nonEmpty) {
             attr.get.getText.getValue.matches(value)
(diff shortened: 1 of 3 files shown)


aquamatthias:
Fixes: #555

kolloch:
Hi @aquamatthias, can you use "Fixes #555 - ..." as commit message format? Or suggest another standard for us indicating the fixed issue in the commit message.

Would be cool.

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -125,4 +126,85 @@ object Constraints {
...
+    * Helper class for easier distribution computation.
+    */
+  private case class GroupByDistribution(constraint: Constraint, distribution: Seq[Set[MarathonTask]]) {
+    def isMoreEvenWithout(selected: Set[MarathonTask]): Boolean = {
+      val diffAfterKill = distributionDifference(selected)
+      //diff after kill is 0=perfect, 1=tolerated or minimizes the difference
+      diffAfterKill <= 1 || distributionDifference() > diffAfterKill
+    }
+
+    def selectTaskToKill(without: Set[MarathonTask], butNot: Set[MarathonTask] = Set.empty): Iterator[MarathonTask] = {
```
gkleiman: is the butNot parameter really needed? It looks like the default is always used.

gkleiman: Also the name of the method is a bit misleading. It returns an iterator over potentially multiple tasks, not just one task as the name implies.

gkleiman: To make it easy to read, I'd add a description of what the method does:

Removes from distribution the tasks in the "without"  parameter and returns:

- An iterator over an empty sequence if the distribution is even.
- An iterator over all the tasks belonging to the value with the largest amount of members otherwise.

kolloch: Also: Without having looked at the context: Avoid using an iterator as a return value since they are mutable. If it is really important that an iterator is used, consider renaming the method to something indicating an iterator.

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -125,4 +126,85 @@ object Constraints {
...
+
+    //short circuit, if there are no constraints to align with
+    if (distributions.isEmpty) return Set.empty
+
+    var toKillTasks = Set.empty[MarathonTask]
+    var flag = true
+    while (flag && toKillTasks.size != toKillCount) {
+      val tried =
+        distributions.toSeq.sortBy(_.distributionDifference(toKillTasks)).reverseIterator //highest diff first
+          .flatMap(_.selectTaskToKill(toKillTasks)) ++ //select all tasks to kill
```
gkleiman: What about changing the comment to something a bit more descriptive like:

For each distribution, select the tasks of the "largest" value group. The tasks in "toKillTasks" will not be counted.

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
```diff
@@ -3,52 +3,79 @@ package mesosphere.mesos
...
 
   import mesosphere.mesos.protos.Implicits._
 
-  def makeSampleTask(id: String, attrs: Map[String, String]) = {
-    val builder = MarathonTask.newBuilder()
-      .setHost("host")
-      .addAllPorts(Lists.newArrayList(999))
-      .setId(id)
+  test("Select tasks to kill for a single group by works") {
+    Given("app with hostname group_by and 50 tasks even distributed on 5 hosts")
```
gkleiman: s/5/2/g

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
```diff
@@ -3,52 +3,79 @@ package mesosphere.mesos
...
-  def makeOffer(hostname: String, attributes: Iterable[Attribute]) = {
-    Offer.newBuilder
-      .setId(OfferID(Random.nextString(9)))
-      .setSlaveId(SlaveID(Random.nextString(9)))
-      .setFrameworkId(FrameworkID(Random.nextString(9)))
-      .setHostname(hostname)
-      .addAllAttributes(attributes.asJava)
-      .build
+  test("Select only tasks to kill for an unbalanced distribution") {
+    Given("app with hostname group_by and 50 tasks even distributed on 5 hosts")
```
gkleiman: This statement is a lie! ;-)

gkleiman:
LGTM

I agree with Peter that it would be good if we agree on a convention for the commit messages. In this case it looks like squashing the commits might also improve the commit history after merging.

gkleiman:
:+1:

PR #1751, opened 2015-07-06
Fix flaky integration tests.

gkleiman:
The tests in the EventSubscribersIntegrationTest suite were flaky, because they
tried to register foobar.com as an endpoint. Fixed them by using localhost
instead of foobar.com.


aquamatthias:
LGTM. There still is a good chance, that we have something local running on port 1337. But perhaps we can ignore this for the moment.

PR #1753, opened 2015-07-06
Fix #1681: Allow administratively getting and resetting the taskLaunchDelay

### LINKED ISSUE #1681: Allow administratively zeroing / resetting the taskLaunchDelay
Currently there is no clean way to change the taskLaunchDelay after Marathon has backed off an app.

One instance that bit recently was a couple apps had been failing to launch for ~10 hours because some slaves had failed. After fixing the slaves, marathon had the taskLaunchDelay set to 52 minutes, and there is no simple way to ask Marathon to reset all apps (there could be dozens / hundreds / thousands) taskLaunchDelay so that things will launch to quickly get services back up after fixing a root issue. 

Generated log line: `Jun 17 20:50:08 marathon[2277]: [2015-06-17 20:50:08,802] INFO Delaying /tools/redirect-server due to backoff. Time left: 52 minutes 55 seconds 995 milliseconds. (mesosphere.marathon.tasks.TaskQueue:113)`

From some discussion with @kolloch and @ConnorDoyle it seems like it would be nice to be able to set this on both a per-app level, as well as per-group. Setting it on the root group would effectively set it for all apps. From a incident response perspective I need the "all apps" variant.

aquamatthias:
The queue endpoint holds the duration.
A new endpoint is added to allow resetting the task launch duration.

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/TaskQueue.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/TaskQueue.scala
@@ -129,4 +129,12 @@ class TaskQueue {
 
     findMatching(sorted)
   }
+
+  /**
+    * Reset the rate limiting delay for the given app definition
+    * @param app the app definition to reset.
+    */
+  def resetDelay(app: AppDefinition): Unit = {
+    rateLimiter.resetDelay(app)
+  }
 }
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/tasks/TaskQueue.scala
```diff
@@ -129,4 +129,12 @@ class TaskQueue {
 
     findMatching(sorted)
   }
+
+  /**
+    * Reset the rate limiting delay for the fgiven app definition
```
gkleiman: s/fgiven/given/

gkleiman:
As we just spoke, the following would be nice:

* Changing the endpoint to DELETE /v2/queue/<appId>/delay
* Adding an integration test. This can be done in another PR in order to unblock the UI work.
* Adding a unit tests that check the timeLeftSeconds value (including one with a deadline in the past, which could result in a negative value).

gkleiman:
Thanks!

PR #1755, opened 2015-07-06
Speedup PortsMatcher

kolloch:
Scala Range contains is implemented naively.
So don't use that.

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -78,8 +78,8 @@ class PortsMatcher(
     */
   private[this] def findPortsInOffer(requiredPorts: Seq[Integer], failLog: Boolean): Option[Seq[PortWithRole]] = {
     takeEnoughPortsOrNone(expectedSize = requiredPorts.size) {
-      requiredPorts.sorted.iterator.map { port =>
-        offeredPortRanges.find(_.range.contains(port)).map { offeredRange =>
+      requiredPorts.sorted.iterator.map { (port: Integer) =>
+        offeredPortRanges.find(_.contains(port)).map { offeredRange =>
           PortWithRole(offeredRange.role, port)
         } orElse {
           if (failLog) log.info(s"Couldn't find host port $port in any offered range for app [${app.id}]")
@@ -124,8 +124,8 @@ class PortsMatcher(
             Option(availablePortsWithoutStaticHostPorts.next())
           }
         case pm: PortMapping =>
-          offeredPortRanges.find(_.range.contains(pm.hostPort)) match {
-            case Some(PortRange(role, _)) =>
+          offeredPortRanges.find(_.contains(pm.hostPort)) match {
+            case Some(PortRange(role, _, _)) =>
               Some(PortWithRole(role, pm.hostPort))
             case None =>
               log.info(s"Cannot find range with host port ${pm.hostPort} for app [${app.id}]")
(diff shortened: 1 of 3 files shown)


aquamatthias:
Just curious: what is naiv in the implementation?
```scala
  final def contains(x: Int) = {
    if (x==end && !isInclusive) false
    else if (step > 0) {
      if (x < start || x > end) false
      else (step == 1) || (((x - start) % step) == 0)
    }
    else {
      if (x < end || x > start) false
      else (step == -1) || (((x - start) % step) == 0)
    }
  }
```

kolloch:
I found the code by profiling (took 20% of CPU time!). I looked at the code that my IDE pointed me to but of course it is overridden in the scala.collection.immutable.Range#contains method. 

Anyways, after changing this, it didn't show up by profiling anymore.

I agree with you that it is strange and I do not understand it. Any ideas?

kolloch:
I realize my testing is not very scientific but the impact is quite obvious. A sampled for 60s, in the version with the patch the deployment was finished before the interval. Without the patch, it was busy the whole time.

# Sample App

```json
{
        "id": "/something/app/service",
        "cmd": "python -m SimpleHTTPServer $PORT",
        "cpus": 0.0001,
        "mem": 8.0,
        "ports": [0]
}
```

# Reproduction Steps

Using httpie.

I also use the `pk/port-matcher-test`, since it's hard to otherwise really execute that code in any wilder fashion ;)

* Start Marathon in Simulation Mode:

```bash
sbt -Djava.library.path=<...include the directories of your native mesos libraries...> \
    "project mesosSimulation" "run --master zk://localhost:2181/mesos --zk zk://localhost:2181/marathon"
```

* Warmup
  * `http --json PUT :8080/v2/apps/something/app/service?force=true instances:=10000`
  * after deployment, scale back:  `http --json PUT :8080/v2/apps/something/app/service?force=true instances:=1`
* Testing
  * `http --json PUT :8080/v2/apps/something/app/service?force=true instances:=10000`
  * Start measuring `jmc` (Mission Control) 

# Results
## With Patch (8837c6b2dc489083795fbec6cf2761f47fba6977)

No entries above >= 5%!

## Without Patch (0ab8a753917fba8a8cbdc9c2e5d3e04cb1fe874b)

Mission Control / Code / Hot Methods, top entries, partially expanded:

```
Stack Trace	Sample Count	Percentage(%)
scala.collection.Iterator$class.exists(Iterator, Function1)	20	26.316
   scala.collection.AbstractIterator.exists(Function1)	20	26.316
      scala.collection.IterableLike$class.exists(IterableLike, Function1)	20	26.316
         scala.collection.AbstractIterable.exists(Function1)	20	26.316
            scala.collection.SeqLike$class.contains(SeqLike, Object)	20	26.316
               scala.collection.AbstractSeq.contains(Object)	20	26.316
                  mesosphere.marathon.tasks.PortsMatcher$$anonfun$findPortsInOffer$1$$anonfun$apply$6.apply(PortsMatcher$PortRange)	20	26.316
                     mesosphere.marathon.tasks.PortsMatcher$$anonfun$findPortsInOffer$1$$anonfun$apply$6.apply(Object)	20	26.316
scala.collection.IndexedSeqLike$Elements.next()	9	11.842
java.lang.Integer.valueOf(int)	5	6.579
akka.actor.LightArrayRevolverScheduler$$anon$8.nextTick()	3	3.947
org.apache.zookeeper.ClientCnxn$SendThread.run()	2	2.632
```

kolloch:
Ok, I understood the original problem with @aquamatthias help:

Range has two different `contains` methods. The efficient one only applies directly to `Int`. The generic one also applies to all values that are convertible to Int. Since we passed a `java.lang.Integer` before this patch, the generic method was used.

PR #1758, opened 2015-07-07
Make integration tests a bit more robust.

gkleiman:
- Used the "eventually" trait to try to make the integration tests less flaky.
- scalastyle fixed the formatting of MarathonClusterIntegrationTest.
- Removed the parentheses from the MarathonFacade#leader definition.

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/EventSubscribersIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/EventSubscribersIntegrationTest.scala
@@ -17,7 +17,7 @@ class EventSubscribersIntegrationTest
     marathon.subscribe("http://localhost:1337").code should be(200)
 
     Then("a notification should be sent to all the subscribers")
-    waitForEvent("subscribe_event").info should contain("callbackUrl" -> "http://localhost:1337")
+    waitForEventWith("subscribe_event", _.info.exists(_ == "callbackUrl" -> "http://localhost:1337"))
 
     And("the subscriber should show up in the list of subscribers")
     marathon.listSubscribers.value.urls should contain("http://localhost:1337")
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/marathon/integration/EventSubscribersIntegrationTest.scala
```diff
@@ -17,7 +22,7 @@ class EventSubscribersIntegrationTest
     marathon.subscribe("http://localhost:1337").code should be(200)
 
     Then("a notification should be sent to all the subscribers")
-    waitForEvent("subscribe_event").info should contain("callbackUrl" -> "http://localhost:1337")
+    eventually { waitForEvent("subscribe_event").info should contain("callbackUrl" -> "http://localhost:1337") }
```
aquamatthias: why eventually?
waitForEvent is doing the same as eventually - it actively waits for an event to arrive. 
You basically extend the timeout, or do I miss something?

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
```diff
@@ -17,7 +29,8 @@ class LeaderIntegrationTest extends IntegrationFunSuite with MarathonClusterInte
 
   test("the leader abdicates when it receives a DELETE") {
     Given("a leader")
-    val leader = marathon.leader().value
+    eventually { marathon.leader.code should be (200) }
```
aquamatthias: The response will not change, after it is created. eventually trying again does not make sense.

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
```diff
@@ -27,6 +40,6 @@ class LeaderIntegrationTest extends IntegrationFunSuite with MarathonClusterInte
     (result.entityJson \ "message").as[String] should be ("Leadership abdicted")
 
     And("the leader must have changed")
-    marathon.leader().value shouldNot be(leader)
+    eventually { marathon.leader.value shouldNot be(leader) }
```
aquamatthias: I like Eventually, but we already have WaitTestSupport. We should strive for consistency.

Issue #375, opened 2014-07-02
Support for arbitrary resource types


ConnorDoyle:
Apache Mesos allows user-defined resource types via the  
`--resources=name(role):value;name(role):value...`  
argument to the `mesos-slave` program.  It would be great if there were a way to require these as part of an app's definition in the Marathon API, in addition to the standard resource types (`cpus`, `mem`, `disk`, `ports`) already supported.

syogesh:
Is anyone working on this? If not I would like to take this up. Is there anything I need to know before I start?

ConnorDoyle:
That's great!  Nobody is currently working on this (so far as I know).  You might save some time by posting your plan and some proposed API examples before you begin hacking.

syogesh:
I might only have time to add support for scalar custom resources is that alright? 

*Basic Plan*
AppDefinition will get a customResources map that maps from resourceName to value. ResourceMatcher will check that the app's custom resources exist in the offer.

*API Example*
The API change will be minimal. Like cpus and mem, there will be an additional key "customResources" on the create /v2/apps POST request that maps from resource name to value. Ex:

{
    ...,
    "cpus": 2.0,
    "customResources": { "bandwidth": 100, "foo": 2 },
    ...
}

syogesh:
Here's an alternate format which is much more extensible and supports the other resource types:

```
"customResources": [
                    {
                        "name": "foo",
                        "type": "scalar",
                        "value": 100.0
                    },
                    {
                        "name": "bar",
                        "type": "range",
                        "value": [100, 200]
                    }
]
```

kolloch:
That looks pretty good. The range resources can specify multiple ranges, though.

syogesh:
Ah, okay. Also I want to change the format to make it more concise, have a structure similar to Resources, and not have to deal with the same key having multiple possible types. How about this: 
```
"customResources": [
                    {
                        "name": "foo",
                        "scalar": 100.0
                    },
                    {
                        "name": "bar",
                        "range": [[100, 200], [300, 400]]
                    },
                    {
                        "name": "foobar",
                        "set": ["a", "b"]
                    }
                   ]
```

kolloch:
Ranges are difficult. What are the use cases besides ports? Maybe IPs?

Anyways, I stick with the port use case for now, even though I am not sure that it generalizes to other use cases.

[...]

PR #1759, opened 2015-07-07
Support for arbitrary resources (fixes issue #375)

syogesh:
Fixes: https://github.com/mesosphere/marathon/issues/375

This is my first time working with Scala so please suggest anything that I can improve with this request. Also this is kind of a pre-emptive pull request in the sense that I want to confirm that the changes I'm making are the preferred way to do things. Any requests for additional tests are welcome.

Example api request:

```
{
  "id": "arbitrary-resource-example",
  "cmd": "./run.sh",
  "cpus": 2.0,
  "mem": 2048.0,
  "urls": ["http://files.my.org/program.tar.gz"],
  "customResources": {
    "foo": {
      "scalar": {
        "value": 100.0
      }
    },
    "bar": {
      "set": {
        "value": ["a", "b", "c"],
        "numberRequired": 1
      }
    },
    "baz": {
      "ranges": {
        "value": [
          {
            "begin": 100,
            "end": 200,
            "numberRequired": 2
          },
          {
            "numberRequired": 1
          }
        ]
      }
    }
  }
}
```


aquamatthias:
Thanks for this PR. We will try to look into this.

kolloch:
Hi @syogesh , thank you very much for your PR. Unfortunately, we are not ready quite yet. See the discussion in https://github.com/mesosphere/marathon/issues/375. After agreeing on an API format, I would like to move ahead with a PR focussing on scalar custom resources only.

PR #1767, opened 2015-07-08
Improve documentation for offer matching tuning CLI flags

kolloch:


pierlo-upitup:
One question @kolloch :
Does the "v0.8.2" label before the arguments mean that they apply ONLY to that version, or SINCE that version?

kolloch:
It means, "since". Maybe we change it into >0.8.2?

There was also some criticism of the approach. Maybe we should have separate pages per marathon version.

pierlo-upitup:
Yeah, on the other hand `>0.8.2` actually might suggest "greather than" (and not equal to)... 
I guess until there is consensus it's OK to keep it in-line with the rest of the document. Mergin'!

PR #1774, opened 2015-07-09
Change description of "Companies using Marathon"

janisz:
Remove Allegro Group from list of companies using Marathon. We use it but not at 100,000 nodes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ To develop on the web UI look into the instructions of the [Marathon UI](https:/
 
 ## Companies using Marathon
 
-Marathon is managing applications on more than 100,000 nodes at these companies, and many more:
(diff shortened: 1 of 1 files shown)


gkleiman:
We don't want to claim that each one of the companies use Marathon with more than 100,000 nodes, but that it is managing over those many nodes *across* all installations. However I do see how that sentence can be misinterpreted.

What would you think about keeping Allegro Group in the list and changing the sentence to this:

"Across all installations Marathon is managing applications on more than 100,000 nodes world-wide. These are some of the companies using it:"

janisz:
Sounds good to me.

gkleiman:
:+1: Thanks for updating the PR. Would it be possible for you to squash the commits?

janisz:
Done

gkleiman:
Thanks!

PR #1792, opened 2015-07-13
Port guessing for servicerouter.py

woegjiub:
Splits the pingback address, uses the port from the end if present.


lloesche:
@woegjiub Thank you! LGTM.

PR #1807, opened 2015-07-15
Fixes #1254 - Add flag to define Marathon PORT prefix.

### LINKED ISSUE #1254: Add flag to define Marathon PORT prefix
Marathon overrrides the PORT env inside the container. This can be a nice feature but is also causing trouble if you need to connect to the port the container app did set to PORT.

A flag to set env prefix for the ports marathon sets would solve this.

gkleiman:


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -91,7 +91,7 @@ case class AppDefinition(
   }
 
   def toProto: Protos.ServiceDefinition = {
-    val commandInfo = TaskBuilder.commandInfo(this, None, None, Seq.empty)
+    val commandInfo = TaskBuilder.commandInfo(this, None, None, Seq.empty, None)
```
kolloch: Named a parameters would be nice here. Intellij can do that automatically.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -178,12 +179,16 @@ class TaskBuilder(app: AppDefinition,
...
+                  taskId: Option[TaskID],
+                  host: Option[String],
+                  ports: Seq[Long],
+                  envPrefix: Option[String]): CommandInfo = {
     val containerPorts = for (pms <- app.portMappings) yield pms.map(_.containerPort)
     val declaredPorts = containerPorts.getOrElse(app.ports)
     val envMap: Map[String, String] =
       taskContextEnv(app, taskId) ++
-        portsEnv(declaredPorts, ports) ++ host.map("HOST" -> _) ++
+        addPrefix(envPrefix, portsEnv(declaredPorts, ports)) ++ host.map("HOST" -> _) ++
```
kolloch: Mention "HOST" as exception in the documentation?

kolloch:
Explain/Reference new configuration option here as well: task-environment-vars.md

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -572,7 +572,8 @@ class TaskBuilderTest extends MarathonSpec {
         ),
         Some(TaskID("task-123")),
         Some("host.mega.corp"),
-        Seq(1000, 1001)
+        Seq(1000, 1001),
+        None
```
kolloch: use named arguments at least for things like "None"

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -633,6 +636,32 @@ class TaskBuilderTest extends MarathonSpec {
...
+    assert("1000" == env("CUSTOM_PREFIX_PORT"))
+
+    assert(!env.contains("PORT0"))
+    assert("1000" == env("CUSTOM_PREFIX_PORT0"))
+    assert("1000" == env("CUSTOM_PREFIX_PORT_8080"))
+
+    assert(!env.contains("PORT1"))
+    assert("1001" == env("CUSTOM_PREFIX_PORT1"))
+    assert("1001" == env("CUSTOM_PREFIX_PORT_8081"))
+  }
```
kolloch: Explicitly check for all Variables not having the prefix and compare it against a list of exceptions.

That's more future proof in my opinion.

kolloch:
Also add a changelog entry.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -267,6 +272,13 @@ object TaskBuilder {
     }
   }
 
+  def addPrefix(envVarsPrefix: Option[String], env: Map[String, String]): Map[String, String] = {
```
kolloch: You COULD write this as:

```scala
envVarsPrefix.fold(env)(prefix => env.map { case (k,v) => (prefix + k) -> v })
```

but your version is fine, too

gkleiman:
Here goes an updated version.

If you like this, then I'll write something for the changelog and ask Matthias to add it to mv/changelog-0.10.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -619,24 +621,55 @@ class TaskBuilderTest extends MarathonSpec {
...
+    assert("1000" == env("CUSTOM_PREFIX_PORT0"))
+    assert("1000" == env("CUSTOM_PREFIX_PORT_8080"))
+
+    assert("1001" == env("CUSTOM_PREFIX_PORT1"))
+    assert("1001" == env("CUSTOM_PREFIX_PORT_8081"))
+
+    assert("host.mega.corp" == env("CUSTOM_PREFIX_HOST"))
+
+    assert(Seq("HOST", "PORTS", "PORT0", "PORT1").forall(k => !env.contains(k)))
+    assert(Seq("MESOS_TASK_ID", "MARATHON_APP_ID", "MARATHON_APP_VERSION").forall(env.contains))
```
kolloch: I would rather check for all variables that do not start with the prefix and ascertain that all exceptions are in the list.

If you add an unprefixed variable now, no test will fail.

kolloch:
LGTM. Don't forget the changelog!

PR #1821, opened 2015-07-16
Fixes #1820: the default for missing headers is an empty enumeration instead of null

### LINKED ISSUE #1820: The SetRequestDefaultsFilter does not handle missing headers correctly
The contract for missing headers in the ServletRequest is as follows:

- `getHeader` - returns null
- `getHeaders` - returns an empty enumeration.

The latter returns null which is wrong.

aquamatthias:


gkleiman:
LGTM!

PR #1824, opened 2015-07-17
change permissions and add logging

lloesche:
The default package permissions of /etc/haproxy/haproxy.cfg are 0644 on EL and Debian (http://pkgs.fedoraproject.org/cgit/haproxy.git/tree/haproxy.spec#n131). However after running servicerouter they are changed to 0600. This fixes that.

It also adds some additional output which is useful for marathon installations with lots of tasks.

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -435,13 +435,16 @@ class Marathon(object):
         return self.api_req('POST', ['apps'], app_json)
 
     def get_app(self, appid):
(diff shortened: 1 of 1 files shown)


kolloch:
Wouldn't it be cooler if it preserved the permissions of the config_file? E.g. read the permissions of that file and apply them to the temporary file?

lloesche:
@kolloch great idea! implemented

joerg84:
LGTM

PR #1826, opened 2015-07-17
servicerouter: make haproxy port overrideable

lloesche:
In Marathon the service port has to be unique.

The use case is to bind a webserver on two different IPs on the same Port. For the IPs I can already define HAPROXY_{0}_BIND_ADDR.

This PR adds HAPROXY_{n}_PORT so I can also specify which port haproxy should bind on.


joerg84:
LGTM

bin/servicerouter.py
```diff
@@ -59,6 +59,11 @@
     Bind to the specific address for the service.
     Ex: HAPROXY_0_BIND_ADDR = '10.0.0.42'
 
+  HAPROXY_{n}_PORT
```
joerg84: We should have a consistent order. Below it is always Mode followed by port...

joerg84:
LGTM

PR #1833, opened 2015-07-20
Fixes #1829 - Only override acceptedResourceRoles when set in update

### LINKED ISSUE #1829: Marathon forgets value for `acceptedResourceRoles` when app is updated (including scaling)
1. Start an app with `acceptedResourceRoles` set.
1. Update it (doesn't matter how).
1. Observe the value of `acceptedResourceRoles` is set to `null`.

kolloch:


aquamatthias:
LGTM

PR #1834, opened 2015-07-20
Build for jdk 8 target.

aquamatthias:
Docker: rely on the official jdk 8 container and install mesos afterwards.
IntegrationTests: set max heap mem explicitly for all launched applications.

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/setup/ForwarderService.scala
+++ b/src/test/scala/mesosphere/marathon/integration/setup/ForwarderService.scala
@@ -76,6 +76,7 @@ object ForwarderService {
 
     ProcessKeeper.startJavaProcess(
       s"app_${conf.httpPort()}",
+      heapInMegs = 128,
       arguments = List(ForwarderService.className, "helloApp") ++ args,
       upWhen = _.contains("Started SelectChannelConnector"))
   }
@@ -85,6 +86,7 @@ object ForwarderService {
 
     ProcessKeeper.startJavaProcess(
       s"forwarder_${conf.httpPort()}",
+      heapInMegs =  128,
       arguments = List(ForwarderService.className, "forwarder", forwardToPort.toString) ++ args,
       upWhen = _.contains("SelectChannelConnector@"))
   }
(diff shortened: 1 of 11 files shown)


src/test/scala/mesosphere/marathon/integration/setup/ForwarderService.scala
```diff
@@ -75,7 +75,7 @@ object ForwarderService {
     val conf = createConf(args: _*)
 
     ProcessKeeper.startJavaProcess(
-      s"app_${conf.httpPort()}",
+      s"app_${conf.httpPort()}", 128,
```
kolloch: Please use a named parameter for the "128".

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
```diff
@@ -47,7 +47,7 @@ object ProcessKeeper {
     val workDirFile = new File(workDir)
     FileUtils.deleteDirectory(workDirFile)
     FileUtils.forceMkdir(workDirFile)
-    startJavaProcess("zookeeper", args, new File("."), sys.env, _.contains("binding to port"))
+    startJavaProcess("zookeeper", 256, args, new File("."), sys.env, _.contains("binding to port"))
```
kolloch: named parameter, please

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
```diff
@@ -73,17 +73,18 @@ object ProcessKeeper {
     FileUtils.forceMkdir(mesosWorkDirFile)
 
     startJavaProcess(
-      "marathon", argsWithMain, cwd,
+      "marathon", 512, argsWithMain, cwd,
```
kolloch: named parameter, please

kolloch:
Besides my wish for named parameters, that looks excellent!

kolloch:
My only concern is that our tests might not be reproducible because we just use the latest available mesos version. Maybe we can pin the version there?

kolloch:
Maybe we address pinning mesos versions when we have proper cross-testing, though.

aquamatthias:
Pinned the mesos version.
Added named parameters.

kolloch:
LGTM!

PR #1838, opened 2015-07-20
Set 0.9.0 version in docs.

aquamatthias:


docs/docs/index.md
```diff
@@ -92,13 +66,27 @@ give us feedback (particularly on violations of the versioning policy).
...
+1. Start the new version of Marathon and watch the log for a successful start.  
+
+#### Upgrading an HA installation
+Upgrading to a newer version of Marathon should be executed in that order:
+
+1. Tear down all running instances of Marathon except one. This instance will be the leader.
+1. Install the new version of Marathon on one of the nodes with the old version.
+1. Start the instance with the new version of Marathon.
+1. Stop the last node with the old version. Now the new version of Marathon will take over leadership and becomes active.
+1. Watch the log of this instance for a successful start
```
kolloch: It would be cool if we were a bit more specific. How do we look for a successful start? Log line + Checking for ERROR log statements maybe?

Being very picky: There is a full stop missing at the end of this item.

docs/docs/index.md
```diff
@@ -92,13 +66,27 @@ give us feedback (particularly on violations of the versioning policy).
 
 ### Upgrading to a Newer Version
 
-Upgrading to a newer version of Marathon should be seamless.
+We generally recommend creating a backup of the ZooKeeper state before upgrading to be able to downgrade in case of problems after an upgrade. This can be done by creating a copy of ZooKeeper's [data directory](http://zookeeper.apache.org/doc/r3.1.2/zookeeperAdmin.html#The+Data+Directory).
```
kolloch: I love that you link to further information here!

kolloch:
Much better than before. Thank you!

kolloch:
It would be cool, if we could move this to a separate `updating.md` page and link this instead or in addition to "Upgrading/0.6.x to 0.7.0" in the side bar.

kolloch:
Decide if you want to do that as part of this PR or later. I would also merge as is since it is definitely an improvement.

aquamatthias:
@kolloch I added a sentence how to discover a successful start. I agree with all other points as well but would like to defer those changes and have that available soon.

kolloch:
Updated our gh-pages page:

https://mesosphere.github.io/marathon/docs/

Issue #1563, opened 2015-05-22
Inconsistent /help <-> rest-api.md documentation


kolloch:
The documentation accessible via the /help endpoint is not in sync with the rest-api.md.

It is not clear which documentation is more up-to-date. I guess these versions have to be merged.

A solution should be found to prevent this kind of inconsistency in the future, for example by automatically generating the rest-api.md from the /help documents or by adding a build CI check that checks the consistency.

"#1148 Fix REST API doc" seems to be related but is actually much broader in scope. This is why I created this issue focussing on consistency.

Solving this issue, should also solve "#1356 Missing /help pages".

air:
@kolloch reached out to the mailing list but we saw no replies, https://groups.google.com/forum/?hl=en#!searchin/marathon-framework/help/marathon-framework/vAk6_rTaV2U/qAW94sTnodUJ

air:
Current state:
- /help is not discoverable or linked from anywhere. The 'Docs' link in the UI takes you to [the REST API docs](https://mesosphere.github.io/marathon/docs/).
- /help seems to have the same content as rest-api.md but with formatting removed.
- /help is in fixed HTML format, which does not adjust to the Accept: header of the client (e.g. Use curl and you get HTML, not text).
  - Compare with Mesos /help which will return Content-Type: text/x-markdown or HTML depending on the client.
- /help has many broken links. Try the 3rd, 4th or 5th link in the page.
- On the other hand, the static REST docs are not highly readable. #1148 is important to improve this (a separate issue!).

air:
My current thinking:

1. We should make the assumption that the user has enough internet access to look at [static docs](https://mesosphere.github.io/marathon/docs/). It is not practical to be a developer/operator without some internet access (a laptop, a phone).
1. If we had to choose a single doc source it would be static docs, because 1. they’re indexed by Google and 2. they don’t require a running Marathon to read.
1. Having a versatile /help endpoint can be useful for exploring the API. Mesos /help is a good example.
1. The Marathon /help docs aren't particularly versatile/explorable; they duplicate static docs. 

Actions:

1. Question: Are there any ways in which the Marathon /help docs are better than the REST API docs? If so, let's improve the REST API docs to get parity (see #1148).
1. Merge the /help content out to the REST API docs (i.e. migrate out any unique content).
1. Replace the /help endpoint with a Redirect to the static docs URL.

Thoughts please.

aquamatthias:
[...]

PR #1841, opened 2015-07-20
Remove extra slash in @Path for artifacts

bobrik:
#1471 is related, I've found new ones in 0.10-RC3.


kolloch:
Thanks, looks good. We are waiting for the integration tests.

Unfortunately, we have basically abandoned the documentation in `/help` because some of us didn't know about that doc.

See #1563. We could use help with that.

PR #1850, opened 2015-07-22
Use sudo: false in travis.yml

kolloch:
to use container-based builds.

See
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade


meichstedt:
LGTM

PR #1852, opened 2015-07-22
Fixes #1818 - Make Marathon fail on startup errors

### LINKED ISSUE #1818: Marathon should fail during start, if the HTTP port is already bound
In the case of an occupied port, Marathon starts up successfully and announces the host and port in the cluster. But this endpoint is not reachable from any external service. 

Marathon should fail fast and shutdown the service in this case.

gkleiman:
Make Marathon fail and exit if any of the services can't be started, e.g., if
another process is already bound to the the HTTP port.

### DIFF
--- a/src/main/scala/mesosphere/marathon/Main.scala
+++ b/src/main/scala/mesosphere/marathon/Main.scala
@@ -1,9 +1,10 @@
 package mesosphere.marathon
 
-import com.google.inject.Module
+import com.google.common.util.concurrent.ServiceManager
+import com.google.inject.{ Guice, Module }
 import com.twitter.common.quantity.{ Amount, Time }
 import com.twitter.common.zookeeper.ZooKeeperClient
-import mesosphere.chaos.{ App, AppConfiguration }
+import mesosphere.chaos.AppConfiguration
 import mesosphere.chaos.http.{ HttpConf, HttpModule, HttpService }
 import mesosphere.chaos.metrics.MetricsModule
 import mesosphere.marathon.api.MarathonRestModule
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -81,14 +87,39 @@ class MarathonApp extends App {
...
+      case Failure(f) =>
+        log.fatal(s"Failed to start all services. Services by state: ${serviceManager.servicesByState()}", f)
+        shutdownAndWait()
+        System.exit(1)
+    }
+  }
+
+  private def shutdownAndWait(): Unit = {
+    serviceManager.foreach(_.stopAsync())
+    serviceManager.foreach(_.awaitStopped())
```
kolloch: you should set `serviceManager` to `None` afterwards to prevent shutting it down twice (via the shutdown hook and directly), even though it does not appear to cause harm.

Some kind of logging "Now shutting down." "Shutdown completed." or something like that would be nice.

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -81,14 +87,39 @@ class MarathonApp extends App {
...
+      case Success(_) =>
+        log.info("All services up and running.")
+      case Failure(f) =>
+        log.fatal(s"Failed to start all services. Services by state: ${serviceManager.servicesByState()}", f)
+        shutdownAndWait()
+        System.exit(1)
+    }
+  }
+
+  private def shutdownAndWait(): Unit = {
```
kolloch: Does this override the method in the parent class? Strange that you can declare it private. (Why not private[this], though?)

I think it would make sense to actually adjust the chaos code but I understand that you find that annoying.

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -81,14 +87,39 @@ class MarathonApp extends App {
...
+
+    lazy val injector = Guice.createInjector(modules().asJava)
+    val services = List(
+      injector.getInstance(classOf[HttpService]),
+      injector.getInstance(classOf[MarathonSchedulerService])
     )
+    val serviceManager = new ServiceManager(services.asJava)
+    this.serviceManager = Some(serviceManager)
+
+    serviceManager.startAsync
```
kolloch: since this has a side-effect I would use `()`

kolloch:
We agreed to move this to Chaos...

PR #1860, opened 2015-07-23
Fix internal links in REST api docs

### LINKED ISSUE #1859: Links in REST API docs don't work
These anchors don't work: 
https://mesosphere.github.io/marathon/docs/rest-api.html#post-/v2/groups

I'm not sure if this is fixable but if not, they should just be removed, so that it's not confusing.

pierlo-upitup:
This fixes #1859


kolloch:
Cool, thanks for this!

The /help link does not point to anything but it simply doesn't exist.

PR #1877, opened 2015-07-24
Fixes #1818 - Make Marathon fail on startup errors

### LINKED ISSUE #1818: Marathon should fail during start, if the HTTP port is already bound

gkleiman:
Fixes #1818 - Make Marathon fail on startup errors

Upgrade Chaos to the latest version, in order to make
Marathon fail and exit if any of the services can't be started,
e.g., if another process is already bound to the the HTTP port.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -243,7 +243,7 @@ object Dependencies {
 object Dependency {
   object V {
     // runtime deps versions
-    val Chaos = "0.6.5"
+    val Chaos = "0.6.7"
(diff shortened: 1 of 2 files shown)


kolloch:
Thanks a lot! :+1:

Issue #878, opened 2014-12-10
Jump to Mesos sandbox from UI


chendo:
It would be fantastic to be able to jump to the relevant Mesos sandbox for the task in the task list view.

xor-xor:
+1

quamilek:
Currently in [Allegro Group](https://github.com/allegro) we are working on this.
See below visualization for this feature.
![mesos-task-url-example](https://cloud.githubusercontent.com/assets/695791/8649614/44b81b34-296c-11e5-9741-5ce977df4055.png)

What do you think?

kamilchm:
@MarathonDevs - what do you think about the labels?
I think that we could put empty header and ``Sandbox`` link in each row. It'll be consistent with Mesos UI.

philipnrmn:
@quamilek that's great to hear! We'd be really interested to see more. 

@kamilchm I've just done a bit of hallway testing, the consensus seems to be that it would be neat either to have a separate row, or to link to the sandbox directly from the node name (`mesos-slave1.local:12345` in this case). The extra column consumes space, and the sort option is a little confusing.

wendigo:
@philipnrmn maybe we can inline this link after the node name? I think that node name with port alone is a valuable information and personally I would leave it as it is.

janisz:
I'm working on this ([here](https://github.com/janisz/marathon/commit/15110a749e6b0f68198b18133d62316fc3955bec) you can see progress). Simply I extended `MarathonTask` with field that contains URL to sandbox, it's populated in `DefaultTaskFactory.scala` with information from offer. I've problem with obtaining mesos api host and port. On first sight it could be `config.mesosMaster` but it's zk address. 
What is the best way to get mesos api url in Marathon?

quamilek:
I prepared two proposal of link location

**1. located after node name:**
![mesos_details_example2](https://cloud.githubusercontent.com/assets/695791/8716845/39693eae-2b95-11e5-9f97-1e5e7094c81a.png)
**2. located on task details page:**
![mesos_details_example1](https://cloud.githubusercontent.com/assets/695791/8716850/42603634-2b95-11e5-9d25-ce60584d5ee0.png)

pierlo-upitup:
@quamilek thanks for your suggestions, we are going to look into the best solution for this shortly!

pierlo-upitup:
https://github.com/mesosphere/marathon/pull/1882

aldipower:
[...]

Issue #1745, opened 2015-07-02
Save Task SlaveID and use it for reconciliation


kolloch:
This will speedup reconciliation in the presence of disconnected slaves. See API section in 

http://mesos.apache.org/documentation/latest/reconciliation/

This might improve #1279 and #1456.

Of course, the code has to work in the case the SlaveID was not yet saved for a task.

PR #1882, opened 2015-07-25
#878 - Add slaveId in MarathonTask and API to retrieve Mesos Master URL

janisz:
This PR prepare backend to expose sandbox URL in GUI https://github.com/mesosphere/marathon/issues/878. 

Sandbox URL is build from 3 pieces

1. Mesos master URL
2. SlaveID
3. TaskID

So I added SlaveID to MarathonTask, and added `/v2/mesos/` API to obtain current Mesos master.

### DIFF
--- a/docs/docs/rest-api.md
+++ b/docs/docs/rest-api.md
@@ -1967,6 +1967,7 @@ Server: Jetty(8.y.z-SNAPSHOT)
         "local_port_max": 49151,
         "local_port_min": 32767,
         "master": "zk://localhost:2181/mesos",
+        "mesos_master_url": "http://mesos.vm:5050",
         "mesos_role": null,
         "mesos_user": "root",
         "reconciliation_initial_delay": 30000,
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -21,6 +21,11 @@ trait MarathonConf
     required = true,
     noshort = true)
 
+  lazy val mesosMasterUrl = opt[String]("mesos_master_url",
+    descr = "The host and port (e.g. \"http://mesos_host:5050\") of the Mesos master",
+    required = false,
+    noshort = true)
+
   lazy val mesosFailoverTimeout = opt[Long]("failover_timeout",
     descr = "The failover_timeout for mesos in seconds (default: 1 week)",
     default = Some(604800L))
--- a/src/main/scala/mesosphere/marathon/MarathonModule.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonModule.scala
@@ -15,7 +15,7 @@ import com.google.inject._
 import com.google.inject.name.Names
 import com.twitter.common.base.Supplier
 import com.twitter.common.zookeeper.{ Candidate, CandidateImpl, Group => ZGroup, ZooKeeperClient }
-import com.twitter.zk.{ NativeConnector, ZkClient }
+import com.twitter.zk.{ AuthInfo, NativeConnector, ZkClient }
 import mesosphere.chaos.http.HttpConf
 import mesosphere.marathon.api.LeaderInfo
 import mesosphere.marathon.core.launchqueue.LaunchQueue
@@ -36,7 +36,7 @@ import mesosphere.util.SerializeExecution
 import mesosphere.util.state.memory.InMemoryStore
 import mesosphere.util.state.mesos.MesosStateStore
 import mesosphere.util.state.zk.ZKStore
-import mesosphere.util.state.{ FrameworkId, FrameworkIdUtil, PersistentStore }
+import mesosphere.util.state._
 import org.apache.log4j.Logger
 import org.apache.mesos.state.ZooKeeperState
 import org.apache.zookeeper.ZooDefs
@@ -97,6 +97,15 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
       .annotatedWith(Names.named(ModuleNames.NAMED_LEADER_ATOMIC_BOOLEAN))
       .toInstance(leader)
 
+  }
+
+  @Provides
+  @Singleton
+  def provideMesosLeaderInfo(): MesosLeaderInfo = {
+    conf.mesosMasterUrl.get match {
+      case someUrl @ Some(_) => ConstMesosLeaderInfo(someUrl)
+      case None              => new MutableMesosLeaderInfo
+    }
   }
 
   @Named(ModuleNames.NAMED_HTTP_EVENT_STREAM)
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -2,13 +2,15 @@ package mesosphere.marathon
 
 import javax.inject.{ Inject, Named }
 
+import akka.actor.ActorSystem
 import akka.event.EventStream
 import mesosphere.marathon.core.base.Clock
 import mesosphere.marathon.core.launcher.OfferProcessor
 import mesosphere.marathon.core.task.bus.TaskStatusObservables.TaskStatusUpdate
 import mesosphere.marathon.core.task.bus.{ MarathonTaskStatus, TaskStatusEmitter }
 import mesosphere.marathon.event._
-import mesosphere.util.state.FrameworkIdUtil
+import mesosphere.marathon.tasks._
+import mesosphere.util.state.{ FrameworkIdUtil, MesosLeaderInfo }
 import org.apache.log4j.Logger
 import org.apache.mesos.Protos._
 import org.apache.mesos.{ Scheduler, SchedulerDriver }
@@ -26,6 +28,9 @@ class MarathonScheduler @Inject() (
     offerProcessor: OfferProcessor,
     taskStatusEmitter: TaskStatusEmitter,
     frameworkIdUtil: FrameworkIdUtil,
+    mesosLeaderInfo: MesosLeaderInfo,
+    taskIdUtil: TaskIdUtil,
+    system: ActorSystem,
     config: MarathonConf,
     schedulerCallbacks: SchedulerCallbacks) extends Scheduler {
 
@@ -41,13 +46,13 @@ class MarathonScheduler @Inject() (
     master: MasterInfo): Unit = {
     log.info(s"Registered as ${frameworkId.getValue} to master '${master.getId}'")
     frameworkIdUtil.store(frameworkId)
-
+    mesosLeaderInfo.onNewMasterInfo(master)
     eventBus.publish(SchedulerRegisteredEvent(frameworkId.getValue, master.getHostname))
   }
 
   override def reregistered(driver: SchedulerDriver, master: MasterInfo): Unit = {
     log.info("Re-registered to %s".format(master))
-
+    mesosLeaderInfo.onNewMasterInfo(master)
     eventBus.publish(SchedulerReregisteredEvent(master.getHostname))
   }
 
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -71,7 +71,8 @@ trait Formats
       "ports" -> task.getPortsList.asScala,
       "startedAt" -> (if (task.getStartedAt != 0) Timestamp(task.getStartedAt) else JsNull),
       "stagedAt" -> (if (task.getStagedAt != 0) Timestamp(task.getStagedAt) else JsNull),
-      "version" -> task.getVersion
+      "version" -> task.getVersion,
+      "slaveId" -> (if (task.hasSlaveId) task.getSlaveId.getValue else JsNull)
     )
   }
 
--- a/src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
+++ b/src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
@@ -1,0 +1,36 @@
+package mesosphere.util.state
+
+import org.apache.mesos.Protos.MasterInfo
+
+/**
+  * Utility class for keeping track of a Mesos Leader
+  */
+trait MesosLeaderInfo {
+
+  /**
+    * Return information about current Mesos master URL
+    * @return String in format "protocol://host:port/" (e.g. "http://mesos.vm:5050/")
+    */
+  def currentLeaderUrl: Option[String]
+
+  def onNewMasterInfo(master: MasterInfo): Unit
+}
+
+case class ConstMesosLeaderInfo(currentLeaderUrl: Option[String]) extends MesosLeaderInfo {
+
+  override def onNewMasterInfo(ignored: MasterInfo): Unit = {}
+}
+
+class MutableMesosLeaderInfo extends MesosLeaderInfo {
+
+  @volatile private[this] var leaderUrl: Option[String] = None
+
+  override def currentLeaderUrl: Option[String] = {
+    leaderUrl
+  }
+
+  override def onNewMasterInfo(master: MasterInfo): Unit = {
+    leaderUrl = Some(s"http://${master.getHostname}:${master.getPort}/")
+  }
+}
+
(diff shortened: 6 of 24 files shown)


src/main/scala/mesosphere/marathon/api/v2/MesosResource.scala
```diff
@@ -0,0 +1,24 @@
...
+import mesosphere.marathon.MarathonConf
+import mesosphere.util.state.MesosMasterUtil
+
+@Path("v2/mesos")
+class MesosResource @Inject() (
+    mesosMasterUtil: MesosMasterUtil,
+    conf: MarathonConf) {
+  @GET
+  @Produces(Array(MediaType.APPLICATION_JSON))
+  def index(): Response = {
```
kolloch: This is mostly already included in the `InfoResource`.

Just add the `"masterUrl"` field as `"mesos_master_url"`.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -71,7 +71,8 @@ trait Formats
       "ports" -> task.getPortsList.asScala,
       "startedAt" -> (if (task.getStartedAt != 0) Timestamp(task.getStartedAt) else JsNull),
       "stagedAt" -> (if (task.getStagedAt != 0) Timestamp(task.getStagedAt) else JsNull),
-      "version" -> task.getVersion
+      "version" -> task.getVersion,
+      "slaveId" -> task.getSlaveId.getValue
```
kolloch: We cannot rely on the slaveId being present. Does this code correctly deal with that case?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -91,6 +92,11 @@ trait Formats
       withServicePorts ++ Json.obj("healthCheckResults" -> task.healthCheckResults)
     else
       withServicePorts
+
+    if (task.mesosTaskUrl.nonEmpty)
+      enrichedJson ++ Json.obj("mesosTaskUrl" -> task.mesosTaskUrl)
```
kolloch: Json.obj should correctly deal with `None` values, so no need for an if-statement.

src/main/scala/mesosphere/marathon/tasks/TaskUrlUtil.scala
```diff
@@ -0,0 +1,20 @@
+package mesosphere.marathon.tasks
+
+import com.google.inject.Inject
+import mesosphere.marathon.MarathonConf
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.util.state.{ FrameworkIdUtil, MesosMasterUtil }
+
+class TaskUrlUtil @Inject() (
```
kolloch: I like that you split this out into a separate class!

Please add a short description.

src/main/scala/mesosphere/marathon/tasks/TaskUrlUtil.scala
```diff
@@ -0,0 +1,20 @@
...
+    frameworkIdUtil: FrameworkIdUtil,
+    config: MarathonConf) {
+
+  def mesosTaskUrl(task: MarathonTask): Option[String] = {
+    val masterUrl = mesosMasterUtil.masterUrl
+    val slaveId = task.getSlaveId.getValue
+    for {
+      frameworkId <- frameworkIdUtil.fetch()
+    } yield s"$masterUrl/#/slaves/$slaveId/frameworks/${frameworkId.getValue}/executors/${task.getId}/"
+  }
```
kolloch: Maybe add an example here as documentation?

src/main/scala/mesosphere/marathon/tasks/TaskUrlUtil.scala
```diff
@@ -0,0 +1,20 @@
...
+import mesosphere.util.state.{ FrameworkIdUtil, MesosMasterUtil }
+
+class TaskUrlUtil @Inject() (
+    mesosMasterUtil: MesosMasterUtil,
+    frameworkIdUtil: FrameworkIdUtil,
+    config: MarathonConf) {
+
+  def mesosTaskUrl(task: MarathonTask): Option[String] = {
+    val masterUrl = mesosMasterUtil.masterUrl
+    val slaveId = task.getSlaveId.getValue
```
kolloch: treat this as an option. slaveId might not be set.

src/test/scala/mesosphere/marathon/api/TaskKillerTest.scala
```diff
@@ -50,8 +51,8 @@ class TaskKillerTest extends MarathonSpec
 
   test("KillRequested with scaling") {
     val appId = PathId(List("my", "app"))
-    val task1 = MarathonTasks.makeTask("task-1", "host", ports = Nil, attributes = Nil, version = Timestamp.now())
-    val task2 = MarathonTasks.makeTask("task-2", "host", ports = Nil, attributes = Nil, version = Timestamp.now())
+    val task1 = MarathonTasks.makeTask("task-1", "host", ports = Nil, attributes = Nil, version = Timestamp.now(), slaveId = SlaveID.newBuilder.setValue("").build())
```
kolloch: Please wrap these lines, they are getting really long.

kolloch:
BTW: This can actually also be used as a first step for https://github.com/mesosphere/marathon/issues/1745. Nice!

kolloch:
@janisz: My first impression is really good! Thanks for this PR!

I will continue my review but please also make sure that you have read this:

https://mesosphere.github.io/marathon/docs/contributing.html

It would be nice if you squashed your PR into one commit based on the current master.

kolloch:
Hi @janisz,

I finished my first review round. There is still one major weakness in your PR: With your code we hit Zookeeper really often for fetching the FrameworkId and the MasterInfo.

We get notified about FrameworkId/MasterInfo changes in `mesosphere.marathon.MarathonScheduler#registered` and `mesosphere.marathon.MarathonScheduler#reregistered`. We should save this info in main-memory and then use that for the MesosMasterUtil. For new code, we want to keep mutable shared state in actors.

We also need to be careful about relying on new fields like slaveId: We always have to deal gracefully with the case that the slaveId is not set (because the task was created before the upgrade).

We would really like this PR! So it would be great if you let us know if you are willing to continue working on this. If yes, we could also schedule a hangout when talking about a revision of this PR.

janisz:
@kolloch Thanks for review. I hope I'll fix all issues this week. 

In my first draft 87ccbd9b1dea20bd163c126a1ea598822547f163 I actually used `MarathonScheduler` to get notification about Mesos master changes but I was not sure if this was right approach.

kolloch:
Hi @janisz, I guess your first approach wasn't so bad, then.

You should make the field that you change `@volatile`, though, and make it an Option. Then make sure that you deal gracefully with the case that it is `None` (e.g. by simply not including the mesos task url in /v2/info).

I had a further discussion in my team right now: 

* We agreed to not export the `mesosTaskURL` for every task. Exporting the `mesos_master_url` in the info endpoint and the `slaveId` for every task should allow the Marathon UI to construct the URL itself.
* In addition, we would like a global configuration option to disable the automatic detection of the `mesos_master_url` and just set it to a constant configuration value.

The reason is that we are not sure how we expose information from Mesos in the Marathon UI in the future. Directly linking to the Mesos UI is often problematic because not all mesos master hosts might be reachable directly from the outside. Thus, we allow to overwrite this value to the publicly available mesos master URL (maybe load balanced). In the future, we might even proxy some things in Marathon to the Mesos API so that the browser never has to speak with the Mesos master directly.

Does that make sense?

kolloch:
@pierlo-upitup : Can you check if my above comment makes sense? Thanks.

janisz:
@kolloch So I should monitor MesosMaster with actor or with global variable?

kolloch:
@janisz Usually, we would use an actor for this but since it's only one variable, a @volatile Option[String] should be fine.

pierlo-upitup:
@kolloch 

Regarding: 
>  Exporting the mesos_master_url in the info endpoint and the slaveId for every task should allow the Marathon UI to construct the URL itself.

Sure, that's trivial.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -15,6 +15,11 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
     required = true,
     noshort = true)
 
+  lazy val mesosMasterUrl = opt[String]("mesos_master_url",
+    descr = "The host and port (e.g. \"http://mesos_host:5050\") of the Mesos master",
+    required = false,
+    noshort = true)
```
janisz: Should I validate URL?

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -101,6 +101,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
 
   }
 
+  @Provides
+  @Singleton
+  def provideMesosLeaderInfo(): MesosLeaderInfo = {
+    if (conf.mesosMasterUrl.isDefined) {
```
Kornel: getOrElse?

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -101,6 +101,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
...
+      ConstMesosLeaderInfo(Some(conf.mesosMasterUrl()))
```
kolloch: I think `getOrElse` here would be great.

src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def load(host: String, port: Int): Unit
+}
+
+case class ConstMesosLeaderInfo(currentLeaderUrl: Option[String]) extends MesosLeaderInfo {
+
+  override def load(host: String, port: Int): Unit = {}
+}
+
+object MutableMesosLeaderInfo extends MesosLeaderInfo {
```
kolloch: You asked me directly whether you use a global variable for this or not. For some reason, I misinterpreted the question. Otherwise I would have said: No, please don't! ;) Sorry about that.

So please make `MutableMesosLeaderInfo` a class instead of an object.

src/main/scala/mesosphere/marathon/MarathonModule.scala
janisz: I don't know how to use it here.

src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  @volatile var leaderUrl: Option[String] = None
```
kolloch: please make this `private[this]`

src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
```diff
@@ -0,0 +1,30 @@
+package mesosphere.util.state
+
+/**
+  * Utility class for keeping track of a Mesos Leader
+  */
+trait MesosLeaderInfo {
+
+  def currentLeaderUrl: Option[String]
```
kolloch: Maybe document which URL it is exactly with an example. Does it include a trailing slash? etc.

src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def load(host: String, port: Int): Unit
```
kolloch: Why not directly set the full URL here? I think that future Mesos versions will support HTTPS (maybe it is in 0.23 already?).

I also do not like the method name "load" so much. Maybe `onAutoDetectedMasterUrl`? That tries to express that the implementation can choose to use this information or not.

kolloch: Or better: `onNewMasterInfo` and pass the `MasterInfo` as argument. The then the logic of constructing an URL from that is inside this object.

docs/docs/rest-api.md
```diff
@@ -1967,6 +1967,7 @@ Server: Jetty(8.y.z-SNAPSHOT)
         "local_port_max": 49151,
         "local_port_min": 32767,
         "master": "zk://localhost:2181/mesos",
+        "mesos_master_url": "http://mesos.vm:5050",
```
kolloch: You should explain this better. Point out that it is automatically detected by default but you can override it with a certain config option.

kolloch: Also add some documentation in command-line-flags.md.

src/main/scala/mesosphere/util/state/MesosLeaderInfo.scala
janisz: `MasterInfo` has no information about protocol, that's why I decided to use host and port.

kolloch:
Regarding your commit message:

* Your commit message should start with "Fixes #878 - "
* You should mention that you use this to auto detect the master url

src/main/scala/mesosphere/marathon/MarathonModule.scala
kolloch: `getOrElse` doesn't fit, sorry. But match would:

```scala
    conf.mesosMasterUrl.get match {
      case someUrl @ Some(_) => ConstMesosLeaderInfo(someUrl)
      case None              => MutableMesosLeaderInfo
    }
```

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -35,6 +35,7 @@ class MarathonScheduler @Inject() (
     taskTracker: TaskTracker,
     taskQueue: TaskQueue,
     frameworkIdUtil: FrameworkIdUtil,
+    mesosMasterUtil: MesosLeaderInfo,
```
kolloch: call the variable `mesosLeaderInfo`

kolloch:
Hi @janisz, great patch! Thank you.

I made some comments, I hope that they make sense and are reasonably easy to fix.

I think there are still tests missing:

* That the DefaultTaskFactory copies the slaveId from the Offer into the MarathonTask
* For the slaveId in the Task JSON
* That MarathonScheduler.*registered callbacks call the appropriate method in `MesosLeaderInfo`
* For the InfoResource (admittedly there are none so far but we want to have tests for new code at least)

Not so important but easy to fix:

* If you initialize SlaveID in tests, I would prefer you to use `"slaveId"` or something like that instead of an empty string.

src/main/scala/mesosphere/marathon/MarathonConf.scala
kolloch: I think that would be great but not a requirement for me.

janisz:
@kolloch 
Thanks for review.

I hope I fixed most issues. I started with tests for `InfoResource` because I thought it will be easy, but I have no idea how properly mock configuration. With other tests I don't know if there are any good practices or convention. Tests in project look different (e.g. uses different assertions) so I'm not sure how to write them. 

Acording to
> That MarathonScheduler.*registered callbacks call the appropriate method in MesosLeaderInfo

I added check for it in `MarathonSchedulerTest` maybe it could be more explicit.

kolloch:
Hi @janisz, often, you do not need to mock config at all, you simply create it. See for example: https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActorTest.scala#L299-L300

If you need to supply special values for testing, you can do something similar to 
https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/MarathonTestHelper.scala#L18-L44

(I would prefer you not to extend that but do that in your test suite)

kolloch:
This is a good patch. Thank you, @janisz .

You wrote above that you could make the tests more explicit. Do you still want to improve the testing there?

Otherwise this passes my bar for merging now :+1:

janisz:
@kolloch Thanks for hint, I've already found this helper but was wandering if there is any other way than creating full config to test `InfoResouce`.
I was thinking about extracting [this assertion] to separate test but it'll end up with same test twice (only different assertions). So I think there is nothing to improve. 

I see integration tests are failing. It's weird. Only change I made today is commit author because it's was wrong after rebase.

kolloch:
Hi @janisz, some of our integration tests are not reliable, unfortunately. I have restarted them.

janisz:
OK, build is green so I think it could be merge

kolloch:
Thank you @janisz !

PR #1897, opened 2015-07-29
Fix developing VM docs

janisz:


kolloch:
Thanks!

kolloch:
How do you like the vagrant experience? I think all developers here in my team just run Mesos directly on their machine. I was thinking of documenting that instead as the recommended development setup.

janisz:
I started with Vagrant because it was described in docks how to set it up. It's quick and simple solution to check if everything is running correctly. When I faced some problems I installed Mesos locally to debug integration tests with IDE (without using remote debugger). Now I use Vagrant only for final tests on clean environment.

kolloch:
Sounds like it is still useful -- thanks for the feedback.

PR #1907, opened 2015-07-30
fix tomcat url in example

anshulverma:
Seems like `tomcat` was upgraded at gtlib.gatech.edu. The url was updated to: http://www.gtlib.gatech.edu/pub/apache/tomcat/tomcat-7/v7.0.63/bin/apache-tomcat-7.0.63.tar.gz


kolloch:
Thanks for the update!

PR #1911, opened 2015-08-01
Update rest-api.md

cameronpickham:
Fix documentation concerning API route for restarting an app's tasks


kolloch:
Thanks! @cameronpickham

PR #1921, opened 2015-08-03
Introduces RWLock utility class.

ConnorDoyle:
Used same in `MarathonHealthCheckManager`.


kolloch:
Nice patch, thank you.

Issue #687, opened 2014-10-08
Expose recent scheduling decisions in the API


ConnorDoyle:
For each app, what happened when pending tasks were considered against a resource offer?  For example, did we wait to satisfy the backoff deadline, did we refuse due to constraints?  See: #482.

jasongilanfarr:
I believe this is handled by the new Declined Offers functionality.

PR #1925, opened 2015-08-04
Handle missing disk resource in offers.

### LINKED ISSUE #1924: App stuck in deployment because of missing (unneeded) disk resource in the offer
It looks like we return a `None` if the offer doesn't include all of 'cpu', 'mem', and 'disk' resources as seen here: https://github.com/mesosphere/marathon/blob/v0.8.1/src/main/scala/mesosphere/mesos/TaskBuilder.scala#L174.

Mesos will not pass a 'disk' resource if there is 0 available disk on that machine (e.g., if all of the disk is used up by other tasks). Thus, even if there is ample 'cpu' and 'mem' resources Marathon will reject an offer because it does not contain a 'disk' resource even though the disk requirement for a task might be 0.

We've reproduced this behavior with Marathon by launching a one machine cluster and running one task which takes all the 'disk' followed by another task which required 'disk: 0' and confirmed that the second task would not launch.

ConnorDoyle:
A minimal patch to handle the case where the disk resources for Marathon's role have been fully allocated.  When a scalar resource goes to zero, Mesos stops including it in resource offers for that slave.  Previously, Marathon would fail to match offers even when the queued task does not require any disk resources, when the disk resource was missing from the offer.

Fixes #1924 for Marathon version `v0.8.1`.  It looks like although the offer matching logic was refactored a bit since then the basic logic was carried over.  A new patch will be required to fix #1924 for `v0.10.0-RC5`.

This was a little difficult to debug, and more verbose logging in the offer matching logic would be a good addition for future versions of Marathon.  See issue #687 on that point.

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -170,7 +170,13 @@ class TaskBuilder(app: AppDefinition,
       }
     }
 
-    if (cpuRole.isEmpty || memRole.isEmpty || diskRole.isEmpty) {
+    // If a particular resource type for this framework's role has been fully
+    // allocated on a slave, then the resource offer will not include that
+    // resource at all.
+    //
+    // For this reason, we should reject on empty disk role only if this app
+    // requires a positive amount of disk.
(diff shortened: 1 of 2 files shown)


ConnorDoyle:
cc @benh

kolloch:
Thanks for analyzing this, @ConnorDoyle!

PR #1234, opened 2015-02-23
fixes #1228 - Use force parameter when restarting an app

drexin:

Issue #1926, opened 2015-08-04
SSE leader proxy buffer/latency problem


drewrobb:
I'm encountering an issue in 0.10.0-RC4 when using the Server Sent Event hub (/v2/events) from a non-leading marathon instance. The event stream is very clearly being buffered in some way and thus events can be greatly delayed before being visible to consumers if the event rate is sufficiently low. This does not happen when connecting directly to the current marathon leader (avoiding proxy). 

It easy to reproduce this by first creating a stream of small bogus events:

    while true; do curl -s 'http://marathon/v2/apps/fake-app/restart' -X POST  -H 'Content-Type: application/json' > /dev/null; sleep 0.1; done

And then consuming the events:
    curl -N -s http://marathon/v2/events -H "Accept: text/event-stream" -H "Cache-Control: no-cache"

If marathon is the leader, the events will appear one at a time nearly instantaneously, otherwise I receive a large chunk of events all at once.

PR #1930, opened 2015-08-04
Flush after write in transfer

drewrobb:
This addresses #1926. However, I'm guessing this could cause a performance regression, so we might want to have a strategy to apply this more selectively only to steaming connections, or pick another strategy.

### DIFF
--- a/src/main/scala/mesosphere/marathon/io/IO.scala
+++ b/src/main/scala/mesosphere/marathon/io/IO.scala
@@ -70,6 +70,7 @@ object IO {
         val byteCount = in.read(buffer)
         if (byteCount >= 0 && continue) {
           out.write(buffer, 0, byteCount)
+          out.flush()
           read()
         }
       }


kolloch:
Hi @drewrobb, awesome follow up to your issue!

Do you have a chance to analyse whether this truly leads to performance issues?

Since we are not reading byte-by-byte and we are talking about communication inside of one cluster (usually), I am not sure that this is an issue. If we block after write for flushing, more data might accumulate for reading in that time, so the block size increases for the flushes (until the 8KB are reached).

But I definitely would like to be sure about this.

kolloch:
Oh, by the way, to be nit picking, it would be cool, if you prefixed your commit message with "Fixes #1234 - "

If you can think of a way to test this automatically, this would be great, too. 

https://mesosphere.github.io/marathon/docs/contributing.html

kolloch:
I think, it is better to have this than not having this for now.

I would still love it if you tested the performance.

Thanks!

drewrobb:
It is not a proper test, but I've been running this patched into 0.10.0-RC4 on our cluster with a moderate amount of load and didn't see any issues. I'll try to benchmark a bulk transfer with and without this change, and I'll be more aware of the contribution guidelines next time :)

kolloch:
You are great, @drewrobb ! Thanks for the update. :+1:

PR #1932, opened 2015-08-04
Improve 0.9.1

kolloch:


kolloch:
Maybe we should include the log string calculating in `if(log.isInfoEnabled)` block.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -43,9 +43,33 @@ class TaskBuilder(app: AppDefinition,
...
+          else {
+            None
+          }
+
+          val portStrings = Seq(maybeStatic, maybeDynamic).flatten.mkString(" + ")
+
+          val portsString = s"ports=($portStrings)"
+
+          log.info(
+            s"Offer ${offer.getId.getValue}. Insufficient for ${app.id} (need cpus=${app.cpus}, mem=${app.mem}, " +
```
gkleiman: I would add "resources" after "Insufficient".

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -82,7 +82,10 @@ class PortsMatcher(
         offeredPortRanges.find(_.range.contains(port)).map { offeredRange =>
           PortWithRole(offeredRange.role, port)
         } orElse {
-          if (failLog) log.info(s"Couldn't find host port $port in any offered range for app [${app.id}]")
+          if (failLog)
+            log.info(
+              s"Offer ${offer.getId.getValue}. Couldn't find host port $port (of ${requiredPorts.mkString(", ")}) " +
+                s"in any offered range for app [${app.id}]")
```
ConnorDoyle: It might be nice to print the offered ranges as well here, at log level DEBUG.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -96,7 +99,7 @@ class PortsMatcher(
     takeEnoughPortsOrNone(expectedSize = numberOfPorts) {
       shuffledAvailablePorts.map(Some(_))
     } orElse {
-      log.info(s"Couldn't find $numberOfPorts ports in offer for app [${app.id}]")
+      log.info(s"Offer ${offer.getId.getValue}. Couldn't find $numberOfPorts ports in offer for app [${app.id}]")
```
ConnorDoyle: Why do some interpolated values have surrounding square brackets and others do not?  I like the square brackets because they make grepping logs easier, and it also looks obvious when the empty string is interpolated.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -22,18 +36,38 @@ object ResourceMatcher {
...
     val groupedResources = offer.getResourcesList.asScala.groupBy(_.getName)
 
-    def findScalarResourceRole(tpe: String, value: Double): Option[Role] =
-      groupedResources.get(tpe).flatMap {
-        _
-          .filter(resource => acceptedResourceRoles(resource.getRole))
-          .find { resource =>
-            resource.getScalar.getValue >= value
-          }.map(_.getRole)
+    def findScalarResource(tpe: String, requiredValue: Double): Option[ScalarMatch] =
```
ConnorDoyle: s/tpe/name

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -22,18 +36,38 @@ object ResourceMatcher {
 
     val groupedResources = offer.getResourcesList.asScala.groupBy(_.getName)
```
ConnorDoyle: Please add an explicit type annotation to this identifier.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -22,18 +36,38 @@ object ResourceMatcher {
...
-    def findScalarResourceRole(tpe: String, value: Double): Option[Role] =
-      groupedResources.get(tpe).flatMap {
-        _
-          .filter(resource => acceptedResourceRoles(resource.getRole))
-          .find { resource =>
-            resource.getScalar.getValue >= value
-          }.map(_.getRole)
+    def findScalarResource(tpe: String, requiredValue: Double): Option[ScalarMatch] =
+      groupedResources.get(tpe).flatMap { resources =>
+        val matchingScalars = resources.filter { resource =>
```
ConnorDoyle: s/matchingScalars/matchingScalarResources ?

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
gkleiman: They are logged in TaskBuilder.scala:72

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -20,20 +36,40 @@ object ResourceMatcher {
...
+        }
+        else {
+          Some(asMatches.maxBy(_.offeredValue))
+        }
       }
 
-    def cpuRoleOpt: Option[Role] = findScalarResourceRole(Resource.CPUS, app.cpus)
-    def memRoleOpt: Option[Role] = findScalarResourceRole(Resource.MEM, app.mem)
-    def diskRoleOpt: Option[Role] = findScalarResourceRole(Resource.DISK, app.disk)
+    val cpuRoleOpt: Option[ScalarMatch] = findScalarResource(Resource.CPUS, app.cpus)
```
ConnorDoyle: I think naming these *RoleOpt is really confusing, since they are options of ScalarMatch.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -51,20 +87,28 @@ object ResourceMatcher {
       badConstraints.isEmpty
     }
 
-    def portsOpt: Option[Seq[RangesResource]] = new PortsMatcher(app, offer, acceptedResourceRoles).portRanges match {
-      case None =>
-        log.warn("App ports are not available in the offer.")
-        None
+    def portsOpt: Option[Seq[RangesResource]] = new PortsMatcher(app, offer, acceptedResourceRoles).portRanges
+
+    val basicResourceRoles = Map(
```
ConnorDoyle: Again naming, the map's value type is `ScalarMatch`.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -51,20 +87,28 @@ object ResourceMatcher {
...
+      "cpu" -> cpuRoleOpt,
+      "disk" -> diskRoleOpt,
+      "mem" -> memRoleOpt
+    )
 
-      case x @ Some(portRanges) =>
-        log.debug("Met all constraints.")
-        x
+    if (basicResourceRoles.values.exists(_.exists(!_.matches))) {
```
ConnorDoyle: Nit: is thie clearer as `! basicResourceRoles.values.forAll(_.map(_.matches).getOrElse(false))` ?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -51,20 +87,28 @@ object ResourceMatcher {
...
+      log.info(s"Offer ${offer.getId.getValue}. Not all basic resources satisfied: $basicResourceString")
     }
 
     for {
-      cpuRole <- cpuRoleOpt
-      memRole <- memRoleOpt
-      diskRole <- diskRoleOpt
+      cpuRole <- cpuRoleOpt.flatMap(_.matchingRole)
+      memRole <- memRoleOpt.flatMap(_.matchingRole)
+      diskRole <- diskRoleOpt.flatMap(_.matchingRole)
```
ConnorDoyle: OK, we check for matching role here (and that the option is defined) but where do we check the value of `ScalarMatch.matches`?  It looks like that method is only invoked for the purposes of logging above.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -5,14 +5,30 @@ import mesosphere.marathon.state.AppDefinition
...
 import org.apache.mesos.Protos.Offer
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.Seq
+import scala.collection.mutable
 
 object ResourceMatcher {
   type Role = String
 
+  case class ScalarMatch(requiredValue: Double, offeredValue: Double, role: String) {
```
ConnorDoyle: How about making `role` an `Option[String]` instead?  Empty string won't be a valid, so this pushes some complexity to the users of this structure.

ConnorDoyle:
LGTM, besides last remaining comment. _Fix it, then :shipit:_

ConnorDoyle:
Thanks!

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -128,7 +131,8 @@ class PortsMatcher(
             case Some(PortRange(role, _)) =>
               Some(PortWithRole(role, pm.hostPort))
             case None =>
-              log.info(s"Cannot find range with host port ${pm.hostPort} for app [${app.id}]")
+              log.info(s"Offer ${offer.getId.getValue}. " +
```
kolloch: this needs square brackets.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
kolloch: Yeah, that makes sense. Using an empty string was a definite hack.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
ConnorDoyle: @kolloch was this comment addressed?  Sorry... seriously hate GitHub's practice of collapsing comments just because the line changed.

kolloch: "matches" is actually also used by "matchingRole" internally.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -51,20 +87,28 @@ object ResourceMatcher {
...
+    val basicResourceMatches = Map(
+      "cpu" -> cpuMatchOpt,
+      "disk" -> diskMatchOpt,
+      "mem" -> memMatchOpt
+    )
 
-      case x @ Some(portRanges) =>
-        log.debug("Met all constraints.")
-        x
+    if (!basicResourceMatches.values.forAll(_.map(_.matches).getOrElse(false))) {
```
gkleiman: This line doesn't compile anymore.

The previous (correct) version was:

`if (basicResourceMatches.values.exists(_.exists(!_.matches))) {`

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -51,20 +87,28 @@ object ResourceMatcher {
...
+      log.info(s"Offer ${offer.getId.getValue}. Not all basic resources satisfied: $basicResourceString")
     }
 
     for {
-      cpuRole <- cpuRoleOpt
-      memRole <- memRoleOpt
-      diskRole <- diskRoleOpt
+      cpuRole <- cpuMatchOpt.flatMap(_.matchingRole)
+      memRole <- memMatchOpt.flatMap(_.matchingRole)
+      diskRole <- diskMatchOpt.flatMap(_.matchingRole)
```
ConnorDoyle: What do you think about eliminating `ScalarResource.matchingRole` and rewriting this as follows:

```scala
diskMatchOpt.filter(_.matches).flatMap(_.matchingRole)
```

kolloch: To be honest, I do not think that that is better. But I also wouldn't object.

ConnorDoyle: OK, let's leave it as-is.

ConnorDoyle:
Taking this over, addressing remaining issues.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
ConnorDoyle: Let's fix that in a subsequent patch, just to keep this change small-ish.

PR #1933, opened 2015-08-04
Fixes #1931 - CLI args for reviveOffers offer rejection duration

### LINKED ISSUE #1931: Marathon should set a non-default filter when declining offers

```java
  /**
   * callback.
   *
   *
   *
   * @see OfferID
   * @see Filters
   * @see Status
   */
```

and clear them later on.

```
  /**
   * from those filtered slaves.
   *
   *
   * @see Status
   */
  Status reviveOffers();
```

kolloch:
`--revive_offers_for_new_apps` if true, revive offers is called when a new app
is added to the `TaskQueue` or if a task of an app with constraints dies. The
latter is necessary, for example, if there is a constraint that only allows
one task per host. In this case, we might accept an offer that we rejected
previously, simply because no task is running on the host anymore.

Also note, that Mesos only filters offers that are a strict sub set of
a rejected offer.

`--reject_offer_duration` allows configuring the duration for which offers
are declined if not matched in mesos.

`--min_revive_offers_interval` if `--revive_offers_for_new_apps` is specified,
do not call reviveOffers more often than this interval. It defaults to 5 seconds.

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/OfferMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/OfferMatcher.scala
@@ -25,6 +25,11 @@ trait OfferMatcher {
 }
 
 trait IterativeOfferMatcherConfig extends ScallopConf {
+  lazy val rejectOfferDuration = opt[Long]("reject_offer_duration",
+    descr = "(Default: Use mesos default of 5 seconds) " +
+      "The duration (milliseconds) for which to reject offers by default",
+    default = None)
+
   lazy val maxTasksPerOffer = opt[Int]("max_tasks_per_offer",
     descr = "Maximally launch this number of tasks per offer.",
     default = Some(1),
@@ -187,6 +192,10 @@ class IterativeOfferMatcher @Inject() (
   }
 
   private[tasks] def commitOfferUsagesToDriver(driver: SchedulerDriver, finalOfferUsage: OfferUsages): Unit = {
+    val tasksToLaunch = finalOfferUsage.usages.map(_.scheduledTasks.size).sum
+    // if the taskLaunchLimit was hit, we might need to get offers back again faster
+    val taskLaunchLimitHit = tasksToLaunch == maxTasksPerCycle
+
     var usedOffers: Long = 0
     var launchedTasks: Long = 0
     var declinedOffers: Long = 0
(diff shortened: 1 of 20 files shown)


kolloch:
This is untested and not yet done but we would like to hear your opinion about that. Thanks!

Next steps: Testing and throttling reviveOffer calls.

kolloch:
I added the reviveOffers throttling now.

kolloch:
Manual tests that we should perform (of course, with all options enabled):

* Launching apps with many tasks (to test what the code does if we hit the `--max_tasks_per_offer_cycle` limit)
* Launching an app with a `hostname:UNIQUE` constraint (which will not finish) on all slaves by using an instance count == slaves + 1. Now select a running task in the GUI, press "kill&scale". Check that the tasks get quickly restarted and the deployment finishes.
* Some concurrent app deployments.
* and, of course, check if the patch solves the resource starvation problem.

(at least what I can think of right now)

kolloch:
Options to test with could be:

`--revive_offers_for_new_apps --reject_offer_duration 300000 --min_revive_offers_interval 5000`

src/main/scala/mesosphere/marathon/tasks/OfferMatcher.scala
```diff
@@ -205,8 +214,15 @@ class IterativeOfferMatcher @Inject() (
         launchedTasks += tasks.size()
       }
       else {
-        log.debug("declining offer {}", offer.getId)
-        driver.declineOffer(offer.getId)
+        iterativeOfferMatcherConfig.rejectOfferDuration.get match {
```
jsancio: I think we need to remove the `.get`. Looks like the type of `rejectOfferDuration` is `Option[Long]`.

kolloch: No, it's not exactly, even though similar. It is a `ScallopOption[Long]`. The `.get` method actually returns an `Option[Long]`

src/main/scala/mesosphere/marathon/tasks/OfferMatcher.scala
```diff
@@ -187,6 +192,10 @@ class IterativeOfferMatcher @Inject() (
   }
 
   private[tasks] def commitOfferUsagesToDriver(driver: SchedulerDriver, finalOfferUsage: OfferUsages): Unit = {
+    val tasksToLaunch = finalOfferUsage.usages.map(_.scheduledTasks.size).sum
+    // if the taskLaunchLimit was hit, we might need to get offers back again faster
+    val taskLaunchLimitHit = tasksToLaunch == maxTasksPerCycle
```
jsancio: Should this be `val taskLaunchLimitHit = tasksToLaunch == maxTasksPerCycle`?

kolloch:
Closing this PR since it targets an incorrect branch.

PR #1934, opened 2015-08-04
Fixes #1745 - Use SlaveID for reconciliation

### LINKED ISSUE #1745: Save Task SlaveID and use it for reconciliation

janisz:


src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -449,6 +449,12 @@ class SchedulerActions(
         val knownTaskStatuses = appIds.flatMap { appId =>
           taskTracker.get(appId).collect {
             case task: Protos.MarathonTask if task.hasStatus => task.getStatus
+            case task: Protos.MarathonTask if task.hasSlaveId => // staged tasks without status but with slaveId
```
kolloch: Great patch, @janisz!

I would prefer if stored the builder in a var:

`val builder = TaskStatus.newBuilder` ...
and had an `if (task.hasSlaveId)` that adds the slaveId. That makes it more obvious that this case only differs in the slaveId. Now we have two nearly identical code blocks...

kolloch:
Great patch, thanks! Had I minor comment. Does it make sense to you?

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
janisz: How about extracting common code to method?

janisz:
I extracted duplicated code, it looks better.

kolloch:
Thanks, @janisz !

PR #1948, opened 2015-08-05
Fixes #968 - Put SlaveID in TaskFailure

### LINKED ISSUE #968: Expose a way to identify "fragile" marathon apps in the web UI (failure history)
So this is as much a question as an idea, based off of a need and problem we have run into while working on our app.

One of our apps will boot, and show 1/1 (all is good at this point)  ... however a specific action will trigger the task to die, however it will almost instantly switch back to 1/1 (tasks) . The only way to see that something bad happened is to see the side effect of the app going down in the user facing part of the app, or to look at the mesos panel to see the FAILED slave(s). 

It would be dandy to expose some sort of historical (limited to browser session) info on the app tasks. Something like, how often is it dropping tasks, when was the last time it did it... anything really, just to aid a user in the debugging process. 

Thanks again for your time and hearing me out.

janisz:
SlaveID in lastFailureTask will allow us to present url to Mesos

### DIFF
--- a/src/main/java/mesosphere/marathon/Protos.java
+++ b/src/main/java/mesosphere/marathon/Protos.java
@@ -18849,6 +18849,20 @@ public final class Protos {
      */
     com.google.protobuf.ByteString
         getTimestampBytes();
+
+    // optional .mesos.SlaveID slaveId = 8;
+    /**
+     * <code>optional .mesos.SlaveID slaveId = 8;</code>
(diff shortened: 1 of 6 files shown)


kolloch:
We currently experience problems on our TeamCity agents (disk full). I executed the integration tests manually and tested your patch with the latest UI code.

Great work! Thank you.

PR #1949, opened 2015-08-05
Fixes #1931 - Flag for reviveOffers and offer reject duration

### LINKED ISSUE #1931: Marathon should set a non-default filter when declining offers

kolloch:
`--revive_offers_for_new_apps` if true, revive offers is called when a new app
is added to the `TaskQueue` or if a task of an app with constraints dies. The
latter is necessary, for example, if there is a constraint that only allows
one task per host. In this case, we might accept an offer that we rejected
previously, simply because no task is running on the host anymore.

Also note, that Mesos only filters offers that are a strict sub set of
a rejected offer.

`--reject_offer_duration` allows configuring the duration for which offers
are declined if not matched in mesos.

`--min_revive_offers_interval` if `--revive_offers_for_new_apps` is specified,
do not call reviveOffers more often than this interval. It defaults to 5 seconds.

### DIFF
--- a/mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
+++ b/mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
@@ -33,6 +33,9 @@ object DriverActor {
     * `override def reconcileTasks(statuses: util.Collection[TaskStatus]): Status`
     */
   case class ReconcileTask(taskStatus: Seq[TaskStatus])
+
+  // Ignored for now. It is implemented in Marathon 0.11.*
+  case object ReviveOffers
 }
 
 class DriverActor(schedulerProps: Props) extends Actor {
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -1,12 +1,13 @@
 package mesosphere.marathon
 
-import mesosphere.marathon.tasks.IterativeOfferMatcherConfig
+import mesosphere.marathon.tasks.{ OfferReviverConf, IterativeOfferMatcherConfig }
 import org.rogach.scallop.ScallopConf
 import scala.sys.SystemProperties
 
 import mesosphere.marathon.io.storage.StorageProvider
 
-trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMatcherConfig with LeaderProxyConf {
+trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMatcherConfig with LeaderProxyConf
+    with OfferReviverConf {
 
   lazy val mesosMaster = opt[String]("master",
     descr = "The URL of the Mesos master",
@@ -167,4 +168,9 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
     validate = Set("zk", "mesos_zk", "mem").contains,
     default = Some("zk")
   )
+
+  lazy val reviveOffersForNewApps = opt[Boolean]("revive_offers_for_new_apps",
+    descr = "Whether to call reviveOffers for new or changed apps. (Default: do not use reviveOffers) ",
+    default = Some(false))
+
 }
--- a/src/main/scala/mesosphere/marathon/MarathonModule.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonModule.scala
@@ -64,6 +64,8 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
     bind(classOf[HttpConf]).toInstance(http)
     bind(classOf[ZooKeeperClient]).toInstance(zk)
     bind(classOf[LeaderProxyConf]).toInstance(conf)
+    bind(classOf[OfferReviverConf]).toInstance(conf)
+    bind(classOf[IterativeOfferMatcherConfig]).toInstance(conf)
 
     // needs to be eager to break circular dependencies
     bind(classOf[SchedulerCallbacks]).to(classOf[SchedulerCallbacksServiceAdapter]).asEagerSingleton()
@@ -111,10 +113,6 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
 
     system.actorOf(Props(new HttpEventStreamActor(leaderInfo, metrics, handleStreamProps)), "HttpEventStream")
   }
-
-  @Provides
-  @Singleton
-  def provideIterativeOfferMatcherConfig(): IterativeOfferMatcherConfig = conf
 
   @Provides
   @Singleton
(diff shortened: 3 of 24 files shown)


mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
```diff
@@ -33,6 +33,9 @@ object DriverActor {
     * `override def reconcileTasks(statuses: util.Collection[TaskStatus]): Status`
     */
   case class ReconcileTask(taskStatus: Seq[TaskStatus])
+
+  // Ignored for now. It is implemented in Marthon 0.11.*
```
gkleiman: typo: s/Marthon/Marathon/

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -1,12 +1,13 @@
...
 
-import mesosphere.marathon.tasks.IterativeOfferMatcherConfig
+import mesosphere.marathon.tasks.{ OfferReviverConf, IterativeOfferMatcherConfig }
 import org.rogach.scallop.ScallopConf
 import scala.sys.SystemProperties
 
 import mesosphere.marathon.io.storage.StorageProvider
 
-trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMatcherConfig with LeaderProxyConf {
+trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMatcherConfig with LeaderProxyConf
```
gkleiman: Nit: one of these configuration traits is not named the same as the others, but it was not created nor touched in this PR.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -191,6 +191,31 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
       "MarathonScheduler")
   }
 
+  @Provides
+  @Singleton
+  def provideOfferReviveConf(): OfferReviverConf = conf
```
gkleiman: is this better than `bind(classOf[OfferReviverConf]).toInstance(conf)` in MarathonModule#configure?

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -108,12 +109,23 @@ class MarathonScheduler @Inject() (
...
+    lazy val maybeApp: Future[Option[AppDefinition]] = appRepo.currentVersion(appId)
+
     if (status.getState == TASK_ERROR || status.getState == TASK_FAILED || killedForFailingHealthChecks)
-      appRepo.currentVersion(appId).foreach {
+      maybeApp.foreach {
         _.foreach(taskQueue.rateLimiter.addDelay)
       }
     status.getState match {
       case TASK_ERROR | TASK_FAILED | TASK_FINISHED | TASK_KILLED | TASK_LOST =>
+        if (config.reviveOffersForNewApps()) {
```
gkleiman: Remove the ()?

src/main/scala/mesosphere/marathon/MarathonModule.scala
kolloch: Ok, I'll change that.

src/main/scala/mesosphere/marathon/MarathonConf.scala
kolloch: I'll not change that ;)

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
kolloch: Jepp

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
kolloch: ScallopOption has an apply method that gets the current value. Without the `()`, it will not work.

(I do not like that API either)

src/main/scala/mesosphere/marathon/tasks/OfferMatcher.scala
```diff
@@ -187,6 +192,10 @@ class IterativeOfferMatcher @Inject() (
   }
 
   private[tasks] def commitOfferUsagesToDriver(driver: SchedulerDriver, finalOfferUsage: OfferUsages): Unit = {
+    val tasksToLaunch = finalOfferUsage.usages.map(_.scheduledTasks.size).sum
```
gkleiman: Explicitly add the type? This val holds a number, not a lists of tasks.

src/main/scala/mesosphere/marathon/tasks/OfferMatcher.scala
```diff
@@ -187,6 +192,10 @@ class IterativeOfferMatcher @Inject() (
...
+    // if the taskLaunchLimit was hit, we might need to get offers back again faster
+    val taskLaunchLimitHit = tasksToLaunch >= maxTasksPerCycle
```
gkleiman: Add the type, pretty please ;-).

src/main/scala/mesosphere/marathon/tasks/OfferReviverActor.scala
```diff
@@ -0,0 +1,95 @@
...
+
+    nextReviveCancellableOpt.foreach(_.cancel())
+    nextReviveCancellableOpt = None
+
+    eventStream.unsubscribe(self)
+  }
+
+  private[this] def reviveOffers(): Unit = {
+    val now: Timestamp = Timestamp.now()
+    val nextRevive = lastRevive + conf.minReviveOffersInterval().milliseconds
```
gkleiman: Add the explicit type?

src/main/scala/mesosphere/marathon/tasks/OfferReviverActor.scala
```diff
@@ -0,0 +1,95 @@
...
+      nextReviveCancellableOpt = None
+
+      driverHolder.driver.foreach(_.reviveOffers())
+      lastRevive = now
+    }
+    else {
+      lazy val untilNextRevive = now until nextRevive
+      if (nextReviveCancellableOpt.isEmpty) {
+        log.info("=> Schedule next revive at {} in {}, adhering to --{} {} (ms)",
+          nextRevive, untilNextRevive, conf.minReviveOffersInterval.name, conf.minReviveOffersInterval())
```
gkleiman: Remove ()

src/main/scala/mesosphere/marathon/tasks/TaskQueue.scala
```diff
@@ -70,7 +72,11 @@ class TaskQueue {
     val queuedTask = apps.getOrElseUpdate(
       (app.id, app.version),
       QueuedTask(app, new AtomicInteger(0)))
-    queuedTask.count.addAndGet(count)
+    val oldValue = queuedTask.count.getAndAdd(count)
+    if (conf.reviveOffersForNewApps() && oldValue == 0) {
```
gkleiman: Parentheses

gkleiman:
LGTM :+1:

PR #1959, opened 2015-08-06
Fixes #1957 - Rename --mesos_master_url to --mesos_leader_ui_url

### LINKED ISSUE #1957: Rename --mesos_master_url to --mesos_leader_ui_url or similar
To avoid confusion with the `--master` parameter, we should probably rename `--mesos_master_url` to `--mesos_leader_ui_url` or similar.

We also should make clear in the documentation that if you walk this path, the given URL should always load balance to the current master (as is done in DCOS).

janisz:
Require change in UI


kolloch:
That was fast. Thanks!

PR #1961, opened 2015-08-06
Fixes #1958 - Fix JSON schema for command health check

### LINKED ISSUE #1958: "command" for healthCheck incorrect in AppDefinition.json (schema)
The schema says that "command" should be of type "string". Marathon expects a nested "value" attribute (for whatever reason).

kolloch:
@ssk2 You might want to include that fix in the dcos CLI soon. You cannot create apps with the COMMAND health check currently.

(is there actually an option to ignore the json schema?)

### DIFF
--- a/docs/docs/rest-api/mesosphere/marathon/api/v2/AppDefinition.json
+++ b/docs/docs/rest-api/mesosphere/marathon/api/v2/AppDefinition.json
@@ -172,7 +172,15 @@
                 "additionalProperties": false,
                 "properties": {
                     "command": {
-                        "type": "string"
+                        "type": "object",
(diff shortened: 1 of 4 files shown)


ssk2:
Thanks for the notification @kolloch! I'll make sure it gets done.

aquamatthias:
Could you add a test for invalid cases?

kolloch:
yes, I'll do that. Thanks for the suggestion, @aquamatthias .

kolloch:
I hope that the new patch will meet your expectations ;)

aquamatthias:
My high expectations are met. Thank you!

PR #1963, opened 2015-08-06
add additional s3 uris

bblanton:
When fetching files from S3, Mesos uses the hadoop client.  That client supports two additional [URIs](http://wiki.apache.org/hadoop/AmazonS3) that Marathon can leverage.

In some cases the normal ```s3://``` schema would not be able to find the file/directory specified unless it was setup as described by the docs linked to above.   ```s3n``` and ```s3a``` support fetching files from a more typical s3 bucket setup.

Not a huge deal but may add some clarity.


aquamatthias:
Thanks for the addition!

PR #1970, opened 2015-08-08
Fixes #1909 - Add missing config values in /v2/info

### LINKED ISSUE #1909: Useful configuration values missing from `/v2/info` output
Including `--max_tasks_per_offer`.

janisz:


src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala
```diff
@@ -17,7 +21,7 @@ class InfoResource @Inject() (
     schedulerService: MarathonSchedulerService,
     mesosLeaderInfo: MesosLeaderInfo,
     leaderInfo: LeaderInfo,
-    conf: MarathonConf with HttpConf with EventConfiguration with HttpEventConfiguration with LeaderProxyConf) {
+    conf: MarathonConf with HttpConf with EventConfiguration with HttpEventConfiguration with LeaderProxyConf with OfferMatcherManagerConfig with OfferProcessorConfig with DebugConf with LaunchQueueConfig with LaunchTokenConfig) { // scalastyle:ignore
```
janisz: How to configure autoformater to split this into multiple lines instead of joining into long one?

ConnorDoyle: See the scalariform docs: https://github.com/mdr/scalariform#format-onoff

janisz: I hoped there was a better way

kolloch:
Hey @janisz, great patch!

What I really like is that you create a grouping of the configuration patches, this is why I took a while to propose the following solution nevertheless:  The `builder` field in `ScallopConf` contains an `Scallop` object. This object contains fields/methods to actually get all configuration options dynamically.

If the use that information in the info endpoint, we would not have a nice hierarchical structure but we would never forget to export a config option again. If we use this approach, we have to make sure that we do not change the structure of any options that we already export. We should filter hidden options, though.

An alternative would be to use the `Scallop` opts list in a test that ensures that we never forget to add a config option again. That would be more work (also in the future) but allow us to use the hierarchical structure.

What do you think?

janisz:
Hi @kolloch 

I thought about this issue. Main problem with this endpoint is lack of tests and adding unit tests it tedious work and I have no idea how to check if field was added to config but not present in API. I prefer flat structure. Most keys has group prefix or are easily to guess what are they for. On the other hand changing hierarchical structure will break backward compatibility and was main reason I do it in same manner.

If we can break backward compatibility I'll go with flat structure.

kolloch:
Hmm, I am not sure if I understand you completely.

There are some pre-existing top-level configuration options which are exported without nesting. These would have to be excluded or exported twice (no big deal).

The automatically discovered options could all go into the "marathon_config" map.

I guess it would be sufficient to test that a subset of the options gets generated into the "marathon_config" map. If you put the code that extracts the options into a different package private method, you can easily test that.

What do you think?

janisz:
Basically something like this will do the trick
 ```scala
conf.builder.opts.map(option => option.name -> conf.builder.get(option.name))
```

Other groups like  `zookeeper_config` will be marked as deprecated and deleted in next release.

**Comment by kolloch** [comment]
Sounds good. We can leave the deprecated stuff as is.

**Comment by kolloch** [comment]
You still want to do that @janisz or should we close this PR?

**Comment by janisz** [comment]
I have a problem with types:

>InfoResource.scala:44: No Json serializer found for type Option[Any]. Try to implement an implicit Writes or Format for this type.

Have you got any idea how to solve it?

File: src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala [hunk 2]
```diff
@@ -18,32 +24,26 @@ class InfoResource @Inject() (
...
-    "webui_url" -> config.webuiUrl.get,
-    "mesos_role" -> config.mesosRole.get,
-    "task_launch_timeout" -> config.taskLaunchTimeout.get,
-    "reconciliation_initial_delay" -> config.reconciliationInitialDelay.get,
-    "reconciliation_interval" -> config.reconciliationInterval.get,
-    "marathon_store_timeout" -> config.marathonStoreTimeout.get,
-    "mesos_user" -> config.mesosUser.get,
-    "leader_proxy_connection_timeout_ms" -> config.leaderProxyConnectionTimeout.get,
-    "leader_proxy_read_timeout_ms" -> config.leaderProxyReadTimeout.get)
+    ConfigToMap.convertToMap(config).map {
```
ConnorDoyle: `ConfigToMap.convertToMap` explicitly returns a `Map[String, Option[Any]]`.

This return type is basically unusable, so that should also be addressed.  However, to clear the type error you should be able to stringify the values using [Map.stringToMap](http://www.scala-lang.org/api/current/index.html#scala.collection.Map@mapValues[C](f:B=>C):scala.collection.Map[A,C]) as follows:

```scala
ConfigToMap.convertToMap(config).mapValues(_.map(_.toString)).map {
  // now values are of type Option[String],
  // and scalac can find a Writes[String]...
}
```

src/main/scala/mesosphere/util/ConfigToMap.scala
```diff
@@ -0,0 +1,12 @@
+package mesosphere.util
+
+import org.rogach.scallop.ScallopConf
+
+object ConfigToMap {
+
+  def convertToMap(config: ScallopConf): Map[String, Option[Any]] = {
```
ConnorDoyle: With the return type as written this method is unlikely to be useful anywhere else (we generally don't allow blind casts from `Any`).  It might just be better after all to move this code to a `Writes[ScallopConf]` in [Formats.scala](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala).

src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala
janisz: Is it OK when API return all values as strings? If no then building JSON by hand is only way to do it.

src/main/scala/mesosphere/util/ConfigToMap.scala
janisz: But this require building this JSON by hand

janisz:
I think there is problem with Config implementation. `MarathonConf` does too much. Parsing should be split from config object.

kolloch:
I am closing this for now.

janisz:
OK

Issue #1673, opened 2015-06-17
Package the UI as webjar 


aldipower:
Create a script that creates a versioned release of the UI packaged in a .war-file, that can be served by the Marathon backend.

AC

1.    'npm run release'
2.   Version number should be retrieved from the package.json version key
3.    dist/ files should be created from the related Git Tag (version)
4.    Creates a ~~.war~~.jar-file with the contents of the dist-folder (main., img/, index.html)
5.    Released ~~.war~~.jar should be in the repository in an own release-folder

pierlo-upitup:
Update: there was a misunderstanding, this needs to be a WebJar file .

aldipower:
I am sorry, my fault.

philipnrmn:
No worries! There's a lot of similarity between the two, so we can still use some of this work.

aquamatthias:
Basic Steps provided. 
If finished, we need to integrate that into TC.
I create a ticket for the Backend part.

aldipower:
PR created https://github.com/mesosphere/marathon-ui/pull/108, but TC integration is more difficult than expected because I need a Dockerimage with both Node.js and Maven integration.

PR #1973, opened 2015-08-09
Fixes #1861 Serve UI from webjar

### LINKED ISSUE #1861: Load webui from webjar
Depends on #1673

Once we publish the UI as webjar, we need a way to load the resources from that.

aquamatthias:
Loads ui resources from webjar.
Please note: this will not compile as long as the ui project can not be resolved as dependency.
Waiting for #1673

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -233,6 +233,7 @@ object Dependencies {
     jsonSchemaValidator % "compile",
     twitterZk % "compile",
     rxScala % "compile",
+    marathonUI % "compile",
 
     // test
     Test.scalatest % "test",
@@ -262,6 +263,7 @@ object Dependency {
     val PlayJson = "2.3.7"
     val JsonSchemaValidator = "2.2.6"
     val RxScala = "0.25.0"
+    val MarathonUI = "0.11.0-SNAPSHOT"
 
     // test deps versions
     val Mockito = "1.9.5"
@@ -294,6 +296,7 @@ object Dependency {
   val jsonSchemaValidator = "com.github.fge" % "json-schema-validator" % V.JsonSchemaValidator
   val twitterZk = "com.twitter" %% "util-zk" % V.TwitterZk
   val rxScala = "io.reactivex" %% "rxscala" % V.RxScala
+  val marathonUI = "mesosphere.marathon" % "ui" % V.MarathonUI
 
   object Test {
     val scalatest = "org.scalatest" %% "scalatest" % V.ScalaTest
--- a/src/main/scala/mesosphere/marathon/api/WebJarServlet.scala
+++ b/src/main/scala/mesosphere/marathon/api/WebJarServlet.scala
@@ -1,0 +1,76 @@
+package mesosphere.marathon.api
+
+import java.net.URI
+import javax.servlet.http.{ HttpServlet, HttpServletRequest, HttpServletResponse }
+
+import mesosphere.marathon.io.IO
+import org.apache.log4j.Logger
+
+class WebJarServlet extends HttpServlet {
+
+  private[this] val log = Logger.getLogger(getClass)
+
+  //scalastyle:off method.length
+  override def doGet(req: HttpServletRequest, resp: HttpServletResponse): Unit = {
+
+    def sendResource(resourceURI: String, mime: String): Unit = {
+      //scalastyle:off magic.number
+      IO.withResource(resourceURI) { stream =>
+        resp.setContentType(mime)
+        resp.setContentLength(stream.available())
+        resp.setStatus(200)
+        IO.transfer(stream, resp.getOutputStream)
+      } getOrElse {
+        resp.sendError(404)
+      }
+    }
+
+    def sendResourceNormalized(resourceURI: String, mime: String): Unit = {
+      val normalized = new URI(resourceURI).normalize().getPath
+      if (normalized.startsWith("/META-INF/resources/webjars")) sendResource(normalized, mime)
+      else resp.sendError(404, s"Path: $normalized")
+    }
+
+    //extract request data
+    val jar = req.getServletPath // e.g. /ui
+    var resource = req.getPathInfo // e.g. /fonts/icon.gif
+    if (resource.endsWith("/")) resource = resource + "index.html" // welcome file
+    val file = resource.split("/").last //e.g. icon.gif
+    val mediaType = file.split("\\.").lastOption.getOrElse("") //e.g. gif
+    val mime = Option(getServletContext.getMimeType(file)).getOrElse(mimeType(mediaType)) //e.g plain/text
+    val resourceURI = s"/META-INF/resources/webjars$jar$resource"
+
+    //log request data, since the names are not very intuitive
+    if (log.isDebugEnabled) {
+      log.debug(
+        s"""
+         |pathinfo: ${req.getPathInfo}
+         |context: ${req.getContextPath}
+         |servlet: ${req.getServletPath}
+         |path: ${req.getPathTranslated}
+         |uri: ${req.getRequestURI}
+         |jar: $jar
+         |resource: $resource
+         |file: $file
+         |mime: $mime
+         |resourceURI: $resourceURI
+       """.stripMargin)
+    }
+
+    //special rule for accessing root -> redirect to ui main page
+    if (req.getRequestURI == "/") resp.sendRedirect("/ui/")
+    //if a directory is requested, redirect to trailing slash
+    else if (!file.contains(".")) resp.sendRedirect(req.getRequestURI + "/") //request /ui -> /ui/
+    //if we come here, it must be a resource
+    else sendResourceNormalized(resourceURI, mime)
+  }
+
+  private[this] def mimeType(mediaType: String): String = {
+    mediaType.toLowerCase match {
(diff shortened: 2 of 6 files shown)


project/build.scala
```diff
@@ -233,6 +233,7 @@ object Dependencies {
     jsonSchemaValidator % "compile",
     twitterZk % "compile",
     rxScala % "compile",
+    ui % "compile",
```
kolloch: I would prefer `marathonUI` to make it a bit clearer.

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -52,6 +54,14 @@ class MarathonApp extends App {
       new HttpModule(conf) {
         // burst browser cache for assets
         protected override val resourceCacheControlHeader = Some("max-age=0, must-revalidate")
+
+        override def provideHandlerCollection(instrumentedHandler: InstrumentedHandler,
```
kolloch: Is this related or something extra?

Does it provide metrics or something similar? Or is it a performance patch? Please document or make the intention clear by using other means.

src/main/scala/mesosphere/marathon/api/WebJarServlet.scala
```diff
@@ -0,0 +1,67 @@
...
+         |uri: ${req.getRequestURI}
+         |jar: $jar
+         |resource: $resource
+         |file: $file
+         |mime: $mime
+         |resourceURI: $resourceURI
+       """.stripMargin)
+    }
+
+    def sendResource(): Unit = {
```
kolloch: If you use nested methods, I'd prefer if you put them first. So you have the following structure:

* methods that are only used in this method and help with the implementation
* code that uses the methods and "wires" them together.

I do like it if the methods do not simply close over local variables but have them as arguments.

src/main/scala/mesosphere/marathon/io/IO.scala
```diff
@@ -85,6 +85,15 @@ object IO {
     new String(out.toByteArray, "UTF-8")
   }
 
+  def withResource[T](path: String)(fn: InputStream => T): Option[T] = {
```
kolloch: I am not convinced that this code does the proper input path sanitisation. How do we prevent that someone e.g. accesses the configuration files with potential secrets this way?

kolloch:
Tests would be splendid.

kolloch:
After this patch, is there still an option to work with the source of the UI directly in combination with the backend from source?

src/main/scala/mesosphere/marathon/api/WebJarServlet.scala
aquamatthias: OK - I will change that.

src/main/scala/mesosphere/marathon/Main.scala
aquamatthias: The problem is the order of the handler collection. The overridden method has the same handlers in different order. The instrumentedHandler needs to go before the resourceHandler (because it will try to load a resource directly, before we can handle the request in the controller)

aquamatthias: We should fix that in chronos. This was a quick hack.

src/main/scala/mesosphere/marathon/io/IO.scala
aquamatthias: What do you mean with path sanitisation? 
The method here (in the IO class) does not prevent anything.
In the WebJar case, you have only access to /META-INF/resources/webjars/*

aquamatthias:
After this patch we only work with compiled ui assets.

src/main/scala/mesosphere/marathon/api/WebJarServlet.scala
```diff
@@ -0,0 +1,68 @@
...
+    if (req.getRequestURI == "/") resp.sendRedirect("/ui/")
+    //if a directory is requested, redirect to trailing slash
+    else if (!file.contains(".")) resp.sendRedirect(req.getRequestURI + "/") //request /ui -> /ui/
+    //if we come here, it must be a resource
+    else sendResource(resourceURI, mime)
+  }
+
+  private[this] def mimeType(mediaType: String): String = {
+    mediaType.toLowerCase match {
+      case "eot" => "application/vnd.ms-fontobject"
```
kolloch: is that sufficient? Nothing for javascript or our images? Are there no warnings in the browser logs if you use it like this?

src/main/scala/mesosphere/marathon/io/IO.scala
aquamatthias: I compute the normalized path and make sure, only /META-INF/resources/webjars/* is accessible.

src/main/scala/mesosphere/marathon/api/WebJarServlet.scala
aquamatthias: This is only the fallback, if getServletContext.getMimeType(file) does not have the correct mime.

kolloch:
LGTM :+1: !!

philipnrmn:
Fantastic!

bobrik:
What's the point of redirecting from / to /ui? This breaks existing setups behind proxies. I run 2 instances for fast failover behind single nginx.

```
~ λ curl -v http://some.marathon.my.corp/
*   Trying 10.36.16.7...
* Connected to 36s52.in.cfops.it (10.36.16.7) port 80 (#0)
> GET / HTTP/1.1
> Host: some.marathon.my.corp
> User-Agent: curl/7.43.0
> Accept: */*
>
< HTTP/1.1 302 Found
< Date: Thu, 24 Sep 2015 08:51:26 GMT
< Server: Jetty(9.3.z-SNAPSHOT)
< Cache-Control: no-cache, no-store, must-revalidate
< X-Marathon-Leader: http://internal-server:8953
< Expires: 0
< Pragma: no-cache
< Date: Thu, 24 Sep 2015 08:51:26 GMT
< Location: http://internal-server:8953/ui/
< X-Marathon-Via: 1.1 36s52:31176
< Content-Length: 0
<
* Connection #0 to host 36s52.in.cfops.it left intact
```

Here `internal-server` is not accessible from the outside and `some.marathon.my.corp` breaks when users are redirected to access internal address.

Seeing this on 0.11.0-RC3.

aquamatthias:
Hey @bobrik then your nginx configuration does not handle redirects correctly (additionally: you can use /ping to see, if that instance is running).

bobrik:
I have a few questions then:

1. Can you please tell me the correct way of handling redirects to arbitrary hosts? To make it easier to understand my question, imagine that upstream can send redirects to http://mesos.apache.org, http://sdhr, http://asgh.

2. Why do redirect with host? `Location: /ui/` can eliminate problem entirely.

I'm already using `/ping` in health checks.

aquamatthias:
Hey @bobrik see #2306 that should fix your problem.
The patch will be ported back to 0.11.0-RC4

gkleiman:
@bobrik: I am no nginx expert, but I think that the proxy_redirect primitive applies to these cases: http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect

bobrik:
Forget nginx, tell me how things should be replaced if my users access marathon at `marathon.corp` and upstreams send the following headers:

1. `Location: http://mesos.apache.org`
2. `Location: http://foo.bar`
3. `Location: http://fizz.buzz`

gkleiman:
http://internal-server:8953/ui/ -> http://marathon.corp/ui/
http://mesos.apache.org -> http://mesos.apache.org
http://foo.bar -> http://foo.bar
http://fizz.buzz -> http://fizz.buzz

So you'd replace internal-server:8953 with your proxy host+port.

bobrik:
Ha, gotcha! `foo.bar` and `fizz.buzz` are my internal servers. I don't see how they are different from `mesos.apache.org`, these are just hostnames.

gkleiman:
If `marathon.corp` forwards a request to `foo.bar` and `fizz.buzz` (or any other hostname), then it knows it can proxy to them. If it gets a redirect from them and the host in the `Location` header is one of the servers it is configure to forward to, then it should replace that host with `marathon.crop`.

So if Marathon is running in internal-server1 and internal-server2 and marathon.corp is setup in a way that it will proxy to those two hosts, then should translate:

```
http://internal-server1:8953/ -> http://marathon.corp/
http://internal-server2:8953/ -> http://marathon.corp/
http://internal-server1:8953/ui/ -> http://marathon.corp/ui/
http://internal-server2:8953/ui/ -> http://marathon.corp/ui/
http://mesos.apache.org -> http://mesos.apache.org
```
And when it gets a request to http://marathon.corp/*, it should forward it to one of the internal servers.

bobrik:
Okay, more goodies. Marathon runs on `abc.in.xyz` (leader) and `def.in.xyz` (standby). Request is proxied from `marathon.corp` to `def.in.xyz`, reply comes back with `Location: http://abc` because `hostname -s` on `abc.in.xyz` is `abc`.

Even more: nginx may not know about `abc.in.xyz` at all and only talk to `def.in.xyz`.

Not including hostname and scheme in redirect is the right solution, not messing with rewriting URLs and catching redirects.

really

Issue #1847, opened 2015-07-21
LDAP authentication 


felixb:
I'd like to see LDAP authentication for marathon. The implementation should support r/w and r/o groups. You still need /ping be accessible w/o authentication for different kinds of health checks. 

I'd be glad to get rid of my current workaround deploying marathon with an apache reverse proxy in the same docker container.

janisz:
It's duplicate of https://github.com/mesosphere/marathon/issues/128

felixb:
It's not. I'm not asking for multi tenancy at all. Just authentication and restriction to r/o mode for some users.

sepiroth887:
while this is not a solution, i run nginx on the mesos-master which also runs marathon. nginx has the ldap-module enabled which should be able to scope pretty well what you need. If you only allow GET to marathon via nginx you should be able to get what you want. 

That said id still like to see granular permissions with Marathon :)

felixb:
I'm doing the same thing with apache, because nginx-ldap does not support STARTTLS.

Anyway, I don't like the idea to run a webserver next to marathon (or chronos) in a single docker container just to do LDAP authentication.
Using marathons groups to marathon/chronos and apache as single containers isn't a solution either. The marathon/chronos instance would be accessible from outside if you know where it's running.

sepiroth887:
True. The permissions model on containers could generally improve to allow those use cases but thats more in dockers realm. 

The problem though is not just with marathon but a lot of other tools with poir or no ldap support. Actually dockers registry 2.0 does bundle nginx internally for ssl termination iirc. 

I would generally love to see more integrations with oauth similar to what rancher does with github. 
This gives a lot of flexibility and granularity for ACLs

air:
Hey @felixb - thanks for the issue! We definitely want to improve Marathon access control.
 
We're working on adding hooks for this in Marathon, as well as some commercial add-ons as part of [DCOS](https://mesosphere.com/learn/).

Could you give a quick example of r/w, r/o users/groups and how it might work for your use case?

felixb:
We currently run one marathon for each team. There is an LDAP group for each team, one for ops/admin and one for read only/monitoring users. The team and ops are allowed to do any requests. Read only is allowed to do GET requests only.
Access one of the css files is granted without login. This is just for the health checks.
I'm quite fine with the current situation. But I'd rather remove the proxy in front of marathon.

janisz:
[...]

PR #1986, opened 2015-08-12
Fixes - 1847 | Add basic LDAP filter

janisz:
PoC for #1847 Simple filter using shiro to perform authentication.

Require configuration shiro.ini on classpath

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 bin/run-tests-config.sh
+src/main/resources/shiro.ini
 .idea
 *.iml
 docker-volumes
--- a/project/build.scala
+++ b/project/build.scala
@@ -215,6 +215,8 @@ object Dependencies {
     sprayClient % "compile",
     sprayHttpx % "compile",
     chaos % "compile",
+    unboundid % "compile",
+    shiro % "compile",
     mesosUtils % "compile",
     jacksonCaseClass % "compile",
     twitterCommons % "compile",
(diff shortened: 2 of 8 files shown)


felixb:
Thank you very much for your effort.

Sadly, I was not able to set this up:

1. When running with default flags, it sends 401 for everything.
2. When trying to set a shiro config, I get `java.io.IOException: Resource [classpath:shiro.ini] could not be found.` for every request.
3. I was unable to find any documentation, on how to set this up to allow only access for certain groups.

janisz:
Hi I'm sorry for not providing any doc. Here is description of how it's working

- If user didn't provide `--shiro_config_path` then every request is pass through
- If user provide `--shiro_config_path` then shiro try to authenticate user and check if he has permissions to do action. Right now I hardcoded two permission groups
    - `user` - able to perform GET and OPTION methods
    - `admin` - able to perform other methods

Here is simple shiro.ini
```
[users]
peter = parker, user
john = doe, admin, user
bruce = banner, admin
steven = rogers, capitan
```
so if you place it in `/tmp/shiro.ini` and set `--shiro_config_path` to it then 
- peter, john can perform GET/OPTIONS 
- bruce, john can perfrom POST/PUT
- steven can do noting

It's just PoC it need some cleanup and tests but it's working. Also it will be great to provide audit log to see who do what.

kolloch:
Hi @janisz, I think unfortunately we are currently too busy to look at that patch and give you feedback soon. We have our own thoughts on how to do authentication and have to check if the approaches are compatible. Sorry & thank you for the patch.

aquamatthias:
Hey @janisz thanks for the PR.

Let me give a short explanation to that topic:
AuthN+AuthZ is a very important topic and we are aware of that.
We spoke to different users of Marathon and gathered very different requirements for AuthN/AuthZ.
As a result we decided to not use one tool directly (shiro is versatile and great but does not solve everything we want to achieve). 

The idea we came up with is: we will introduce interfaces for AuthN+AuthZ into Marathon along with a plugin mechanism. The idea is, that a provider can implement those interfaces according to the specific needs. So it would be possible to use shiro as one provider.

We will start implementing those features into Marathon 0.12.0 (which will start next week).
Until then we won't merge another solution to that idea.

I will close this PR for the reasons outlined above.

kamilchm:
@aquamatthias, please let us know when you'll have some working code, that we can use to start working on custom auth plugins.
We'll extract code from this PR to separate repo, eg. marathon-shiro or sth.

PR #1990, opened 2015-08-13
docs: point to sbt from git

felixb:
Update README pointing to sbt binary checked into git.


kolloch:
Hi @felixb, thanks for your patch.

Doesn't that depend on where you have sbt installed? I have it in my path and not in a subdirectory of marathon. Thus, the current instructions would work for me and your instructions don't.

Or am I missing something obvious?

kolloch:
I will close this for now. Please reopen & elaborate if you disagree.

PR #1991, opened 2015-08-13
Add replicated log storage provider

aquamatthias:
- make the store_backend parameter available (no hidden internal_store_backend)
- added parameter for replicated log parameters

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -161,12 +161,32 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
     noshort = true
   )
 
-  //Internal settings, that are not intended for external use
-  lazy val internalStoreBackend = opt[String]("internal_store_backend",
-    descr = "The backend storage system to use. One of zk, mesos_zk, mem",
+  lazy val storeBackend = opt[String]("store_backend",
+    descr = "The backend storage system to use. One of zk, mesos_zk, mesos_log, mem",
     hidden = true,
-    validate = Set("zk", "mesos_zk", "mem").contains,
+    validate = Set("zk", "mesos_zk", "mesos_log", "mem").contains,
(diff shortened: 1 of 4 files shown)


src/test/scala/mesosphere/util/state/mesos/MesosStateZKStoreTest.scala
```diff
@@ -9,7 +9,7 @@ import org.scalatest.{ ConfigMap, Matchers }
 
 import scala.concurrent.duration._
 
-class MesosStateStoreTest extends PersistentStoreTest with StartedZookeeper with Matchers {
```
kolloch: Was that on purpose?

kolloch: It was, got it know.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -161,14 +161,34 @@ trait MarathonConf extends ScallopConf with ZookeeperConf with IterativeOfferMat
     noshort = true
   )
 
-  //Internal settings, that are not intended for external use
-  lazy val internalStoreBackend = opt[String]("internal_store_backend",
-    descr = "The backend storage system to use. One of zk, mesos_zk, mem",
+  lazy val storeBackend = opt[String]("store_backend",
```
kolloch: We talked about this: Maybe we want to avoid renaming this parameter.

kolloch:
Don't merge. This is in the process of being tested.

src/main/scala/mesosphere/marathon/MarathonConf.scala
aquamatthias: Nobody should know that parameter. Do you think this is a problem?
I would like to reserve `--internal*` for experimental flags.

aquamatthias:
Rebase on master - close this one.
@kolloch thanks for feedback

PR #1992, opened 2015-08-13
REST API documentation: Remove duplicate info about updateStrategy

kolloch:
Also remove the mentioning of the read-only version attribute. It should
ultimately be documented in AppResource_show.md.


aquamatthias:
Thanks for that contribution @kolloch!

PR #1998, opened 2015-08-13
Added SmartProcure to README.md

lusid:
Added SmartProcure to the list of companies using Marathon. We rely on it heavily to power our entire infrastructure now.


aquamatthias:
Great!

PR #1999, opened 2015-08-13
Update deployments.md to include common cause for Docker deploy failures

jgarcia-mesosphere:
Patch to add Docker timeouts/containerizer issues to Troubleshooting


kolloch:
Thanks!

PR #2000, opened 2015-08-13
Add replicated log storage provider

aquamatthias:
based on master

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -184,11 +184,31 @@ trait MarathonConf
     noshort = true
   )
 
-  //Internal settings, that are not intended for external use
-  lazy val internalStoreBackend = opt[String]("internal_store_backend",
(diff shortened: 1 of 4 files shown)


kolloch:
Can we close that?

aquamatthias:
This is based on master - I hope it makes sense there as well.

cmaloney:
Should look at: https://youtu.be/ZQenXkoG_7o for lots of good insights around this (Case Study of Apache Aurora implementing/using the mesos replicated log)

stephentknight:
Will this assist in recovery of Docker machines to their previous state? I understand it aids in recovery of applications (is that correct), since Docker is considered first class, will this mean that in the event of an unexpected reboot I can recover my stateful Docker container deployed via Marathon?

kolloch:
@stephentknight This PR is only related to the state kept directly by Marathon: The App Definitions, Groups, Running Tasks. What you suggest would require Mesos/Docker support.

kolloch:
Since this PR is not mergeable and I do not see progress, I close this PR for now.

Please feel very free to reopen with an updated PR!

PR #2002, opened 2015-08-14
Add sbt tmp files to .gitignore

felixb:
sbt stores its distribution in `/sbt` when running from checked in `sbt.sh` (#1990)


kolloch:
Why did you install sbt in a sub directory of marathon? I still feel I am missing something obvious...

felixb:
Never mind. I installed sbt locally month ago and assumed it was checked in.

PR #2003, opened 2015-08-14
make logger work in python 2.6

flosell:
Adds a positional argument in log statement to make `--listen` work on python 2.6


cmaloney:
Loogs good to me, merging

PR #2004, opened 2015-08-14
servicerouter: remove callbacks when servicerouter terminates

flosell:


bin/servicerouter.py
```diff
@@ -811,6 +811,11 @@ def wsgi_app(env, start_response):
     httpd.serve_forever()
 
 
+def clear_callbacks(marathon, callback_url):
```
cmaloney: nit: I don't see a lot of value in having this be a function rather than be inline. That said, works fine either way.

cmaloney:
Looks good. I'll try and test it out a little bit tomorrow then commit if everything goes well. This definitely helps cleanup this API usage some. Ideally longer-term we'll move to Marathon's event stream API (http://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-events) which means we won't have to do the registration / de-registration stuff.

PR #2005, opened 2015-08-14
servicerouter: integrate health checks from marathon app definition

felixb:
* set service mode if specified in health check
* allow to configure haproxy health checks for tcp and http
  * introduce new templates, empty by default
  * fetch parameters from marathon health check definition

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -96,6 +96,38 @@ Templates:
   HAPROXY_BACKEND_HTTP_OPTIONS
     Sets HTTP headers, for example X-Forwarded-For and X-Forwarded-Proto.
 
+  HAPROXY_BACKEND_HTTP_HEALTHCHECK_OPTIONS
+    Sets HTTP health check options, for example timeout check and httpchk GET.
+    Parameters of the first health check for this service are exposed as:
+      * healthCheckPortIndex
+      * healthCheckProtocol
+      * healthCheckPath
+      * healthCheckTimeoutSeconds
+      * healthCheckIntervalSeconds
+      * healthCheckIgnoreHttp1xx
+      * healthCheckGracePeriodSeconds
+      * healthCheckMaxConsecutiveFailures
+      * healthCheckFalls is set to healthCheckMaxConsecutiveFailures + 1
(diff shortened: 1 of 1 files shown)


bin/servicerouter.py
```diff
@@ -96,6 +96,32 @@
...
+    Parameters of the first health check for this service are exposed as:
+      * healthCheckPortIndex
+      * healthCheckProtocol
+      * healthCheckPath
+      * healthCheckTimeoutSeconds
+      * healthCheckIntervalSeconds
+      * healthCheckIgnoreHttp1xx
+      * healthCheckGracePeriodSeconds
+      * healthCheckMaxConsecutiveFailures
+      * healthCheckFalls is set to healthCheckMaxConsecutiveFailures + 1
```
cmaloney: What does having this enable that you can't just do with healthCheckMaxConsecutiveFailures? The healthCheckFalls also isn't the most obvious name to me.

bin/servicerouter.py
```diff
@@ -577,13 +687,44 @@ def config(apps, groups):
...
+
             backend_server_options = templater.haproxy_backend_server_options
             backends += backend_server_options.format(
                 host=backendServer.host,
                 port=backendServer.port,
                 serverName=serverName,
                 cookieOptions=' check cookie ' +
-                serverName if app.sticky else ''
+                serverName if app.sticky else '',
+                healthCheckOptions=' ' +
```
cmaloney: Why do we need to add the space here? Could we just default healthCheckOptions to empty string (or empty string) and get rid of this if?

cmaloney:
Generally looks good. Would be great if we could provide reasonable defaults or sample templates for the health check chunks. Some comments on the code itself.

bin/servicerouter.py
felixb: It's for configuring haproxies `fall` statement. You want to drop the backend as unhealthy when there are more than `healthCheckMaxConsecutiveFailures` failures.

felixb:
I moved the necessary white space into the template and added some examples to `--longhelp`.
One example explains the use of `healthCheckFalls`.

PR #2007, opened 2015-08-14
Fix #1984 In multi Marathon setup, the zookeeper path is created concurrently

### LINKED ISSUE #1984: Creation of state path in Zookeeper fails when performed concurrently
Environment:
Marathon: 0.9.1
Mesos: 0.23.0
Zookeeper: 3.5.0-alpha
3 AWS EC2 instances(OS: centos 7.0)

I deploy this environment on AWS using Chef. After the deployment completes, the Marathon service on at least one EC2 instance(not limited to the EC2 instance has the leader of Mesos Master) is not available. I cannot access the RESTful API of Marathon. 

From the log("/var/log/message") 
```
Aug 12 06:32:32 vos-node-03 marathon[2559]: [2015-08-12 06:32:32,282] INFO Session establishment complete on server localhost/0:0:0:0:0:0:0:1:2181, sessionid = 0x34f209c213e0007, negotiated timeout = 40000 (org.apache.zookeeper.ClientCnxn:1235)
Aug 12 06:32:32 vos-node-03 marathon[2559]: [2015-08-12 06:32:32,428] INFO Registering in Zookeeper with hostPort:vos-node-03:8080 (mesosphere.marathon.MarathonModule:207)
Aug 12 06:32:32 vos-node-03 marathon[2559]: [2015-08-12 06:32:32,582] INFO Now standing by. Closing existing handles and rejecting new. (mesosphere.marathon.event.http.HttpEventStreamActor:100)
Aug 12 06:32:32 vos-node-03 marathon[2559]: [2015-08-12 06:32:32,584] INFO http_endpoints(List(http://127.0.0.1:8088/api/v1/marathon_event/)) are specified at startup. Those will be added to subscribers list. (mesosphere.marathon.event.http.HttpEventModule:61)
Aug 12 06:32:32 vos-node-03 marathon[2559]: Exception in thread "main" com.google.inject.CreationException: Guice creation errors:
Aug 12 06:32:32 vos-node-03 marathon[2559]:
Aug 12 06:32:32 vos-node-03 marathon[2559]: 1) Tried proxying mesosphere.marathon.MarathonScheduler to support a circular dependency, but it is not an interface.
... ... 
```

More detail log please refer to https://gist.github.com/howardzhang/02c8184f803d3effefa9

Does anyone get any information on this?

aquamatthias:
Every instance tries to create the path during startup.
The exists->create loop for the hierarchy did not take external creation into account.
Fix: if the creation fails with a NodeExistsException it is translated as success.

### DIFF
--- a/src/main/scala/mesosphere/util/state/zk/ZKStore.scala
+++ b/src/main/scala/mesosphere/util/state/zk/ZKStore.scala
@@ -12,7 +12,7 @@ import mesosphere.util.state.zk.ZKStore._
 import mesosphere.util.state.{ PersistentEntity, PersistentStore }
 import org.apache.log4j.Logger
 import org.apache.zookeeper.KeeperException
-import org.apache.zookeeper.KeeperException.NoNodeException
+import org.apache.zookeeper.KeeperException.{ NodeExistsException, NoNodeException }
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{ Await, Future, Promise }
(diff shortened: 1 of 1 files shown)


aquamatthias:
This problem is in Marathon 0.9 branch and 0.10 branch as well.

aquamatthias:
@spahl This is the fix. Once it is merged, we need to cherry pick that onto the 0.9, 0.10 release

PR #2008, opened 2015-08-14
servicerouter: select http backend by http header X-Marathon-App-Id

felixb:
you can use the following config to specify backends by http header 'X-MARATHON-APP-ID' with the following config:

HAPROXY_HTTP_FRONTEND_ACL:

    acl app_{cleanedUpHostname} hdr(x-marathon-app-id) -i {appId}
    use_backend {backend} if app_{cleanedUpHostname}

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -83,6 +83,15 @@ Templates:
     An HTTP frontend that binds to port *:80 by default and gathers
     all virtual hosts as defined by the HAPROXY_{n}_VHOST variable.
 
+  HAPROXY_HTTP_FRONTEND_APPID_HEAD
+    An HTTP frontend that binds to port *:81 by default and gathers
+    all apps in http mode.
+    To use this frontend to forward to your app, configure the app with
+    "HAPROXY_0_MODE=http" then you can access it via a call to the :81 with
+    the header "X-Marathon-App-Id" set to the Marathon AppId.
+    Note multiple http ports being exposed by the same marathon app are not
+    supported. Only the first http port is available via this frontend.
+
   HAPROXY_HTTPS_FRONTEND_HEAD
     An HTTPS frontend for encrypted connections that binds to port *:443 by
     default and gathers all virtual hosts as defined by the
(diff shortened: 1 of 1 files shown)


cmaloney:
Any particular reason to not just hard-code this to always be the header X-Marathon-AppId? That way we end up with one standard way of configuring this, I don't see a lot of value in allowing other arbitrary headers to be used here.

It seems like it should be possible to have this not conflict with the virtual hosts stuff, and instead output two frontend definitions (one for if _VHOST is given, and another for the x-marathon-appid header).

Then this basically becomes a flag to "enable appid routing". I get people might not want to always use vhost and appid routing at the same time, but I can see a number of use cases where it would be handy to have.

felixb:
How about this solution?

bin/servicerouter.py
```diff
@@ -547,17 +577,38 @@ def config(apps, groups):
...
+            if http_appid_frontends:
+                http_appid_frontend_acl = templater.haproxy_http_frontend_appid_acl
+                http_appid_frontends += http_appid_frontend_acl.format(
+                    cleanedUpAppId=cleanedUpAppId,
+                    hostname=app.hostname,
+                    appId=app.appId,
+                    backend=backend
+                )
+
+        if app.mode == 'http':
```
cmaloney: I don't see what the value of this extra if is, couldn't this statement just be part of the previous `app.mode == 'http'`? Possibly reorder the line up a little bit higher so it isn't after the http_appid_frontends bit for readability.

bin/servicerouter.py
```diff
@@ -83,6 +83,10 @@
     An HTTP frontend that binds to port *:80 by default and gathers
     all virtual hosts as defined by the HAPROXY_{n}_VHOST variable.
 
+  HAPROXY_HTTP_FRONTEND_APPID_HEAD
+    An HTTP frontend that binds to port *:81 by default and gathers
+    all apps in http mode.
```
cmaloney: Could you add documentation on how an app should use this. Something like:

To use this frontend to forward to your app, configure the app with "HAPROXY_0_MODE=http" then you can access it via  a call to the <servicerouter_host>:81 with the header "X-Marathon-App-Id" set to the Marathon AppId. Note multiple ports being exposed by the same marathon app are not supported


What happens when the same App exposes multiple ports as http? Can we explicitly emit a warning when this happens (and ignore the app entirely) or make the forwarding {app-id}/{port-index}?

cmaloney:
This looks good to me. If you squash down and fix the last pep8 error TravisCI found, I will do a final sanity check and merge it.

felixb:
Squashed and documented.

PR #2026, opened 2015-08-17
Fix #1819 First step of migration should make sure the zookeeper path exists

### LINKED ISSUE #1819: First step of migration should make sure the zookeeper path exists
We have some reports from users, that started marathon and then removed the zookeeper node /marathon/state. Subsequent leaders tried to migrate, but that failed, since the path did not exist.
During startup Marathon creates the path if not existing, but afterwards not.

aquamatthias:


src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -60,8 +60,14 @@ class Migration @Inject() (
...
   }
 
+  def initializeStore: Future[Unit] = store match {
+    case manager: PersistentStoreManagement => manager.initialize()
+    case _: PersistentStore                 => Future.successful(())
+  }
+
   def migrate(): StorageVersion = {
     val result = for {
+      _ <- initializeStore
```
kolloch: since this method has side effects, I prefer `()`

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -90,22 +87,26 @@ class ZKStore(val client: ZkClient, rootNode: ZNode) extends PersistentStore {
...
+    def nodeExists(node: ZNode): Future[Boolean] = node.exists().asScala
+      .map(_ => true)
+      .recover { case ex: NoNodeException => false }
+      .recover(exceptionTransform("Can not query for exists"))
+
+    def createNode(node: ZNode): Future[ZNode] = node.create().asScala
+      .recover { case ex: NodeExistsException => node }
+      .recover(exceptionTransform("Can not create"))
+
+    def createParent(node: ZNode): Future[ZNode] = {
```
kolloch: Better use `createParentsAndNode` or something like that. This way it is misleading.

src/main/scala/mesosphere/util/state/PersistentStore.scala
```diff
@@ -79,3 +79,16 @@ trait PersistentStore {
...
+/**
+  * Optional interface if the store also has management infrastructure.
+  */
+trait PersistentStoreManagement {
+
+  /**
+    * Initialize the store.
+    * This method is called before any PersistentStore operation is called.
+    * Specific setup logic should be performed here.
+    * @return A future to indicate, when the initialization logic is finished.
```
kolloch: No comma before when (ignore if you want ;) )

kolloch:
Some minor comments, otherwise good. But what about tests? I know that at least the initialize is executed but it would be good to:

* test it with deeply nested paths
* existing paths
* and test that the migration calls it correctly

aquamatthias:
Incorporated your feedback

kolloch:
Awesome. Merge, bam. :+1:

PR #2028, opened 2015-08-18
Use consolidated info of the LaunchQueue to determine

kolloch:
how many more tasks we have to launch.

I hope that will have an effect on the phenomena that we see during scaling: For large numbers of instances that are launched quickly, hitting the target count exactly takes a long time.


gkleiman:
:+1: :+1:

PR #2034, opened 2015-08-18
Update visualiser to 1.6 stable

pierlo-upitup:


aquamatthias:
That looks awesome to me!

PR #2045, opened 2015-08-19
Check servicerouter is PEP-8 compliant

cmaloney:


stephentknight:
Gents,

I appreciate this is not a support page, but I am looking for some community help on servicerouter.py -- can you point me to the right place for some expert advice?

Thanks

PR #2056, opened 2015-08-20
WIP! Pk/app version info

kolloch:
I would appreciate if you already had a look. I will write tests for this still


kolloch:
I will divide this further into its sub problems...

PR #2073, opened 2015-08-24
Fixes #722 - Use play JSON for JSON serialization

### LINKED ISSUE #722: Replace JSON libraries with play-json
We currently use different JSON libraries for the endpoints and the events API. Additionally the reflection used by json4s currently causes exceptions in the events api (see: https://github.com/mesosphere/marathon/issues/646). Another benefit of using play-json exclusively is that we can remove the somewhat verbose code for custom Jackson de/serialization.

kolloch:
I think that the serialization tests should actually be moved to the json package and consolidated.
I did not want to do that as part of this patch as well, though.


kolloch:
@gkleiman: I don't get the compile error on TeamCity. Any ideas?

src/main/scala/mesosphere/marathon/api/v2/DeploymentsResource.scala
```diff
@@ -51,16 +53,19 @@ class DeploymentsResource @Inject() (
 
   private def toInfo(
     deployment: DeploymentPlan,
-    currentStepInfo: DeploymentStepInfo): Map[String, Any] =
-    Map(
+    currentStepInfo: DeploymentStepInfo): JsObject = {
+
+    val steps = deployment.steps.map(step => step.actions.map(actionToMap)).map(Json.toJson(_))
```
gkleiman: is the indentation here correct?

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
```diff
@@ -110,37 +112,26 @@ trait MarathonTestHelper {
...
     val factory = JsonSchemaFactory.byDefault()
     factory.getJsonSchema(appDefinition)
   }
-  val appSchema = getAppSchema()
 
   def validateJsonSchema(app: V2AppDefinition, valid: Boolean = true) {
-    val appStr = schemaMapper.writeValueAsString(app)
+    import mesosphere.marathon.api.v2.json.Formats._
+    // TODO: Revalidate the decision to disallow null values in schema
+    // Possible resolution: Do not render null values in our formats by default anymore.
```
gkleiman: I like the idea of not rendering null values anymore.

src/test/scala/mesosphere/marathon/api/MarathonExceptionMapperTest.scala
```diff
@@ -18,12 +22,36 @@ class MarathonExceptionMapperTest extends MarathonSpec with GivenWhenThen with M
...
+    val mapper = new MarathonExceptionMapper()
+
+    When("The mapper creates a response from this exception")
+    val response = mapper.toResponse(ex)
+
+    Then("The correct response is created")
+    response.getStatus should be(422)
+    val entityString = response.getEntity.asInstanceOf[String]
+    val entity = Json.parse(entityString)
+    (entity \ "message").as[String] should be("Bean is not valid")
```
gkleiman: I know that this is the current error as well, but "bean" is a java concept that we shouldn't leak to the users.

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
kolloch: Me, too. But this isn't part of this PR in my opinion.

I created https://github.com/mesosphere/marathon/issues/2081

gkleiman:
LGTM, except for the failing integration test. It might be flaky one, but the log mentions Jackson...

gkleiman:
Merged, goodbye Jackson ;-)

PR #2080, opened 2015-08-25
Fix TaskBuilderTest: Allow multiple port ranges in various tests

kolloch:
...which might happen (rarely) because of the port randomization.


meichstedt:
LGTM

PR #2083, opened 2015-08-25
servicerouter.py: Only generate backends with valid hostname

first-it-consulting:
Implement new command line parameter `--check-hostname` to only generate backends can be resolved by a dns server

Reason:
HAProxy has the same check when validating the configuration, so generated configuration with invalid hostnames would always fail reloading

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -187,6 +187,8 @@ Templates:
 Operational Notes:
   - When a node in listening mode fails, remove the callback url for that
     node in marathon.
+  - If run in listening mode, DNS isn't re-resolved. Restart the process
+    periodically to force re-resolution if desired.
 """
 
 from logging.handlers import SysLogHandler
@@ -205,6 +207,7 @@ import re
 import requests
 import subprocess
 import sys
+import socket
 import time
 
 
@@ -286,7 +289,7 @@ class ConfigTemplater(object):
 '''
 
     HAPROXY_BACKEND_SERVER_OPTIONS = '''\
-  server {serverName} {host}:{port}{cookieOptions}{healthCheckOptions}
+  server {serverName} {host_ipv4}:{port}{cookieOptions}{healthCheckOptions}
 '''
 
     HAPROXY_BACKEND_SERVER_HTTP_HEALTHCHECK_OPTIONS = ''
@@ -593,6 +596,22 @@ def has_group(groups, app_groups):
 
     return False
 
+ip_cache = dict()
+
+
+def resolve_ip(host):
+    cached_ip = ip_cache.get(host, None)
+    if cached_ip:
+        return cached_ip
+    else:
+        try:
+            logger.debug("trying to resolve ip address for host %s", host)
+            ip = socket.gethostbyname(host)
+            ip_cache[host] = ip
+            return ip
+        except socket.gaierror:
+            return None
+
 
 def config(apps, groups):
     logger.info("generating config")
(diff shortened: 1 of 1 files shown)


cmaloney:
I'm worried some about performance here (That could be a lot of extra getaddrinfo calls), as well as differences in what is seen between when we're doing the getaddrinfo checks vs. what haproxy sees.

I don't see how this eliminates entirely the chances of giving haproxy a config it will reject / refuse to reload since it is very possible (And if using mesos-dns somewhat likely) that when servicerouter runs and does it's checks everything has valid host names, but when haproxy runs it does the same getaddrinfo calls and doesn't find the hostnames.

Was there some particular system failure you saw with this? Should we maybe configure the backends purely with ip addresses rather than host-names? Was this just a hard to detect error?

first-it-consulting:
Indeed, we had a problem in our production environment that the haproxy wrote a wrong configuration file and could not reload. 

This was because a slave went down but tasks where still active in marathon and therefore a new config was written pointing to a hostname which doesn't exits anymore. 

You are right that this might slow down the performance because of extra getaddrinfo calls, but it would solve at least some problems. 

We are working on reducing the getaddrinfo calls.

cmaloney:
So I think what we want to do with this then is to resolve the hostnames in servicerouter, and write the config always using IP addresses. Then we'll never have haproxy refuse to load the config script because the hostnames don't resolve. If a hostname doesn't resolve, then that particular config entry is dropped and we emit a log message (that way monitoring software can pick it up). That makes it so the hostnames are only ever resolved once, and if a hostname is resolvable when servicerouter runs but not when haproxy reloads, haproxy will still reload the configuration.

Sound reasonable?

flosell:
Sounds like a good idea. We reworked the pull request this way. 

You can now use `host_ipv4` in the `HAPROXY_BACKEND_SERVER_OPTIONS` template. IP addresses for hostnames are cached and backends where IP addresses could not be resolved are ignored (and not cached)

cmaloney:
Generally looks good. 

Could the ip lookup cache only live for a single iteration of the config() function though, I know of a couple clusters where hostnames change what they point to quite a bit, and just caching over a single run of config() should work well as a caching point (All the tasks which are on the same host the hostname is only resolved once, we re-resolve each run once). With moving the caching there, we should cache failed lookups so we don't have growing failures (Hostnames start failing to resolve -> increased load through dozens/hundredes of hostname lookup failures for the tasks that were on those hosts -> more hosts / services fail)

Can you change the default template to use host_ipv4? I'd rather have people have the better operational behavior by default, rather than running into problems at some point down the line and finding out "oh, there was a flag I should have changed". I don't see an instance where that change would be particularly bad for any operator since haproxy is doing the same resolution itself.

felixb:
When run with http callbacks `config()` is run every few seconds. That's why we moved the cache to the global context. 

What do you think of a cache with TTL of a few minutes?

cmaloney:
With #2115 perf should be considerably better so there should be a lot more headroom in the refresh interval (I hope)... Really should teach to consume server sent events from marathon for listening mode and do partial updates (Would fix a bunch of these issues)...

I think the logic around TTL is going to be complex / confusing to understand. I'm up for just leaving the cache global if it's that much of a time saver and adding some documentation of "If run in listening mode, DNS isn't re-resolved. Restart the process periodically to force re-resolution if desired".

flosell:
updated documentation and default template.

PR #2087, opened 2015-08-25
Fixes #1898 - Sync AppRepository writes with GroupRepository writes

### LINKED ISSUE #1898: Rewrite the GroupRepository/AppRepository logic and make it consistent
Current state:
- all applications are stored as part of the root group in the group repository for every user intended change
- all applications are stored again in the app repository, if the deployment of that application starts

Future plan:
- the applications should be always loaded via the groupManager or groupRepository.
- the app repository is used to load/store versions of the application
- the app repository is not used for load/store app definitions
- all references to app repo are cleaned up

kolloch:
Reads still partially go to the AppRepository for the latest
app version which should be fine after this change.

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -84,7 +84,11 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.MethodLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxLength"><![CDATA[50]]></parameter>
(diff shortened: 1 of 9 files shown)


meichstedt:
LGTM – only minor nit picking regarding the test and the the comment; and I'd like to see the appRepo dependency removed. Otherwise, yearning to merge :8ball:

Issue #2091, opened 2015-08-26
Multiple ports are incorrectly matched


bobrik:
Marathon 0.10.0, Mesos 0.23.0, slaves announce 8000-10000 port range.

```json
{
  "id": "/hey/test",
  "apps": [
    {
      "id": "/hey/test/sleeper",
      "cmd": "echo $PORT0 $PORT1 && sleep 3600",
      "ports": [9200, 9300],
      "cpus": 0.1,
      "instances": 1,
      "mem": 16
    }
  ]
}

```

Task gets 2 random ports.

If you only ask for `[9200]`, tasks always get port 9200. I expect the same with many specified ports.

gkleiman:
If you want Marathon not to assign random ports, then you have to set `requirePorts` to `true` in the app definition.

bobrik:
@gkleiman thanks, that worked.

> If you need more control and want to specify your host ports in advance, you can set requirePorts to true. This way the ports you have specified are used as host ports. That also means that Marathon can schedule the associated tasks only on hosts that have the specified ports available.

> The specified ports need to be in the local port range specified by the --local_port_min and --local_port_max flags. See Command Line Flags doc page).

I didn't change local port range, why did it work?

Also, I'd love to see some consistency. If apps gets assigned random ports with 2 ports, it should get random ports with 1 port as well. People start with 1 port and make false assumptions with the way it works now.

gkleiman:
The following statement is wrong and I've just removed it from our documentation:

>The specified ports need to be in the local port range specified by the --local_port_min and --local_port_max flags. See Command Line Flags doc page).

I agree that it would be more consistent to always assign random ports if `requirePorts` is `false`. @kolloch agrees with me and will do the necessary code change soon.

Thanks for having opened this issue!

bobrik:
:+1: 

Please don't forget to mention that when you do a release, this can be a surprising change for many.

PR #2096, opened 2015-08-26
Improves #2091 - Remove incorrect statement from the docs

gkleiman:
The --local_port_min and --local_port_max flags define the range of
ports to use when automatically assigning service ports and don't
influence the app port allocation.


kolloch:
Awesome!

PR #2110, opened 2015-08-27
Fixes #2011 - Provide app versioning information

### LINKED ISSUE #2011: Export summary about the most recent configuration change (API)
For debugging problematic apps, it is useful to have a quick way 

* when the last real configuration change was (e.g. the command changed, the resources changed, ...)
* when the last scaling operation was (including manual restarting of all tasks but excluding manual killing of single tasks)  

See #1951

kolloch:
In addition:
* Provides a migration for #1898 as well (latest app info in group has to win)
* Nicer logging of DeploymentPlans


changelog.md
```diff
@@ -4,9 +4,15 @@
 
 * `--revive_offers_for_new_apps` is now the default. If you want to avoid resetting filters
   if new tasks need to be started, you can disable this by `--disable_revive_offers_for_new_apps`.
-
+  
 ### Overview
 
+#### New versioning information in API and UI
+
+We know have `"versionInfo"` with `"lastConfigChangeAt"` and `"lastScalingAt"` in the apps JSON of our API.
```
meichstedt: s/know/now

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -215,24 +215,11 @@ class AppsResource @Inject() (
     val appId = id.toRootPath
     val newVersion = Timestamp.now()
     def setVersionOrThrow(opt: Option[AppDefinition]) = opt
-      .map(_.copy(version = newVersion))
+      .map(app => app.copy(versionInfo = AppDefinition.VersionInfo.NoVersion))
```
meichstedt: `setVersionOrThrow` now doesn't set the version anymore. At this point, it's not clear to me whether `NoVersion` (with a `version = Timestamp(0)`) actually is wanted, or overridden later on.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -243,7 +230,14 @@ class AppsResource @Inject() (
     def updateApp(current: AppDefinition) = validateApp(appUpdate(current))
     def rollback(version: Timestamp) = service.getApp(appId, version).getOrElse(throw new UnknownAppException(appId))
     def updateOrRollback(current: AppDefinition) = appUpdate.version.map(rollback).getOrElse(updateApp(current))
-    existing.map(updateOrRollback).getOrElse(createApp()).copy(version = newVersion)
+
+    existing match {
```
meichstedt: That's definitely more readable – thanks :)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -237,17 +259,80 @@ case class AppDefinition(
...
+      */
+    case class OnlyVersion(version: Timestamp) extends VersionInfo
+
+    /**
+      * @param version The versioning timestamp (we are currently assuming that this is the same as lastChangeAt)
+      * @param lastScalingAt The time stamp of the last change including scaling or restart changes
+      * @param lastConfigChangeAt The time stamp of the last change that changed configuration
+      *                               besides scaling or restarting
+      */
+    case class FullVersionInfo(
```
meichstedt: minor nit-picking, but since it's about naming I'd like to share my thoughts:
What I like about `NoVersion` and `OnlyVersion` is, that both make clear, that these objects are somewhat exceptional/corner cases. However, `FullVersionInfo`, besides being the only one with `Info` suffix, _sounds_ like it's somewhat _unusual_ because of the `full` prefix, which is not the case. I currently have no better proposal at hand, but if you agree, let's have a quick chat about a nicer name :)

PR #2111, opened 2015-08-27
Fixes #2107 - Don't try to use "ports" as host ports anymore

### LINKED ISSUE #2107: Do not try to deterministically assign host ports anymore
The values in the ports field of the AppDefinition are either 0 or they specify the service ports of the app. Confusingly, if available, our code still currently tries to assign these ports as host ports if they are available in the offer. If not we fall back to randomized port assignment.

That can lead to the misunderstanding, the the values in the port arrays are host ports.

We should always use randomized port assignment.

kolloch:


gkleiman:
:+1:

PR #2115, opened 2015-08-27
servicerouter: improve performance

flosell:
instead of pulling all tasks and then pulling all apps from marathon, this pull request changes `get_apps` to pull all apps and tasks at once using `/v2/apps?embed=apps.tasks`

On a marathon with ~50 apps, this reduced the time to regenerate the config after an event from about 30 seconds to less than two (measured as in #2093):

Before:
```
2015-08-26 16:05:34,551 servicerouter DEBUG updating tasks finished, took 43.9271931648 seconds
2015-08-26 16:06:48,061 servicerouter DEBUG updating tasks finished, took 39.2305569649 seconds
2015-08-26 16:07:23,728 servicerouter DEBUG updating tasks finished, took 35.6545169353 seconds
2015-08-26 16:07:56,529 servicerouter DEBUG updating tasks finished, took 32.7971820831 seconds
```

After:
```
2015-08-26 16:36:56,770 servicerouter DEBUG updating tasks finished, took 1.21059083939 seconds
2015-08-26 16:37:19,629 servicerouter DEBUG updating tasks finished, took 1.67731595039 seconds
2015-08-26 16:37:21,033 servicerouter DEBUG updating tasks finished, took 1.39936995506 seconds
2015-08-26 16:37:22,669 servicerouter DEBUG updating tasks finished, took 1.63086915016 seconds
2015-08-26 16:37:23,522 servicerouter DEBUG updating tasks finished, took 0.84465098381 seconds
2015-08-26 16:37:36,860 servicerouter DEBUG updating tasks finished, took 1.35650086403 seconds
```

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -499,8 +499,8 @@ class MarathonService(object):
 
 class MarathonApp(object):
 
-    def __init__(self, marathon, appId):
-        self.app = marathon.get_app(appId)
+    def __init__(self, marathon, appId, app):
+        self.app = app
         self.groups = frozenset()
         self.appId = appId
 
@@ -523,11 +523,9 @@ class Marathon(object):
     def api_req_raw(self, method, path, body=None, **kwargs):
         for host in self.__hosts:
             path_str = os.path.join(host, 'v2')
-            if len(path) == 2:
-                assert(path[0] == 'apps')
-                path_str += '/apps/{0}'.format(path[1])
-            else:
-                path_str += '/' + path[0]
+
+            for path_elem in path:
+                path_str = path_str + "/" + path_elem
             response = requests.request(
                 method,
                 path_str,
@@ -537,6 +535,8 @@ class Marathon(object):
                 },
                 **kwargs
             )
+
+            logger.debug("%s %s", method, response.url)
             if response.status_code == 200:
                 break
         if 'message' in response.json():
@@ -559,7 +559,8 @@ class Marathon(object):
     # Lists all running apps.
     def list(self):
         logger.info('fetching apps')
-        return self.api_req('GET', ['apps'])["apps"]
+        return self.api_req('GET', ['apps'],
+                            params={'embed': 'apps.tasks'})["apps"]
 
     def tasks(self):
         logger.info('fetching tasks')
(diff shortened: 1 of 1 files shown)


bin/servicerouter.py
```diff
@@ -498,8 +498,11 @@ def __repr__(self):
 
 class MarathonApp(object):
 
-    def __init__(self, marathon, appId):
-        self.app = marathon.get_app(appId)
+    def __init__(self, marathon, appId, app=None):
```
cmaloney: I'd just make app mandatory here, these classes are only used for servierouter.

flosell:
fixed, made the app mandatory

bin/servicerouter.py
```diff
@@ -558,7 +562,7 @@ def get_app(self, appid):
     # Lists all running apps.
     def list(self):
         logger.info('fetching apps')
-        return self.api_req('GET', ['apps'])["apps"]
+        return self.api_req('GET', ['apps', '?embed=apps.tasks'])["apps"]
```
cmaloney: To pass in this sort of arguments we want to use kwargs to pass it through to the requests library like http://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls

So `return self.api_req('GET', ['apps'], params={'embed': 'apps.tasks'})["apps"]`

cmaloney:
Still WIP on the review / working through it, should be more before too long. I'll comment again when I'm done (in and out of meetings)

flosell:
fixed the kwargs issue

bin/servicerouter.py
```diff
@@ -823,64 +824,60 @@ def compareWriteAndReloadConfig(config, config_file):
...
-
-            app = apps[appId]
-            if servicePort not in app.services:
-                app.services[servicePort] = MarathonService(
-                    appId, servicePort, get_health_check(app, i))
+            service_ports = app['ports']
+            for i in xrange(len(service_ports)):
+                servicePort = service_ports[i]
+                port = task['ports'][i] if len(task['ports']) else servicePort
+                if servicePort not in marathon_app.services:
```
cmaloney: So how this was all setup because it did a task list then lazily fetched the apps is structurally a bit backwards from what it should look like now that we are starting at the "apps" and then iterating over the tasks inside (Wish I'd been aware apps had the tasks available if you asked right...).

I'm fine in this PR just merging with the old code structure since the speed improvement is so drastic and things work.

Longer term what should happen is there is a loop over the "ports" array in the base app, that builds out the set of services. Then in looping over the tasks just attach them to services, rather than doing the "we aren't sure if this port/service exists yet" check.

I think there may be some latent issues around deployments which will need to be sorted out (There are two versions of an app during a deployment with different sets of ports, ServiceRouter config, etc.). Fortunately most apps don't have changing sets of ports so it hasn't come up yet. (Switching to be mesos task label based rather than marathon could help there).

cmaloney:
Finished reviewing, only thing which is left to fix would be to make it so we fill out services first then just assume the service entry exists when looping over tasks rather than the existing sort of inverted logic currently there. Welcome to fix now or just leave it (Code works fine as is, and this is a massive improvement).

When you're happy with things rebase down to a single commit and I'll sanity-check and merge the PR

flosell:
squashed commits and refactored `get_apps` along the lines you outlined.

PR #2118, opened 2015-08-28
WIP: servicerouter: add SSE capabilities

felixb:
This enables servicerouter to switch to SSE instead of HTTP callbacks for processing events.

Still to do:

* [x] sseclient maintainer needs to merge and release patch to support marathons event stream (see https://bitbucket.org/btubbs/sseclient/pull-requests/5/support-r-n-r-n-line-separators)
* [ ] update app from event instead of fetching it from API
* [ ] make sure to follow events from marathon leader
* [ ] follow marathon leader if it changes

### DIFF
--- a/bin/servicerouter.py
+++ b/bin/servicerouter.py
@@ -195,6 +195,8 @@ from shutil import move
 from tempfile import mkstemp
 from textwrap import dedent
 from wsgiref.simple_server import make_server
+from sseclient import SSEClient
+
 import argparse
 import json
 import logging
@@ -577,6 +579,12 @@ class Marathon(object):
                 ['eventSubscriptions'],
                 params={'callbackUrl': callbackUrl})
 
+    def get_event_stream(self):
+        url = self.__hosts[0]+"/v2/events"
+        logger.info(
+            "SSE Active, trying fetch events from from {0}".format(url))
+        return SSEClient(url)
+
 
 def has_group(groups, app_groups):
     # All groups / wildcard match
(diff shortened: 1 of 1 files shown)


aquamatthias:
@felixb @flosell this PR is open for a month now. Do you still work on this?

flosell:
Sorry, we didn't get around to it lately. We'll probably continue once v0.11.0 is released since the fix for #1926 should make it way easier: If we get this right, we could probably just read from any marathon without following the leader.

lloesche:
@felixb did you test this in a reasonably busy Marathon? I realise that this is using the same slow mechanism the subscriber method was using. The current implementation spends a lot of time in https://github.com/otto-de/marathon/blob/servicerouter/sse/bin/servicerouter.py#L864
Changing this requires a major rewrite. So I'm not sure if we should add more code to an already bad implementation.

flosell:
This is one of the tasks that is still open: "update app from event instead of fetching it from API"

cmaloney:
@lloesche the performance profile has already gotten a lot better for servicerouter because of https://github.com/mesosphere/marathon/pull/2115 and https://github.com/mesosphere/marathon/pull/2083, the time taken is quite a lot shorter / considerably more performant.

lloesche:
@cmaloney Awesome! I wasn't aware this had been fixed. In that case I guess this is fine.

lloesche:
@flosell would you like this to be merged as is or do you want to keep it open and add to it later?

flosell:
It's very much work in progress right now, I don't think it's ready to merge. 
Unfortunately, our teams focus has shifted for now since the existing servicerouter works well enough for us after the performance improvements so we probably won't be able to finish SSE in the immediate future. 

Feel free to close this pull request for now and we'll reopen it once we get to it.

discordianfish:
@flosell Thanks for the PR though, I cherry-picked it for mesos-lb which will (likely) replace servicerouter here.

PR #2119, opened 2015-08-28
Make AppDeployIntegrationTest less shaky

aquamatthias:


src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -314,22 +313,18 @@ class AppDeployIntegrationTest
 
   test("kill all tasks of an App with scaling") {
     Given("a new app with multiple tasks")
```
gkleiman: If you change the instances count, then you also have to change this description ;-).

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -314,22 +313,18 @@ class AppDeployIntegrationTest
...
-    val app = v2AppProxy(testBasePath / "tokill", "v1", instances = 2, withHealth = false)
+    val app = v2AppProxy(testBasePath / "tokill", "v1", instances = 1, withHealth = false)
```
gkleiman: If you have only 1 instance, how can you be sure that killAllTasks will kill > 1 task?

kolloch: And we DID have errors like that in the past. I also would like it if you kept the count at 2.

aquamatthias: The only reason I switched to 1 is: to be able to identify the kill event __and__ the deployment success. Reverted and removed the waitForEvents method.

PR #2122, opened 2015-08-28
Use jetty9 SSE Servlet

aquamatthias:
Send event, not only data (event: eventType, data: json)
Improve the delivery by using not more than one thread per SSEHandle


meichstedt:
Nice! Besides the mentioning of Akka FSM this LGTM

PR #2126, opened 2015-08-30
Fix #1931 decline with configurable duration.

### LINKED ISSUE #1931: Marathon should set a non-default filter when declining offers

aquamatthias:
Add command line flag and decline with the configured duration.
The timeout case for the Offer is not handled in this PR: https://github.com/mesosphere/marathon/blob/mv/fix_1931_b/src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala#L132-L135


aquamatthias:
@kolloch would be interested in your feedback as well

src/main/scala/mesosphere/marathon/core/matcher/base/OfferMatcher.scala
```diff
@@ -24,8 +24,13 @@ object OfferMatcher {
     * A MatchedTasks reply does not guarantee that these tasks can actually be launched.
     * The launcher of message should setup some kind of timeout mechanism and handle
     * taskLaunchAccepted/taskLaunchRejected calls appropriately.
+    *
+    * @param offerId the identifier of the offer
+    * @param tasks the tasks that should be launched on that offer
+    * @param resendThisOffer true, if this offer could not be processed completely (e.g. timeout)
+    *                        and should be resend and reprocessed again
```
kolloch: Nit: s/reprocessed/processed/ or omit the "again" (otherwise it's a bit repetitive)

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
```diff
@@ -210,15 +217,18 @@ private class OfferMatcherManagerActor private (
           .recover {
             case NonFatal(e) =>
               log.warning("Received error from {}", e)
-              MatchedTasks(data.offer.getId, Seq.empty)
+              MatchedTasks(data.offer.getId, Seq.empty, resendThisOffer = false)
```
kolloch: should be true if this is a temporary error.

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
```diff
@@ -132,14 +134,14 @@ private class OfferMatcherManagerActor private (
...
-        OfferMatcher.MatchedTasks(offer.getId, Seq.empty))
+        MatchTimeout(offer.getId))
 
       // process offer for the first time
       scheduleNextMatcherOrFinish(data)
   }
 
   private[this] def receiveMatchedTasks: Receive = {
-    case OfferMatcher.MatchedTasks(offerId, addedTasks) =>
+    case OfferMatcher.MatchedTasks(offerId, addedTasks, _) =>
```
kolloch: We should evaluate the `resendThisOffer` flag here. If any sub matcher responded with `resendThisOffer=true`, the manager should do that as well.

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
```diff
@@ -178,6 +180,11 @@ private class OfferMatcherManagerActor private (
         case None =>
           addedTasks.foreach(_.reject(s"offer '${offerId.getValue}' already timed out"))
       }
+
+    case MatchTimeout(offerId) =>
+      // When the timeout is reached, we will answer with all matching tasks we found until then
+      // Since we can not be sure if we found all matching tasks, we set resendThisOffer to true
```
kolloch: Nits: s/can not/cannot/, end sentences with full stop.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -75,7 +75,7 @@ class OfferProcessorImplTest extends MarathonSpec {
     Await.result(offerProcessor.processOffer(offer), 1.second)
 
     verify(offerMatcher).matchOffer(deadline, offer)
-    verify(taskLauncher).declineOffer(offerId)
+    verify(taskLauncher).declineOffer(offerId, None)
```
kolloch: A named parameter would improve readability.

src/test/scala/mesosphere/marathon/core/launcher/impl/TaskLauncherImplTest.scala
```diff
@@ -47,13 +47,19 @@ class TaskLauncherImplTest extends MarathonSpec {
   test("declineOffer without driver") {
     driverHolder.driver = None
 
-    launcher.declineOffer(offerId)
+    launcher.declineOffer(offerId, None)
```
kolloch: A named parameter would improve readability.

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
```diff
@@ -59,6 +59,8 @@ private[manager] object OfferMatcherManagerActor {
       )
     }
   }
+
+  private case class MatchTimeout(offerId: OfferID)
```
kolloch: I like this. Much clearer than reusing another more general message :+1:

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
kolloch: This test would be more fun if we actually also tested with a custom decline duration.

src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
```diff
@@ -56,7 +56,9 @@ private[launcher] class OfferProcessorImpl(
             }
           }
           else {
-            taskLauncher.declineOffer(offerId)
+            //if the offer should be resend, than we ignore the configured decline offer duration
```
gkleiman: s/resend/resent/

PR #2136, opened 2015-08-31
Fixes #2130 - Store task failures in Zookeeper

### LINKED ISSUE #2130: Last task failures are not persisted
The `TaskFailureRepository` references an `EntityStore` but doesn't use it. It uses an in-memory map instead to save the last task failures. Even though, the last task failures are not the most important state, this can still lead to strange behavior on fail over or restart when API consumers see the "lastTaskFailure" disappear.

kolloch:


kolloch:
I wanted to move this to 0.12 but it was so easy to write tests for this that I proceeded.

src/main/scala/mesosphere/marathon/state/TaskFailureRepository.scala
```diff
@@ -1,17 +1,20 @@
...
+class TaskFailureRepository(
+  protected val store: EntityStore[TaskFailure],
+  protected val maxVersions: Option[Int] = Some(1),
+  protected val metrics: Metrics)
+    extends EntityRepository[TaskFailure] {
 
-  def store(id: PathId, value: TaskFailure): Unit =
-    synchronized { taskFailures(id) = value }
+  def store(id: PathId, value: TaskFailure): Future[TaskFailure] =
+    super.storeWithVersion(id.safePath, value.version, value)
```
aquamatthias: Why do we store versions of a failure? 
Do we store only the last failure? (Should that be reflected in the class name?)
Should there be one entry per application (and version?) ?

src/test/scala/mesosphere/marathon/state/TaskFailureRepositoryTest.scala
```diff
@@ -0,0 +1,74 @@
...
+
+    And("fetch it")
+    val readFailure = f.taskFailureRepo.current(PathId("/some/app")).futureValue
+
+    Then("the result is None")
+    readFailure should be(None)
+  }
+
+  class Fixture {
+    lazy val inMemoryStore = new InMemoryStore()
```
aquamatthias: Debatable if this should be run as unit or integration test...

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -15,7 +15,7 @@ class MarathonStore[S <: MarathonState[_, S]](
     store: PersistentStore,
     metrics: Metrics,
     newState: () => S,
-    prefix: String = "app:")(implicit ct: ClassTag[S]) extends EntityStore[S] {
+    prefix: String)(implicit ct: ClassTag[S]) extends EntityStore[S] {
```
aquamatthias: The default really was wrong...

aquamatthias:
Generally is looking good. Made some comments.

src/test/scala/mesosphere/marathon/state/TaskFailureRepositoryTest.scala
kolloch: It's fast. You do not need zookeeper or mesos. So why not always run it?

What do you think?

src/main/scala/mesosphere/marathon/state/TaskFailureRepository.scala
kolloch: We only expose one task failure in our API, so in theory we could just store that.

All our stores are "versioned" so I just implemented it like this. Our current API of the `EntityRepository` only supports "versioned" stores.

But you are right, these are not really versions. 

I don't feel that we should persist the lastTaskFailure for every app version we retain. This would be quite some effort (synchronizing this with the versions of the AppRepo etc) but wouldn't have such great effect. The lastTaskFailure itself contains information about the version of the task.

You are asking a lot of (valuable) questions without providing your opinion. What do you think is best?

These are the options I see:

A) Leave it as is. There is a technically unnecessary overhead for versioning, even though we do not use the different versions.
B) Extend the EntityRepository to support unversioned storage and use that.
C) Use the an `EntityStore` directly.

I think that A) might lead to performance problems in load situations but otherwise OK.

B) is probably needed for a better task storage anyway (we do not want to version our task infos, do we?)

C) is ugly.

kolloch:
I changed it to storing just one version per app.

src/test/scala/mesosphere/marathon/state/TaskFailureRepositoryTest.scala
aquamatthias: I agree it's fast.

src/main/scala/mesosphere/marathon/state/TaskFailureRepository.scala
```diff
@@ -1,17 +1,20 @@
...
+class TaskFailureRepository(
+  protected val store: EntityStore[TaskFailure],
+  protected val maxVersions: Option[Int] = Some(1),
+  protected val metrics: Metrics)
+    extends EntityRepository[TaskFailure] {
 
-  def store(id: PathId, value: TaskFailure): Unit =
-    synchronized { taskFailures(id) = value }
+  def store(id: PathId, value: TaskFailure): Future[TaskFailure] =
+    super.storeWithoutVersion(id.safePath, value)
```
aquamatthias: Just curious: why super?

src/main/scala/mesosphere/marathon/state/EntityRepository.scala
```diff
@@ -84,9 +84,13 @@ trait EntityRepository[T <: MarathonState[_, T]] extends StateMetrics {
 
   protected def storeWithVersion(id: String, version: Timestamp, t: T): Future[T] = {
     for {
-      alias <- this.store.store(id, t)
-      result <- this.store.store(id + ID_DELIMITER + version, t)
+      alias <- storeWithoutVersion(id, t)
+      result <- storeWithoutVersion(id + ID_DELIMITER + version, t)
```
aquamatthias: You use the method storeWithoutVersion to store a version here...

PR #2139, opened 2015-08-31
Saner extended AppInfo

kolloch:
This pull request depends on pull request #2136 .


src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -0,0 +1,123 @@
...
+    marathonSchedulerService: MarathonSchedulerService,
+    taskFailureRepository: TaskFailureRepository) {
+  import AppInfoBaseData.log
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  lazy val runningDeploymentsByAppFuture: Future[Map[PathId, Seq[Identifiable]]] = {
+    log.debug("Retrieving running deployments")
+
+    val allRunningDeploymentsFuture: Future[Seq[DeploymentPlan]] =
+      marathonSchedulerService.listRunningDeployments().map(_.map(_._1))
```
aquamatthias: something like `.map(_.map(_._1))` is hard to grasp. Perhaps listRunningDeployments can give a better result?

kolloch: I agree. I wanted to finish the PR off without changing even more code but hey, I changed it.

src/main/scala/mesosphere/marathon/api/v2/AppInfoEmbedResolver.scala
```diff
@@ -0,0 +1,56 @@
...
+  private[this] val EmbedCounts = "counts"
+
+  /**
+    * Converts embed arguments to our internal representation.
+    *
+    * Accepts the arguments with all prefixes or even no prefix at all
+    * to avoid subtle user errors confusing the two.
+    */
+  def resolve(embed: util.Set[String]): Set[Embed] = {
+    def mapEmbedStrings(prefix: String, withoutPrefix: String): Set[AppInfo.Embed] = withoutPrefix match {
```
aquamatthias: Isn't that a good use case for an extractor in Embed, so you can match directly?

src/main/scala/mesosphere/marathon/api/v2/LabelSelector.scala
```diff
@@ -1,16 +1,17 @@
...
 
+import mesosphere.marathon.core.appinfo.AppSelector
 import mesosphere.marathon.state.AppDefinition
 import org.apache.log4j.Logger
 
 import scala.util.control.NonFatal
 import scala.util.parsing.combinator.RegexParsers
 
-case class LabelSelector(key: String, fn: String => Boolean, value: List[String]) {
+case class LabelSelector(key: String, fn: String => Boolean, value: List[String]) extends AppSelector {
```
aquamatthias: Since the LabelSelector now is an AppSelector - should the class be moved to the core.appinfo(.impl)

src/main/scala/mesosphere/marathon/api/v2/AppInfoEmbedResolver.scala
kolloch: Do you mean an object with `unapply` with extractor? Is it important to you?

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultAppInfoService.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  override def queryForAppId(id: PathId, embed: Set[Embed]): Future[Option[AppInfo]] = {
+    appRepository.currentVersion(id).flatMap {
+      case Some(app) => newBaseData().appInfoFuture(app, embed).map(Some(_))
+      case None      => Future.successful(None)
+    }
+  }
+
+  override def queryAll(selector: AppSelector, embed: Set[Embed]): Future[Seq[AppInfo]] = {
+    appRepository
```
aquamatthias: As long as there is a difference between groupRepo and appRepo, I would use the groupManager to fetch the app.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultAppInfoService.scala
```diff
@@ -0,0 +1,50 @@
...
+import scala.concurrent.Future
+
+private[appinfo] class DefaultAppInfoService(
+    groupManager: GroupManager,
+    appRepository: AppRepository,
+    newBaseData: () => AppInfoBaseData) extends AppInfoService {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  override def queryForAppId(id: PathId, embed: Set[Embed]): Future[Option[AppInfo]] = {
+    appRepository.currentVersion(id).flatMap {
```
aquamatthias: As long as there is a difference between groupRepo and appRepo, I would use the groupManager to fetch the app.

src/main/scala/mesosphere/marathon/api/v2/LabelSelector.scala
kolloch: I am not sure. Currently, it is only referenced in the V2 API. It depends on whether we want to use it in future APIs. It does not depend on implementing AppSelector or not (in my opinion).

It should not go into `impl` because it is used outside of the appinfo package. Maybe it even deserves its own package.

Can we make that change part of another PR if you like this?

src/main/scala/mesosphere/marathon/api/v2/AppInfoEmbedResolver.scala
aquamatthias: "tasks"=> Tasks, "deployments"=>Deployments should be part of embed. The resolve method makes sense here, since it removes the prefix and creates a set...

aquamatthias:
Some minor comments - otherwise Cool! :+1: 
ship that shit!

src/main/scala/mesosphere/marathon/api/v2/LabelSelector.scala
aquamatthias: Sure.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultAppInfoService.scala
kolloch: I changed this to groupRepo.rootGroup because it is more efficient.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultAppInfoService.scala
kolloch: The appRepository.store now happens in groupManager.update. So this should be more efficient.

kolloch:
Changed and squashed as noted above.

kolloch:
Thanks for the superfast review!

PR #2151, opened 2015-09-01
Fixes #1900 - Add a marathon-interface project

### LINKED ISSUE #1900: Add an interface module that can be used from external projects
Create a new sbt compile unit containing interface definitions that can be implemented by external modules.

The resulting artifact should be made publicly available (e.g. downloads.mesosphere.io).

JIRA: MARATHON-227

gkleiman:


project/build.scala
```diff
@@ -13,10 +16,24 @@ import ReleasePlugin._
 import ReleaseStateTransformations._
 
 object MarathonBuild extends Build {
+  lazy val marathonInterface: Project = Project(
+    id = "marathon-interface",
+    base = file("marathon-interface"),
+    settings = baseSettings ++
+      asmSettings ++
```
aquamatthias: The formatting here and in root.settings is different.

aquamatthias:
Generally looks good: does the publishing work in TC?

PR #2154, opened 2015-09-02
Add a way to override the port for tcp/http/https healthchecks

EvanKrall:
This makes it possible to have Marathon healthcheck a different (fixed) port on the host where a task is running, or to healthcheck a different host entirely.

For context, we use uber/hacheck as part of our load balancing layer. This allows operators to remove machines from service before performing maintenance, by running `hadown all`. This will cause all healthchecks that go through hacheck to fail. Unfortunately, Marathon has no idea that a box is marked down, so it will happily deploy tasks to that box, even though they will never be load balanced.

We would like to have Marathon healthcheck http://${host}:3333/spool/all/0/status, which will return 5xx if a box is `hadown`ed. This patch would allow this by setting up a healthcheck that looks like:

```
{
  "protocol": "http",
  "path": "/spool/all/0/status",
  "overridePort": 6666,
  ...
}
```

### DIFF
--- a/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md
+++ b/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md
@@ -82,6 +82,15 @@ Here is an example of an application JSON which includes all fields.
             "maxConsecutiveFailures": 3
         },
         {
+            "protocol": "HTTP",
+            "path": "/machinehealth",
+            "gracePeriodSeconds": 3,
+            "intervalSeconds": 10,
+            "overridePort": 3333,
+            "timeoutSeconds": 10,
+            "maxConsecutiveFailures": 3
+        },
+        {
             "protocol": "TCP",
             "gracePeriodSeconds": 3,
             "intervalSeconds": 5,
--- a/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
+++ b/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
@@ -181,6 +181,17 @@ Transfer-Encoding: chunked
                 "portIndex": 0,
                 "protocol": "HTTP",
                 "timeoutSeconds": 10
+            },
+            {
+                "command": null,
+                "gracePeriodSeconds": 5,
+                "intervalSeconds": 10,
+                "maxConsecutiveFailures": 6,
+                "path": "/machinehealth",
+                "portIndex": 0,
+                "overridePort": 3333,
+                "protocol": "HTTP",
+                "timeoutSeconds": 10
             }
         ],
         "id": "/toggle",
@@ -211,6 +222,14 @@ Transfer-Encoding: chunked
                         "lastFailure": null,
                         "lastSuccess": "2014-09-13T00:25:07.506Z",
                         "taskId": "toggle.802df2ae-3ad4-11e4-a400-56847afe9799"
+                    },
+                    {
+                        "alive": true,
+                        "consecutiveFailures": 0,
+                        "firstSuccess": "2014-09-13T00:20:28.101Z",
+                        "lastFailure": null,
+                        "lastSuccess": "2014-09-13T00:25:07.506Z",
+                        "taskId": "toggle.802df2ae-3ad4-11e4-a400-56847afe9799"
                     }
                 ],
                 "host": "10.141.141.10",
@@ -232,6 +251,14 @@ Transfer-Encoding: chunked
                         "lastFailure": null,
                         "lastSuccess": "2014-09-13T00:25:07.508Z",
                         "taskId": "toggle.7c99814d-3ad4-11e4-a400-56847afe9799"
+                    },
+                    {
+                        "alive": true,
+                        "consecutiveFailures": 0,
+                        "firstSuccess": "2014-09-13T00:20:28.101Z",
+                        "lastFailure": null,
+                        "lastSuccess": "2014-09-13T00:25:07.506Z",
+                        "taskId": "toggle.802df2ae-3ad4-11e4-a400-56847afe9799"
                     }
                 ],
                 "host": "10.141.141.10",
--- a/src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
+++ b/src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
@@ -49,10 +49,10 @@ class HealthCheckWorkerActor extends Actor with ActorLogging {
   }
 
   def doCheck(task: MarathonTask, check: HealthCheck): Future[Option[HealthResult]] =
-    task.getPortsList.asScala.lift(check.portIndex) match {
-      case None =>
-        Future { Some(Unhealthy(task.getId, task.getVersion, "Invalid port index")) }
-
+    getPort(task, check) match {
+      case None => Future {
+        Some(Unhealthy(task.getId, task.getVersion, "Invalid port index and no override port specified"))
+      }
       case Some(port) => check.protocol match {
         case HTTP  => http(task, check, port)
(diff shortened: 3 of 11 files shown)


carroux:
+1 –– This would be a useful option to have

src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
```diff
@@ -73,6 +73,13 @@ class HealthCheckWorkerActor extends Actor with ActorLogging {
       }
     }
 
+  def getPort(task: MarathonTask, check: HealthCheck): Option[Integer] = {
```
kolloch: this method can be `private[this]` or embedded in the `doCheck` method. What do you think?

kolloch:
Hi @EvanKrall, thanks for the pull request!

I think that functionality makes sense. I put @gkleiman in the loop as further filter to prevent potential feature bloat ;)

For the pull request do get accepted, you would need to:

* add tests
* add documentation (REST API, changelog and other places where health checks are mentioned)

gkleiman:
What @EvanKrall describes will no longer be necessary once https://issues.apache.org/jira/browse/MESOS-1474 is completed, but it is completely valid until then and I see no reason for not adding this feature once it's fully tested and documented.

EvanKrall:
Thanks for the feedback. I'll get tests and docs written.

src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
EvanKrall: I had originally inlined this code in doCheck, but the complexity checker complained about that. I can make it private.

kolloch: Yeah, this complexity checker is not very smart. It should treat inline defs as separate methods.

aquamatthias:
@EvanKrall  any updates here?

EvanKrall:
I've rebased against master, added tests, and put overridePort in a few more places in documentation. I think it'd be a bit overkill to put it everywhere `healthChecks` is mentioned in documentation, as I think most people will not use this, so I somewhat arbitrarily chose a few spots.

Please let me know if there are places in the docs that would be more appropriate, or if there are specific tests you want me to write.

aquamatthias:
@gkleiman You already looked into the first submission - can you do the review?

EvanKrall:
Rebased again and regenerated Protos.java to deal with conflicts.

gkleiman:
@EvanKrall thanks a lot! I will take a look tomorrow.

docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md
```diff
@@ -82,6 +82,16 @@ Here is an example of an application JSON which includes all fields.
             "maxConsecutiveFailures": 3
         },
         {
+            "protocol": "HTTP",
+            "path": "/machinehealth",
+            "gracePeriodSeconds": 3,
+            "intervalSeconds": 10,
+            "overridePort": 3333,
+            "portIndex": 0,
```
gkleiman: I think we remove portIndex from here.

Having both `overridePort` and `portIndex` is a bit confusing.

docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
```diff
@@ -181,6 +181,17 @@ Transfer-Encoding: chunked
...
                 "protocol": "HTTP",
                 "timeoutSeconds": 10
+            },
+            {
+                "command": null,
+                "gracePeriodSeconds": 5,
+                "intervalSeconds": 10,
+                "maxConsecutiveFailures": 6,
+                "path": "/machinehealth",
+                "portIndex": 0,
```
gkleiman: I'd remove this (see previous comment).

src/test/scala/mesosphere/marathon/health/HealthCheckTest.scala
```diff
@@ -31,6 +31,29 @@ class HealthCheckTest extends MarathonSpec {
     assert(10 == proto.getGracePeriodSeconds)
     assert(60 == proto.getIntervalSeconds)
     assert(0 == proto.getMaxConsecutiveFailures)
+    assert(!proto.hasOverridePort)
+  }
+
+  test("ToProtoOverridePort") {
```
gkleiman: Can you also add the field in MergeFromProto to make sure that the overridePort is read when loading protocol buffers?

gkleiman:
@EvanKrall thanks again!

This looks good, but there are a few outstanding comments and I'm afraid that you'll have to rebase the branch again.

docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md
EvanKrall: I can remove it here; it'll default to 0 even though it's not used.

docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
EvanKrall: This actually does show, since portIndex is not nullable.

This also shows on "protocol": "COMMAND": https://gist.github.com/EvanKrall/026de225cabe1843762e

src/test/scala/mesosphere/marathon/health/HealthCheckTest.scala
EvanKrall: done

EvanKrall:
I've addressed feedback and rebased.

EvanKrall:
Rebased again to keep up to date.

Anything else I should address?

mauri:
Any particular reason why the option to override the host completely wasn't included? As there has been a lot of improvements in container networking like calico project or docker overlay networking it would be desired to health check the actual host/ip-address of the app.

EvanKrall:
@mauri I thought about doing that, but my particular use case doesn't need it, and I thought it would complicate the diff. If there's a desire for that, I could try to add it, but I don't think I'll have a chance to do that for at least a few days.

mauri:
@EvanKrall No problem, I can add it later

gkleiman:
Thanks @EvanKrall for keeping this PR up-to-date and sorry for not having gotten back to you sooner.

We've got a new use case for this feature - support for per-container IPs.

I think it would be better to rename the new optional field "port" and make "portIndex" optional. This means that we need some validation to make sure that either "port" or "portIndex" are defined, but not both.

We'd like to merge this in the next couple of days. If you have time to work on this, we'd be really happy to collaborate, otherwise we can make the changes ourselves.

src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
```diff
@@ -49,10 +49,10 @@ class HealthCheckWorkerActor extends Actor with ActorLogging {
   }
 
   def doCheck(task: MarathonTask, check: HealthCheck): Future[Option[HealthResult]] =
-    task.getPortsList.asScala.lift(check.portIndex) match {
-      case None =>
-        Future { Some(Unhealthy(task.getId, task.getVersion, "Invalid port index")) }
-
+    getPort(task, check) match {
+      case None => Future {
```
gkleiman: If you want to improve our current code as well, you can use Future.successful and spare a thread.

EvanKrall:
I'm happy to hand off this work -- I have basically zero experience with Scala so it would probably be faster for you to make those changes than me.

IP-per-container does make this important. For my use case (the per-host healthcheck daemon), I may want to add an overrideIP if we start using IP-per-container.

kolloch:
Thanks a lot @EvanKrall. Continuing this in PR #2731.

PR #2156, opened 2015-09-02
Add github issues to the help section

pierlo-upitup:
Hey guys,

I noticed `marathon` doesn't have a CONTRIBUTING.md file and it doesn't really tell our users how to use the issues page. I know it's trivial info, but we can at least use this to spend some words on how to use our `gui` label. Maybe it helps?


gkleiman:
Good idea!

PR #2160, opened 2015-09-02
Modify the marathon the default log directory

ss75710541:


xiaods:
+1

aquamatthias:
Hey @ss75710541 thanks for that PR.
I will not merge that PR, since it will break the existing installations that rely on the current configuration.
You can change the log4j configuration, by passing the -Dlog4j.configuration={path to file} to the start command.

PR #2168, opened 2015-09-02
Add troubleshooting step for secret too short

jgarcia-mesosphere:
When trying to set up framework auth today, I had a persistent issue with starting Marathon because of:

```ERROR Failed to take over leadership: java.io.IOException: Error reading authentication secret from file [file:///tmp/crd] (mesosphere.marathon.MarathonSchedulerService:263)```

After a bit of kicking it around, I added more chars to the secret string. This had the unexpected effect of resolving the issue.


aquamatthias:
Thanks @jgarcia-mesosphere!

PR #2172, opened 2015-09-03
Fix #2171 deployments should be part of the index automatically (back…

### LINKED ISSUE #2171: Application has no deployments object
Reproduced on 0.11.0-SNAPSHOT

If fetching the applications with ```v2/apps/``` (without any embed param) the applications have no deployment object assigned, while they are deploying. Although these deploying applications are listed in ```v2/deployments/```.

That means that the application status is listed as RUNNING in the application list, which is wrong.

deployment

aquamatthias:
…ward compatibility)


gkleiman:
can i haz tests?

gkleiman:
LGTM!

PR #2177, opened 2015-09-03
Fixes #2176 - Make AppMockFacade retries less verbose and

### LINKED ISSUE #2176: Make HTTP Retries in AppMockFacade less verbose and more reliable
We often have the situation in the test code that we send an HTTP query to an app which might or might not be already started up completely.

We retry with high frequency. This results in lots of logging output and led to stackoverflows because our retry logic used recursion without tailrec.

kolloch:
not crash because of StackOverflows.


aquamatthias:
LGTM!

PR #2188, opened 2015-09-04
Fix #1901 - Implement an external plugin loader

### LINKED ISSUE #1901: Create a plugin loader, that is able to load externally packaged plugins
Create a mechanism to load external plugins that implement Marathon's public interfaces.

JIRA: MARATHON-228

gkleiman:
This commit is based on top of gk/fix_1900. I'm creating the PR for review only, but it shouldn't be merged until the branch for v0.11.0 is cut.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/plugin/PluginManager.scala
+++ b/src/main/scala/mesosphere/marathon/core/plugin/PluginManager.scala
@@ -1,0 +1,76 @@
+package mesosphere.marathon.core.plugin
+
+import java.io.File
+import java.net.{ URL, URLClassLoader }
+import java.util.ServiceLoader
+
+import org.slf4j.{ Logger, LoggerFactory }
+
+import scala.collection.JavaConverters._
+import scala.reflect.ClassTag
+
+/**
+  * Instances of this class holds a reference to a plugin of a specific type.
+  * @param classTag the class tag of the specific type.
+  * @param plugins all the instances of this plugin type.
+  * @tparam T the type implemented by the plugins.
+  */
+case class PluginReference[T](classTag: ClassTag[T], plugins: Seq[T])
+
+/**
+  * The plugin manager can load plugins from given urls.
+  * @param urls the urls pointing to plugins.
+  */
+class PluginManager(val urls: Seq[URL]) {
+
+  private[this] val log: Logger = LoggerFactory.getLogger(getClass)
+
+  private[this] var pluginReferences: List[PluginReference[_]] = List.empty[PluginReference[_]]
+
+  val classLoader: URLClassLoader = new URLClassLoader(urls.toArray, this.getClass.getClassLoader)
+
+  /**
+    * Load plugin for a specific type.
+    */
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/plugin/PluginManager.scala
```diff
@@ -0,0 +1,63 @@
+package mesosphere.marathon.plugin
```
aquamatthias: I thought the idea is to move all new code to core (core.plugin) and align to the new structure?

src/main/scala/mesosphere/marathon/plugin/PluginManager.scala
```diff
@@ -0,0 +1,63 @@
...
+import scala.collection.JavaConverters._
+import scala.reflect.ClassTag
+
+/**
+  * Instances of this class holds a reference to a plugin of a specific type.
+  * @param classTag the class tag of the specific type
+  * @param plugins all the instances of this plugin
+  * @tparam T the specific type of this plugin.
+  */
+case class PluginReference[T](classTag: ClassTag[T], plugins: Seq[T])
```
aquamatthias: The Seq here is eager. The iterator returned by the ServiceLoader is lazy. Perhaps we should maintain the lazyness?

src/main/scala/mesosphere/marathon/plugin/PluginModule.scala
```diff
@@ -0,0 +1,30 @@
+package mesosphere.marathon.plugin
+
+import java.net.URL
+
+import com.google.inject.{ Singleton, Provides, AbstractModule }
+import org.rogach.scallop.ScallopConf
+
+trait PluginConfiguration extends ScallopConf {
+  lazy val pluginUris = opt[List[String]]("plugin_uris",
```
aquamatthias: URIs are powerful, but also complicated. My proposal would be to use plugin_dir. That directory is listed and all items are added to the CP.
Example: `/path/to/marathon/plugins`

Perhaps it make sense to use that option multiple times.

aquamatthias:
Please rebase.

gkleiman:
Rebased

src/main/scala/mesosphere/marathon/core/plugin/PluginManager.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  def plugin[T](className: String)(implicit ct: ClassTag[T]): Option[T] = {
+    plugins[T].find(plugin => className.equals(plugin.getClass.getName))
+  }
+}
+
+object PluginManager {
+  def apply(conf: PluginConfiguration): PluginManager = {
+    conf.pluginDir.get.map { pluginDir: String =>
+      val jars: Array[File] = IO.listFiles(new File(pluginDir), ".*jar$".r)
```
aquamatthias: My first approach also was to filter jars. 
But you can have directories (class dirs like target) or config files.
Perhaps it is better to not be clever and filter jars.

aquamatthias:
What about test cases?

PR #2191, opened 2015-09-05
Remove git submodule wreckage

okuryu:
Git Submodules have been removed in #1973, but some stuffs still
seem to remain. This change discard them completely.


gkleiman:
Thanks! :+1:

PR #2199, opened 2015-09-07
Make Scaling Test more robust

gkleiman:


gkleiman:
The goal of this PR was to address the Scaling Test failure in https://teamcity.mesosphere.io/viewLog.html?buildId=43959&buildTypeId=Oss_Marathon_IntegrationScaleTests&tab=buildLog

But after taking a deeper look and taking with @kolloch, it became evident that the test failed because of a bug, so I'm opening an issue and closing this PR.

PR #2209, opened 2015-09-08
Use 3 seconds as default timeout for waiting for Futures

kolloch:
it was 150ms before and there were a lot failures in code using this.

An alternative would be to use test duration scaling but I like this better.

(A lot of the other durations should not be scaled.)


gkleiman:
:+1:

PR #2213, opened 2015-09-08
Save tasks just before launching

kolloch:


src/main/scala/mesosphere/marathon/core/launcher/OfferProcessorConfig.scala
```diff
@@ -9,6 +9,11 @@ trait OfferProcessorConfig extends ScallopConf {
     descr = "Offer matching timeout (ms). Stop trying to match additional tasks for this offer after this time.",
     default = Some(1000))
 
+  lazy val saveTasksToLaunchTimeout = opt[Int]("save_tasks_to_launch_timeout",
```
gkleiman: This new flag has to be added to the docs and the changelog.

From this description alone, I don't really get which tasks Marathon will try to save and what would happen if it times out.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -7,94 +7,205 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher
...
+      dummySource, task, MarathonTestHelper.makeTaskFromTaskInfo(task)))
     val offerProcessor = createProcessor()
 
     val deadline: Timestamp = clock.now() + 1.second
-    when(offerMatcher.matchOffer(deadline, offer)).thenReturn(
-      Future.successful(MatchedTasks(offerId, tasksWithSource))
-    )
-    when(taskLauncher.launchTasks(offerId, tasks)).thenReturn(true)
 
+    And("an cooperative offerMatcher and taskTracker")
```
gkleiman: s/an /a /

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -7,94 +7,205 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher
...
   test("match successful, launch tasks unsuccessful") {
+    Given("an offer")
+    val dummySource = new DummySource
+    val tasksWithSource = tasks.map(task => TaskWithSource(
+      dummySource, task, MarathonTestHelper.makeTaskFromTaskInfo(task)))
+
+    val offerProcessor = createProcessor()
+
+    val deadline: Timestamp = clock.now() + 1.second
+    And("an cooperative offerMatcher and taskTracker")
```
gkleiman: s/an /a /

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -7,94 +7,205 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher
...
+
+    Then("we saw the matchOffer request and the task launch attempt")
+    verify(offerMatcher, times(1)).matchOffer(deadline, offer)
+    verify(taskLauncher, times(1)).launchTasks(offerId, tasks)
+
+    And("all task launches were rejected")
+    assert(dummySource.accepted.isEmpty)
+    assert(dummySource.rejected.toSeq.map(_._1) == tasks)
+
+    And("the tasks where first stored and then expunged again")
```
gkleiman: It sounds like you care about the order of the following calls.

Mockito doesn't care about the order of the calls unless you use an inOrder object.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -7,94 +7,205 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher
...
     val dummySource = new DummySource
-    val tasksWithSource = tasks.map(TaskWithSource(dummySource, _))
+    val tasksWithSource = tasks.map(task => TaskWithSource(
+      dummySource, task, MarathonTestHelper.makeTaskFromTaskInfo(task)))
 
     val offerProcessor = createProcessor()
 
     val deadline: Timestamp = clock.now() + 1.second
-    when(offerMatcher.matchOffer(deadline, offer)).thenReturn(
+    And("an cooperative offerMatcher that takes really long")
```
gkleiman: s/an /a /

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -7,94 +7,205 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher
...
+    When("processing the offer")
     Await.result(offerProcessor.processOffer(offer), 1.second)
 
-    verify(offerMatcher).matchOffer(deadline, offer)
-    verify(taskLauncher).launchTasks(offerId, tasks)
+    Then("we saw the matchOffer request")
+    verify(offerMatcher, times(1)).matchOffer(deadline, offer)
 
+    And("all task launches were rejected")
     assert(dummySource.accepted.isEmpty)
```
gkleiman: What about adding:

And("the processor didn't try to launch the tasks")
verify(taskLauncher, never()).launchTasks(offerId, tasks)

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -7,94 +7,205 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher
...
-    when(offerMatcher.matchOffer(deadline, offer)).thenReturn(
-      Future.successful(MatchedTasks(offerId, Seq.empty))
-    )
+    offerMatcher.matchOffer(deadline, offer) returns Future.successful(MatchedTasks(offerId, Seq.empty))
 
     Await.result(offerProcessor.processOffer(offer), 1.second)
 
-    verify(offerMatcher).matchOffer(deadline, offer)
-    verify(taskLauncher).declineOffer(offerId, refuseMilliseconds = Some(conf.declineOfferDuration()))
+    verify(offerMatcher, times(1)).matchOffer(deadline, offer)
```
gkleiman: times(1) is used by default, any reason for adding it here explicitly?

src/main/scala/mesosphere/marathon/core/launcher/OfferProcessorConfig.scala
kolloch: Since this is rather special, I didn't want to expose this config option at all. That's why it is "hidden".

Maybe it is better to document it. This documentation could be extended as well:

"[...] about to launch. [Added from here] When reaching the timeout, only the tasks that we could already save are also launched. All other task launches are temporarily rejected and retried later."

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
kolloch: It is not added by default in the `Mockito` class that @aquamatthias created and that is used here.

src/main/scala/mesosphere/marathon/core/launcher/OfferProcessorConfig.scala
gkleiman: I like the addition, it makes it much clearer.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
kolloch: Ok

kolloch:
updated

gkleiman:
LGTM, I'll let @aquamatthias chime in before merging.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
aquamatthias: Now uses inOrder.

gkleiman:
Merging, thanks a lot!

PR #2221, opened 2015-09-09
Improve simulation and task confirmation timeout

kolloch:
It is in additional feature but it might be need in the new version because we persist unconfirmed task launches in 0.11.

### DIFF
--- a/mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
+++ b/mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
@@ -4,7 +4,7 @@ import java.util.UUID
 
 import akka.actor.{ Actor, ActorRef, Cancellable, Props }
 import akka.event.LoggingReceive
-import mesosphere.mesos.simulation.DriverActor.{ ReviveOffers, KillTask, LaunchTasks }
+import mesosphere.mesos.simulation.DriverActor.{ ChangeTaskStatus, ReconcileTask, ReviveOffers, KillTask, LaunchTasks }
 import mesosphere.mesos.simulation.SchedulerActor.ResourceOffers
 import org.apache.mesos.Protos._
 import org.apache.mesos.SchedulerDriver
@@ -12,6 +12,7 @@ import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConversions._
 import scala.concurrent.duration._
+import scala.util.Random
 
 object DriverActor {
   case class DeclineOffer(offerId: OfferID)
@@ -38,6 +39,8 @@ object DriverActor {
   case class ReconcileTask(taskStatus: Seq[TaskStatus])
 
   case object ReviveOffers
+
+  private case class ChangeTaskStatus(taskStatus: TaskStatus)
 }
 
 class DriverActor(schedulerProps: Props) extends Actor {
@@ -45,8 +48,23 @@ class DriverActor(schedulerProps: Props) extends Actor {
 
   private[this] val numberOfOffersPerCycle: Int = 10
 
+  // use random seed to get reproducable results
+  private[this] val random = {
+    val seed = System.currentTimeMillis()
+    log.info(s"Random seed for this test run: $seed")
+    new Random(new java.util.Random(seed))
+  }
+
   private[this] var periodicOffers: Option[Cancellable] = None
   private[this] var scheduler: ActorRef = _
+
+  private[this] var tasks: Map[String, TaskStatus] = Map.empty.withDefault { taskId =>
+    TaskStatus.newBuilder()
+      .setSource(TaskStatus.Source.SOURCE_SLAVE)
+      .setTaskId(TaskID.newBuilder().setValue(taskId).build())
+      .setState(TaskState.TASK_LOST)
+      .build()
+  }
 
   //scalastyle:off magic.number
   private[this] def offer: Offer = {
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -131,7 +136,7 @@ trait MarathonConf
   lazy val reconciliationInitialDelay = opt[Long]("reconciliation_initial_delay",
     descr = "This is the length of time, in milliseconds, before Marathon " +
       "begins to periodically perform task reconciliation operations",
-    default = Some(15000L)) // 15 seconds
+    default = Some(10000L)) // 15 seconds
```
gkleiman: Please update (or remove) the comment as well.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -141,7 +146,7 @@ trait MarathonConf
   lazy val scaleAppsInitialDelay = opt[Long]("scale_apps_initial_delay",
     descr = "This is the length of time, in milliseconds, before Marathon " +
       "begins to periodically attempt to scale apps",
-    default = Some(15000L)) // 15 seconds
+    default = Some(20000L)) // 15 seconds
```
gkleiman: Please update (or remove) the comment as well.

gkleiman:
Can you please rebase this on top of master?

kolloch:
Rebased, squashed and documented. Baam! :boom:

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
```diff
@@ -45,9 +48,21 @@ class DriverActor(schedulerProps: Props) extends Actor {
 
   private[this] val numberOfOffersPerCycle: Int = 10
 
+  private val randomSeed: Long = 123
+  // use seed for getting more deterministic behavior
+  private[this] val random = new Random(new java.util.Random(randomSeed))
```
gkleiman: What's the reasoning behind this? Wouldn't be potentially cover more cases in the long run if we don't use a constant seed?

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
```diff
@@ -103,32 +118,94 @@ class DriverActor(schedulerProps: Props) extends Actor {
...
-      tasks.foreach { taskInfo: TaskInfo =>
-        scheduler ! TaskStatus.newBuilder()
-          .setSource(TaskStatus.Source.SOURCE_EXECUTOR)
-          .setTaskId(taskInfo.getTaskId)
-          .setState(TaskState.TASK_STAGING)
-          .build()
+      simulateTaskLaunch(offers, tasks)
+
+    case KillTask(taskId) =>
+      log.debug(s"kill tasksToLaunch $taskId")
```
gkleiman: s/tasksToLaunch/task to kill/ ?

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
```diff
@@ -103,32 +118,94 @@ class DriverActor(schedulerProps: Props) extends Actor {
...
       }
-      tasks.foreach { taskInfo: TaskInfo =>
-        scheduler ! TaskStatus.newBuilder()
+  }
+
+  def simulateTaskLaunch(offers: Seq[OfferID], tasksToLaunch: Seq[TaskInfo]): Unit = {
+    if (random.nextDouble() > 0.001) {
+      log.debug(s"launch tasksToLaunch $offers, $tasksToLaunch")
+      tasksToLaunch.foreach { taskInfo: TaskInfo =>
+        val msg = TaskStatus.newBuilder()
```
gkleiman: The code inside all 3 branches is very similar. Can you extract it into a local function?

src/test/scala/mesosphere/marathon/tasks/TaskTrackerTest.scala
```diff
@@ -493,35 +493,53 @@ class TaskTrackerTest extends MarathonSpec with Matchers with GivenWhenThen {
...
     When("We check which tasks should be killed because they're not yet staged")
-    val stagedTasks = taskTracker.determineOverdueTasks(now)
+    val overdueTasks = taskTracker.determineOverdueTasks(now)
 
     Then("Both the overdueUnstagedTask and the overdueStagedTask tasks are returned")
-    assert(stagedTasks == Seq(overdueStagedTask, overdueUnstagedTask))
+    assert(overdueTasks.map(_.getId).toSet ==
+      Set(overdueStagedTask, overdueUnstagedTask, unconfirmedOverdueTask).map(_.getId))
 
+    assert(overdueTasks.toSet ==
```
gkleiman: doesn't this assert imply the previous assert as well?

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
kolloch: I have a dejavu. Didn't I answer this yet?

Well, we have an integration test that has a timeout. There are some random scenarios where this will fail so I wanted to keep this fixed for now. 

We can change this later or make this configurable to have the tests testing more scenarios. But this is for sure better than what we did before. 

Does that make sense to you?

src/test/scala/mesosphere/marathon/tasks/TaskTrackerTest.scala
kolloch: yes exactly. But the earlier assert is easier to debug if it fails.

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
meichstedt: @gkleiman In order to have more deterministic tests, I'd keep this as suggested by @kolloch.
Is that ok for you?

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
meichstedt: I called it `kill task`, since it's just a KillTask message with the taskId

mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
meichstedt: Did that, and also applied that to the 2 similar blocks in the `override def receive` above.

src/test/scala/mesosphere/marathon/tasks/TaskTrackerTest.scala
meichstedt: For the test's sake, the first assert would suffice, since it's about checking whether the right tasks are picked. The second assert _should_ never fail in this context, because we expect nothing else to change. But _if_ something changes (different version, whatever), we might be happy to have this. From this perspective, the second assert (testing that determineOverdueTasks has no side effects) should go into a different test, but I don't care enough to change this.
@gkleiman ok for you?

meichstedt:
Hey @gkleiman Can you please have another look?

gkleiman:
Looking good, thanks!

PR #2229, opened 2015-09-10
Fix #2228 use mesos version 0.23.0

### LINKED ISSUE #2228: Marathon 0.11 should use mesos-utils 0.23

aquamatthias:


gkleiman:
Thanks!

PR #2233, opened 2015-09-10
Introduce Interfaces for AuthN and AuthZ.

aquamatthias:
Introduce the hooks that make sure, only authenticated and authorized Identities can access the resources.
Introduce test cases that test that new functionality, while maintaining backward compatibility.


marathon-interface/src/main/scala/mesosphere/marathon/interface/auth/Authenticator.scala
```diff
@@ -0,0 +1,39 @@
...
+  * The Identity is passed to the [[Authorizer]].
+  * It should hold all information to make decisions possible for authorization.
+  *
+  * The Resources will make sure, that the [[Authenticator#authenticate]]
+  * will call this method only once per request.
+  *
+  * The authenticate method method is most probably one that will query external identity providers,
+  * and uses a Future as result.
+  */
+trait Authenticator extends Plugin {
```
gkleiman: https://github.com/mesosphere/marathon/pull/2188 doesn't include the Plugin trait. Do you think it is necessary?

marathon-interface/src/main/scala/mesosphere/marathon/interface/auth/Authenticator.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+    * Read data from the http request and derive the calling identity.
+    * @param request the request to read the identity information from.
+    * @return Some(Identity) if the identity can be authenticated from the request or None if not.
+    */
+  def authenticate(request: HttpRequest): Future[Option[Identity]]
+
+  /**
+    * If the identity can not be derived from the request, this method is called.
```
gkleiman: This method will be called if the identity can not be derived from a request.

marathon-interface/src/main/scala/mesosphere/marathon/interface/auth/Authenticator.scala
```diff
@@ -0,0 +1,39 @@
...
+    * This purpose of this method is to customize the response send to the user.
```
gkleiman: s/send/sent/

marathon-interface/src/main/scala/mesosphere/marathon/interface/auth/Identity.scala
```diff
@@ -0,0 +1,14 @@
...
+/**
+  * Base trait that represents an identity.
+  */
+trait Identity {
+
+  /**
+    * The string identifier of that identity.
+    * @return the unique identifier of that user.
+    */
+  def id: String
```
gkleiman: Is this really needed?

I can imagine a few scenarios where no id is needed:

- Mocks in Tests
- Very rudimentary security where there's only one user with access to everything.
- Plugins that rely on class hierarchies instead of an id: AdminIdentity vs. UserIdentity

marathon-interface/src/main/scala/mesosphere/marathon/interface/http/HttpResponse.scala
```diff
@@ -0,0 +1,40 @@
...
+  def status(code: Int)
+
+  /**
+    * The redirect to send.
+    * @param url the url to redirect.
+    */
+  def sendRedirect(url: String)
+
+  /**
+    * Set new cookie with name to a value.
```
gkleiman: s/set/Send/?

marathon-interface/src/main/scala/mesosphere/marathon/interface/http/HttpResponse.scala
```diff
@@ -0,0 +1,40 @@
...
+    * @param name the name of the cookie.
+    * @param value the value of the cookie.
+    */
+  def cookie(name: String, value: String)
```
gkleiman: Do we also want to add an optional maxAge parameter? Security cookies normally expire.

marathon-interface/src/main/scala/mesosphere/marathon/interface/auth/Authenticator.scala
aquamatthias: I don't think the Plugin trait is technically necessary. For me it is a Marker that says: this is a Plugin interface. We do provide a lot of traits here and you don't see provided/required.

src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -0,0 +1,86 @@
...
+                               action: AuthorizedAction[Resource],
+                               resources: Resource*)(fn: Identity => Response): Response = {
+    def isAllowed(id: Identity) = resources.forall(authorizer.isAuthorized(id, action, _))
+    def response(fn: HttpResponse => Unit): Response = {
+      val responseFacade = new ResponseFacade
+      fn(responseFacade)
+      responseFacade.response
+    }
+    val requestFacade = new RequestFacade(request)
+    result(authenticator.authenticate(requestFacade)) match {
```
gkleiman: new line? I also like a new line between functions, but that's just my personal taste.

src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -0,0 +1,86 @@
...
+      case Some(identity) if isAllowed(identity) => fn(identity)
+      case Some(identity)                        => response(authorizer.handleNotAuthorized(identity, requestFacade, _))
+      case None                                  => response(authenticator.handleNotAuthenticated(requestFacade, _))
+    }
+  }
+
+  def isAllowed(pathId: PathId)(implicit identity: Identity): Boolean = {
```
gkleiman: isAllowedToView?

src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -0,0 +1,86 @@
...
+    authorizer.isAuthorized(identity, ViewAppOrGroup, pathId)
+  }
+
+  def selectAuthorized(fn: => AppSelector)(implicit identity: Identity): AppSelector = {
```
gkleiman: I'd name this authorizedAppSelector

src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -0,0 +1,86 @@
...
+    AppSelector.forall(Seq(new AuthorizeSelector(identity), fn))
+  }
+
+  private class AuthorizeSelector(identity: Identity) extends AppSelector {
```
gkleiman: private class ViewableAppsSelector(implicit identity: Identity) extends AppSelector {
    override def matches(app: AppDefinition): Boolean = isAllowedToView(app.id)
  }

marathon-interface/src/main/scala/mesosphere/marathon/interface/auth/Identity.scala
aquamatthias: At the moment this is a guess and should probably be replaced:
- we need to log which identity performs which action (or we do it on the plugin side)
- we will need to pass the user identity down to mesos

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -45,140 +48,161 @@ class AppsResource @Inject() (
...
 
   @POST
   @Timed
-  def create(@Context req: HttpServletRequest, body: Array[Byte],
-             @DefaultValue("false")@QueryParam("force") force: Boolean): Response = {
-
+  def create(body: Array[Byte],
+             @DefaultValue("false")@QueryParam("force") force: Boolean,
+             @Context req: HttpServletRequest, @Context resp: HttpServletResponse): Response = {
     val now = clock.now()
```
gkleiman: I know you didn't change this in your commits, but this val is unused.

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -55,6 +62,15 @@ case class PathId(path: List[String], absolute: Boolean = true) extends Ordered[
 
   def toHostname: String = path.reverse.mkString(".")
 
+  override def includes(definition: interface.PathId): Boolean = {
+    //scalastyle:off return
+    if (path.size < definition.path.size) return false
+    path.zip(definition.path).forall{
```
gkleiman: missing space before {

gkleiman:
This looks VERY good =).

Does any of the test check that the app/group index endpoints return only the apps/groups that the principal is allowed to see?

aquamatthias:
No there are none. If there are no general concerns, I can add tests for those cases.

marathon-interface/src/main/scala/mesosphere/marathon/interface/PathId.scala
```diff
@@ -0,0 +1,23 @@
+package mesosphere.marathon.interface
```
meichstedt: We already talked about that, but in order to keep track:
I really much like that package name – but how about Java interoperability?

aquamatthias:
@gkleiman added auth module, plugin flags and wiring. Rebased. Ready to merge.

src/main/scala/mesosphere/marathon/core/auth/AuthConfig.scala
```diff
@@ -0,0 +1,18 @@
+package mesosphere.marathon.core.auth
+
+import org.rogach.scallop.ScallopConf
+
+trait AuthConfig extends ScallopConf {
+
+  lazy val authorizer = opt[String] (
```
gkleiman: should we mention that AuthAllowEverything will be used by default?

src/main/scala/mesosphere/marathon/core/auth/AuthModule.scala
```diff
@@ -0,0 +1,25 @@
...
+import mesosphere.marathon.core.plugin.PluginManager
+import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer }
+
+import scala.reflect.ClassTag
+
+class AuthModule(pluginManager: PluginManager, authConfig: AuthConfig) {
+
+  private[this] def plugin[T](className: String)(implicit ct: ClassTag[T]): T = {
+    pluginManager.plugin[T](className).getOrElse(
+      throw WrongConfigurationException(s"No ${ct.runtimeClass.getSimpleName} with className $className found!")
```
gkleiman: nit:

s"${ct.runtimeClass.getSimpleName} with className $className not found!" sounds better to me.

janisz:
I wonder if writing plugin interface in Scala is good decision. I think plain old Java interface suits better here because it's easier to implement in any JVM language, has less dependencies, so it's more lightweight and has backward compatibility with new releases of JVM languages which is not common in Scala world. In other world when jar with plugin-interface is build with specific version of Scala and some day you will update Scala in in this project all libs will need to be rebuild even if there were no change in interface itself.

@aquamatthias would you push this jar to some Maven snapshot repo, so I can easily start working on plugin?

aquamatthias:
Hey @janisz. I don't see a problem in defining interfaces in scala. I think the point is, that this interface should be usable from Java as well. And I completely agree with this. We will create an example Plugin in Java to show this is possible.

aquamatthias:
@janisz we will publish the interface jar publicly in the future. For now you can do an `sbt plugin-interface/publishLocal` to have it in the local repository.

janisz:
I'm just worried about binary compatibility of this interface. Will my plugin works after you update Scala version (no changes in code)? If yes then there is no problem with Scala.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/Authenticator.scala
```diff
@@ -0,0 +1,39 @@
...
+  /**
+    * Read data from the http request and derive the calling identity.
+    * @param request the request to read the identity information from.
+    * @return Some(Identity) if the identity can be authenticated from the request or None if not.
+    */
+  def authenticate(request: HttpRequest): Future[Option[Identity]]
+
+  /**
+    * This method will be called if the identity can not be derived from a request.
+    * This purpose of this method is to customize the response sent to the user.
```
janisz: This should be more specific. I'm not sure if it will be called when `Option` is empty or `Future` fails.

gkleiman: I agree the documentation should be more specific.

I think that this method should be called on both cases described by @janisz.

janisz:
@aquamatthias I created simple Shiro based plugin https://github.com/janisz/marathon-shiro-authentication-plugin
How to enable it in marathon?
I used following configuration flags `--plugin_dir /tmp/shiro --plugin_conf /tmp/plugin_conf.json` and `plugin_conf.json`:
```json
{
  "plugins":[
    {
      "plugin":"mesosphere.marathon.plugin.auth.Authenticator",
      "implementation":"pl.allegro.tech.marathon.plugin.auth.ShiroAuthenticator"
    },
    {
      "plugin":"mesosphere.marathon.plugin.auth.Authorizer",
      "implementation":"pl.allegro.tech.marathon.plugin.auth.ShiroAuthorizer"
    }
  ]
}
```
but application throws 
> Plugin not found: PluginDefinition(mesosphere.marathon.plugin.auth.Authorizer,pl.allegro.tech.marathon.plugin.auth.ShiroAuthorizer,None)

plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/Authenticator.scala
aquamatthias: Added a more specific comment. Hope that is more clear?

aquamatthias:
Hey @janisz Cool! Without looking at your code, I suspect that you did not define the ServiceLoader manifest? (https://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html)
I just pushed an example plugin PR here: https://github.com/mesosphere/marathon-example-plugins/pull/1

Note: everything is in flux at that moment. But feedback to the interface is highly appreciated!

janisz:
Cool! I'll take a look at your example

aquamatthias:
@janisz I also added a simple java based AuthN+Z Plugin into that PR for reference.

PR #2241, opened 2015-09-12
allow disabling syslog via /etc/marathon/conf/?no-logger'

rboyer:
For people using upstart/sysvinit and wish to not have the logs go to syslog, they can drop `/etc/marathon/conf/?no-logger` on the system without resorting to editing the distributed upstart/sysvinit scripts to pass `--no-logger`

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -77,7 +77,13 @@ function load_options_and_log {
 
   if [[ "${cmd[@]} $@" == *'--no-logger'* ]]
   then
-    "${cmd[@]}" ${@//--no-logger/}
+    local clean_cmd=()
+    for i in "${cmd[@]}" "$@"; do
+      if [[ "${i}" != "--no-logger" ]]; then
+        clean_cmd+=( "${i}" )
+      fi
+    done
+    "${clean_cmd[@]}"
   else
     logged marathon "${cmd[@]}" "$@"
   fi


kolloch:
Hey @cmaloney, can you please have a look?

cmaloney:
I really don't know bash. @lingmann?

aquamatthias:
Hey @lingmann any objections?

aquamatthias:
@lloesche No answer from @lingmann. Can you have a quick look?

lloesche:
Hi @rboyer thank you for your PR!

I'm wondering, would replacing
```bash
"${cmd[@]}" ${@//--no-logger/}
```
with
```bash
"${cmd[@]//--no-logger/}" ${@//--no-logger/}
```
produce the same result?

rboyer:
No, that results in an extra double-single-quote argument being passed to the script.  There is no way AFAIK to contract an element out of a bash list like this without rebuilding the list.

Example script showing your proposed change:

```bash
#!/bin/bash
set -x
echo "extraneous quotes:" "${@//--no-logger}"
```

Running that results in an extra double-single-quote argument being passed in lieu of the contracted argument.  This is because it shrinks the argument to zero length and then ensures that the correct number of arguments get quote-escaped (and empty strings get escaped as ```''```).

```
$ bash tester.sh --something --no-logger --else
+ echo 'extraneous quotes:' --something '' --else
extraneous quotes: --something  --else
```

This is different than what is done to the ```$@``` array:

```bash
#!/bin/bash
set -x
echo "no quotes, but no shell-escape safety:" ${@//--no-logger}
```

Running that results in the argument being contracted as before, but since we don't quote anything the empty string vanishes.  Unfortunately this isn't awesome because it doesn't ensure shell-escaping safety:

```
+ echo 'no quotes, but no shell-escape safety:' --something --else
no quotes, but no shell-escape safety: --something --else
```

lloesche:
@rboyer that makes sense, thanks for clarifying.

LGTM

PR #2258, opened 2015-09-16
Move to Chaos 0.7.0

aquamatthias:


gkleiman:
LGTM!

Don't we also have to change the target jvm in buld.scala?

PR #2268, opened 2015-09-17
Fixes #2267 Remove absolute path from redirect.

### LINKED ISSUE #2267: Redirect to /ui/ doesn't work if Marathon is proxied
Works fine:
```
http://marathon/proxied/ui/
```

This is redirected by Marathon itself to the wrong destination without the ```proxied```-path:
```
http://marathon/proxied/ -> http://marathon/ui/  (404)
```

aquamatthias:


gkleiman:
Looks like a proxy misconfiguration, not a Marathon issue.

Removing "proxy_redirect off" from the nginx configuration should fix the issue.

PR #2286, opened 2015-09-21
Fixes #2264 - Fix validation of app id in AppDefinition.json

### LINKED ISSUE #2264: Cannot submit job with id containing internal slashes
Tried to submit:

```
{
  "id":"/tools/docker/registry",
  "container":{ "type":"DOCKER", "docker":{ "image":"registry" }},
  "env":{
    "SETTINGS_FLAVOR":"local",
    "STORAGE_PATH":"/docker_storage"
  },
  "cpus":0.5,
  "mem":1024,
  "instances":1
}
```

And got:

```
$ dcos marathon app add registry.json
Error: '/tools/docker/registry' does not match '^/?(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])$'
Path: id
Value: "/tools/docker/registry"
```

kolloch:
allowing nested paths.

This effects the CLI. So please notify @tamarrow or @ssk2 after merging.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/ModelValidation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/ModelValidation.scala
@@ -17,6 +17,15 @@ object ModelValidation {
 
   //scalastyle:off null
 
+  /**
+    * This regular expression is used to validate each path segment of an ID.
+    *
+    * If you change this, please also change "pathType" in AppDefinition.json and
+    * notify the maintainers of the DCOS CLI.
+    */
+  private[this] val ID_PATH_SEGMENT_PATTERN =
+    "^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$".r
+
   // TODO: Re-implement this method on it's own terms
   def checkGroup(
     group: Group,
@@ -227,8 +236,7 @@ object ModelValidation {
   }
 
   def idErrors[T: ClassTag](t: T, base: PathId, id: PathId, path: String): Iterable[ConstraintViolation[T]] = {
-    val p = "^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$".r
-    val valid = id.path.forall(p.pattern.matcher(_).matches())
+    val valid = id.path.forall(ID_PATH_SEGMENT_PATTERN.pattern.matcher(_).matches())
     val errors =
       if (!valid)
         List(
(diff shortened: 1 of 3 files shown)


src/test/scala/mesosphere/marathon/api/v2/json/V2AppDefinitionTest.scala
```diff
@@ -49,11 +49,13 @@ class V2AppDefinitionTest extends MarathonSpec with Matchers {
     shouldViolate(app, "id", idError)
 
     app = app.copy(id = "-dash-disallowed-at-start".toRootPath)
-    validateJsonSchema(app, false)
+    // The JSON schema does allow this. It is not as strict as our internal validation to keep the regular
+    // expression reasonably simple.
```
tamarrow: I think you can use the regex from the cli on group ids. It is more complex though... See [here](https://github.com/mesosphere/dcos-cli/blob/master/cli/dcoscli/data/marathon-group-schema.json#L210)

kolloch: At least, we are using the regular expression somewhere else then. Thank you for the pointer!

BTW, does that mean we have group support in the CLI? Any chance that will be available for universe packages soon?

tamarrow: We have group support in the cli! For Universe maybe in the future, but no plans atm.

src/main/scala/mesosphere/marathon/api/v2/ModelValidation.scala
```diff
@@ -17,6 +17,14 @@ object ModelValidation {
 
   //scalastyle:off null
 
+  /**
+    * This regular expression is used to validate each path segment of an ID.
+    *
+    * If you change this, please also change AppDefinition.json and notify the maintainers of the DCOS CLI.
+    */
+  private[this] val ID_PATH_SEGMENT_PATTERN =
+    "^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$".r
```
aquamatthias: Can we use (almost, there is a difference I know) the same pattern as the schema?

src/test/scala/mesosphere/marathon/api/validation/V2AppDefinitionValidatorTest.scala
```diff
@@ -23,6 +25,109 @@ class V2AppDefinitionValidatorTest extends MarathonSpec {
     validateJsonSchema(app)
   }
 
+  for (
```
aquamatthias: Not a big fan of programmatically created test functions - it is harder to find the problem, if it breaks. But that is personal taste.

kolloch: Ok. I see your point. Without the programmatic creation we have more code and its more redundant but easier to debug a single test. I changed it.

kolloch:
Ok, I did NOT resuse the regexp from the internal validation because that results in a very complex regexp in the schema. The schema is more relaxed than our internal validation which is fine in my opinion.

Issue #1988, opened 2015-08-12
Consolidate logging


kolloch:
Currently, we do not use logging consistently.

Some of our code uses log4j, other parts slf4j, other parts Akka logging.

We should probably

* use `slf4j` with the logback backend
* Configure akka to log via slf4j
* use appropriate adapters (log4j, ...) so that legacy libraries also ultimately log via logback

Furthermore, logging configuration should be appropriately documented.

bobrik:
What I'd love to see is ability to grep for distinct offer processing lines. Right now lines look like this:

```
Sep 11 16:49:38 pewpew docker[25833]: [2015-09-11 16:49:38,358] INFO Offer [20150908-164039-319497226-5050-29034-O6855483]. Insufficient resources for [/pss/hhh/zedsg/shhr/bzz] (need cpus=2.0, mem=49152.0, disk=0.0, ports=(1 dynamic), available in offer:
```

Here offer is on the next line and it's not really convenient it tools like kibana.

kolloch:
Related #1127

kolloch:
Hi @bobrik, 

I am not sure what you are saying exactly and what solution you are suggesting. So feel free to clarify if I misunderstood that.

Generally, since our current logging isn't consistent, aggregating log statements across multiple lines for Kibana is hard. It should be reasonably easy to do once we consolidated the log format. You can do it, e.g. with logstash: http://logstash.net/docs/1.4.1/filters/multiline

I would rather not log all information for one offer on one line because that becomes rather unreadable.

gkleiman:
Hi @kolloch,

I am not sure we can easily migrate to the `logback` backend, because some of our dependencies depend on the sl4j to log4j bridge (slf4j-log4j12):

```
> whatDependsOn org.slf4j slf4j-log4j12 1.7.5
[info] org.slf4j:slf4j-log4j12:1.7.5
[info]   +-org.apache.hadoop:hadoop-auth:2.4.1
[info]   | +-org.apache.hadoop:hadoop-common:2.4.1
[info]   |   +-mesosphere:marathon_2.11:0.12.0-SNAPSHOT [S]
[info]   |
[info]   +-org.apache.hadoop:hadoop-common:2.4.1
[info]   | +-mesosphere:marathon_2.11:0.12.0-SNAPSHOT [S]
[info]   |
[info]   +-org.apache.zookeeper:zookeeper:3.4.6
[info]     +-com.twitter.common.zookeeper:candidate:0.0.76
[...]

PR #2300, opened 2015-09-23
Improves #1988 - Consolidate logging

gkleiman:
Removed direct usage of log4j from all classes, except for DebugModule,
where it is used to set the root logger level.

Added akka configuration to log via slf4j.

Pending: move logging backend to logback


kolloch:
Looks good!

PR #2302, opened 2015-09-24
improve handling of exceptions for /v2/apps

tamarrow:


src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala
```diff
@@ -70,6 +70,8 @@ class MarathonExceptionMapper extends ExceptionMapper[Exception] {
       )
     case e: JsonParseException =>
       Json.obj("message" -> e.getOriginalMessage)
+    case e: JsonMappingException =>
+      Json.obj("message" -> "Please specify data in JSON format")
```
kolloch: You introduced this to deal with parsing messages which referenced to "Source@234234234" or something like that, correct?

Even though the message was bad, it contained some more information since it actually said where the parse error was in the input, e.g. like this:

```
{
    "message": "Unexpected close marker '}': expected ']' (for ARRAY starting at [Source: [B@534b9d7e; line: 1, column: 0])"
}
```

It would be nice if you preserved that info.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -60,9 +60,20 @@ class AppsResource @Inject() (
              @DefaultValue("false")@QueryParam("force") force: Boolean): Response = {
 
     val now = clock.now()
+
+    val bodyJson = Json.parse(body)
+    if (bodyJson.toString() == "[]")
```
kolloch: That seems to be a very special purpose check. You want to check if the user specified an empty array?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -60,9 +60,20 @@ class AppsResource @Inject() (
...
+      throw new BadRequestException("Please specify an AppDefinition to create")
+
     val app =
-      validateApp(Json.parse(body).as[V2AppDefinition].withCanonizedIds().toAppDefinition)
-        .copy(versionInfo = AppDefinition.VersionInfo.OnlyVersion(now))
+      bodyJson.asOpt[V2AppDefinition] match {
```
kolloch: If I am not mistaken, we got `JsResultExceptions` before this change which had special treatment in `MarathonExceptionMapper` and actually indicating detail on the errors.

What is it that you would like to improve specifically?

src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala
tamarrow: This happens when you don't specify required data, and used to return: 
`No content to map due to end-of-input\n at [Source: [B@1457c640; line: 1, column: 1]`

I just found it not useful, but I'll put it back in so we don't lose any information.

kolloch:
Thanks for the patch! 

See comments above. Maybe we need to talk about a error handling strategy. Also make sure that the tests work fine. You can make a quick check with `test` in sbt.

Have a look at https://mesosphere.github.io/marathon/docs/contributing.html

kolloch:
:+1:

Issue #2299, opened 2015-09-23
Writes to EntityRepositories should be visible in following reads


kolloch:
Our code assumes that writes are visible for any read following a write.

But it turns out that Zookeeper writes are not immediately visible for subsequent reads. "Consistency Guarantees - ZooKeeper is a high performance, scalable service. Both reads and write operations are designed to be fast, though reads are faster than writes. The reason for this is that in the case of reads, ZooKeeper can serve older data, which in turn is due to ZooKeeper's consistency guarantees." (https://zookeeper.apache.org/doc/trunk/zookeeperProgrammers.html)

This should probably be handled:

* by serving read requests from memory (this is consistent because we only have one leader)
* using the sync request when a new leader starts up to make sure that it sees all prior writes

We will first implement this for tasks: https://github.com/mesosphere/marathon/issues/462

Symptoms:

* Deployments and Apps are not immediately visible after the PUT creating the deployment (see #1407)
* other races

aquamatthias:
Fixed by #2411

PR #2305, opened 2015-09-24
Make TaskTracker use EntityStore

### LINKED ISSUE #2282: TaskTracker should use EntityStore
Parent epic: #462

As a first step to refactoring the TaskTracker, usage of PersistentStore should be replaced with usage of EntityStore. A generic MarathonStore or equivalent should be injected.

meichstedt:
Fixes #2282. Minimally invasive implementation to use EntityStore in TaskTracker. MarathonTask is generated and final and cannot directly extend MarathonState. To enable usage of the existing EntityStore and MarathonStore implementations, I added the MarathonTaskState wrapper, that is used to wrap a MarathonTask into a representation that can be used to parameterise EntityStore and MarathonStore.

Suggested review order:
1. MarathonTaskState
1. TaskTracker
1. MarathonModule
1. everything else


src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -326,13 +326,26 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
...
+  @Singleton
   def provideMigration(
     store: PersistentStore,
     appRepo: AppRepository,
     groupRepo: GroupRepository,
+    taskTracker: TaskTracker,
     metrics: Metrics,
     config: MarathonConf): Migration = {
-    new Migration(store, appRepo, groupRepo, config, metrics)
+    new Migration(store, appRepo, groupRepo, taskTracker, config, metrics)
```
aquamatthias: Is that necessary?

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -151,6 +154,38 @@ class Migration @Inject() (
   }
 }
 
+object LegacyTaskMigrationHelper {
+
+  private[this] val log = Logger.getLogger(getClass.getName)
+
+  // TODO is this still needed? it was introduced in 4c1c5d994aa303a90d54985efb8490e6cc69dd65 on 08/07/2014 ...
```
aquamatthias: I would say no.
The Migration need to fail, when migrating from such an old version.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -1,45 +1,47 @@
...
 
-import scala.collection.JavaConverters._
 import scala.collection._
 import scala.collection.concurrent.TrieMap
 import scala.collection.immutable.Set
 import scala.concurrent.{ Await, Future }
 
 class TaskTracker @Inject() (
-    store: PersistentStore,
+    store: EntityStore[MarathonTaskState],
```
aquamatthias: Please create a separate TaskRepository instead of using the entity store directly.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -1,45 +1,47 @@
...
     config: MarathonConf,
     val metrics: Metrics) extends StateMetrics {
 
   import mesosphere.marathon.tasks.TaskTracker._
   import mesosphere.util.ThreadPoolContext.context
 
+  import scala.language.implicitConversions
```
aquamatthias: why is that needed?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -1,45 +1,47 @@
...
-  val PREFIX = "task:"
   val ID_DELIMITER = ":"
 
   private[tasks] val apps = TrieMap[PathId, InternalApp]()
 
-  private[tasks] def fetchFromState(id: String): Option[PersistentEntity] = timedRead {
-    Await.result(store.load(id), timeout)
+  private[tasks] def fetchFromState(id: String): Option[MarathonTask] = timedRead {
+    val res = store.fetch(id).flatMap {
+      case Some(taskState) => Future.successful(Some(taskState.toProto))
```
aquamatthias: Both cases have Future.successful together with flatMap -> make a simple map.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -1,45 +1,47 @@
...
+      case _               => Future.successful(None)
+    }
+    Await.result(res, timeout)
   }
 
   private[tasks] def getKey(appId: PathId, taskId: String): String = {
-    PREFIX + appId.safePath + ID_DELIMITER + taskId
+    appId.safePath + ID_DELIMITER + taskId
```
aquamatthias: move that logic to TaskIdUtil

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -315,6 +255,8 @@ class TaskTracker @Inject() (
 
 object TaskTracker {
 
+  val STORE_PREFIX = "task:"
```
aquamatthias: Move to repo.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -1,45 +1,47 @@
...
   private[this] val log = LoggerFactory.getLogger(getClass.getName)
 
-  val PREFIX = "task:"
   val ID_DELIMITER = ":"
 
   private[tasks] val apps = TrieMap[PathId, InternalApp]()
 
-  private[tasks] def fetchFromState(id: String): Option[PersistentEntity] = timedRead {
-    Await.result(store.load(id), timeout)
+  private[tasks] def fetchFromState(id: String): Option[MarathonTask] = timedRead {
```
aquamatthias: Move of Repository

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -230,74 +232,12 @@ class TaskTracker @Inject() (
...
-    sink.write(task.toByteArray)
-    sink.flush()
-  }
-
-  def store(appId: PathId, task: MarathonTask): Future[PersistentEntity] = {
-    val byteStream = new ByteArrayOutputStream()
-    val output = new ObjectOutputStream(byteStream)
-    serialize(task, output)
-    val bytes = byteStream.toByteArray
+  def store(appId: PathId, task: MarathonTask): Future[MarathonTask] = {
```
aquamatthias: move to repository

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -230,74 +232,12 @@ class TaskTracker @Inject() (
     fetchTask(getKey(appId, taskId))
 
   private[tasks] def fetchTask(taskKey: String): Option[MarathonTask] = {
```
aquamatthias: Does this method serve any purpose? Just use the repo directly.

aquamatthias:
Most of the comments suggest, that all the persistence logic should be moved out of the taskTracker. I would suggest to put that logic into a separate Repository. Look into GroupRepository for reference.

src/main/scala/mesosphere/marathon/MarathonModule.scala
meichstedt: Only as long as we want to provide the current migration functionality

src/main/scala/mesosphere/marathon/state/Migration.scala
meichstedt: I'd be fine with that. Should we make that explicit? E.g., Upgrade instructions might state that upgrading 0.7 to 0.10 is not possible.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: Done.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: Leftover, is removed now.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: D'oh! Done.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: Done.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: Done.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: See comment below.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
meichstedt: See comment below.

meichstedt:
The scope of this ticket was only to replace the custom implementation with existing  generalisation. I added a TaskRepo, but left the accessors provided by TaskTracker in place – refactoring that should be part of #2299. Refactoring reads/writes to TaskTracker is more involved and requires state synchronisation (of TaskTracker internal inMemory state and the persistence layer).

src/main/scala/mesosphere/marathon/state/TaskRepository.scala
```diff
@@ -0,0 +1,41 @@
...
+  def tasksKeys(appId: PathId): Future[Iterable[String]] = {
+    allIds().map(_.filter(name => name.startsWith(appId.safePath + ID_DELIMITER)))
+  }
+
+  def store(appId: PathId, task: MarathonTask): Future[MarathonTask] = {
+    val key: String = TaskIdUtil.getKey(appId, task.getId)
+    this.store.store(key, MarathonTaskState(task)).map(_.toProto)
+  }
+
+  // TODO: this is needed because EntitRepo does not allow colons as separators and won't return any keys in allIds
```
aquamatthias: The task name should be `tasks:<appid>.<uuid>` - that should be covered by the default implementation

src/main/scala/mesosphere/marathon/state/Migration.scala
aquamatthias: Can we omit the dependency to the TaskTracker here?

src/main/scala/mesosphere/marathon/tasks/TaskIDUtil.scala
```diff
@@ -7,14 +7,19 @@ import org.apache.mesos.Protos.TaskID
...
+// there should rather be a layer that provides access to tasks by appId and taskId
 class TaskIdUtil {
-  val taskDelimiter = "."
+  val appDelimiter = "."
   val TaskIdRegex = """^(.+)[\._]([^_\.]+)$""".r
   val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
 
+  // TODO: using a colon here breaks EntityRepository.allIds() because that only collects names w/o ":"
+  // changing this does not really remove the problem and would require a migration
+  val idDelimiter = ":"
```
aquamatthias: This should not be needed, once we migrated to the correct naming scheme.

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -116,11 +119,11 @@ class Migration @Inject() (
   }
 
   private def changeTasks(fn: InternalApp => InternalApp): Future[Any] = {
```
aquamatthias: Please add a Migration, which renames:
`tasks:appid:appid.uuid` to `tasks:appid.uuid`

src/main/scala/mesosphere/marathon/state/Migration.scala
meichstedt: Yes, that one is replaced by a taskRepo dependency. I added functionality to prevent migration from old versions – that is currently set to 0.3.0; I guess we have no knowledge of whether a migration would work from < 0.3.0.

aquamatthias: No - I think this is pretty safe to assume :)

src/main/scala/mesosphere/marathon/state/TaskRepository.scala
meichstedt: Changed that. Migration is provided – if this will not go into 0.12 I'd update the names …

src/main/scala/mesosphere/marathon/tasks/TaskIDUtil.scala
meichstedt: Indeed, I got rid of a lot of binary functions. Most of them now only take the taskId, which makes mapping easier in various occasions.

src/main/scala/mesosphere/marathon/state/Migration.scala
meichstedt: Added.

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -244,6 +290,29 @@ class MigrationTo0_11(groupRepository: GroupRepository, appRepository: AppReposi
...
+  private[this] val log = LoggerFactory.getLogger(getClass)
+
+  val entityStore = taskRepository.entityStore
+
+  def migrateTasks(): Future[Unit] = {
+    log.info("Start 0.12 migration")
+
+    def migrateKey(legacyKey: String): Future[Unit] = {
+      entityStore.fetch(legacyKey).flatMap {
+        case Some(taskState) => taskRepository.store(taskState.task).flatMap { _ =>
```
aquamatthias: Can you add a short comment why you are doing this (task:foo:foo:version -> task:foo:verison)

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -54,10 +58,21 @@ class Migration @Inject() (
           case NonFatal(e) => throw new RuntimeException("while migrating storage to 0.11", e)
         }
       } yield ()
+    },
+    StorageVersions(0, 12, 0) -> { () =>
```
aquamatthias: Sorry I was to slow - it is 0.13 already.

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -54,10 +58,21 @@ class Migration @Inject() (
...
+      for {
```
aquamatthias: why this extra for loop?

src/main/scala/mesosphere/marathon/state/TaskRepository.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def tasksKeys(appId: PathId): Future[Iterable[String]] = {
+    allIds().map(_.filter(name => name.startsWith(appId.safePath)))
+  }
+
+  def store(task: MarathonTask): Future[MarathonTask] = {
+    this.store.store(task.getId, MarathonTaskState(task)).map(_.toProto)
+  }
+
+  private[state] def entityStore: EntityStore[MarathonTaskState] = store
```
aquamatthias: Just curious: why not make store package protected?

src/main/scala/mesosphere/marathon/tasks/TaskIDUtil.scala
```diff
@@ -7,29 +7,30 @@ import org.apache.mesos.Protos.TaskID
...
-    appId.safePath + taskDelimiter + uuidGenerator.generate()
-  }
-
   def appId(taskId: TaskID): PathId = appId(taskId.getValue)
 
+  // this should actually be a try or option
   def appId(taskId: String): PathId = {
     taskId match {
       case TaskIdRegex(appId, uuid) => PathId.fromSafePath(appId)
+      case _                        => throw new MatchError(s"taskId $taskId is no valid identifier")
```
aquamatthias: Interesting solution :)

aquamatthias:
Looks good!
Please rename 0.12 -> 0.13. Other than that only minor changes.
If you change this I will merge.

src/main/scala/mesosphere/marathon/state/Migration.scala
meichstedt: np, changed that.

src/main/scala/mesosphere/marathon/state/Migration.scala
meichstedt: Sure, done.

src/main/scala/mesosphere/marathon/state/Migration.scala
meichstedt: good point – c&p from the 0.11 block. changed that in both cases.

src/main/scala/mesosphere/marathon/state/TaskRepository.scala
meichstedt: not sure; I wasn't fond of exposing it at all but maybe thought that this is more explicit. never mind, made store package protected and removed the accessor.

src/main/scala/mesosphere/marathon/tasks/TaskIDUtil.scala
meichstedt: Well, yes. Actually, appId should return an option. But the regex match was at some point introduced without doing that, so MatchErrors would throw without any meaningful message, only something like `scala.MatchError:  (of class java.lang.String)` in case of an empty String. I was too lazy to change the signature to return an Option, but at least invalid taskIds are logged now :)

meichstedt:
Thanks for the review! I addressed your comments, fixed, squashed and pushed.

aquamatthias:
LGTM - waiting for CI

aquamatthias:
:+1:

PR #2306, opened 2015-09-24
Send redirects as relative URL (without Host)

aquamatthias:
sendRedirect does always specify the host.
use a separate method for that.


aquamatthias:
Cherry pick for 0.11

aquamatthias:
Fixes #2267

gkleiman:
:+1: 

But I still think people should fix their proxy configurations instead of relying on this =).

bobrik:
@gkleiman I still don't get how to "fix" proxy configuration (see 1 https://github.com/mesosphere/marathon/pull/1973#issuecomment-142875243).

:+1: for this PR, though.

Issue #1763, opened 2015-07-08
Proxying with "Expect: 100-continue" does not work correctly


sttts:
# Workarounds

* Don't use "Expect: 100-continue" (obviously)
* Do use Java 8 to run Marathon
* Always connect to the Marathon-Leader directly

# Original Text

I deploy a redis cluster. srv005 is marathon leader. Then:
```bash
$ curl -v -f -H "Content-Type: application/json" -X POST http://srv004:5080/v2/groups -d@redis.json
* Hostname was NOT found in DNS cache
*   Trying 10.0.0.4...
* Connected to srv004 (10.0.0.4) port 5080 (#0)
> POST /v2/groups HTTP/1.1
> User-Agent: curl/7.37.1
> Host: srv004:5080
> Accept: */*
> Content-Type: application/json
> Content-Length: 2009
> Expect: 100-continue
> 
< HTTP/1.1 100 Continue
< HTTP/1.1 200 OK
< Content-Length: 0
* Server Jetty(8.y.z-SNAPSHOT) is not blacklisted
< Server: Jetty(8.y.z-SNAPSHOT)
< 
* Connection #0 to host srv004 left intact
```

But, talking to srv005 directly gives http 400:
```bash
$ curl -v -f -H "Content-Type: application/json" -X POST http://srv005:5080/v2/groups -d@redis.json
* Hostname was NOT found in DNS cache
*   Trying 10.0.0.5...
* Connected to srv005 (10.0.0.5) port 5080 (#0)
> POST /v2/groups HTTP/1.1
> User-Agent: curl/7.37.1
> Host: srv005:5080
> Accept: */*
> Content-Type: application/json
> Content-Length: 2009
> Expect: 100-continue
> 
< HTTP/1.1 100 Continue
* The requested URL returned error: 400 Bad Request
* Closing connection 0
curl: (22) The requested URL returned error: 400 Bad Request
```

In the logs of srv004:
```
Jul  8 09:52:00 srv004 marathon[6896]: [2015-07-08 09:52:00,372] INFO Proxying request to GET http://srv005:5080/v2/apps from srv004:5080 (mesosphere.marathon.api.JavaUrlConnectionRequestForwarder$:2
80)
Jul  8 09:52:14 srv004 marathon[6896]: [2015-07-08 09:52:14,708] WARN /v2/groups (org.eclipse.jetty.servlet.ServletHandler:563)
[...]

PR #2310, opened 2015-09-25
Adding the use of JAVA_HOME within marathon-framework.

sledigabel:
This would allow to start marathon using the non-default jdk.
Particularly useful to work around issue https://github.com/mesosphere/marathon/issues/1763

Change-Id: I101c9a17fa38e6b88ca562d75bb02b2c1c557279

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -96,7 +96,13 @@ function run_jar {
     vm_opts+=( ${j_opt} )
   done
   # TODO: Set main class in pom.xml and use -jar
-  exec java "${vm_opts[@]}" -cp "$marathon_jar" mesosphere.marathon.Main "$@"
+  if [ -n "$JAVA_HOME" ]
+    then
+      JAVA_BIN="$JAVA_HOME/bin/java" 
+    else
+      JAVA_BIN="java"
+  fi
+  exec $JAVA_BIN "${vm_opts[@]}" -cp "$marathon_jar" mesosphere.marathon.Main "$@"
 }
 
 function logged {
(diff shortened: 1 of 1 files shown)


pegerto:
+1:  Really useful for shared environments

gkleiman:
I tested this on OS X with `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin14) ` and I get the following error:

`target/marathon-runnable.jar: line 99: JAVA_HOME: unbound variable`

http://stackoverflow.com/questions/11362250/in-bash-how-do-i-test-if-a-variable-is-defined-in-u-mode might be useful.

sledigabel:
@gkleiman Sorry about this and thanks for this feedback, I thought this case should be covered by the [ -n "$JAVA_HOME" ]. This is usually how I test if a variable is set or not, seems to be fine for me on OSX:

$ set | grep JAVA
$ #so basically JAVA_HOME is not set
$ if [ -n "$JAVA_HOME" ]; then echo "set"; else echo "not set"; fi
not set

I'll update the PR to take into account the unbound variable error you had.

gkleiman:
@sledigabel: Thank you for the PR!

A bit more info - I just tested on Ubuntu 15.04 - GNU bash, version 4.3.30(1)-release (x86_64-pc-linux-gnu) and I got the same error.

sledigabel:
Updated to fix the mentioned error.
Thanks a lot for your assistance.

sledigabel:
Mmmh the TeamCity tests seem to have failed i'm not sure why.

gkleiman:
This works now, thanks a lot!

Could you please squash your changes to a single commit as described [here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html).

aquamatthias:
@sledigabel Thanks!

PR #2316, opened 2015-09-28
ResourceMatcher: reduce loglevel for unmet constraints

felixb:
It now uses the same log level as for offers with unmet resource requirements.
It was spitting out a lot of warnings although it was running totally fine.


aquamatthias:
@felixb I agree it should have the same logging level. Thanks!

PR #2327, opened 2015-09-29
Fixes #2249 - Added killOldTasksDelaySeconds to UpgradeStrategy

### LINKED ISSUE #2249: Update delay before killing old tasks
My team would like to have more control over the speed of app updates. I would like to add support for a delay between when a new task is healthy and an old task is taken down. Something like Aurora's watch_secs:
http://aurora.apache.org/documentation/latest/configuration-tutorial/#defining-job-objects

This is somewhat similar to https://github.com/mesosphere/marathon/issues/1504, but a different use case.

Would the Marathon team be open to a PR adding this feature? If so, would TaskReplaceActor be the right place to add it?

pgkelley4:
Added the killOldTasksDelaySeconds option to UpgradeStrategy.
The amount of time in seconds to wait before killing an old
task when a new task becomes healthy.

This only effects updates. It does not delay immediate killing
of tasks to meet the minimumHealthCapacity at the beginning of
an update.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,8 @@
 
 ### Fixed Issues
 
-- #2294 - Make boolean command line flags use Scallop's 'toggle'
+- \#2294 - Make boolean command line flags use Scallop's 'toggle'
+- \#2249 - Update delay before killing old tasks
 
 ## Changes from 0.10.0 to 0.11.0
 
--- a/docs/docs/rest-api.md
+++ b/docs/docs/rest-api.md
@@ -1530,7 +1530,8 @@ Transfer-Encoding: chunked
                 "maxLaunchDelaySeconds" : 3600,
                 "upgradeStrategy" : {
                    "minimumHealthCapacity" : 1,
-                   "maximumOverCapacity" : 1
+                   "maximumOverCapacity" : 1,
+                   "killOldTasksDelaySeconds" : 0
                 },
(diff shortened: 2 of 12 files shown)


pgkelley4:
Went ahead and implemented this for https://github.com/mesosphere/marathon/issues/2249. Open to any changes or suggestions.

One thing I was considering was whether or not a task failing during this killOldTaskDelay period should cancel the kill. Right now I have it killing the old task no matter what after the delay.

aquamatthias:
@kolloch can you look into this one? Thanks!

pgkelley4:
This is blocked while https://github.com/mesosphere/marathon/issues/2688 is fixed.

kolloch:
See https://github.com/mesosphere/marathon/pull/2689

felixb:
Is there any way to force marathon to keep old instances for a given time with #2689?

kolloch:
Not that I know of.

I am not sure if Mesos finally caught up with this but generally, processes should get a SIGTERM first and then, only after not responding, a SIGKILL. In theory, processes should be able to delay their death by this but there is hard-coded time-limit in Mesos: https://issues.apache.org/jira/browse/MESOS-1571

If your concern is in the layer above (e.g. load balancing), I think you have to do blue-green deployments by having multiple applications to model the groups.

PR #2328, opened 2015-09-30
Variable in /etc/default/marathon should be exported to take effect

surdy:
Setting MARATHON_* in /etc/default/marathon has no effect as the variables defined here do not make it to the environment and hence cannot be used for setting command line options using `env | grep MARATHON`


aquamatthias:
Hey @surdy Thanks for this!

PR #2340, opened 2015-10-02
Define also slf4j as provided plugin dependency.

aquamatthias:


aquamatthias:
Should be cherry picked into 0.12.

PR #2351, opened 2015-10-05
Genesis of RAML based Rest API documentation.

aquamatthias:
- [x] Add intro raml documentation 
- [x] Add a link to the UI
- [x] Question: there is a raml2html tool which we could use to create one static html file for the docs folder. Do you think this is better?
- [x] Remove the "try it out" button from the static documentation

### DIFF
--- a/docs/docs/rest-api/public/api/api.raml
+++ b/docs/docs/rest-api/public/api/api.raml
@@ -1,0 +1,39 @@
+#%RAML 0.8
+
+title: Marathon REST
+version: 2.0
+mediaType: application/json
+baseUri: /
+schemas:
+  - AppDefinition: !include v2/schema/AppDefinition.json
+  - Group: !include v2/schema/Group.json
+
+traits: !include traits.raml
+
+
+#documentation:
+# - title: Using the Marathon REST API
+#   content: Make me real!
+
+# considered ofthen used endpoints in alphabetical order
+/v2/apps: !include v2/apps.raml
+/v2/deployments: !include v2/deployments.raml
+/v2/groups: !include v2/groups.raml
+/v2/groups/{group_id}: !include v2/groups.raml
+/v2/tasks: !include v2/tasks.raml
+
+# considered less used endpoints in alphabetical order
+/v2/artifacts: !include v2/artifacts.raml
+/v2/events: !include v2/events.raml
+/v2/eventSubscriptions: !include v2/eventSubscriptions.raml
+/v2/info: !include v2/info.raml
+/v2/leader: !include v2/leader.raml
+/v2/queue: !include v2/queue.raml
+
+# no real API endpoints
+/ping: !include general/ping.raml
+/metrics: !include general/metrics.raml
+
+
+
+
--- a/docs/docs/rest-api/public/api/traits.raml
+++ b/docs/docs/rest-api/public/api/traits.raml
@@ -1,0 +1,52 @@
+
+- secured:
+    responses:
+      401:
+        description: Unauthorized. Authentication is enabled and you did not provide enough or wrong information to authenticate that request.
+        body:
+          application/json:
+            example: |
+              { "message": "Invalid username or password." }
+      403:
+        description: Forbidden. Authorization is granted but the identity provided does not have sufficient access rights to do that action.
+        body:
+          application/json:
+            example: |
+              { "message": "Not Authorized to perform this action!" }
+
+- deployable:
+    description:
+      Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+      You can query the deployments endoint to see the status of the deployment.
+    queryParameters:
+      force:
+        required: false
+        description:
+          Only one deployment can be applied to one application at the same time.
+          If the existing deployment should be canceled by this change, you can set force=true.
+
+          Caution&#58; setting force=true will cancel the current deployment. This paramter should be used only, if the current deployment is unsuccessful!
+        type: boolean
+        default: false
+    responses:
+      200:
+        description:
+          A deployment is started which has a unique deployment identifier.
+          The related deployment can be fetched from the /v2/deployments endpoint.
+          If the deployement is gone from the list of deployments, than this means it is finished.
+          As long as the deployment runs, the effect of that change operation is visible only partially.
+        body:
+          application/json:
+            example: !include v2/examples/deployments_result.json
+      409:
+        description:
+          There is an already running deployment, that affects this application.
+          To override this deployment, use the force=true flag
+        body:
+          application/json:
+            example: |
+              {
+               "message":"App is locked by one or more deployments. Override with the option '?force=true'. View details at '/v2/deployments/<DEPLOYMENT_ID>'.",
+               "deployments":[{"id":"97c136bf-5a28-4821-9d94-480d9fbb01c8"}]
+              }
+
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -1,0 +1,416 @@
+
+###
+# Get a list of applications
+#
+get:
+  description:
+    Get the list of running applications. Several filters can be applied via the following query parameters.
+  is: [ secured ]
+  queryParameters:
+    cmd:
+      required: false
+      description: Filter the result to only return apps whose `cmd` field contains the given value
+      example: java
+    id:
+      required: false
+      description: Filter the result to only return apps whose `id` is or contains the given value
+      example: /us-east/database/memsql
+    label:
+      required: false
+      description:
+        A label selector query contains one or more label selectors, which are comma separated.
+        Marathon supports three types of selectors existence-based, equality-based and set-based.
+        In the case of multiple selectors, all must be satisfied so comma separator acts as an AND logical operator.
+        Labels and values must consist of alphanumeric characters plus `-` `_` and `.` `-A-Za-z0-9_.`.
+        Any other character is possible, but must be escaped with a backslash character.
+
+        * <code>Existence based Selector Query</code> Matches the existence of a label <br/>
+          Example&#58; my_label,environment
+
+        * <code>Equality based Selector Query</code> Matches existence of labels and the (non) equality of the value.<br/>
+          Example&#58; environment==production, tier!=frontend
+
+        * <code>Set based Selector Query</code> Matches existence of labels and the (non) existence of the value in a given set<br/>
+          Example&#58; environment in (stage,production), tier notin (frontend, service)
+
+      example: my_label, environment==production, tier!=frontend\ tier, deployed in (us, eu), deployed notin (aa, bb)
+    embed:
+      required: false
+      description:
+        Embeds nested resources that match the supplied path.
+        You can specify this parameter multiple times with different values.
+
+        - <code>apps.tasks</code> embed all tasks of each application<br/>
+        Note&#58; if this embed is definded, it automatically sets <code>apps.deployments</code> but this will change in a future release.
+        Please define all embeds explicitly.
+
+        - <code>apps.counts</code> embed all task counts (tasksStaged, tasksRunning, tasksHealthy, tasksUnhealthy) <br/>
+        Note&#58; currently embedded by default but this will change in a future release.
+        Please define all embeds explicitly.
+
+        - <code>apps.deployments</code> embed all deployment identifier, if the related app currently is in deployment.
+
+        - <code>apps.lastTaskFailure</code> embeds the lastTaskFailure for the application if there is one.
+
+        - <code>apps.failures</code> Shorthand for apps.lastTaskFailure, apps.tasks, apps.counts and apps.deployments.<br/>
+        Note&#58; deprecated and will be removed in future versions
+        Please define all embeds explicitly.
+
+        - <code>apps.taskStats</code> exposes task statatistics in the JSON.
+
+      enum: [ apps.tasks, apps.count, apps.deployments, apps.lastTaskFailure, apps.failures, apps.taskStats ]
+      repeat: true
+      example: apps.tasks
+  responses:
+    200:
+      description: "The list of applications that match the defined filters"
+      body:
+        application/json:
+          example: !include examples/app_list.json
+
+###
+# update a list of applications
+#
+put:
+  description:
+    Change multiple applications either by upgrading existing ones or creating new ones.
+    If there is an update to an already running application, the application gets upgraded.
+    All instances of this application get replaced by the new version.
+    The order of dependencies will be applied correctly.
+    The upgradeStrategy defines the behaviour of the upgrade.
+
+    If the id of the application is not known, the application gets started.
+    The order of dependencies will be applied correctly.
+    It is possible to mix upgrades and installs.
+
+    If you have more complex scenarios with upgrades, use the groups endpoint.
+
+    Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+     You can query the deployments endoint to see the status of the deployment.
+  is: [ secured, deployable ]
+  body:
+    application/json:
+      example: !include examples/apps_create.json
+  responses:
+    409:
+      description: There is an already deployed application with this name
+      body:
+        application/json:
+          example: |
+            {"message":"An app with id [/existing_app] already exists."}
+    400:
+      description: The application definition provided in the body is not valid.
+      body:
+        application/json:
+          example: |
+            {"message":"Invalid JSON","details":[{"path":"/id","errors":["error.expected.jsstring"]}]}
+    422:
+      description: The entity send can not be preocessed, since there are validation errors
+      body:
+        application/json:
+          example: |
+            {
+              "message": "Bean is not valid",
+              "errors": [
+                {
+                  "attribute": "upgradeStrategy.minimumHealthCapacity",
+                  "error": "is greater than 1"
+                }
+              ]
+            }
+
+
+
+
+###
+# Create application
+#
+post:
+  description:
+    Create and start a new application.
+
+    Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+     You can query the deployments endoint to see the status of the deployment.
+  is: [ secured ]
+  body:
+    application/json:
+      example: !include examples/app.json
+      schema: AppDefinition
+  responses:
+    201:
+      description: The application has been created and a deployment is started.
+      body:
+        application/json:
+          example: !include examples/app.json
+    409:
+      description: There is an already deployed application with this name
+      body:
+        application/json:
+          example: |
+            {"message":"An app with id [/existing_app] already exists."}
+    400:
+      description: The application definition provided in the body is not valid.
+      body:
+        application/json:
+          example: |
+            {"message":"Invalid JSON","details":[{"path":"/id","errors":["error.expected.jsstring"]}]}
+    422:
+      description: The entity send can not be preocessed, since there are validation errors
+      body:
+        application/json:
+          example: |
+            {
+              "message": "Bean is not valid",
+              "errors": [
+                {
+                  "attribute": "upgradeStrategy.minimumHealthCapacity",
+                  "error": "is greater than 1"
+                }
+              ]
+            }
+
+
+/{app_id}:
+
+  ###
+  # Get a specific app
+  #
+  get:
+    description:
+      Get the application with id `app_id`.
+      The response includes some status information besides the current configuration of the app.
+      You can specify optional embed arguments, to get more embedded information.
+    is: [ secured ]
+    queryParameters:
+      embed:
+        description:
+          Embeds nested resources that match the supplied path.
+          You can specify this parameter multiple times with different values. <br/>
+
+          - <code>app.tasks</code>. embed tasks
+          Note&#58; if this embed is definded, it automatically sets <code>apps.deployments</code> but this will change in a future release.
+          Please define all embeds explicitly.
+
+          - <code>app.counts</code>. embed all task counts (tasksStaged, tasksRunning, tasksHealthy, tasksUnhealthy) <br/>
+          Note&#58; currently embedded by default but this will change in a future release.
+          Please define all embeds explicitly.
+
+          - <code>app.deployments</code>. embed all deployment identifier, if the related app currently is in deployment.
+
+          - <code>app.lastTaskFailure</code> embeds the lastTaskFailure for the application if there is one.
+
+          - <code>app.failures</code> Shorthand for apps.lastTaskFailure, apps.tasks, apps.counts and apps.deployments.<br/>
+          Note&#58; deprecated and will be removed in future versions
+          Please define all embeds explicitly.
+
+          - <code>app.taskStats</code> exposes task statatistics in the JSON.
+
+        enum: [ app.tasks, app.count, app.deployments, app.lastTaskFailure, app.failures, app.taskStats ]
+        repeat: true
+        example: embed=app.deployments&embed=app.lastTaskFailure
+    responses:
+      200:
+        body:
+          application/json:
+            example: !include examples/app_list.json
+      404:
+        description: No task found with this `app_id`.
+        body:
+          application/json:
+            example: |
+              { "message": "App '/not_existent' does not exist" }
+
+  ###
+  # Update a specific app
+  #
+  put:
+    description:
+      Replaces parameters of a running application. If no application with the given id
+      exists, it will be created. If there is an application with this id, all running instances get
+      upgraded to the new definition.
+
+
+      Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+       You can query the deployments endoint to see the status of the deployment.
+    is: [ secured, deployable ]
+    body:
+      application/json:
+        example: !include examples/app.json
+        schema: AppDefinition
+    responses:
+      404:
+        description: No task found with this `app_id`.
+        body:
+          application/json:
+            example: |
+              { "message": "App '/not_existent' does not exist" }
+      400:
+        description: The application definition provided in the body is not valid.
+        body:
+          application/json:
+            example: |
+              {"message":"Invalid JSON","details":[{"path":"/id","errors":["error.expected.jsstring"]}]}
+      422:
+        description: The entity send can not be preocessed, since there are validation errors
+        body:
+          application/json:
+            example: |
+              {
+                "message": "Bean is not valid",
+                "errors": [
+                  {
+                    "attribute": "upgradeStrategy.minimumHealthCapacity",
+                    "error": "is greater than 1"
+                  }
+                ]
+              }
+
+
+
+  ###
+  # Delete a specific app
+  #
+  delete:
+    description: Destroy an application. All data about that application will be deleted.
+
+      Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+      You can query the deployments endoint to see the status of the deployment.
+    is: [ secured, deployable ]
+    responses:
+      404:
+        description: No app with this id known.
+        body:
+          application/json:
+            example: |
+              {"message":"App '/not-existing' does not exist"}
+
+  /restart:
+    ###
+    # Restart this application
+    #
+    post:
+      description: Restart all tasks of this application.
+      is: [ secured, deployable ]
+      responses:
+        404:
+          description: No task found with this `app_id`.
+          body:
+            application/json:
+              example: |
+                { "message": "App '/not_existent' does not exist" }
+
+  /tasks:
+    ###
+    # Get all tasks of an application
+    #
+    get:
+      description: List all running tasks for application `app_id`.
+      is: [ secured ]
+      responses:
+        200:
+          description: The list of running tasks for application `app_id`.
+          body:
+            application/json:
+              example: !include examples/app_tasks.json
+            text/plain:
+              example: |
+                minecraft_survival-world	10013	srv7.hw.ca1.mesosphere.com:31756
+        404:
+          description: No task found with this `app_id`.
+          body:
+            application/json:
+              example: |
+                { "message": "App '/not_existent' does not exist" }
+
+    ###
+    # Kill tasks of an application
+    #
+    delete:
+      description: Kill tasks that belong to the application `app_id`
+      is: [ secured, deployable ]
+      queryParameters:
+        host:
+          description: all tasks of that application on the supplied slave are killed
+        scale:
+          type: boolean
+          description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks.
+          default: false
+      responses:
+        200:
+          description: If scale=false, all tasks that were killed are returned.
+            If scale=true, than a deployment is triggered and the deployment is returned.
+          body:
+            application/json:
+              example: !include examples/app_tasks.json
+        404:
+          description: No task found with this `app_id`.
+          body:
+            application/json:
+              example: |
+                { "message": "App '/not_existent' does not exist" }
+
+    /{task_id}:
+      ###
+      # Create application
+      #
+      delete:
+        description: Kill the task with ID `task_id` that belongs to the application `app_id`.
+        is: [ secured, deployable ]
+        queryParameters:
+          scale:
+            type: boolean
+            description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks.
+            default: false
+        responses:
+          200:
+            description: If scale=false, the task that was killed is returned.
+              If scale=true, than a deployment is triggered and the deployment is returned.
+            body:
+              application/json:
+                example: !include examples/app_task.json
+          404:
+            description: No task found with this task_id.
+            body:
+              application/json:
+                example: |
+                  {"message":"Task 'not-existing' does not exist"}
+
+  ###
+  # List versions
+  #
+  /versions:
+    get:
+      description: List the versions of the application with id `app_id`
+      is: [ secured ]
+      responses:
+        200:
+          description: The list of versions of the application
+          body:
+            application/json:
+              example: |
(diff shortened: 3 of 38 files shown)


docs/docs/rest-api/public/api/api.raml
```diff
@@ -0,0 +1,29 @@
+#%RAML 0.8
+
+title: Marathon REST API
+version: 2.0
```
meichstedt: This value is not rendered in the output html. The entries are rather self-explanatory since they include the `/v2/` path element, but I think this should be used to render `Marathon REST API 2.0`

docs/docs/rest-api/public/api/api.raml
```diff
@@ -0,0 +1,29 @@
...
+baseUri: /
+schemas:
+  - AppDefinition: !include v2/schema/AppDefinition.json
+  - Group: !include v2/schema/Group.json
+
+traits: !include traits.raml
+
+documentation:
+ - title: Using the Marathon REST API
+   content: Make me real!
```
meichstedt: I need to click on the title in order to open a layover that renders the content. Feels like there should be text rendered directly underneath `Documentation` – expanding the `Using the Marathon REST API` makes no sense, and I don't think we need too much explanation for the documentation header.

docs/docs/rest-api/public/api/traits.raml
```diff
@@ -0,0 +1,15 @@
+
+- secured:
+    responses:
+      401:
+        description: Authentication is enabled and you did not provide enough or wrong information to authenticate that request.
```
meichstedt: I'd place * Unauthorized* in front, so one does not have to read the whole sentence ;)

docs/docs/rest-api/public/api/traits.raml
```diff
@@ -0,0 +1,15 @@
...
+        body:
+          application/json:
+            example: |
+              { "message": "description for authentication failed" }
+      403:
+        description: Authorization is granted but the identity provided does not have sufficient access rights to do that action.
```
meichstedt: As above: **Forbidden** in front of the remaining description

docs/docs/rest-api/public/api/traits.raml
```diff
@@ -0,0 +1,15 @@
...
+        body:
+          application/json:
+            example: |
+              { "message": "description for autorization failed" }
```
meichstedt: Are the messages placeholders? Should they contain the actual message with which we respond, or do we respond with different messages? If we do not document the exact message (I guess we won't), I'd put the message in `<>` braces.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
+
+###
+# Get a list of applications
+#
+get:
+  description: Get the list of running applications. Several filters can be applied to filter the result list.
```
meichstedt: ... via the following query parameters.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+  is: [ secured ]
+  queryParameters:
+    cmd:
+      description: the command of this application
```
meichstedt: "Filter the result to only return apps whose `cmd` field contains the given `cmd`, Default: `""`"

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+      example: java
+    id:
+      description: the id or part of the id of the application
```
meichstedt: "Filter the result to only return apps whose `id` is or contains the given value"

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+        Labels and values must consist of alphanumeric characters plus `-` `_` and `.` `-A-Za-z0-9_.`.
+        Any other character is possible, but must be escaped with a backslash character.
+
+        * <code>Existence based Selector Query</code> Matches the existence of a label
+
+        * <code>Equality based Selector Query</code> Matches existence of labels and the (non) equality of the value.
+
+        * <code>Set based Selector Query</code> Matches existence of labels and the (non) existence of the value in a given set
+
+      example: environment==production, tier!=frontend\ tier, deployed in (us, eu), deployed notin (aa, bb)
```
meichstedt: An example for an existence based selector is missing; only equality and set based are given. As far as I understand this, setting label to `environment` would match apps that have an `environment` set – no matter to which value.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
+
+###
+# Get a list of applications
+#
+get:
+  description: Get the list of running applications. Several filters can be applied to filter the result list.
+  is: [ secured ]
```
meichstedt: As far as I can see, this leads to the output having a `Security Schemes` section with one `Anonymous` entry – this might be a little irritating, as it looks like _the anonymous security scheme is applied_ …

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+        * <code>Set based Selector Query</code> Matches existence of labels and the (non) existence of the value in a given set
+
+      example: environment==production, tier!=frontend\ tier, deployed in (us, eu), deployed notin (aa, bb)
+    embed:
+      description:
+        Embeds nested resources that match the supplied path.
+        You can specify this parameter multiple times with different values.
+
+        - <code>apps.tasks</code> Apps tasks are not embedded in the response
+        by default. This currently implies apps.deployments currently but that
```
meichstedt: drop one of the two _currently_'s – I'd hunt down the second one.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+        * <code>Existence based Selector Query</code> Matches the existence of a label
+
+        * <code>Equality based Selector Query</code> Matches existence of labels and the (non) equality of the value.
+
+        * <code>Set based Selector Query</code> Matches existence of labels and the (non) existence of the value in a given set
+
+      example: environment==production, tier!=frontend\ tier, deployed in (us, eu), deployed notin (aa, bb)
+    embed:
+      description:
+        Embeds nested resources that match the supplied path.
```
meichstedt: I figure this is copied from the former docs, but in general, I think it'd be easier to read if all flags followed an imperative (_eek_) style along with italic annotations, like 
`"apps.counts"`: embed task counts (tasksStaged, tasksRunning, tasksHealthy, tasksUnhealthy). _Note: these are currently embedded by default but this will change in a future release._

docs/docs/rest-api/public/api/v2/apps.raml
meichstedt: Should we mention that the call is/might get expensive? Is there a raml-way of adding a note/mark other than mis-using traits for that?

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+            }
+
+
+
+
+###
+# Create application
+#
+post:
+  description: Create and start a new application.
```
meichstedt: In the old API docs, we have documented most of the body-JSON-properties as well – this is and supposedly cannot be integrated into the raml docs because raml does not support documentation ob body fields. I don't think this is a problem, it's even better than having properties described in the schema _and_ in the docs. We would need to make sure that all properties are correctly described in the JSON schema, though.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+      description: There is an already deployed application with this name
+      body:
+        application/json:
+          example: |
+            {"message":"An app with id [/existing_app] already exists."}
+    400:
+      description: The application definition provided in the body is not valid.
+      body:
+        application/json:
+          example: |
```
meichstedt: Maybe the example should be short and only show a basic example – and the schema open by default? If the example is long, the schema-button is a long way down the page.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+                ]
+              }
+
+
+
+  ###
+  # Delete a specific app
+  #
+  delete:
+    description: Destroy an application. All data about that application will be deleted.
```
meichstedt: Shouldn't we document/explain what exactly happens, and that there might still be running tasks when the response is returned?

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+    is: [ secured ]
+    responses:
+      204:
+        description: The response will contain an empty body.
```
meichstedt: Actually that's a 2012 and it will contain a deployment result.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+                  }
+                ]
+              }
+
+
+
+  ###
+  # Delete a specific app
+  #
+  delete:
```
meichstedt: The `force` parameter is undocumented

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+                minecraft_survival-world	10013	srv7.hw.ca1.mesosphere.com:31756
+        404:
+          description: No task found with this `app_id`.
+          body:
+            application/json:
+              example: |
+                { "message": "App '/not_existent' does not exist" }
+
+    ###
+    # Create application
```
meichstedt: s/Create/Delete

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+          To override this deployment, use the force=true flag
+        body:
+          application/json:
+            example: |
+              {
+               "message":"App is locked by one or more deployments. Override with the option '?force=true'. View details at '/v2/deployments/<DEPLOYMENT_ID>'.",
+               "deployments":[{"id":"97c136bf-5a28-4821-9d94-480d9fbb01c8"}]
+              }
+
+  /restart:
```
meichstedt: undocumented: `force` query param

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+      is: [ secured ]
+      queryParameters:
+        host:
+          description: all tasks of that application on the supplied slave are killed
+        scale:
+          type: boolean
+          description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks.
+          default: false
+      responses:
+        200:
```
meichstedt: Actually, the response is different depending on the scale param:
- if `scale == true`, the result will be a deployment info
- otherwise, the result will be a list of tasks

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+      delete:
+        description: Kill the task with ID `task_id` that belongs to the application `app_id`.
+        queryParameters:
+          scale:
+            type: boolean
+            description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks.
+            default: false
+        responses:
+          200:
+            description: The tasks that has been killed.
```
meichstedt: - if `scale == true`, a deployment result
- otherwise _the_ task that has been killed (in a field called `task`)

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+        404:
+          description: No task found with this `app_id`.
+          body:
+            application/json:
+              example: |
+                { "message": "App '/not_existent' does not exist" }
+
+    /{task_id}:
+      ###
+      # Create application
```
meichstedt: s/Create application/Delete one task

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,451 @@
...
+              application/json:
+          404:
+            description: No task found with this task_id.
+            body:
+              application/json:
+                example: |
+                  {"message":"Task 'not-existing' does not exist"}
+
+  ###
+  # Create application
```
meichstedt: s/Create/List versions of an

I get the feeling that there is the same amount of misleading comments and of correct ones :see_no_evil: 
Do we need really need them?

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+#
+put:
+  description:
+    Change parameters of a deployed application group.
+    The new group parameters get applied.
+
+
+    * Changes to application parameters will result in a restart of this application.
+
+    * A new application added to the group is started.
```
meichstedt: s/is/will be

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+
+    * An existing application removed from the group gets stopped.
```
meichstedt: s/gets/will be

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+
+
+    If there are no changes to the application definition, no restart is triggered.
+    During restart marathon keeps track, that the configured amount of minimal running instances are _always_ available.
+
+    This method alows 2 special modes for the update operation>
```
meichstedt: s/alows/allows

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+
+    * Provide only the `version` field in the group definition. This will rollback the group to that given version
+
+    * Provide only the `scaleBy` field will scale all transitive applications by the given factor.
```
meichstedt: s/applications/applications' instance counts

Also: an explanation how rounding will be applied might be helpful. Imagine I have a group with 3 apps:
- app-1, instance count: 100
- app-2, instance count: 100
- app-3, instance count: 1

people will ask (themselves or us): what happens to app-3 when I scaleBy 0.1?

docs/docs/rest-api/public/api/v2/groups.raml
meichstedt: Add: When one of `version` or `scaleBy` are set, nothing else than a version change or transitive instance count scaling will be applied. If both `version` and `scaleBy` are set, only a version rollback will be performed – the scaleBy value will not be applied.

- There should be tests for this implementation detail
- It might be better to throw validation errors than to silently drop body parts by performing `versionChange orElse scaleChange getOrElse update.apply`.

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+
+    A deployment can run forever. This is the case, when the new application has a problem and does not become healthy.
```
meichstedt: This paragraph is rendered relative to the above bullet point which, I guess, is unintended. You need to insert another blank line above to fix this …

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+    In this case, human interaction is needed with 2 possible choices
+
+
+    * Rollback to an existing older version
+
+    * Update with a newer version of the group which does not have the problems of the old one.
+
+
+    If there is an upgrade process already in progress, a new update will be rejected unless the force flag is set.
+    With the force flag given, a running upgrade is terminated and a new one is started.
```
meichstedt: `force` should be documented via `queryParameters:`

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+
+    Since the deployment of the group can take a considerable amount of time, this endpoint returns immediatly with a version.
```
meichstedt: s/immediatly/immediately

docs/docs/rest-api/public/api/v2/groups.raml
meichstedt: This can also have other reasons, e.g. lack of suitable resource offers. In that case, the deployment will stay in state `Waiting`, and there's no problem in terms of an error in the AppDefinition.

In general, there's a lot of prose written here. We should consider moving prose to the general documentation and only documenting the rough edges here, while maintaining links from either part to the other. Of course, we might decide to put everything in here, but I think that would be beyond the scope of a REST API doc.

docs/docs/rest-api/public/api/v2/groups.raml
meichstedt: We should describe somewhere, what happens with tasks created during a later on cancelled deployment :speak_no_evil:

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+  body:
+    application/json:
+      example: !include examples/group.json
+      schema: Group
+  responses:
+    200:
+      description:
+        A deployment is started which has a unique deployment identifier.
+        The related deployment can be fetched from the /v2/deployments endpoint.
+        If the deployement is gone from the list of deployments, than this means it is finished.
```
meichstedt: s/deployement/deployment

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+    * Rollback to an existing older version
+
+    * Update with a newer version of the group which does not have the problems of the old one.
+
+
+    If there is an upgrade process already in progress, a new update will be rejected unless the force flag is set.
+    With the force flag given, a running upgrade is terminated and a new one is started.
+
+    Since the deployment of the group can take a considerable amount of time, this endpoint returns immediatly with a version.
+    The failure or success of the action is signalled via event. There is a group_change_success and group_change_failed with
```
meichstedt: s/event/events

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+      body:
+        application/json:
+          example: !include examples/deployments_result.json
+    400:
+      description: The group definition provided in the body is not valid.
+      body:
+        application/json:
+          example: |
+            {"message":"Invalid JSON","details":[{"path":"/id","errors":["error.expected.jsstring"]}]}
+    409:
```
meichstedt: The 409 does not belong here but only to the POST section. A PUT will always update the group as long as there is no validation error.

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+      body:
+        application/json:
+          example: |
+            {
+             "message":"App is locked by one or more deployments. Override with the option '?force=true'. View details at '/v2/deployments/<DEPLOYMENT_ID>'.",
+             "deployments":[{"id":"97c136bf-5a28-4821-9d94-480d9fbb01c8"}]
+            }
+
+
+/{group_id}:
```
meichstedt: docs for /group/{group_id} are the same as for /group – I'd extract those into a separate raml file and include that in both cases. Should we point out that you can shoot against nested groups?

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -0,0 +1,376 @@
...
+                  "error":"identifier /app is not child of /group. Hint: use relative paths."
+                }
+              ]
+            }
+
+
+###
+# Create a new application group
+#
+post:
```
meichstedt: The underlying implementation has a `force` flag that is undocumented but will allow changing the group even if there's a running deployment.

docs/docs/rest-api/public/api/v2/tasks.raml
```diff
@@ -0,0 +1,43 @@
+get:
+  description:
+    List all running tasks.
+  is: [ secured ]
+  queryParameters:
+    status:
+      description: Filter the list of tasks by status
+      enum: [ TASK_STAGING, TASK_STARTING, TASK_RUNNING, TASK_FINISHED, TASK_FAILED, TASK_KILLED, TASK_LOST, TASK_ERROR ]
```
meichstedt: Not quite. According to the [old docs](https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-tasks) and [the implementation](https://github.com/mesosphere/marathon/blob/d098e5f73905d91e9511dbbdf970472656d4649b/src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala#L135-L139), possible values are `[running,staging]`

docs/docs/rest-api/public/api/v2/tasks.raml
```diff
@@ -0,0 +1,43 @@
...
+      repeat: true
+  responses:
+    200:
+      description: The list of running tasks.
```
meichstedt: The list of all tasks disregarding their status, or a list of tasks matching the specified `status` filter.

docs/docs/rest-api/public/api/v2/tasks.raml
```diff
@@ -0,0 +1,43 @@
...
+        type: boolean
+        description: If `scale=true` is specified, then the related application is scaled down by the number of killed tasks.
+        default: false
+    body:
+      application/json:
+        example: |
+          { "ids": [ "task1", "task2" ] }
+    responses:
+      200:
+        description: No content is send with this reponse.
```
meichstedt: Should we document this: if `scale == true`, a `DeploymentPlan` will be created for each app corresponding to one or more of the given tasks. These are not returned, so one cannot track the deployments in order to see when the killing of multiple tasks with scaling eventually finished.

btw: I am not convinced that this functionality makes sense and it used (killing tasks of multiple apps at once while scaling their instance count to whatever remains). Maybe we should deprecate that and see if anyone complains.

docs/docs/rest-api/public/api/v2/events.raml
```diff
@@ -0,0 +1,33 @@
+get:
+  description:
+    Attach to the marathon event stream.
```
meichstedt: The `Try it` section for events is not working

docs/docs/rest-api/public/api/v2/queue.raml
```diff
@@ -0,0 +1,34 @@
...
+        application/json:
+          example: !include examples/queue.json
+
+
+
+/{app_id}/delay:
+
+  delete:
+    description:
+      If an application fails to often in a specified amount of time (according to the application definition),
```
meichstedt: s/to/too

docs/docs/rest-api/public/api/v2/queue.raml
```diff
@@ -0,0 +1,34 @@
...
+      the task launch will be delayed.
+      This delay can be removed by calling this endpoint.
+      The effect is, that the tasks of this application will be launched immediately.
+    responses:
+      200:
+        description: the delay is resetted and no body is returned
```
meichstedt: s/resetted/reset

meichstedt:
Phew, that was a lot to comment, but not too many critical things. In general, I like the structure and maintainability, along with the option to `try it out`.

- several sections contain misleading or wrong information, at least some of those were ported from the old doc
- In various locations, there is more info than just API documentation. That's not bad at all, maybe even better than having these separately in the _general_ documentation. However, having a lot of prose in here creates the need to read a lot, and prevents from quickly scanning through available params. Therefore, I'd rather move too detailed explanations into prose based docs and let the API docs only outline the API. If we do that, then:
  - there ought to be a lot of very specific cross linkages, so people can find the API doc to a given feature, or vice versa
  - we need to provide generated html with anchors for specific endpoints so we can directly link them from the general documentation.

Using raml2html for generating static docs without the `Try it` would make sense. Those static pages should reside on the web site relative to the version for which it was generated. I'm thinking about `/0.13/v2/api` vs `/0.14/v2/api` vs `/0.14/v3/api`, so that there's a specific documentation for each Marathon version and API version.

Please fix the commented issues and let's discuss how we can incorporate this!

docs/docs/rest-api/public/api/api.raml
aquamatthias: Not exactly sure what to do here. Add 2.0 to the title?

docs/docs/rest-api/public/api/traits.raml
aquamatthias: The problem here is, that this message is generated from a plugin. But I agree if this is an example, we should make it one.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: The description can be defined as markdown. So you can highlight anything you want in the description. There is no specific note element.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: The security schema is a different thing - with that you can define how to access the API (e.g. basic auth). The secured trait is attached to all endpoints, that would need that behavior.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Like the idea. Will rewrite that part.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: I have started to integrate this documentation into the json schema. It would be the natural place for raml documentation. Unfortunately there is no nice visualization. So my suggestion would be to add the model description to the introduction part of the documentation.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: I changed that example back and forth: 
- I think most of the people will not try to understand json schema
- I think an example is better than the schema
- I think a short example is easy to understand, but is always only half of the truth.

I decided to use a real example here.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Added a sentence with respect to deployments.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Good catch.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Made a deployable trait, which documents the force flag and added that to all related endpoints.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Add deployable trait.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: True. Added that.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Added that.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: Done.

docs/docs/rest-api/public/api/v2/apps.raml
aquamatthias: While writing this, they help me find the single endpoints more easily. I corrected all - hope this helps also in the future.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: Added. I agree with that checks - should be part of validation though.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: 1) Thanks for looking into the rendered result. The markdown parser is rather clumsy...
2) Part of the problem is also: that the endpoint is able to do __a lot__

Generally I agree: we want to have short/sharp reference docs and longer prose as introduction.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: added deployable

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: removed.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: Done.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: Done.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: Good idea. Removed everything under {group_id} and imported the same file twice.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: forceable -> deployable

docs/docs/rest-api/public/api/v2/tasks.raml
aquamatthias: In theory we could report tasks in other states. From an API perspective that could make sense. In order to document the current status I agree.

docs/docs/rest-api/public/api/v2/tasks.raml
aquamatthias: Added documentation for the 2 options.

docs/docs/rest-api/public/api/v2/events.raml
aquamatthias: This is a question of your expectation: 
- if you add the Accept Header, the Try it out function connects to the event stream and will not return -> you see nothing
- if you omit the Accept Header, you get a 405 response.

That means this functionality can not be used in the console. 
I will add a sentence to the docs.

aquamatthias:
Cool - had the same idea with versioning.

aquamatthias:
@meichstedt Thanks for all the good hints! I hope I addressed them all.
I still feel there is a way for really good documentation, but at least it feels like a good start.

Changes:
- Incorporated (hopefully) all suggestions
- biggest change probably is the usage of the deployable trait (made a lot things easier - thanks for the suggestion)
- removed api-console from the docs folder and use the static generated file (not versioned for now)
- index: moved current rest-api to rest-api (deprecated) and add the new rest-api 
- deployed version 0.1.0 of api-console and use this as runtime dependency for /help

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -0,0 +1,416 @@
...
+      example: my_label, environment==production, tier!=frontend\ tier, deployed in (us, eu), deployed notin (aa, bb)
+    embed:
+      required: false
+      description:
+        Embeds nested resources that match the supplied path.
+        You can specify this parameter multiple times with different values.
+
+        - <code>apps.tasks</code> embed all tasks of each application<br/>
+        Note&#58; if this embed is definded, it automatically sets <code>apps.deployments</code> but this will change in a future release.
+        Please define all embeds explicit.
```
meichstedt: Good idea to add that hint. However, please `s/explicit/explicitly`, as this is an adverb to `define`. This does also hold for the 2 occurrences below.

docs/docs/rest-api/public/api/v2/apps.raml
meichstedt: Ok, I think that would be fine as well.

docs/docs/rest-api/public/api/v2/apps.raml
meichstedt: I agree, a real world example probably explains best. (Additionally, if we get the fancy strict mode implemented, then the result of trying out a JSON snippet will speak for itself).

docs/docs/rest-api/public/api/v2/apps.raml
meichstedt: :+1:

docs/docs/rest-api/public/api/v2/groups.raml
meichstedt: That's a better name indeed.

docs/docs/rest-api/public/api/v2/tasks.raml
meichstedt: 1) the implementation will match the given string like this
```
    case "running" => Some(TaskState.TASK_RUNNING)
    case "staging" => Some(TaskState.TASK_STAGING)
    case _         => None
```
so passing `TASK_STAGING ` will **not** work but return all tasks disregarding the status. In that sense, the documentation is still wrong.

2) Yes, the base for the result of this endpoint is the TaskTracker list, so integrating any other task states would be quite easy.

Please fix 1)

meichstedt:
Hey @aquamatthias looks fine – just 2 little issues left:
- wrong grammar in `apps.raml`
- wrong documentation of the `status` queryParam in `tasks.raml`

I made some more comments but as far as I can see these are the only 2 with need for action.

The latter one prevents me from merging, please fix that.

aquamatthias:
@meichstedt addressed your comments. Thanks. If you agree, I will squash the commits.

meichstedt:
@aquamatthias perfect – please squash and wait for the sound of the foghorn :ship:

PR #2363, opened 2015-10-06
Cache data on leader elected and use in in subsequent reads.

aquamatthias:


src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -53,6 +50,12 @@ object ModuleNames {
   final val NAMED_SERVER_SET_PATH = "SERVER_SET_PATH"
   final val NAMED_SERIALIZE_GROUP_UPDATES = "SERIALIZE_GROUP_UPDATES"
   final val NAMED_HTTP_EVENT_STREAM = "HTTP_EVENT_STREAM"
+
+  final val STORE_APP = "AppStore"
```
kolloch: the other constants are prefixed with "NAMED_". It would be nice to adhere to that convention.

src/main/scala/mesosphere/marathon/health/HealthCheck.scala
```diff
@@ -16,24 +16,23 @@ import scala.concurrent.duration._
 @JsonIgnoreProperties(ignoreUnknown = true)
 case class HealthCheck(
 
-  @FieldJsonDeserialize(contentAs = classOf[java.lang.String]) path: Option[String] = HealthCheck.DefaultPath,
+    @FieldJsonDeserialize(contentAs = classOf[java.lang.String]) path: Option[String] = HealthCheck.DefaultPath,
```
kolloch: why this change?

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,84 @@
...
+import mesosphere.util.ThreadPoolContext
+import org.slf4j.LoggerFactory
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.Future
+
+/**
+  * The entity store cache caches all current versions from the underlying store.
+  * Idea:
+  * - old versions (versioned entries) are never cached
```
kolloch: update the description. (Mention the prefetch on startup)

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,84 @@
...
+  private[this] val log = LoggerFactory.getLogger(getClass)
+
+  override def fetch(key: String): Future[Option[T]] = {
+    if (isUnversionedEntry(key)) {
+      Future.successful(cache.get(key))
+    }
+    else {
+      val unversioned = key.substring(0, key.indexOf(':'))
+      cache.get(unversioned) match {
+        case Some(t) if key == unversioned + ':' + t.version => Future.successful(Some(t))
```
kolloch: maybe add a comment here

src/test/scala/mesosphere/marathon/state/AppRepositoryTest.scala
```diff
@@ -130,7 +130,7 @@ class AppRepositoryTest extends MarathonSpec {
     assert(res.forall(identity), "Should succeed")
 
     verify(store).names()
-    verify(store).expunge("app1")
+    verify(store).expunge("app1", null)
```
kolloch: that's really strange. We probably should fix this later. Mark it?

kolloch:
We should add tests for the cache.

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  override def modify(key: String, onSuccess: (T) => Unit = _ => ())(update: Update): Future[T] = {
+    def onModified(t: T): Unit = {
+      if (isUnversionedEntry(key)) cache.update(key, t)
+      onSuccess(t)
+    }
+    store.modify(key, onModified)(update)
+  }
+
+  override def names(): Future[Seq[String]] = Future.successful(cache.keys.toSeq)
```
kolloch: That's incorrect. The list of names has to include older versions as well.

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
```diff
@@ -38,27 +39,35 @@ class MarathonStore[S <: MarathonState[_, S]](
...
-        bytesWritten.update(updatedEntity.bytes.length)
-        store.update(updatedEntity)
-      case None =>
-        val created = f(() => newState()).toProtoByteArray
-        bytesWritten.update(created.length)
-        store.create(prefix + key, created)
+  def modify(key: String, onSuccess: (S) => Unit = _ => ())(f: Update): Future[S] = {
+    lockManager.executeSequentially(key) {
+      log.debug(s"Modify $prefix$key")
+      val res = store.load(prefix + key).flatMap {
```
kolloch: Generally, this is incorrect.

As described in #2299, writes are not immediately visible. So instead of a fetch here, we should actually do a lookup in our cache. But this is probably an issue for another PR.

src/main/scala/mesosphere/marathon/health/HealthCheck.scala
aquamatthias: we removed the MarathonState extension - so this is just a reformat?

src/main/scala/mesosphere/marathon/state/MarathonStore.scala
aquamatthias: That one is really complicated. But I agree.

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
aquamatthias: Added the versioned entries as well.

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,102 @@
...
+      cache.get(unversioned) match {
+        case Some(Some(t)) if key == unversioned + ':' + t.version => Future.successful(Some(t))
+        case _ => store.fetch(key)
+      }
+    }
+  }
+
+  override def modify(key: String, onSuccess: (T) => Unit = _ => ())(update: Update): Future[T] = {
+    def onModified(t: T): Unit = {
+      if (isUnversionedEntry(key)) cache.update(key, Some(t))
```
kolloch: you need to add `None` for versioned entries

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,102 @@
...
+      unversionedNames.foldLeft[Future[Unit]](Future.successful(())) { (completed, nextName) =>
+        completed.flatMap { _ => preloadEntry(nextName) }
+      }
+    }
+
+    def handleEntries(names: Seq[String]): Future[Unit] = {
+      //add keys with None for version entries
+      names.filterNot(isUnversionedEntry).foreach(cache.put(_, None))
+      //add key with loaded values
+      preloadEntries(names)
```
kolloch: would be cool to make it more obvious that this works on the unversioned stuff

aquamatthias:
Added test cases.

aquamatthias:
Create a new PR that is based on master

PR #2376, opened 2015-10-07
The ec for async IO should not use the bounded ForkJoin Pool.

aquamatthias:


gkleiman:
LGTM, waiting for the tests to pass.

PR #2377, opened 2015-10-07
Make marathon ssl certificate dynamic

titilambert:


lloesche:
Thanks for the PR!

Instead of passing that option through all those methods you can create a template file called HAPROXY_HTTPS_FRONTEND_HEAD which contains your site specific certificate settings.

PR #2387, opened 2015-10-09
Efficiently create groups/apps in  time (was exponential).

aquamatthias:
Group store computed values and provide efficient equals/hashCode implementation.

See Before/After:

screenshot_01

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -84,7 +84,7 @@ case class AppDefinition(
     * port mappings.
     */
   def portIndicesAreValid(): Boolean = {
-    val validPortIndices = 0 until hostPorts.size
+    val validPortIndices = hostPorts.indices
     healthChecks.forall { hc =>
       hc.protocol == Protocol.COMMAND || (validPortIndices contains hc.portIndex)
     }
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -251,6 +251,15 @@ case class AppDefinition(
 
   def isUpgrade(to: AppDefinition): Boolean =
     this != to.copy(instances = instances, version = version)
+
+  override def equals(obj: Any): Boolean = {
```
kolloch: you may not do that because that makes the "isUpgrade" stuff and other things fail.

At least, we need to modify some other things for that.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -110,11 +110,11 @@ case class Group(
     }
   }
 
-  def transitiveApps: Set[AppDefinition] = this.apps ++ groups.flatMap(_.transitiveApps)
+  lazy val transitiveApps: Set[AppDefinition] = this.apps ++ groups.flatMap(_.transitiveApps)
```
kolloch: nice catch!

kolloch:
nice analysis! Unfortunately, there are some invalid shortcuts included in this patch.

aquamatthias:
@kolloch Thanks for pointing that out. I implemented AppDefinition.isUpgrade. Should that be enough?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -179,6 +182,15 @@ case class Group(
...
   def withoutChildren: Group = copy(apps = Set.empty, groups = Set.empty)
+
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case that: Group => (that eq this) || (that.id == id && that.version == version)
+      case _           => false
+    }
+  }
+
+  override def hashCode(): Int = id.hashCode()
```
kolloch: please document your decision. Why did you only include the ID here?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -84,7 +84,7 @@ case class AppDefinition(
     * port mappings.
     */
   def portIndicesAreValid(): Boolean = {
-    val validPortIndices = 0 until hostPorts.size
+    val validPortIndices = hostPorts.indices
```
kolloch: why? just beauty or has this a performance impact?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -179,6 +182,15 @@ case class Group(
 
   @JsonIgnore
   def withoutChildren: Group = copy(apps = Set.empty, groups = Set.empty)
+
+  override def equals(obj: Any): Boolean = {
```
kolloch: where is this called?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
aquamatthias: beauty - I can revert that, if you want.

src/main/scala/mesosphere/marathon/state/Group.scala
aquamatthias: Group.groups is a Set.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
kolloch: I don't care so much. It would be better to revert it. Just to keep it minimal

kolloch:
Let's clean this up later.

PR #2389, opened 2015-10-09
Only throttle non-GET requests

meichstedt:


meichstedt:
LGTM

src/main/scala/mesosphere/marathon/api/LimitConcurrentRequestsFilter.scala
```diff
@@ -15,8 +15,22 @@ class LimitConcurrentRequestsFilter(concurrentOption: Option[Int]) extends Filte
   val filterFunction = concurrentOption.map(_ => withSemaphore _).getOrElse(pass _)
 
   def withSemaphore(request: ServletRequest, response: ServletResponse, chain: FilterChain): Unit = {
+    request match {
+      case r: HttpServletRequest if r.getMethod != "GET" =>
```
aquamatthias: Can you extend the test class, so we make sure GET requests are not blocked.

aquamatthias: I added a test case.

aquamatthias:
LGTM

Issue #1407, opened 2015-04-17
Racy deployment creation


bobrik:
I noticed that some deployments from [ansible role](https://github.com/Topface/ansible-marathon_app) report as finished immediately (under ~200ms) while marathon ui thinks that they are running.

I wrote a simple script that creates an app with `PUT /v2/apps`, gets deployment id and checks if deployment appears immediately in `GET /v2/deployments`. I was't able to replicate results over ssh tunnel, but in datacenter it revealed racy behavior.

Here's my test script:

```sh
#!/bin/sh

set -e

MARATHON=$1
if [ -z "${MARATHON}" ]; then
    echo "usage: <marathon uri> <app id> <app definition>"
    exit 1
fi

MARATHON_APP=$2
if [ -z "${MARATHON_APP}" ]; then
    echo "usage: <marathon uri> <app definition>"
    exit 1
fi

MARATHON_APP_ID=$(echo "${MARATHON_APP}" | jq -r .id)

echo "creating ${MARATHON_APP_ID}"

MARATHON_DEPLOYMENT_ID=$(curl -s -H "Content-type: application/json" -X PUT "${MARATHON}/v2/apps/${MARATHON_APP_ID}" -d "${MARATHON_APP}" | jq -r .deploymentId)

echo "got deployment id: ${MARATHON_DEPLOYMENT_ID}"

curl -s "${MARATHON}/v2/deployments" | jq .

sleep 1

curl -s "${MARATHON}/v2/deployments" | jq .
```

Here's the output:

```
# /tmp/test.sh http://web488:8080 '{"id":"/racy/test1","cmd":"sleep 60","cpus":0.1,"mem":16}'
creating /racy/test1
got deployment id: 917eea45-fc73-4951-89a4-0d5b17c761f4
[]
[
  {
    "currentStep": 1,
    "currentActions": [
      {
        "action": "RestartApplication",
        "app": "/racy/test1"
      }
    ],
    "affectedApps": [
      "/racy/test1"
    ],
    "version": "2015-04-17T13:24:50.601Z",
    "id": "917eea45-fc73-4951-89a4-0d5b17c761f4",
    "totalSteps": 1,
    "steps": [
      [
        {
          "action": "RestartApplication",
          "app": "/racy/test1"
        }
      ]
    ]
  }
]
```

I'm on 0.8.1-SNAPSHOT, not very current one.

drexin:
[...]

PR #2391, opened 2015-10-09
Cache GET request to /v2/apps, /v2/tasks, /v2/deployments

meichstedt:


kolloch:
Let's not merge this to an official release branch. If it is needed for a specific customer, let's make a special branch for this.

It will break things for other people, e.g. https://github.com/mesosphere/marathon/issues/1407.

aquamatthias:
I agree - we should not merge that on master.
Tobi want to have that as band aid.

kolloch:
Please don't include this in a public release!

PR #2393, opened 2015-10-09
Improve LimitConcurrentRequestsFilterTest

kolloch:
the `availablePermits` test was ineffective.


aquamatthias:
Fixed that already - but thanks for the PR.

PR #2394, opened 2015-10-09
Warn about 0.11 in changelog.

kolloch:


gkleiman:
Should we warn here as well? https://mesosphere.github.io/marathon/

kolloch:
probably, I'll prepare another pull request if you fine with that right now.

Do we have any easy way to make people install 0.10 by default using our stable repositories?

gkleiman:
On Ubuntu/Debian you can do:

`apt-get install marathon=<version>`

But the version is different in each distribution version =/. For example for Ubuntu 14.04 you have to do:

`apt-get install marathon=0.10.1-1.0.416.ubuntu1404`

PR #2399, opened 2015-10-12
Efficiently create groups/apps in constant time (was exponential).

aquamatthias:
Group store computed values and provide efficient equals/hashCode implementation.


gkleiman:
LGTM

PR #2400, opened 2015-10-12
Remove dead code

gkleiman:


aquamatthias:
This can be used to graph the dependencies. But is really not used.

PR #2411, opened 2015-10-12
Add storage cache layer.

aquamatthias:
The layer is enabled by default, but can be disabled via command line flag.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonModule.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonModule.scala
@@ -39,16 +39,25 @@ import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
 import scala.concurrent.Await
+import scala.reflect.ClassTag
 import scala.util.control.NonFatal
+import scala.collection.immutable.Seq
 
 object ModuleNames {
-  final val NAMED_CANDIDATE = "CANDIDATE"
-  final val NAMED_HOST_PORT = "HOST_PORT"
+  final val CANDIDATE = "CANDIDATE"
+  final val HOST_PORT = "HOST_PORT"
 
-  final val NAMED_LEADER_ATOMIC_BOOLEAN = "LEADER_ATOMIC_BOOLEAN"
-  final val NAMED_SERVER_SET_PATH = "SERVER_SET_PATH"
-  final val NAMED_SERIALIZE_GROUP_UPDATES = "SERIALIZE_GROUP_UPDATES"
-  final val NAMED_HTTP_EVENT_STREAM = "HTTP_EVENT_STREAM"
+  final val LEADER_ATOMIC_BOOLEAN = "LEADER_ATOMIC_BOOLEAN"
+  final val SERVER_SET_PATH = "SERVER_SET_PATH"
+  final val SERIALIZE_GROUP_UPDATES = "SERIALIZE_GROUP_UPDATES"
+  final val HTTP_EVENT_STREAM = "HTTP_EVENT_STREAM"
+
+  final val STORE_APP = "AppStore"
+  final val STORE_TASK_FAILURES = "TaskFailureStore"
+  final val STORE_DEPLOYMENT_PLAN = "DeploymentPlanStore"
+  final val STORE_FRAMEWORK_ID = "FrameworkIdStore"
+  final val STORE_GROUP = "GroupStore"
+  final val STORE_TASK = "TaskStore"
 }
 
 class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
@@ -79,7 +88,7 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
     bind(classOf[HealthCheckManager]).to(classOf[MarathonHealthCheckManager]).asEagerSingleton()
 
     bind(classOf[String])
-      .annotatedWith(Names.named(ModuleNames.NAMED_SERVER_SET_PATH))
+      .annotatedWith(Names.named(ModuleNames.SERVER_SET_PATH))
       .toInstance(conf.zooKeeperServerSetPath)
 
     bind(classOf[Metrics]).in(Scopes.SINGLETON)
@@ -88,7 +97,7 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
     // If running in single scheduler mode, this node is the leader.
     val leader = new AtomicBoolean(!conf.highlyAvailable())
     bind(classOf[AtomicBoolean])
-      .annotatedWith(Names.named(ModuleNames.NAMED_LEADER_ATOMIC_BOOLEAN))
+      .annotatedWith(Names.named(ModuleNames.LEADER_ATOMIC_BOOLEAN))
       .toInstance(leader)
   }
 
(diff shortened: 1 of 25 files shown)


src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -49,6 +51,13 @@ object ModuleNames {
   final val NAMED_SERVER_SET_PATH = "SERVER_SET_PATH"
   final val NAMED_SERIALIZE_GROUP_UPDATES = "SERIALIZE_GROUP_UPDATES"
   final val NAMED_HTTP_EVENT_STREAM = "HTTP_EVENT_STREAM"
+
+  final val NAMED_STORE_APP = "AppStore"
```
meichstedt: ModuleNames.NAMED_STORE_APP is quite verbose. IMHO ModuleNames.STORE_APP or Named.STORE_APP would be OK since it's only in local scope of that file. Benefit of a rename would be shorter and more readable lines where the vals are used in annotations.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -101,6 +110,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
     }
   }
 
+  @Provides
+  @Singleton
+  def provideLeadershipCallbacks(
```
meichstedt: This is missing the TaskStore. When using an EntityStoreCache one could enforce that the Cache is leadership aware – you could automate this :)

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -414,4 +361,111 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
 
     groupManager
   }
+
+  // persistence fucntionality ----------------
```
meichstedt: s/fucntionality/functionality

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,102 @@
...
+          case _       => None
+        }
+      }
+    }
+    else {
+      //if we need to fetch a versioned entry, try if this is the latest version we have in the cache
+      //otherwise let the underlying store fetch that entry.
+      val unversioned = key.substring(0, key.indexOf(':'))
+      cache.get(unversioned) match {
+        case Some(Some(t)) if key == unversioned + ':' + t.version => Future.successful(Some(t))
```
meichstedt: (De-)construction of keys is a repetition of EntityRepository implementation (while it is not enforced that an EntityRepository is actually used). Maybe introduce a trait with implementation of versionKey and means to deconstruct? Might be considered over engineered, but the current handling of the version prefix is quite unobvious.

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -0,0 +1,102 @@
...
+        case _ => store.fetch(key)
+      }
+    }
+  }
+
+  override def modify(key: String, onSuccess: (T) => Unit = _ => ())(update: Update): Future[T] = {
```
meichstedt: Why did you add the `onSuccess` to the EntityStore?  No one passes an onSuccess into an EntityStore.modify/expunge call but the EntityStoreCache. Both modify and expunge return a Future that could be mapped upon in EntityStoreCache without having to change their signature.

src/test/scala/mesosphere/marathon/state/AppRepositoryTest.scala
```diff
@@ -146,7 +146,7 @@ class AppRepositoryTest extends MarathonSpec {
     assert(res.forall(identity), "Should succeed")
 
     verify(store).names()
-    verify(store).expunge("app1")
+    verify(store).expunge("app1", null) //TODO: remove the null matcher here
```
meichstedt: A `TODO` made it's way into the commit.

src/test/scala/mesosphere/marathon/state/EntityStoreCacheTest.scala
```diff
@@ -0,0 +1,162 @@
...
+import mesosphere.marathon.Protos.MarathonApp
+import mesosphere.marathon.{ MarathonSpec, Protos }
+import mesosphere.util.Mockito
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.{ BeforeAndAfter, GivenWhenThen, Matchers }
+
+import scala.collection.mutable
+import scala.concurrent.Future
+
+class EntityStoreCacheTest extends MarathonSpec with GivenWhenThen with Matchers with BeforeAndAfter with ScalaFutures with Mockito {
```
meichstedt: I like the clear structure and easy readability of this test :)

src/test/scala/mesosphere/marathon/state/EntityStoreCacheTest.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  test("The onDefeated trigger will clear the state") {
+    Given("A pre-filled entityCache")
+    val names = Set("a", "b", "c")
+    entityCache.cache ++= names.map(t => t -> Some(new TestApp(t)))
+
+    When("On defeated is called on the cache")
+    entityCache.onDefeated
+
+    Then("All values are cached")
```
meichstedt: s/cached/removed from the cache

src/test/scala/mesosphere/marathon/state/EntityStoreCacheTest.scala
```diff
@@ -0,0 +1,162 @@
...
+  test("Fetching a versioned entry will delegate to the store") {
+    Given("A pre-filled entityCache")
+    val names = Set("a", "b", "c")
+    val fixed = Timestamp.now()
+    content ++= names.map(t => s"$t:$fixed" -> TestApp(t, fixed))
+
+    When("A known versioned entry is fetched")
+    val a = entityCache.fetch(s"a:$fixed").futureValue
+
+    Then("The store is used to fetch that version")
```
meichstedt: verify that the store was used?

src/test/scala/mesosphere/marathon/state/EntityStoreCacheTest.scala
```diff
@@ -0,0 +1,162 @@
...
+    Given("A pre-filled entityCache")
+    val names = Set("a", "b", "c")
+    val update = TestApp("a", Timestamp.now())
+    content ++= names.map(t => t -> TestApp(t))
+    entityCache.onElected.futureValue
+
+    When("Modify an entity")
+    val a = entityCache.modify("a")(_ => update).futureValue
+
+    Then("The value is returned without interaction to the underlying store")
```
meichstedt: _With_ interaction to the store! Verify?

meichstedt:
I like the idea of just having a wrapper around the EntityStore that does the caching, and it makes sense to only cache the current version. My main comments address the details of maintaining & modifying the cache, as well as the construction and deconstruction of entity store/repo keys.

src/main/scala/mesosphere/marathon/MarathonModule.scala
aquamatthias: I completely agree and did that in my first attempt. Peter made exactly the opposite claim, so I changed it in that way.

src/main/scala/mesosphere/marathon/MarathonModule.scala
aquamatthias: At the time of writing this store was not born - So good catch! 

Please note: if you disable the cache, all direct stores are not leadership aware.

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
aquamatthias: If you look into MarathonStore, you see that every change is locked by LockManager. The onSuccess method is called while that lock still is acquired. That means the change to the cache is also synchronized, while the onSuccess handler is called _after_ the store change.

src/main/scala/mesosphere/marathon/state/VersionedEntry.scala
```diff
@@ -0,0 +1,42 @@
...
+    */
+  def noVersionKey(key: String): Boolean = !isVersionKey(key)
+
+  /**
+    * Create the key without timestamp part.
+    */
+  def versionKeyPrefix(name: String): String = name + VERSION_SEPARATOR
+
+  /**
+    * Gives the id part of th
```
meichstedt: Please finish your sentence.

meichstedt:
Thanks – please finish that one doc line and I'll merge.

aquamatthias:
Finished the sentence. Thanks!

meichstedt:
LGTM, waiting for tests to pass …

aquamatthias:
@meichstedt rebased.

PR #2413, opened 2015-10-13
Add configurable zk node compression.

aquamatthias:
Compression information is stored as part of the surrounding protobuf.
Compression can be turned on/off via cmd line flags and is enabled by default.

### DIFF
--- a/src/main/scala/mesosphere/marathon/ZookeeperConf.scala
+++ b/src/main/scala/mesosphere/marathon/ZookeeperConf.scala
@@ -35,6 +35,21 @@ trait ZookeeperConf extends ScallopConf {
     default = Some(25)
   )
 
+  lazy val zooKeeperCompressionEnabled = toggle("zk_compression",
+    descrYes = "Enable compression of zk nodes, if the size of the node is bigger than the configured threshold.",
+    descrNo = "Disable compression of zk nodes",
+    noshort = true,
+    prefix = "disable_",
+    default = Some(true)
+  )
+
+  lazy val zooKeeperCompressionThreshold = opt[Long]("zk_compression_threshold",
+    descr = "Threshold in bytes, when compression is applied to the zk node (Default: 64 KB).",
+    noshort = true,
+    validate = _ >= 0,
+    default = Some(64 * 1024)
+  )
+
   def zooKeeperStatePath: String = "%s/state".format(zkPath)
   def zooKeeperLeaderPath: String = "%s/leader".format(zkPath)
   def zooKeeperServerSetPath: String = "%s/apps".format(zkPath)
--- a/src/main/scala/mesosphere/util/state/zk/ZKStore.scala
+++ b/src/main/scala/mesosphere/util/state/zk/ZKStore.scala
@@ -6,18 +6,21 @@ import com.fasterxml.uuid.impl.UUIDUtil
 import com.google.protobuf.{ ByteString, InvalidProtocolBufferException }
 import com.twitter.util.{ Future => TWFuture }
 import com.twitter.zk.{ ZNode, ZkClient }
+import mesosphere.marathon.io.IO
 import mesosphere.marathon.{ Protos, StoreCommandFailedException }
 import mesosphere.util.ThreadPoolContext
 import mesosphere.util.state.zk.ZKStore._
 import mesosphere.util.state.{ PersistentEntity, PersistentStore }
 import org.apache.log4j.Logger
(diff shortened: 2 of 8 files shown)


src/main/scala/mesosphere/marathon/ZookeeperConf.scala
```diff
@@ -35,6 +35,21 @@ trait ZookeeperConf extends ScallopConf {
     default = Some(25)
   )
 
+  lazy val zooKeeperCompressionEnabled = toggle("zk_compression",
+    descrYes = "Compression of zk nodes is enabled, if the size of the node is bigger than the configured threshold",
```
gkleiman: Our flags description are usually written in imperative.

For example:

```
    descrYes = "(Default) Enable metric measurement of service method calls",
    descrNo = "Disable metric measurement of service method calls",
```

src/main/scala/mesosphere/marathon/ZookeeperConf.scala
```diff
@@ -35,6 +35,21 @@ trait ZookeeperConf extends ScallopConf {
...
+    descrNo = "Compression of zk nodes is disabled",
+    noshort = true,
+    prefix = "disable_",
+    default = Some(true)
+  )
+
+  lazy val zooKeeperCompressionThreshold = opt[Long]("zk_compression_threshold",
+    descr = "Threshold in bytes, when compression is applied to the zk node.",
```
gkleiman: I'ad add "(default: 64kb)" to the description.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -115,17 +118,25 @@ case class ZKEntity(node: ZNode, data: ZKData, version: Option[Int] = None) exte
 }
 
 case class ZKData(name: String, uuid: UUID, bytes: IndexedSeq[Byte] = Vector.empty) {
-  def toProto: Protos.ZKStoreEntry = Protos.ZKStoreEntry.newBuilder()
-    .setName(name)
-    .setUuid(ByteString.copyFromUtf8(uuid.toString))
-    .setValue(ByteString.copyFrom(bytes.toArray))
-    .build()
+  def toProto(conf: CompressionConf): Protos.ZKStoreEntry = {
```
gkleiman: I'd rename conf to compressionConf or even compression.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
```diff
@@ -6,18 +6,21 @@ import com.fasterxml.uuid.impl.UUIDUtil
...
 import org.apache.log4j.Logger
 import org.apache.zookeeper.KeeperException
-import org.apache.zookeeper.KeeperException.{ NodeExistsException, NoNodeException }
+import org.apache.zookeeper.KeeperException.{ NoNodeException, NodeExistsException }
 
 import scala.concurrent.duration.Duration
 import scala.concurrent.{ Await, Future, Promise }
 
-class ZKStore(val client: ZkClient, rootNode: ZNode) extends PersistentStore {
+case class CompressionConf(enabled: Boolean, sizeLimit: Long)
```
gkleiman: Our *Conf classes always inherit from ScallopCon. Is this one different for any special reason?

Maybe naming it CompressionSettings could make it clearer that it is not like the other *Conf classes we have.

src/test/scala/mesosphere/util/state/zk/ZKStoreTest.scala
```diff
@@ -49,11 +49,37 @@ class ZKStoreTest extends PersistentStoreTest with StartedZookeeper with Matcher
     zkLoadUpdated.get.bytes should be("Hello again".getBytes)
   }
 
+  test("Entity nodes greater then the compression limit get compressed") {
```
gkleiman: s/then/than/

gkleiman:
Looks pretty good to me.

Made just a few comments regarding the flag description and the CompressionConf class.

gkleiman:
**Reminder:** we have to cherrypick this in master.

src/main/scala/mesosphere/util/state/zk/ZKStore.scala
aquamatthias: I don't want to use ScallopConf objects other than setup logic, mainly to make testing code easy.

gkleiman:
We should add the new flags to our docs.

PR #2415, opened 2015-10-13
Fixes #2369 - Increase default --task_launch_confirm_timeout

### LINKED ISSUE #2369: Large file URIs cause "Failed to fetch all URIs for container" error when pulling from HDFS.
Hello,

After upgrading to Marathon 0.11 today, a number of containers that we tried to start would fail.  Looking at the logs, we could see the "Failed to fetch all URIs for container" error.  Checking HDFS, everything looked fine.  We could even manually retrieve the files from HDFS.

After a good amount of debugging, it was discovered that it look like it's only the larger files (100MB and larger) that have issues.  It looks like something times out either during the download or during the extraction of the tar files.  Is there a new setting that we're missing or something else?

We tested this with a web server instead of HDFS with no issues.

FYI, we're still on Mesos 0.23.0.

gkleiman:


kolloch:
You still need to change the documentation. `command-line-flags.md`

PR #2423, opened 2015-10-13
Fix the seed.

aquamatthias:
Does that help?


gkleiman:
This doesn't compile =/

gkleiman:
The content of the commit is good, but the message is from another commit?

I'd also change the comment to:

 // use a fixed seed to get reproducible results

gkleiman:
The scale tests failed, we could try with another fixed seed, but I think there might be a bug behind these failures...

PR #2428, opened 2015-10-13
Add documentation and set the default value explicitly.

aquamatthias:
The default is already false but that should be defined explicitly.

### DIFF
--- a/src/main/proto/marathon.proto
+++ b/src/main/proto/marathon.proto
@@ -188,5 +188,9 @@ message ZKStoreEntry {
   required string name = 1;
   required bytes uuid = 2;
   required bytes value = 3;
-  optional bool compressed = 4;
+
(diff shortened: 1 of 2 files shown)


src/main/proto/marathon.proto
```diff
@@ -188,5 +188,9 @@ message ZKStoreEntry {
   required string name = 1;
   required bytes uuid = 2;
   required bytes value = 3;
-  optional bool compressed = 4;
+
+  // This flag indicates, if the byte array in value is gzip compressed
+  // Introduced in Marathon 0.12
+  // Optional to be backward compatible and also with mesos zk storage.
```
gkleiman: I'd remove the " and also with mesos zk storage"

gkleiman:
LGTM.

Do you want to document the new flags in this PR or should I merge this?

PR #925, opened 2014-12-30
Slim docker image

bobrik:
Using mesos image as base, removing caches and unneded files after sbt.

Image size is much smaller, 1.316gb vs 1.951gb for current image. Moreover, if you run marathon on the same node as mesos (which is common case), you get 1.172gb shared with mesos-master image. It makes 150mb on disk instead of 1951mb. Layer where marathon jar is built:

```
47f807a08209        23 seconds ago      /bin/sh -c sbt assembly &&     mv $(find targ   59.39 MB
```

And old one:

```
7fc8b0dc7d69        6 weeks ago         /bin/sh -c sbt assembly                         483.9 MB
```

However, ideal case would be (single `RUN` step):

1. Install jdk, scala, sbt
2. Build marathon jar
3. Remove jdk, scala, sbt
4. Install jre

This way image would only contain needed files to run marathon, not to build it. Disk and bandwidth savings compared to proposed version are not that big to do it now.

Bonus: mesos version is now more obvious.

PR #2437, opened 2015-10-14
Dockerfile*: Separate resolving dependencies

kolloch:
from the rest of the build to speed up local cached builds.

not essential but helps while testing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,15 @@ RUN apt-key adv --keyserver keyserver.ubuntu.com --recv E56151BF && \
     apt-get update && \
     apt-get install --no-install-recommends -y --force-yes mesos=0.23.0-1.0.debian81 sbt
 
-COPY . /marathon
 WORKDIR /marathon
+
+# The build configuration including dependencies changes
+# less frequently than the source code. By separating
+# these steps we can greatly speed up cached local docker builds
+COPY project /marathon/project
+RUN sbt update test:update
+
+COPY . /marathon
 
 RUN sbt -Dsbt.log.format=false assembly && \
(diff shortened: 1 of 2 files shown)


kolloch:
Also merge into master!

Dockerfile
```diff
@@ -6,9 +6,16 @@ RUN apt-key adv --keyserver keyserver.ubuntu.com --recv E56151BF && \
     apt-get update && \
     apt-get install --no-install-recommends -y --force-yes mesos=0.23.0-1.0.debian81 sbt
 
-COPY . /marathon
 WORKDIR /marathon
 
+# The build configuration including dependencies changes
+# less frequently than the source code. By separating
+# these steps we can greatly speed up cached local docker builds
+ADD project /marathon/project
```
gkleiman: Are you using `ADD` instead of `COPY` for any particular reason?

According to Docker, `COPY` should be preferred over `ADD` (https://docs.docker.com/articles/dockerfile_best-practices/#add-or-copy):

> Although ADD and COPY are functionally similar, generally speaking, COPY is preferred. That’s because it’s more transparent than ADD. COPY only supports the basic copying of local files into the container, while ADD has some features (like local-only tar extraction and remote URL support) that are not immediately obvious. Consequently, the best use for ADD is local tar file auto-extraction into the image, as in ADD rootfs.tar.xz /.

kolloch: No. I'll try with COPY instead.

kolloch:
Changed ADD to COPY

bobrik:
* `v0.11.0` 388mb
* `v0.11.1-RC1` 579mb

I see no reason to change release builds when you can only change dev builds.

bobrik:
https://github.com/mesosphere/marathon/pull/925#issuecomment-68431390

PR #2448, opened 2015-10-15
[0.11] [Discuss Merge option] Add storage cache layer.

aquamatthias:
The layer is enabled by default, but can be disabled via command line flag.


aquamatthias:
Not in 0.11.1. Close for now.

PR #2449, opened 2015-10-15
Integration tests should be retried before they get reported as failed.

aquamatthias:


kolloch:
I disagree. Shaky integration tests have to get fixed. I admit that we have too often just retried running tests instead of analyzing them but I do not want to codify that practice.

kolloch:
I'll close this for now. Please reopen if you disagree.

PR #2462, opened 2015-10-15
Update Marathon for Mesos 0.25.0.

ConnorDoyle:
Mesos added a new method to the _SchedulerDriver_ interface to suppress resource offers (more ergonomic than declining successive offers with long filter timeouts), so there were some small updates required in _SimulatedDriver.scala_.

Regenerated the Marathon protobuf classes based on the 0.25.0 _mesos.proto_.

Ran unit and integration tests, successfully launched and killed tasks on Mesos 0.25.0 using the _mesos-local_ utility.

### DIFF
--- a/mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
+++ b/mesos-simulation/src/main/scala/mesosphere/mesos/simulation/DriverActor.scala
@@ -4,7 +4,7 @@ import java.util.UUID
 
 import akka.actor.{ Actor, ActorRef, Cancellable, Props }
 import akka.event.LoggingReceive
-import mesosphere.mesos.simulation.DriverActor.{ ChangeTaskStatus, ReconcileTask, ReviveOffers, KillTask, LaunchTasks }
+import mesosphere.mesos.simulation.DriverActor._
 import mesosphere.mesos.simulation.SchedulerActor.ResourceOffers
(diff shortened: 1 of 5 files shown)


ConnorDoyle:
Integration tests failed the first time through... restarted them prior to investigation in case it was due to flakiness.

tobilg:
Would be nice to see this merged. I deployed Mesos 0.25.0 yesterday, and I can't create apps now with the current Marathon version. Quite sure I'm not the only one...

gkleiman:
@tobilg AFAIK Marathon should work on Mesos 0.25.0 even without this PR.

Could you please add some logs and your app definition to this PR?

tobilg:
@gkleiman Sorry, I already rolled back to 0.24.1. As I run the Masters/Slaves as Docker images on CoreOS, the logs are now gone unfortunately.

App definition was simple:

```
{
    "id": "sleep-unique",
    "cmd": "sleep 60",
    "instances": 3,
    "constraints": [["hostname", "UNIQUE"]]
  }
```

When I started the app, it was acknowledged by Marathon, but apparantly it couldn't allocate resources. I had a look at the logs of Marathon and the Slaves, but couldn't find anything concerning errors etc. It more seemed like a communication problem between Marathon and the Mesos 0.25.1 masters. Sorry that I can't be more specific. Next time I need to save the logs.

PR #2470, opened 2015-10-16
[0.11] Add datadog and graphite reporter

aquamatthias:


src/main/scala/mesosphere/marathon/metrics/MetricsReporterService.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  def startUp() {
+    this.graphite = config.graphite.get.map(startGraphiteReporter)
+    this.datadog = config.dataDog.get.map(startDatadog)
+  }
+
+  def shutDown() {
+    graphite.foreach(_.stop)
+    datadog.foreach(_.stop())
```
gkleiman: Why do you have the parens here and not in the previous line?

gkleiman:
Looks good to me, but I think we should document the reporters.

gkleiman:
I just noticed that this is a backport to 0.11.1.

I think my comments only make sense for master. If you agree with them, then I'll just merge this and create an issue.

src/main/scala/mesosphere/marathon/metrics/MetricsReporterService.scala
aquamatthias: No reason and should be consistent.

aquamatthias:
I agree.

PR #2474, opened 2015-10-16
Fixes #1988 - Move from log4j to logback

### LINKED ISSUE #1988: Consolidate logging

gkleiman:


src/main/resources/logback.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<configuration>
+    <appender name="stdout" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>[%d] %p %m \(%c:%t\)%n</pattern>
```
kolloch: I like the long format specifiers better (%date, %message, ...) because otherwise I always have to look up the meaning.

Also, I would align and length limit fields such as level and class.

Can you include some sample logging output?

Appending the logger name and the thread in the end in the format given is quite unusual. Isn't it? The meaning is not the same as in the log4j configuration.

project/build.scala
```diff
@@ -316,18 +318,22 @@ object Dependency {
...
-  val hadoopCommon = "org.apache.hadoop" % "hadoop-common" % V.Hadoop excludeAll(excludeMortbayJetty, excludeJavaxServlet)
+  val hadoopHdfs = "org.apache.hadoop" % "hadoop-hdfs" % V.Hadoop excludeAll(excludeMortbayJetty, excludeJavaxServlet,
+    excludeLog4j)
+  val hadoopCommon = "org.apache.hadoop" % "hadoop-common" % V.Hadoop excludeAll(excludeMortbayJetty,
+    excludeJavaxServlet, excludeSlf4jLog4j12, excludeLog4j)
   val beanUtils = "commons-beanutils" % "commons-beanutils" % "1.9.2"
   val scallop = "org.rogach" %% "scallop" % V.Scallop
   val jsonSchemaValidator = "com.github.fge" % "json-schema-validator" % V.JsonSchemaValidator
-  val twitterZk = "com.twitter" %% "util-zk" % V.TwitterZk
+  val twitterZk = "com.twitter" %% "util-zk" % V.TwitterZk excludeAll(excludeSlf4jLog4j12, excludeLog4j)
```
kolloch: rather do the exclusions for all dependencies with a map than do it individually.

src/main/resources/logback.xml
gkleiman: This file was created using the (log4j.properties to logback.xml Translator)[http://logback.qos.ch/translator/].

The output is the same we had with log4j in master (see https://github.com/mesosphere/chaos/commit/9ceaa5a6ba823fde19db5d01486739bf58d632f4#diff-df4f6f706036a0c5d315105bff3710adR5):

`[2015-10-16 16:28:41,913] INFO Task task_a has been killed. Waiting for 1 more tasks to be killed. (mesosphere.marathon.upgrade.AppStopActor:System-akka.actor.default-dispatcher-6)`

I agree our format is not the nicest, but I'd change it in a different PR.

src/main/scala/mesosphere/marathon/DebugConf.scala
```diff
@@ -82,7 +82,10 @@ class DebugModule(conf: DebugConf) extends AbstractModule {
 
   override def configure(): Unit = {
     //set trace log level
-    conf.logLevel.get.foreach(level => Logger.getRootLogger.setLevel(Level.toLevel(level.toUpperCase)))
+    conf.logLevel.get.foreach { level =>
```
aquamatthias: The description in `logging_level` says fatal. Please remove that rom the description. Do yo think we should translate fatal to error?

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
```diff
@@ -43,7 +43,7 @@ object ProcessKeeper {
   }
 
   def startZooKeeper(port: Int, workDir: String) {
-    val args = "org.apache.zookeeper.server.ZooKeeperServerMain" :: port.toString :: workDir :: Nil
+    val args = "-Dzookeeper.jmx.log4j.disable=true" :: "org.apache.zookeeper.server.ZooKeeperServerMain" :: port.toString :: workDir :: Nil
```
aquamatthias: This is the _trick_ to not rely on log4j server?

src/main/resources/logback.xml
aquamatthias: I agree with you @gkleiman with not changing the format here. I still like the idea of the long format specification.

project/build.scala
```diff
@@ -270,13 +272,13 @@ object Dependencies {
     Test.scalatest % "test",
     Test.mockito % "test",
     Test.akkaTestKit % "test"
-  )
+  ).map(_.excludeAll(excludeSlf4jLog4j12).excludeAll(excludeLog4j))
```
aquamatthias: Nice.

aquamatthias:
Thanks. 2 small comments, but generally good!

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
gkleiman: Yup, that's the magical parameter...

gkleiman:
@aquamatthias thanks for the review! I addressed all your comments.

aquamatthias:
:100: Thanks!

PR #2475, opened 2015-10-16
Fixes #2147 - Remove unnecessary assertions

### LINKED ISSUE #2147: Can't remove cmd from application without ugly hack
- Create a simple application eg `cmd: "sleep 10"`
- Update the config with a docker image and remove the `cmd` field
- The `cmd` field persists if not set or set to null (as expected), and responds with a `minLength` validation failure if set to blank. 

The UI currently works around this by sending a single space character.

janisz:
Deleted assertions are covered by `PortIndicesValidator`
I added new test case in `V2AppdefinitionTest` that was broken
before, because of those assertions.


aquamatthias:
Hey @janisz thanks for that!

PR #2481, opened 2015-10-19
Exclude JCL from the dependencies.

gkleiman:
Remove transitive dependency on Commons Logging, because it conflicts
with jcl-over-slf4j, breaking 'sbt assembly'.


aquamatthias:
why did the PR that introduced logback not break?

gkleiman:
Because it runs "sbt integration:test", but not "sbt assembly".

Maybe we should add assembly to the TeamCity task.

PR #2483, opened 2015-10-20
Fixes #1835 by creating a deployment for kill with scale, which will …

### LINKED ISSUE #1835: No error received when a DELETE is sent to a task in deployment
See https://github.com/mesosphere/marathon/issues/1780 .

When a `DELETE` is sent, for example 

`curl -X DELETE "http://localhost:8080/v2/apps/foobar/tasks/foobar.5cca1ad9-2ce5-11e5-ad2d-aa4a641fc847?scale=true"` 

the server returns

```
{
  "task": {
    "id": "foobar.5cca1ad9-2ce5-11e5-ad2d-aa4a641fc847",
    "host": "172.18.6.103",
    "ports": [
      31479
    ],
    "startedAt": null,
    "stagedAt": "2015-07-18T00:38:52.055Z",
    "version": "2015-07-17T13:37:07.842Z"
  }
}
```
–  the task is actually not being deleted. 

Shouldn't it return an error in this case?

aquamatthias:
…be denied if the same app is locked.

The kill operation always returns a result:
 - either a deployment (scale=true)
 - or the tasks that will be killed.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
@@ -8,21 +8,22 @@ import javax.ws.rs.core.{ Context, MediaType, Response }
 
 import com.codahale.metrics.annotation.Timed
 import mesosphere.marathon.Protos.MarathonTask
-import mesosphere.marathon.api._
 import mesosphere.marathon.api.v2.json.Formats._
-import mesosphere.marathon.api.{ EndpointsHelper, MarathonMediaType, TaskKiller }
+import mesosphere.marathon.api.{ EndpointsHelper, MarathonMediaType, TaskKiller, _ }
 import mesosphere.marathon.core.appinfo.EnrichedTask
 import mesosphere.marathon.health.HealthCheckManager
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
```diff
@@ -106,6 +108,7 @@ class TasksResource @Inject() (
   @Path("delete")
   def killTasks(
     @QueryParam("scale")@DefaultValue("false") scale: Boolean,
+    @QueryParam("force")@DefaultValue("false") force: Boolean,
```
meichstedt: Can you add tests for `force = true`? This param is
- only evaluated for `scale = true`
- not documented at all
Fixing the second point could clarify the first.

meichstedt:
Can you please add test(s) as commented above? Apart from that: LGTM, will merge as soon as there's a meaningful test :)

aquamatthias:
Added a test case for killing with force.

meichstedt:
Thanks – waiting for tests to pass …

PR #2484, opened 2015-10-20
Cleanup MarathonModule

gkleiman:
MarthonConf is an attribute of MarathonModule, so there is no need to
pass it to provider methods.


aquamatthias:
LGTM.

PR #2491, opened 2015-10-21
[WIP] Network isolation support

ConnorDoyle:
TODO:

- [ ]  This patch should also take ports into account, e.g. how do users communicate service ports to load balancer configuration programs without consuming port resources on the host? Solution for the MVP: do not allow ports and network interfaces.
- [ ] Allow only COMMAND healthchecks in app definitions with non-empty “networkIntefaces”.


ConnorDoyle:
NB: There are pending changes to the `NetworkInfo` message that have ramifications for this patch.  See https://reviews.apache.org/r/39531/

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -68,9 +70,15 @@ trait Formats
   }
 
   implicit lazy val MarathonTaskWrites: Writes[MarathonTask] = Writes { task =>
+    val containerAddress = {
+      val result = MarathonTasks.hostAddress(task)
+      if (result.nonEmpty) JsString(result) else JsNull
```
aquamatthias: In which case will this String be null?

src/main/scala/mesosphere/marathon/core/CoreGuiceModule.scala
```diff
@@ -84,9 +84,18 @@ class CoreGuiceModule extends AbstractModule {
...
+
+      // Update the task tracker _first_.
+      //
+      // Subsequent steps (for example, the health check subsystem) depend on
+      // task tracker lookup to determine the routable host address for running
+      // tasks.  In case this status update is the first TASK_RUNNING update
+      // in IP-per-container mode, we need to store the assigned container
+      // address reliably before attempting to initiate health checks, or
+      // publish events to the bus.
+      updateTaskTrackerStepImpl,
```
aquamatthias: @kolloch what was the reason to update the task tracker as 3rd step. I see the point of the health check manager. What about the rate limiter?

src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
```diff
@@ -74,7 +75,7 @@ class HealthCheckWorkerActor extends Actor with ActorLogging {
     }
 
   def http(task: MarathonTask, check: HealthCheck, port: Int): Future[Option[HealthResult]] = {
-    val host = task.getHost
+    val host = MarathonTasks.hostAddress(task)
```
aquamatthias: The most natural way would be a call to task (task.hostAddress). Since this is a proto, this is not possible. We should either think about to not use the proto object directly or use pimp my library extension pattern.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -69,6 +69,8 @@ case class AppDefinition(
 
   acceptedResourceRoles: Option[Set[String]] = None,
 
+  network: Option[Seq[Network]] = AppDefinition.DefaultNetwork,
```
aquamatthias: Is there a difference between None and Nil? Is Some(Nil) valid?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -227,7 +237,13 @@ case class AppDefinition(
       upgradeStrategy =
         if (proto.hasUpgradeStrategy) UpgradeStrategy.fromProto(proto.getUpgradeStrategy)
         else UpgradeStrategy.empty,
-      dependencies = proto.getDependenciesList.asScala.map(PathId.apply).toSet
+      dependencies = proto.getDependenciesList.asScala.map(PathId.apply).toSet,
+      network =
+        if (proto.hasNetwork) {
```
aquamatthias: make this a separate function

aquamatthias:
Practically, ports resources always refer to the Mesos agent's host interface so Marathon should avoid consuming ports when also sending NetworkInfo. This functionality is missing?

aquamatthias:
Did you had a chance to test the functionality provided here?

aquamatthias:
Deprecate this PR by #2698

PR #2495, opened 2015-10-21
Omit classpath from ProcessKeeper output

gkleiman:
Omit the classpath in order to avoid cluttering the integration tests output.


aquamatthias:
LGTM

PR #2503, opened 2015-10-21
corrects documentation for current step of deployment events

pbnsilva:
Current documentation does not match actual format.


aquamatthias:
@pbnsilva Thanks!

PR #2504, opened 2015-10-21
Use sessionTimeout for session timeout.

aquamatthias:
Use a timer for the storage connection.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonModule.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonModule.scala
@@ -14,6 +14,7 @@ import com.google.inject._
 import com.google.inject.name.Names
 import com.twitter.common.base.Supplier
 import com.twitter.common.zookeeper.{ Candidate, CandidateImpl, Group => ZGroup, ZooKeeperClient }
+import com.twitter.util.JavaTimer
 import com.twitter.zk.{ NativeConnector, ZkClient }
 import mesosphere.chaos.http.HttpConf
 import mesosphere.marathon.Protos.MarathonTask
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -140,10 +141,9 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
   @Singleton
   def provideStore(): PersistentStore = {
     def directZK(): PersistentStore = {
-      implicit val timer = com.twitter.util.Timer.Nil
       import com.twitter.util.TimeConversions._
-      val sessionTimeout = conf.zooKeeperSessionTimeout.get.map(_.millis).getOrElse(30.minutes)
-      val connector = NativeConnector(conf.zkHosts, None, sessionTimeout, timer)
+      val sessionTimeout = conf.zooKeeperSessionTimeout().millis
+      val connector = NativeConnector(conf.zkHosts, None, sessionTimeout, new JavaTimer(true))
```
gkleiman: Please include the name of the parameter when creating the JavaTimer (isDaemon).

src/main/scala/mesosphere/marathon/ZookeeperConf.scala
```diff
@@ -17,11 +17,11 @@ trait ZookeeperConf extends ScallopConf {
 
   lazy val zooKeeperTimeout = opt[Long]("zk_timeout",
```
gkleiman: What's the difference between these two timeouts?

We should add more info to the descriptions.

aquamatthias: We used the zktimeout also as session timeout, which is just wrong.

gkleiman:
Brilliant!

PR #2510, opened 2015-10-22
Fix SingleAppScalingTest by omitting the reconciliation case.

aquamatthias:


meichstedt:
That looks just awesome :+1: 
I created #2513 in order to keep track of the removed IT complexity – ITs should be able to run without preventing task reconciliation.

merging …

PR #2516, opened 2015-10-22
Use mesos 0.25.0 for docker images.

aquamatthias:


aquamatthias:
Created a mesos bug report here: https://issues.apache.org/jira/browse/MESOS-3793

aquamatthias:
Mesos 0.25 does not work with cgroups enabled (see jira ticket).
mesos local is now started with posix isolator.
Do we want to wait until the cgroups problem is fixed?

kolloch:
Yeah, let's wait for a fix. Your last log looks like a hard dependency on systemd. If that is wanted, the "fix" can also be to include systemd in the docker image.

aquamatthias:
@kolloch I just asked Jojy for an ETA for the fix. Given that we still use 0.24 I am not sure if a recent version of Mesos with posix isolator is better than an older version...

aquamatthias:
According to that ticket, the posix launcher was used before per default.
> There is some recent change that tries to use linux launcher whenever possible (running under linux with root permission). Previously, posix launcher is always used for local cluster.
Since mesos local might create multiple slaves on the same host, we should always use posix launcher (linux launcher does not work if there're multiple slaves running on the same host).

meichstedt:
LGTM – but isn't this worth a changelog entry?

aquamatthias:
@meichstedt: you are right - added a small section in the changelog.

meichstedt:
:+1: can you please squash?

PR #2517, opened 2015-10-22
Fixes EntityStoreCacheTest: Fetching a versioned entry will delegate …

meichstedt:
…to the store and unknown versions will result in None


aquamatthias:
Thanks!

PR #2526, opened 2015-10-23
The RAML Console loads the files from /public/api/api.raml

aquamatthias:
This does not work on dcos.
Changed the path to relative url: ../public/api/api.raml and created a new version.


meichstedt:
LGTM!

Issue #2532, opened 2015-10-23
LeadershipCallback.onElected: the default zk_timeout is too low with real world data


aquamatthias:
The whole zk state has to be read during startup time:
-groups
-apps
-tasks
In my example setup I had 500 apps with each one task. 
I got a timeout exception during leader election.

gomes:
This worries me, I have 264 apps with 770 tasks currently. I had to decrease the number of versions in Zookeeper to 3.

gkleiman:
@gomes: this doesn't affect any of the released versions and is a blocker, which means that it will be fixed before we ship the next release candidate (v0.13.0-RC1).

kolloch:
Fixed by #2533

PR #2533, opened 2015-10-23
Fixes 2532 by defining a specific timeout, which is by default bigger than the zk timeout

aquamatthias:


meichstedt:
@aquamatthias The commit message should say `Fixes #2532`, not `5232`.

meichstedt:
LGTM – please update the commit message and I'll merge.

PR #2551, opened 2015-11-03
Fixes 2550 by changing the order if the leader got elected.

aquamatthias:
The second problem is the failing caching mechanism, when the prefix is empty (as this is the case for frameworkId)
Added a migration from frameworkId->framework:id

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
@@ -17,6 +17,9 @@ object MarathonSchedulerDriver {
                 httpConfig: HttpConf,
                 newScheduler: MarathonScheduler,
                 frameworkId: Option[FrameworkID]): SchedulerDriver = {
+
+    log.info(s"Create new Scheduler Driver with frameworkId: $frameworkId")
+
     val frameworkInfoBuilder = FrameworkInfo.newBuilder()
       .setName(config.frameworkName())
       .setFailoverTimeout(config.mesosFailoverTimeout().toDouble)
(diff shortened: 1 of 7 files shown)


src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
```diff
@@ -17,6 +17,9 @@ object MarathonSchedulerDriver {
                 httpConfig: HttpConf,
                 newScheduler: MarathonScheduler,
                 frameworkId: Option[FrameworkID]): SchedulerDriver = {
+
+    log.info(s"Create new Scheduler Driver with frameworkId: $frameworkId")
```
kolloch: Thanks! Nice addition :+1:

src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala
```diff
@@ -83,6 +83,7 @@ class EntityStoreCache[T <: MarathonState[_, T]](store: EntityStore[T])
 
     def handleEntries(names: Seq[String]): Future[Unit] = {
       val (unversionedNames, versionedNames) = names.partition(noVersionKey)
+      log.debug(s"$store Preload and cache entries: $unversionedNames and versioned entries $versionedNames")
```
kolloch: These can result in a lot of "names". So please wrap the debug statement in an `if (log.isDebugEnabled)`.

(We should really switch to suitable Scala logging macros)

aquamatthias: Done - and I agree. Any proposals?

kolloch:
LGTM!

PR #2571, opened 2015-11-05
Fixes #2314 - Do not remove health checks that are still needed

### LINKED ISSUE #2314: Health state gets lost sporadically
Multiple users seem to report that the health state is not preserved in the context of upgrades.

See #2311 and #2234.

We assumed that happens if you migrate from really old versions but maybe the problem is more wide-spread.

I could imagine that reconciliation does not preserve health status or something like that.

We have verified the following and not found errors:

* Health state of an app with command health checks is preserved when Marathon restarts.
* Health state of an app with command health checks is preserved when migrating from 0.10.x to 0.11.x.

kolloch:
I refactored the code slightly, e.g. make it more clear for what block a lock
is active (it was misleading before due to Futures). The code should probably
be converted into an actor or something similar because the synchronization
is probably not sufficient as it is.

I also made the code perform slightly better during reconciliation
if no health checks need to be added.

I also noticed that the health check requires that all app versions for which tasks
are still active are kept in the AppRepository. Our current maxVersions policy does
not ensure that. I made an appropriate FIXME for that.

This patch also improves the logging of health checks by including app versioning information.
This conveys better that the HealthCheckActor is actually instantiated per app version.

### DIFF
--- a/src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
+++ b/src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
@@ -115,30 +115,43 @@ class MarathonHealthCheckManager @Inject() (
   override def reconcileWith(appId: PathId): Future[Unit] =
     appRepository.currentVersion(appId) flatMap {
       case None => Future(())
-      case Some(app) => appHealthChecks.writeLock { ahcs =>
-        val versionTasksMap: Map[String, MarathonTask] =
-          taskTracker.getTasks(app.id).iterator.map { task => task.getVersion -> task }.toMap
+      case Some(app) =>
+        log.info(s"reconcile [$appId] with latest version [${app.version}]")
 
-        // remove health checks for which the app version is not current and no tasks remain
-        // since only current version tasks are launched.
-        for {
-          (version, activeHealthChecks) <- ahcs(appId)
-          if version != app.version && !versionTasksMap.contains(app.id.toString)
-          activeHealthCheck <- activeHealthChecks
-        } {
-          remove(appId, version, activeHealthCheck.healthCheck)
+        val tasks: Iterable[MarathonTask] = taskTracker.getTasks(app.id)
+        val activeAppVersions: Set[String] = tasks.iterator.map(_.getVersion).toSet + app.version.toString
+
+        val healthCheckAppVersions: Set[String] = appHealthChecks.writeLock { ahcs =>
+          // remove health checks for which the app version is not current and no tasks remain
+          // since only current version tasks are launched.
+          for {
+            (version, activeHealthChecks) <- ahcs(appId)
+            if version != app.version && !activeAppVersions.contains(version.toString)
+            activeHealthCheck <- activeHealthChecks
+          } remove(appId, version, activeHealthCheck.healthCheck)
+
+          ahcs(appId).iterator.map(_._1.toString).toSet
         }
 
         // add missing health checks for the current
         // reconcile all running versions of the current app
-        val res = versionTasksMap.keys map { version =>
+        val appVersionsWithoutHealthChecks: Set[String] = activeAppVersions -- healthCheckAppVersions
+        val res: Iterator[Future[Unit]] = appVersionsWithoutHealthChecks.iterator map { version =>
           appRepository.app(app.id, Timestamp(version)) map {
-            case None             =>
-            case Some(appVersion) => addAllFor(appVersion)
+            case None =>
+              // FIXME: If the app version of the task is not available anymore, no health check is started.
+              // We generated a new app version for every scale change. If maxVersions is configured, we
+              // throw away old versions such that we may not have the app configuration of all tasks available anymore.
+              log.warn(
+                s"Cannot find health check configuration for [$appId] and version [$version], " +
+                  "using most recent one.")
+
+            case Some(appVersion) =>
+              log.info(s"addAllFor [$appId] version [$version]")
+              addAllFor(appVersion)
           }
         }
         Future.sequence(res) map { _ => () }
-      }
     }
 
   override def update(taskStatus: TaskStatus, version: Timestamp): Unit =
--- a/src/test/scala/mesosphere/marathon/health/MarathonHealthCheckManagerTest.scala
+++ b/src/test/scala/mesosphere/marathon/health/MarathonHealthCheckManagerTest.scala
@@ -1,12 +1,14 @@
 package mesosphere.marathon.health
 
+import akka.actor.ActorDSL._
 import akka.actor._
 import akka.event.EventStream
-import akka.testkit.EventFilter
+import akka.testkit.{ EventFilter, TestProbe }
 import com.codahale.metrics.MetricRegistry
 import com.typesafe.config.ConfigFactory
 import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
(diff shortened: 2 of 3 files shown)


src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -115,30 +115,41 @@ class MarathonHealthCheckManager @Inject() (
...
+            activeHealthCheck <- activeHealthChecks
+          } remove(appId, version, activeHealthCheck.healthCheck)
+
+          ahcs
         }
 
         // add missing health checks for the current
         // reconcile all running versions of the current app
-        val res = versionTasksMap.keys map { version =>
+        val healthCheckVersions: Set[String] = ahcs(appId).iterator.map(_._1.toString).toSet
```
kolloch: this ahcs access must be inside the ahcs block

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -115,30 +115,41 @@ class MarathonHealthCheckManager @Inject() (
...
+        val ahcs = appHealthChecks.writeLock { ahcs =>
+          // remove health checks for which the app version is not current and no tasks remain
+          // since only current version tasks are launched.
+          for {
+            (version, activeHealthChecks) <- ahcs(appId)
+            if version != app.version && !activeVersions.contains(version.toString)
+            activeHealthCheck <- activeHealthChecks
+          } remove(appId, version, activeHealthCheck.healthCheck)
+
+          ahcs
```
kolloch: don't export ahcs. that's the whole point of the lock.

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -115,30 +115,40 @@ class MarathonHealthCheckManager @Inject() (
...
-          activeHealthCheck <- activeHealthChecks
-        } {
-          remove(appId, version, activeHealthCheck.healthCheck)
+      case Some(app) =>
+        log.info(s"reconcile [$appId] with latest version [${app.version}]")
+
+        val tasks: Iterable[MarathonTask] = taskTracker.getTasks(app.id)
+        val activeAppVersions: Set[String] = tasks.iterator.map(_.getVersion).toSet + app.version.toString
+
+        val healthCheckAppVersions: Set[String] = appHealthChecks.writeLock { ahcs =>
```
aquamatthias: This not only computes the `healthCheckAppVersions` but also removes health checks. Please extract the remove operation.

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -115,30 +115,40 @@ class MarathonHealthCheckManager @Inject() (
...
+        val appVersionsWithoutHealthChecks: Set[String] = activeAppVersions -- healthCheckAppVersions
+        val res: Iterator[Future[Unit]] = appVersionsWithoutHealthChecks.iterator map { version =>
           appRepository.app(app.id, Timestamp(version)) map {
-            case None             =>
-            case Some(appVersion) => addAllFor(appVersion)
+            case None =>
+              // FIXME: If the app version of the task is not available anymore, no health check is started.
+              // We generated a new app version for every scale change. If maxVersions is configured, we
+              // throw away old versions such that we may not have the app configuration of all tasks available anymore.
+              log.warn(s"Cannot find health check configuration for [$appId] and version [$version]")
```
aquamatthias: Thanks for adding a log statement here....

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -115,30 +115,40 @@ class MarathonHealthCheckManager @Inject() (
...
+            case Some(appVersion) =>
+              log.info(s"addAllFor [$appId] version [$version]")
+              addAllFor(appVersion)
```
aquamatthias: You remove HealthChecks in a write lock. I missed it for the add operation ans saw it is done in the addAllFor operation. Perhaps make it congruent?

src/test/scala/mesosphere/marathon/health/MarathonHealthCheckManagerTest.scala
```diff
@@ -256,33 +261,85 @@ class MarathonHealthCheckManagerTest extends MarathonSpec with Logging {
...
+  private[this] def captureEvents(block: => Unit): Seq[MarathonEvent] = {
+    // yes, this is ugly. Since we only access it in the actor until it terminates, we do have
+    // the correct thread sync boundaries in place.
+
+    var capture = Vector.empty[MarathonEvent]
+    val captureEventsActor = actor {
+      new Act {
+        become {
+          case captureMe: MarathonEvent =>
+            println(s"event: $captureMe")
```
aquamatthias: Left over? Please remove.

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -115,30 +115,40 @@ class MarathonHealthCheckManager @Inject() (
...
 
         // add missing health checks for the current
         // reconcile all running versions of the current app
-        val res = versionTasksMap.keys map { version =>
+        val appVersionsWithoutHealthChecks: Set[String] = activeAppVersions -- healthCheckAppVersions
+        val res: Iterator[Future[Unit]] = appVersionsWithoutHealthChecks.iterator map { version =>
           appRepository.app(app.id, Timestamp(version)) map {
-            case None             =>
-            case Some(appVersion) => addAllFor(appVersion)
+            case None =>
```
aquamatthias: Do you think it would make sense to use the latest app version here?
It is guessing, but probably better than doing nothing?

aquamatthias:
LGTM. Some minor styling issues.

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
kolloch: it's hard to do it better without many changes.

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
kolloch: I know that it is inconsistent. The lock wasn't active before my changes because this is executed in a future. In fact, that lock could have led to extra contention.

It's hard to do it better without many changes. So I would leave it as is following our discussion.

src/test/scala/mesosphere/marathon/health/MarathonHealthCheckManagerTest.scala
kolloch: ack

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
kolloch: I wouldn't do that in this changeset but feel free to create an issue and I'll do it next.

kolloch:
Ok, I removed the `println` from the test code.

aquamatthias:
Cool.

PR #2572, opened 2015-11-05
Remove servicerouter and point docs to marathon-lb

discordianfish:
I've extracted the servicerouter and wrapped it in a Docker container which can easily be deployed on top of marathon and DCOS. To avoid fragmentation we should now remove the servicerouter.py from this repository. The script can still be used directly, so nothing (beside where to find it) should change for existing users: https://github.com/mesosphere/marathon-lb/blob/master/marathon-update-haproxy.py

I also tried my best to keep the scripts commit history, just in case someone might be worried about their contributions not attributed: https://github.com/mesosphere/marathon-lb/commits/master?page=3


aquamatthias:
@discordianfish thanks! We probably should also remove the haproxy-marathon-bridge  and point to marathon-lb.

discordianfish:
@aquamatthias Yes, I've discussed this with @spahl and @tparekh73 - we'll change all our examples to use marathon-lb and then we can remove all the other marathon LB scripts.

PR #2590, opened 2015-11-09
Simplify awk

hayderimran7:
fix the comments about variables output, the format is with --xx yy . Also avoid for loop in favor of known variable to be printed ; here its $2 so simple  print $2 would suffice.

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -35,7 +35,7 @@ require_arg() {
 }
 
 # Filter out MARATHON_APP* env variables. All MARATHON_XXX=yyy variables are interpreted as --xxx yyy application parameter
-for env_op in `env | grep -v ^MARATHON_APP | grep ^MARATHON_ | awk '{gsub(/MARATHON_/,""); gsub(/=/," "); printf("%s%s ", "--", tolower($1)); for(i=2;i<=NF;i++){printf("%s ", $i)}}'`; do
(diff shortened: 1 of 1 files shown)


aquamatthias:
@lloesche can you review please?

hayderimran7:
a comparison:
```
export MARATHON_IMR=imr
export MARATHON_MANI=mani
 export MARATHON_HAYDER=hayder
```

Old:
```
 for env_op in `env | grep -v ^MARATHON_APP | grep ^MARATHON_ | awk '{gsub(/MARATHON_/,""); gsub(/=/," "); printf("%s%s ", "--", tolower($1)); for(i=2;i<=NF;i++){printf("%s ", $i)}}'`; do echo $env_op; done

--hayder
hayder
--mani
mani
--imr
imr
```

New:
```
for env_op in `env | grep -v ^MARATHON_APP | grep ^MARATHON_ | awk '{gsub(/MARATHON_/,""); gsub(/=/," "); printf("%s%s ", "--", tolower($1)); print $2}'`; do echo $env_op; done

--hayder
hayder
--mani
mani
--imr
imr
```

lloesche:
Hi, thanks for the PR!

I don't think the rest of the code allows args with spaces anyways, but just to maintain the current behaviour I'd `$1=""; print $0` instead of printing $2.
Otherwise the new awk code breaks for things like
```
export MARATHON_foo="some arg with spaces"
```
which the old awk code handles correctly.

aquamatthias:
Hey @hayderimran7 any updates here?

hayderimran7:
@aquamatthias  thanks . i have updated PR with proposed change and it takes care of args with spaces.

PR #2595, opened 2015-11-09
Improves #2440 - unit tests for PostToEventStreamStepImpl

kolloch:


kolloch:
Added `  test("process running notification of missing task") { ...` to achieve 100% coverage of that class.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/PostToEventStreamStepImpl.scala
```diff
@@ -38,22 +38,34 @@ class PostToEventStreamStepImpl @Inject() (
       case TASK_ERROR | TASK_FAILED | TASK_FINISHED | TASK_KILLED | TASK_LOST =>
         // Remove from our internal list
         maybeTask match {
-          case Some(task) => postEvent(appId, status, task)
-          case None       => log.warn(s"Task not found. Do not post event for '{}'", taskId.getValue)
+          case Some(task) =>
+            postEvent(timestamp, appId, status, task)
+          case None =>
+            log.warn(s"Received terminal task event but task not found. " +
```
meichstedt: There is no substitution on this first line – no need to use String interpolation.
I've talked with our _Onsite English Expert_ and came up with a neater (and shorter) message:
`log.warn(s"Received TASK_RUNNING for unknown task [${taskId}] of [$appId]. Not posting event.")`

Btw: the `taskId` is only used for logging, so you can as well initialize it with `val taskId = status.getTaskId.getValue` – then the messages fit in one line. Yay ;)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/PostToEventStreamStepImpl.scala
```diff
@@ -38,22 +38,34 @@ class PostToEventStreamStepImpl @Inject() (
...
+              s"Do not post event for [${taskId.getValue}] of app [$appId].")
         }
 
       case TASK_RUNNING if !maybeTask.exists(_.hasStartedAt) => // staged, not running
-        maybeTask.foreach { task => postEvent(appId, status, task) }
+        maybeTask match {
+          case Some(task) =>
+            postEvent(timestamp, appId, status, task)
+          case None =>
+            log.warn(s"Received TASK_RUNNING event but task not found. " +
```
meichstedt: See my comment regarding the log message above.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/steps/PostToEventStreamStepImplTest.scala
```diff
@@ -0,0 +1,226 @@
...
+        timestamp = updateTimestamp,
+        appId = appId,
+        maybeTask = existingTask,
+        status = status
+      )
+    }
+
+    Then("no event gets posted")
+    events should be (empty)
+    And("only sending event info gets logged")
```
meichstedt: And("a warning is logged for the unknown task")

src/test/scala/mesosphere/marathon/core/task/tracker/impl/steps/PostToEventStreamStepImplTest.scala
```diff
@@ -0,0 +1,226 @@
...
+  private[this] val stagedMarathonTask =
+    MarathonTask
+      .newBuilder()
+      .setId(taskId.getValue)
+      .setHost(host)
+      .addAllPorts(portsList.asJava)
+      .setVersion(version.toString)
+      .build()
+
+  private[this] val runningMarathonTask = stagedMarathonTask.toBuilder.setStartedAt(2).build()
```
meichstedt: Unused val

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/PostToEventStreamStepImpl.scala
kolloch: Thx! Ack.

kolloch: Refactored it a bit.

In the short future, that logging statement becomes defacto unreachable but it's probably ok to leave it in there for robustness.

kolloch:
updated the PR according to your suggestions.

src/test/scala/mesosphere/marathon/health/MarathonHealthCheckManagerTest.scala
```diff
@@ -306,38 +305,5 @@ class MarathonHealthCheckManagerTest extends MarathonSpec with Logging {
...
-      eventStream.unsubscribe(captureEventsActor)
-      captureEventsActor ! PoisonPill
-      val probe = TestProbe()
-      probe.watch(captureEventsActor)
-      probe.expectMsgClass(classOf[Terminated])
-    }
-
-    capture
-  }
+  def captureEvents = CaptureEvents(eventStream)
```
meichstedt: Not sure I like this application of a case class. There is actually no pattern matching etc required or done, and imho the type of the def should be a function type – currently, it's implicitly referencing the apply method and its type is `CaptureEvents` while it should be `block: => Unit`.

Event though something like
`eventCaptor.capture {…}`
looks less elegant than your solution, I'd prefer this is production code. I can live with that in test code, just wanted to mention it :palm_tree:

meichstedt:
LGTM, please address the above comments and I'll merge.

src/test/scala/mesosphere/marathon/health/MarathonHealthCheckManagerTest.scala
kolloch: I like your suggestion for the `capture` method. Making it less concise but clearer is fine with me. I'll do it in a follow up patch.

I like `new CaptureEvents` less than `CaptureEvents` and case classes give me the object apply method free-of-charge. I don't care much for unused auto-generated code but that's a personal preference. So what's your argument there? Is the use of a case class misleading or do you have performance gripes?

meichstedt: No, performance was not bothering me here at all. I completely understand that you like the syntactic sugar a case class comes with; I'd just prefer having a concise syntax where I can see the type of what I'm supposed to use. The way you implemented it is nice and dotless – however, one has to _know_ that `CaptureEvents` is a case class, and that a reference to it can therefore be used to _implicitly_ call the apply method. I can imagine thinking differently about that tomorrow; let's just leave it like that.

PR #2597, opened 2015-11-09
Use EntityCacheStore for EventSubscribers

aquamatthias:
Cherry pick for 0.13


kolloch:
Can we somehow learn from this mistake and prevent something like that happening again in the future?

Otherwise: LGTM!

aquamatthias:
Guess the main reason for that was the fact, that all repos were created in one place, but not this one. Do you have a suggestion?

kolloch:
cherry picked to releases/0.13

kolloch:
probably you are right. That's sufficient.

I was thinking along the lines of checking whether there is an entityStore bound by guice that's not cached, making the uncached EntityStore package private, stuff like that.

PR #2598, opened 2015-11-09
Improves #2440 - Test for NotifyLaunchQueueStepImpl

kolloch:
...also uses futureValue to wait for the Future returned by processUpdate
even if the implementation is synchronous (more robust blackbox testing).

...also refactor the CaptureEvents code a bit.


meichstedt:
:+1:

PR #2604, opened 2015-11-10
Fixes #2603 by defining log templates and binding variables.

### LINKED ISSUE #2603: ActorLogging warning does not expand Throwable
`log.warning(message, throwable)` is not handled the way other logging frameworks handle that and lead to wrong log messages like: `WARNING arguments left: 1`

aquamatthias:
After merge cherry pink on master.

### DIFF
--- a/src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
+++ b/src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
@@ -2,16 +2,15 @@ package mesosphere.marathon.event.http
 
 import akka.actor._
 import akka.pattern.ask
+import mesosphere.marathon.api.v2.json.Formats._
 import mesosphere.marathon.event._
 import mesosphere.marathon.event.http.SubscribersKeeperActor.GetSubscribers
-import play.api.libs.json.{ JsValue, Json }
-import spray.client.pipelining.{ sendReceive, _ }
-import spray.http.{ HttpRequest, HttpResponse }
+import spray.client.pipelining.{sendReceive, _}
+import spray.http.{HttpRequest, HttpResponse}
 import spray.httpx.PlayJsonSupport
-import mesosphere.marathon.api.v2.json.Formats._
 
 import scala.concurrent.Future
-import scala.util.{ Failure, Success }
+import scala.util.{Failure, Success}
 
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -45,10 +44,9 @@ class HttpEventActor(val subscribersKeeper: ActorRef) extends Actor with ActorLo
     response.onComplete {
       case Success(res) =>
         if (res.status.isFailure)
-          log.warning(s"Failed to post $event to $urlString")
-
+          log.warning(s"Post {} to {} resulted in error response {}", event, urlString, res.status)
```
meichstedt: I suppose your point was to replace scala string interpolation with the log.warning varargs approach – you missed removing the string interpolation prefix that will lead to a call to `StringContext.s()``

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamHandleActor.scala
```diff
@@ -76,11 +76,11 @@ class HttpEventStreamHandleActor(
       //Do not act any longer on any event.
       context.become(Actor.emptyBehavior)
     case _ =>
-      log.warning(s"Could not send message to $handle", ex)
+      log.warning("Could not send message to {} reason: {}", handle, ex)
   }
 
   private[this] def dropEvent(event: MarathonEvent): Unit = {
-    log.warning(s"Ignore event $event for handle $handle (slow consumer)")
+    log.warning("Ignore event {} for handle {} (slow consumer)", event, handle)
```
meichstedt: Is there a specific reason you chose varargs substitution over the scala way? As far as I can see, doing it this way and adding another arg to the list will still result in a `WARNING arguments left: 1` error.

meichstedt:
See my comments above. I am not sure, whether we should use varargs here or go for scala string interpolation in general – I think we use of string interpolation mor often than not, so we might do here as well?

src/main/scala/mesosphere/marathon/event/http/HttpEventStreamHandleActor.scala
aquamatthias: Looked like the default usage in our code base. 
My preference would be to use a clever scala logging system, where we don't have to decide which way to use...

PR #2613, opened 2015-11-10
Fixes #2440 - Finish testing and refactoring task update steps

### LINKED ISSUE #2440: Create unit tests for PR #2381

kolloch:
* Kill unknown tasks in processor so that TaskUpdateSteps only get executed on known tasks.
* Also handle acknowledgement directly in processor.
* Simplify step logic, especially for the updateTaskTracker step.
* Create UpdateTaskTrackerStepImplTest.
* Ensure that the TaskStatusUpdateProcessorImplTest uses all steps specified in Guice module.


src/test/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImplTest.scala
```diff
@@ -0,0 +1,201 @@
...
+import mesosphere.marathon.test.Mockito
+import org.apache.mesos.Protos.{ SlaveID, TaskState, TaskStatus }
+import org.apache.mesos.SchedulerDriver
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.{ FunSuite, GivenWhenThen, Matchers }
+
+import scala.collection.JavaConverters._
+import scala.concurrent.Future
+
+class UpdateTaskTrackerStepImplTest extends FunSuite with Matchers with ScalaFutures with Mockito with GivenWhenThen {
```
kolloch: After writing this test I agreed with @meichstedt that we want to simplify the logic of the `UpdateTaskTrackerStepImpl`. Thus, I am not sure if it is really worth a lot of time reviewing this test ;)

src/main/scala/mesosphere/marathon/core/task/tracker/TaskStatusUpdateStep.scala
```diff
@@ -15,5 +15,5 @@ trait TaskStatusUpdateStep {
   def name: String
 
   def processUpdate(
-    timestamp: Timestamp, appId: PathId, maybeTask: Option[MarathonTask], mesosStatus: TaskStatus): Future[_]
+    timestamp: Timestamp, appId: PathId, task: MarathonTask, mesosStatus: TaskStatus): Future[_]
```
aquamatthias: Definitively an enhancement to not think about non existing tasks in every step!

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -43,24 +48,41 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
+        }
+    }
+  }
+
+  private[this] def acknowledge(taskStatus: TaskStatus): Unit = {
+    driverHolder.driver.foreach(_.acknowledgeStatusUpdate(taskStatus))
+  }
+
+  private[this] def killTask(taskId: TaskID): Unit = {
+    driverHolder.driver.foreach(_.killTask(taskId))
```
aquamatthias: Not related to this PR: could the DriverHolder implement the Scheduler interface by delegating to the real driver or handling the non existence? I see a lot of noise by the optional driver and the handling of non existence driver...

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
```diff
@@ -18,13 +18,21 @@ class ContinueOnErrorStep(wrapped: TaskStatusUpdateStep) extends TaskStatusUpdat
...
   override def processUpdate(
-    timestamp: Timestamp, appId: PathId, maybeTask: Option[MarathonTask], mesosStatus: TaskStatus): Future[_] = {
+    timestamp: Timestamp, appId: PathId, task: MarathonTask, mesosStatus: TaskStatus): Future[_] = {
 
     import scala.concurrent.ExecutionContext.Implicits.global
-    wrapped.processUpdate(timestamp, appId, maybeTask, mesosStatus).recover {
-      case NonFatal(e) =>
-        log.error("while executing step {} for [{}], continue with other steps",
-          wrapped.name, mesosStatus.getTaskId.getValue, e)
+    val maybeProcessed: Option[Future[_]] = Option(wrapped.processUpdate(timestamp, appId, task, mesosStatus))
```
aquamatthias: why should a processor return null instead of a future?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
```diff
@@ -18,13 +18,21 @@ class ContinueOnErrorStep(wrapped: TaskStatusUpdateStep) extends TaskStatusUpdat
...
+    maybeProcessed match {
+      case Some(processed) =>
+        processed.recover {
+          case NonFatal(e) =>
+            log.error("while executing step {} for [{}], continue with other steps",
+              wrapped.name, mesosStatus.getTaskId.getValue, e)
+        }
+      case None =>
```
aquamatthias: why do we handle this explicitly?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/NotifyRateLimiterStepImpl.scala
```diff
@@ -18,35 +18,34 @@ class NotifyRateLimiterStepImpl @Inject() (
...
-      case _ =>
-        Future.successful(())
+    task: MarathonTask): Future[_] = {
+    if (status.getState != TaskState.TASK_KILLED) {
+      import scala.concurrent.ExecutionContext.Implicits.global
+      appRepository.app(appId, Timestamp(task.getVersion)).map { maybeApp =>
+        // It would be nice if we could make sure that the delay gets send
+        // to the AppTaskLauncherActor before we continue but that would require quite some work.
+        //
+        // In production, it should not matter if we do not delay some task launches so much.
```
aquamatthias: The delay is a result of an effect that happened in the past. The message abut that effect also takes time. I am tempted to say: please remove this comment, because we react as prompt as possible...

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
```diff
@@ -1,65 +1,37 @@
...
-    marathonSchedulerDriverHolder: MarathonSchedulerDriverHolder) extends TaskStatusUpdateStep {
-
-  private[this] val log = LoggerFactory.getLogger(getClass)
+    taskTracker: TaskTracker) extends TaskStatusUpdateStep {
 
   def name: String = "updateTaskTracker"
 
-  //TODO: fix style issue and enable this scalastyle check
-  //scalastyle:off cyclomatic.complexity method.length
   def processUpdate(
```
aquamatthias: Removed the cyclomatic complexity! +1
And I have to admit scalastyle is right - it reads a lot easier...

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
```diff
@@ -1,65 +1,37 @@
...
-      case TASK_RUNNING if !maybeTask.exists(_.hasStartedAt) => // staged, not running
+      case TASK_RUNNING if !task.hasStartedAt => // staged, not running
         taskTracker.running(appId, status)
-          .recover {
-            case NonFatal(t) =>
-              log.warn(s"Task could not be saved: '${taskId.getValue}'", t)
-              driverOpt.foreach(_.killTask(status.getTaskId))
-          }
-
-      case TASK_STAGING if !taskTracker.contains(appId) =>
```
aquamatthias: why can we ignore this case?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
```diff
@@ -1,65 +1,37 @@
...
     import org.apache.mesos.Protos.TaskState._
-
-    import scala.concurrent.ExecutionContext.Implicits.global
     status.getState match {
       case TASK_ERROR | TASK_FAILED | TASK_FINISHED | TASK_KILLED | TASK_LOST =>
         // Remove from our internal list
         taskTracker.terminated(appId, taskId.getValue)
 
-      case TASK_RUNNING if !maybeTask.exists(_.hasStartedAt) => // staged, not running
+      case TASK_RUNNING if !task.hasStartedAt => // staged, not running
```
aquamatthias: please change comment to // was staged and is now running

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
```diff
@@ -1,65 +1,37 @@
...
         taskTracker.running(appId, status)
-          .recover {
-            case NonFatal(t) =>
-              log.warn(s"Task could not be saved: '${taskId.getValue}'", t)
```
aquamatthias: If we get the same tasks again during reconciliation - don't we have the same logic: kill the task?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStatusUpdateProcessorImplTest.scala
```diff
@@ -6,11 +6,13 @@ import akka.testkit.TestProbe
...
 import mesosphere.marathon.MarathonSchedulerActor.ScaleApp
 import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.core.CoreGuiceModule
 import mesosphere.marathon.core.base.ConstantClock
 import mesosphere.marathon.core.launchqueue.LaunchQueue
+import mesosphere.marathon.core.task.bus.TaskStatusObservables.TaskStatusUpdate
 import mesosphere.marathon.core.task.bus.{ TaskStatusEmitter, TaskStatusUpdateTestHelper }
 import mesosphere.marathon.core.task.tracker.TaskStatusUpdateProcessor
-import mesosphere.marathon.core.task.tracker.impl.steps.{ AcknowledgeTaskUpdateStepImpl, NotifyHealthCheckManagerStepImpl, NotifyRateLimiterStepImpl, PostToEventStreamStepImpl, ScaleAppUpdateStepImpl, TaskStatusEmitterPublishStepImpl, UpdateTaskTrackerStepImpl }
+import mesosphere.marathon.core.task.tracker.impl.steps._
```
aquamatthias: thanks!

src/test/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImplTest.scala
aquamatthias: I thought the original task was to write tests for the new logic. I see changed logic, removed tests and the only added test is not for review...
Do you want me to merge or is this just WIP?

aquamatthias:
Travis and TC builds are red. Does not look like problems of this PR. I restarted the tests - let's see.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
meichstedt: That's only happened in a test where the appropriate result was not mocked

aquamatthias: Then please fix the test.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
meichstedt: again, tests and it does not make the code less robust. Actually the opposite ;)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/NotifyRateLimiterStepImpl.scala
meichstedt: The worst case would be that we restart one or few tasks without delay – this is unlikely but possible.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
meichstedt: Thx! Good suggestion :+1:

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
meichstedt: Killing unknown tasks has moved to processor. See commit/pr description.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
meichstedt: Killing unknown tasks has moved to processor. See commit/pr description.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStatusUpdateProcessorImplTest.scala
meichstedt: that was an accident. I actually don't like it. Code might break by adding classes => horror. But we'll leave it as is ;)

src/test/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImplTest.scala
meichstedt: Well, testing revealed deficiencies that we were able to remove. This is not WIP, this should work in production. I should not have written the comment.

If you insist, we can create a new issue for this.

meichstedt:
@aquamatthias We incorporated your feedback and added changes in a separate commit. We can squash after you reviewed that.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
aquamatthias: I hope we don't start the Nullpointer discussion. Let's make our code never return null, so the compiler can reason about.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
aquamatthias: OK

aquamatthias:
Two points were not addressed. Otherwise LGTM. Please address and squash.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
meichstedt: We fixed the test. But if we have a faulty test or faulty production code we will get better error feedback.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/ContinueOnErrorStep.scala
meichstedt: We are using a test framework that leads to returning "null" values when a call is not mocked. When the code changes and the test fails, not handling this case explicitly led to a situation that was hard to debug. This code makes it easy.

In addition, the code does not make the production code less robust -- it actually gets more robust.

If you really insist, we'll remove the code but we both would prefer to keep it.

kolloch:
As discussed, we did not remove the Option logic from ContinueOnError but squashed and rebased the pull request.

aquamatthias:
I really don't like the intent of this, but I do not insist.
For amusement: http://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare
Otherwise looks good.

kolloch:
We are not in favor of null pointer references but the tools that we use for testing use them as default.

PR #2625, opened 2015-11-12
Fixes #2618 - Consolidate statusUpdate logic in TaskTracker life cycle methods

### LINKED ISSUE #2618: Consolidate task update logic in TaskTracker
In #2440 we noticed that the code in `UpdateTaskTrackerStepImpl` duplicates the structure of the running and taskUpdate methods in TaskTracker. This led to fragile interaction between those classes.

We want to simplify this design and consolidate the task update logic completely in the TaskTracker.

kolloch:
* Added some documentation and logging.
* Move lifecycle hooks next to each other
* Assert that task that is passed to TaskTracker.created does not exist yet.
* Separate processUpdate method from statusUpdate to make scalastyle happy.
* Make methods only used by processUpdate method private.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
@@ -19,19 +19,7 @@ class UpdateTaskTrackerStepImpl @Inject() (
 
   def processUpdate(
     timestamp: Timestamp, appId: PathId, task: MarathonTask, status: TaskStatus): Future[_] = {
-    val taskId = status.getTaskId
 
-    import org.apache.mesos.Protos.TaskState._
-    status.getState match {
-      case TASK_ERROR | TASK_FAILED | TASK_FINISHED | TASK_KILLED | TASK_LOST =>
-        // Remove from our internal list
-        taskTracker.terminated(appId, taskId.getValue)
-
-      case TASK_RUNNING if !task.hasStartedAt => // was staged and is now running
-        taskTracker.running(appId, status)
-
-      case _ =>
-        taskTracker.statusUpdate(appId, status)
-    }
+    taskTracker.statusUpdate(appId, status)
   }
 }
--- a/src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
@@ -5,6 +5,7 @@ import javax.inject.Inject
 import mesosphere.marathon.MarathonConf
 import mesosphere.marathon.Protos._
 import mesosphere.marathon.state.{ PathId, TaskRepository, Timestamp }
+import org.apache.mesos.Protos.TaskState._
 import org.apache.mesos.Protos.{ TaskState, TaskStatus }
 import org.slf4j.LoggerFactory
 
@@ -13,6 +14,23 @@ import scala.collection.concurrent.TrieMap
 import scala.collection.immutable.Set
 import scala.concurrent.{ Await, Future }
 
+/**
+  * The TaskTracker contains the latest known state for every task.
+  *
+  * It also processes new status information as sent by Mesos (see taskUpdate)
+  * and persist it in the associated task state.
+  *
+  * Known deficiencies:
+  *
+  * * There are race conditions in this code involving the synchronization of the repo and the cachedApps
+  *   data structure.
+  * * There are probably race conditions in checking whether to remove the entry for an app.
+  * * Some methods such as list only look into the cached tasks. If the first reconciliation has not yet
+  *   happened, this might not contain all apps. Could be fixed by preloading all tasks on election.
+  * * expungeOrphanedTasks is probably not a good idea. See issue #2620.
+  * * Some methods or data structures use Set[MarathonTask] which leads to performance problems because
+  *   hashCode on MarathonTask is relatively expensive.
+  */
 class TaskTracker @Inject() (
     repo: TaskRepository,
     config: MarathonConf) {
@@ -24,13 +42,14 @@ class TaskTracker @Inject() (
 
   private[this] val log = LoggerFactory.getLogger(getClass.getName)
 
-  private[tasks] val apps = TrieMap[PathId, InternalApp]()
+  private[tasks] val cachedApps = TrieMap[PathId, InternalApp]()
 
+  // FIXME: remove, see documented deficiencies above
   def get(appId: PathId): Set[MarathonTask] =
     getTasks(appId).toSet
 
   def getTasks(appId: PathId): Iterable[MarathonTask] = {
-    getInternal(appId).values
+    getTaskMap(appId).values
   }
 
   def getVersion(appId: PathId, taskId: String): Option[Timestamp] =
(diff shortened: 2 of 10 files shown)


src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -5,6 +5,7 @@ import javax.inject.Inject
 import mesosphere.marathon.MarathonConf
```
aquamatthias: We put a lot of effort into TaskTracker recently. Does it make sense to extract a trait and move it into core/*?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -50,60 +51,102 @@ class TaskTracker @Inject() (
...
-          .build
-        getInternal(appId) += (task.getId -> task)
-        store(appId, task).map(_ => task)
-
-      case Some(oldTask) => // running
-        val msg = s"Task for ID $taskId already running, ignoring"
-        log.warn(msg)
-        Future.failed(new Exception(msg))
+    val maybeAppTask: Option[(InternalApp, MarathonTask)] =
+      for { app <- apps.get(appId); task <- app.tasks.get(taskId) } yield (app, task)
```
aquamatthias: Use getInternal instead of apps?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -50,60 +51,102 @@ class TaskTracker @Inject() (
...
+    }
+  }
+
+  /**
+    * Remove the task for the given app with the given ID completely.
+    *
+    * If the task does not exist, the method will not fail.
+    */
+  def terminated(appId: PathId, taskId: String): Future[_] = {
+    val maybeInternalApp = apps.get(appId)
```
aquamatthias: Probably does not happen. But why no use getInternal instead of apps?

kolloch: Because getInternal does create the InternalApp if it does not exist yet.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -50,60 +51,102 @@ class TaskTracker @Inject() (
...
+
+    Await.result(repo.expunge(taskId), timeout)
+
+    log.info(s"Task $taskId expunged and removed from TaskTracker")
+
+    if (app.shutdown && app.tasks.isEmpty) {
+      // Are we shutting down this app? If so, remove it
+      remove(app.appName)
     }
   }
```
aquamatthias: shutdown: use getInternal instead of apps?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: It is still full of race conditions, unfortunately.

But the steps of refactorings make it easier to eventually replace it (clearer scope in the functionality).

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: We do not want to create an internal app if it doesn't exist yet.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: And what does not happen?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: see above.

kolloch:
@meichstedt and me worked on this BTW.

I think, I addressed your comments by comments (which is only fair ;) ).

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
aquamatthias: Ok.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
aquamatthias: I probably do not understand the concept of apps here: 
It looks like a lazily loaded cache, that is backed by a store.
If I get a status update why would I not want to load that app, if it is not in the cache?

aquamatthias: Furthermore: please rename apps and getInternal to reflect the real meaning.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
aquamatthias: See comment above.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
aquamatthias: Why: `apps.getOrElseUpdate(appId, fetchApp(appId))` is also done in getInternal. Probably with some purpose?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
```diff
@@ -19,19 +19,7 @@ class UpdateTaskTrackerStepImpl @Inject() (
...
-        // Remove from our internal list
-        taskTracker.terminated(appId, taskId.getValue)
-
-      case TASK_RUNNING if !task.hasStartedAt => // was staged and is now running
-        taskTracker.running(appId, status)
-
-      case _ =>
-        taskTracker.statusUpdate(appId, status)
-    }
+    taskTracker.statusUpdate(appId, status)
```
aquamatthias: Cool

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -5,6 +5,7 @@ import javax.inject.Inject
 import mesosphere.marathon.MarathonConf
 import mesosphere.marathon.Protos._
 import mesosphere.marathon.state.{ PathId, TaskRepository, Timestamp }
+import org.apache.mesos.Protos.TaskState._
 import org.apache.mesos.Protos.{ TaskState, TaskStatus }
 import org.slf4j.LoggerFactory
```
aquamatthias: Should the TaskTracker also be prepared during onElected hook?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
aquamatthias: Obviously there was (is?) logic related to task tracking outside of TaskTracker. Can you make a note in TaskTracker, what logic is in/out of TaskTracker?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -117,26 +160,6 @@ class TaskTracker @Inject() (
     log.warn(s"App $appId removed from TaskTracker")
```
aquamatthias: Why should we warn here?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
aquamatthias: For example: killing an unknown task is not in scope of the TaskTracker. Why?

aquamatthias:
TaskTracker becomes more focused with this PR. 
See my comments.
Additional things that come to my mind reviewing the code:

- there are still a lot of Awaits in the code
- didn't we agree to remove expungeOrphanedTasks?
- TaskTracker.get should be removed
- TaskTracker.fetchTask should be removed
- TaskTracker.list returns a Map
- TaskTracker.shutdown should be renamed (sounds like system shutdown)
- why do we have App and InternalApp?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: As described in the documentation of this method, taskUpdate will fail if the task does not exist yet. So it should never create an InternalApp or Task. In fact, it can't because it wouldn't have all necessary information.

Since we did neither introduce nor change apps and getInternal in this patch, I would like to do that in another patch (but I'll take care of that).

How should we rename `apps` and `getInternal`? 

* `apps` -> `appTasks` ?
* `getInternal` -> `getOrCreateAppTasks` ?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: I do not get "comment above". Please elaborate or explain it to me personally.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: but that would create the InternalApp! We do not want that if we want to remove a task.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: Its not part of this patch. Feel free to suggest a change outside of this PR.

kolloch:
@aquamatthias: I wish to keep this pull request focussed and not overload it with a complete refactoring of the `TaskTracker`. I think that we benefit from small incremental steps.

Furthermore, I did not make the design choices that went to the original creation of the TaskTracker. So please ask the original authors.

This is not the pull request to make the TaskTracker perfect.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: Aaaah, @aquamatthias you made the comment above the method. github does not display the relevant code snippet then. Now I get you. `shutdown` should not use `getOrElseUpdate`. But since we did not touch this code, I typically would't want to change this code in this PR. But I will...

kolloch:
If have updated the PR.

kolloch:
Thanks for pointing out some crucial points! I didn't understand them at once -- therefore the long-winded discussions...

src/main/scala/mesosphere/marathon/core/task/tracker/impl/steps/UpdateTaskTrackerStepImpl.scala
kolloch: The code in the taskTracker mostly duplicated the branches in this method. That was silly. In theory, the TaskTracker should guarantee the atomicity of processing these updates so it made sense to move the logic the the TaskTracker.

Killing an unknown task does not relate to updating the state of a task, thus it is handled outside of the TaskTracker. Furthermore, it would introduce a dependency from the TaskTracker to the MesosSchedulerDriver.

aquamatthias:
@kolloch I know this is pretty awful code here - I think FixMe's are a first good step, because they make you think. We need to get rid of those urgently...

aquamatthias:
But for this PR - thanks!

PR #2632, opened 2015-11-13
Fix flaky unit test

gkleiman:
This should fix https://travis-ci.org/mesosphere/marathon/builds/90919136

### DIFF
--- a/src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
+++ b/src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
@@ -24,6 +24,30 @@ class TaskBuilderTest extends MarathonSpec {
   import mesosphere.mesos.protos.Implicits._
 
   test("BuildIfMatches") {
+    val offer = makeBasicOffer(cpus = 1.0, mem = 128.0, disk = 2000.0, beginPort = 31000, endPort = 32000).build
+
+    val task: Option[(TaskInfo, Seq[Long])] = buildIfMatches(
+      offer,
+      AppDefinition(
+        id = "/product/frontend".toPath,
+        cmd = Some("foo"),
+        cpus = 1.0,
+        mem = 64.0,
+        disk = 1.0,
+        executor = "//cmd",
+        ports = Seq(8080, 8081)
+      )
+    )
+
+    assert(task.isDefined)
+
+    val (taskInfo, taskPorts) = task.get
+    assertTaskInfo(taskInfo, taskPorts, offer)
+
+    assert(!taskInfo.hasLabels)
+  }
+
+  test("BuildIfMatches works with duplicated resources") {
     val offer = makeBasicOffer(cpus = 1.0, mem = 128.0, disk = 2000.0, beginPort = 31000, endPort = 32000)
       .addResources(ScalarResource("cpus", 1))
       .addResources(ScalarResource("mem", 128))
@@ -132,11 +156,7 @@ class TaskBuilderTest extends MarathonSpec {
   }
 
   test("BuildIfMatchesWithLabels") {
-    val offer = makeBasicOffer(cpus = 1.0, mem = 128.0, disk = 2000.0, beginPort = 31000, endPort = 32000)
-      .addResources(ScalarResource("cpus", 1))
-      .addResources(ScalarResource("mem", 128))
-      .addResources(ScalarResource("disk", 2000))
-      .build
+    val offer = makeBasicOffer(cpus = 1.0, mem = 128.0, disk = 2000.0, beginPort = 31000, endPort = 32000).build
 
     val labels = Map("foo" -> "bar", "test" -> "test")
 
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -24,11 +24,7 @@ class TaskBuilderTest extends MarathonSpec {
   import mesosphere.mesos.protos.Implicits._
 
   test("BuildIfMatches") {
-    val offer = makeBasicOffer(cpus = 1.0, mem = 128.0, disk = 2000.0, beginPort = 31000, endPort = 32000)
-      .addResources(ScalarResource("cpus", 1))
-      .addResources(ScalarResource("mem", 128))
-      .addResources(ScalarResource("disk", 2000))
-      .build
+    val offer = makeBasicOffer(cpus = 1.0, mem = 128.0, disk = 2000.0, beginPort = 31000, endPort = 32000).build
```
kolloch: This is just an unrelated cleanup, right?

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -860,8 +856,8 @@ class TaskBuilderTest extends MarathonSpec {
     // The taskName is the elements of the path, reversed, and joined by dots
     assert("frontend.product" == taskInfo.getName)
     assert(2 == taskPorts.size)
-    assert(taskPorts.head == range.get.getBegin.toInt)
-    assert(taskPorts(1) == range.get.getEnd.toInt)
+    assert(taskPorts.contains(range.get.getBegin.toInt))
+    assert(taskPorts.contains(range.get.getEnd.toInt))
```
kolloch: You are not testing for the size of the range. 

Basically, if I understand it correctly, we want to test that the same ports are contained in the range and the taskPorts sequence. Can't we say that a bit more nicely?

kolloch:
Better than before. I think it is possible to improve even more.

Please note that travis build links are not immutable. If you restart the build you discard the old build log. Thus it would be great if you put the original task failure into the commit message.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
gkleiman: Yes it is, I cleaned up the rest of the file and put those changes in a separate commit.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
gkleiman: The original test checked that the task got assigned exactly the first and last ports of the offered range.

This is often, but not always the case. I think that it is more appropriate to check only that all the ports it got are contained in the offer.

gkleiman:
Addressed all the comments, assigning this back to @kolloch.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -853,15 +838,15 @@ class TaskBuilderTest extends MarathonSpec {
...
       .map(r => r.getRanges.getRange(0))
-    assert(range.isDefined)
+    assert(maybeRange.isDefined)
+    val range = scala.collection.immutable.Range.Long.inclusive(maybeRange.get.getBegin, maybeRange.get.getEnd, 1)
     // The taskName is the elements of the path, reversed, and joined by dots
     assert("frontend.product" == taskInfo.getName)
     assert(2 == taskPorts.size)
-    assert(taskPorts.head == range.get.getBegin.toInt)
-    assert(taskPorts(1) == range.get.getEnd.toInt)
+    assert(taskPorts.forall(range.contains(_)))
```
kolloch: Wouldn't it be better to compare the sets of ports? E.g. `taskPorts.toSet == range.toSet`

In theory, it is possible that we end up with multiple ranges but maybe I shouldn't be so picky. If I am right, this should happen to 1 in thousand test runs.

gkleiman:
@kolloch: thanks for the feedback. I hope the third is the charm ;-)

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -853,15 +866,15 @@ class TaskBuilderTest extends MarathonSpec {
   }
 
   private def assertTaskInfo(taskInfo: TaskInfo, taskPorts: Seq[Long], offer: Offer): Unit = {
-    val range = taskInfo.getResourcesList.asScala
+    val maybeTaskInfoPorts = taskInfo.getResourcesList.asScala
       .find(r => r.getName == Resource.PORTS)
       .map(r => r.getRanges.getRange(0))
```
kolloch: It still only considers one range. What do you thing of the following?

```scala
  private def assertTaskInfo(taskInfo: TaskInfo, taskPorts: Seq[Long], offer: Offer): Unit = {
    // The taskName is the elements of the path, reversed, and joined by dots
    assert("frontend.product" == taskInfo.getName)

    val portsFromTaskInfo = {
      val asScalaRanges = for {
        resource <- taskInfo.getResourcesList.asScala if resource.getName == Resource.PORTS
        range <- resource.getRanges.getRangeList.asScala
      } yield range.getBegin to range.getEnd
      asScalaRanges.flatMap(_.iterator).toSet
    }

    assert(2 == taskPorts.size)
    assert(portsFromTaskInfo == taskPorts.toSet)

    assert(!taskInfo.hasExecutor)
   // ...
```

gkleiman:
PR Updated

aquamatthias:
LGTM. Please squash.

PR #2636, opened 2015-11-13
Add Oracle Linux to OS check

heichblatt:
This tiny patch adds Oracle Linux to the list of OSs recognized by the script.


aquamatthias:
Thanks!

PR #2641, opened 2015-11-16
PortWithRoleRandomPortsFromRangesTest: Increase benchmark/expected du…

kolloch:
…ration multiplicator

The old factor seems to occasionally break if the test is run with coverage.


aquamatthias:
Terrific!

PR #2648, opened 2015-11-16
Stop-gap warning about deployment design document

kolloch:
until we decide where to put design documents.


kolloch:
Hey @mwasn, how do you want to proceed with this pull request?

PR #2654, opened 2015-11-17
Fix bashism

heichblatt:
Somehow I managed to slip a tiny typo into my tiny patch. This fixes that.


aquamatthias:
Did't see that. Thanks.

PR #2666, opened 2015-11-17
Docs: Add "requiring ports" to common problems

kolloch:


docs/docs/common-problems.md
```diff
@@ -2,6 +2,34 @@
 title: Common Problems
 ---
 
+# An app stays in "Waiting" forever
+ 
+This means that Marathon does not receive "Resource Offers" from Mesos that allow it to start tasks of
+this application. The simplest failure is that there are not sufficient resources available in the cluster or another
+framework hords all these resources. You can check the Mesos UI for available resources. Note that the required resources
+(such as CPU, Mem, Disk) have to be all available on a single host.
```
aquamatthias: I think we should add constraints as one possibility. What do you think?

kolloch:
Also differentiate between troubleshooting and setup

aquamatthias:
Thanks for the effort! You want to add constraints to Waiting problem?

kolloch:
Hey Matthias, that would make sense and indeed I started with that. Right now, I approach the topic from the "Recipes" side: How to partition your cluster statically? 

I am also unclear about what to write about constraints in the Troubleshooting section. Since what would you write for constraints? That they are a likely candidate for deployment problems? That you have to make sure the spelling is correct?

Another topic I would like to add eventually is interference with Mesos roles.

Do you think that the documentation is in a bad state if we first merge as is? We can add more "problems" as we go.

BTW I renamed the page title since I didn't want it to start with Marathon. I wanted the prefix be significant. If you have nice suggestions, shoot ahead.

aquamatthias:
Like it. I think there are a lot points to add to that topic - let's do them one by one.

docs/_layouts/docs.html
```diff
@@ -65,8 +65,8 @@
         </a>
       </li>
       <li>
-        <a href="{{ site.baseurl }}/docs/common-problems.html">
-          Common Problems
+        <a href="{{ site.baseurl }}/docs/setup.html">
+          Configuration Tips for Marathon
```
aquamatthias: Don't know if "tip" is correct here. Hint, Recommendation?

docs/docs/troubleshooting.md
```diff
@@ -0,0 +1,61 @@
...
+  [...]
+```
+
+In this case, make sure that your local CA is included in your keystore by executing something like:
+
+```
+keytool -import -trustcacerts -file ./cacert.cer -alias yourca --keystore /my/path/to/marathon.jks
+```
+
+## Error reading authentication secret from file
```
aquamatthias: Do we have proof, that this is true? I mean it make sense to have secret longer than 8 chars, but at least I did not find any constraint like that.

aquamatthias:
2 questions - other than that it looks good to me.

docs/_layouts/docs.html
kolloch: It's a long word otherwise good ;) I'll take it.

docs/docs/troubleshooting.md
kolloch: Meta: Please comment below the line in question. That makes it easier in github to see the context.

I didn't write this for this PR. I just moved it to another file.

This was contributed by a user in an issue.

kolloch:
I changed the title. I still don't like it but can't think of something better.

aquamatthias:
Thanks.

PR #2669, opened 2015-11-18
Docs: How to use static partitioning with Marathon

kolloch:


docs/docs/recipes.md
```diff
@@ -189,3 +189,69 @@ since `0.7.4`).  Requires Mesos version `0.20.0` or later.
   "uris": ["http://downloads.mesosphere.com/misc/toggle.tgz"]
 }
 ```
+
+## Constraining task placement / static partitioning
+
+Usually, Mesos and Marathon collaboratively decide where to start your application tasks without your interference.
```
meichstedt: I'd either leave out the word _collaboration_ and only write something like _Marathon decides where to start your tasks based on resource offers it receives from Mesos_. I know this is basically the same, but if you don't define what that _collaboration_ is, it might be confusing to mention that.

Apart from that: I like this little introduction before diving into the details. (It might make sense to enhance that, put it into a more general location and reference here later on – no need to do that now.)

docs/docs/recipes.md
```diff
@@ -189,3 +189,69 @@ since `0.7.4`).  Requires Mesos version `0.20.0` or later.
...
+That is desirable because it truly makes your infrastructure easily scalable. All resources are treated equally. 
+However, there are times when you have
```
meichstedt: s/have to/want to

docs/docs/recipes.md
```diff
@@ -189,3 +189,69 @@ since `0.7.4`).  Requires Mesos version `0.20.0` or later.
...
+to separate a few machines from the common pool and ascertain that only some special tasks are scheduled there. For
```
meichstedt: I'd leave out _a few_ – I can imagine assigning half of the cluster to a specific use.

docs/docs/recipes.md
```diff
@@ -189,3 +189,69 @@ since `0.7.4`).  Requires Mesos version `0.20.0` or later.
...
+example,
+
+* you might have some nodes with publicly reachable network addresses. You only want to start tasks there
+  that you have marked explicitly.
```
meichstedt: s/that you have marked explicitly/that make use of public addresses/

docs/docs/recipes.md
```diff
@@ -189,3 +189,69 @@ since `0.7.4`).  Requires Mesos version `0.20.0` or later.
...
+Configuring Mesos Master: `--roles=<...other roles...>,yourrole`
+
+Configuring the Mesos Agents in question: Either use `--default_role yourrole` to assign all resources of that
+agent to your role or use the `--resources` configuration to assign individual resources to that role (such
+as a certain port range). See
+[the Mesos command line documentation](http://mesos.apache.org/documentation/latest/configuration/) for details.
+
+### Preventing Accidental Usage of Special Roles inside Marathon
+
+With the base setup, all Marathon applications by default will use all resources either assigned to the
```
meichstedt: s/by default will/will by default/ – imho

docs/docs/recipes.md
```diff
@@ -189,3 +189,69 @@ since `0.7.4`).  Requires Mesos version `0.20.0` or later.
...
+    "id": "/my.special.app",
+    "acceptedResourceRoles": [ "yourrole" ]
+    // ... more config ...
+}
+```
+
+### Limitations of Static Partitioning with Roles
+
+The biggest limitation is that one framework can currently only be assigned to one role. So if you have
+multiple types of special nodes you cannot handle them within one framework -- you need to have separate frameworks
```
meichstedt: An [em dash](http://www.thepunctuationguide.com/em-dash.html) would be typographically correct here. _sorry for nit-picking!_

meichstedt:
Some minor remarks from my side – otherwise LGTM.

docs/docs/recipes.md
kolloch: actually in LaTeX "--" gets translated to one. I though markdown would to that, too.

meichstedt:
:+1: Thanks!

PR #2686, opened 2015-11-20
Fixes #2681 - Circumvent cache if not yet preloaded

### LINKED ISSUE #2681: 0.11.1 -> 0.13.0-RC4 upgrade losing and then killing old tasks
I was testing upgrade from 0.11.1 to 0.13.0-RC4. Every attempted upgrade results in the new 0.13 leader not recognizing existing tasks, and killing all old tasks at once upon the first reconciliation. I can fail over leaders without upgrading in both 0.11->0.11 or 0.13->0.13 without losing tasks, it is only the upgrade 0.11->0.13  that has this issue. By greatly increasing `--reconciliation_initial_delay` I can make it so that the new tasks are launched before the old tasks get killed, however this is far from ideal IMO. 

Attached is part of the logs from the new 0.13 master, running a single app hello with 3 healthy tasks before the upgrade. 3 more tasks are started and the old 3 tasks are killed. I would expect that the old 3 tasks would be recognized and continue running.

[log.txt](https://github.com/mesosphere/marathon/files/39754/log.txt)

kolloch:
I am not completely happy with this patch and still have to test the migration but it should demonstrate the problem.

### DIFF
--- a/src/test/scala/mesosphere/marathon/test/Mockito.scala
+++ b/src/test/scala/mesosphere/marathon/test/Mockito.scala
@@ -28,6 +28,10 @@ trait Mockito extends MockitoSugar {
     M.verifyNoMoreInteractions(mocks: _*)
   }
 
+  def reset(mocks: AnyRef*): Unit = {
+    M.reset(mocks: _*)
+  }
+
   class MockAnswer[T](function: Array[AnyRef] => T) extends Answer[T] {
     def answer(invocation: InvocationOnMock): T = {
       function(invocation.getArguments)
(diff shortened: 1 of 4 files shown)


kolloch:
I tested changed it to using an Option. If you like the design, I will fix the tests.

src/test/scala/mesosphere/marathon/test/Mockito.scala
```diff
@@ -28,6 +28,10 @@ trait Mockito extends MockitoSugar {
     M.verifyNoMoreInteractions(mocks: _*)
   }
 
+  def reset(mocks: AnyRef*): Unit = {
+    M.reset(mocks: _*)
```
kolloch: unused now in the tests but it can be useful so I kept it.

meichstedt:
:+1: for adding missing test cases!

PR #2689, opened 2015-11-20
Fixes #2688 - Updates should handle failover

### LINKED ISSUE #2688: Failover during update scales below minimum health capacity
While working on https://github.com/mesosphere/marathon/issues/2249 I noticed that TaskReplaceActor may not be setup to handle Marathon failovers.

I put together a quick test to verify this. I started an update on a 2 instance app, then restarted the master. The new master picked up the deployment plan just fine, but it would kill too many tasks and fall below the minimum health capacity (Not every time).

I believe this is because the TaskReplaceActor.preStart method assumes it will be run once at the beginning of an update and then:
- tasksToKill doesn't take into account that some may be the new version and might not be healthy (often the case in an update). Because of this nrToKillImmediately is wrong and will kill tasks right away.
- reconcileNewTasks doesn't know that tasks were previously started.
- the backoff delay is reset every time.

This can be fixed by changing the preStart method to assume things could have already been started. I will put up a PR soon with some unit tests to illustrate the problem and propose a solution.

pgkelley4:
The TaskReplaceActor now assesses the current
state of an app instead of assuming all existing
apps are of the old version and healthy.

This change fixes #2688 where apps could be scaled
below their minimum health capacity during an
update if Marathon failed over.

Added new unit tests to show issue.

### DIFF
--- a/src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
+++ b/src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
@@ -5,10 +5,13 @@ import akka.event.EventStream
 import mesosphere.marathon.TaskUpgradeCanceledException
 import mesosphere.marathon.core.launchqueue.LaunchQueue
 import mesosphere.marathon.event.{ HealthStatusChanged, MesosStatusUpdateEvent }
+import mesosphere.marathon.health.HealthCheckActor.AppHealth
+import mesosphere.marathon.health.HealthCheckManager
 import mesosphere.marathon.state.AppDefinition
 import mesosphere.marathon.tasks.TaskTracker
 import mesosphere.marathon.upgrade.TaskReplaceActor._
 import org.apache.mesos.Protos.TaskID
+import org.apache.mesos.Protos.TaskState._
 import org.apache.mesos.SchedulerDriver
 
 import scala.collection.mutable
@@ -19,19 +22,21 @@ class TaskReplaceActor(
     driver: SchedulerDriver,
     taskQueue: LaunchQueue,
     taskTracker: TaskTracker,
+    healthCheckManager: HealthCheckManager,
     eventBus: EventStream,
     app: AppDefinition,
     promise: Promise[Unit]) extends Actor with ActorLogging {
   import context.dispatcher
 
-  val tasksToKill = taskTracker.get(app.id)
+  val newTasks = taskTracker.get(app.id).filter(_.getVersion == app.version.toString)
   val appId = app.id
   val version = app.version.toString
   var healthy = Set.empty[String]
-  var newTasksStarted: Int = 0
-  var oldTaskIds = tasksToKill.map(_.getId)
+  var newTasksStarted: Int = newTasks.size
+  var oldTaskIds = taskTracker.get(app.id).map(_.getId) -- newTasks.map(_.getId)
   val toKill = oldTaskIds.to[mutable.Queue]
   var maxCapacity = (app.instances * (1 + app.upgradeStrategy.maximumOverCapacity)).toInt
+  val minHealthy = (app.instances * app.upgradeStrategy.minimumHealthCapacity).ceil.toInt
   var outstandingKills = Set.empty[String]
   val periodicalRetryKills: Cancellable = context.system.scheduler.schedule(15.seconds, 15.seconds, self, RetryKills)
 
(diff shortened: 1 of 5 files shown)


pgkelley4:
This is a WIP. If this looks like a decent approach I have two TODOs in TaskReplaceActor that I would like an opinion on.

My changes to the unit tests do demonstrate the issue though. I added some new tests and fixed the existing ones as they break when TaskReplaceActor is fixed because they don't take app or task version into account.

pgkelley4:
Also I don't think either of these tests failed because of my changes and it doesn't look like I have permission to retrigger.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -2,33 +2,41 @@ package mesosphere.marathon.upgrade
...
     taskQueue: LaunchQueue,
     taskTracker: TaskTracker,
+    healthCheckManager: HealthCheckManager,
     eventBus: EventStream,
     app: AppDefinition,
     promise: Promise[Unit]) extends Actor with ActorLogging {
   import context.dispatcher
 
-  val tasksToKill = taskTracker.get(app.id)
+  private[this] val log = LoggerFactory.getLogger(getClass.getName)
```
kolloch: `getClass` is sufficient. The `.getName` is not necessary.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -2,33 +2,41 @@ package mesosphere.marathon.upgrade
...
+
+  val newTasks = taskTracker.get(app.id).filter(_.getVersion == app.version.toString)
+  val tasksToKill = taskTracker.get(app.id) -- newTasks
```
kolloch: jepp :+1: 

Unfortunately, using `Set[MarathonTask]` is kind of an anti-pattern that already caused all kinds of performance problems in the past. The problem is that `hashCode` on `MarathonTask` is fairly expensive. It is better if you work with `Set[TaskID]` or `Map[TaskID, MarathonTask]`.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -36,11 +44,34 @@ class TaskReplaceActor(
   val periodicalRetryKills: Cancellable = context.system.scheduler.schedule(15.seconds, 15.seconds, self, RetryKills)
 
   override def preStart(): Unit = {
+    // TODO there is a race condition if this come before or after the subscribe. If it goes before
+    // TODO   the subscribe then we miss updates during that time. If it comes after, updates could
+    // TODO   still be received out of sync with this request for status.
+    if (app.healthChecks.nonEmpty) {
+      // TODO what should the timeout be for this?
+      val health = Await.result(healthCheckManager.statuses(appId), 5.seconds)
```
kolloch: Even though we use it elsewhere, I do not like the `Await` approach much. Better use non-blocking constructs like multiple receives in different states.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -36,11 +44,34 @@ class TaskReplaceActor(
...
+      def is_task_healthy(marathonTask: MarathonTask): Boolean = {
```
kolloch: camel-case is customary in Scala

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -36,11 +44,34 @@ class TaskReplaceActor(
   val periodicalRetryKills: Cancellable = context.system.scheduler.schedule(15.seconds, 15.seconds, self, RetryKills)
 
   override def preStart(): Unit = {
+    // TODO there is a race condition if this come before or after the subscribe. If it goes before
+    // TODO   the subscribe then we miss updates during that time. If it comes after, updates could
+    // TODO   still be received out of sync with this request for status.
```
kolloch: We anyway get the health check updates with a delay. It is a distributed system. The task is already healthy before that state reaches us. So we just have to make sure that we are not staying in an outdated state forever.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
pgkelley4: Ahh ok. It looks like you made `TaskTracker.getTaskMap` private so I went with`Set[TaskID]` for this operation. It might be cleaner to get the map though... Let me know if you think I should make it public and use the map.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
pgkelley4: Whoops! Switching between this and Python.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -19,19 +23,23 @@ class TaskReplaceActor(
...
     taskQueue: LaunchQueue,
     taskTracker: TaskTracker,
+    healthCheckManager: HealthCheckManager,
     eventBus: EventStream,
     app: AppDefinition,
     promise: Promise[Unit]) extends Actor with ActorLogging {
   import context.dispatcher
 
-  val tasksToKill = taskTracker.get(app.id)
+  private[this] val log = LoggerFactory.getLogger(getClass)
```
gkleiman: We actually don't need this, since TaskReplaceActor uses the ActorLogging mixin.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -125,6 +148,14 @@ class TaskReplaceActor(
     checkFinished()
   }
 
+  def checkForTasksToKill(): Int = {
```
gkleiman: I'd name this killOldTasks, because it actually requests the old tasks to be killed.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -155,7 +186,7 @@ class TaskReplaceActor(
 
   def retryKills(): Unit = {
     outstandingKills.foreach { id =>
-      log.warning(s"Retrying kill of old task $id")
+      log.warn(s"Retrying kill of old task $id")
```
gkleiman: ActorLogging provides log#warning, not log#warn.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
pgkelley4: Sure, made the change.

pgkelley4:
Looks like my note got deleted when I made fixes for gkleiman.

But FYI @kolloch, with this implementation I have never actually seen the AppHealth message received by TaskReplaceActor in my testing. I believe this is because when the new master takes over and TaskReplaceActor is started there aren't any health checks results available yet. In practice the health of all tasks is received via `HealthStatusChanged` events, even after a failover.

However, I still think requesting AppHealth on startup of TaskReplaceActor is necessary to eliminate the potential race condition if there were health checks that had been run already.

kolloch:
Hi @pgkelley4, to review these patches, we would need considerable time -- maybe more than implementing a fix. Probably, we need to change some concepts.

Time has shown that our priorities are elsewhere. I am not happy about closing this PR like this but I don't think that we'll proceed with this any time soon.

Sorry! Alone your feedback about the bugs has been really valuable.

PR #2690, opened 2015-11-23
Fixes #2558 - If driver finishes with error, Marathon does not abdicate

### LINKED ISSUE #2558: If driver finishes with error, Marathon does not abdicate
(Not seen in the wild but found while performing a code review of related code)

If the `MesosDriver.run` throws an exception, The marathon instance is not abdicating its leadership.

This was tested with the following modification of the `runDriver` method of `MarathonSchedulerService`:

```scala
def runDriver(abdicateCmdOption: Option[ExceptionalCommand[JoinException]]): Unit = {
    log.info("Running driver")

    // The following block asynchronously runs the driver. Note that driver.run()
    // blocks until the driver has been stopped (or aborted).
    Future {
      //      driver.foreach(_.run())
      throw new RuntimeException("FAAAAAAIIIIIL")
    } onComplete {
      case Success(_) =>
        log.info("Driver future completed. Executing optional abdication command.")

        // If there is an abdication command we need to execute it so that our
        // leadership is given up. Note that executing the abdication command
        // does a few things: - It causes onDefeated() to be executed (which is
        // part of the Leader interface).  - It removes us as a leadership
        // candidate. We must offer out leadership candidacy if we ever want to
        // become the leader again in the future.
        //
        // If we don't have a abdication command we simply mark ourselves as
        // not the leader
        abdicateCmdOption match {
          case Some(cmd) => cmd.execute()
          case _         => leader.set(false)
        }

        // If we are shutting down then don't offer leadership. But if we
        // aren't then the driver was stopped via external means. For example,
        // our leadership could have been defeated or perhaps it was
        // abdicated. Therefore, for these cases we offer our leadership again.
        if (isRunning) {
          offerLeadership()
        }
      case Failure(t) =>
        log.error("Exception while running driver", t)
    }
  }
```

As you can see, the failure branch does not execute the abdicate option.

There are probably other problems as well since this code does different calls when what would happen if an exception is thrown in `onElected`:

```scala
      case NonFatal(e) => // catch Scala and Java exceptions
        log.error("Failed to take over leadership", e)

        increaseOfferLeadershipBackOff()

        abdicateLeadership()

        if (!driverHandlesAbdication) {
          // here the driver is not running yet and therefore it cannot execute
          // the abdication command and offer the leadership. So we do it here
          abdicateCmd.execute()
          offerLeadership()
        }
```

meln1k:
The MarathonSchedulerService now performs abdication action if driver ends with error.

This change fixes #2558 where abdication action wasn't performed in case of driver failure.

Added new unit test to show issue.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
@@ -202,6 +202,12 @@ class MarathonSchedulerService @Inject() (
   }
 
   def runDriver(abdicateCmdOption: Option[ExceptionalCommand[JoinException]]): Unit = {
+
+    def executeAbdicationCommand() = abdicateCmdOption match {
+      case Some(cmd) => cmd.execute()
+      case _         => leader.set(false)
+    }
+
     log.info("Running driver")
 
     // The following block asynchronously runs the driver. Note that driver.run()
@@ -221,10 +227,7 @@ class MarathonSchedulerService @Inject() (
         //
         // If we don't have a abdication command we simply mark ourselves as
         // not the leader
-        abdicateCmdOption match {
-          case Some(cmd) => cmd.execute()
-          case _         => leader.set(false)
-        }
+        executeAbdicationCommand()
 
         // If we are shutting down then don't offer leadership. But if we
         // aren't then the driver was stopped via external means. For example,
@@ -235,6 +238,7 @@ class MarathonSchedulerService @Inject() (
         }
       case Failure(t) =>
         log.error("Exception while running driver", t)
+        abdicateAfterFailure(() => executeAbdicationCommand(), runAbdicationCommand = true)
     }
   }
 
@@ -291,17 +295,7 @@ class MarathonSchedulerService @Inject() (
     catch {
       case NonFatal(e) => // catch Scala and Java exceptions
         log.error("Failed to take over leadership", e)
-
-        increaseOfferLeadershipBackOff()
-
-        abdicateLeadership()
-
-        if (!driverHandlesAbdication) {
-          // here the driver is not running yet and therefore it cannot execute
-          // the abdication command and offer the leadership. So we do it here
-          abdicateCmd.execute()
-          offerLeadership()
-        }
+        abdicateAfterFailure(() => abdicateCmd.execute(), runAbdicationCommand = !driverHandlesAbdication)
     }
   }
   //End Leader interface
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -235,6 +238,9 @@ class MarathonSchedulerService @Inject() (
         }
       case Failure(t) =>
         log.error("Exception while running driver", t)
+        executeAbdicationCommand()
```
kolloch: we probably need to increase the leadership backoff, right?

kolloch:
Thanks! 

What I would like also is to make it "obvious" that all these failure branches are handled the same. Maybe refactor it into common methods?

meln1k:
@kolloch does it look better now? If so, I'll squash my commits and push it to the pull request.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -421,4 +417,18 @@ class MarathonSchedulerService @Inject() (
       reconciliationInitialDelay.toMillis + reconciliationInterval.toMillis
     )
   }
+
+  private def startAbdication(abdicationCommand: () => Unit, driverHandlesAbdication: Boolean): Unit = {
```
kolloch: We should find a better name for this method and also the parameter `driverHandlesAbdication`. I would even question the utility of this flag. I know that you are preserving existing functionality but I think the functionality does not make sense. The flag is only set to `false` if `resetOfferLeadershipBackOff` throws an exception. Not very likely.

kolloch: Maybe, `abdicateAfterFailure` would be a good method name?

kolloch:
There are other things that are seriously wrong with the class at this moment but I think we can merge your code as an iterative improvement when you address the comments.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
meln1k: I agree, ```abdicateAfterFailure``` definitely sounds better. 

Actually, this flag initially set to false, and this allows us to handle migration failure, driver creation failure and failure in prepareStart method. Maybe this flag deserves to stay here?

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -292,16 +296,8 @@ class MarathonSchedulerService @Inject() (
       case NonFatal(e) => // catch Scala and Java exceptions
         log.error("Failed to take over leadership", e)
 
-        increaseOfferLeadershipBackOff()
+        abdicateAfterFailure(() => abdicateCmd.execute(), !driverHandlesAbdication)
```
kolloch: A named parameter for the boolean would be nice.

kolloch: I mean `abdicateAfterFailure(() => abdicateCmd.execute(), runAbdicationCommand = !driverHandlesAbdication)`

kolloch:
Let's address the last comment, squash and then ship it :) :ship:

meln1k:
Done :)

kolloch:
Thx! :ship: :rocket:

PR #2712, opened 2015-11-24
Fixes #2509 by listening for connection drops and abdicating directly

### LINKED ISSUE #2509: Leadership not abdicated on ZK connection loss
A Marathon Leader should abdicate if it loses its connection to ZK.

Right now it keeps leadership while trying to reconnect.

aquamatthias:


src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -105,6 +107,41 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, zk: ZooKeeperClient)
...
+    @Named(EventModule.busName) eventStream: EventStream,
+    leadershipCallbacks: Seq[LeadershipCallback] = Seq.empty): MarathonSchedulerService = {
+
+    val service = new MarathonSchedulerService(leadershipCoordinator, healthCheckManager, candidate, config,
+      frameworkIdUtil, leader, appRepository, taskTracker, driverFactory, system, migration, schedulerActor,
+      eventStream, leadershipCallbacks)
+
+    // The current candidate implementation does not handle ZK connection loss
+    // We register a listener on the candidate zk connection and handle disconnetions explicitly
+    zk.get().register(new Watcher {
```
kolloch: It would be nice to move this out of the Guice logic.

It also looks strange to me that you couple this with creating the MarathonSchedulerService.

Couldn't that simply be inside a `WhenLeaderActor` that has the `ZooKeeperClient` and the `MarathonSchedulerService` as dependencies?

kolloch:
@aquamatthias Thanks for taking care of this! I don't like placing the logic in the guice module, though...

aquamatthias:
@kolloch I agree. I have created a separate actor that handles this. 
Side Note: The biggest problem here was wiring everything. We need to improve this.

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
```diff
@@ -169,16 +171,25 @@ object ProcessKeeper {
     }
   }
 
+  def stopProcess(name: String): Option[Int] = {
+    log.info(s"Stop Process $name")
+    processes.get(name).map { process =>
+      Try(process.destroy())
+      processes -= name
+      process.exitValue()
```
gkleiman: Isn't this affected by the race conditions mentioned in L188?

It looks odd that stopAllProcesses actually doesn't call this method for each process.

gkleiman:
LGTM, just a minor comment regarding one of the new tests.

aquamatthias:
@kolloch Since Gaston is on vacation - can you review?

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
```diff
@@ -0,0 +1,40 @@
...
+import mesosphere.marathon.LeadershipAbdication
+import org.apache.zookeeper.{ WatchedEvent, Watcher }
+
+object AbdicateOnConnectionLossActor {
+  def props(zk: ZooKeeperClient, leader: LeadershipAbdication): Props = {
+    Props(new AbdicateOnConnectionLossActor(zk, leader))
+  }
+}
+
+class AbdicateOnConnectionLossActor(zk: ZooKeeperClient,
```
kolloch: minor: could be `private[impl]`

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
```diff
@@ -0,0 +1,40 @@
+package mesosphere.marathon.core.leadership.impl
+
+import akka.actor.{ Actor, ActorLogging, Props }
+import com.twitter.common.zookeeper.ZooKeeperClient
+import mesosphere.marathon.LeadershipAbdication
+import org.apache.zookeeper.{ WatchedEvent, Watcher }
+
+object AbdicateOnConnectionLossActor {
```
kolloch: minor: could be `private[leadership]`

src/test/scala/mesosphere/marathon/core/leadership/AlwaysElectedLeadershipModule.scala
```diff
@@ -1,21 +1,27 @@
...
   * Provides an implementation of the [[LeadershipModule]] which assumes that we are always the leader.
   *
   * This simplifies tests.
   */
-object AlwaysElectedLeadershipModule {
+object AlwaysElectedLeadershipModule extends Mockito {
   def apply(shutdownHooks: ShutdownHooks): LeadershipModule = {
     val actorsModule = new ActorsModule(shutdownHooks)
-    new AlwaysElectedLeadershipModule(actorsModule.actorRefFactory)
+    val zk = mock[ZooKeeperClient]
```
kolloch: This is test code which is used to stub the LeadershipModule. So it would be even better if we avoided starting nested actors which are unused. Unfortunately, the general pattern of using "startWhenLeader" currently is to do it in the module constructors (can we change that easily?) which makes it hard to prevent. If you have a simple idea...

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
```diff
@@ -0,0 +1,40 @@
...
+    zk.unregister(watcher)
+  }
+
+  def disconnected(): Unit = {
+    log.warning("ZooKeeper connection has been dropped. Abdicate Leadership.")
+    leader.abdicateLeadership()
+  }
+
+  override def receive: Receive = {
+    case event: WatchedEvent if event.getState == Watcher.Event.KeeperState.Disconnected => disconnected()
```
kolloch: That appears to be rather specific. I would rather like to test for a healthy state. Other than that we might want to deal with `Expired` as well?

These are possible states:

```java
        public enum KeeperState {
            /** Unused, this state is never generated by the server */
            @Deprecated
            Unknown (-1),

            /** The client is in the disconnected state - it is not connected
             * to any server in the ensemble. */
            Disconnected (0),

            /** Unused, this state is never generated by the server */
            @Deprecated
            NoSyncConnected (1),

            /** The client is in the connected state - it is connected
             * to a server in the ensemble (one of the servers specified
             * in the host connection parameter during ZooKeeper client
             * creation). */
            SyncConnected (3),

            /**
             * Auth failed state
             */
            AuthFailed (4),

            /**
             * The client is connected to a read-only server, that is the
             * server which is not currently connected to the majority.
             * The only operations allowed after receiving this state is
             * read operations.
             * This state is generated for read-only clients only since
             * read/write clients aren't allowed to connect to r/o servers.
             */
            ConnectedReadOnly (5),

            /**
              * SaslAuthenticated: used to notify clients that they are SASL-authenticated,
              * so that they can perform Zookeeper actions with their SASL-authorized permissions.
              */
            SaslAuthenticated(6),

            /** The serving cluster has expired this session. The ZooKeeper
             * client connection (the session) is no longer valid. You must
             * create a new client connection (instantiate a new ZooKeeper
             * instance) if you with to access the ensemble. */
            Expired (-112);
           // ...
}
```

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
```diff
@@ -0,0 +1,40 @@
...
+                                    leader: LeadershipAbdication) extends Actor with ActorLogging {
+
+  private[impl] val watcher = new Watcher {
+    val reference = self
+    override def process(event: WatchedEvent): Unit = reference ! event
+  }
+
+  override def preStart(): Unit = {
+    log.info("Register as ZK Listener")
+    zk.register(watcher)
```
kolloch: To avoid race-conditions, you should test the current zookeeper connection state after registering for events. The connection might already have died.

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
kolloch: My assumption is that we use the zookeeper connection for leadership election. So we can assume a working connection when this code is called.

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
kolloch: Short class comment would be nice.

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
```diff
@@ -50,4 +50,24 @@ class LeaderIntegrationTest extends IntegrationFunSuite
     And("the leader must have changed")
     WaitTestSupport.waitUntil("the leader changes", 30.seconds) { marathon.leader().value != leader }
   }
+
+  test("abdicate if the zk connection is dropped") {
```
kolloch: nice to have an integration test for this!

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
```diff
@@ -50,4 +50,24 @@ class LeaderIntegrationTest extends IntegrationFunSuite
...
+    ProcessKeeper.stopProcess("zookeeper")
+
+    Then("There is no leader in the cluster")
+    WaitTestSupport.waitUntil("a leader has been abdicated", 30.seconds) {
+      //TODO: should it be always 404?
+      Seq(404, 503).contains(marathon.leader().code)
+    }
+
+    When("Zookeeper starts again")
+    ProcessKeeper.startZooKeeper(config.zkPort, "/tmp/foo/single", wipeWorkDir = false)
```
kolloch: Why are you using a different working directory here?

kolloch:
Hey @aquamatthias, looks good. Mostly minor nitpicks and one race-condition. The only major thing is that I am not sure if the error detection code is exhaustive.

src/main/scala/mesosphere/marathon/core/leadership/impl/AbdicateOnConnectionLossActor.scala
aquamatthias: Check connection in preStart()

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
aquamatthias: This is the same path as in SingleMarathonIntegrationTest.

kolloch: Maybe move it to a constant in SingleMarathonIntegrationTest to make that clear?

kolloch:
Thanks for the changes. I would suggest moving that Zookeeper path to a constant and referencing it in the test. Otherwise I would keep being confused about the meaning of that path in the test. Otherwise looks good!

aquamatthias:
Extracted a def with parameter default values.
Rebased and Squashed.

kolloch:
:shipit: :rocket:

Issue #2709, opened 2015-11-24
Prototype per-container IP support


gkleiman:
Create a prototype of per-container IP support.

See #2698.

Acceptance criteria:

1. Have a docker image and setup instructions ready for user test.

PR #2731, opened 2015-11-26
Towards #2709 - Allow port instead of portIndex for health checks

kolloch:
* Improved health check documentation
  * Replaced JInt with scala Int in HealthCheck case class
  * Automatically add `"portIndex": 0` when necessary to preserve backwards compatibility

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -350,14 +350,14 @@ trait HealthCheckFormats {
     (
       (__ \ "path").formatNullable[String] ~
       (__ \ "protocol").formatNullable[Protocol].withDefault(DefaultProtocol) ~
-      (__ \ "portIndex").formatNullable[Integer].withDefault(DefaultPortIndex) ~
+      (__ \ "portIndex").formatNullable[Int] ~
       (__ \ "command").formatNullable[Command] ~
       (__ \ "gracePeriodSeconds").formatNullable[Long].withDefault(DefaultGracePeriod.toSeconds).asSeconds ~
       (__ \ "intervalSeconds").formatNullable[Long].withDefault(DefaultInterval.toSeconds).asSeconds ~
       (__ \ "timeoutSeconds").formatNullable[Long].withDefault(DefaultTimeout.toSeconds).asSeconds ~
-      (__ \ "maxConsecutiveFailures").formatNullable[Integer].withDefault(DefaultMaxConsecutiveFailures) ~
+      (__ \ "maxConsecutiveFailures").formatNullable[Int].withDefault(DefaultMaxConsecutiveFailures) ~
       (__ \ "ignoreHttp1xx").formatNullable[Boolean].withDefault(DefaultIgnoreHttp1xx) ~
-      (__ \ "overridePort").formatNullable[Integer]
+      (__ \ "port").formatNullable[Int]
     )(HealthCheck.apply, unlift(HealthCheck.unapply))
   }
 }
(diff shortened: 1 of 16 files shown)


src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -446,6 +447,26 @@ trait V2Formats {
...
+    *
+    * In the past, healthCheck.portIndex was required and had a default value 0. When we introduced healthCheck.port, we
+    * made it optional (also with ip-per-container in mind) and we have to re-add it in cases where it makes sense.
+    */
+  private[this] def addHealthCheckPortIndexIfNecessary(v2App: V2AppDefinition): V2AppDefinition = {
+    val hasPortMappings = v2App.container.exists(_.docker.exists(_.portMappings.exists(_.nonEmpty)))
+    val portIndexesMakeSense = v2App.ports.nonEmpty || hasPortMappings
+    if (portIndexesMakeSense) {
+      v2App.copy(healthChecks = v2App.healthChecks.map { healthCheck =>
+        if (healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty && healthCheck.protocol != Protocol.COMMAND) {
```
aquamatthias: Can you extract those checks as flag: `val needPortIndex = x&y&z ` and then have only one if: `if (needPortIndex && portIndexesMakeSense) app.copy(...)`

src/main/scala/mesosphere/marathon/health/HealthCheck.scala
```diff
@@ -54,7 +58,13 @@ case class HealthCheck(
       path =
         if (proto.hasPath) Some(proto.getPath) else None,
       protocol = proto.getProtocol,
-      portIndex = proto.getPortIndex,
+      portIndex =
+        if (proto.hasPortIndex)
+          Some(proto.getPortIndex)
+        else if (!proto.hasPort && proto.getProtocol != Protocol.COMMAND)
```
aquamatthias: Can we run into this situation?

src/test/scala/mesosphere/marathon/health/HealthCheckTest.scala
```diff
@@ -31,12 +30,35 @@ class HealthCheckTest extends MarathonSpec {
...
     assert(60 == proto.getIntervalSeconds)
     assert(0 == proto.getMaxConsecutiveFailures)
+    assert(!proto.hasPort)
+  }
+
+  test("ToProtoOverridePort") {
+    val healthCheck = HealthCheck(
+      path = Some("/health"),
+      protocol = Protocol.HTTP,
+      portIndex = Some(0),
```
aquamatthias: Isn't that something we forbid to create: portIndex and port?

aquamatthias:
Only minor comments. Looks good! :+1: 
Please address and :ship:

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
kolloch: You mean like this?

```scala

  /**
    * Ensure backwards compatibility by adding portIndex to health checks when necessary.
    *
    * In the past, healthCheck.portIndex was required and had a default value 0. When we introduced healthCheck.port, we
    * made it optional (also with ip-per-container in mind) and we have to re-add it in cases where it makes sense.
    */
  private[this] def addHealthCheckPortIndexIfNecessary(v2App: V2AppDefinition): V2AppDefinition = {
    val hasPortMappings = v2App.container.exists(_.docker.exists(_.portMappings.exists(_.nonEmpty)))
    val portIndexesMakeSense = v2App.ports.nonEmpty || hasPortMappings
    v2App.copy(healthChecks = v2App.healthChecks.map { healthCheck =>
        def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty && healthCheck.protocol != Protocol.COMMAND
        if (portIndexesMakeSense && needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
        else healthCheck
    })
  }
```

Slightly less efficient but more readable, I like it.

src/main/scala/mesosphere/marathon/health/HealthCheck.scala
kolloch: We are not sure. We might be for backwards compatibility. As written below, portIndex = 0 was the default int protobufs. We don't know if we can get into this situation by a combination of using an old Marathon version and migrations so we thought that we better cover it.

src/test/scala/mesosphere/marathon/health/HealthCheckTest.scala
kolloch: Yes, we do that. This test only covers the serialization. Should we separate this to two different tests?

kolloch:
@aquamatthias I hopefully addressed all your concerns in an additional commit. I can squash our two commits if you are happy (I would like to preserve Evans to give him credit where it's due).

aquamatthias:
Thanks! LGTM please squash and merge.

kolloch:
It is squashed! :)  We will not squash Evans patch.

aquamatthias:
Looks good. Thanks!

PR #2752, opened 2015-12-02
Add trivago as a user of Marathon

andygrunwald:
We ([trivago](http://www.trivago.com/)) are using Mesos, Marathon and Chronos in one of our datacenters. This pull request adds trivago as a Marathon user.


gkleiman:
Thanks a lot, it's always good to hear from new users!

PR #2756, opened 2015-12-03
Fixes #2730 by removing the unused schema.

### LINKED ISSUE #2730: Don't have AppDefinition.json twice
We have the `AppDefinition.json` file twice in our project repository.

@aquamatthias is that necessary?

It is really error prone because you easily forget to update one of them.

aquamatthias:
Enhance the schemas with documentation and validation, where applicable.


docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
```diff
@@ -150,18 +173,21 @@
...
         "$ref": "#/definitions/pathType"
-      },
-      "type": "array"
+      }
     },
     "disk": {
-      "minimum": 0,
-      "type": "number"
+      "type": "number",
+      "description": "How many disk space is needed for this application. This number does not have to be integer, but can be a fraction.",
```
gkleiman: s/How many/How much/
s/be integer/be an integer/ ?

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
```diff
@@ -185,47 +213,57 @@
...
-            "minimum": 0,
-            "type": "integer"
+            "type": "integer",
+            "description": "Number of consecutive health check failures after which the unhealthy task should be killed.",
+            "minimum": 0
           },
           "path": {
-            "type": "string"
+            "type": "string",
+            "description": "Path to endpoint exposed by the task that will provide health status. Example: /path/to/health. Note: only used if protocol == HTTP."
```
gkleiman: s/HTTP/HTTP[S]/ ?

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
```diff
@@ -234,83 +272,100 @@
       "description": "Unique identifier for the app consisting of a series of names separated by slashes. Each name must be at least 1 character and may only contain digits (`0-9`), dashes (`-`), dots (`.`), and lowercase letters (`a-z`). The name may not begin or end with a dash."
     },
     "instances": {
-      "minimum": 0,
-      "type": "integer"
+      "type": "integer",
+      "description": "The number of instances of this application to start. Please note: this number can be changed every time as needed to scale the application.",
```
gkleiman: s/every time/any time/

gkleiman:
Added a few comments.

The changes break V2AppDefinitionTest.

aquamatthias:
@gkleiman Thanks for the feedback.
1) Most of the documentation comes from our GH pages doc - we should update that as well.
2) The failing test was wrong (as I see it) the expectation was to fail in the case of ports=[0, 0] which is perfectly valid (and I removed that restriction from the schema).

gkleiman:
Thanks! :rocket:

PR #2743, opened 2015-11-30
Update index.md

mrajashree:
The images weren't being displayed

PR #2759, opened 2015-12-03
Fixes #2743 by only using the available assigned ports instead of all…

aquamatthias:
… dynamically assigned ports.

Problem in short:
step 1: user send ports=[0] this will result in ports=[10000]
step 2: user updates ports to ports=[10000, 0] 
the logic would see 10000 as available free port, remembers that and will it assign to the next dynamic value. This will result in [10000, 10000] which breaks validation.


aquamatthias:
Sorry - wrong issue number

PR #2765, opened 2015-12-04
Do not remove frameworkId automatically.

aquamatthias:


src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -1,22 +1,20 @@
 package mesosphere.marathon
 
-import javax.inject.{ Inject, Named }
+import javax.inject.{Inject, Named}
```
gkleiman: It looks like your IDE insists on reformatting the imports =(

Issue #2781, opened 2015-12-04
subprocess call should close file descriptors in child


cgbaker:
_Note: this is only an issue in python2, which is the version used by servicerouter.py in 0.11.x, 0.12.x and 0.13.x branches. python3 uses `close_fd=True` by default._

I've run into a strange problem with servicerouter.py. I'm using it in listening mode, like so: 
python ./servicerouter.py --marathon=http://localhost:8080 --listening=http://localhost:8081
I noticed that after shutting down the servicerouter (Ctrl-C/SIGTERM), I couldn't start it back up listening on the same port. This only occurred if the process did a reload of the haproxy config. Netstat tells me that my listening port (8081) is owned by haproxy.

Apparently what is happening is that the haproxy reload subprocess spawned by the servicerouter script is inheriting the file descriptors (specifically, the one corresponding to listening socket). When it spawns a new haproxy, that inherits them as well. The reload process finishes and terminates, so that if I kill the servicerouter some time later, the new haproxy is left holding these file descriptors. 

A simple workaround is to restart haproxy before trying to run the servicerouter again. 

I'm not sure whether there's an underlying issue with the WSGI library, i.e., whether the web server isn't closing the sockets appropriately on shutdown. One solution I've found is to start the subprocesses from the Python servicerouter in such a way that they don't inherit file descriptors. This seems to be as simple as changing the call from
`subprocess.check_call(reloadCommand)`
to 
`subprocess.check_call(reloadCommand, close_fds=True)`
I've tested this and it settles the problem as I've observed it.

Details:
I'm using servicerouter.py from Marathon 0.11.1:
https://github.com/mesosphere/marathon/blob/v0.11.1/bin/servicerouter.py
on Centos 6.7 (6-7.el6.centos.12.3), with python 2.6.6.

aldipower:
Hi @cgbaker !
I see, this is annoying problem, thanx for reporting this.

@lloesche

lloesche:
@cgbaker thanks for the report. We migrated servicerouter.py into it's own project called marathon-lb. Your fix is incorporated in https://github.com/mesosphere/marathon-lb/pull/30

P.S. in marathon-lb you can also use SSE (https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2-events) instead of Callbacks for reloading the config. That way you don't need a connection from your Marathon hosts back to the load balancer machines.

PR #2782, opened 2015-12-04
modified call to subprocess.check_call for haproxy reload to close all file descriptors in the child

cgbaker:
for history, see discussion from marathon_lb project: 
https://github.com/mesosphere/marathon-lb/pull/29

references #2781


cgbaker:
also, if there will be additional patch releases on 0.11.x and 0.12.x, this should be backpatched there

lloesche:
:shipit:

PR #2787, opened 2015-12-07
Expose additional app definition variables

wendigo:
I would like to expose additional appDefinition fields as environment variables

### DIFF
--- a/docs/docs/task-environment-vars.md
+++ b/docs/docs/task-environment-vars.md
@@ -44,9 +44,17 @@ names through the `--env_vars_prefix` command line flag.
   was used to start this task. For example `2015-04-02T09:37:00.596Z`.
 - `MARATHON_APP_DOCKER_IMAGE` contains the value of the app definition 
   `container.docker.image` value. For example `mesosphere/marathon:latest`, `nginx` or `nginx:1.9.3`.
+- `MARATHON_APP_RESOURCE_CPUS` contains the value of the app definition `cpus` value.
+- `MARATHON_APP_RESOURCE_MEM` contains the value of the app definition `mem` value, expressed in megabytes.
+- `MARATHON_APP_RESOURCE_DISK`  value of the app definition `disk` value, expressed in megabytes.
+- `MARATHON_APP_LABELS` contains list of labels of the corresponding app definition. For example: `label1 label2 label3`
+- `MARATHON_APP_LABEL_NAME` contains value of label "NAME" of the corresponding app definition. 
 
 _Note: No custom prefix will be added to these variable names, even if one is
 specified via the `--env_vars_prefix` command line flag.
+
+_Note: Label name will be sanitized: all characters except a-z A-Z 0-9 _ will be replaced with _. 
+All keys and values longer than 512 chars will be omitted.
 
 ## Task Metadata
 
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -1,6 +1,6 @@
 package mesosphere.mesos
 
-import com.google.protobuf.{ TextFormat, ByteString }
+import com.google.protobuf.{ ByteString, TextFormat }
 import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
 import mesosphere.marathon.Protos.MarathonTask
 import mesosphere.marathon._
@@ -16,6 +16,7 @@ import play.api.libs.json.Json
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.Seq
+import scala.collection.mutable
 
 class TaskBuilder(app: AppDefinition,
                   newTaskId: PathId => TaskID,
@@ -208,6 +209,10 @@ class TaskBuilder(app: AppDefinition,
 
 object TaskBuilder {
 
+  val maxEnvironmentVarLength = 512
+  val labelEnvironmentKeyPrefix = "MARATHON_APP_LABEL_"
+  val maxVariableLength = maxEnvironmentVarLength - labelEnvironmentKeyPrefix.length
+
   def commandInfo(app: AppDefinition,
                   taskId: Option[TaskID],
                   host: Option[String],
@@ -318,10 +323,26 @@ object TaskBuilder {
         "MESOS_TASK_ID" -> taskId.map(_.getValue),
         "MARATHON_APP_ID" -> Some(app.id.toString),
         "MARATHON_APP_VERSION" -> Some(app.version.toString),
-        "MARATHON_APP_DOCKER_IMAGE" -> app.container.flatMap(_.docker.map(_.image))
+        "MARATHON_APP_DOCKER_IMAGE" -> app.container.flatMap(_.docker.map(_.image)),
+        "MARATHON_APP_RESOURCE_CPUS" -> Some(app.cpus.toString),
+        "MARATHON_APP_RESOURCE_MEM" -> Some(app.mem.toString),
+        "MARATHON_APP_RESOURCE_DISK" -> Some(app.disk.toString)
       ).collect {
           case (key, Some(value)) => key -> value
-        }.toMap
+        }.toMap ++ labelsToEnvVars(app.labels)
     }
+  }
+
+  def labelsToEnvVars(labels: Map[String, String]): Map[String, String] = {
+    def escape(name: String): String = name.replaceAll("[^a-zA-Z0-9_]+", "_").toUpperCase
+
+    val validLabels = labels.collect {
+      case (key, value) if key.length < maxVariableLength
+        && value.length < maxEnvironmentVarLength => escape(key) -> value
+    }
+
+    val names = Map("MARATHON_APP_LABELS" -> validLabels.keys.mkString(" "))
+    val values = validLabels.map { case (key, value) => s"$labelEnvironmentKeyPrefix$key" -> value }
+    names ++ values
   }
 }
--- a/src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
+++ b/src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
@@ -596,7 +596,11 @@ class TaskBuilderTest extends MarathonSpec {
       env == Map(
         "MESOS_TASK_ID" -> "taskId",
         "MARATHON_APP_ID" -> "/app",
-        "MARATHON_APP_VERSION" -> "2015-02-03T12:30:00.000Z"
+        "MARATHON_APP_VERSION" -> "2015-02-03T12:30:00.000Z",
+        "MARATHON_APP_RESOURCE_CPUS" -> AppDefinition.DefaultCpus.toString,
+        "MARATHON_APP_RESOURCE_MEM" -> AppDefinition.DefaultMem.toString,
+        "MARATHON_APP_RESOURCE_DISK" -> AppDefinition.DefaultDisk.toString,
+        "MARATHON_APP_LABELS" -> ""
       )
     )
   }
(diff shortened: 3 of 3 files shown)


gkleiman:
LGTM

wendigo:
Waiting for comments then :)

docs/docs/task-environment-vars.md
```diff
@@ -44,6 +44,10 @@ names through the `--env_vars_prefix` command line flag.
   was used to start this task. For example `2015-04-02T09:37:00.596Z`.
 - `MARATHON_APP_DOCKER_IMAGE` contains the value of the app definition 
   `container.docker.image` value. For example `mesosphere/marathon:latest`, `nginx` or `nginx:1.9.3`.
+- `MARATHON_APP_CPUS` contains the value of the app definition `cpus` value.
+- `MARATHON_APP_MEMORY` contains the value of the app definition `mem` value, expressed in megabytes.
```
gkleiman: Can we please follow the naming schema used in the API, i.e., 'MEM' and 'DISK'?

docs/docs/task-environment-vars.md
```diff
@@ -44,6 +44,10 @@ names through the `--env_vars_prefix` command line flag.
...
+- `MARATHON_APP_DISK_SPACE`  value of the app definition `disk` value, expressed in megabytes.
+- `MARATHON_APP_LABELS` contains list of labels of the corresponding app definition. For example: `label1:value1;label2:value2`
```
gkleiman: The following variables would be easier to use from shell scripts:

MARATHON_APP_LABELS=“LABEL1 LABEL2 LABEL3”
MARATHON_APP_LABEL_LABEL1=value1
MARATHON_APP_LABEL_LABEL2=value2

We would however have to make sure we convert invalid characters/spaces in the label names to underscores.

wendigo: OK. Will do

docs/docs/task-environment-vars.md
wendigo: OK.

docs/docs/task-environment-vars.md
wendigo: I will follow discussion here: http://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names to provide environment variable name safety

aquamatthias:
Hey @wendigo thanks for this PR.

1) Resources
Please change the names of the resources to MARATHON_APP_RESOURCE_*:
MARATHON_APP_RESOURCE_MEM
MARATHON_APP_RESOURCE_CPUS
MARATHON_APP_RESOURCE_DISK

2) Labels
There is a size limit for env vars: See here: http://man7.org/linux/man-pages/man2/execve.2.html
"Limits one size of arguments and environment"
I guess it makes sense to only include labels, that do not exceed a defined length (key and value).
We should make that length adjustable (default 512 chars?).

Thanks again! Matthias

wendigo:
Hi @aquamatthias:

1) Resources - I've changed variable names according to your comment
2) Implemented. Limit (512) is hardcoded for now. I think I can manage to make it configurable if you would like to.

P.S. I'm not a Scala dev so all comments are appreciated :)

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -1,6 +1,6 @@
 package mesosphere.mesos
 
-import com.google.protobuf.{ TextFormat, ByteString }
+import com.google.protobuf.{ ByteString, TextFormat }
```
wendigo: IntelliJ :)

docs/docs/task-environment-vars.md
```diff
@@ -44,6 +44,11 @@ names through the `--env_vars_prefix` command line flag.
   was used to start this task. For example `2015-04-02T09:37:00.596Z`.
 - `MARATHON_APP_DOCKER_IMAGE` contains the value of the app definition 
   `container.docker.image` value. For example `mesosphere/marathon:latest`, `nginx` or `nginx:1.9.3`.
+- `MARATHON_RESOURCE_CPUS` contains the value of the app definition `cpus` value.
```
aquamatthias: Please use MARATHON_APP_RESOURCE_CPUS

docs/docs/task-environment-vars.md
```diff
@@ -44,6 +44,11 @@ names through the `--env_vars_prefix` command line flag.
...
+- `MARATHON_RESOURCE_MEM` contains the value of the app definition `mem` value, expressed in megabytes.
```
aquamatthias: Please use MARATHON_APP_RESOURCE_MEM

docs/docs/task-environment-vars.md
```diff
@@ -44,6 +44,11 @@ names through the `--env_vars_prefix` command line flag.
...
+- `MARATHON_RESOURCE_DISK`  value of the app definition `disk` value, expressed in megabytes.
```
aquamatthias: Please use MARATHON_APP_RESOURCE_DISK

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -318,10 +321,38 @@ object TaskBuilder {
...
+      case (key, value) => {
+        variables = variables ++ Map(s"${labelEnvironmentKeyPrefix}${escapeEnvironmentVariableName(key)}" -> value)
+      }
+    }
+
+    variables
+  }
+
+  def escapeEnvironmentVariableName(name: String): String = {
+    val variableName = name.replaceAll("[^a-zA-Z0-9_]+", "_").toUpperCase
```
aquamatthias: This would replace subsequent invalid chars into a single underscore: e.g. `1234$$$$` into `1234_`. If you omit the leading `+`, then the string would be replaced to `1234____`
I am not exactly sure which one is better...

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -318,10 +321,38 @@ object TaskBuilder {
...
+
+    if (variableName.length > maxVariableLength) {
+      variableName.substring(0, maxVariableLength)
```
aquamatthias: I rather would completely omit a variable if it is longer than the maximum.
Truncating could lead to strange behavior as well...

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -318,10 +321,38 @@ object TaskBuilder {
...
+  }
+
+  def labelsToEnvVars(labels: Map[String, String]): Map[String, String] = {
+    var variables = Map(
+      "MARATHON_APP_LABELS" -> labels.keys.map(escapeEnvironmentVariableName).mkString(" ")
+    )
+
+    labels.foreach {
+      case (key, value) => {
+        variables = variables ++ Map(s"${labelEnvironmentKeyPrefix}${escapeEnvironmentVariableName(key)}" -> value)
```
aquamatthias: What about the values. They could exceed size as well, or?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -318,10 +321,38 @@ object TaskBuilder {
         "MESOS_TASK_ID" -> taskId.map(_.getValue),
         "MARATHON_APP_ID" -> Some(app.id.toString),
         "MARATHON_APP_VERSION" -> Some(app.version.toString),
-        "MARATHON_APP_DOCKER_IMAGE" -> app.container.flatMap(_.docker.map(_.image))
+        "MARATHON_APP_DOCKER_IMAGE" -> app.container.flatMap(_.docker.map(_.image)),
+        "MARATHON_RESOURCE_CPUS" -> Some(app.cpus.toString),
+        "MARATHON_RESOURCE_MEM" -> Some(app.mem.toString),
+        "MARATHON_RESOURCE_DISK" -> Some(app.disk.toString)
```
aquamatthias: See naming scheme

src/main/scala/mesosphere/mesos/TaskBuilder.scala
wendigo: Ok :)

src/main/scala/mesosphere/mesos/TaskBuilder.scala
aquamatthias: You can do something like, so you don't create a new map with every entry:
```scala
variables += key->value
```

src/main/scala/mesosphere/mesos/TaskBuilder.scala
wendigo: My mistake. Adjusted naming as you suggested.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -318,10 +322,38 @@ object TaskBuilder {
...
+  def labelsToEnvVars(labels: Map[String, String]): Map[String, String] = {
+    var variables = mutable.Map(
+      "MARATHON_APP_LABELS" -> labels.keys
+        .map(escapeEnvironmentVariableName)
+        .filter(_.length <= maxVariableLength)
+        .mkString(" ")
+    )
+
+    labels.foreach {
+      case (key, value) => {
```
aquamatthias: You can check the length of key and value here:
`case (key, value) if key.length<512 && value.length<512 => ...`

docs/docs/task-environment-vars.md
```diff
@@ -44,10 +44,18 @@ names through the `--env_vars_prefix` command line flag.
...
+- `MARATHON_APP_RESOURCE_MEM` contains the value of the app definition `mem` value, expressed in megabytes.
+- `MARATHON_APP_RESOURCE_DISK`  value of the app definition `disk` value, expressed in megabytes.
+- `MARATHON_APP_LABELS` contains list of labels of the corresponding app definition. For example: `label1 label2 label3`
+- `MARATHON_APP_LABEL_NAME` contains value of label "NAME" of the corresponding app definition. 
 
 _Note: No custom prefix will be added to these variable names, even if one is
 specified via the `--env_vars_prefix` command line flag.
 
+_Note: Label name will be sanitized: all characters except a-z A-Z 0-9 _ will be replaced with _ and 
+label will be truncated to 512 chars (by default).
```
aquamatthias: Not truncated, but filtered out.

aquamatthias:
@wendigo Thanks for the update. 
Please also check the length of the label value. 
If you could squash the commits, I can merge the functionality.

wendigo:
Commit squashed.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -318,10 +323,42 @@ object TaskBuilder {
...
+        "MARATHON_APP_RESOURCE_MEM" -> Some(app.mem.toString),
+        "MARATHON_APP_RESOURCE_DISK" -> Some(app.disk.toString)
       ).collect {
           case (key, Some(value)) => key -> value
-        }.toMap
+        }.toMap ++ labelsToEnvVars(app.labels)
+    }
+  }
+
+  def labelsToEnvVars(labels: Map[String, String]): Map[String, String] = {
```
aquamatthias: Sorry to be picky: collect already filters automatically. May I suggest the following code:
```scala
def labelsToEnvVars(labels: Map[String, String]): Map[String, String] = {
  def escape(name: String): String = name.replaceAll("[^a-zA-Z0-9_]+", "_").toUpperCase
  val validLabels = labels.collect {
    case (key, value) if key.length<maxEnvLen && value.length<maxEnvLen => escape(key) -> value
  }

  val names = Map("MARATHON_APP_LABELS" -> validLabels.keys.mkString(" "))
  val values = validLabels.map { case (key, value) => s"$labelEnvironmentKeyPrefix$key" -> value }
  names ++ values
}
```

wendigo: Done :)

aquamatthias: Great - will wait for the tests.

wendigo: Passed :)

aquamatthias:
Thanks!

lloesche:
@wendigo I love this feature, thank you!

PR #2798, opened 2015-12-08
Remove airbnb from companies list

ljharb:
We're no longer using Marathon internally at @Airbnb. Noticed per https://twitter.com/CleverRoffey/status/674334494040567808


gkleiman:
Thanks for the update. It would be interesting to know what led @Airbnb to stop using Marathon!

PR #2801, opened 2015-12-09
IP per Container

aquamatthias:


aquamatthias:
__TODO: Fix mesos dependency before merge!__

Issue #2735, opened 2015-11-26
Change validation routines to non-BeanValidation ones


alexanderweber:
Accord seems to be a nice alternative for model validation. (http://wix.github.io/accord/)
The goal is to get rid of the ModelValidation and provide implicit validation for relevant classes.

kolloch:
#2723 is very similar. Probably we should merge these issues.

alexanderweber:
Fixed with #2804.

Issue #2426, opened 2015-10-13
Get rid of V2 model objects


kolloch:
Now that we use play json serialization, the structure of the model objects does not have to correspond to the JSON structure and we can get rid of one level of indirection.

(@connor, do you agree?)

connor:
@kolloch I think you mis-tagged me :)

gkleiman:
@connor: sorry for that, I believe he meant @ConnorDoyle

ConnorDoyle:
@kolloch :+1:

Couple of related thoughts:

- What do you thnk about moving the `marathon.api.validation` logic into the play-json formatters and also get rid of hibernate validation?  If you need API-independent model validation, maybe a stackable Validation trait in the state package makes sense?
- You should also be able to drop the project dependency on the `jackson-case-class-module` artifact.  I wrote that thing as a shim for handling default values for Scala case classes.  Happy to see it go away :)  https://github.com/mesosphere/jackson-case-class-module .
- We shouldn't forget to allow HTTP event subscribers to choose the API version formatters to use for serializing events before introducing v3 formatters.

PR #2804, opened 2015-12-09
#2735 - replacement of ModelValidation

alexanderweber:
@aquamatthias @kolloch @gkleiman @meichstedt Hey guys, we decided to give [Accord Validation](https://github.com/wix/accord) a try and replace our ModelValidation, which was still using Java's BeanValidation. The new library is capable of keeping track of the path down to the occurrence of the error. It is also using macros and detects property names automatically, so there is no need in propagating them explicitly. Validation rules can be constructed in an easy manner like this:
```scala
implicit val appDefinitionValidator = validator[V2AppDefinition] { appDef =>
    appDef.id is valid
    appDef.dependencies is valid
    appDef.upgradeStrategy is valid
    appDef.storeUrls is every(urlsCanBeResolvedValidator)
}
```
As it is an implicit validator, you could easily validate an `appDefinition` from type `V2AppDefinition` like this:
```scala
...
validate(appDefintion)
...
```
We already tried the new validation out on some API methods. E.g. a following POST request in /v2/apps
```json
{
  "id": "/test",
  "dependencies": ["-_","213","---_--_"]
}
```
will return an error sequence, which looks very similar to the one we are already used to:
```json
{
  "errors": [
    {
      "attribute": "dependencies[0]",
      "error": "must match regular expression '^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$'"
    },
    {
      "attribute": "dependencies[2]",
      "error": "must match regular expression '^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$'"
    }
  ]
} 
```
How do you feel about the new way of validating objects? You could have a look at e.g. `AppsResource.scala` in order to see the new validation in action.


src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
```diff
@@ -36,4 +42,76 @@ object V2Group {
...
+
+    group is validPorts
+  }
+
+  def v2GroupWithConfigValidator(conf: MarathonConf)
+                                (implicit validator: Validator[V2Group]): Validator[V2Group] = {
+    new Validator[V2Group] {
+      override def apply(group: V2Group): Result = {
+        conf.maxApps.get.map { num =>
+          RuleViolation(group,
```
aquamatthias: Shouldn't you compare num to the number of apps ?

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
```diff
@@ -36,4 +42,76 @@ object V2Group {
       groups = group.groups.map(V2Group(_)),
       dependencies = group.dependencies,
       version = group.version)
+
+  implicit val v2GroupValidator: Validator[V2Group] = validator[V2Group] { group =>
+    group.id is valid
+    group.apps is valid
+    group.groups is valid
+    group.groups is noAppsAndGroupsWithSameName(group.apps)
```
aquamatthias: Should be group is noAppsAndGroupsWithSameName(group.apps), since group.groups will do the recursion.
Why do you have to pass group.apps?

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
```diff
@@ -36,4 +42,76 @@ object V2Group {
...
+                |(Override with command line option --max_apps)""".stripMargin,None)
+        } match {
+          case Some(v) => Failure(Set(v))
+          case _ => Success
+        }
+      } and validator(group)
+    }
+  }
+
+  private def noAppsAndGroupsWithSameName(apps: Set[V2AppDefinition]): Validator[Set[V2Group]] =
```
aquamatthias: remove argument.

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
```diff
@@ -36,4 +42,76 @@ object V2Group {
       groups = group.groups.map(V2Group(_)),
       dependencies = group.dependencies,
       version = group.version)
+
+  implicit val v2GroupValidator: Validator[V2Group] = validator[V2Group] { group =>
+    group.id is valid
+    group.apps is valid
+    group.groups is valid
+    group.groups is noAppsAndGroupsWithSameName(group.apps)
+    group.dependencies is noCyclicDependencies(group)
```
aquamatthias: Minor: checking this in the root group should be sufficient.

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
```diff
@@ -58,4 +63,28 @@ object V2GroupUpdate {
...
+
+    group.id is optional(PathId.validChild(group.id.map(_.canonicalPath(PathId.empty)).getOrElse(PathId.empty)))
+    group.apps is valid
+    group.groups is valid
+  }
+
+  def hasOnlyOneDefinedOption[A <: Product: ClassTag, B]: Validator[A] =
+  new Validator[A] {
+    def apply( product: A ) = {
+      if (product.productIterator.count {
```
aquamatthias: assign the count to a separate val

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
```diff
@@ -58,4 +63,28 @@ object V2GroupUpdate {
...
+        case Some(_) => true
+        case _ => false
+      } <= 1)
```
aquamatthias: `== 1` ?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -431,4 +432,12 @@ object AppDefinition {
 
   def fromProto(proto: Protos.ServiceDefinition): AppDefinition =
     AppDefinition().mergeFromProto(proto)
+
+  import com.wix.accord.dsl._
```
aquamatthias: move the import to the top

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -157,7 +162,10 @@ class GroupManager @Singleton @Inject() (
       from <- rootGroup()
       (toUnversioned, resolve) <- resolveStoreUrls(assignDynamicServicePorts(from, change(from)))
       to = GroupVersioningUtil.updateVersionInfoForChangedApps(version, from, toUnversioned)
-      _ = BeanValidation.requireValid(ModelValidation.checkGroup(to, "", PathId.empty, config.maxApps.get))
+      _ = {
+        // TODO AW: test
+        validateOrThrow(V2Group(to))(validator = V2Group.v2GroupWithConfigValidator(config))
```
aquamatthias: why V2Group?

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -88,4 +88,37 @@ object PathId {
...
+    pathId.path.each should matchRegex(ID_PATH_SEGMENT_PATTERN.pattern)
+    // pathId is validChild(pathId.parent)
+  }
+
+  import com.wix.accord._
+  import scala.util.Try
+  def validChild(parent: PathId): Validator[PathId] = {
+    new Validator[PathId] {
+      override def apply(child: PathId): Result = {
+        Try(child.canonicalPath(parent)) match {
```
aquamatthias: `parent` should always be an absolute path (an exception is thrown if not). Handling that case here (and also catching all exceptions) I would try to avoid.

src/test/scala/mesosphere/marathon/api/validation/V2AppDefinitionValidatorTest.scala
```diff
@@ -1,5 +1,6 @@
 package mesosphere.marathon.api.validation
-
+// TODO AW: test
```
aquamatthias: reenable...

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -64,26 +68,29 @@ class AppsResource @Inject() (
...
-        maybeDeployments = Some(Seq(Identifiable(plan.id)))
-      )
-
-      maybePostEvent(req, appWithDeployments.app)
-      Response
-        .created(new URI(app.id.toString))
-        .entity(jsonString(appWithDeployments))
-        .build()
+    withValid(Json.parse(body).as[V2AppDefinition].withCanonizedIds()) { appDef =>
+      withValid(appDef.toAppDefinition) { app =>
```
aquamatthias: why nested withValid?

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
alexanderweber: I think there is no recursion in here. `noAppsAndGroupsWithSameName` Validator does not execute an implicit validator of type `V2Group`. It takes in `group.groups` as the `apply` argument and compares it against `group.apps`. Correct me if I am wrong.

The only reason why I build a validator like that is, that the property name `groups` could be resolved in the failure message. Instead I could do a `group is noAppsAndGroupsWithSameName` (apps can be derived from `group.apps` within the validator) and explicitly put the property name `groups` into the `RuleValidation`'s `description` property. This is the way I had it first, though automatically derived property name seemed a better choice. Should I change it to `group is noAppsAndGroupsWithSameName` and explicitly provide the property name?

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
alexanderweber: Do you mean `apps: Set[V2AppDefinition]`? See explanation in previous comment.

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
alexanderweber: Good point. Should the root group still always be checked when an arbitrary `V2Group` is validated? Or should this one be extracted in a separate Validator and explicitly be called when appropriate?

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
alexanderweber: It seemed like a valid solution instead of checking `group.version.isDefined` and then calling the `hasOnlyOneDefinedOption`. I did it because I could not simply make an `if(group.version.isDefined) { group.version is hasOnlyOneDefinedOption }`. There is an implicit `and` conjunction between the rules. Is `<=` not a valid work around?

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
alexanderweber: I committed this change.

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
alexanderweber: Thanks, you are right. I committed a change on that.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
alexanderweber: Originally `ModelValidation.checkGroup` looked like this:

```scala
// TODO: Re-implement this method on it's own terms
  def checkGroup(
    group: Group,
    path: String,
    parent: PathId,
    maxApps: Option[Int]): Iterable[ConstraintViolation[V2Group]] = {

    val v2Group = V2Group(group)
    val maximumApps = maxApps.filter(group.transitiveApps.size > _).map{ num =>
      Seq(violation(v2Group, "apps", "",
        s"This Marathon instance may only handle up to $num Apps! (Override with command line option --max_apps)"))
    }.getOrElse(Seq.empty[ConstraintViolation[V2Group]])
    checkGroup(v2Group, path, parent) ++ maximumApps
  }
```
Besides the maxApps check, the main validation is done with the V2Group object in a separate `checkGroup` function. I decided to have only the V2Group validator as it contains all the major validation.

Alternatively I could make a `Group` validator, cast the group to a `V2Group` in there and call its validator. Seems actually like the right thing to do. This way the caller could validate an `Group` without explicitly transforming it to the `V2Group` (the transformation will be done within the Group validator)

src/test/scala/mesosphere/marathon/api/validation/V2AppDefinitionValidatorTest.scala
alexanderweber: Yes, reenabling the tests is the next step I wanted to do. Until then there should be no merge.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
alexanderweber: You are right, the validators are identical. Fix committed.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
aquamatthias: Understand. We need to get rid of V2(Group|App) asap

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
aquamatthias: Just a performance consideration - ignore in the moment.

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
aquamatthias: The recursion is in group.groups or?
I still think you do not have to pass apps and make the validator a val. Obviously I miss something. Let's discuss f2f

aquamatthias:
What do you think @kolloch @meichstedt @gkleiman ??

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
alexanderweber: Change commited

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
alexanderweber: Change commited.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
kolloch: Right. @alexanderweber Can you do that in a separate PR as a preparation? Thanks!

https://github.com/mesosphere/marathon/issues/2426

kolloch:
@alexanderweber I just read your nice summary and it looks very good.

You say that the new error messages "look very similar". Does that mean they contain the same JSON fields in error responses, just that the messages are slightly different? We should only add fields to the JSON wherever possible to keep compatibility.

Furthermore, I would like to ensure that the following things are possible (I would expect it but better be certain). You don't have to address them at once in the patch, I am more interested in feasibility.

* Error codes along with error messages: Currently, our error reporting sucks. I would much like to report an error code along with the error message. That could be the I18N text ID. It would also be nice to have a map of named arguments for the I18N template in the JSON, e.g.:

```javascript
{
  "errors": [
    {
      "attribute": "dependencies.[0].path",
      "errorCode": "string.maxLengthExceeded", // just a silly example
      "errorArgs": {
          "value": "/this/is/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/long/for/a/path",
         "maxLength": 256
      },
      "error": "The value '/this/is/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/really/long/for/a/path' exceed the maximum length of 256 characters."
    }
  ]
}

```

That design allows automated parsing of errors and also allows (optionally) I18n by the frontend.

* How can we make sure that the fieldNames in our messages correspond to field names in our JSON?

* "Transforming" the error paths according to other API versions: We may have to support multiple API versions in the same code base that use different JSON attribute names or even a different structure.

alexanderweber:
@kolloch Hey Peter, I did mean exactly that - JSON field names will stay the same, but the content of messages might be slightly different.

* I am not sure about the error codes at this point to be honest. Violation classes of the Accord framework expect Strings as error messages. I was actually talking to @aquamatthias about related topic: if we could match a violation against its type instead of the message string in our testing routines. An error code would do the job there. But I need to think about a possible realisation.

* The correspondence is provided by property names in validated classes. The framework figures out the property name through macros. As long as property names of the case class correspond to the JSON field names (which is the case as we are using play json, correct me if I am wrong), everything should correspond correctly.

* That might be a problem, but I can still imagine it being possible, as field names can be explicitly be set during the validation as well.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
alexanderweber: @kolloch Will do. I assigned #2426 to myself for now. @aquamatthias do you mind me trying to get rid of V2?

alexanderweber:
@aquamatthias I am constantly getting this message in TeamCity:
`[error] /marathon/src/main/scala/mesosphere/marathon/api/RestResource.scala:55: File line length exceeds 120 characters`
The line is definitely less than 120 characters long. IntelliJ is not complaining as well. I already tried `scaliest:off method.length`, but that did not help out. Do you know the reason for that?

src/main/scala/mesosphere/marathon/state/GroupManager.scala
alexanderweber: @kolloch #2836 is the pull request regarding V2 deletion.

mwasn:
@kolloch Regarding error codes: The lack of error codes respectively missing documentation of those has actually been a demand of a customer. I´m going to create a separate story for implementation and documentation.

alexanderweber:
Merged master into this branch. @aquamatthias Please get back to me if you are ready to merge this pull request.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -127,18 +133,18 @@ class AppsResource @Inject() (
...
-
-      val now = clock.now()
-      val plan = result(groupManager.updateApp(appId, updateOrCreate(appId, _, appUpdate, now), now, force))
-
-      val response = plan.original.app(appId)
-        .map(_ => Response.ok())
-        .getOrElse(Response.created(new URI(appId.toString)))
-      maybePostEvent(req, plan.target.app(appId).get)
-      deploymentResult(plan, response)
+    // TODO AW: test
```
aquamatthias: Forgot to remove comment or test?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -208,8 +215,9 @@ class AppsResource @Inject() (
                              existing: Option[AppDefinition],
                              appUpdate: V2AppUpdate,
                              newVersion: Timestamp): AppDefinition = {
-    def createApp() = validateApp(appUpdate(AppDefinition(appId)))
-    def updateApp(current: AppDefinition) = validateApp(appUpdate(current))
+    // TODO AW: test
```
aquamatthias: Forgot to remove comment or test?

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
```diff
@@ -136,21 +137,22 @@ class GroupsResource @Inject() (
...
     doIfAuthorized(req, resp, UpdateAppOrGroup, id.toRootPath) { implicit identity =>
-      val update = Json.parse(body).as[V2GroupUpdate]
-      BeanValidation.requireValid(ModelValidation.checkGroupUpdate(update, needsId = false))
-      if (dryRun) {
-        val planFuture = groupManager.group(id.toRootPath).map { maybeOldGroup =>
-          val oldGroup = maybeOldGroup.getOrElse(Group.empty)
-          Json.obj(
-            "steps" -> DeploymentPlan(oldGroup, update.apply(V2Group(oldGroup), Timestamp.now()).toGroup()).steps
-          )
+      // TODO AW: test
```
aquamatthias: Forgot to remove comment or test?

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
```diff
@@ -100,14 +100,15 @@ class GroupsResource @Inject() (
...
     doIfAuthorized(req, resp, CreateAppOrGroup, id.toRootPath) { implicit identity =>
-      val update = Json.parse(body).as[V2GroupUpdate]
-      BeanValidation.requireValid(ModelValidation.checkGroupUpdate(update, needsId = true))
-      val effectivePath = update.id.map(_.canonicalPath(id.toRootPath)).getOrElse(id.toRootPath)
-      val current = result(groupManager.rootGroup()).findGroup(_.id == effectivePath)
-      if (current.isDefined)
-        throw ConflictingChangeException(s"Group $effectivePath is already created. Use PUT to change this group.")
-      val (deployment, path, version) = updateOrCreate(id.toRootPath, update, force)
-      deploymentResult(deployment, Response.created(new URI(path.toString)))
+      // TODO AW: test
```
aquamatthias: Forgot to remove comment or test?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        if (violations.isEmpty) Success
+        else Failure(Set(GroupViolation(seq, "seq contains elements, which are not valid", None, violations.toSet)))
+      }
+    }
+  }
+
+  implicit lazy val failureWrites: Writes[Failure] = Writes { f =>
+    // TODO AW: get rid of toSeq
```
aquamatthias: Forgot to remove comment ?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  implicit lazy val ruleViolationWrites: Writes[RuleViolation] = Writes { v =>
+    Json.obj(
+      "attribute" -> v.description,
+      "error" -> v.constraint
+    )
+  }
+
+  implicit lazy val groupViolationWrites: Writes[GroupViolation] = Writes { v =>
+    // TODO AW: get rid of toSeq
```
aquamatthias: Forgot to remove comment ?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -0,0 +1,139 @@
...
+        case 1 => violationToJsValue(v.children.head, v.description)
+        case _ => JsArray(v.children.toSeq.map(c =>
+          violationToJsValue(c, v.description, parentSeq = true)
+        ))
+      }
+    }
+  }
+
+  private def concatPath(parent: String, child: Option[String], parentSeq: Boolean): String = {
+    // TODO AW: fix not point in array issue
```
aquamatthias: Forgot to remove comment?

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
```diff
@@ -2,17 +2,24 @@ package mesosphere.marathon.api.v2.json
...
 import mesosphere.marathon.state._
+import mesosphere.marathon.state.PathId._
 import org.apache.mesos.{ Protos => mesos }
 
 import scala.collection.immutable.Seq
 import scala.concurrent.duration._
 
+import com.wix.accord.dsl._
+
+// TODO AW: delete this one
```
aquamatthias: Forgot to remove comment or action?

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -155,7 +159,10 @@ class GroupManager @Singleton @Inject() (
       from <- rootGroup()
       (toUnversioned, resolve) <- resolveStoreUrls(assignDynamicServicePorts(from, change(from)))
       to = GroupVersioningUtil.updateVersionInfoForChangedApps(version, from, toUnversioned)
-      _ = BeanValidation.requireValid(ModelValidation.checkGroup(to, "", PathId.empty, config.maxApps.get))
+      _ = {
+        // TODO AW: test
```
aquamatthias: Forgot to remove comment or action?

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
```diff
@@ -20,8 +23,8 @@ class ModelValidationTest
   test("A group update should pass validation") {
     val update = V2GroupUpdate(id = Some("/a/b/c".toPath))
 
-    val violations = ModelValidation.checkGroupUpdate(update, true)
-    violations should have size 0
+    // TODO AW: test
```
aquamatthias: Forgot to remove comment or action?

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
```diff
@@ -31,49 +34,45 @@ class ModelValidationTest
       createServicePortApp("/c".toPath, 0).toAppDefinition
     ))
 
-    val validations = ModelValidation.checkGroup(group, "", PathId.empty, maxApps = Some(2))
-    validations should not be Nil
-    validations.find(_.getMessage.contains("This Marathon instance may only handle up to 2 Apps!")) should be ('defined)
+    // TODO AW: test
```
aquamatthias: Forgot to remove comment or action?

alexanderweber:
@aquamatthias I committed last relevant changes. Could you have a look at the PR again?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -24,6 +25,9 @@ import play.api.libs.json.Json
 import scala.collection.immutable.Seq
 import scala.concurrent.Future
 
+// Accord validation
```
aquamatthias: unnecessary comment

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -0,0 +1,123 @@
...
+        else Failure(Set(RuleViolation(seq, "Elements must be unique", None)))
+      }
+    }
+  }
+
+  case class ViolationMessageAndProperty(message: String, property: Option[String]) {
+    override def toString: String = s"Property: $property Message: $message"
+  }
+
+  def getAllRuleConstrains(r: Result): Seq[ViolationMessageAndProperty] = {
```
aquamatthias: Is that method only be used in tests? If yes: can we move it to test scope?

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
```diff
@@ -2,17 +2,23 @@ package mesosphere.marathon.api.v2.json
...
 import mesosphere.marathon.state._
+import mesosphere.marathon.state.PathId._
 import org.apache.mesos.{ Protos => mesos }
 
 import scala.collection.immutable.Seq
 import scala.concurrent.duration._
 
+import com.wix.accord.dsl._
+
 @PortIndices
```
aquamatthias: Do we still need this?

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
```diff
@@ -2,17 +2,23 @@ package mesosphere.marathon.api.v2.json
...
 @ValidV2AppDefinition
```
aquamatthias: Same here. Also the whole package mesosphere.marathon.api.validation can be removed?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -0,0 +1,123 @@
...
+          val desc = concatPath(p, g.description, attachDot)
+          allRuleViolationsWithFullDescription(c, Some(desc), g.description.isDefined)
+        } getOrElse {
+          allRuleViolationsWithFullDescription(c, g.description, g.description.isDefined)
+        }
+      }
+    }
+  }
+
+  def urlsCanBeResolvedValidator: Validator[String] = {
```
aquamatthias: s/urls/url (no plural)

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
```diff
@@ -128,4 +134,47 @@ object V2AppDefinition {
       upgradeStrategy = app.upgradeStrategy, labels = app.labels, acceptedResourceRoles = app.acceptedResourceRoles,
       ipAddress = app.ipAddress, version = app.version, versionInfo = maybeVersionInfo)
   }
+
+  implicit val appDefinitionValidator = validator[V2AppDefinition] { appDef =>
```
aquamatthias: Clean and easy to understand. Nice!

src/main/scala/mesosphere/marathon/api/v2/json/V2Group.scala
```diff
@@ -36,4 +39,72 @@ object V2Group {
       groups = group.groups.map(V2Group(_)),
       dependencies = group.dependencies,
       version = group.version)
+
+  implicit val v2GroupValidator: Validator[V2Group] = validator[V2Group] { group =>
+    group.id is valid
+    group.apps is valid
+    group.groups is valid
+    group is noAppsAndGroupsWithSameName
+    (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
```
aquamatthias: :+1: Cool!

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
```diff
@@ -58,4 +63,53 @@ object V2GroupUpdate {
     V2GroupUpdate(Some(id), if (apps.isEmpty) None else Some(apps), if (groups.isEmpty) None else Some(groups))
   }
   def empty(id: PathId): V2GroupUpdate = V2GroupUpdate(Some(id))
+
+  implicit val v2GroupUpdateValidator: Validator[V2GroupUpdate] = validator[V2GroupUpdate] { group =>
+    group is notNull
+
+    group.version is isTheOnlyDefinedOptionIn(group)
```
aquamatthias: Can you rename isTheOnlyDefineIption into theOnlyDefinedOption
Will read like: 
```scala
group.version is theOnlyDefinedOptionIn(group)
```

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
```diff
@@ -58,4 +63,53 @@ object V2GroupUpdate {
...
+    group.scaleBy is isTheOnlyDefinedOptionIn(group)
+
+    group.id is valid
+    group.apps is valid
+    group.groups is valid
+  }
+
+  def hasOnlyOneDefinedOption[A <: Product: ClassTag, B](prop: Option[B]): Validator[A] =
```
aquamatthias: Is not specific to V2GroupUpdate and can be moved to Validation?

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
```diff
@@ -58,4 +63,53 @@ object V2GroupUpdate {
...
+            if (n == 1)
+              Success
+            else
+              Failure(Set(RuleViolation(prop, s"not allowed in conjunction with other properties.", None)))
+          case None => Success
+        }
+      }
+    }
+
+  def isTheOnlyDefinedOptionIn[A <: Product: ClassTag, B](product: A): Validator[Option[B]] =
```
aquamatthias: Move to Validation?

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -155,7 +159,9 @@ class GroupManager @Singleton @Inject() (
       from <- rootGroup()
       (toUnversioned, resolve) <- resolveStoreUrls(assignDynamicServicePorts(from, change(from)))
       to = GroupVersioningUtil.updateVersionInfoForChangedApps(version, from, toUnversioned)
-      _ = BeanValidation.requireValid(ModelValidation.checkGroup(to, "", PathId.empty, config.maxApps.get))
+      _ = {
+        validateOrThrow(V2Group(to))(validator = V2Group.v2GroupWithConfigValidator(config.maxApps.get))
+      }
```
aquamatthias: 1) Remove the curly braces.
2) Can we create a validator in Group that delegates to V2Group (until we remove it)
```scala
validateOrThrow(to)(Group.groupValidator(config.maxApps())
```
```

File: src/main/scala/mesosphere/marathon/state/PathId.scala [hunk 34]
```diff
@@ -88,4 +91,43 @@ object PathId {
...
+  }
+
+  /**
+    * Check if path's parent, if it exists, is a valid parent indeed.
+    * @return
+    */
+  def validChild: Validator[PathId] = {
+    new Validator[PathId] {
+      override def apply(pathId: PathId): Result = {
+        if (pathId.parent == "".toPath) Success
```
aquamatthias: s/"".toPath/empty

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -88,4 +91,43 @@ object PathId {
...
+        else {
+          if (pathId.parent.absolute) {
```
aquamatthias: make it part of the outer if/else
if (root)
else if (absolute)
else //relative

src/test/scala/mesosphere/marathon/api/v2/json/V2AppDefinitionTest.scala
```diff
@@ -17,50 +16,58 @@ import play.api.libs.json.{ JsError, Json }
...
+          )
+      }
     }
 
     var app = V2AppDefinition(id = "a b".toRootPath)
-    val idError = "path contains invalid characters (allowed: lowercase letters, digits, hyphens, \".\", \"..\")"
+    val idError = "must fully match regular expression '^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$'"
     validateJsonSchema(app, false)
-    shouldViolate(app, "id", idError)
+    shouldViolate(app, "id.path", idError)
```
aquamatthias: Probably minor - we already talked about that:
the user defines an id -> we give him a response for id.path

aquamatthias:
@alexanderweber Thanks for this PR. It really looks pretty and I like it.
I added some minor suggestions.
I think also that we should remove the old JSR validation logic, since it is not used any longer (mesosphere.marathon.api.validation)

wendigo:
Nice one :) This will simplify validation I've added in other issue I'm working on. When will this be merged to master so I can rebase against it?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
alexanderweber: Do you mean `elementsAreUnique` method? For the time being it is only used once in `V2AppDefinition`. I am moving it there in the following commit. Just for future reference - should I avoid defining general Validators?

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
alexanderweber: From my point of view we do. Let me give an example in the mentioned class:
```scala
  implicit val appDefinitionValidator = validator[V2AppDefinition] { appDef =>
    appDef.id is valid
    appDef.dependencies is valid
    appDef.upgradeStrategy is valid
    appDef.storeUrls is every(urlsCanBeResolvedValidator)
    appDef.ports.filterNot { _ == AppDefinition.RandomPortValue } as "ports" is elementsAreUnique
    appDef.executor should matchRegexFully("^(//cmd)|(/?[^/]+(/[^/]+)*)|$")
    appDef is containsCmdArgsContainerValidator
    appDef is portIndicesAreValid
    appDef.instances.intValue should be >= 0
  }
```
Here we have a predefined validator `matchRegexFully`. It could of course be written by us as well, but I found it convenient to use the provided one.
Functions like `is` or `should` are also part of the package. In order to define our class validators, we need to import that package. An alternative would be defining that class validators outside of the class file. We decided to do it this way back then, but I don't mind thinking about alternatives again.

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
alexanderweber: It is the same file as above. You probably meant some other class, like V2Group or something. Same reasons go for that ones, as I described in the previous one.

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
alexanderweber: Thanks!

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
alexanderweber: Makes sense!

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
alexanderweber: I moved it to `mesosphere.marathon.api.v2.ValidationHelper`. Is that one ok?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
alexanderweber: Committed that.

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
alexanderweber: Yep, committed that.

src/main/scala/mesosphere/marathon/api/v2/json/V2AppDefinition.scala
alexanderweber: @aquamatthias Removing annotation validation, I realised, that there is also a HealthCheck validation. I implemented that one. Still there is a comment which caught my attention:
```scala
/**
    * FIXME: This is not called when an V2AppDefinition is called because the recursion apparently doesn't work
    * with Scala Seq.
    *
    * We cannot re-enable it because it would break backwards compatibility in weird ways. If users hard already
    * one invalid app definition, each deployment would cause a complete revalidation of the root group
    * including the invalid one. Until the user changed all invalid apps, the user would get weird validation
    * errors for every deployment potentially unrelated to the deployed apps.
    */
```
That comment is regarding HealthCheck validation. Could you get back to me on that?

src/main/scala/mesosphere/marathon/state/GroupManager.scala
alexanderweber: For the time being, I committed that forwarding.

src/main/scala/mesosphere/marathon/state/PathId.scala
alexanderweber: Check

src/test/scala/mesosphere/marathon/api/v2/json/V2AppDefinitionTest.scala
alexanderweber: This one was a bit trickier than I thought. There is a `PathId` Validator, which I am calling from `V2AppDefinition` on `id` property. That causes it to be a sub-violation of `V2AppDefinition` validation, which is internally correct. The problem is, that the JSON parser traverses down the violation tree and there is no way to mark `PathId`'s violation descriptions as 'should be ignored'. What I had to do is to extract the validation functionality out of PathId's validator and call it manually on V2AppDefinition's `id` property:
```scala
implicit val appDefinitionValidator = validator[V2AppDefinition] { appDef =>
    appDef.id is validPath
```
`validPath` being a public validation function defined in `PathId`. The nicer former version was:
```scala
implicit val appDefinitionValidator = validator[V2AppDefinition] { appDef =>
    appDef.id is valid
```
which was using the implicit validator for `PathId` classes. Are you fine with that change? You can look it up in `V2AppDefinition` and `PathId` classes of `Extracted PathId validation into a function for external usage. Needed to avoid drilling down the violation chain.` commit.

src/main/scala/mesosphere/marathon/state/PathId.scala
alexanderweber: This change caused an integration test error. `PathId.empty` is not equal to `"".toPath`. The first produces a path id having `absolute = true`, the former a path id having `absolute = false`. I changed it back to `"".toPath` if you don't have another suggestion.

alexanderweber:
@aquamatthias Peter was faster than me with his PR yesterday, so I had to merge master into my branch again. Now everything is squashed into 45e0152. All checks look fine.

alexanderweber:
@wendigo Nice to hear. Let me know if you'll have any issues using the new approach. I am hoping for a merge during today.

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
aquamatthias: Mmmh - I still see it here...

src/main/scala/mesosphere/marathon/health/HealthCheck.scala
```diff
@@ -122,4 +122,36 @@ object HealthCheck {
   val DefaultMaxConsecutiveFailures = 3
   val DefaultIgnoreHttp1xx = false
   val DefaultPort = None
+
+  implicit val healthCheck = validator[HealthCheck] { hc =>
+    (hc.portIndex.nonEmpty is true) or (hc.port.nonEmpty is true)
+    hc is validProtocol
+  }
+
+  /**
```
aquamatthias: This comment here is:
- outdated: since the problem with the seq is not valid any longer
- in the wrong place: it probably should be moved to V2AppUpdate.appDefinitionValidator, where we _don't_ do `appDef.healthChecks is valid`

alexanderweber: Regarding your second remark: do you mean the *comment* should be moved to appDefintionValidator?

src/main/scala/mesosphere/marathon/api/v2/json/V2GroupUpdate.scala
alexanderweber: Fixed in c4cb9e2.

wendigo:
@alexanderweber sure thing!

src/test/scala/mesosphere/marathon/api/validation/V2AppDefinitionValidatorTest.scala
```diff
@@ -1,234 +0,0 @@
-package mesosphere.marathon.api.validation
```
aquamatthias: Hey @alexanderweber sorry that I did not mention that before: why did you remove the whole test class here? Would it not make sense to use the new validation logic here as well?

alexanderweber: @aquamatthias Sorry for late reply. As I was deleting the whole `validation` folder with annotation validation, I did not pay attention, that there was also one test included in there. Sorry for that, I added it back again adjusting it to new validation model.

aquamatthias:
@alexanderweber: Thanks. Please rebase.

aquamatthias:
This is a great addition. Thanks!

wendigo:
Nice! :)

holograph:
Hi everyone, sorry to raise the dead but I missed the original discussion when it was still ongoing. I'm the maintainer of Accord, and would absolutely love your input on how to add these features cleanly!

In particular, Accord is currently quite naïve about how it handles violations, in that it simply encodes them as strings. There's an open issue (wix/accord#21) that deals with the various implementation options, but lacking external input I was never sufficiently happy with any of them. It's a somewhat long-winded discussion, but if you have some fresh ideas on what this should look like (e.g. how to maintain type safety and/or offer a clean DSL), please do weigh in! There's also a [Gitter channel](https://gitter.im/wix/accord) if you prefer the more personal touch.

Lastly, I'm curious about the "path" generation/transformation properties mentioned by @kolloch; Accord automatically generates a string description for each "thing" being validated. A "thing" could be property access, modifier (`.each` or `.size`) or an arbitrary expression, and in all cases Accord does its best to figure out a good representation of the inspected property. This can be overridden per-rule using `as`, but the description generation mechanism is not yet pluggable, is fairly naïve, and could use a lot of improvement. Our own services don't normally externalize validation errors, and so again I would dearly appreciate input on how this could be modelled.

Thank you for trying out Accord, I hope you've been enjoying it so far!

nnordrum:
This also appears to have broken relative paths in dependencies, or at least changed how they are represented. 
```
"dependencies":["../hub"]
```
I have this in my deployment of a selenium grid, and this worked before 0.15.3, but doesn't now.

How do you represent this relatively now? All the examples and tests seem to be full path, which I would rather not do for what I think would think would be relatively obvious reasons.

PR #2817, opened 2015-12-14
Fixes #2620 - Remove TaskTracker.expungeOrphanedTasks

### LINKED ISSUE #2620: Remove dubious functionality to remove orphaned tasks from TaskTracker storage
`TaskTracker.expungeOrphanedTasks` removes tasks from (ZK-) storage if they are not present in memory. It is called once after startup – hopefully after reconciliation has finished, but this is only controlled via a delay.

If `TaskTracker.expungeOrphanedTasks` id called before all apps are loaded/reconciled (e.g. when reconciliation takes longer than expected), non-orphaned tasks will be removed from storage by mistake.

@kolloch and I cannot think of a use case where this method makes things better, so we propose to remove it.

kolloch:


aquamatthias:
:+1: Please rebase.

kolloch:
Rebased

PR #2819, opened 2015-12-14
Fixes #2818 - Remove TaskTracker.get returning Set[MarathonTask]

### LINKED ISSUE #2618: Consolidate task update logic in TaskTracker

### LINKED ISSUE #2818: Remove `Set[MarathonTask]` from TaskTracker
The TaskTracker and related code uses `Set[MarathonTask]` for various tasks. This is nearly never wanted and has been found to be a performance issue in many cases.

kolloch:
and refactor related code that uses Set[MarathonTask].

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -407,7 +407,7 @@ class SchedulerActions(
     healthCheckManager.removeAllFor(app.id)
 
     log.info(s"Stopping app ${app.id}")
-    val tasks = taskTracker.get(app.id)
+    val tasks = taskTracker.getTasks(app.id)
 
     for (task <- tasks) {
       log.info(s"Killing task ${task.getId}")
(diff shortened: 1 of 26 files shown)


aquamatthias:
One Question: we have a lot of places where we use Seq. You now introduced a lot of Iterable. Would it be a problem to use a Seq instead of Iterable or is there any downside?

kolloch:
Hi @aquamatthias, 

"we have a lot of places where we use Seq"

do you mean in general in our code or in relation to the `TaskTracker`?

The difference between `Seq` and `Iterable` in Scala is mostly that a `Seq` has a defined order, an `Iterable` not. The order of the tasks is not important in this case, so `Iterable` is fine and `Seq` is too specific in this case.

In many places in our code, `Iterable` would actually be fine instead of `Seq`. I feel that we should use it more rather than less where order does not matter, especially in method arguments (as opposed to fields in case classes where having a defined order usually is appreciated).

The main technical reason that I used "Iterable" was that this is the type returned by `MapLike[...].values`. All other types would require a translation in that method resulting in unnecessary wrapping. Furthermore, I do not need to change the tests so much because `Set` is also an `Iterable` but not a `Seq`.

aquamatthias:
Got it. Thanks!

PR #1455, opened 2015-05-04
Check-in mesos.proto file.

kolloch:
This ensures that protoc-calls are reproducible. Otherwise, different
developers might use different mesos.proto versions.

Issue #1445, opened 2015-04-30
Multiple health checks


dsiegel:
It appears that when I have multiple health checks, the deployment of a task will complete as soon as a single health check succeeds rather than waiting for all health checks to succeed. That means the old task will be killed before the new task has reached a healthy state.

Is this expected behavior? I'd rather it wait for all health checks to succeed.

I've observed this through the UI, but I can't see it in the logs, because I don't see the log statements from the HealthCheckActor.

Here is the JSON I am posting to marathon:

```
{
    "backoffFactor": 2,
    "backoffSeconds": 10,
    "container": {
        "docker": {
            "image": "mytestcontainer",
            "network": "BRIDGE",
            "portMappings": [
                {
                    "containerPort": 22925,
                    "hostPort": 0,
                    "protocol": "tcp"
                },
                {
                    "containerPort": 13001,
                    "hostPort": 0,
                    "protocol": "tcp"
                }
            ]
        },
        "type": "DOCKER"
    },
    "cpus": 1,
    "healthChecks": [
        {
            "gracePeriodSeconds": 90,
            "intervalSeconds": 20,
            "maxConsecutiveFailures": 3,
            "path": "/status",
            "portIndex": 0,
            "protocol": "HTTP"
        },
        {
            "gracePeriodSeconds": 90,
            "intervalSeconds": 20,
            "maxConsecutiveFailures": 3,
            "portIndex": 1,
            "protocol": "TCP"
        }
    ],
    "id": "testing",
    "instances": 1,
    "mem": 1024.0
} 
```

dsiegel:
Bump.

kolloch:
Hi @dsiegel, I agree with you that a task should succeed if all health checks are reporting success.

Have you tried using COMMAND health checks as a work around? They are more scalable anyways since they are not handled from the single marathon master instance but by the slaves.

dsiegel:
Are you suggesting a single COMMAND check, like this:

curl X && curl Y && curl Z, instead of 3 separate health checks?

Seems like that should work fine. 

What is the point of the marathon based health checks then? Perhaps the marathon TCP/HTTP checks should just generate COMMAND health checks?

bobrik:
Forking to curl is quite expensive operation. It'd be better to support tcp/http health checks from slave processes. That could hide problems like "everything resolves to localhost and not accessible from the outside", though.

[...]

PR #2822, opened 2015-12-14
Fixes #1455 - Multiple health checks DO NOT MERGE

meln1k:
Here is a start of process of fixing issue #1445.

The idea is to create a new actor called HealthCheckAppActor, which is responsible to aggregate failed health checks from HealthCheckActor. Based on this checks, HealthCheckAppActor should be able to correctly determine if an instance is working, and don't miss any failed health checks.

What should be done next: we should move the application killing logic from HealthCheckActor to HealthCheckAppActor, now it's responsible to making decisions based on task health information. Read TODOs for more information.


aquamatthias:
Thanks @meln1k - we need to improve that logic started here. I close the PR for now.

Issue #462, opened 2014-08-18
Redesign TaskTracker


drexin:
The current implementation of the `TaskTracker` has several issues that need to be fixed. After finding several races, a temporary fix added synchronization on several state changing methods, but still mutable state leaks out. See also: https://github.com/mesosphere/marathon/issues/438

Acceptance Criteria 
- [ ] task state is held in an actor
- [ ] the actor uses the WhenLeader infrastructure
- [ ] reads should be served from memory, writes propagated to Zookeeper and only confirmed when Zookeeper confirmed the write 
- [ ] acceptable performance
- [ ] easy to get consistent view of all tasks for an app and to get updates for this view (task update acknowledgements?)
- [ ] concept for task reconciliation
- [ ] think about Zookeeper consistence guarantees and what they mean in this context (probably we need a sync request on regaining leadership)

kolloch:
It should also not directly use the PersistentStore but use EntityStore.

mwasn:
Acceptance Criteria 
* task state is held in an actor
* the actor uses the WhenLeader infrastructure
* reads should be served from memory, writes propagated to Zookeeper and only confirmed when Zookeeper confirmed the write 
* acceptable performance
* easy to get consistent view of all tasks for an app and to get updates for this view (task update acknowledgements?)
* concept for task reconciliation
* think about Zookeeper consistence guarantees and what they mean in this context (probably we need a sync request on regaining leadership)

kolloch:
I opened two more specific issues with most of the acceptance criteria from this issue: https://github.com/mesosphere/marathon/issues/2916 and https://github.com/mesosphere/marathon/issues/2917 

These two acceptance criteria got lost on purpose:

- [ ] easy to get consistent view of all tasks for an app and to get updates for this view (task update acknowledgements?)
- [ ] concept for task reconciliation

@aquamatthias or others: Do you agree that we don't need them anymore? Otherwise, please open new issues that outline what needs to be done.

PR #2830, opened 2015-12-16
Toward #462 - create interface traits for TaskTracker

kolloch:
..to ease later replacement of the implementation.

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/TaskCreator.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/TaskCreator.scala
@@ -1,0 +1,24 @@
+package mesosphere.marathon.tasks
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.state.PathId
+
+import scala.concurrent.Future
+
+/**
+  * Notifies the [[TaskTracker]] of task creation and termination.
+  */
+trait TaskCreator {
+  /**
+    * Create a new task.
+    *
+    * If the task exists already, the returned Future will fail.
+    */
+  def created(appId: PathId, task: MarathonTask): Future[MarathonTask]
+
+  /**
+    * Remove the task for the given app with the given ID completely.
+    *
+    * If the task does not exist, the returned Future will not fail.
+    */
+  def terminated(appId: PathId, taskId: String): Future[_]
+}
--- a/src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
@@ -1,259 +1,32 @@
 package mesosphere.marathon.tasks
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.state.{ Timestamp, PathId }
 
-import javax.inject.Inject
-
-import mesosphere.marathon.MarathonConf
-import mesosphere.marathon.Protos._
-import mesosphere.marathon.state.{ PathId, TaskRepository, Timestamp }
-import org.apache.mesos.Protos.TaskState._
-import org.apache.mesos.Protos.{ TaskState, TaskStatus }
-import org.slf4j.LoggerFactory
-
-import scala.collection._
-import scala.collection.concurrent.TrieMap
-import scala.collection.immutable.Set
-import scala.concurrent.{ Await, Future }
+import scala.collection.{ Iterable, Map }
 
 /**
-  * The TaskTracker contains the latest known state for every task.
+  * The TaskTracker exposes the latest known state for every task.
   *
-  * It also processes new status information as sent by Mesos (see taskUpdate)
-  * and persist it in the associated task state.
-  *
-  * Known deficiencies:
-  *
-  * * There are race conditions in this code involving the synchronization of the repo and the cachedApps
-  *   data structure.
-  * * There are probably race conditions in checking whether to remove the entry for an app.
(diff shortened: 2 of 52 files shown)


kolloch:
@gaston: BTW the build error is probably due to d7954c17c2f9f8344afb0577d52c1389091ec61c.

I have asked @aquamatthias whether to revert it or what to do next...

kolloch:
Rebased towards the fixed version.

kolloch:
yet another rebase...

src/main/scala/mesosphere/marathon/tasks/TaskCreator.scala
```diff
@@ -0,0 +1,24 @@
...
+    * Create a new task.
+    *
+    * If the task exists already, the returned Future will fail.
+    */
+  def created(appId: PathId, task: MarathonTask): Future[MarathonTask]
+
+  /**
+    * Remove the task for the given app with the given ID completely.
+    *
+    * If the task does not exist, the method will not fail.
```
gkleiman: s/the method/the returned Future/?

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
```diff
@@ -1,259 +1,32 @@
...
-        Future.successful(())
-      }
-    }
+  def getVersion(appId: PathId, taskId: String): Option[Timestamp]
 
-    def updateTask(updatedTask: MarathonTask): Future[_] = {
-      val appId = app.appName
-      val taskId: String = updatedTask.getId
-      log.debug(s"updating task [$taskId] of app [$appId]: {}", updatedTask)
+  def list: Map[PathId, TaskTracker.App]
```
gkleiman: There are some FIXME comments in the original code and now in TaskTrackerImpl about removing `list` and `contains`. Should we also put a comment here?

src/main/scala/mesosphere/marathon/tasks/TaskTrackerImpl.scala
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  private[this] def remove(appId: PathId): Unit = {
+    cachedApps.remove(appId)
+    log.info(s"App $appId removed from TaskTracker")
+  }
+
+  // FIXME: See deficiencies above, should probably be replaced by preloading all tasks on elected
+  private[tasks] def fetchApp(appId: PathId): InternalApp = {
+    def fetchTask(taskId: String): Option[MarathonTask] = Await.result(repo.task(taskId), timeout)
```
gkleiman: The only difference between this class and the original TaskTracker is that fetchTask was moved to inside this method?

src/test/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiterActorTest.scala
```diff
@@ -65,7 +65,7 @@ class RateLimiterActorTest extends MarathonSpec {
     appRepository = mock[AppRepository]
     updateReceiver = TestProbe()
     val props = RateLimiterActor.props(
-      rateLimiter, taskTracker, appRepository, updateReceiver.ref)
+      rateLimiter, appRepository, updateReceiver.ref)
```
gkleiman: I think this now fits in one line.

src/main/scala/mesosphere/marathon/tasks/TaskTrackerImpl.scala
kolloch: And `determineOverdueTasks` was moved to another class. That's what I remember.

src/main/scala/mesosphere/marathon/tasks/TaskTracker.scala
kolloch: I don't think that that is necessary. Do you think it is? The FIXMEs relate to the implementation and not the trait, correct?

We still have the FIXMEs (and we remove TaskTrackerImpl soon).

gkleiman: Well, the comment says:

  // FIXME: remove, see documented deficiencies above

So without any context, I assume that the complete method should be removed, which would implied that it will also be removed from the trait and should be used in new code.

kolloch: Ah, I understand. Maybe we should move that list to the TaskTrackerImpl?

gkleiman:
@kolloch: LGTM.

I am still not sure if the comments regarding TaskTracker#list and TaskTracker#contains means that the methods will be removed or that they will be reimplemented. I guess that the comments hints that the method should be kept, but implemented in another way. But if the idea is not to use these methods in new code that uses TaskTracker, then we should add comments to TaskTracker. Otherwise ignore my this and merge!

This PR was a bit difficult to review. Given its size, It would have been much easier if the commit message provided some more context and a summary of the changes, such as:

- Created 3 traits out of TaskTracker: ....
- Removed the fetchTask method (and maybe a reason for this)
- Renamed shutdown to removeUnknownAppAndItsTasks
- Moved determineOverdueTasks into an actor - this change was big enough that it IMHO it should have been in a different commit/PR.

kolloch:
Thanks for reviewing this demanding PR! We'll try to do it better next time.

Currently, the FIXME methods still have to be used because there are no alternatives. The plan is to provide an alternative implementation -- see the follow up PRs.

Alright, two more changes:

* s/the method/the returned Future/
* Folded test code into one line ;)

PR #2833, opened 2015-12-16
Remove unused functionality from WhenLeaderActor

kolloch:


kolloch:
Yet another rebase...

aquamatthias:
LGTM. Why don't we need this functionality any longer? Even if we don't need it now, could it make sense to have it in the future?

kolloch:
@aquamatthias I can't predict the future. I am not sure what's best. If actors use this functionality, they are coupled to the WhenLeaderActor logic and they save only a few lines of code.

aquamatthias:
LGTM. Thanks!

PR #2836, opened 2015-12-16
#2426 get rid of v2 model objects

alexanderweber:
Using play-json, we are no longer in need for `V2AppDefinition` or `V2Group` classes anymore. This pull request should get rid of those classes.


aquamatthias:
@alexanderweber this branch has everything that is in the ModelValidation branch - correct? You can create a branch on top of another branch. That way we see only the changes for Model Objects.

alexanderweber:
@aquamatthias you are right. Should I close this one and open a new pull request having the ModelValidation replacement as base?

aquamatthias:
Yes please.

alexanderweber:
New pull request created: #2837

PR #2837, opened 2015-12-16
#2426 get rid of v2 model objects

alexanderweber:
Using play-json, we are no longer in need for V2AppDefinition or V2Group classes anymore. This pull request should get rid of those classes.


aquamatthias:
@alexanderweber will close this PR - please reopen, if it is ready.

PR #2838, opened 2015-12-16
Addition of sbt-idea plugin.

kermitas:
Hi,
I just added sbt-idea plugin. Also refactor .gitignore and plugins.sbt to make them easier to read.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,37 @@
+# === This file contains a list of files that will be ignored by GIT. 
+
+# === uncategorized
 bin/run-tests-config.sh
-.idea
-*.iml
-docker-volumes
-target
-.reviewboardrc
 dependency-reduced-pom.xml
 node_modules
-.project
-.classpath
-.settings
 .Rproj.user
 start.sh
+
+# === Review Board (https://www.reviewboard.org/) specific files
+.reviewboardrc
+
+# === Docker (https://www.docker.com/) specific files
+docker-volumes
+
+# === SBT (http://www.scala-sbt.org/) and scalac (http://www.scala-lang.org/) secific files
+target
+.project
+.history
+.cache/
+.lib/
+
+# === Scala-IDE (http://scala-ide.org/) and Eclipse (https://eclipse.org/downloads/) specific files
+.classpath
+.settings
+.scala_dependencies
+.worksheet
+
+# === IntelliJ Idea IDE (https://www.jetbrains.com/idea/) specific files
+*.iml
+.idea
+.idea_modules
+projectFilesBackup
+
+# === OS X (Mac) specific files
 .DS_Store
+
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,29 +1,42 @@
+// This file contains a list of SBT plugins.
+
+// === resolvers
 resolvers += Classpaths.typesafeReleases
-
-resolvers += "sonatype-releases" at "https://oss.sonatype.org/content/repositories/releases/"
-
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.11.2")
-
-addSbtPlugin("com.github.gseitz" % "sbt-release" % "0.8.3")
-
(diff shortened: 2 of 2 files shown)


gkleiman:
Thank you for contributing to Marathon!

The name of the commit related to .gitignore is misleading, as new entries are added as well.

It would also be good if you could follow our [Contributor Guidelines](https://mesosphere.github.io/marathon/docs/contributing.html) and either squash the commits or create one PR per commit.

I'll let others comment on the idea of adding the sbt-idea plugin. I personally use IntelliJ IDEA, but creating a project from existing sources has been always enough for me.

kermitas:
"The name of the commit related to .gitignore is misleading, as new entries are added as well." - that is right, sorry for that (few new entries were added, for ignoring Idea and Eclipse files)

"It would also be good if you could follow our Contributor Guidelines and either squash the commits or create one PR per commit." - ok, thanks for the hint, I will

"I'll let others comment on the idea of adding the sbt-idea plugin. I personally use IntelliJ IDEA, but creating a project from existing sources has been always enough for me." - for me it was crappy for a years so I am using sbt-idea; I have no idea how well Idea reads SBT's configuration (that one can put into series of *.scala and *.sbt files) but sbt-idea gives you simple `gen-idea` command that generates up to date project (with whatever was in SBT configuration)

.gitignore
```diff
@@ -1,14 +1,37 @@
...
-docker-volumes
-target
-.reviewboardrc
 dependency-reduced-pom.xml
 node_modules
+.Rproj.user
+start.sh
+
+# === Review Board (https://www.reviewboard.org/) specific files
+.reviewboardrc
```
aquamatthias: We don't use review board

aquamatthias:
Hey @kermitas thanks for this PR. I added some suggestions. Please address and I will merge.

project/plugins.sbt
```diff
@@ -1,29 +1,42 @@
+// This file contains a list of SBT plugins.
+
+// === resolvers
 resolvers += Classpaths.typesafeReleases
+resolvers += Classpaths.sbtPluginReleases
+resolvers += Resolver.sonatypeRepo("releases")
+resolvers += Resolver.sonatypeRepo("snapshots")
```
aquamatthias: We try to use only stable dependencies

project/plugins.sbt
```diff
@@ -1,29 +1,42 @@
...
+resolvers += "Era7 maven releases" at "http://releases.era7.com.s3.amazonaws.com"
 
-resolvers += "sonatype-releases" at "https://oss.sonatype.org/content/repositories/releases/"
+// === dependecy graph, https://github.com/jrudolph/sbt-dependency-graph (e.g. `dependencyGraphMl` command)
+addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.7.5")
 
+// === IntelliJ Idea project generator, https://github.com/mpeltonen/sbt-idea (e.g. `gen-idea` command)
+addSbtPlugin("com.github.mpeltonen" % "sbt-idea" % "1.7.0-SNAPSHOT")
```
aquamatthias: Such plugins are best kept as sbt global dependencies

.gitignore
kermitas: .reviewboardrc was here before my changes; if it is not needed we can remove it

project/plugins.sbt
kermitas: Even for SBT plugins (that affects only SBT)?

project/plugins.sbt
kermitas: Great, thanks for hint!
http://www.scala-sbt.org/0.13/tutorial/Using-Plugins.html#Global+plugins

project/plugins.sbt
aquamatthias: To have reproducible builds, I would say so.

kermitas: Ok.

.gitignore
aquamatthias: I was not aware - thanks.

aquamatthias:
Close this, since the original purpose seem obsolete.

PR #2839, opened 2015-12-16
Upgrade sbt plugins

kermitas:
Hi,
I just upgraded SBT plugins to their latest versions.

Unfortunately, as in every upgrade like that I can not guarantee that all will work fine for everybody.
Each plugin should be tested one by one to check if nothing was broken.


kermitas:
This pull request has more than one problem (and does not follow [Contributor Guidelines](https://mesosphere.github.io/marathon/docs/contributing.html). Closing.

PR #2859, opened 2015-12-18
Integration Tests & ProcessKeeper: Do not use duplicate process names

kolloch:
...since that leads to process leaking...


aquamatthias:
Thanks!

PR #2863, opened 2015-12-18
Fixes #2916 - Actor Implementation of TaskTracker, TaskCreator, TaskUpdater

### LINKED ISSUE #2916: Actor-based implementation of the TaskTracker
Originally part of #462.

Acceptance Criteria 
- [ ] task state is held in an actor
- [ ] the actor uses the WhenLeader infrastructure
- [ ] reads should be served from memory, writes propagated to Zookeeper and only confirmed when Zookeeper confirmed the write 
- [ ] acceptable performance

kolloch:
Done (I hope ;) ).

* Also moved the TaskTracker, TaskUpdater, TaskCreator traits
  to the core packages (towards the new implementation)
* Got rid of the recently introduced TaskReconciler because
  it is not needed anymore with the new implementation. Supposedly
  for concurrency reasons, the old implementation had a concept
  "Apps" in the TaskTracker. In the new implementation,
  tasks are grouped by app id but app ids are not retained when no
  tasks are known for the app anymore.
* Introduced async variants for all TaskTracker methods to encourage
  usage of asynchronous programming. I would like it better
  to rename the synchronous methods to "*Sync" but that would
  have conflated the size of this PR.


kolloch:
rebased and added a TaskTrackerDelegate metric

kolloch:
The test error is probably related to https://github.com/mesosphere/marathon/pull/2931

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerConfig.scala
```diff
@@ -0,0 +1,17 @@
+package mesosphere.marathon.core.task.tracker
+
+import org.rogach.scallop.ScallopConf
+
+trait TaskTrackerConfig extends ScallopConf {
+  //scalastyle:off magic.number
+
+  lazy val taskTrackerRequestTimeout = opt[Int]("task_tracker_request_timeout",
```
aquamatthias: can we rename this parameter to internalXXX so that we easily see it is internal and hidden by default?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerConfig.scala
```diff
@@ -0,0 +1,17 @@
...
+    descr = "INTERNAL TUNING PARAMETER: Timeout (in ms) for requests to the taskTracker.",
+    hidden = true,
+    default = Some(10000))
+
+  lazy val taskUpdateRequestTimeout = opt[Int]("task_update_request_timeout",
```
aquamatthias: same as above: internalXXX

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
```diff
@@ -0,0 +1,39 @@
...
+  *
+  * FIXME: To allow introducing the new asynchronous [[TaskTracker]] without needing to
+  * refactor a lot of code at once, synchronous methods are still available.
+  */
+trait TaskTracker {
+
+  def getTasks(appId: PathId): Iterable[MarathonTask]
+  def getTasksAsync(appId: PathId)(implicit ec: ExecutionContext): Future[Iterable[MarathonTask]]
+
+  def getTask(appId: PathId, taskId: String): Option[MarathonTask]
```
aquamatthias: We probably should add @deprecated here?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerModule.scala
```diff
@@ -0,0 +1,40 @@
...
+class TaskTrackerModule(
+    clock: Clock,
+    metrics: Metrics,
+    config: TaskTrackerConfig,
+    leadershipModule: LeadershipModule,
+    taskRepository: TaskRepository) {
+  lazy val taskTracker: TaskTracker = new TaskTrackerDelegate(Some(metrics), config, taskTrackerActorRef)
+
+  def taskUpdater: TaskUpdater = taskTrackerCreatorAndUpdater
+  def taskCreator: TaskCreator = taskTrackerCreatorAndUpdater
```
aquamatthias: I think I can not create tasks with the TaskCreator - what about  `TaskCreationListener` or equivalent to better match the nature?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoader.scala
```diff
@@ -0,0 +1,9 @@
+package mesosphere.marathon.core.task.tracker.impl
+import scala.concurrent.Future
+
+/**
+  * Loads all task data into an [[AppDataMap]].
+  */
+private[tracker] trait TaskLoader {
```
aquamatthias: Generally I like traits very much. Do we want to ouse them also internal object that are not used out of this context?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoaderImpl.scala
```diff
@@ -0,0 +1,32 @@
...
+  */
+private[tracker] class TaskLoaderImpl(repo: TaskRepository) extends TaskLoader {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  private[this] val log = LoggerFactory.getLogger(getClass.getName)
+
+  override def loadTasks(): Future[AppDataMap] = {
+    for {
+      names <- repo.allIds()
+      _ = log.info(s"About to load ${names.size} tasks")
```
aquamatthias: debug?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoaderImpl.scala
```diff
@@ -0,0 +1,32 @@
...
+      tasks <- Future.sequence(names.map(repo.task(_))).map(_.flatten)
+    } yield {
+      log.info(s"Loaded ${tasks.size} tasks")
```
aquamatthias: debug?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -0,0 +1,155 @@
...
+
+  override def process(op: Operation)(implicit ec: ExecutionContext): Future[Unit] = {
+    op.action match {
+      case Action.Update(task) => // used for a create or as a result from a UpdateStatus action
+        repo.store(task).map { _ =>
+          taskTrackerRef ! TaskTrackerActor.TaskUpdated(op.appId, task, TaskTrackerActor.Ack(op.sender))
+        }.recoverWith(tryToRecover(op)(expectedTaskState = Some(task)))
+      case Action.Expunge => // used for task termination or as a result from a UpdateStatus action
+        repo.expunge(op.taskId).map { _ =>
+          taskTrackerRef ! TaskTrackerActor.TaskRemoved(op.appId, op.taskId, TaskTrackerActor.Ack(op.sender))
```
aquamatthias: The action has a sender. The sender is informed for Noop/Fail but not for update /Expunge/UpdateStatus.
Why?
Perhaps add this as comment?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -0,0 +1,155 @@
...
+  */
+private[tracker] class TaskOpProcessorImpl(
+    taskTrackerRef: ActorRef,
+    repo: TaskRepository,
+    statusUpdateActionResolver: StatusUpdateActionResolver) extends TaskOpProcessor {
+  private[this] val log = LoggerFactory.getLogger(getClass)
+
+  import TaskOpProcessor._
+
+  override def process(op: Operation)(implicit ec: ExecutionContext): Future[Unit] = {
```
aquamatthias: Should we check the deadline of this operation?

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -42,14 +37,10 @@ class Migration @Inject() (
     */
   def migrations: List[MigrationAction] = List(
     StorageVersions(0, 5, 0) -> { () =>
```
aquamatthias: Guess jou can remove it for 0.5, if 0.7 is not supported.

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -42,14 +37,10 @@ class Migration @Inject() (
...
-      changeApps(app => app.copy(id = app.id.toString.toLowerCase.replaceAll("_", "-").toRootPath))
+      Future.failed(new IllegalStateException("migration from 0.5.x not supported anymore"))
     },
     StorageVersions(0, 7, 0) -> { () =>
-      for {
-        _ <- changeTasks(app => new InternalApp(app.appName.canonicalPath(), app.tasks, app.shutdown))
-        _ <- changeApps(app => app.copy(id = app.id.canonicalPath()))
-        _ <- putAppsIntoGroup()
-      } yield ()
+      Future.failed(new IllegalStateException("migration from 0.7.x not supported anymore"))
```
aquamatthias: Yay.

aquamatthias:
Looks good. 
I left a couple of comments.
Please also rebase.

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerConfig.scala
kolloch: nice suggestion

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
kolloch: Then the caller code will not compile anymore with our settings.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoader.scala
kolloch: What would you suggest instead? Just mock the Impl directly? I like this approach better -- not much more work -- but I do not care that much. If you want to, I can change this.

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerModule.scala
kolloch: I don't like the name either. But, unfortunately, I cannot think of something better. `TaskCreationListener` sounds to passive. The methods in `TaskCreator` do result in the task being "created/stored" in the `TaskTracker` or a task being removed from the TaskTracker.

Better suggestions?

Maybe we should keep the renaming to another PR, though.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoaderImpl.scala
kolloch: It is displayed once on startup. I would keep it at info. You do see if Marathon was stuck loading the tasks that way.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoaderImpl.scala
kolloch: See above.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
kolloch: If we have to update the state of the TaskTracker, we have to go through the TaskTracker for acknowledging the change (See Ack). If we did not change anything, then we can inform the sender directly.

I'll try to come up with good documentation for that.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
kolloch: This method is only called by the `TaskUpdateActor`. I chose to do the deadline checking there but the decision was quite arbitrary -- I just avoid a context switch there.

src/main/scala/mesosphere/marathon/state/Migration.scala
kolloch: I think that you are right.

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerModule.scala
aquamatthias: `taskCreationHandler`?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
aquamatthias: True - probably too much hassle in the moment.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoader.scala
aquamatthias: Not important: since this is a simple helper class, I wonder if we need a trait for that.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskLoaderImpl.scala
aquamatthias: OK

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
aquamatthias: Understant. Comment would help.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
aquamatthias: I would expect that check in process(Operation).  Perhaps add a line of documentation?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTrackerModule.scala
kolloch: ack

kolloch:
Thanks for the review @aquamatthias. I addressed your comments in four additional commits and also rebased the PR to the latest master.

aquamatthias:
@kolloch Failing test.

kolloch:
Sorry, forgot to commit/push my test fix.

aquamatthias:
Thanks @kolloch can you squash the commits?

PR #2877, opened 2015-12-23
Fixes #2872 - Make it possible to update the ipAddress

### LINKED ISSUE #2872: Update IPAddress of a running app is not possible
V2AppUpdate does not have the ipAddress field, so it will be ignored via an update.

gkleiman:


aquamatthias:
Looks good to me. Thanks!

PR #2887, opened 2016-01-01
Fixes #1604 - Add simple Docker integration tests

### LINKED ISSUE #1604: Create integration tests starting apps via docker
Currently, our integration tests do not test starting containers via docker. This is one of the major use cases for Marathon, therefore we should test that.

Ideally this includes:

* port mapping
* volume mounts
* parameters

gkleiman:
Fixes #1604 - Add simple Docker integration tests

These integration tests make Marathon start Docker containers, by making
the Mesos Agent use a downloaded Docker binary connecting to the Docker
daemon through the Docker unix socket.

They don't work when run on OS X.


Dockerfile.build-base
```diff
@@ -1,5 +1,5 @@
 #
-# This is the docker image used for ./bin/run-tests.sh and development tasks.
```
kolloch: Please delete this file if it is not used anymore by the ./bin/run-tests.sh script -- that was its only purpose.

kolloch: Or am I missing something?

bin/run-tests.sh
```diff
@@ -109,25 +111,43 @@ if [ "$CLEANUP_CONTAINERS_ON_EXIT" = "true" ]; then
...
+    -e "PROJECT_DIRS=$PROJECT_DIRS"
+    -e "SBT_DIR=$SBT_DIR"
+    -e "TARGET_DIRS=$TARGET_DIRS"
+    -e "MARATHON_MAX_TASKS_PER_OFFER=$MARATHON_MAX_TASKS_PER_OFFER"
+    -v "$SBT_DIR:/root/.sbt"
+    -v "$IVY2_DIR:/root/.ivy2"
+    -v "$TARGET_DIRS/main:/marathon/target"
+    -v "$TARGET_DIRS/project:/marathon/project/target"
+    -v "$TARGET_DIRS/mesos-simulation:/marathon/mesos-simulation/target"
+    -v "/etc/hosts:/etc/hosts"
```
kolloch: you should explain this :)

bin/run-tests.sh
```diff
@@ -109,25 +111,43 @@ if [ "$CLEANUP_CONTAINERS_ON_EXIT" = "true" ]; then
...
+    -i
+)
+
+if [ "$MOUNT_DOCKER_SOCKET" = "true" ]; then
+    docker_args+=(-v /var/run/docker.sock:/var/run/docker.sock)
+fi
+
+docker_args+=(
+    "marathon-buildbase:$BUILD_ID"
+    -c 'sbt -Dsbt.log.format=false test integration:test && sbt -Dsbt.log.format=false "project mesos-simulation" integration:test "test:runMain mesosphere.mesos.scale.DisplayAppScalingResults"'
```
kolloch: I know that it was like this before but it would be nice to break this very long line into multiple lines.

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -30,6 +30,52 @@ class AppDeployIntegrationTest
   //clean up state before running the test case
   before(cleanUp())
 
+  // FIXME (gkleiman): Docker tests only work when the integration tests are run on Linux using bin/run-tests.sh
+  if (sys.env.contains("BUILD_ID")) {
```
kolloch: this only tests if it was run via ./bin/run-tests.sh, right? If you did that on Mac OS, you would get a weird error. We do recommend running this script: https://mesosphere.github.io/marathon/docs/contributing.html#submitting-code-changes-to-marathon

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -30,6 +30,52 @@ class AppDeployIntegrationTest
...
+    test("deploy a simple Docker app") {
+      Given("a new Docker app")
+      val app = AppDefinition(
+        id = testBasePath / "dockerapp",
+        cmd = Some("true"),
```
kolloch: this will immediately stop the app and restart it -- is that really what we want here?

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -30,6 +30,52 @@ class AppDeployIntegrationTest
...
+        container = Some(
+          Container(
+            docker = Some(
+              mesosphere.marathon.state.Container.Docker(
+                image = "busybox"
+              )))
+        ),
+        cpus = 0.2,
+        mem = 256.0,
```
kolloch: lots of memory ;) I think we can specify less.

src/test/scala/mesosphere/marathon/integration/setup/SingleMarathonIntegrationTest.scala
```diff
@@ -160,11 +160,42 @@ trait SingleMarathonIntegrationTest
...
   }
 
-  def appProxy(appId: PathId, versionId: String, instances: Int, withHealth: Boolean = true, dependencies: Set[PathId] = Set.empty): AppDefinition = {
+  def appProxy(appId: PathId, versionId: String, instances: Int, withHealth: Boolean = true, dependencies: Set[PathId] = Set.empty, docker: Boolean = false): AppDefinition = {
     val mainInvocation = appProxyMainInvocation
-    val exec = Some(s"""echo APP PROXY $$MESOS_TASK_ID RUNNING; $mainInvocation $appId $versionId http://localhost:${config.httpPort}/health$appId/$versionId""")
     val health = if (withHealth) Set(HealthCheck(gracePeriod = 20.second, interval = 1.second, maxConsecutiveFailures = 10)) else Set.empty[HealthCheck]
-    AppDefinition(appId, exec, executor = "//cmd", instances = instances, cpus = 0.5, mem = 128.0, healthChecks = health, dependencies = dependencies)
+
+    if (docker) {
```
kolloch: since we branch for most of the method and you are actually not even using the "shared" health part, I would suggest two different methods. What do you think?

kolloch:
I have some comments. I am excited that we are finally close to getting docker test coverage!

src/test/scala/mesosphere/marathon/integration/setup/SingleMarathonIntegrationTest.scala
gkleiman: The `health` variable is used both on both branches, see line 191 and 197.

`mainInvocation` is used only in the else branch, so I'll move it there. I could create a new `dockerAppProxy` method, but I don't want to duplicate the health check creation logic.

It might however be necessary to create a new method once we add tests that exercise the port mappings.

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
gkleiman: Will change it for a looooong sleep, so that mesos has to kill the container when setting up the next test.

src/test/scala/mesosphere/marathon/integration/setup/SingleMarathonIntegrationTest.scala
kolloch: The `health` variable is used in both branches but your docker test with a health check doesn't use it. Correct?

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -30,6 +30,52 @@ class AppDeployIntegrationTest
...
+      result.code should be(201) // Created
+      extractDeploymentIds(result) should have size 1
+      waitForEvent("deployment_success")
+      waitForTasks(app.id, 1) // The app has really started
+    }
+
+    test("create a simple docker app using http health checks with HOST networking") {
+      Given("a new app")
+      val app = appProxy(testBasePath / "docker-http-app", "v1", instances = 1, withHealth = false, docker = true).
+        copy(healthChecks = Set(healthCheck))
```
kolloch: Why do you not use the healthCheck created in appProxy?

gkleiman:
@kolloch thanks for the review!

I addressed all your comments, reassigning the PR to you =).

bin/run-tests.sh
```diff
@@ -109,25 +123,45 @@ if [ "$CLEANUP_CONTAINERS_ON_EXIT" = "true" ]; then
     trap cleanup EXIT
 fi
 
+DOCKER_VERSION=`docker --version | sed -e 's/^.* version \([0-9.]*\).*$/\1/g'`
```
kolloch: Why do you set the docker version?

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -30,6 +30,51 @@ class AppDeployIntegrationTest
   //clean up state before running the test case
   before(cleanUp())
 
+  // FIXME (gkleiman): Docker tests don't work under Docker Machine yet. So they can be disabled through an env variable.
+  if (sys.env.getOrElse("RUN_DOCKER_INTEGRATION_TESTS", "true") == "true") {
```
kolloch: Move to an extra-suite

src/test/scala/mesosphere/marathon/integration/setup/SingleMarathonIntegrationTest.scala
```diff
@@ -160,11 +160,42 @@ trait SingleMarathonIntegrationTest
...
   }
 
-  def appProxy(appId: PathId, versionId: String, instances: Int, withHealth: Boolean = true, dependencies: Set[PathId] = Set.empty): AppDefinition = {
-    val mainInvocation = appProxyMainInvocation
-    val exec = Some(s"""echo APP PROXY $$MESOS_TASK_ID RUNNING; $mainInvocation $appId $versionId http://localhost:${config.httpPort}/health$appId/$versionId""")
+  def appProxy(appId: PathId, versionId: String, instances: Int, withHealth: Boolean = true, dependencies: Set[PathId] = Set.empty, docker: Boolean = false): AppDefinition = {
     val health = if (withHealth) Set(HealthCheck(gracePeriod = 20.second, interval = 1.second, maxConsecutiveFailures = 10)) else Set.empty[HealthCheck]
-    AppDefinition(appId, exec, executor = "//cmd", instances = instances, cpus = 0.5, mem = 128.0, healthChecks = health, dependencies = dependencies)
+
+    if (docker) {
```
kolloch: split it into two methods

gkleiman:
@kolloch another try =)

I added moved `extractDeploymentIds` to `MarathonCallbackTestSupport`. I am not 100% happy with the new place, but couldn't come up with a better one.

src/test/scala/mesosphere/marathon/integration/setup/MarathonCallbackTestSupport.scala
```diff
@@ -81,4 +85,15 @@ trait MarathonCallbackTestSupport extends ExternalMarathonIntegrationTest {
 
     receivedEventsForKinds.groupBy(_.eventType)
   }
+
+  def extractDeploymentIds(app: RestResult[AppDefinition]): Seq[String] = {
```
kolloch: That's not a good place because the topic of this class is actually dealing with callbacks. I also do not like mixins -- especially not for pure methods such as this.

kolloch: This call is related to the MarathonAPI. So you could for example put it into its own class in `mesosphere.marathon.integration.setup` or maybe as part of the MarathonFacade?

kolloch:
I think that you chose an unfortunate place for `extractDeploymentIds` other than that looks good.

kolloch:
:rocket: :+1: :rocket: :ship:

PR #2890, opened 2016-01-04
Fixed duplicate assertion in ConstraintsTest.

medzin:
In ConstraintsTest.scala file in one of the test cases was a duplicated assertion. It seems that the second assertion should check "rack-2" not "rack-1".


kolloch:
Thanks, @medzin!

PR #2891, opened 2016-01-04
Introduce trait for LeadershipModule

kolloch:
...to simplify test implementation.


src/test/scala/mesosphere/marathon/core/leadership/AlwaysElectedLeadershipModule.scala
```diff
@@ -12,17 +13,22 @@ import mesosphere.marathon.test.Mockito
   * This simplifies tests.
   */
 object AlwaysElectedLeadershipModule extends Mockito {
```
aquamatthias: Does it make sense to move that to the test package?

aquamatthias:
LGTM

PR #2913, opened 2016-01-07
Update Marathon UI version to 0.15-SNAPSHOT

philipnrmn:


project/build.scala
```diff
@@ -299,7 +299,7 @@ object Dependency {
     val PlayJson = "2.4.3"
     val JsonSchemaValidator = "2.2.6"
     val RxScala = "0.25.0"
-    val MarathonUI = "0.14.1"
+    val MarathonUI = "0.15-SNAPSHOT"
```
gkleiman: I think the version string should be `0.15.0-SNAPSHOT`.

philipnrmn: d'oh

philipnrmn:
Fixed it up. Thanks for the catch.

gkleiman:
¡Muchas gracias! :rocket:

PR #2921, opened 2016-01-07
Fixes #2919 - correct metrics for state reads/writes

### LINKED ISSUE #2919: StateMetrics.timed(Read/Write) incorrectly used for methods returning Futures
Therefore we only capture the time the calls need to return the `Future`s and no real data on ZooKeeper performance.

kolloch:


gkleiman:
:rocket:

PR #2931, opened 2016-01-08
Do not use user-supplied version timestamp for new applications

kolloch:
Reapply lost parts of f3a5ccfbd623de6f6ee4d3792c51419a8392c474


aquamatthias:
Thanks!

PR #2935, opened 2016-01-08
Do not enable tracing by default

kolloch:
...and even allow disabling it!

### DIFF
--- a/src/main/scala/mesosphere/marathon/DebugConf.scala
+++ b/src/main/scala/mesosphere/marathon/DebugConf.scala
@@ -22,7 +22,7 @@ trait DebugConf extends ScallopConf {
     noshort = true,
     prefix = "disable_")
 
-  lazy val deprecatedDebugTracing = opt[Boolean]("enable_tracing", hidden = true)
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/DebugConf.scala
```diff
@@ -22,7 +22,7 @@ trait DebugConf extends ScallopConf {
     noshort = true,
     prefix = "disable_")
 
-  lazy val deprecatedDebugTracing = opt[Boolean]("enable_tracing", hidden = true)
```
gkleiman: This change doesn't modify the default value, it just makes the flag show in the output of `--help`.

`opt[Boolean]` flags are actually [disabled by default](https://github.com/scallop/scallop/wiki/Option-types-and-definitions):

> Special case of plain options are the flag options. They take no arguments, and evaluate to true if option was present on the command line or to false if it was not

gkleiman:
@kolloch it's almost Friday 8 PM, so I might be missing something, but I think that the default was already false, and that this change would only make the deprecated flag show up in the output of `--help`.

kolloch:
I am closing this for now...

PR #2937, opened 2016-01-08
Report state metrics to the same classes as before

kolloch:
(the intermediate version did only port them under StateMetricsTemplate name)


gkleiman:
Thanks!

PR #2959, opened 2016-01-12
Fixes #2927 - cap maximum number of parallel status updates

### LINKED ISSUE #2927: Cap parallel task status update processing
With PR https://github.com/mesosphere/marathon/pull/2863 status update processing has become "less blocking". Therefore, if there are many status updates as happens for reconciliation, a lot of parallel requests are sent to ZooKeeper which leads to problems.

Experimentation has shown that if we throttle the status updates as they come into Marathon, Marathon behaves much better under load.

This throttling is also an optimal place for some metrics for monitoring:

* how many status updates are currently processed?
* how many status updates are still queued for processing?

kolloch:
that are processed. Also unify on one "serialization solution".


kolloch:
Rebased.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -224,4 +225,13 @@ trait MarathonConf
     descr = "The timeout for preparing the Marathon instance when elected as leader.",
     default = Some(3 * 60 * 1000L) //3 minutes
   )
+
+  lazy val maxQueuedRootGroupUpdates = opt[Int]("max_queued_root_group_updates",
```
aquamatthias: Can we name all our internal parameters `internal` - so it is easy to reason about (the val and the cmd line parameter)?

kolloch: Ack for renaming the vars.

we haven't renamed the new options for the TaskTrackerActor "internal*" on the command line. They are marked as "hidden" though and have "INTERNAL TUNING PARAMETER" as description.

The command line parameters get quite long... Anyway, either way we go we should treat all new internal options.

aquamatthias: Kind of like the idea to have the var named internalXX, but the parameter not.
If we think we should unhide this parameter, we can make this change without breaking any existing external usage.

src/main/scala/mesosphere/marathon/core/task/update/TaskStatusUpdateConfig.scala
```diff
@@ -0,0 +1,22 @@
+package mesosphere.marathon.core.task.update
+
+import org.rogach.scallop.ScallopConf
+
+trait TaskStatusUpdateConfig extends ScallopConf {
+  //scalastyle:off magic.number
+
+  lazy val maxParallelStatusUpdates = opt[Int]("max_parallel_status_updates",
```
aquamatthias: Please use `internalMaxParallelStatusUpdates`

src/main/scala/mesosphere/marathon/core/task/update/TaskStatusUpdateConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+    descr = "INTERNAL TUNING PARAMETER: The maximum number of status updates that get processed in parallel.",
+    noshort = true,
+    hidden = true,
+    default = Some(20)
+  )
+
+  lazy val maxQueuedStatusUpdates = opt[Int]("max_queued_status_updates",
```
aquamatthias: please use `internalMaxQueuedStatusUpdates`

src/main/scala/mesosphere/marathon/core/task/update/impl/ThrottlingTaskStatusUpdateProcessor.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object ThrottlingTaskStatusUpdateProcessor {
+  /**
+    * A tag used for dependency injection to disambiguate the dependencies of this processor from
+    * other instances with the same type.
+    */
+  final val dependencyTag = "ThrottlingTaskStatusUpdateProcessor"
+}
+
+private[core] class ThrottlingTaskStatusUpdateProcessor @Inject() (
```
aquamatthias: Can we do the wiring outside of impl? 
And can we move the guice wiring to the CoreGuice level?

kolloch: Please not. What you suggest means that the `@Named` annotations have a dependency on the wiring stuff. That's kind of the opposite that you want to achieve by DI.

Yes, I would like to wire without using `@Named` altogether but with our current DI situation I do not know how.

kolloch: (I assumed you mean declaring this constant by "wiring" since you attached the comment to it. The wiring itself, of course, is not done here.)

src/test/scala/mesosphere/util/CapConcurrentExecutionsTest.scala
```diff
@@ -0,0 +1,119 @@
...
+    with ScalaFutures {
+  def capMetrics: CapConcurrentExecutionsMetrics = new CapConcurrentExecutionsMetrics(
+    new Metrics(new MetricRegistry),
+    classOf[CapConcurrentExecutionsTest]
+  )
+
+  test("submit successful futures after each other") {
+    val serialize = CapConcurrentExecutions(capMetrics, system, "serialize1", maxParallel = 1, maxQueued = 10)
+    try {
+      val result1 = serialize(Future.successful(1)).futureValue
```
aquamatthias: To be honest - I don't see the point of those tests: you basically block and wait for the future to succeed. Proposal:
```scala
      var count = 0
      serialize(Future{ Thread.sleep(50); count += 1; count should be(1) })
      serialize(Future{ count += 1; count should be(2) })
      serialize(Future{ count += 1; count should be(3) })
```

src/test/scala/mesosphere/util/CapConcurrentExecutionsTest.scala
```diff
@@ -0,0 +1,119 @@
...
+      result1 should be(1)
+      val result2 = Future.successful(2).futureValue
```
aquamatthias: serialize missing - becomes Future(2).get==2

kolloch: Thanks.

src/test/scala/mesosphere/util/CapConcurrentExecutionsTest.scala
kolloch: This should a basic baseline test -- just checking that the serialization does not interfere with the future execution. Something like that (but without the sleep) is used below.

src/main/scala/mesosphere/marathon/core/task/update/TaskStatusUpdateConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.rogach.scallop.ScallopConf
+
+trait TaskStatusUpdateConfig extends ScallopConf {
+  //scalastyle:off magic.number
+
+  lazy val internalMaxParallelStatusUpdates = opt[Int]("max_parallel_status_updates",
+    descr = "INTERNAL TUNING PARAMETER: The maximum number of status updates that get processed in parallel.",
+    noshort = true,
+    hidden = true,
+    default = Some(20)
```
aquamatthias: How can we find a "good" number here?

src/main/scala/mesosphere/marathon/core/task/update/TaskStatusUpdateConfig.scala
```diff
@@ -0,0 +1,22 @@
...
+  )
+
+  lazy val internalMaxQueuedStatusUpdates = opt[Int]("max_queued_status_updates",
+    descr = "INTERNAL TUNING PARAMETER: The maximum number of status updates that we queue for processing." +
+      " Mesos will resent status updates which we do not acknowledge.",
+    noshort = true,
+    hidden = true,
+    default = Some(10000)
```
aquamatthias: Feels safe. :)

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
```diff
@@ -0,0 +1,157 @@
...
+
+class CapConcurrentExecutions private (
+    metrics: CapConcurrentExecutionsMetrics,
+    actorRefFactory: ActorRefFactory,
+    actorName: String,
+    maxParallel: Int,
+    maxQueued: Int) {
+  import CapConcurrentExecutions.log
+
+  private[this] val serializeExecutionActorProps =
```
aquamatthias: Do you need this props other than to create the acorRef one line below?

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
```diff
@@ -0,0 +1,157 @@
...
+    RestrictParallelExecutionsActor.props(metrics, maxParallel = maxParallel, maxQueued = maxQueued)
+  private[this] val serializeExecutionActorRef = actorRefFactory.actorOf(serializeExecutionActorProps, actorName)
+
+  def apply[T](block: => Future[T]): Future[T] = {
+    PromiseActor.askWithoutTimeout(
+      actorRefFactory, serializeExecutionActorRef, RestrictParallelExecutionsActor.Execute(() => block))
+  }
+
+  @PreDestroy
```
aquamatthias: What is the purpose of this?

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+  * Accepts execute instructions containing functions returning `scala.concurrent.Future`s.
+  * It only allows `maxParallel` parallel executions and queues the other operations.
+  * It will not queue more than `maxQueued` execute instructions.
+  */
+private[util] class RestrictParallelExecutionsActor(
+    metrics: CapConcurrentExecutionsMetrics, maxParallel: Int, maxQueued: Int) extends Actor {
+
+  import RestrictParallelExecutionsActor.{ log, Execute, Queued }
```
aquamatthias: My IDE says log is unused.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
```diff
@@ -0,0 +1,157 @@
...
+    actorName: String,
+    maxParallel: Int,
+    maxQueued: Int) {
+  import CapConcurrentExecutions.log
+
+  private[this] val serializeExecutionActorProps =
+    RestrictParallelExecutionsActor.props(metrics, maxParallel = maxParallel, maxQueued = maxQueued)
+  private[this] val serializeExecutionActorRef = actorRefFactory.actorOf(serializeExecutionActorProps, actorName)
+
+  def apply[T](block: => Future[T]): Future[T] = {
```
aquamatthias: Shouldn't this also have an EC as implicit parameter?
On which EC is that block executed?

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
```diff
@@ -0,0 +1,157 @@
...
+  private[this] var active: Int = 0
+  private[this] var queue: Queue[Queued] = Queue.empty
+
+  override def preStart(): Unit = {
+    super.preStart()
+
+    metrics.reset()
+
+    for (execute <- queue) {
+      execute.sender ! Status.Failure(new IllegalStateException(s"$self actor stopped"))
```
aquamatthias: Should the queue be cleared after sending failed to all clients?

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
```diff
@@ -0,0 +1,157 @@
...
+
+        import context.dispatcher
+        import akka.pattern.pipe
+        val future: Future[Any] =
+          try metrics.processingTimer.timeFuture(next.func())
+          catch { case NonFatal(e) => Future.failed(e) }
+        future.pipeTo(next.sender)
+
+        val myself = self
+        future.onComplete(_ => myself ! Finished)(CallerThreadExecutionContext.callerThreadExecutionContext)
```
aquamatthias: Curious: why do you use the callerThreadEC here?

aquamatthias:
Added a couple of questions regarding EC. 
Generally I like the idea of this PR  :+1:

src/main/scala/mesosphere/marathon/core/task/update/TaskStatusUpdateConfig.scala
kolloch: Testing. I used something bigger before and it was not a success. I think `20` is erring on the side of non-optimal performance but no havoc.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
kolloch: Submitting the future for execution.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
kolloch: that can be entirely done in the block that is given as an argument. No need for a separate argument.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
kolloch: I guess. But since we really should do this in postStop, it doesn't matter that much ;)

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
kolloch: To avoid the overhead of yet another context switch. Sending a message to an actor not be (significantly) more expensive than submitting a task to an execution context.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
kolloch: I guess you want to encourage me to fold this into the next line?

kolloch:
Thanks. I made some review-inspired changes in an extra commit.

kolloch:
I also exchanged the askActor thing with a promise in a second commit.

src/main/scala/mesosphere/marathon/core/task/update/impl/ThrottlingTaskStatusUpdateProcessor.scala
aquamatthias: My hope was, that we do not use Inject and stuff inside core, but only in CoreGuiceModule.

src/test/scala/mesosphere/util/CapConcurrentExecutionsTest.scala
aquamatthias: Ok. I try again: 
```
val result1 = serialize(Future.successful(1))
val result2 = serialize(Future.successful(2))
val result3 = serialize(Future.successful(3))
result1.futureValue  should be(1)
result2.futureValue  should be(1)
result3.futureValue  should be(3)
```
It just looks wrong to await the result on each line...

src/test/scala/mesosphere/util/CapConcurrentExecutionsTest.scala
```diff
@@ -0,0 +1,163 @@
...
+      result2 should be(2)
+      val result3 = serialize(Future.successful(3)).futureValue
+      result3 should be(3)
+    }
+    finally {
+      serialize.close()
+    }
+  }
+
+  test("concurrent executions are serialized if maxParallel has been reached") {
```
aquamatthias: Really cool test! :+1: 
And also interesting to assert on the metrics :)

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
aquamatthias: Yes. The Future is created with one and we don't do (map etc) anything with it.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
aquamatthias: Got it. Thanks.

src/main/scala/mesosphere/util/CapConcurrentExecutions.scala
aquamatthias: Never used this. Can you add a comment for further readers?

aquamatthias:
Looks good! Proposed some minor changes. Please address and squash - I will merge it.

src/test/scala/mesosphere/util/CapConcurrentExecutionsTest.scala
kolloch: Why does it look wrong? Agreed, it is a basic test but why is it wrong? Sorry, I don't want to deflect criticism, I just don't get it.

aquamatthias:
Please squash.

kolloch:
rebased/squashed.

aquamatthias:
Thanks!

PR #2963, opened 2016-01-12
Fixes #2957 - improve usage of io threadpool

### LINKED ISSUE #2957: Unbounded ThreadPool is used for too many operations
`ThreadPoolContext` provides an unbounded cached ThreadPool named `context`. The comment specifies that this should only be used for blocking IO operations but in reality this context is used for many operations.

This can lead to `OutOfMemoryException`s and other resource exhaustion problems.

kolloch:
* Use it only for bounded operations
* Make it use a fixed number of threads


kolloch:
rebased

gkleiman:
Thanks!

:rocket:

PR #2965, opened 2016-01-12
Update Tomcat examples

guenter:
* update the Tomcat example to the current version of Tomcat
* add an example of Tomcat in Docker


aquamatthias:
Thanks!

PR #2966, opened 2016-01-12
Remove misleading logging on GET /v2/info

gkleiman:


kolloch:
Looks good!

kolloch:
Maybe also remove the FIXME?

gkleiman:
The FIXME is still true. That method is only called by `InfoResource`, so it might be better to pass a `FrameworkIdUtil` instance to it and remove the method and `frameworkIdUtil` from `MarathonSchedulerService`.

kolloch:
:rocket:

PR #2975, opened 2016-01-13
Adding 'message' property and changing 'errors' to 'details'

alexanderweber:
Regarding #2804, there is a test been commented but not uncommented during the PR. Sorry for that, the test is being committed with this PR.


aquamatthias:
LGTM! Wait for the tests to get green.

alexanderweber:
@aquamatthias "message" -> "Object is not valid" is added to JSON reply now. "errors" is renamed to "details". Please get back to me if there is something I did not think of here.

alexanderweber:
@aquamatthias tests are green now, thanks for looking into it

PR #2977, opened 2016-01-13
Fix error when installing sbt

kolloch:
Basically, codify the instructions from

http://www.scala-sbt.org/0.13/tutorial/Manual-Installation.html

Fixes this error:

[10:46:32][Step 3/3] W: GPG error: http://dl.bintray.com  Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 99E82A75642AC823
[10:46:32][Step 3/3] W: Failed to fetch http://httpredir.debian.org/debian/dists/jessie-backports/main/binary-amd64/Packages  Hash Sum mismatch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,12 +12,15 @@ FROM java:8-jdk
 COPY . /marathon
 WORKDIR /marathon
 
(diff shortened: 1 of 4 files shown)


aquamatthias:
Thanks.

PR #2978, opened 2016-01-13
Fixes #2426 - get rid of v2 model objects

### LINKED ISSUE #2426: Get rid of V2 model objects

alexanderweber:
Using play-json, we are no longer in need for V2AppDefinition or V2Group classes anymore. This pull request should get rid of those classes.


src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
```diff
@@ -61,13 +61,13 @@ class GroupsResource @Inject() (
...
-        case ListApps(gid)              => groupResponse(gid.toRootPath, _.transitiveApps.map(V2AppDefinition(_)))
-        case ListRootApps()             => groupResponse(PathId.empty, _.transitiveApps.map(V2AppDefinition(_)))
+        case ListApps(gid)              => groupResponse(gid.toRootPath, _.transitiveApps)
+        case ListRootApps()             => groupResponse(PathId.empty, _.transitiveApps)
         case ListVersionsRE(gid)        => ok(jsonString(result(groupManager.versions(gid.toRootPath))))
         case ListRootVersionRE()        => ok(jsonString(result(groupManager.versions(PathId.empty))))
-        case GetVersionRE(gid, version) => groupResponse(gid.toRootPath, V2Group(_), version = Some(Timestamp(version)))
-        case GetRootVersionRE(version)  => groupResponse(PathId.empty, V2Group(_), version = Some(Timestamp(version)))
-        case _                          => groupResponse(id.toRootPath, V2Group(_))
+        case GetVersionRE(gid, version) => groupResponse(gid.toRootPath, { g => g }, version = Some(Timestamp(version)))
```
alexanderweber: @aquamatthias Could you think of a better way defining the function, or is that one ok?

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala
```diff
@@ -17,10 +18,10 @@ class V2AppDefinitionFormatsTest
   import Formats.PathIdFormat
 
   object Fixture {
-    val a1 = V2AppDefinition(
+    val a1 = AppDefinition(
       id = "app1".toPath,
       cmd = Some("sleep 10"),
-      version = Timestamp(1)
+      versionInfo = AppDefinition.VersionInfo.OnlyVersion(Timestamp(1))
```
alexanderweber: There is no version in AppDefinition compared to V2AppDefinition. Is that the correct equivalent?

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
alexanderweber: @aquamatthias Do you think of a better way of defining a function which only forwards the group object?

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala
alexanderweber: @aquamatthias AppDefintion does have VersionInfo instead of the version Property which was used in V2AppDefinition. Is this the correct equivalent?

kolloch:
@alexanderweber If you use "Fixes #1234 - " as title prefix, waffle and github can automatically associate the PR with the issue that they fix. Please also prefix your commits like this.

See https://mesosphere.github.io/marathon/docs/contributing.html

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
aquamatthias: There is the `identity` function in Predef. But you probably can completely remove the function in groupResponse and directly use the group?

aquamatthias:
@alexanderweber Great. Please rename your commit as suggested by @kolloch and I will merge.

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
alexanderweber: Thanks! I did change it to `Predef.identity`

alexanderweber:
@kolloch Thanks Peter! I did change it.

aquamatthias:
Thanks!

PR #2979, opened 2016-01-13
Improve TaskReplaceActorTest "Wait until the tasks are killed"

kolloch:
* Speed up from more than 5 seconds down to a couple of milliseconds
* Actually test more

### DIFF
--- a/src/test/scala/mesosphere/marathon/upgrade/TaskReplaceActorTest.scala
+++ b/src/test/scala/mesosphere/marathon/upgrade/TaskReplaceActorTest.scala
@@ -14,6 +14,7 @@ import mesosphere.marathon.upgrade.TaskReplaceActor.RetryKills
 import org.apache.mesos.Protos.{ Status, TaskID }
 import org.apache.mesos.SchedulerDriver
 import org.mockito.Matchers.any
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/marathon/upgrade/TaskReplaceActorTest.scala
```diff
@@ -566,12 +567,13 @@ class TaskReplaceActorTest
     for (i <- 0 until app.instances)
       ref ! MesosStatusUpdateEvent("", s"task_$i", "TASK_RUNNING", "", app.id, "", Nil, Nil, app.version.toString)
 
-    intercept[Exception] {
```
gkleiman: The goal of this test was to make sure that the Actor doesn't complete the future until all pending kills are completed.

So if the actor sends the kills and then completes right away, without waiting for the TASK_KILLED updates, then the test should fail. This was the behaviour before the test was introduced.

I might be wrong, but I think that with this change, the test would pass even if the actor doesn't wait for the kills.

kolloch:
I hopefully addressed your concern.

gkleiman:
:shipit:

PR #2994, opened 2016-01-14
Fixes #2989 - report total task count metrics

### LINKED ISSUE #2989: Report task count metrics
Task counts are one of the most important characteristics of Marathon usage.

Since we now have a synchronized place for task creation/update/deletion, we can now provide reliable task stats reasonably cheap. I'd suggest counts analogous to our API:

* staged
* running

Not part of this issue since these require querying the health check manager:

* healthy
* unhealthy

kolloch:
running/staged

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/appinfo/TaskCounts.scala
+++ b/src/main/scala/mesosphere/marathon/core/appinfo/TaskCounts.scala
@@ -13,16 +13,34 @@ import org.apache.mesos.{ Protos => mesos }
   * @param tasksUnhealthy snapshot of the number of unhealthy tasks (does not include tasks without health info)
   */
 case class TaskCounts(
-  tasksStaged: Int,
-  tasksRunning: Int,
-  tasksHealthy: Int,
-  tasksUnhealthy: Int)
+    tasksStaged: Int,
+    tasksRunning: Int,
+    tasksHealthy: Int,
+    tasksUnhealthy: Int) {
+  def +(counts: TaskCounts): TaskCounts = {
+    copy(
+      tasksRunning = tasksRunning + counts.tasksRunning,
+      tasksStaged = tasksStaged + counts.tasksStaged,
+      tasksHealthy = tasksHealthy + counts.tasksHealthy,
+      tasksUnhealthy = tasksUnhealthy + counts.tasksUnhealthy
+    )
+  }
+
+  def -(counts: TaskCounts): TaskCounts = {
+    copy(
+      tasksRunning = tasksRunning - counts.tasksRunning,
+      tasksStaged = tasksStaged - counts.tasksStaged,
+      tasksHealthy = tasksHealthy - counts.tasksHealthy,
+      tasksUnhealthy = tasksUnhealthy - counts.tasksUnhealthy
+    )
+  }
+}
 
 object TaskCounts {
   def zero: TaskCounts = TaskCounts(tasksStaged = 0, tasksRunning = 0, tasksHealthy = 0, tasksUnhealthy = 0)
 
-  def apply(appTasks: Iterable[MarathonTask], statuses: Map[String, Seq[Health]]): TaskCounts = {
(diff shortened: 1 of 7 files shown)


src/main/scala/mesosphere/marathon/core/appinfo/TaskCounts.scala
```diff
@@ -13,16 +13,34 @@ import org.apache.mesos.{ Protos => mesos }
...
+      tasksUnhealthy = tasksUnhealthy + counts.tasksUnhealthy
+    )
+  }
+
+  def -(counts: TaskCounts): TaskCounts = {
+    copy(
+      tasksRunning = tasksRunning - counts.tasksRunning,
+      tasksStaged = tasksStaged - counts.tasksStaged,
+      tasksHealthy = tasksHealthy - counts.tasksHealthy,
+      tasksUnhealthy = tasksUnhealthy - counts.tasksUnhealthy
```
meichstedt: For the sake of the subtraction, the implementation is fine – semantically, something like `TaskCounts.zero - TaskCounts(1, 1, 1, 1)` should at least result in `TaskCounts.zero`, not in `TaskCounts(-1, -1, -1, -1)`

src/main/scala/mesosphere/marathon/core/task/tracker/impl/AppData.scala
```diff
@@ -32,9 +33,8 @@ private[impl] case class AppDataMap private (appTasks: Map[PathId, AppData]) {
     task <- app.taskMap.get(taskId)
   } yield task
 
-  def toTaskTrackerAppMap: Map[PathId, TaskTracker.App] = {
-    appTasks.mapValues(_.toTaskTrackerApp)
-  }
+  def toTaskTrackerAppMap: Map[PathId, TaskTracker.App] = appTasks.mapValues(_.toTaskTrackerApp)
+  def tasks: Iterable[MarathonTask] = appTasks.view.flatMap(_._2.tasks)
```
meichstedt: nit-picking: in other locations, we have names like `allIds` (which I'm not necessarily fond of). Here, we have `getTasks(appId)` vs `tasks`. What do you think about aligning and calling it `allTasks` instead? Otherwise, I'd tend to use `tasksFor(appId)` and `tasks` to make it explicit.

src/main/scala/mesosphere/marathon/core/appinfo/TaskCounts.scala
kolloch: Well, it wouldn't work like I use it. And I don't agree. Why should it? I actually think if you really want to restrict the counts to positive numbers, your example should result in an exception. But I think the chosen semantic is completely valid.

Semantically, you get a TaskCount diff -- I could introduce another type for that but I think that this is not necessary. I can't imagine someone using this method without thinking about the semantics. So the error potential is small.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/AppData.scala
kolloch: I like `tasksForApp(appId)` and `allTasks`

kolloch: If I change that now, I will get into major conflicts. I created another issue for this:

https://github.com/mesosphere/marathon/issues/3003

src/main/scala/mesosphere/marathon/core/appinfo/TaskCounts.scala
meichstedt: I see. I added the comment before I got to the usage; from that perspective it's ok. The only strange thing is that TaskCounts can actually be negative – yet I don't think an Exception would be any help here so just leave it like that.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/AppData.scala
meichstedt: :+1:

meichstedt:
LGTM!

PR #2996, opened 2016-01-14
Allow obtaining client IP in authe plugin

janisz:
Add X-Forwarded-For header for proxy
Add client IP in plugin request


aquamatthias:
Thanks for this PR!

PR #3002, opened 2016-01-15
Mark blocking calls in MarathonSchedulerService

kolloch:
with scala.concurrent.blocking


aquamatthias:
That will block for a long time :) Good Catch!

PR #3006, opened 2016-01-15
Make sure to not match on every zk option.

cmluciano:
The script allows for the passing of zk run commands as files under the
proper conf_dir. If an option like zk_timeout is changed with this
method, the wildcard will attempt to match that as the option to pass to
the --zk command. The extra space after the --zk line in this script
should only match if someone is actually passing that exact command.

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -66,7 +66,7 @@ function load_options_and_log {
   # Default zk and master option
   if [[ -s /etc/mesos/zk ]]
   then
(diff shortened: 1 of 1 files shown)


aquamatthias:
@cmluciano Thanks for this PR. It looks good from my point of view.
@lloesche can you verify please?

lloesche:
LGTM

cmluciano:
Thanks! Is there any chance of getting this backported to 0.11.1?

PR #3023, opened 2016-01-19
Move the haproxy-marathon-bridge into examples directory and add a deprecation warning

aquamatthias:


lloesche:
LGMT

kolloch:
Thanks!

kolloch:
@aquamatthias we should also mention the deprecation in our service discovery/load balancing documentation -- and update the paths there. (Sorry, thought about this after merging).

PR #3027, opened 2016-01-19
Fix constraints definition schema

janisz:


kolloch:
Nice, @janisz. Thank you! :rocket:

PR #3037, opened 2016-01-20
Update build.scala for 0.16.0-snapshot

Poltergeist:


gkleiman:
Thanks @Poltergeist!

We want master to always use the latest marathon-ui snapshot: `v0.16.0-SNAPSHOT`.

This commit should be merged into the release branch for 0.15.0 (`releases-0.15`) as soon as it is created.

PR #3038, opened 2016-01-20
Fixes #2868 - Don't assign duplicated service ports

### LINKED ISSUE #2868: Marathon sometimes tries (and fails) to assign duplicated service ports.
Steps to reproduce:

1. Start Marathon with the default service ports range: `[10000, 20000]` and no apps.
2. POST /v2/apps with:

```json
{
  "id": "/foo",
  "cmd": "sleep 1000",
  "ports": [0, 10000],
  "requirePorts": false
}
```

Marathon will then try to assign the following service ports: `[10000, 10000]`, fail and return a 422 with the following body:

```json
{
  "message": "Bean is not valid",
  "errors": [
    {
      "attribute": "ports",
      "error": "Elements must be unique"
    }
  ]
}
```

It should instead be cleverer and assign `[10001, 10000]`.

cc/ @vishnu2kmohan

gkleiman:


aquamatthias:
Thanks

PR #3072, opened 2016-01-22
Fixes #3047 by introducing a exponential backoff for unresponsive hosts.

### LINKED ISSUE #3047: Event callbacks hold on to unresponsive hosts
When using event callbacks, an unresponsive host will be tried six times for every event. We'd like the option of a specified number of failed endpoint connections or a specified period of endpoint unavailability to trigger the removal of that endpoint.

aquamatthias:
Also a couple of metrics are introduced to have better insights.

### DIFF
--- a/src/main/java/mesosphere/marathon/metrics/Metrics.scala
+++ b/src/main/java/mesosphere/marathon/metrics/Metrics.scala
@@ -1,11 +1,11 @@
 package mesosphere.marathon.metrics
 
 import java.util.concurrent.TimeUnit
-import java.util.concurrent.atomic.{AtomicInteger, AtomicLong}
+import java.util.concurrent.atomic.AtomicInteger
 
 import com.codahale.metrics.{Gauge, MetricRegistry}
 import com.google.inject.Inject
-import mesosphere.marathon.metrics.Metrics.{Histogram, Meter, Timer}
+import mesosphere.marathon.metrics.Metrics.{ Histogram, Meter, Timer, Counter }
 import org.aopalliance.intercept.MethodInvocation
 
 import scala.collection.concurrent.TrieMap
@@ -26,6 +26,10 @@ class Metrics @Inject() (val registry: MetricRegistry) {
     finally {
       timer.update(System.nanoTime() - startTime, TimeUnit.NANOSECONDS)
     }
+  }
+
+  def counter(name: String): Counter = {
+    new Counter(registry.counter(name))
   }
 
   def timer(name: String): Timer = {
@@ -64,6 +68,11 @@ class Metrics @Inject() (val registry: MetricRegistry) {
 }
 
 object Metrics {
+  class Counter(counter: com.codahale.metrics.Counter) {
+    def inc(): Unit = counter.inc()
+    def dec(): Unit = counter.dec()
+  }
+
   class Timer(timer: com.codahale.metrics.Timer) {
     def apply[T](block: => T): T = {
       val startTime = System.nanoTime()
@@ -74,6 +83,8 @@ object Metrics {
         timer.update(System.nanoTime() - startTime, TimeUnit.NANOSECONDS)
       }
     }
+
+    def updateMillis(time: Long): Unit = timer.update(time, TimeUnit.MILLISECONDS)
   }
 
   class Histogram(histogram: com.codahale.metrics.Histogram) {
@@ -87,13 +98,9 @@ object Metrics {
   }
 
   class Meter(meter: com.codahale.metrics.Meter) {
-    def mark(): Unit = {
-      meter.mark()
-    }
-
-    def mark(n: Long): Unit = {
-      meter.mark(n)
-    }
+    def mark(): Unit = meter.mark()
+    def mark(n: Long): Unit = meter.mark(n)
+    def mark(n: Int): Unit = meter.mark(n.toLong)
   }
 
   class AtomicIntGauge extends Gauge[Int] {
--- a/src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
+++ b/src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
@@ -1,54 +1,114 @@
 package mesosphere.marathon.event.http
 
+import java.lang.System.currentTimeMillis
+import javax.inject.Inject
+
 import akka.actor._
 import akka.pattern.ask
+import mesosphere.marathon.api.v2.json.Formats._
 import mesosphere.marathon.event._
+import mesosphere.marathon.event.http.HttpEventActor._
 import mesosphere.marathon.event.http.SubscribersKeeperActor.GetSubscribers
-import play.api.libs.json.{ JsValue, Json }
+import mesosphere.marathon.metrics.{ MetricPrefixes, Metrics }
 import spray.client.pipelining.{ sendReceive, _ }
 import spray.http.{ HttpRequest, HttpResponse }
 import spray.httpx.PlayJsonSupport
-import mesosphere.marathon.api.v2.json.Formats._
 
 import scala.concurrent.Future
+import scala.concurrent.duration._
 import scala.util.{ Failure, Success }
 
-class HttpEventActor(val subscribersKeeper: ActorRef) extends Actor with ActorLogging with PlayJsonSupport {
+/**
+  * This actor subscribes to the event bus and distributes every event to all http callback listener.
+  * The list of active subscriptions is handled in the subscribersKeeper.
+  * If a callback handler can not be reached or is slow, an exponential backoff is applied.
+  */
+object HttpEventActor {
+  case class NotificationFailed(url: String)
+  case class NotificationSuccess(url: String)
+
+  case class EventNotificationLimit(failedCount: Long, backoffUntil: Option[Deadline]) {
+    def nextFailed: EventNotificationLimit = {
+      val next = failedCount + 1
+      EventNotificationLimit(next, Some(math.pow(2, next.toDouble).seconds.fromNow))
+    }
+    def notLimited: Boolean = backoffUntil.fold(true)(_.isOverdue())
+    def limited: Boolean = !notLimited
+  }
+  val NoLimit = EventNotificationLimit(0, None)
+
+  class HttpEventActorMetrics @Inject() (metrics: Metrics) {
+    private val pre = MetricPrefixes.SERVICE
+    private val clazz = classOf[HttpEventActor]
+    // the number of requests that are open without response
+    val outstandingCallbacks = metrics.counter(metrics.name(pre, clazz, "outstanding-callbacks"))
+    // the number of events that are broadcast
+    val eventMeter = metrics.meter(metrics.name(pre, clazz, "events"))
+    // the number of events that are not send to callback listeners due to backoff
+    val skippedCallbacks = metrics.meter(metrics.name(pre, clazz, "skipped-callbacks"))
+    // the number of callbacks that have failed during delivery
+    val failedCallbacks = metrics.meter(metrics.name(pre, clazz, "failed-callbacks"))
+    // the response time of the callback listeners
+    val callbackResponseTime = metrics.timer(metrics.name(pre, clazz, "callback-response-time"))
+  }
+}
+
+class HttpEventActor(conf: HttpEventConfiguration, val subscribersKeeper: ActorRef, metrics: HttpEventActorMetrics)
+    extends Actor with ActorLogging with PlayJsonSupport {
 
   implicit val ec = HttpEventModule.executionContext
   implicit val timeout = HttpEventModule.timeout
-
-  val pipeline: HttpRequest => Future[HttpResponse] = (
-    addHeader("Accept", "application/json")
-    ~> sendReceive)
+  val pipeline: HttpRequest => Future[HttpResponse] = addHeader("Accept", "application/json") ~> sendReceive
+  var limiter = Map.empty[String, EventNotificationLimit].withDefaultValue(NoLimit)
 
   def receive: Receive = {
-    case event: MarathonEvent =>
-      broadcast(event)
-    case _ =>
-      log.warning("Message not understood!")
+    case event: MarathonEvent     => broadcast(event)
+    case NotificationSuccess(url) => limiter += url -> NoLimit
+    case NotificationFailed(url)  => limiter += url -> limiter(url).nextFailed
+    case _                        => log.warning("Message not understood!")
   }
 
   def broadcast(event: MarathonEvent): Unit = {
+    metrics.eventMeter.mark()
     log.info("POSTing to all endpoints.")
-    (subscribersKeeper ? GetSubscribers).mapTo[EventSubscribers].foreach {
-      _.urls.foreach { post(_, event) }
+    val me = self
+    (subscribersKeeper ? GetSubscribers).mapTo[EventSubscribers].foreach { subscribers =>
+      val (active, limited) = subscribers.urls.partition(limiter(_).notLimited)
+      if (limited.nonEmpty) {
+        log.info(s"""Will not send event ${event.eventType} to unresponsive hosts: ${limited.mkString(" ")}""")
+      }
+      //remove all unsubscribed callback listener
+      limiter = limiter.filterKeys(subscribers.urls).iterator.toMap.withDefaultValue(NoLimit)
+      metrics.skippedCallbacks.mark(limited.size)
+      active.foreach(post(_, event, me))
     }
   }
 
-  def post(urlString: String, event: MarathonEvent): Unit = {
-    log.info("Sending POST to:" + urlString)
+  def post(url: String, event: MarathonEvent, eventActor: ActorRef): Unit = {
+    log.info("Sending POST to:" + url)
 
-    val request = Post(urlString, eventToJson(event))
+    metrics.outstandingCallbacks.inc()
+    val request = Post(url, eventToJson(event))
     val response = pipeline(request)
+    val start = currentTimeMillis()
 
     response.onComplete {
+      case _ =>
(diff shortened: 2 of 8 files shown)


aquamatthias:
I guess we changed settings in a way, that travis can not build any longer...
On my machine the tests were green.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
 import spray.httpx.PlayJsonSupport
-import mesosphere.marathon.api.v2.json.Formats._
 
 import scala.concurrent.Future
-import scala.util.{ Failure, Success }
+import scala.concurrent.duration._
+import scala.util.Success
+import System.currentTimeMillis
+
+object HttpEventActor {
```
kolloch: Not the fault of this PR but some documentation about what this actor does and what the implementation goals are would be nice.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
+  case class NotificationFailed(url: String)
+  case class NotificationSuccess(url: String)
+  case class NotificationSlow(url: String)
+
+  case class EventNotificationLimit(failed: Long, deadLine: Option[Deadline]) {
```
kolloch: what does `failed` mean? docs... what's the deadline? a deadline of a single call??

kolloch: ah, deadline is more like `backOffUntil`, and `failed` is a somehow scoped failure counter...

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
   implicit val ec = HttpEventModule.executionContext
   implicit val timeout = HttpEventModule.timeout
+  val pipeline: HttpRequest => Future[HttpResponse] = addHeader("Accept", "application/json") ~> sendReceive
+  var limiter = Map.empty[String, EventNotificationLimit].withDefaultValue(NoLimit)
 
-  val pipeline: HttpRequest => Future[HttpResponse] = (
-    addHeader("Accept", "application/json")
-    ~> sendReceive)
+  //metrics
+  val metricOutstandingRequests = metrics.counter("service.mesosphere.marathon.event.http.outstanding-requests")
```
kolloch: metrics in codahale may only be registered once. Since actors can be restarted, please create an additional object for the metrics that survives actor restarts.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
   def receive: Receive = {
-    case event: MarathonEvent =>
-      broadcast(event)
-    case _ =>
-      log.warning("Message not understood!")
+    case event: MarathonEvent     => broadcast(event)
+    case NotificationSuccess(url) => limiter += url -> NoLimit
+    case NotificationSlow(url)    => limiter += url -> limiter(url).nextFailed
+    case NotificationFailed(url)  => limiter += url -> limiter(url).nextFailed
+    case _                        => log.warning("Message not understood!")
```
kolloch: Since you treat slow/failed the same right now, I would simplify the code by getting rid of `NotificationSlow`. Or do you distinguish somewhere else?

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
kolloch: documentation for the metrics would be nice, too, but I don't insist ;)

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
     log.info("POSTing to all endpoints.")
-    (subscribersKeeper ? GetSubscribers).mapTo[EventSubscribers].foreach {
-      _.urls.foreach { post(_, event) }
+    val me = self
+    (subscribersKeeper ? GetSubscribers).mapTo[EventSubscribers].foreach { subscribers =>
+      val (active, limited) = subscribers.urls.partition(limiter(_).notLimited)
+      if (limited.nonEmpty) log.info(s"Will not send event ${event.eventType} to unresponsive hosts: $limited")
+      metricDroppedEvent.mark(limited.size)
+      active.foreach(post(_, event, me))
     }
```
kolloch: This would be a good place to delete limiter entries for subscribers that no longer exist... Or am I missing the cleanup code?

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
+    }
+    response.onComplete {
       case Success(res) =>
-        if (res.status.isFailure)
-          log.warning(s"Failed to post $event to $urlString")
-
-      case Failure(t) =>
-        log.warning(s"Failed to post $event to $urlString", t)
+        if (res.status.isFailure) log.debug(s"No success response for post $event to $url")
+        val inTime = (currentTimeMillis() - start) < conf.slowConsumerTimeout
```
kolloch: So HTTP-level failures (500, ...) are not treated as failures? I think it is more polite doing so because if the receiving end is overloaded, we will desist. Of course, this changes the behavior a bit.

Then, (because we back off and prevent spamming the logs), I would at least log the HTTP status code + status string as warning.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
+        eventActor ! (if (inTime) NotificationSuccess(url) else NotificationSlow(url))
+      case _ =>
+        log.warning(s"Failed to post $event to $url")
```
kolloch: You do not want to spam the log with stack traces here? Or you want to keep using akka logging? 

Anyway, maybe the exception name+message would already provide some more info.

src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
```diff
@@ -26,8 +25,17 @@ trait HttpEventConfiguration extends ScallopConf {
     required = false,
     noshort = true).map(parseHttpEventEndpoints)
 
+  lazy val httpEventCallbackSlowConsumerTimeout = opt[Long]("http_endpoint_slow_consumer_duration",
```
kolloch: scala variable name does not match CLI string. I guess the string is incorrect.

src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
```diff
@@ -26,8 +25,17 @@ trait HttpEventConfiguration extends ScallopConf {
...
+    descr = "A http consumer is considered slow, if the delivery takes longer than this timeout (ms)",
```
kolloch: you probably want to adjust this as well, even if it is abstract enough to match callbacks, too, in principle ;)

src/test/scala/mesosphere/marathon/event/http/HttpEventActorTest.scala
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+
+  test("If a message is send to non existing subscribers") {
+    Given("A HttpEventActor with 2 subscribers")
+    val aut = TestActorRef(new NoHttpEventActor(Set("host1", "host2")))
+    responseAction = () => throw new RuntimeException("Can not connect")
+
+    When("An event is send to the actor")
+    aut ! EventStreamAttached("remote")
```
kolloch: you also can use `out.receive` here to synchronously call the actor.

src/test/scala/mesosphere/marathon/event/http/HttpEventActorTest.scala
```diff
@@ -0,0 +1,112 @@
...
+    Then("The callback listener is rate limited")
+    waitUntil("Wait for rate limiting 2 subscribers", 1.second) {
+      aut.underlyingActor.limiter("host1").deadLine.isDefined && aut.underlyingActor.limiter("host2").deadLine.isDefined
+    }
+  }
+
+  test("If a message is send to a slow subscriber") {
+    Given("A HttpEventActor with 1 subscriber")
+    val aut = TestActorRef(new NoHttpEventActor(Set("host1")))
+    responseAction = () => { Thread.sleep(100); response }
```
kolloch: Ouch, sleeps in tests. Should I less that pass for releasing-0.10? I wouldn't like that on master...

Alternatively, you can virtualize your time by making the method to determine the time `protected` and overridable so that you can mock it for testing.

src/test/scala/mesosphere/marathon/event/http/HttpEventActorTest.scala
```diff
@@ -0,0 +1,112 @@
...
+    EventFilter.info(start = "Will not send event event_stream_attached to unresponsive hosts: Set(host1)") intercept {
+      aut ! EventStreamAttached("remote")
+    }
+
+    And("The message is send to the other subscriber")
+    waitUntil("Wait for 1 subscribers to get notified", 1.second) {
+      aut.underlyingActor.requests.size == 1
+    }
+  }
+
```
kolloch: Nice tests!

Some more test cases:

* limiters for subscribers that do not exist anymore are cleaned up.
* HTTP status != 2xx are logged.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
 
   def broadcast(event: MarathonEvent): Unit = {
+    metricSendEvent.mark()
     log.info("POSTing to all endpoints.")
-    (subscribersKeeper ? GetSubscribers).mapTo[EventSubscribers].foreach {
-      _.urls.foreach { post(_, event) }
+    val me = self
+    (subscribersKeeper ? GetSubscribers).mapTo[EventSubscribers].foreach { subscribers =>
+      val (active, limited) = subscribers.urls.partition(limiter(_).notLimited)
+      if (limited.nonEmpty) log.info(s"Will not send event ${event.eventType} to unresponsive hosts: $limited")
```
kolloch: Nit-pick: I don't like "`Set(...)`" in log messages. But that's not blocking a merge...

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
+  val pipeline: HttpRequest => Future[HttpResponse] = addHeader("Accept", "application/json") ~> sendReceive
+  var limiter = Map.empty[String, EventNotificationLimit].withDefaultValue(NoLimit)
 
-  val pipeline: HttpRequest => Future[HttpResponse] = (
-    addHeader("Accept", "application/json")
-    ~> sendReceive)
+  //metrics
+  val metricOutstandingRequests = metrics.counter("service.mesosphere.marathon.event.http.outstanding-requests")
+  val metricSendEvent = metrics.meter("service.mesosphere.marathon.event.http.send")
+  val metricDroppedEvent = metrics.meter("service.mesosphere.marathon.event.http.dropped-events")
```
kolloch: I think that the metric name is slightly misleading. The events themselves are not dropped but only calling certain callbacks. It is confusing that "send" is measured per event and "dropped-events" per callback. I do like the current position of the metrics, though. So keep the semantic but rename the metrics? E.g.:

* "send" => "events-processed"
* "dropped-events" => "skipped-rate-limited-callbacks" or "skipped-callbacks"

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -2,53 +2,91 @@ package mesosphere.marathon.event.http
...
+  val metricResponseTime = metrics.timer("service.mesosphere.marathon.event.http.response-time")
```
kolloch: "callback-response-time" would be clear to me.

kolloch: Intuitively, I am missing "failed-callbacks"/"slow-callbacks" as metrics. What do you think?

src/main/java/mesosphere/marathon/metrics/Metrics.scala
```diff
@@ -64,6 +68,11 @@ class Metrics @Inject() (val registry: MetricRegistry) {
 }
 
 object Metrics {
+  class Counter(counter: com.codahale.metrics.Counter) {
```
kolloch: Actually, this is great. We missed the `Counter` type before and used stuff like the `AtomicIntGauge` instead. I am just not sure whether it's a good idea to introduce this in this patch.

Either way, it probably makes sense to create an issue and change `master` to using counters instead of `AtomicIntGauge`?

Also, please document that it is a gauge, since counter is not a metric category in statsd, I wouldn't know where to find it otherwise.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
kolloch: Usually, we use "metrics.name" to create the metrics name in combination with `MetricPrefixes`. I am really not sure if I like that approach but for consistency it would be better to stick to that, I guess.

kolloch:
Great functionality!

I wrote some comments...

src/test/scala/mesosphere/marathon/event/http/HttpEventActorTest.scala
aquamatthias: I will ignore this test. In master we should use clock instead of System.currentTimeMillis

src/test/scala/mesosphere/marathon/event/http/HttpEventActorTest.scala
aquamatthias: Not 100% sure - the synchronous thing will help here.

src/main/java/mesosphere/marathon/metrics/Metrics.scala
aquamatthias: Counters and Gauges are a separate thing in codahale metrics. Don't know what to document here if it is statsd specific?

aquamatthias:
Thanks for the comments! Hope I addressed them all.

src/test/scala/mesosphere/marathon/event/http/HttpEventActorTest.scala
aquamatthias: Added a test for "unlimiting" a previous limited subscriber.

src/main/java/mesosphere/marathon/metrics/Metrics.scala
kolloch: You are right. For some reason, I didn't realize that it is a distinct metrics type.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
kolloch: Ack.

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
kolloch: Ack.

src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
kolloch: Ack

src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala
kolloch: Ack

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
kolloch: Ack, thanks.

src/main/java/mesosphere/marathon/metrics/Metrics.scala
kolloch: Ack.

kolloch:
Please investigate test failures. Otherwise, LGTM!

src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala
```diff
@@ -1,54 +1,114 @@
...
 
-  def post(urlString: String, event: MarathonEvent): Unit = {
-    log.info("Sending POST to:" + urlString)
+  def post(url: String, event: MarathonEvent, eventActor: ActorRef): Unit = {
+    log.info("Sending POST to:" + url)
 
-    val request = Post(urlString, eventToJson(event))
+    metrics.outstandingCallbacks.inc()
+    val request = Post(url, eventToJson(event))
     val response = pipeline(request)
```
kolloch: Is it difficult to specify timeouts for individual requests?

aquamatthias:
@kolloch the tests consistently fail without this patch - pretty sure this is not introduced by this PR. Hopefully fixed the shaky assertions.

aquamatthias:
I run into all shaky integration tests with this PR. I am pretty sure, the problem is not related to this PR.
The unit and integration tests run fine on my local machine.
I will create another PR based on master, so we can be sure about that.

kolloch:
:ship: !

PR #3084, opened 2016-01-25
Fixes #2692 - Document important metrics

### LINKED ISSUE #2692: Document the most important metrics
Recently, someone ask on the mailing list whether we have documentation on which metrics we do expose.

Documenting all metrics in a generally understandable way is probably a lot of work. But we should identify a subset of metrics that is important to most of our customers. I guess our guideline should be which metrics can quantify "Marathon health" and be used to make operations aware of likely problems, e.g. very slow responses.

We appreciate community feedback on what kind of metrics are most important.

kolloch:


kolloch:
@graham-murphy / @jgarcia-mesosphere : I'd really appreciate your feedback on this. We have to keep it focused, though because we didn't allocate time for this. Thanks a lot!

docs/docs/command-line-flags.md
```diff
@@ -281,7 +281,7 @@ The Web Site flags control the behavior of Marathon's web site, including the us
 
 * <span class="label label-default">v0.13.0</span> `--[disable_]metrics` (Optional. Default: enabled):
     Expose the execution time per method via the metrics endpoint (/metrics).
-    This metrics measurement can be disabled with --disable_metrics.
+    These measurements can be disabled with --disable_metrics. Other metrics are not affected.
```
gkleiman: We should do this change also in DebugConf.scala.

docs/docs/metrics.md
```diff
@@ -0,0 +1,150 @@
...
+
+Marathon stores its permanent state in "Repositories". These are the most important ones:
+
+* `GroupRepository` for app configurations and groups.
+* `TaskRepository` for the latest known task state. This is the repository with the largest
+  data churn.
+
+There are also the
+
+* AppRepository for versioned app configuration.
```
gkleiman: I'd quote the repository names to be consistent with the rest of the document.

docs/docs/metrics.md
```diff
@@ -0,0 +1,150 @@
...
+* DeploymentRepository for currently running deployments.
+* TaskFailureRepository for the last failure for every application.
+* and more...
+
+For each of the repositories we have statistics about read and write requests. Substitute `*` with
+the repository of interest:
+ 
+`service.mesosphere.marathon.state.*.read-request-time.count` - the number of read requests.
```
gkleiman: Would this be more readable as a list instead of different paragraphs?

docs/docs/metrics.md
```diff
@@ -0,0 +1,150 @@
...
+
+`service.mesosphere.marathon.state.*.read-request-time.mean` - the exponential weighted average of
```
gkleiman: I think the 1-, 5-, and 15-minute moving averages are more useful than the "global" mean. Do we expose them?

docs/docs/metrics.md
```diff
@@ -0,0 +1,150 @@
...
+
+`jvm.threads.count` (meter) - the total number of threads. If this is above >500, this
+is generally a bad sign.
+
+`jvm.memory.total.used` (meter) - the total number of bytes used by the Marathon JVM.
+
+## Instrumented method calls
+
+These metrics are created automatically by instrumenting certain classes in our code base which can be disabled
+by `--disable_metrics` which will not disable all metrics. Generally, these timers can be very valuable in diagnosing
```
gkleiman: I'm very bad with prepositions in english. But I think the following sounds better: s/by `--disable_metrics`/with `disable_metrics`

docs/docs/command-line-flags.md
kolloch: Ack!

docs/docs/metrics.md
kolloch: Not we hourselves but other metric solutions might layer on top of that. But as with the mean I don't really know how to interpret them other than "oh, that's really long", "that's probably not the problem". The problem is that we don't really report this reliably for reporting intervals in combination with technologies as datadog, statsd, graphite.

```
"service.mesosphere.marathon.state.AppRepository.write-request-time": {
      "count": 43,
      "max": 3,
      "mean": 1.1844599999513041,
      "min": 1,
      "p50": 1.0,
      "p75": 1.0,
      "p95": 2.0,
      "p98": 2.0,
      "p99": 2.0,
      "p999": 2.0,
      "stddev": 0.3878588768731083
    },
```

kolloch:
Thanks for you suggestions! BTW @philipnrmn agrees with your preference of "with" in this case :)

docs/docs/metrics.md
kolloch: Do you mean prefixing the items with "* " in the markdown?

Actually, for me the dot for the list would only contribute useless clutter. I guess it is a matter of taste.

Have you looked at the rendered output?

If you feel strongly about this, I can change it.

kolloch:
Added a link from the README, thanks for the hint!

gkleiman:
:ship: :shipit: :rocket:

PR #3098, opened 2016-01-26
Prefer Scala Numeric classes over Java classes

gkleiman:
Now that we have play-json, I see no reason for using java.lang.integer and
java.lang.double instead of Scala's classes.

cc/ @aquamatthias


src/main/scala/mesosphere/marathon/api/EndpointsHelper.scala
```diff
@@ -35,7 +35,7 @@ object EndpointsHelper {
           sb.append(cleanId).append(delimiter).append(port).append(delimiter)
 
           for (task <- tasks if task.getStatus.getState == TaskState.TASK_RUNNING) {
-            val taskPort = Option(task.getPortsList.get(i): java.lang.Integer).getOrElse(java.lang.Integer.valueOf(0))
+            val taskPort: Integer = Option(task.getPortsList.get(i): Integer).getOrElse(0)
             sb.append(task.getHost).append(':').append(taskPort).append(delimiter)
```
kolloch: `Integer` is `java.lang.Integer`... `Int` is Scala.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -147,9 +145,9 @@ trait Formats
     Reads.of[Int].map(Int.box),
     Writes[Integer] { i => JsNumber(i.intValue) }
   )
-  implicit lazy val DoubleFormat: Format[JDouble] = Format(
+  implicit lazy val DoubleFormat: Format[java.lang.Double] = Format(
```
kolloch: Unneeded, IntegerFormat as well.

kolloch:
Scalas integer type is `Int`

gkleiman:
@kolloch thanks for the review :blush:

I addressed your comments, now we have `Int` almost everywhere =).

kolloch:
Thank you! Nice endeavour! :+1: :shipit:

PR #3099, opened 2016-01-26
Fixes #3059 - Reset task's delay only after viability threshold has been reached

### LINKED ISSUE #3059: Task launch delay reset every 10 seconds
TasksLaunchDelays are reset every 10 seconds by CleanupOverdueDelays
events scheduled there:

https://github.com/mesosphere/marathon/blob/v0.14.0/src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiterActor.scala#L50

This event triggers the cleanUpOverdueDelays method, which drops
current launchDelays information of every task, which deadline has
been reached:

https://github.com/mesosphere/marathon/blob/v0.14.0/src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala#L24

As a consequence, if the CleanupOverdueDelays event occurs between the
launch of a task and its failure, the task's launch delay is
reset.

The longer the task takes to fail, the higher the probability of its
TaskLaunchDelay to be reset. And if task takes more than 10 seconds to
fail, its TaskLaunchDelay is systematically reset.

This bug disables the Exponential Backoff mechanism.

FlorentFlament:


src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala
```diff
@@ -19,13 +19,21 @@ private[launchqueue] class RateLimiter(clock: Clock) {
...
 
-  def cleanUpOverdueDelays(): Unit = {
+  def getMinimumTaskExecutionSeconds: Long = minimumTaskExecutionSeconds
+
+  /**
+    * Reset delay for tasks that have reached the viability
+    * threshold. The deadline indicates when the task has been
+    * launched for the last time.
+    */
+  def resetViableTasksDelays(): Unit = {
```
gkleiman: `resetDelaysOfViableTasks` might be a better name, but I have no strong opinion about this.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala
```diff
@@ -69,10 +77,16 @@ private[launchqueue] class RateLimiter(clock: Clock) {
 }
 
 private object RateLimiter {
+  /**
+    * The viability threshold may be tuned later, once the mechanism
+    * has been validated.
+    */
+  private val minimumTaskExecutionSeconds: Long = 60
```
gkleiman: I'd make this a hidden flag, see for example: https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueueConfig.scala

gkleiman:
@FlorentFlament  Thanks for the PR! I added a few comments, but in general it looks very good to me =).

@kolloch is much more familiar with this part of Marathon, so I would feel more confident if he would look over the PR before merging it.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala
FlorentFlament: ok

src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala
FlorentFlament: Oh, it's even better if the parameter can be tuned through command line flags .. What do you think about not having this parameter hidden ?

FlorentFlament:
Hi, any news on this PR ?

jasongilanfarr:
@aquamatthias and @meichstedt   This looks pretty good to me, I can handle getting it up to date. Any thoughts?

aquamatthias:
@jasongilanfarr I agree - please proceed.

pierrecdn:
Hi,
I just rebased this on master and (hopefully) solved the conflicts.
We use this patch since 0.13 (was cherry-picked on our side for 0.14.0, 0.15.0, 0.15.3, 1.1.1, 1.3.0)

pierrecdn:
Solved newly introduced conflicts.
Don't really know if this was addressed by another commit at some point.

aquamatthias:
ok to test

aquamatthias:
Thanks @FlorentFlament and sorry for the (too) long time to handle this PR!
Since this change did not compile any longer - I took the liberty to cherry pick your commit and put the necessary changes on top as a separate commit. See #4915

PR #3106, opened 2016-01-27
Enable http based plugin extensions

aquamatthias:
By defining an abstract http handler trait, as well as a PluginsResource to handle plugin requests.
Idea:
-GET /v2/plugins will list all available HttpHandler implementations
-GET/HEAD/PUT/POST/DELETE /v2/plugins/<context_path>/xxx/yyy will be handled by the plugin  that handles the defined context path (defined by the plugin). The path /xxx/yyy has to be handled by the plugin.

I will create an example in the marathon-example-plugins project.

@orlandohohmeier @aldipower: comments welcome

### DIFF
--- a/plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
+++ b/plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
@@ -1,0 +1,38 @@
+package mesosphere.marathon.plugin.http
+
+import java.net.{ URL, URLConnection }
+import com.google.common.io.Resources
+
+/**
+  * The HttpRequestHandlerBase extends the HttpRequestHandler by
+  * providing functionality to serve content packaged as resource.
+  */
+//scalastyle:off magic.number
+abstract class HttpRequestHandlerBase extends HttpRequestHandler {
+
+  protected[this] def serveResource(path: String, response: HttpResponse): Unit = {
+    val content = withResource(path) { url =>
+      response.body(mediaMime(url), Resources.toByteArray(url))
+      response.status(200)
+    }
+    content.getOrElse(response.status(404))
+  }
+
+  protected[this] def withResource[T](path: String)(fn: URL => T): Option[T] = {
+    Option(getClass.getResource(path)).map(fn)
+  }
+
+  protected[this] def mediaMime(url: URL): String = {
+    url.getPath.split("\\.").lastOption.flatMap(wellKnownMimes.get)
+      .orElse(Option(URLConnection.guessContentTypeFromName(url.toString)))
+      .getOrElse("application/octet-stream")
+  }
+
+  protected[this] val wellKnownMimes = Map (
+    "css" -> "text/css",
+    "js" -> "application/javascript",
+    "eot" -> "application/vnd.ms-fontobject",
+    "svg" -> "image/svg+xml",
+    "ttf" -> "application/font-ttf"
+  )
+}
--- a/src/main/scala/mesosphere/marathon/api/v2/PluginsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/PluginsResource.scala
@@ -1,0 +1,66 @@
+package mesosphere.marathon.api.v2
+
+import javax.inject.Inject
+import javax.servlet.http.HttpServletRequest
+import javax.ws.rs._
+import javax.ws.rs.core.{ Context, Response }
+
+import mesosphere.marathon.MarathonConf
+import mesosphere.marathon.api.v2.json.Formats._
+import mesosphere.marathon.api.{ MarathonMediaType, RequestFacade, ResponseFacade, RestResource }
+import mesosphere.marathon.core.plugin.PluginDefinitions
+import mesosphere.marathon.plugin.http.{ HttpRequest, HttpRequestHandler, HttpResponse }
+
+@Path("v2/plugins")
+class PluginsResource @Inject() (val config: MarathonConf,
+                                 requestHandlers: Seq[HttpRequestHandler],
+                                 definitions: PluginDefinitions) extends RestResource {
+
+  val pluginIdToHandler = definitions.plugins
+    .filter(_.plugin == classOf[HttpRequestHandler].getName)
+    .flatMap { d => requestHandlers.find(_.getClass.getName == d.implementation).map(d.id -> _) }
+    .toMap
+
+  @GET
+  @Produces(Array(MarathonMediaType.PREFERRED_APPLICATION_JSON))
+  def plugins(): Response = ok(jsonString(definitions))
+
+  @GET
+  @Path("""{pluginId}/{path:.+}""")
+  def get(@PathParam("pluginId") pluginId: String,
+          @PathParam("path") path: String,
+          @Context req: HttpServletRequest): Response = handleRequest(pluginId, path, req)
+
+  @HEAD
+  @Path("""{pluginId}/{path:.+}""")
+  def head(@PathParam("pluginId") pluginId: String,
+           @PathParam("path") path: String,
+           @Context req: HttpServletRequest): Response = handleRequest(pluginId, path, req)
+
+  @PUT
+  @Path("""{pluginId}/{path:.+}""")
+  def put(@PathParam("pluginId") pluginId: String,
+          @PathParam("path") path: String,
+          @Context req: HttpServletRequest): Response = handleRequest(pluginId, path, req)
+
+  @POST
+  @Path("""{pluginId}/{path:.+}""")
+  def post(@PathParam("pluginId") pluginId: String,
+           @PathParam("path") path: String,
+           @Context req: HttpServletRequest): Response = handleRequest(pluginId, path, req)
+
+  @DELETE
+  @Path("""{pluginId}/{path:.+}""")
+  def delete(@PathParam("pluginId") pluginId: String,
+             @PathParam("path") path: String,
+             @Context req: HttpServletRequest): Response = handleRequest(pluginId, path, req)
+
+  private[this] def handleRequest(pluginId: String, path: String, req: HttpServletRequest): Response = {
+    pluginIdToHandler.get(pluginId).map { handler =>
+      val request = new RequestFacade(req, path)
+      val response = new ResponseFacade
+      handler.serve(request, response)
+      response.response
+    }.getOrElse(notFound(s"No plugin with this pluginId: $pluginId"))
+  }
+}
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -6,6 +6,7 @@ import mesosphere.marathon.Protos.ResidencyDefinition.TaskLostBehavior
 import mesosphere.marathon.Protos.{ Constraint, MarathonTask }
 import mesosphere.marathon.core.appinfo._
 import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.plugin.{ PluginDefinitions, PluginDefinition }
 import mesosphere.marathon.event._
 import mesosphere.marathon.event.http.EventSubscribers
 import mesosphere.marathon.health.{ Health, HealthCheck }
@@ -55,6 +56,7 @@ trait Formats
     with DeploymentFormats
     with EventFormats
     with EventSubscribersFormats
+    with PluginFormats
     with IpAddressFormats {
   import scala.collection.JavaConverters._
 
@@ -463,6 +465,7 @@ trait FetchUriFormats {
 }
 
 trait AppAndGroupFormats {
+
   import Formats._
 
   implicit lazy val IdentifiableWrites = Json.writes[Identifiable]
@@ -473,7 +476,7 @@ trait AppAndGroupFormats {
     (
       (__ \ "minimumHealthCapacity").readNullable[Double].withDefault(DefaultUpgradeStrategy.minimumHealthCapacity) ~
       (__ \ "maximumOverCapacity").readNullable[Double].withDefault(DefaultUpgradeStrategy.maximumOverCapacity)
-    )(UpgradeStrategy(_, _))
+    ) (UpgradeStrategy(_, _))
   }
 
   implicit lazy val ConstraintFormat: Format[Constraint] = Format(
@@ -527,7 +530,7 @@ trait AppAndGroupFormats {
       .withDefault(AppDefinition.DefaultMaxLaunchDelay.toSeconds).asSeconds ~
       (__ \ "container").readNullable[Container] ~
       (__ \ "healthChecks").readNullable[Set[HealthCheck]].withDefault(AppDefinition.DefaultHealthChecks)
-    )((
+    ) ((
         id, cmd, args, maybeString, env, instances, cpus, mem, disk, executor, constraints, storeUrls,
         requirePorts, backoff, backoffFactor, maxLaunchDelay, container, checks
       ) => AppDefinition(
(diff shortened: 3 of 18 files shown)


aquamatthias:
Example plugin can be found here: mesosphere/marathon-example-plugins#8

gkleiman:
I like what I see =)

orlandohohmeier:
Yeah! This is very near to what I had in mind thinking about the Marathon UI Plugin integration. Thanks a bunch :bouquet: 

I would like to propose a slightly adjusted plugin configuration as well as path mapping and loading process,  after playing with your WIP implementation.

## Context Path

Consider having two plugins serving files, if both of them use the same `contextPath` (e.g. `/static`) you might get unwanted results. This could easily avoided if we define (unique) plugin ids and use those as context path.

This is how `plugin-conf.json` with ids could like:
 
```
{
  "experiment": {
    "plugin": "mesosphere.marathon.plugin.http.HttpRequestHandler",
    "implementation": "mesosphere.marathon.example.plugin.experiment.WebJarHandler"
  },
  "http": {
    "plugin": "mesosphere.marathon.plugin.http.HttpRequestHandler",
    "implementation": "mesosphere.marathon.example.plugin.http.WebJarHandler"
  },
  "example-authorizer": {
    "plugin": "mesosphere.marathon.plugin.auth.Authorizer",
    "implementation": "my.company.ExampleAuthorizer"
  },
  "example-authenticator": {
    "plugin": "mesosphere.marathon.plugin.auth.Authenticator",
    "implementation": "my.company.ExampleAuthenticator",
    "configuration": {
      "users": [
        {
          "user": "ernie",
          "password": "ernie",
          "permissions": [
            {
              "allowed": "create",
              "on": "/dev/"
            }
          ]
        }
      ]
    }
  }
}
```

*BWT: I was wondering if we really need all the plugin/implementation stuff...*

## Isolation

At the moment I have a one plugin serving just some files (static) and another one serving a gui plugin, but due to the lack of isolation both serve the same files as they use the same resource path (`/META-INF/resources/webjars/static/`). 

Plugin endpoint (`/v2/plugins`) response:              

```
[
    {
        "contextPath": "gui", 
        "implementation": "mesosphere.marathon.example.plugin.gui.WebJarHandler"
    }, 
    {
        "contextPath": "static", 
        "implementation": "mesosphere.marathon.example.plugin.http.WebJarHandler"
    }
]
```

Static Plugin (`/v2/plugins/static/package.json`) response:
```
{	
	"description": "Example Plugin", 
    "main": "main.js" 
}
```

Gui Plugin (`/v2/plugins/gui/package.json`) response:
```
{	
	"description": "Example Plugin", 
    "main": "main.js" 
}
```


Is there a way to restrict/isolate them? It would also be nice if only plugins (jars) that are actually defined within the `plugin-conf.json` are loaded as the active plugins even serve files from inactive plugins. 

--

What do you think?

aquamatthias:
@orlandohohmeier: Cool! And thanks for the feedback.

Let me elaborate a little bit on the plugin mechanism:
- Plugins are (java/scala) classes identified by name. 
- There can be only one class with the same name.
- It would be possible to have multiple instances of the same class.
Having two plugins defining the same class will not work. There can be multiple plugins that implement the same plugin interface, but the class has to be different. 

The contextPath is defined in the plugin interface itself: [WebJarHandler](https://github.com/mesosphere/marathon-example-plugins/blob/mv/http_example/http/src/main/scala/mesosphere/marathon/example/plugin/http/WebJarHandler.scala#L16)
So your example of having:
-`mesosphere.marathon.example.plugin.gui.WebJarHandler` serving `/v2/plugins/static`
-`mesosphere.marathon.example.plugin.http.WebJarHandler` serving `/v2/plugins/gui` 
would work perfectly well. It would be also very easy to define the contextPath in the configuration.

Example:
```json
{
  "plugins": [
    {
      "plugin": "mesosphere.marathon.plugin.http.HttpRequestHandler",
      "implementation": "mesosphere.marathon.example.plugin.http.WebJarHandler",
      "configuration": {
        "contextPath": "gui"
      }
    }
  ]
}
```

I like your proposal of defining the list of plugins as objects with unique identifier instead of an array.
Need to think about it.

orlandohohmeier:
@aquamatthias The current implementation does not serve javascript files. I always get `500 Internal Server Error` with the following JSON response body: `{ "message": null }`. Could you please have a look?

aquamatthias:
@orlandohohmeier good catch. The problem was the media mime mapping - the nio logic does not work for zip file entries (I still do not understand why). I replaced that part - should work now.

orlandohohmeier:
@aquamatthias Thanks for the fix! Works like a charm :100:

aquamatthias:
@gkleiman any updates here?

plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.{ ByteArrayOutputStream, InputStream, OutputStream }
+import java.net.{ URL, URLConnection }
+import java.nio.file.{ Files, Paths }
+
+import scala.annotation.tailrec
+import scala.util.Try
+
+//scalastyle:off magic.number
+abstract class HttpRequestHandlerBase extends HttpRequestHandler {
```
gkleiman: Maybe add some comments describing the class?

plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  protected[this] def serveResource(path: String, response: HttpResponse): Unit = {
+    def transfer(in: InputStream, out: OutputStream) {
```
gkleiman: Have you thought about using Streams + transferTo? The code would look cleaner and I think it uses sendFile under-the-hood, so it would also perform better.

scalastyle-config.xml
```diff
@@ -3,7 +3,7 @@
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"/>
  <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxFileLength"><![CDATA[800]]></parameter>
```
gkleiman: If we do this, then we can remove the `// scalastyle:off file.size.limit` line from Formats.scala

src/main/scala/mesosphere/marathon/api/v2/PluginsResource.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import mesosphere.marathon.MarathonConf
+import mesosphere.marathon.api.v2.json.Formats._
+import mesosphere.marathon.api.{ MarathonMediaType, RequestFacade, ResponseFacade, RestResource }
+import mesosphere.marathon.core.plugin.PluginsDescriptor
+import mesosphere.marathon.plugin.http.{ HttpRequest, HttpRequestHandler, HttpResponse }
+
+@Path("v2/plugins")
+class PluginsResource @Inject() (val config: MarathonConf,
+                                 requestHandler: Seq[HttpRequestHandler],
```
gkleiman: s/requestHandler/requestHandlers/

src/main/scala/mesosphere/marathon/api/v2/PluginsResource.scala
```diff
@@ -0,0 +1,66 @@
...
+                                 descriptor: PluginsDescriptor) extends RestResource {
+
+  val handlerMap = descriptor.plugins
```
gkleiman: What about calling this pluginIdToHandler?

src/main/scala/mesosphere/marathon/api/v2/PluginsResource.scala
```diff
@@ -0,0 +1,66 @@
...
+    .filter(_.plugin == classOf[HttpRequestHandler].getName)
+    .flatMap { d => requestHandler.find(_.getClass.getName == d.implementation).map(d.id -> _) }
+    .toMap
+
+  @GET
+  @Produces(Array(MarathonMediaType.PREFERRED_APPLICATION_JSON))
+  def plugins(): Response = ok(jsonString(descriptor))
+
+  @GET
+  @Path("""{path}/{pluginPath:.+}""")
```
gkleiman: `@Path("""{pluginId}/{path:.+}""")` might be a bit clearer.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -782,3 +784,16 @@ trait V2Formats {
...
   )(Group(_, _, _, _, _), unlift(Group.unapply))
 }
+
+trait PluginFormats {
+
+  implicit lazy val pluginDefinitionFormat: Writes[PluginDefinition] = (
+    (__ \ "id").write[String] ~
+    (__ \ "plugin").write[String] ~
+    (__ \ "implementation").write[String] ~
+    (__ \ "tags").writeNullable[Set[String]] ~
```
gkleiman: what are the tags for?

src/main/scala/mesosphere/marathon/core/plugin/PluginDescriptor.scala
```diff
@@ -0,0 +1,12 @@
...
+import play.api.libs.json.JsObject
+
+case class PluginDefinition(id: String,
+                            plugin: String,
+                            implementation: String,
+                            tags: Option[Set[String]],
+                            configuration: Option[JsObject],
+                            info: Option[JsObject])
+
+case class PluginsDescriptor(plugins: Seq[PluginDefinition])
```
gkleiman: I find `PluginsDescriptor` and `PluginDefinition` maybe too similar. What do you think about renaming `PluginsDescriptor` to `PluginsConfiguration` or maybe just `Plugins`?

src/main/scala/mesosphere/marathon/core/plugin/impl/PluginManagerImpl.scala
```diff
@@ -18,7 +23,7 @@ import scala.reflect.ClassTag
   * The plugin manager can load plugins from given urls.
   * @param urls the urls pointing to plugins.
   */
-private[plugin] class PluginManagerImpl(val descriptor: PluginsDescriptor, val urls: Seq[URL]) extends PluginManager {
+class PluginManagerImpl(val descriptor: PluginsDescriptor, val urls: Seq[URL]) extends PluginManager {
```
gkleiman: Why did you change the visibility of this class? It doesn't seem like it is being used outside of the plugin package.

gkleiman:
@aquamatthias sorry for the delay, made a couple of comments.

This also needs to be rebased and I think we also should update the plugins documentation and add something to the change log.

orlandohohmeier:
I'd like to propose another plugin descriptor update which also supports plugins that not only use the `HttpRequestHandler` but also other plugin interfaces.

```
{
  "plugins": {
    "my-cool-plugin-id": {
      "implementations": {
            "mesosphere.marathon.plugin.http.HttpRequestHandler":
                "mesosphere.marathon.plugin.myCoolPlugin.HttpRequestHandler",
            "mesosphere.marathon.plugin.other.CoolPluginThing":
                "mesosphere.marathon.plugin.myCoolPlugin.other.CoolPluginThing"
        }
      "info": {
        "name": "My Cool Marathon Plugin",
        "modules": ["ui", "service"],
        "description": ""
      }
    }
  }
}
```

What do you think?

plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
aquamatthias: Done.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
aquamatthias: No. Can you make a simple example?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
aquamatthias: Added a sentence in the documentation. Basically `tags` and `info` is there to add meta data to the plugin.

aquamatthias:
@orlandohohmeier not sure what the goal of this proposal is: 
- one jar can provide multiple plugins. 
- there can be multiple implementations of the same plugin interface
Can you describe the problem/goal more specific?

Second: your proposal breaks identity: there is one id, but 2 plugins.

orlandohohmeier:
@aquamatthias I know that it is possible to bundle several plugins in one jar. And it probably boils down to me still thinking about plugins as Marathon plugins and not `ServiceLoader` plugins. In my point of view using and configuring a plugin should be as easy as possible. I could think of cases in which one plugin or part of a plugin is useless without the other thus I think they should be defined and handle as one Marathon plugin. So yeah, it's one Marathon plugin (plugin id) and several  `ServiceLoader` plugins.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/http/HttpRequestHandlerBase.scala
gkleiman: Forget what I said about transferTo, that's only useful if you're writing directly to a socket (using a Channel).

But Marathon already depends on Guava, so we can get rid of this method and do:

```scala
val content = withResource(path) { url =>
  response.body(mediaMime(url), Resources.toByteArray(url))
  response.status(200)
}
content.getOrElse(response.status(404))
```

aquamatthias: Then guava becomes a dependency to all plugins.

aquamatthias: I tried `Files.copy` which looked promising. But it does not handle files from a zip file... :(

gkleiman: Since Marathon already depends on it, it would be a `provided` dependency, like logback.

aquamatthias: I agree - this makes it very easy. Thanks for the hint.

aquamatthias:
@gkleiman Hope I addressed all comments. Squashed. Ready to merge :)

src/main/scala/mesosphere/marathon/core/plugin/PluginDescriptor.scala
```diff
@@ -0,0 +1,12 @@
+package mesosphere.marathon.core.plugin
```
gkleiman: Nit: The file is called `PluginDescriptor.scala`, but there's no `PluginDescriptor` class anymore.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
gkleiman: What's the goal of this meta data? It is not available to the plugins nor used by Marathon itself.

Is it to work around not being able to write comments in the configuration file? If that's the case, isn't `info` enough?

gkleiman:
LGTM.

Just a comment about the renaming of PluginDescriptor to Plugins not being complete.

aquamatthias:
@gkleiman Renamed the file and squashed again.

aquamatthias:
@Renamed the parameter and all usages. Added RAML documentation.

gkleiman:
:rocket:

Thanks! Waiting for the tests to pass before mergin'.

PR #3115, opened 2016-01-28
Fixes #3040 - use acceptOffer to launch tasks

### LINKED ISSUE #3040: Use acceptOffers to launch tasks
Extend our Mesos API Interface to include the new acceptOffers calls to allow specifying `Offer:Operation:Reserve` and `Offer:Operation:Create` operations.

kolloch:
Note that some renames are in a separate commit to make the real changes better to see.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/TaskLauncher.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/TaskLauncher.scala
@@ -1,6 +1,7 @@
 package mesosphere.marathon.core.launcher
 
-import org.apache.mesos.Protos.{ OfferID, TaskInfo }
+import mesosphere.marathon.core.matcher.base.OfferMatcher.TaskOp
+import org.apache.mesos.Protos.OfferID
 
 /**
   * A TaskLauncher launches tasks on an offer or declines an offer.
@@ -12,7 +13,7 @@ trait TaskLauncher {
     *
     * @return `true` if we could communicate the task launch to Mesos and `false` otherwise
     */
-  def launchTasks(offerID: OfferID, taskInfos: Seq[TaskInfo]): Boolean
+  def acceptOffer(offerID: OfferID, taskOps: Seq[TaskOp]): Boolean
 
   /**
     * Decline the offer. We cannot use the offer afterwards anymore.
--- a/src/main/scala/mesosphere/marathon/core/launcher/TaskLauncher.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/TaskLauncher.scala
@@ -8,7 +8,7 @@ import org.apache.mesos.Protos.OfferID
   */
 trait TaskLauncher {
   /**
-    * Launch the given tasks on the given offer. The offer is consumed afterwards and
+    * Send operations on given offer to Mesos. The offer is consumed afterwards and
     * cannot be used anymore.
     *
     * @return `true` if we could communicate the task launch to Mesos and `false` otherwise
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLauncherImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLauncherImpl.scala
@@ -5,8 +5,10 @@ import java.util.Collections
 import mesosphere.marathon.MarathonSchedulerDriverHolder
 import mesosphere.marathon.core.base.Clock
 import mesosphere.marathon.core.launcher.TaskLauncher
+import mesosphere.marathon.core.matcher.base.OfferMatcher
+import mesosphere.marathon.core.matcher.base.OfferMatcher.TaskOp
 import mesosphere.marathon.metrics.{ MetricPrefixes, Metrics }
-import org.apache.mesos.Protos.{ OfferID, Status, TaskInfo }
+import org.apache.mesos.Protos.{ Offer, OfferID, Status }
 import org.apache.mesos.{ Protos, SchedulerDriver }
 import org.slf4j.LoggerFactory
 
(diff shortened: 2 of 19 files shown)


kolloch:
Fixed the mesos-simulation in an additional commit.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLauncherImpl.scala
```diff
@@ -21,17 +23,19 @@ private[launcher] class TaskLauncherImpl(
...
       }
-      driver.launchTasks(Collections.singleton(offerID), taskInfos.asJava)
+
+      val operations = taskOps.flatMap(_.offerOperations)
+      driver.acceptOffers(Collections.singleton(offerID), operations.asJava, Protos.Filters.newBuilder().build())
     }
     if (launched) {
       usedOffersMeter.mark()
-      launchedTasksMeter.mark(taskInfos.size.toLong)
+      taskOps.collect { case OfferMatcher.Launch(_, _, _) => launchedTasksMeter.mark(taskOps.size.toLong) }
```
meichstedt: Did you mean `collectFirst`? Afaict we want to call ` launchedTasksMeter.mark(taskOps.size.toLong)` only once, not for every `Launch` operation.

src/main/scala/mesosphere/marathon/core/launcher/TaskLauncher.scala
```diff
@@ -12,7 +13,7 @@ trait TaskLauncher {
     *
     * @return `true` if we could communicate the task launch to Mesos and `false` otherwise
     */
-  def launchTasks(offerID: OfferID, taskInfos: Seq[TaskInfo]): Boolean
+  def acceptOffer(offerID: OfferID, taskOps: Seq[TaskOp]): Boolean
```
meichstedt: I'd change the scaladoc to something like _Schedule Launch operations … _

meichstedt: Can you change the scaladoc to something like _Schedule Launch operations … _?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -205,33 +206,30 @@ private class AppTaskLauncherActor(
...
 
-    case TaskLaunchSourceDelegate.TaskLaunchRejected(
-      taskInfo, AppTaskLauncherActor.TASK_LAUNCH_REJECTED_TIMEOUT_REASON
+    case TaskOpSourceDelegate.TaskLaunchRejected(
+      op, AppTaskLauncherActor.TASK_LAUNCH_REJECTED_TIMEOUT_REASON
       ) =>
 
-      log.debug("Unnecessary timeout message. Ignoring task launch rejected for task id '{}'.",
-        taskInfo.getTaskId.getValue)
+      log.debug("Unnecessary timeout message. Ignoring task launch rejected for task id '{}'.", op.taskId)
```
meichstedt: Curiosity: Does the log message say that the log message is unnecessary or how should this be understood? Is this a _should not happen_ case?

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLauncherImpl.scala
kolloch: Thx. This is wrong.

We want to mark once for every launch.

src/main/scala/mesosphere/marathon/core/launcher/TaskLauncher.scala
kolloch: Jepp.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -336,7 +334,7 @@ private class AppTaskLauncherActor(
         case Some(CreatedTask(mesosTask, marathonTask)) =>
           def updateActorState(): Unit = {
             tasksMap += marathonTask.getId -> marathonTask
-            inFlightTaskLaunches += mesosTask.getTaskId -> None
+            inFlightTaskLaunches += mesosTask.getTaskId.getValue -> None
```
meichstedt: Not sure if I'm nit-picking or if this is on purpose: `marathonTask.getId` is the same as `mesosTask.getTaskId.getValue` – to make it obvious that both keys are the same, I'd rather write
```
tasksMap += marathonTask.getId -> marathonTask
inFlightTaskLaunches += marathonTask.getId -> None
```

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
kolloch: This is a the message send via a scheduled timeout. If we receive launch confirmation, we cancel the timer. But there is a small race condition such that we might get the message after canceling the timer. In this case, we simply ignore it.

src/main/scala/mesosphere/marathon/core/matcher/base/OfferMatcher.scala
```diff
@@ -1,15 +1,51 @@
...
+    def offerOperations: Seq[Offer.Operation]
+  }
+
+  /** Launch a task on the offer. */
+  case class Launch(taskInfo: TaskInfo, newTask: MarathonTask, oldTask: Option[MarathonTask] = None) extends TaskOp {
+    def applyToOffer(offer: Offer): Offer = {
+      import scala.collection.JavaConverters._
+      ResourceUtil.consumeResourcesFromOffer(offer, taskInfo.getResourcesList.asScala)
+    }
+    def offerOperations: Seq[Offer.Operation] = Seq(OfferOperation.launch(taskInfo))
```
meichstedt: Can't this be an Iterable?

src/main/scala/mesosphere/marathon/core/matcher/base/OfferMatcher.scala
```diff
@@ -17,25 +53,35 @@ object OfferMatcher {
     * could not match the offer in any way it should simply leave the tasks
     * collection empty.
     *
-    * To increase fairness between matchers, each normal matcher should only launch as
-    * few tasks as possible per offer -- usually one. Multiple tasks could be used
-    * if the tasks need to be colocated. The OfferMultiplexer tries to summarize suitable
+    * To increase fairness between matchers, each normal matcher should only execute as few operations
```
meichstedt: II would prefer the term _scheduling_ in this context, since these Operations do not necessarily get executed. This affects the scaladocs in the whole `OfferMatcher. If you agree, please update the docs to say `schedule` (or something else that does not imply any execution/processing).

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
kolloch: ok, I do something similar

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Ok, got it

src/main/scala/mesosphere/marathon/core/matcher/base/OfferMatcher.scala
kolloch: Ack

meichstedt:
Apart from the comments you already addressed only minor ones – please address as well; otherwise looks good to me!

src/main/scala/mesosphere/marathon/core/matcher/base/OfferMatcher.scala
kolloch: I used a mix of "send to Mesos" and scheduling -- depending on the context.

kolloch:
I addressed your comments.

meichstedt:
Yay :shipit:

PR #3117, opened 2016-01-28
Type annotations to make compilation in Intellij work

kolloch:
Without these, my Intellij issues a compilation warning and fails the build.

I think they are acceptable...


gkleiman:
:+1:

PR #3119, opened 2016-01-28
Hack to make it possible to run integration tests

kolloch:
on my machine.

Should not do harm otherwise since all other directories are checked first.


aquamatthias:
sounds promising... :)

kolloch:
Cool, heh?

aquamatthias:
Terrific!

PR #3166, opened 2016-02-05
Ports documentation, fixes #2236.

### LINKED ISSUE #2236: Ports documentation
The ports documentation is nonsensical: https://mesosphere.github.io/marathon/docs/rest-api.html and https://mesosphere.github.io/marathon/docs/service-discovery-load-balancing.html

> This port is used to ensure no two applications can be run using Marathon with overlapping port assignments (i.e. that the ports are globally unique).

I can run apps on different hosts with the same port, so what does "globally unique" mean here?

> However, this is not the port that each instance of that application is assigned when running on slave nodes within the Mesos cluster.

Then what is it?

> Since multiple instances may run on the same node, each instance (or task) is assigned a random port.

Isn't this only if I specify "0".  What if I specify multiple "0"s?  Do I get multiple random ports?  How do I access them?

The list goes on...

ssk2:
Add new doc on ports. Clean up service discovery page.

### DIFF
--- a/docs/docs/ports.md
+++ b/docs/docs/ports.md
@@ -1,0 +1,236 @@
+---
+title: Ports
+---
+
+# Ports
+
+Port configuration for applications in Marathon can be confusing and there is [an outstanding issue](https://github.com/mesosphere/marathon/issues/670) to redesign the ports API. This page attempts to explain more clearly how they work.
+
+## Definitions
+
+*containerPort*: A _container port_ specifies a port within a container. This is only necessary as part of a _port mapping_ when using `BRIDGE` mode networking with a Docker container.
+
+*hostPort*: A _host port_ specifies a port on the host to bind to. When used with `BRIDGE` mode networking, you specify a _port mapping_ from a _host port_ to a _container port_. In `HOST` networking, requested ports are _host ports_ by default. Note that only _host ports_ are made available to a task through environment variables.
+
+*BRIDGE networking*: `BRIDGE` mode networking is used by Docker applications that specify `BRIDGE` mode networking. In this mode, container ports (a port within the container) are mapped to host ports (a port on the host machine). In this mode, applications bind to the specified ports within the container and Docker networking binds to the specified ports on the host.
+
+*HOST networking*: `HOST` networking is used by non-Docker Marathon applications and Docker applications that use `HOST` mode networking. In this mode, applications bind directly to one or more ports on the host machine.
+
+*portMapping*: A _port mapping_ is necessary for Docker applications that use `BRIDGE` mode networking and is a tuple containing a host port, container port, service port and protocol. Multiple _port mappings_ may be specified for a Marathon application.
+
+*ports*: The _ports_ array is used to define ports that should be considered as part of a resource offer. It is necessary only to define this array if you are using `HOST` networking and no port mappings are specified.
+
+*protocol*: _Protocol_ specifies the internet protocol to use for a port (e.g. `tcp` or `udp`). This is only necessary as part of a _port mapping_ when using `BRIDGE` mode networking with a Docker container.
+
+*requirePorts*: _requirePorts_ is a property that specifies whether Marathon should specifically look for specified ports in the resource offers it receives. This ensures that these ports are free and available to be bound to on the Mesos agent. This does not apply to `BRIDGE` mode networking.
+
+*servicePort*: A _service port_ is a port used to describe the port that a service should made available at. Marathon does not bind to the service ports specified but will ensure that you cannot have multiple applications that use the same service port running on the same host. Service ports are typically only used by external applications (e.g. HAProxy) to make the application available at the specified port. See [Service Discovery & Load Balancing](service-discovery-load-balancing.html) for more information.
+
+## Random Port Assignment
+
+Using the value 0 for any port settings indicates to Marathon that you would like a random port assignment. However, if `containerPort` is set to 0 within a `portMapping`, it is set to the same value as `hostPort`.
+
+## Environment Variables
+
+Each *host port* value is exposed to the running application instance via environment variables `$PORT0`, `$PORT1`, etc. Each Marathon application is given a single port by default, so `$PORT0` is always available. These variables are available inside a Docker container being run by Marathon too.
+
+When using `BRIDGE` mode networking, be sure to bind your application to the `containerPort`s you have specified in your `portMapping`s. However, if you have set `containerPort` to 0 then this will be the same as `hostPort` and you can use the `$PORT` environment variables.
+
+## Example Configuration
+
+### Host Mode
+
+Host mode networking is the default networking mode for Docker containers and the only networking mode for non-Docker applications. Note that it not necessary to `EXPOSE` ports in your Dockerfile.
+
+#### Enabling Host Mode
+
+Host mode is enabled by default for containers. If you wish to be explicit, you can also specify it manually through the `network` property:
+
+```json
+  "container": {
+    "type": "DOCKER",
+    "docker": {
+      "image": "my-image:1.0",
+      "network": "HOST"
+    }
+  },
+```
+
+For non-Docker applications, you don't need to specify anything.
+
+#### Specifying Ports
+
+You can specify the ports that are available through the `ports` array:
+
+```json
+    "ports": [
+        0, 0, 0
+    ],
+```
+
+In this example, we specify three randomly assigned host ports which would then be available to our command via the environment variables `$PORT0`, `$PORT1` and `$PORT2`. Marathon will also randomly assign three service posts in addition to these three host ports.
+
+You can also specify specific service ports:
+
+```json
+    "ports": [
+        2001, 2002, 3000
+    ],
+```
+
+In this case, host ports `$PORT0`, `$PORT1` and `$PORT3` remain randomly assigned. However, the three service ports for this application are now `2001`, `2002` and `3000`. 
+
+In this example, as with the previous one, it is necessary to use a service discovery solution such as HAProxy to proxy requests from service ports to host ports.
+
+If you want the applications service ports to be equal to its host ports, you can set `requirePorts` to `true` (`requirePorts` is `false` by default). This will tell Marathon to only schedule this application on agents which have these ports available:
+
+```json
+    "ports": [
+        2001, 2002, 3000
+    ],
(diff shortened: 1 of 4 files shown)


docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+
+# Ports
+
+Port configuration for applications in Marathon can be confusing and there is [an outstanding issue](https://github.com/mesosphere/marathon/issues/670) to redesign the ports API. This page attempts to explain more clearly how they work.
+
+## Definitions
+
+*containerPort*: A _container port_ specifies a port within a container. This is only necessary as part of a _port mapping_ when using `BRIDGE` mode networking with a Docker container.
+
+*hostPort*: A _host port_ specifies a port on the host to bind to. This is only necessary as part of a _port mapping_ when using `BRIDGE` mode networking with a Docker container.
```
kolloch: "normal" non-docker ports are also really host ports. One of the important things is that this is how you other tasks can reach this task.

The host port is also then made available in the environment variables (and not the container port).

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+
+*servicePort*: A _service port_ is a port used to describe the port that a service should made available at. Marathon does not bind to the service ports specified but will ensure that you cannot have multiple applications that use the same service port running on the same host. Service ports are typically only used by external applications (e.g. HAProxy) to make the application available at the specified port. See [Service Discovery & Load Balancing](service-discovery-load-balancing.html) for more information.
+
+## Random Port Assignment
+
+Using the value 0 for any port settings indicates to Marathon that you would like a random port assignment. However, if containerPort is set to 0 within a portMapping, it is set to the same value as hostPort.
+
+## Environment Variables
+
+Each port value is exposed to the running application instance via environment variables `$PORT0`, `$PORT1`, etc. Each Marathon application is given a single port by default, so `$PORT0` is always available. These variables are available inside a Docker container being run by Marathon too.
```
kolloch: Explain that these are the host ports and therefore the ports on which you reach the task from outside of the container on the host ip in bridge networking. Therefore, you should NOT use these port environment variables to bind your application to when using bridge networking.

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Example Configuration
+
+### Host Mode
+
+Host mode networking is the default networking mode for Docker containers and the only networking mode for non-Docker applications. Note that it not necessary to `EXPOSE` ports in your Dockerfile.
+
+#### Enabling Host Mode
+
+Host mode is enabled by default for containers. If you wish to be explicit, you can also specify it manually through the `network` property:
```
kolloch: I didn't know that ;)

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+
+You can specify the ports that are available through the `ports` array:
+
+```json
+    "ports": [
+        0, 0, 0
+    ],
+```
+
+In this example, we specify three randomly assigned ports which would then be available to our command via the environment variables `$PORT0`, `$PORT1` and `$PORT2`.
```
kolloch: I would call them host ports to be clear.

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+
+These ports also become service ports for your application and are reserved cluster wide.
```
kolloch: Incorrect and therefore very confusing.

The ports that are assigned to your task as host ports are different from the service ports. (except with requirePorts where every last bit of sense of service ports falls apart).

With `requirePorts = false` (default) Marathon will only look at the port count to assign the correct number host ports to the task. It will assign unique service ports, though (at least unique for this Marathon instance) and return them via the API in this exact field. This is the horrible design decision that keeps on confusing scores of Marathon users.

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+The service ports and the environment variables`$PORT0`, `$PORT1`, and `$PORT2` are both now `2001`, `2002` and `3000` respectively.
+
+This property is useful if you don't use a service discovery solution such as HAProxy to proxy requests on service ports.
+
+#### Referencing Ports
+
+You can reference ports in the Dockerfile for our fictitious app as follows:
+
+```sh
+CMD ./my-app --http-port=$PORT0 --https-port=$PORT1 --monitoring-port=$PORT2
```
kolloch: I know that this is in the `HOST` networking section but I would mention that this only works for `HOST` networking here again. This is a constant source of confusion.

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+      "portMappings": [
+        { "containerPort": 80, "hostPort": 0 },
+        { "containerPort": 443, "hostPort": 0 },
+        { "containerPort": 4000, "hostPort": 0 }
+      ]
+    }
+  },
+```
+
+In this case, Marathon will randomly allocate host ports and map these to ports `80`, `443` and `4000` respectively. It's important to note that the `$PORT` variables refer to the host ports. In this case, `$PORT0` will be set to the value of `hostPort` for the first mapping and so on.
```
kolloch: Jepp!

docs/docs/ports.md
```diff
@@ -0,0 +1,234 @@
...
+        { "containerPort": 443, "hostPort": 0, "protocol": "tcp", "servicePort": 2001 },
+        { "containerPort": 4000, "hostPort": 0, "protocol": "udp", "servicePort": 3000}
+      ]
+    }
+  },
+```
+
+In this example, the host ports `$PORT0`, `$PORT1` and `$PORT3` remain randomly assigned. However, the service ports for this application are now `2001`, `2002` and `3000`. An external proxy, like HAProxy, should be configured to route from the service ports to the host ports.
+
+#### Referencing Ports
```
kolloch: Really nice section!

docs/docs/service-discovery-load-balancing.md
```diff
@@ -4,51 +4,31 @@ title: Service Discovery & Load Balancing
...
+There are several mechanisms available to accomplish this:
+* [Mesos-DNS](https://github.com/mesosphere/mesos-dns) provides service discovery through the domain name system ([DNS](http://en.wikipedia.org/wiki/Domain_Name_System))
+* [Marathon-lb](https://github.com/mesosphere/marathon-lb) provides port based service discovery using HAProxy, a lightweight TCP/HTTP proxy
+* [haproxy-marathon-bridge](https://github.com/mesosphere/marathon/blob/master/examples/haproxy-marathon-bridge) is an example script which configures a local HAProxy installation
+
+For a detailed description of how ports work in Marathon, see [Ports](ports.html).
+
+## Mesos-DNS
+
+Mesos-DNS generates a hostname for each Mesos task (including Marathon application instances) and translates these names to the IP address and port on the machine currently running each application. Mesos-DNS is particularly useful if the apps are launched through multiple frameworks (not just Marathon). Check the Mesos-DNS [documentation and tutorials page](http://mesosphere.github.io/mesos-dns/) for further information.
```
kolloch: It also creates a `SRV` record with which you cannot only lookup the IP but also the port. The latter is necessary if you use random host port assignments. Maybe you could point out that Mesos DNS is especially nice if you use one IP per task and make that routable. (We don't support that well, though)

docs/docs/service-discovery-load-balancing.md
```diff
@@ -4,51 +4,31 @@ title: Service Discovery & Load Balancing
...
+
+## Marathon-lb
+
+An alternative way to implement service discovery is to run a TCP/HTTP proxy on each host in the cluster, and transparently forward connections to the static service port on localhost to the dynamically assigned host/port combinations of the individual Marathon application instances (running Mesos *tasks*). That way, clients simply connect to the well known defined service port, and do not need to know the implementation details of discovery.This approach is sufficient if all apps are launched through Marathon.
+
+Marathon-lb is a Dockerized application that includes both HAProxy an application that uses Marathon's REST API to regenerate the HAProxy configuration. It supports more advanced functionality like SSL offloading, sticky connections and VHost based load balancing, allowing you to specify virtual hosts for your Marathon applications.
+
+See the [Marathon-lb repository](https://github.com/mesosphere/marathon-lb) for more information or check out [the tutorial on the Mesosphere blog](https://mesosphere.com/blog/2015/12/04/dcos-marathon-lb/).
```
kolloch: Maybe note that if you use this approach, you should not use `requirePorts=true`

docs/docs/service-discovery-load-balancing.md
```diff
@@ -4,51 +4,31 @@ title: Service Discovery & Load Balancing
...
+
+## haproxy-marathon-bridge
+
+Marathon ships with a simple shell script called `haproxy-marathon-bridge` which uses Marathon's REST API to create a config file for HAProxy. The `haproxy-marathon-bridge` provides a minimum set of functionality and is easier to understand for beginners or as a good starting point for a custom implementation.
```
kolloch: It's deprecated.

kolloch:
Great to see this effort! Thanks a lot @ssk2!

docs/docs/ports.md
ssk2: I'm confused :D. 

With `requirePorts = false` and three values of 0, you get (in this example) 3 random host ports and 3 random service ports? 

With `requirePorts = false` and three set values (e.g. 2001, 2002 and 3000), you get 3 random host ports and 3 service ports set to 2001, 2002 and 3000?

With `requirePorts = true` and three set values (e.g. 2001, 2003 and 3000), you get 3 host ports  and 3 services ports all set to 2001, 2003 and 3000?

ssk2:
@kolloch - updated! One outstanding question from me.

docs/docs/ports.md
kolloch: For some reason, I don't see this comment on github. But you are right!

It is beautiful, isn't it?

kolloch:
Very nice work! Since you need to rebase anyways, I'd like you to make the deprecation of the haproxy script more prominent.

ssk2:
Thanks for the review @kolloch. Comments addressed and rebased.

kolloch:
Thanks for your service to the Marathon community :) :rocket:

kolloch:
Do you publish the current master to ghpages for us? @ssk2

ssk2:
@kolloch sure, will do now

PR #3168, opened 2016-02-05
Read environment variables from /etc/default/marathon file.

aquamatthias:
This allows docker images, that extend the Marathon docker image to pass env variables.


gkleiman:
:ship: :rocket: :train: :tractor:

Issue #3041, opened 2016-01-20
Resident Tasks: Extend MarathonTask state


meichstedt:
* Add information on pending reservations volumes.
* Decompose the `MarathonTask` protocol buffers object into a better scala-style abstraction.

Checklist to verify that this is finished:

Not done as part of this issue, now part of #3311:

  - [ ] check remaining `MarathonTask` occurrences (persistence related only)
  - [ ] `TaskFailures` (`Task.Id` instead of Mesos `TaskID`)
  - [ ] Check for other usages of Mesos `TaskID`, should only be used for communication with Mesos
  - [ ] `TaskSerializer` is package private
  - [ ] `MarathonTestHelper.*taskProto` methods are gone  

Done:

  - [x] check remaining `Task.Id` conversions (only towards the communication boundaries, e.g. towards Mesos)
  - [x] check remaining `Task.Id.idString` calls (only towards the communication boundaries, e.g. towards Mesos)
  - [x] `MarathonTasks` gone (all methods moved into `Task` and its sub structures)
  - [x] `TaskIDUtil` gone
  - [x] `Task.marathonTask` gone

kolloch:
Done. Left-overs tracked in other issues.

PR #3171, opened 2016-02-05
Towards #3041 - Representing all state relating to a task in TaskState

kolloch:
and eventually get rid of MarathonTask.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
@@ -1,6 +1,8 @@
 package mesosphere.marathon.core.task.tracker
 
 import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.tracker.impl.TaskSerializer
 import mesosphere.marathon.state.PathId
 import org.slf4j.LoggerFactory
 
@@ -22,8 +24,10 @@ trait TaskTracker {
   def appTasksSync(appId: PathId): Iterable[MarathonTask]
   def appTasks(appId: PathId)(implicit ec: ExecutionContext): Future[Iterable[MarathonTask]]
 
-  def taskSync(appId: PathId, taskId: String): Option[MarathonTask]
-  def task(appId: PathId, taskId: String)(implicit ec: ExecutionContext): Future[Option[MarathonTask]]
+  def marathonTaskSync(appId: PathId, taskId: String): Option[MarathonTask]
+  def marathonTask(appId: PathId, taskId: String)(implicit ec: ExecutionContext): Future[Option[MarathonTask]]
+
+  def task(appId: PathId, taskId: Task.Id)(implicit ec: ExecutionContext): Future[Option[Task]]
 
   def tasksByAppSync: TaskTracker.TasksByApp
   def tasksByApp()(implicit ec: ExecutionContext): Future[TaskTracker.TasksByApp]
@@ -50,10 +54,15 @@ object TaskTracker {
       appTasksMap.get(appId).map(_.tasks).getOrElse(Iterable.empty)
     }
 
-    def task(appId: PathId, taskId: String): Option[MarathonTask] = for {
+    def marathonTask(appId: PathId, taskId: String): Option[MarathonTask] = for {
       app <- appTasksMap.get(appId)
       task <- app.taskMap.get(taskId)
     } yield task
+
+    def taskState(appId: PathId, taskId: Task.Id): Option[Task] = for {
+      app <- appTasksMap.get(appId)
+      taskState <- app.taskStateMap.get(taskId.id)
+    } yield taskState
 
     def allTasks: Iterable[MarathonTask] = appTasksMap.values.view.flatMap(_.tasks)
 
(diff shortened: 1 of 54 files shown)


src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
```diff
@@ -39,6 +42,7 @@ object TaskTracker {
   /**
     * Contains all tasks grouped by app ID.
     */
+  // FIXME: Rename to TaskStatesByApp
```
aquamatthias: Is that a fixme you want to address in this commit?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateSerializer.scala
```diff
@@ -0,0 +1,118 @@
...
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.core.task.tracker.TaskState
+import mesosphere.marathon.state.Timestamp
+import org.apache.mesos.{ Protos => MesosProtos }
+
+/**
+  * Converts between [[TaskState]] objects
+  * and their serialized representation: MarathonTask.
+  */
+object TaskStateSerializer {
```
aquamatthias: Would be nice to have a type class approach like Play Json Reader/Writer for all our proto stuff.

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
kolloch: No, because that would cause havoc for reviews.

Let's stabilize the concepts in this PR and then bundle the renames in a separate PR. That was a pattern for review happiness in the past.

src/main/scala/mesosphere/marathon/core/task/tracker/TaskState.scala
```diff
@@ -0,0 +1,86 @@
...
+  * The state for launching a task. This might be a launched task or a reservation for launching a task or both.
+  */
+case class TaskState(
+    taskId: TaskState.TaskId,
+    agentInfo: TaskState.AgentInfo,
+    reservationWithVolume: Option[TaskState.ReservationWithVolume.type],
+    launchCounter: Long,
+    launchedTask: Option[TaskState.LaunchedTask]) {
+
+  lazy val appId = TaskIdUtil.appId(taskId.id)
```
aquamatthias: could also be part of TaskId -> taskId.appId
I know there are a couple of places that need this.

src/test/scala/mesosphere/marathon/SchedulerActionsTest.scala
```diff
@@ -60,35 +60,11 @@ class SchedulerActionsTest extends MarathonActorSupport with MarathonSpec with M
...
-      .setId("task_3")
-      .setSlaveId(SlaveID("slave 1"))
-      .build()
-
-    val stagedWithSlaveIdStatus = TaskStatus.newBuilder
-      .setTaskId(TaskID.newBuilder.setValue(stagedTaskWithSlaveId.getId))
-      .setSlaveId(stagedTaskWithSlaveId.getSlaveId)
-      .setState(TaskState.TASK_STAGING)
-      .build()
+    val runningTask = MarathonTestHelper.runningTask("task_1")
```
aquamatthias: Clean!

src/test/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseDataTest.scala
```diff
@@ -115,16 +111,8 @@ class AppInfoBaseDataTest extends MarathonSpec with GivenWhenThen with Mockito w
...
-      .setId("task1")
-      .setStatus(Protos.TaskStatus.newBuilder().setState(Protos.TaskState.TASK_STAGING).buildPartial())
-      .buildPartial()
-    val running = MarathonTask
-      .newBuilder()
-      .setId("task2")
-      .setStatus(Protos.TaskStatus.newBuilder().setState(Protos.TaskState.TASK_RUNNING).buildPartial())
-      .buildPartial()
+    val staged = MarathonTestHelper.stagedTask("task1")
+    val running = MarathonTestHelper.runningTask("task2")
```
aquamatthias: Thanks for removing the mess!

src/main/scala/mesosphere/marathon/core/task/tracker/TaskState.scala
```diff
@@ -0,0 +1,86 @@
...
+    * * it saves storage space
+    * * we already have the infrastructure to look for a specific task ID, we do not have
+    *   to build a second one to look things up for a volume ID.
+    */
+  case object ReservationWithVolume
+
+  /**
+    * Represents a task which has been launched (i.e. sent to Mesos for launching).
+    */
+  case class LaunchedTask(
```
aquamatthias: Why do you nest the types inside the object?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskState.scala
```diff
@@ -0,0 +1,86 @@
+package mesosphere.marathon.core.task.tracker
```
aquamatthias: Can we move classes like TaskState to `mesosphere.marathon.core.task`?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskState.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+    * Legacy conversion to MarathonTask. Cache result to speed up repeated uses.
+    * Should be removed before releasing 0.16.
+    */
+  lazy val marathonTask: MarathonTask = TaskStateSerializer.marathonTask(this)
+}
+
+object TaskState {
+  case class TaskId(id: String) {
```
aquamatthias: Really appreciate, that we have a type for this.
Can we move it outside of TaskState? or rename to Id so it becomes Task.Id?

aquamatthias:
Thanks. No comments about the logic - just naming proposals.

PR #3173, opened 2016-02-05
Fixes #3140 - Also default to HealthCheck#portIndex=0 when an app is updated

### LINKED ISSUE #3140: breaking API change on portIndex?
We recently upgraded from Marathon 0.11 to 0.14.1 and noticed that all our deployments were failing. Looking at the logs, we found this:

```
[Unhealthy(service.1d56c262-c9ca-11e5-a454-0242c88d2ac8,2016-02-02T16:00:05.537Z,Missing/invalid port index and no explicit port specified,2016-02-02T16:30:29.359Z)] (mesosphere.marathon.health.HealthCheckActor:marathon-akka.actor.default-dispatcher-7)
```

We did not specify the port index. To my understanding, it used to default to 0. Once we passed it explicitly, deployments went through just fine.

I scanned through the release notes but couldn't find any mentioning of a breaking API change in this regards. The only thing I could find was [a section](https://github.com/mesosphere/marathon/blob/master/changelog.md#health-checks-allow-port-instead-of-portindex-1) describing the additional `port` field. I understand this as supposedly non-breaking, however. The description of [this commit](https://github.com/mesosphere/marathon/commit/38ce1b3f1be6ad99379fa9f859bb8df59451c1ac) from last November also indicates that backwards compatibility should be maintained.

Was this a deliberate API modification, and did I miss a note somehow? Or rather an inadvertent change?

Side note: The [deprecated REST API documentation](https://mesosphere.github.io/marathon/docs/rest-api.html) says that `portIndex` defaults to zero. I can't find any description of the field in the [new documentation](https://mesosphere.github.io/marathon/docs/generated/api.html) though. Where did this piece of specification go?

Thanks.

gkleiman:


aquamatthias:
LGTM.

PR #3175, opened 2016-02-05
Resolves #3039 API changes for Resident Tasks

### LINKED ISSUE #3039: Allow configuration of Resident Tasks
subtasks
- [x] API changes
- [x] Model changes
- [x] Validation

meichstedt:
Resolves #3039 by allowing to create AppDefinitions that define persistent volumes, allow container type `MESOS` and enable defining residency information. *Note*: the added models are not yet handled anywhere, this is just the API adjustment part. Changes in particular:

- updated Protobuf to allow for persistent volumes and residency information
- moved serialization logic to dedicated serializers
- added json formats
- updated AppDefinition json schema
- added validation
- added unit test and integration tests

### DIFF
--- a/src/main/proto/marathon.proto
+++ b/src/main/proto/marathon.proto
@@ -82,6 +82,7 @@ message ServiceDefinition {
   optional int64 last_scaling_at = 23;
   optional int64 last_config_change_at = 24;
   optional IpAddress ipAddress = 25;
+  optional ResidencyDefinition residency = 26;
 }
 
 message ResourceRoles {
@@ -114,7 +115,7 @@ message ContainerInfo {
   repeated bytes options = 2;
 }
 
-// Copied from mesos.proto
+// Copied from mesos.proto and adjusted to meet the needs for persistent volumes
 // Extends DockerInfo.PortMapping to include `service_port`.
 message ExtendedContainerInfo {
 
@@ -148,9 +149,42 @@ message ExtendedContainerInfo {
   }
 
   required mesos.ContainerInfo.Type type = 1;
-  repeated mesos.Volume volumes = 2;
+  repeated Volume volumes = 2;
 
   optional DockerInfo docker = 3;
+}
+
+// Copied from mesos.proto and adjusted to meet the needs of persistent volumes
+// Describes a volume mapping either from host to container or vice
+// versa. Both paths can either refer to a directory or a file.
+message Volume {
+
+  // Defining properties of persistent volumes
+  message PersistentVolumeInfo {
+    required uint64 size = 1;
+  }
+
+  required mesos.Volume.Mode mode = 3;
+
+  // Path pointing to a directory or file in the container. If the
+  // path is a relative path, it is relative to the container work
+  // directory. If the path is an absolute path, that path must
+  // already exist.
+  required string container_path = 1;
+
+  // The following specifies the source of this volume. At most one of
+  // the following should be set.
+
+  // Absolute path pointing to a directory or file on the host or a
+  // path relative to the container work directory.
+  optional string host_path = 2;
+
+  // The source of the volume is an Image which describes a root
+  // filesystem which will be provisioned by Mesos.
+  optional mesos.Image image = 4;
+
+  // Defines a Persistent Volume. In this case hostPath must not be set
+  optional PersistentVolumeInfo persistent = 5;
 }
 
 message EventSubscribers {
@@ -208,4 +242,14 @@ message ZKStoreEntry {
   // Introduced in Marathon 0.12
   // Optional to be backward compatible
   optional bool compressed = 4 [default = false];
+}
+
+message ResidencyDefinition {
+  enum TaskLostBehavior {
+    RELAUNCH_AFTER_TIMEOUT = 0;
+    WAIT_FOREVER = 1;
+  }
+
+  optional int64 relaunchEscalationTimeoutSeconds = 1;
+  optional TaskLostBehavior taskLostBehavior = 2;
 }
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -2,13 +2,14 @@ package mesosphere.marathon.api.v2.json
 
 import mesosphere.marathon.Protos.Constraint.Operator
 import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
+import mesosphere.marathon.Protos.ResidencyDefinition.TaskLostBehavior
 import mesosphere.marathon.Protos.{ Constraint, MarathonTask }
 import mesosphere.marathon.core.appinfo._
 import mesosphere.marathon.event._
 import mesosphere.marathon.event.http.EventSubscribers
 import mesosphere.marathon.health.{ Health, HealthCheck }
 import mesosphere.marathon.state.Container.Docker.PortMapping
-import mesosphere.marathon.state.Container.{ Docker, Volume }
+import mesosphere.marathon.state.Container.Docker
 import mesosphere.marathon.state._
 import mesosphere.marathon.tasks.MarathonTasks
 import mesosphere.marathon.upgrade._
@@ -209,13 +210,16 @@ trait ContainerFormats {
   )(Docker(_, _, _, _, _, _), unlift(Docker.unapply))
 
   implicit lazy val ModeFormat: Format[mesos.Volume.Mode] =
-    enumFormat(mesos.Volume.Mode.valueOf, str => s"$str is not a valid mode")
+    enumFormat(mesos.Volume.Mode.valueOf, str => s"$str is not a valid mde")
+
+  implicit lazy val PersistentVolumeInfoFormat: Format[PersistentVolumeInfo] = Json.format[PersistentVolumeInfo]
 
   implicit lazy val VolumeFormat: Format[Volume] = (
     (__ \ "containerPath").format[String] ~
-    (__ \ "hostPath").format[String] ~
-    (__ \ "mode").format[mesos.Volume.Mode]
-  )(Volume(_, _, _), unlift(Volume.unapply))
+    (__ \ "hostPath").formatNullable[String] ~
+    (__ \ "mode").format[mesos.Volume.Mode] ~
+    (__ \ "persistent").formatNullable[PersistentVolumeInfo]
+  )(Volume(_, _, _, _), unlift(Volume.unapply))
 
   implicit lazy val ContainerTypeFormat: Format[mesos.ContainerInfo.Type] =
     enumFormat(mesos.ContainerInfo.Type.valueOf, str => s"$str is not a valid container type")
@@ -535,7 +539,8 @@ trait AppAndGroupFormats {
           labels: Map[String, String],
           acceptedResourceRoles: Option[Set[String]],
           ipAddress: Option[IpAddress],
-          version: Timestamp)
+          version: Timestamp,
+          residency: Option[Residency])
 
         val extraReads: Reads[ExtraFields] =
           (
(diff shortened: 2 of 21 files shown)


src/main/proto/marathon.proto
```diff
@@ -148,11 +149,42 @@ message ExtendedContainerInfo {
...
+
+  // The source of the volume is an Image which describes a root
+  // filesystem which will be provisioned by Mesos.
+  optional mesos.Image image = 4;
+
+  // Defines a Persistent Volume. In this case hostPath must not be set
+  optional PersistentVolumeInfo persistent = 5;
+}
+
+message PersistentVolumeInfo {
```
aquamatthias: Can you define that message inside Volume?
So you have all Volume related type definitions together.

src/main/proto/marathon.proto
```diff
@@ -148,11 +149,42 @@ message ExtendedContainerInfo {
   }
 
   required mesos.ContainerInfo.Type type = 1;
-  repeated mesos.Volume volumes = 2;
+  repeated Volume volumes = 2;
```
aquamatthias: Have you tested to read volume information from older Marathon versions?
Do we need a migration? (Not sure how protobuf handles different types)

src/main/scala/mesosphere/marathon/state/Residency.scala
```diff
@@ -0,0 +1,16 @@
+package mesosphere.marathon.state
+
+import com.wix.accord.dsl._
+import mesosphere.marathon.Protos.ResidencyDefinition.TaskLostBehavior
+
+case class Residency(relaunchEscalationTimeoutSeconds: Long, taskLostBehavior: TaskLostBehavior)
```
aquamatthias: Minor: specifying a relaunchEscalationTimeoutSeconds only makes sense for a certain behavior.

src/main/proto/marathon.proto
meichstedt: Makes sense, will do.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -586,6 +594,31 @@ trait V2Formats {
     })
   }
 
+  implicit lazy val taskLostBehaviorWrites = Writes[TaskLostBehavior] { taskLostBehavior =>
+    JsString(taskLostBehavior.name())
+  }
+  private[this] lazy val allowedTaskLostBehaviorString =
```
aquamatthias: can you move it into the None case (only usage)?

src/main/scala/mesosphere/marathon/api/v2/serialization/ContainerSerializer.scala
```diff
@@ -0,0 +1,165 @@
...
+
+    builder.setImage(docker.image)
+
+    docker.network foreach builder.setNetwork
+
+    docker.portMappings.foreach { pms =>
+      builder.addAllPortMappings(pms.map(PortMappingSerializer.toProto).asJava)
+    }
+
+    builder.setPrivileged(docker.privileged)
```
aquamatthias: why not user 
```scala
xxx.newBuilder
    .setXXX
    .setYYY
```
?

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -203,4 +184,33 @@ object Container {
 
   }
 
+  // We need validation based on the container type, but don't have dedicated classes. Therefore this approach manually
+  // delegates validation to the matching validator
+  implicit val containerValidator: Validator[Container] = {
+    val volumeValidator: Validator[Volume] = new Validator[Volume] {
```
aquamatthias: Move to Container companion?

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -203,4 +184,33 @@ object Container {
...
+      override def apply(volume: Volume): Result = volume match {
+        case pv: PersistentVolume => valid[PersistentVolume](PersistentVolume.persistentVolumeValidator).apply(pv)
+        case dv: DockerVolume     => valid[DockerVolume](DockerVolume.dockerVolumeValidator).apply(dv)
+      }
+    }
+
+    val dockerContainerValidator: Validator[Container] = validator[Container] { container =>
+      container.docker is notEmpty
```
aquamatthias: Add `container.docker is valid` ?

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -172,23 +170,6 @@ object Container {
         protocol: String = "tcp") {
 
       require(protocol == "tcp" || protocol == "udp", "protocol can only be 'tcp' or 'udp'")
```
aquamatthias: Question: do we want to use require or validator?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -89,13 +88,96 @@ class AppsResourceTest extends MarathonSpec with Matchers with Mockito with Give
...
+        |}
+      """.stripMargin.getBytes("UTF-8")
+    groupManager.updateApp(any, any, any, any, any) returns Future.successful(plan)
+
+    When("The request is processed")
+    val response = appsResource.create(body, false, auth.request, auth.response)
+
+    Then("The return code indicates that the hostPath of volumes[0] is missing") // although the wrong field should fail
+    response.getStatus should be(422)
+    response.getEntity.toString should include("container.volumes.[0]hostPath")
```
aquamatthias: Should that be `container.volumes[0].hostPath`?

src/main/scala/mesosphere/marathon/api/v2/serialization/ContainerSerializer.scala
```diff
@@ -0,0 +1,165 @@
+package mesosphere.marathon.api.v2.serialization
```
aquamatthias: Why is the proto serialization in api.v2 ?

src/test/scala/mesosphere/marathon/api/validation/AppUpdateValidatorTest.scala
```diff
@@ -0,0 +1,28 @@
...
+import mesosphere.marathon.MarathonSpec
+import mesosphere.marathon.api.v2.Validation._
+import mesosphere.marathon.api.v2.json.AppUpdate
+import mesosphere.marathon.state.{ Container, PathId }
+import org.apache.mesos.{ Protos => mesos }
+import org.scalatest.Matchers
+
+class AppUpdateValidatorTest extends MarathonSpec with Matchers {
+
+  class Fixture {
```
aquamatthias: Minor: can we move the fixture after the tests?

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -1,143 +1,141 @@
...
+    persistent: PersistentVolumeInfo,
+    mode: Mesos.Volume.Mode)
+      extends Volume
+
+  object PersistentVolume {
+    import org.apache.mesos.Protos.Volume.Mode
+    implicit val persistentVolumeValidator = validator[PersistentVolume] { vol =>
+      vol.containerPath is notEmpty
+      vol.persistent is valid
+      vol.mode is equalTo(Mode.RW)
```
aquamatthias: I think this is valid. Just to make sure: do we have to enforce this?

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -1,143 +1,141 @@
...
-      containerPath: String,
-      hostPath: String,
-      mode: mesos.Volume.Mode) {
-    def toProto(): mesos.Volume =
-      mesos.Volume.newBuilder
-        .setContainerPath(containerPath)
-        .setHostPath(hostPath)
-        .setMode(mode)
-        .build
+  sealed trait Volume {
```
aquamatthias: Can we move the complete Volume trait into a separate file?

aquamatthias:
Generally looks good - please see my comments.

src/main/scala/mesosphere/marathon/state/Residency.scala
meichstedt: The actual idea behind as discussed was to use `relaunchEscalationTimeoutSeconds` for all terminal tasks (FAILED, FINISHED, etc) and to have the `taskLostBehavior` define the behavior for lost tasks.
That means that
- if a task fails, Marathon will wait for an offer matching its reservation labels and volumes for the amount of time defined by `relaunchEscalationTimeoutSeconds`. When the timeout is reached, it will relaunch the task on another agent.
- if a task is lost, Marathon will either do nothing (`WAIT_FOREVER`), or apply the same logic as above (although it most probably won't get an offer because the node is gone).

It might make sense to immediately relaunch elsewhere upon `TASK_LOST`, and to apply a `taskTerminationBehavior` instead to all kinds of terminal task states, respectively. All based on assumption though …

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/api/v2/serialization/ContainerSerializer.scala
meichstedt: Did that where it's cheap – setNetwork doesn't allow null so I can't chain that call, and for portMappings it'd be a little ugly because I'd have to fold over the Option and pass an empty list if there aren't any portMappings – easier to read is it is imho.

src/main/scala/mesosphere/marathon/state/Container.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/state/Container.scala
meichstedt: I added a validator but kept the require … there are various usages of require that we might remove, but I'll leave that for a separate PR.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
meichstedt: Ack. Yet I would have to dig into accord implementation to fix this, and I really don't want to do this right now.

src/test/scala/mesosphere/marathon/api/validation/AppUpdateValidatorTest.scala
meichstedt: moved.

src/main/scala/mesosphere/marathon/state/Container.scala
meichstedt: Mesos Persistent Volumes only allows `RW` – other volume types allow `RO` as well, so I guess it makes sense to enforce this here (we can't validate this constraint via the json schema). Plus, I want Alice to receive a validation error if she consciously tries to set `RO`.

src/main/scala/mesosphere/marathon/state/Container.scala
meichstedt: moved.

src/main/scala/mesosphere/marathon/api/v2/serialization/ContainerSerializer.scala
meichstedt: Moved to api. We should consider moving the models from v2 up to api as well. I don't want to incorporate that change into this PR, though.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
aquamatthias: @alexanderweber: ^^^ do you know why?

src/main/proto/marathon.proto
meichstedt: I have tested with this app definition
```
{
  "id": "docker-with-vol",
  "instances": 1,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "thomasr/sleep"
    },
    "volumes": [{
      "containerPath": "/my-mount",
      "hostPath": "/tmp",
      "mode": "RW"
    }]
  }
}
```

Created in 0.15.1, still working in 0.16.0-SNAPSHOT and can be scaled/restarted.

meichstedt:
addressed your comments in a separate commit. If that's ok, I'll squash.

aquamatthias:
Thanks! LGTM. Please squash and I will merge.

meichstedt:
squashed!

aquamatthias:
TUUUUUUUUUUUUUUUUUUUUT
```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__            
         __|_o_o_o\__         
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^   
  
```

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
alexanderweber: @aquamatthias the issue is resolved in #3209. Sorry for the inconvenience.

PR #3177, opened 2016-02-05
Haproxy marathon bridge deprecated

sascala:
Removed the discussion of haproxy-marathon-bridge. Pointed users to the marathon-lb docs.


aquamatthias:
Thanks @sascala!

PR #3187, opened 2016-02-08
Corrected example of max_tasks_per_offer default.

joerg84:


aquamatthias:
The docs are more accurate...
It will become the default, if #3190 is merged.

PR #3188, opened 2016-02-08
Towards #3041 - Use new "Task" in more places instead of "MarathonTask"

kolloch:
* Get rid of appId in some place where it is redundant
* Use Task.Id in Map structures inside the TaskTracker and the AppTaskLaunchActor

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -9,6 +9,7 @@ import mesosphere.marathon.MarathonSchedulerActor.ScaleApp
 import mesosphere.marathon.api.LeaderInfo
 import mesosphere.marathon.api.v2.json.AppUpdate
 import mesosphere.marathon.core.launchqueue.LaunchQueue
+import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.tracker.TaskTracker
 import mesosphere.marathon.event.{ AppTerminatedEvent, DeploymentFailed, DeploymentSuccess, LocalLeadershipEvent }
 import mesosphere.marathon.health.HealthCheckManager
@@ -160,7 +161,7 @@ class MarathonSchedulerActor private (
       val origSender = sender()
       withLockFor(appId) {
         val promise = Promise[Unit]()
-        val tasksToKill = taskIds.flatMap(taskId => taskTracker.marathonTaskSync(appId, taskId))
+        val tasksToKill = taskIds.flatMap(taskId => taskTracker.marathonTaskSync(Task.Id(taskId)))
         context.actorOf(Props(classOf[TaskKillActor], driver, appId, taskTracker, eventBus, tasksToKill, promise))
         val res = for {
           _ <- promise.future
@@ -436,7 +437,7 @@ class SchedulerActions(
     healthCheckManager.removeAllFor(app.id)
 
     log.info(s"Stopping app ${app.id}")
-    val tasks = taskTracker.appTasksSync(app.id)
+    val tasks = taskTracker.marathonAppTasksSync(app.id)
 
     for (task <- tasks) {
       log.info(s"Killing task ${task.getId}")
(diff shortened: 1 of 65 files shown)


kolloch:
Added some renames in a second commit.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -160,7 +161,7 @@ class MarathonSchedulerActor private (
       val origSender = sender()
       withLockFor(appId) {
         val promise = Promise[Unit]()
-        val tasksToKill = taskIds.flatMap(taskId => taskTracker.marathonTaskSync(appId, taskId))
+        val tasksToKill = taskIds.flatMap(taskId => taskTracker.marathonTaskSync(Task.Id(taskId)))
```
aquamatthias: Is there a reason why KillTasks has taskids as string?
`MarathonSchedulerService.killTasks` convert from taskId->String and here back again...

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
```diff
@@ -50,21 +53,25 @@ object TaskTracker {
...
     def hasAppTasks(appId: PathId): Boolean = appTasksMap.contains(appId)
 
-    def appTasks(appId: PathId): Iterable[MarathonTask] = {
+    def appTasks(appId: PathId): Iterable[Task] = {
       appTasksMap.get(appId).map(_.tasks).getOrElse(Iterable.empty)
     }
 
-    def marathonTask(appId: PathId, taskId: String): Option[MarathonTask] = for {
-      app <- appTasksMap.get(appId)
+    def marathonAppTasks(appId: PathId): Iterable[MarathonTask] = {
```
aquamatthias: Just to get the idea: we want to delete this method at some point in the future?

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -122,11 +123,11 @@ class AppInfoBaseData(
 
     lazy val enrichedTasksFuture: Future[Seq[EnrichedTask]] = {
       def statusesToEnrichedTasks(
-        tasksById: Map[String, MarathonTask],
+        tasksById: Map[Task.Id, MarathonTask],
         statuses: Map[String, collection.Seq[Health]]): Seq[EnrichedTask] = {
         for {
           (taskId, healthResults) <- statuses.to[Seq]
-          task <- tasksById.get(taskId)
+          task <- tasksById.get(Task.Id(taskId))
```
aquamatthias: The HealthCheckManager could give us a TaskId instead of a String.
If not done in this PR: can we somehow mark those occurrences? (It will be hard to identify all places afterwards)

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -59,11 +81,15 @@ object Task {
   /**
     * Contains information about the status of a launched task including timestamps for important
     * state transitions.
+    *
+    * @param stagedAt Despite its name, stagedAt is set on task creation and before the TASK_STAGED notification from
```
aquamatthias: Can we rename it then?

aquamatthias:
It looks half way trough - I can merge if we agree on the goal.

aquamatthias:
See my comments - and please squash

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
kolloch: Well, I drew the line somewhere. The idea is to change the usage of Strings for `Task.Id`s everywhere, eventually.

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
kolloch: Why is it harder to identify afterwards? How do you want them marked?

src/main/scala/mesosphere/marathon/core/task/Task.scala
kolloch: If using a fresh approach, I would call it `launchedAt`.

If we rename it now, it might be confusing because in the API it is called `stagedAt`. What do you think?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskTracker.scala
kolloch: exactly. We want to get rid of all references to `MarathonTask` anywhere except for storing `Task` via the `TaskRepository` as `MarathonTask`.

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
kolloch: Generally my strategy towards the end of this work will be:

* Search for `MarathonTask`. Make sure it's only left as an implementation detail for storing `Task`s.
* Search for `Task.Id` conversions. Check whether we can get rid of them by converting some data structure from `String` to `Task.Id`.

To come up with incremental changes, I cannot do it all at once.

kolloch:
Squashed.

PR #3190, opened 2016-02-09
Update OfferMatchingManager.scala to safe max-tasks-per-offer setting

jgarcia-mesosphere:
The [docs list](https://mesosphere.github.io/marathon/docs/command-line-flags.html) `--max-tasks-per-offer` as defaulted to 1, which is a safe setting. We default this to 100 in the code and 1000 in DCOS; our tests indicate that anything higher than 5 is unsafe for AWS. Adjusting this to a safe setting so less risk-averse users can adjust on their own.

> --max_tasks_per_offer (Optional. Default: 1): Launch at most this number of tasks per Mesos offer. Usually, there is one offer per cycle and slave. You can speed up launching tasks by increasing this number.

We need to get this important safety feature out ASAP.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ start.sh
 .DS_Store
 *.ipr
 *.iws
+.metadata/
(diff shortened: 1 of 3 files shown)


aquamatthias:
@jgarcia-mesosphere: I agree. You also have to adapt tests.

aquamatthias:
Handled in #3229

PR #3193, opened 2016-02-09
Fixes #3169 by validating the resource values of the AppDefinition.

### LINKED ISSUE #3169: Possible to start app with negative resources
Where resources are expressed as numbers, it's possible to send negative numbers and have your app configuration treated as valid. 

Example configuration:
```
{
  "id": "invalid-application",
  "cmd": "sleep 10000",
  "memory": -128,
  "ports": [-8080]
}
```

We should check that they are >= 0.

aquamatthias:


src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -116,6 +116,11 @@ object Validation {
     }
   }
 
+  val isPositive: Validator[Int] = new Validator[Int] {
+    override def apply(v1: Int): Result =
+      if (v1 >= 0) Success else Failure(Set(RuleViolation(v1, "isPositive", None)))
```
jimenez: Should the message here be "Element must be positive" instead of "isPositive"?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -502,8 +502,8 @@ trait AppAndGroupFormats {
       (__ \ "args").readNullable[Seq[String]] ~
       (__ \ "user").readNullable[String] ~
       (__ \ "env").readNullable[Map[String, String]].withDefault(AppDefinition.DefaultEnv) ~
-      (__ \ "instances").readNullable[Int](minValue(0)).withDefault(AppDefinition.DefaultInstances) ~
-      (__ \ "cpus").readNullable[Double](greaterThan(0.0)).withDefault(AppDefinition.DefaultCpus) ~
+      (__ \ "instances").readNullable[Int].withDefault(AppDefinition.DefaultInstances) ~
```
jimenez: This validations should also be removed line 739 and 740

jimenez: Also the greaterThan function is not used anymore

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -470,12 +470,16 @@ object AppDefinition {
...
-    appDef.ports is elementsAreUnique(filterOutRandomPorts)
+    appDef.ports is elementsAreUnique(filterOutRandomPorts) and every(isPositive)
     appDef.executor should matchRegexFully("^(//cmd)|(/?[^/]+(/[^/]+)*)|$")
     appDef is containsCmdArgsContainerValidator
     appDef is portIndicesAreValid
     appDef.instances.intValue should be >= 0
     appDef.fetch is every(fetchUriIsValid)
+    appDef.mem should be >= 0.0
+    appDef.cpus should be >= 0.0
+    appDef.instances should be >= 0
```
jimenez: I think this is already checked line 477

aquamatthias:
@jimenez thanks for the comments. Hope I addressed them.

jimenez:
LGTM :+1:

gkleiman:
:rocket:

PR #3195, opened 2016-02-09
update mesosphere.io to mesosphere.com

lloesche:
This PR should update all relevant references to downloads.mesosphere.io and repos.mesosphere.io to use the .com domain instead.

(Note that the S3 bucket name hasn't changed)


meichstedt:
LGTM, waiting for tests

PR #3196, opened 2016-02-09
Towards #3041 - Use new Task for counts and some other things

kolloch:


src/main/scala/mesosphere/marathon/core/appinfo/TaskLifeTime.scala
```diff
@@ -14,7 +14,7 @@ case class TaskLifeTime(
   medianSeconds: Double)
 
 object TaskLifeTime {
-  def forSomeTasks(now: Timestamp, tasks: Iterable[MarathonTask]): Option[TaskLifeTime] = {
+  def forSomeTasks(now: Timestamp, tasks: Iterable[Task]): Option[TaskLifeTime] = {
```
aquamatthias: if there is a `forSomeTasks`, will there be a `forALotOfTasks`?

kolloch: Not part of this PR but I admit your comment is funny ;)

aquamatthias: My suggestion would be to rename it: `forTasks`

aquamatthias:
LGTM

PR #3201, opened 2016-02-09
Towards #3041 - better Tasks in health

kolloch:


src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -138,6 +139,11 @@ trait Formats
     Writes[PathId] { id => JsString(id.toString) }
   )
 
+  implicit lazy val TaskIdFormat: Format[Task.Id] = Format(
+    Reads.of[String](Reads.minLength[String](3)).map(Task.Id(_)),
```
aquamatthias: why 3? since we use uuid's it should be at least 36?

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -26,7 +26,9 @@ class TaskReplaceActor(
 
   val tasksToKill = taskTracker.marathonAppTasksSync(app.id)
   val appId = app.id
-  val version = app.version.toString
+  val version = app.version
+  // FIXME: Remove!
```
aquamatthias: Please mark as FIXME(PK) and create a ticket for you to remove all occurrences of FIXMEs.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -3,7 +3,7 @@ package mesosphere.marathon.core.task
 import mesosphere.marathon.Protos.MarathonTask
 import mesosphere.marathon.core.task.Task.Launched
 import mesosphere.marathon.core.task.tracker.impl.TaskSerializer
-import mesosphere.marathon.state.{ PathId, Timestamp }
+import mesosphere.marathon.state.{ AppDefinition, PathId, Timestamp }
 import mesosphere.marathon.tasks.TaskIdUtil
```
aquamatthias: Do we still need TaskIdUtil?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -30,6 +30,16 @@ case class Task(
   def withLaunched(update: Launched => Launched): Task =
     copy(launched = launched.map(update))
 
+  def ipAddresses: Iterable[MesosProtos.NetworkInfo.IPAddress] = launched.map(_.ipAddresses).getOrElse(Iterable.empty)
```
aquamatthias: Thanks for moving that functionality here.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
kolloch: Admittedly, a fairly random choice. Basically, we did not validate task IDs before. Do we also want such a high length limit during testing? If the ID doesn't match an existing task, it is mood anyways.

 I think we do not really need to verify it here but my reasoning was that there needed to be at least an app prefix, a dot, and a uuid.

I can use `36` if it doesn't break any tests.

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
kolloch: I shouldn't have conflated this PR with changing Timestamp strings to Timestamps. I created an issue https://github.com/mesosphere/marathon/issues/3207 and I will remove the FIXME. (Otherwise I'll have to create a lot of FIXMEs ;) )

The issue contains specific instructions on how to find these occurrences. Prioritize as you like.

src/main/scala/mesosphere/marathon/core/task/Task.scala
kolloch:
```scala
/**
  * Utility functions for dealing with TaskIDs
  */
// FIXME(PK): Remove, merge into Task.Id
class TaskIdUtil {
```

src/main/scala/mesosphere/marathon/core/task/Task.scala
kolloch: :)

kolloch:
Thanks for the comments! I added a commit with `TaskTest` which tests the added methods -- I forgot that. `MarathonTasks.*` and `MarathonTasksTest` still exist since the code still has usages.

Since there was some doubt what constitutes to finishing #3041, I added my private checklist to this issue. Check it out and tell me if you agree.

Thanks a lot.

src/main/scala/mesosphere/marathon/core/task/Task.scala
aquamatthias: :+1:

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
aquamatthias: Is there something like nonEmpty? otherwise I don't care.

aquamatthias:
Thanks: Please squash and I will merge.

kolloch:
Thanks! Squashed.

aquamatthias:
Thanks!

PR #3210, opened 2016-02-10
Migrate upgrade package to new Task and get rid of TaskIDUtil

kolloch:


src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -470,11 +469,7 @@ class SchedulerActions(
     appRepository.allPathIds().map(_.toSet).flatMap { appIds =>
       taskTracker.tasksByApp().map { tasksByApp =>
         val knownTaskStatuses = appIds.flatMap { appId =>
-          tasksByApp.marathonAppTasks(appId).collect {
-            case task: Protos.MarathonTask if task.hasStatus => task.getStatus
-            case task: Protos.MarathonTask => // staged tasks, which have no status yet
-              taskStatus(task)
-          }
+          tasksByApp.appTasks(appId).flatMap(_.mesosStatus)
```
aquamatthias: Do we handle the task staging case elsewhere?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -25,6 +26,32 @@ case class Task(
     */
   lazy val marathonTask: MarathonTask = TaskSerializer.marathonTask(this)
 
+  lazy val launchedMesosId: Option[MesosProtos.TaskID] = launched.map { _ =>
+    // it doesn't make sense for an unlaunched task
+
+    var id = taskId.idString
+    if (launchCounter > 0) {
```
aquamatthias: Kind of implicit - launchCounter>0 == resident task?
Perhaps create a separate def and use it to make it more explicit?
```scala 
def isResident = launchCounter>0
````

File: src/main/scala/mesosphere/marathon/core/task/Task.scala [hunk 3]
```diff
@@ -43,16 +70,37 @@ case class Task(
 }
 
 object Task {
+  def tasksById(tasks: Iterable[Task]): Map[Task.Id, Task] = tasks.iterator.map(task => task.taskId -> task).toMap
+
   case class Id(idString: String) {
+    // FIXME(PK): For dynamic reservations, the mesosTaskId has to take the launchCounter into
```
aquamatthias: Confusing...
Task.Id can be either the taskId (simple task) or the static part of a reserved task?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -43,16 +70,37 @@ case class Task(
...
+
+    def appId(taskId: String): PathId = {
+      taskId match {
+        case TaskIdRegex(appId, uuid) => PathId.fromSafePath(appId)
+        case _                        => throw new MatchError(s"taskId $taskId is no valid identifier")
+      }
+    }
+
+    // FIXME(PK): If we get this from mesos, we also need to extract the launch counter.
+    // Probably better to introduce a launch ID which can be translated into a TaskId.
```
aquamatthias: Then we would have a type and have to think about that.

aquamatthias:
Some comments and one open question - let's have a short discussion.
Otherwise looks good.
Squash commits.

src/main/scala/mesosphere/marathon/core/task/Task.scala
kolloch: Will remove launchCounter...

src/main/scala/mesosphere/marathon/core/task/Task.scala
kolloch: Pending removal.

src/main/scala/mesosphere/marathon/core/task/Task.scala
kolloch: Pending removal.

meichstedt:
LGTM – various unused `MarathonTask` imports in the tests – see `mesosphere.marathon.upgrade` package

meichstedt:
Thanks! waiting for tests …

PR #3211, opened 2016-02-10
Fixes #3209 - fixes misplacement of dot in validation messages.

### LINKED ISSUE #3209: Validation point position for containers is placed wrong according to #3175
Validation for container objects yields a validation error message, which contains a misplaced point as shown here:
`"container.volumes.[0]hostPath"`
This issue was discovered in #3175

alexanderweber:
Fixes an issue of point misplacement as described in #3209

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -52,23 +52,26 @@ object Validation {
   }
 
   private def allRuleViolationsWithFullDescription(violation: Violation,
-                                                   parentDesc: Option[String] = None,
-                                                   attachDot: Boolean = true): Set[RuleViolation] = {
+                                                   parentDesc: Option[String] = None): Set[RuleViolation] = {
     def concatPath(parent: String, child: Option[String], attachDot: Boolean): String = {
-      child.map(c => parent + { if (attachDot) "." else "" } + c).getOrElse(parent)
+      child.map(c => parent + c + { if (attachDot) "." else "" }).getOrElse(parent)
     }
 
     violation match {
       case r: RuleViolation => Set(parentDesc.map(p =>
-        r.withDescription(concatPath(p, r.description, attachDot)))
+        r.withDescription(concatPath(p, r.description, attachDot = false)))
         .getOrElse(r))
       case g: GroupViolation => g.children.flatMap { c =>
-        parentDesc.map { p =>
-          val desc = concatPath(p, g.description, attachDot)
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -52,22 +52,28 @@ object Validation {
...
       case r: RuleViolation => Set(parentDesc.map(p =>
-        r.withDescription(concatPath(p, r.description, attachDot)))
+        r.withDescription(concatPath(p, r.description, attachDot = false)))
         .getOrElse(r))
       case g: GroupViolation => g.children.flatMap { c =>
         parentDesc.map { p =>
-          val desc = concatPath(p, g.description, attachDot)
-          allRuleViolationsWithFullDescription(c, Some(desc), g.description.isDefined)
+          val forceNoDot = g.value match {
+            case x :: xs => false
```
aquamatthias: Here you pattern match for a non empty list. Perhaps you want to do something like: 
```scala
case _: Iterable => false
```

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -52,22 +52,28 @@ object Validation {
...
+            case _       => true
+          }
+          val desc = concatPath(p, g.description, forceNoDot)
+          allRuleViolationsWithFullDescription(c, Some(desc))
         } getOrElse {
-          allRuleViolationsWithFullDescription(c, g.description, g.description.isDefined)
+          val desc = if (g.description.isDefined)
```
aquamatthias: `g.description` is already n option. You could use it like:
```scala
g.description.map( concatPath("", _, attachDot = true))
```

aquamatthias:
Thanks! 
Small improvement suggestions - otherwise fine.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
alexanderweber: You are right, I did change that to the suggested line.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
alexanderweber: Missed that one, thanks for pointing that out!

alexanderweber:
@aquamatthias your suggestions made it into a new commit. Tests are good, looking forward to merging.

aquamatthias:
Thanks! :+1:

PR #3217, opened 2016-02-11
Fixes #3216 by setting the default offer timeout to 0 while accepting an offer.

### LINKED ISSUE #3216: When accepting an Offer, no offer timeout should be applied.
When accepting an offer, the rest of the offer is declined automatically with a filter.
This filter should have no timeout, so the same offer is send in the next allocator cycle.

aquamatthias:


aquamatthias:
Cheery Pick for 0.15

kolloch:
:+1: :rocket: 

waiting for tests

PR #3226, opened 2016-02-11
Fix naming irregularities in TaskCreationHandlerAndUpdaterDelegateTest

meichstedt:
Test description and val naming was wrong and misleading.


kolloch:
a wonderful PR! :100:

PR #3227, opened 2016-02-11
Resolves #3200 by adding an integration test for persistent volumes

### LINKED ISSUE #3200: Resident Tasks: Write Integration Tests

meichstedt:
Note that the test is ignored currently and we will need some adjustments for mesos-local (framework auth/role registration) after the reserve/create functionality is added in order to let this test succeed.

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
@@ -1,0 +1,105 @@
+package mesosphere.marathon.integration
+
+import mesosphere.marathon.api.v2.json.AppUpdate
+import mesosphere.marathon.integration.setup.MarathonFacade._
+import mesosphere.marathon.integration.setup.{ SingleMarathonIntegrationTest, IntegrationFunSuite }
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
```diff
@@ -0,0 +1,105 @@
...
+    create(f.appWithPersistentVolume(
+      appId,
+      containerPath,
+      cmd = s"""echo "data" > $containerPath/data """))
+
+    // wait until the task exited
+    waitForStatusUpdates(TASK_RUNNING, TASK_FINISHED)
+
+    // suspend the app
+    suspend(appId)
```
aquamatthias: Don't know if this is the most simple test case.
My initial idea would be to use a simple kill.
But your proposal solves the initial create and exists logic.

aquamatthias: Is there any other way to check the existence of a file (from the mesos side)?

aquamatthias:
LGTM - Thanks!

PR #3249, opened 2016-02-16
Update docs version to 0.15.2

philipnrmn:
This raises the version number on the downloads page at mesosphere.github.io/marathon.


gkleiman:
Thanks!

PR #3255, opened 2016-02-17
Fixes #3241 - consistent JSON error output.

### LINKED ISSUE #3241: Serverside validation messages are inconsistent
There seem to be two kinds of error response objects when validation errors occur in the API:

Case 1) Generic errors
Case 2) Field-specific errors

**Case 1** errors trigger a `422` response with this structure:
```
{
  "message": "Object is not valid",
  "details": [
    {
      "attribute": "value",
      "error": "Groups and Applications may not have the same identifier: /hello"
    }
  ]
}
```
(side-note: `value` seems to be a placeholder for non-specific attribute paths)

**Case 2** errors trigger a `400` response with this structure:
```
{
  "message": "Invalid JSON",
  "details": [
    {
      "path": "/cpus",
      "errors": [
        "error.expected.jsnumber"
      ]
    }
  ]
}
```

I suggest the following result for consistency:

**Case 1)** `422`
```
{
  "message": "Object is not valid",
  "details": [
    {
      "path": "general",
      "errors": [
        "Groups and Applications may not have the same identifier: /hello"
      ]
    }
  ]
}
```

**Case 2)** `400`
Unchanged.

Please correct if I'm wrong and looking forward to hearing your thoughts on this.

alexanderweber:
Fixes #3241 applying following changes to JSON output of errors:

1. `attribute` field renamed to `path`
2. `error` field renamed to `errors`
3. `errors` field contains a list of errors caused by an attribute, which location can be found in `path`
4. root `path` information is no longer labeled as `value`, but as `self`
5. `.value` should never terminate a path (e.g. `apps[0].value` is replaced by `apps[0]`)
6. `"path": "self"` if there is no description about the point of failure at all

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
@@ -1,7 +1,11 @@
 package mesosphere.marathon.api.v2
 
+import java.util.regex.Pattern
+
+import com.wix.accord.{ Success, Failure }
+import com.wix.accord.dsl._
 import mesosphere.marathon.MarathonSpec
-import mesosphere.marathon.api.v2.json.{ GroupUpdate, GroupUpdate$ }
+import mesosphere.marathon.api.v2.json.GroupUpdate
 import mesosphere.marathon.state.Container.Docker.PortMapping
 import mesosphere.marathon.state.Container._
 import mesosphere.marathon.state.PathId._
@@ -10,8 +14,10 @@ import org.apache.mesos.Protos.ContainerInfo.DockerInfo.Network
 import org.scalatest.{ BeforeAndAfterAll, Matchers, OptionValues }
 
 import mesosphere.marathon.api.v2.Validation._
+import play.api.libs.json.{ JsObject, Json }
 
 import scala.collection.immutable.Seq
+import scala.util.matching.Regex
 
 class ModelValidationTest
     extends MarathonSpec
@@ -74,6 +80,36 @@ class ModelValidationTest
       v.message == "Requested service port 3200 conflicts with a service port in app /app2") should be(true)
   }
 
+  test("Multiple errors within one field of a validator should be grouped into one array") {
+    val empty = ImportantTitle("")
+
+    validate(empty) match {
+      case Success => fail()
+      case f: Failure =>
+        val errors = (Json.toJson(f) \ "details").as[Seq[JsObject]]
+        errors should have size 1
+        (errors.head \ "path").as[String] should be("name")
+        (errors.head \ "errors").as[Seq[String]] should have size 2
+    }
+  }
+
+  test("Validators should not produce 'value' string at the end of description.") {
+    val group = Group("/test".toPath, groups = Set(
+      Group("/test/group1".toPath, Set(
+        AppDefinition("/test/group1/valid".toPath, cmd = Some("foo")),
+        AppDefinition("/test/group1/invalid".toPath))
+      ),
+      Group("/test/group2".toPath)))
+
+    validate(group) match {
+      case Success => fail()
+      case f: Failure =>
+        val errors = (Json.toJson(f) \ "details").as[Seq[JsObject]]
+        errors should have size 1
+        (errors.head \ "path").as[String] should be("groups[0].apps[1]")
+    }
+  }
+
   private def createServicePortApp(id: PathId, servicePort: Int) =
     AppDefinition(
       id,
@@ -86,4 +122,10 @@ class ModelValidationTest
       ))
     )
 
+  case class ImportantTitle(name: String)
+
+  implicit val mrImportantValidator = validator[ImportantTitle] { m =>
+    m.name is equalTo("Dr.")
+    m.name is notEmpty
+  }
 }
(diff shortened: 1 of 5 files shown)


alexanderweber:
@aquamatthias I did apply the changes we've been talking about in #3241. You can find the list of changes in the description of this PR. Glad to receive any feedback!
cc/ @pierlo-upitup

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
```diff
@@ -74,6 +80,34 @@ class ModelValidationTest
...
 
+  test("Multiple errors within one field of a validator should be grouped into one array") {
+    val empty = ImportantTitle("")
+
+    validate(empty) match {
+      case Success => fail()
+      case f: Failure =>
+        val errors = (Json.toJson(f) \ "details").as[Seq[JsObject]]
+        errors should have size 1
+        (errors.head \ "path").as[String] should be("name")
```
aquamatthias: If we see self as top level object. Shouldn't it be `self.name` here?

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
```diff
@@ -74,6 +80,34 @@ class ModelValidationTest
...
+      AppDefinition("/test/group1/valid".toPath, cmd = Some("foo")),
+      AppDefinition("/test/group1/invalid".toPath)
+    ))
+
+    validate(group) match {
+      case Success => fail()
+      case f: Failure =>
+        val errors = (Json.toJson(f) \ "details").as[Seq[JsObject]]
+        errors should have size 1
+        (errors.head \ "path").as[String] should be("apps[1]")
```
aquamatthias: Question: should the path be `self.groups[0].apps[1]` ?

aquamatthias:
@alexanderweber questions about the goal - let's have a short discussion.

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
alexanderweber: The way i got it last time is, that self appears if the error is on the top level only. Meaning if the error is in an attribute of the object, the path yields only the attribute. So self would appear as a standalone information, if the error is not in an attribute, but the object itself. I agree that prepending it in general is a coherent solution, if it does not confuse the user, I am also in for that.

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
alexanderweber: The `self` part is described in the upper comment. Regarding `groups[0]` - there is only one group here, containing two apps. The validation is done on that group object, so there is no list of groups in it.

alexanderweber:
@aquamatthias Sure thing, I left my note on both comments. Basically I am in for prepending `self.` in general in all error messages, as it is describing the error tree most accurately. Let me know if I should do the change.

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
aquamatthias: @pierlo-upitup also interested in your opinion on that.

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
aquamatthias: You are right - I looked into the group path, which is nested, but the object hierarchy is not.
Can you make an example with a more nested problem?

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
pierlo-upitup: As @alexanderweber  wrote
> So self would appear as a standalone information, if the error is not in an attribute, but the object itself.

That's how I got it too. If we have precise information regarding which attribute failed validation, then we don't need `self`. What's described at point 6 makes absolute sense to me:

> "path": "self" if there is no description about the point of failure at all

pierlo-upitup: Also 

https://github.com/mesosphere/marathon/issues/3241#issuecomment-184762112

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
alexanderweber: Sure, I added a depth level to the example above.

alexanderweber:
@aquamatthias I left it the way Pierluigi agreed on it and extended the test in terms of nesting. Let me know if I missed something important out.

aquamatthias:
@alexanderweber Thanks!

aquamatthias:
@pierlo-upitup can you please retest?

pierlo-upitup:
Thanks, opening an issue https://github.com/mesosphere/marathon/issues/3337

pierlo-upitup:
@alexanderweber thanks for this. 

@aquamatthias  question: doesn't this require an update to the docs? 

screen shot 2016-02-29 at 12 37 04

alexanderweber:
Thanks for pointing that out. Docs are adjusted in #3387.

PR #3256, opened 2016-02-17
Fix unrendered examples in docs

janisz:
Docs for [`GET /v2/apps?label={labelSelectorQuery}`](https://mesosphere.github.io/marathon/docs/rest-api.html) are not rendered properrly. Text in `{{ }}` is not rendered into html. Replaced with code formatting.


aquamatthias:
@janisz Thanks!

PR #3279, opened 2016-02-19
Fix markdown emphasis mistakes for notes in task env var doc.

timoreimann:
All "Note:" paragraphs on the Task Environment Variables page start with an underscore for italic emphasis but lack the closing one. One side effect of this is that the `--env_vars_prefix` command line flag is incorrectly displayed as `--envvars_prefix`. (The first underscore would be considered as the closing mark for the italic emphasis.)

This change fixes the problem by using strong emphasis just around the "Note:" label. Note that not doing italics for the entire paragraph makes it easier to read `inline code` parts.


aquamatthias:
Thanks @timoreimann !

PR #3282, opened 2016-02-19
Edit event bus

sascala:


docs/docs/event-bus.md
```diff
@@ -5,7 +5,8 @@ title: Event Bus
 # Event Bus
 
 Marathon has an internal event bus that captures all API requests and scaling events. This is useful for integrating with load balancers, keeping stats, etc.
```
sascala: Could we briefly explain how this is useful? The stats use makes immediate sense, but the load balancer use case could use some explanation.

aquamatthias: By subscribing to the event bus, you get informed about every change instantly (without pulling).  
All entities that act on the state of Marathon (e.g. load balancer) benefit from this functionality.

aquamatthias:
@sascala the event bus has 2 APIs: 

- the callback endpoint: you register a http callback listener. Every event is transferred via a separate HTTP Request to this listener.
- the event stream (see /v2/events)

The latter one is not described at all here (but should).
We like people to use the event stream instead of the callback listener, since it has many advantages:

- it is easier to setup
- the delivery is faster (since no separate connection, request/response cycle)
- the events are transferred in order 

Ultimately we want to deprecate the callback listener.

aquamatthias:
Close for now - please reopen when ready.

PR #3292, opened 2016-02-22
WIP on resident task offer matching and reserve/launch operations

### LINKED ISSUE #3042: Resident Tasks: Offer matching
Match offers with reservations/volumes on queued Resident Tasks

meichstedt:
Goal: resolve #3042 

Changes:
- moved TaskOps into a dedicated object TaskOp
- creation of TaskOps is delegated from AppTaskLauncherActor to taskOpLogic
- adjusted AppTaskLauncherActor to handle TaskOps Launch and ReserveAndCreate
- introduced an OfferOperationFactory – reserve and create both need the framework role and principal and are thus easier to handle as wired class objects rather than singleton objects
- adjusted DefaultTaskOpLigic to infer TaskOps depending on whether the app is resident or not.
- new functions in the TaskBuilder targeting resident apps expect a ResourceMatch/VolumeMatch to be passed in (in order to decide which taskOp to generate, we need to know whether there's a volumeMatch or not – before using tha taskBuilder)

Open points:
- validate the responsibilities of TaskOpLogic/TaskBuilder and related classes
- validate the counting logic inside AppTaskLauncherActor
- write unit/integration tests for resident tasks that verify the behavior on each layer
- role and principal are currently passed along from DefaultTaskOpLogic to TaskOpFactory to OfferOperationFactory. This is ugly but the alternative would be to wire all these via Guice.
- ... surely some more open points


src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -334,44 +335,56 @@ private class AppTaskLauncherActor(
...
-              taskId.mesosTaskId == mesosTask.getTaskId,
-              "marathon task id and mesos task id must be equal"
-            )
-
-            tasksMap += taskId -> marathonTask
-            inFlightTaskLaunches += taskId -> None
-            tasksToLaunch -= 1
-            OfferMatcherRegistration.manageOfferMatcherStatus()
-          }
+      def updateActorState(task: Task, mesosTask: Option[TaskInfo]): Unit = {
```
aquamatthias: Hard to read. Can you move the handling of that case into a separate method?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -334,44 +335,56 @@ private class AppTaskLauncherActor(
...
 
           log.info("Request to launch task with id '{}', version '{}'. {}",
             mesosTask.getTaskId.getValue, app.version, status)
 
-          val launch = OfferMatcher.Launch(mesosTask, marathonTask)
-          self ! AppTaskLauncherActor.ScheduleTaskLaunchNotificationTimeout(launch)
+          self ! AppTaskLauncherActor.ScheduleTaskOpNotificationTimeout(launch)
           sender() ! MatchedTaskOps(offer.getId, Seq(TaskOpWithSource(myselfAsLaunchSource, launch)))
 
+        case Some(reserveAndCreate @ TaskOp.ReserveAndCreateVolumes(newTask, _, _, _, _)) =>
```
aquamatthias: Personal taste: Either match on the type or the content. why mix both?
```scala
case Some(reserveAndCreate : TaskOp.ReserveAndCreateVolumes) =>
```

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -334,44 +335,56 @@ private class AppTaskLauncherActor(
...
+          self ! AppTaskLauncherActor.ScheduleTaskOpNotificationTimeout(reserveAndCreate)
+          sender() ! MatchedTaskOps(offer.getId, Seq(TaskOpWithSource(myselfAsLaunchSource, reserveAndCreate)))
+
         case None => sender() ! MatchedTaskOps(offer.getId, Seq.empty)
       }
   }
 
   protected val receiveScheduleTaskLaunchNotificationDelay: Receive = {
-    case AppTaskLauncherActor.ScheduleTaskLaunchNotificationTimeout(launch) if inFlight(launch) =>
+    case AppTaskLauncherActor.ScheduleTaskOpNotificationTimeout(launch) if inFlight(launch) =>
```
aquamatthias: Do we need to handle reservations and launches differently here?

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  private[this] val log = LoggerFactory.getLogger(getClass)
+
+  // FIXME (217): log.warn in these cases? or only if we really need and access them?
+
+  private[this] lazy val principal: String = principalOpt match {
+    case Some(value) => value
+    case _ =>
+      log.error("No principal set. See --mesos_authentication_principal")
+      "n/a"
```
aquamatthias: might be hard to debug, why "n/a" is not a valid principal. Perhaps we need a special validation

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private[this] val role: String = roleOpt match {
+    case Some(value) => value
+    case _ =>
```
aquamatthias: same as principal

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -99,20 +101,33 @@ object Task {
...
   case class ReservationWithVolumes(volumeIds: Iterable[LocalVolumeId])
 
-  case class LocalVolumeId(idString: String) {
+  case class LocalVolume(id: LocalVolumeId, persistentVolume: PersistentVolume)
+
+  case class LocalVolumeId(appId: PathId, containerPath: String, uuid: String) {
+    import LocalVolumeId._
+    lazy val idString = appId.safePath + appDelimiter + containerPath + volumeDelimiter + uuid
+
     override def toString: String = s"LocalVolume [$idString]"
```
aquamatthias: Suggestion: `def toString = idString`
The compiler does not complain about string coercion.

src/main/scala/mesosphere/marathon/tasks/DefaultTaskOpLogic.scala
```diff
@@ -0,0 +1,162 @@
...
+    val principalOpt = config.mesosAuthenticationPrincipal.get
+    val roleOpt = config.mesosRole.get
+
+    new TaskOpFactory(principalOpt, roleOpt)
+  }
+
+  override def inferTaskOp(app: AppDefinition, offer: Mesos.Offer, tasks: Iterable[Task]): Option[TaskOp] = {
+    log.debug(s"inferTaskOp")
+
+    if (app.residency.isDefined) {
```
aquamatthias: Make a help function in app: `app.isResident`

src/main/scala/mesosphere/mesos/PersistentVolumeMatcher.scala
```diff
@@ -0,0 +1,54 @@
...
+            Some(found.flatten)
+          }
+          else {
+            None
+          }
+      }
+    }
+
+    val satisfiedTasks = for {
+      task <- waitingTasks
```
aquamatthias: Hint: if you use waitingTasks.iterator, you make this computation lazy.

src/main/scala/mesosphere/mesos/PersistentVolumeMatcher.scala
```diff
@@ -0,0 +1,54 @@
...
+      resources <- resourcesForTask(task)
+    } yield (task, resources)
+
+    satisfiedTasks.headOption match {
```
aquamatthias: Use `map` instead of `match`.

src/main/scala/mesosphere/mesos/PersistentVolumeMatcher.scala
```diff
@@ -0,0 +1,54 @@
...
+        resource.getDisk.getPersistence.getId -> resource
+    }.toMap
+
+    def resourcesForTask(task: Task): Option[Iterable[Mesos.Resource]] = {
+      task.reservationWithVolumes match {
+        // in case there are no required volumes, return a match
+        case None => Some(Nil)
+
+        case Some(reservation) =>
+          val found = reservation.volumeIds.map { id =>
```
aquamatthias: use `flatMap` instead of `map`

aquamatthias:
@meichstedt Thanks! - made my comments.
There are a lt of FIXME's - I am not sure how important they are and when they get fixed.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Ack, defined the handling of these 3 cases in separate methods.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Makes sense, now only matching against the type since most of the vals are not of interest.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Could be – the default timeout for a Launch is currently 3 seconds. I'll try to find out whether reserve/create can expected to take longer than that.

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
meichstedt: As discussed, this will now throw if accessed and not defined – since we don't know whether a Marathon user will use this feature, we can only validate the flags when POSTing an AppDefinition and reserving/creating.

src/main/scala/mesosphere/marathon/tasks/DefaultTaskOpLogic.scala
meichstedt: Done.

src/main/scala/mesosphere/mesos/PersistentVolumeMatcher.scala
meichstedt: thanks :+1:

src/main/scala/mesosphere/mesos/PersistentVolumeMatcher.scala
meichstedt: indeed nicer – I also added comments to these 3 cases and another test case that makes sure the correct task/volume match is returned if there are multiple tasks/volumes

meichstedt:
Thanks @aquamatthias – I addressed your comments.
FIXME's are removed and summarized in #3311. These should be prioritized before going after them.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -264,11 +265,23 @@ private class AppTaskLauncherActor(
...
-    inFlightTaskLaunches -= taskId
+    inFlightTaskOperations.get(taskId).foreach(_.foreach(_.cancel()))
+    inFlightTaskOperations -= taskId
     tasksMap -= taskId
   }
 
+  private[this] def unlaunchTask(taskId: Task.Id): Unit = {
+    tasksMap.get(taskId) match {
+      case Some(task) =>
+        log.info(">>> resident task unlaunched")
```
aquamatthias: Left over debug message? Either make a proper message or remove.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -321,8 +335,9 @@ private class AppTaskLauncherActor(
     sender() ! QueuedTaskCount(
       app,
       tasksLeftToLaunch = tasksToLaunch,
-      taskLaunchesInFlight = inFlightTaskLaunches.size,
-      tasksLaunchedOrRunning = tasksMap.size - inFlightTaskLaunches.size,
+      taskLaunchesInFlight = inFlightTaskOperations.size,
+      // don't count tasks that are not launched in the tasksMap
+      tasksLaunchedOrRunning = tasksMap.values.count(_.launched.isDefined) - inFlightTaskOperations.size,
```
aquamatthias: why subtract `inFlightTaskOpertions.size`?
launchedOrRunning ==> all tasks that have a launched option defined - or?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -334,55 +349,65 @@ private class AppTaskLauncherActor(
...
-              taskId.mesosTaskId == mesosTask.getTaskId,
-              "marathon task id and mesos task id must be equal"
-            )
-
-            tasksMap += taskId -> marathonTask
-            inFlightTaskLaunches += taskId -> None
-            tasksToLaunch -= 1
-            OfferMatcherRegistration.manageOfferMatcherStatus()
-          }
+      val taskOp: Option[TaskOp] = taskOpLogic.inferTaskOp(app, offer, tasksMap.values)
```
aquamatthias: Much more readable! :+1:

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Right, removed that and another one!

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -334,55 +349,65 @@ private class AppTaskLauncherActor(
...
+    log.info("Request: Reserve/Create for task '{}', version '{}'. {}", reserve.taskId.idString, app.version, status)
 
-        case None => sender() ! MatchedTaskOps(offer.getId, Seq.empty)
-      }
+    updateActorState(reserve.newTask, None)
+    self ! AppTaskLauncherActor.ScheduleTaskOpNotificationTimeout(reserve)
+    sender() ! MatchedTaskOps(offer.getId, Seq(TaskOpWithSource(myselfAsLaunchSource, reserve)))
+  }
+
+  private[this] def updateActorState(task: Task, mesosTask: Option[Mesos.TaskInfo]): Unit = {
```
aquamatthias: Since this method is now a class level member (context is gone), can we have a better name for this?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -334,55 +349,65 @@ private class AppTaskLauncherActor(
...
-            mesosTask.getTaskId.getValue, app.version, status)
+    updateActorState(launch.newTask, Some(launch.taskInfo))
+    self ! AppTaskLauncherActor.ScheduleTaskOpNotificationTimeout(launch)
+    sender() ! MatchedTaskOps(offer.getId, Seq(TaskOpWithSource(myselfAsLaunchSource, launch)))
+  }
 
-          val launch = OfferMatcher.Launch(mesosTask, marathonTask)
-          self ! AppTaskLauncherActor.ScheduleTaskLaunchNotificationTimeout(launch)
-          sender() ! MatchedTaskOps(offer.getId, Seq(TaskOpWithSource(myselfAsLaunchSource, launch)))
+  private[this] def handleReserveOp(reserve: TaskOp.ReserveAndCreateVolumes, offer: Mesos.Offer): Unit = {
```
aquamatthias: Just noting that handleLaunch and handleReserveOp do exactly the same (other than a different log message). Do you think we can be more DRY here?

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -0,0 +1,94 @@
...
+      .setType(Mesos.Offer.Operation.Type.LAUNCH)
+      .setLaunch(launch)
+      .build()
+  }
+
+  def reserve(resources: Iterable[Mesos.Resource]): Mesos.Offer.Operation = {
+    import scala.collection.JavaConverters._
+    val reservedResources = resources.map { resource =>
+      Mesos.Resource.newBuilder(resource)
+        .setRole(role) // actually needed or already in the resource?
```
aquamatthias: Don't understand the question in the comment. Question for the mesos folks, we should answer before merge?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: The original line before any ResidentTasks changes was
```
tasksLaunchedOrRunning = tasksMap.size - inFlightTaskLaunches.size
```
The scala doc states that `tasksLaunchedOrRunning` is _the number of tasks which are running or at least have been confirmed to be launched_ – so that was hard to understand from the beginning. The way I see it, we have
- `tasksLeftToLaunch` – not yet launched at all
- `taskLaunchesInFlight` – we created a TaskOp but that's not been accepted internally == not yet send to the scheduler via `acceptOffer()`
- `tasksLaunchedOrRunning` – either running or at least already processed via `acceptOffer()` – these are NOT in flight, therefore the inFlightTaskOperations.size is subtracted.

I think the naming here is really misleading and we should either make that crystal clear with good names and/or clarify with tests. Any suggestions?

meichstedt: I think the name `tasksLaunchedOrRunning` is most misleading in that it implies these tasks might be inFlight, which they arent – maybe just rename that to `tasksLaunched` or `tasksRunning`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -99,25 +101,27 @@ object Task {
...
   object LocalVolumeId {
-    def apply(appId: PathId, path: String): LocalVolumeId = {
-      //FIXME: mock implementation from ME
-      LocalVolumeId(s"${appId.safePath}.$path.random")
-    }
-
+    private val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
+    private val appDelimiter = "."
+    private val volumeDelimiter = "."
     private val LocalVolumeEncoderRE = "^([^.]+).([^.]+).([^.]+)$".r
```
aquamatthias: Since we have delimiter, we should use them in the regexp:
```scala
private val LocalVolumeEncoderRE = s"^([^.]+)[$appDelimiter]([^.]+)[$volumeDelimiter]([^.]+)$".r
```

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -99,25 +101,27 @@ object Task {
...
 
   object LocalVolumeId {
-    def apply(appId: PathId, path: String): LocalVolumeId = {
-      //FIXME: mock implementation from ME
-      LocalVolumeId(s"${appId.safePath}.$path.random")
-    }
-
+    private val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
+    private val appDelimiter = "."
+    private val volumeDelimiter = "."
```
aquamatthias: As long as we have the same delimiter - only use one?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -99,25 +101,27 @@ object Task {
...
   }
 
   object LocalVolumeId {
-    def apply(appId: PathId, path: String): LocalVolumeId = {
-      //FIXME: mock implementation from ME
-      LocalVolumeId(s"${appId.safePath}.$path.random")
-    }
-
+    private val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
+    private val appDelimiter = "."
```
aquamatthias: An app can contain '.' so this is a bad delimiter. 
- '_' is used for the safe path.
- '/' will be used in the container path (and is used in the file system)
- ':' is handled specifically on the file system (at least osx)

Suggestion: '@', '#', '-'

src/main/scala/mesosphere/marathon/tasks/TaskOp.scala
```diff
@@ -0,0 +1,48 @@
+package mesosphere.marathon.tasks
```
aquamatthias: Can we move the TaskOp to `core/*`?

src/main/scala/mesosphere/marathon/tasks/TaskOpLogic.scala
```diff
@@ -0,0 +1,15 @@
+package mesosphere.marathon.tasks
```
aquamatthias: Can we move the TaskOp to `core/*`?

src/main/scala/mesosphere/marathon/tasks/TaskOpFactory.scala
```diff
@@ -0,0 +1,36 @@
+package mesosphere.marathon.tasks
```
aquamatthias: Can we move the TaskOpFactory to `core/*`?

src/main/scala/mesosphere/marathon/tasks/TaskOpLogic.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.state.AppDefinition
+import org.apache.mesos.{ Protos => Mesos }
+
+/** Infers which TaskOps to crate for given app definitions and offers. */
+trait TaskOpLogic {
```
aquamatthias: Just thinking about the name, which is not very clear to me.
Isn't that the TaskOpFactory? (Where the other one is just a helper?)

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
aquamatthias: The tasks in flight already have the launched data - correct? 
That explains why we subtract. 
And yes - I renamed it in the other PR.

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -0,0 +1,94 @@
+package mesosphere.marathon.core.matcher.base.util
+
+import mesosphere.marathon.WrongConfigurationException
+import mesosphere.marathon.core.task.Task.LocalVolume
+import mesosphere.marathon.state.PathId
+import org.apache.mesos.Protos.Resource.ReservationInfo
+import org.apache.mesos.{ Protos => Mesos }
+import org.slf4j.LoggerFactory
+
+class OfferOperationFactory(
```
aquamatthias: Perhaps we should add some test cases for the logic here?

src/main/scala/mesosphere/marathon/tasks/DefaultTaskOpLogic.scala
```diff
@@ -0,0 +1,161 @@
...
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.Task.{ LocalVolume, LocalVolumeId, ReservationWithVolumes }
+import mesosphere.marathon.state.{ AppDefinition, PersistentVolume }
+import mesosphere.mesos.{ PersistentVolumeMatcher, ResourceMatcher, TaskBuilder }
+import org.apache.mesos.{ Protos => Mesos }
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConverters._
+
+class DefaultTaskOpLogic @Inject() (
```
aquamatthias: Perhaps we should add some test cases for the logic here?

aquamatthias:
Hey @meichstedt this looks great!
Made some improvement suggestions.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: exactly, if there was a `TaskOp.Launch` scheduled for them they already have a launched info

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: I merged handleLaunchOp and handleReserveOp as proposed and re-integrated updateStatus into a handleTaskOp – therefore the context is given again :)

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Absolutely, thanks for the hint! -> `handleTaskOp`

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
meichstedt: Ah, that was a note to self with a `FIXME` – removed it.

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: :+1:

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: Yeah why not, changed that.

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: Fair point, I'll change that
(Note that containerPaths including slashes are currently disallowed by Mesos, but that might change of course)

src/main/scala/mesosphere/marathon/tasks/TaskOp.scala
meichstedt: Ok

src/main/scala/mesosphere/marathon/tasks/TaskOpLogic.scala
meichstedt: moved to `core/launcher/*`

src/main/scala/mesosphere/marathon/tasks/TaskOpFactory.scala
meichstedt: moved to `core/launcher/*`

src/main/scala/mesosphere/marathon/tasks/TaskOpLogic.scala
meichstedt: Yes I tried to distinguish the two a little – but I agree, this one's the Factory so I renamed the other one to `...Helper`

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
meichstedt: Done.

src/main/scala/mesosphere/marathon/tasks/DefaultTaskOpLogic.scala
meichstedt: You're right, did that as well

meichstedt:
Hey @aquamatthias, thanks for reviewing and your suggestions – I've addressed them all. Please have a look on those changes as well, I'll squash them if you're OK with them.

meichstedt:
Thanks @aquamatthias I addressed your comments and rebased onto master – let the ship out of the dock!

aquamatthias:
Thanks @meichstedt!!
Wow that's big! The rocket will start soon...

```



                           *     .--.
                                / /  `
               +               | |
                      '         \ \__,
                  *          +   '--'  *
                      +   /\
         +              .'  '.   *
                *      /======\      +
                      ;:.  _   ;
                      |:. (_)  |
                      |:.  _   |
            +         |:. (_)  |          *
                      ;:.      ;
                    .' \:.    / `.
                   / .-'':._.'`-. \
                   |/    /||\    \|
                 _..--"""````"""--.._
           _.-'``                    ``'-._
         -'                                '-
```

PR #3301, opened 2016-02-22
Kill tasks in batches.

blacklab:
Instead of send kill requests for all tasks at once we send them in batches.


aquamatthias:
Thanks!

aquamatthias:
Superseded by #3513

PR #3330, opened 2016-02-24
Fixes #3328 - nicer way of PortDefinition validation.

### LINKED ISSUE #3328: Nicer PortDefinition validation is needed
According to [this comment](https://github.com/mesosphere/marathon/pull/3302/files#diff-08840cca496c5a1088fdc59fda2c84f9) in #3302, there should be a nicer way for validating port definitions. The validation should be implemented element wise, as well as on the sequence itself.

alexanderweber:
Does following:
- removes `Some` which was generated in test messages
- simplifies `PortDefintion` validation by introducing the `elementsAreUniqueBy`, `elementsAreUniqueByOptional`, `elementsAreUniqueFilterWith`.

The `elementsAreUniqueBy` accepts mapping functions like `flatMap`, `map`, but is also able to accept combinations of map and filter like:
```Scala
pd is elementsAreUniqueBy(_.port, filter = { port: Int => port != 42 })
```

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -7,6 +7,7 @@ import mesosphere.marathon.{ AllConf, ValidationFailedException }
 import mesosphere.marathon.state.FetchUri
 import play.api.libs.json._
 
+import scala.collection.GenTraversableOnce
 import scala.reflect.ClassTag
 import scala.util.Try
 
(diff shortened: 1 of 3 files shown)


alexanderweber:
@aquamatthias I had to modify the concept of `fn: A => B` argument in `elementsAreUniqueBy` to `fn: Seq[A] => Seq[B]`, as `map` is not the only combinator one might use. In the case of `PortDefintion` we are in need for a `map`, `flatMap` and a combination with `filterNot`. Providing those externally like: `portDefinitions is elementsAreUniqueBy(_.map(_.port).filterNot(_ == 42)` gives the most flexibility. Let me know if that is ok with you. 
/cc @gkleiman

aquamatthias:
@alexanderweber Thanks for this improvement - very nice!

aquamatthias:
@gkleiman the Some(prop) did not came from accord, but from our own toString method...

PR #3359, opened 2016-02-29
Fixes #3231 by introducing a validator for DeploymentPlan

### LINKED ISSUE #3231: Resident Tasks: Restrictive AppDefinition validation
We should not allow AppDefinitions which lead to badly supported scenarios. That means:

* [ ] Disallow resource/volume config changes.
* [ ] Disallow a rolling upgrade strategy which uses overcapacity.

aquamatthias:
and introducing validators for AppDefinition, that take the original AppDefinition into account.


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -526,4 +530,46 @@ object AppDefinition {
...
   }
+
+  def residentUpdateIsValid(from: AppDefinition): Validator[AppDefinition] = {
+    def changeNoVolumes: Validator[AppDefinition] = new Validator[AppDefinition] {
+      override def apply(to: AppDefinition): Result = {
+        val fromVolumes = from.persistentVolumes
+        val toVolumes = to.persistentVolumes
+        def sameSize = fromVolumes.size == toVolumes.size
+        def noChange = from.persistentVolumes.forall { fromVolume =>
+          toVolumes.find(_.containerPath == fromVolume.containerPath).contains(fromVolume)
```
meichstedt: please add a check for `persistentVolume.persistent.size` – that's the size of the volume itself which will also need to be reserved and must not change.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -526,4 +530,46 @@ object AppDefinition {
...
+        }
+        if (sameSize && noChange) Success
+        else Failure(Set(RuleViolation(to, "Persistent volumes can not be changed!", None)))
```
meichstedt: Just speaking my mind: Originally, we decided to call these `Local Volumes`. However, the underlying Mesos functionality is `Persistent Volumes` and I begin to feel that renaming these will only lead to problems. My proposal: leave this here as it is. We should make sure in a cleanup phase that we have consistent naming (see `LocalVolumeId` as an example ...)

src/main/scala/mesosphere/marathon/state/UpgradeStrategy.scala
```diff
@@ -23,4 +24,9 @@ object UpgradeStrategy {
     strategy.minimumHealthCapacity is between(0.0, 1.0)
     strategy.maximumOverCapacity is between(0.0, 1.0)
   }
+
+  lazy val validForResidentTasks: Validator[UpgradeStrategy] = validator[UpgradeStrategy] { strategy =>
+    strategy.minimumHealthCapacity should be <= 0.5
+    strategy.maximumOverCapacity should be <= 0.00001
```
meichstedt: does `should be 0` not work?

src/test/scala/mesosphere/marathon/api/validation/AppDefinitionValidatorTest.scala
```diff
@@ -362,4 +324,106 @@ class AppDefinitionValidatorTest extends MarathonSpec with Matchers {
     assert(validate(container).isFailure)
   }
 
+  test("Validation for update of resident apps") {
```
meichstedt: Nice test!

src/test/scala/mesosphere/marathon/upgrade/DeploymentPlanTest.scala
```diff
@@ -385,4 +388,59 @@ class DeploymentPlanTest extends MarathonSpec with Matchers with GivenWhenThen {
...
+  }
+
+  test("Deployment plan validation fails for invalid changes in resident tasks") {
+    Given("All options are supplied and we have a valid group change")
+    AllConf.SuppliedOptionNames = Set("mesos_authentication_principal", "mesos_role", "mesos_authentication_secret_file")
+    val f = new Fixture()
+    val validator = DeploymentPlan.changeIsValid(f.config)
+
+    When("We update the upgrade strategy to the default strategy")
+    val app2 = f.validResident.copy(upgradeStrategy = AppDefinition.DefaultUpgradeStrategy, instances = 2)
```
meichstedt: Why `instances = 2`? Should work without that, too.

meichstedt:
@aquamatthias Only minor comments here – however, the ITs fail and it looks like you need to adjust some test format:
```
GroupDeployIntegrationTest:
 *** ABORTED *** (1 minute, 42 seconds)
  java.lang.IllegalArgumentException: could not parse as mesosphere.marathon.integration.setup.ITDeploymentResult:
```

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
aquamatthias: The `contains` method will use equals, wich does a deep comparison of the volumes.
If the size (or any parameter) changes, it will fail.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
aquamatthias: Good point - I agree with your suggestion.

src/main/scala/mesosphere/marathon/state/UpgradeStrategy.scala
aquamatthias: I learned to never use `==` on doubles. Scala implements equality on Double - I can use it.

src/test/scala/mesosphere/marathon/upgrade/DeploymentPlanTest.scala
aquamatthias: Removed.

aquamatthias:
@meichstedt: the underlying problem was something else: 
The validation of a DeploymentPlan happens after the DeploymentPlan has been created.
The validation before checked, if the target group is valid.
With circular dependencies, I got StackOverflowException's - so we have to validate twice: (see GroupManager):
1) if the target group is valid
2) create the deployment plan and validate that plan.

meichstedt:
LGTM!

PR #3370, opened 2016-03-01
Fixes #2404 by introducing GroupInfo for rendering groups and using A…

### LINKED ISSUE #2404: Allow embed fields in /v2/groups endpoint
In the `/v2/apps` endpoint, it's possible to specify a `embed=...` parameter specifying `deployments` and `counts`. 

It would be useful to have this same functionality in the `/v2/groups` endpoint, as the application table in the UI needs this information in order to build the applications table. This can currently only be done by calling both `/v2/groups` and `/v2/apps` and interpolating the results.

aquamatthias:
…ppInfo instead of AppDefinitions directly.

With this change one can  add embed parameters  to the GET groups endpoint.
The default embed parameters match the existing json structure (apps and groups).


src/main/scala/mesosphere/marathon/api/v2/InfoEmbedResolver.scala
```diff
@@ -23,13 +20,16 @@ private[v2] object AppInfoEmbedResolver {
...
+  private[v2] val EmbedApps = "group.apps"
+
   /**
     * Converts embed arguments to our internal representation.
     *
     * Accepts the arguments with all prefixes or even no prefix at all
     * to avoid subtle user errors confusing the two.
     */
-  def resolve(embed: util.Set[String]): Set[Embed] = {
+  def resolveApp(embed: Set[String]): Set[AppInfo.Embed] = {
```
kolloch: Much better with prefix. Thank you!

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,79 @@
...
+      .group(groupId)
+      .map(_.map(_.transitiveApps).getOrElse(Seq.empty))
+      .flatMap(resolveAppInfos(_, embed))
+  }
+
+  override def queryForGroup(group: Group,
+                             appEmbed: Set[AppInfo.Embed],
+                             groupEmbed: Set[GroupInfo.Embed]): Future[GroupInfo] = {
+
+    def queryGroups =
```
kolloch: For me, the return type here is not obvious and would help me in understanding this method.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  override def queryAllInGroup(groupId: PathId, embed: Set[AppInfo.Embed]): Future[Seq[AppInfo]] = {
+    log.debug(s"queryAllInGroup $groupId")
+    groupManager
+      .group(groupId)
+      .map(_.map(_.transitiveApps).getOrElse(Seq.empty))
+      .flatMap(resolveAppInfos(_, embed))
+  }
+
+  override def queryForGroup(group: Group,
```
kolloch: The other methods are careful to prevent round-trips to the taskTracker and other query methods -- which also has the advantage of providing at a consistent snapshots of deployments, tasks. By using recursion here, you are requerying this information multiple times. I guess that this information will be eventually polled in the UI so that performance optimization should be worth it.

docs/docs/rest-api/public/api/v2/groups.raml
```diff
@@ -5,6 +5,32 @@ get:
...
+
+        - <code>group.apps.deployments</code> embed all deployment identifier, if the related app currently is in deployment.
+
+        - <code>group.apps.lastTaskFailure</code> embeds the lastTaskFailure for the application if there is one.
+
+        - <code>group.apps.taskStats</code> exposes task statatistics in the JSON.
+
+      enum: [ group.groups, group.apps, group.apps.tasks, group.apps.count, group.apps.deployments, group.apps.lastTaskFailure, group.apps.failures, group.apps.taskStats ]
+      repeat: true
+      example: group.apps
```
kolloch: I fear the still need to add this to the REST markdown documentation as well.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -13,6 +13,7 @@ import mesosphere.marathon.api.v2.json.AppUpdate
 import mesosphere.marathon.api.v2.json.Formats._
 import mesosphere.marathon.api.{ AuthResource, MarathonMediaType, RestResource }
 import mesosphere.marathon.core.appinfo.{ AppInfo, AppInfoService, AppSelector, TaskCounts }
+import mesosphere.marathon.core.appinfo.AppInfo.Embed
```
kolloch: I like `AppInfo.Embed` in the code much better -- it's clearer to what the embed options apply.

src/main/scala/mesosphere/marathon/api/v2/InfoEmbedResolver.scala
kolloch: and the non-java set as well.

src/main/scala/mesosphere/marathon/api/v2/InfoEmbedResolver.scala
```diff
@@ -46,13 +46,30 @@ private[v2] object AppInfoEmbedResolver {
     }
 
     def separatePrefix(embedMe: String): (String, String) = {
-      val removablePrefix = EmbedAppsPrefixes.find(embedMe.startsWith(_)).getOrElse("")
+      val removablePrefix = EmbedAppsPrefixes.find(embedMe.startsWith).getOrElse("")
```
kolloch: do you really like that better? With Connor I agreed that we mostly like the `(_)` syntax better. That syntax has also the advantage of working with additional default arguments. (Which doesn't matter with "startsWith" admittedly)

kolloch: and I agree that we use the syntax you used here below. It would be interesting to hear what you like better, though.

src/main/scala/mesosphere/marathon/api/v2/InfoEmbedResolver.scala
```diff
@@ -46,13 +46,30 @@ private[v2] object AppInfoEmbedResolver {
...
+    val embedWithSeparatedPrefixes = embed.map(separatePrefix)
+    embedWithSeparatedPrefixes.flatMap { case (prefix, withoutPrefix) => mapEmbedStrings(prefix, withoutPrefix) }
+  }
+
+  def resolveGroup(embeds: Set[String]): Set[GroupInfo.Embed] = {
+    embeds.flatMap {
+      case EmbedGroups => Some(GroupInfo.Embed.Groups)
+      case EmbedApps   => Some(GroupInfo.Embed.Apps)
+      case unknown: String =>
+        log.warn(s"unknown embed argument: $unknown")
```
kolloch: maybe use "unknown group embed argument: ..." here and "unknown app embed argument: ..." above so that we can distinguish the two log statements?

src/main/scala/mesosphere/marathon/core/appinfo/GroupInfoService.scala
```diff
@@ -0,0 +1,17 @@
+package mesosphere.marathon.core.appinfo
+
+import mesosphere.marathon.state.Group
+
+import scala.concurrent.Future
+
+/**
+  * Queries for extended group information.
+  */
+trait GroupInfoService {
```
kolloch: Thanks for a separate trait.

src/test/scala/mesosphere/marathon/api/v2/GroupsResourceTest.scala
```diff
@@ -224,12 +226,24 @@ class GroupsResourceTest extends MarathonSpec with Matchers with Mockito with Gi
   var groupRepository: GroupRepository = _
   var groupsResource: GroupsResource = _
   var auth: TestAuthFixture = _
+  var groupInfo: GroupInfoService = _
+  var embed: java.util.Set[String] = _
```
kolloch: you could use a `val` here and `Collections.emptySet`

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
```diff
@@ -250,15 +255,21 @@ class GroupsResource @Inject() (
...
     val parents = visibleGroups.flatMap(_.allParents) -- visibleGroups
 
-    root.updateGroup { subGroup =>
+    val viewable = root.updateGroup { subGroup =>
       if (visibleGroups.contains(subGroup.id)) Some(subGroup)
       else if (parents.contains(subGroup.id)) Some(subGroup.copy(apps = subGroup.apps.filter(authorizedToViewApp)))
       else None
     }.getOrElse(Group.empty) // fallback, if root is not allowed
+
+    //for backward compatibility, we always render apps and groups, if nothing is given
```
kolloch: Can we maybe get rid of the inline comment and create a constant "DefaultGroupEmbeds" or something like that with the comment?

src/main/scala/mesosphere/marathon/core/appinfo/GroupInfoService.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  /**
+    * Query info for an existing group.
+    */
+  def queryForGroup(group: Group, appEmbed: Set[AppInfo.Embed], groupEmbed: Set[GroupInfo.Embed]): Future[GroupInfo]
```
kolloch: For me, it feels a bit unnatural that you pass the group to this method. Is it because we do the security stuff in the resource layer? Wouldn't it be better to pass the "Identity" into this method and do the filtering here?

kolloch:
I suggested some changes. Tell me what you think!

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
aquamatthias: Added type info.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
aquamatthias: Totally makes sense. Only one call to `resolveAppInfos` should suffice.

docs/docs/rest-api/public/api/v2/groups.raml
aquamatthias: Added the same content into GroupsResource_show.md

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
aquamatthias: We are in AppsResource - it should be clear. 
But I added the long version with line breaks.

src/main/scala/mesosphere/marathon/api/v2/InfoEmbedResolver.scala
aquamatthias: Passing functions should be pretty natural - so yes it pleases my eye more than `(_)`.

src/main/scala/mesosphere/marathon/api/v2/InfoEmbedResolver.scala
aquamatthias: good hint

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
aquamatthias: To whom does this constant belong to?
Added to the GroupsResource.
Not sure I like that solution more.

src/main/scala/mesosphere/marathon/core/appinfo/GroupInfoService.scala
aquamatthias: Generally I agree with your point. The way we handle this in the AppsResource is to implement a special AppSelector. We could implement something like a GroupSelector which extends AppSelector.
But I see this in a separate PR.

aquamatthias:
@kolloch addressed all your points.

docs/docs/rest-api/mesosphere/marathon/api/v2/GroupsResource_show.md
```diff
@@ -1,7 +1,26 @@
...
+Unknown embed parameters are ignored.
+If you omit this parameter, it defaults to `group.groups`, `group.apps`
+
+- `group.groups` embed all child groups of each group<br/>
+
+- `group.apps` embed all apps of each group<br/>
+
+- `group.apps.tasks` embed all tasks of each application<br/>
+
+- `group.apps.counts` embed all task counts (tasksStaged, tasksRunning, tasksHealthy, tasksUnhealthy) <br/>
```
pierlo-upitup: Extra `<br/>` ;)

pierlo-upitup: Are `<br/>`s needed in the `.md` file? (same for all 4 lines above)

src/main/scala/mesosphere/marathon/core/appinfo/GroupInfoService.scala
kolloch: I am not sure if I agree. You call this "queryForGroup", similar to the other methods. Yet, the other methods take care of loading all required data. This is more like "enrichGroupInfo" or something like this

The selector-based approach would at least use the same principle for groups that is used for apps. It feels strange to use something different for such a similar method in the same implementation class.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,85 @@
...
+          else
+            None
+        val apps: Option[Seq[AppInfo]] =
+          if (groupEmbed(GroupInfo.Embed.Apps))
+            Some(ref.apps.toIndexedSeq.flatMap(a => infoById.get(a.id)).sortBy(_.app.id))
+          else
+            None
+        GroupInfo(ref, apps, groups)
+      }
+      queryGroup(group)
```
kolloch: I like it :+1:

kolloch:
I still feel that `queryForGroup` should not get the root Group as argument. Your suggestion with the `GroupSelector` makes sense. Since you introduce this as a new method, I think that you also should introduce the GroupSelector in this PR.

docs/docs/rest-api/mesosphere/marathon/api/v2/GroupsResource_show.md
aquamatthias: No. copied from the raml file (there it is needed).

pierlo-upitup: OK – I wrote that comment because the lines below don't have the `br`s so I thought it was strange.

aquamatthias:
Added Selectors to the InfoService and use them for access control filtering.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,105 @@
...
+    groupManager
+      .group(groupId)
+      .map(_.map(_.transitiveApps.filter(selector.matches)).getOrElse(Seq.empty))
+      .flatMap(resolveAppInfos(_, embed))
+  }
+
+  override def selectGroup(groupId: PathId, groupSelector: GroupSelector,
+                           appEmbed: Set[Embed], groupEmbed: Set[GroupInfo.Embed]): Future[Option[GroupInfo]] = {
+    groupManager.group(groupId).flatMap {
+      case Some(group) => queryForGroup(group, groupSelector, appEmbed, groupEmbed).map(Some(_))
```
kolloch: this method should return `None`if the returned group is not matched by the `groupSelector`.

I would recommend letting `queryForGroup` to return a `Future[Option[GroupInfo]]`as well.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,105 @@
...
+      case None        => Future.successful(None)
+    }
+  }
+
+  override def selectGroupVersion(groupId: PathId, version: Timestamp, groupSelector: GroupSelector,
+                                  groupEmbed: Set[GroupInfo.Embed]): Future[Option[GroupInfo]] = {
+    groupManager.group(groupId, version).flatMap {
+      case Some(group) => queryForGroup(group, groupSelector, Set.empty, groupEmbed).map(Some(_))
```
kolloch: Needs to filter group by `groupSelector`.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,105 @@
...
+    val appInfos = {
+      if (groupEmbed(GroupInfo.Embed.Apps))
+        resolveAppInfos(group.transitiveApps.filter(groupSelector.matches), appEmbed)
+      else
+        Future.successful(Seq.empty)
+    }
+
+    appInfos.map { apps =>
+      val infoById = apps.map(info => info.app.id -> info).toMap
+      def queryGroup(ref: Group): GroupInfo = {
```
kolloch: Possible implementation: let this return `Option[GroupInfo]` and filter the group itself with the `groupSelector`.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,105 @@
...
+        val groups: Option[Seq[GroupInfo]] =
+          if (groupEmbed(GroupInfo.Embed.Groups))
+            Some(ref.groups.toIndexedSeq.filter(groupSelector.matches).map(queryGroup).sortBy(_.group.id))
```
kolloch: ...then you can use a `flatMap` here and remove the `filter` here.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  override def selectApp(id: PathId, selector: AppSelector, embed: Set[AppInfo.Embed]): Future[Option[AppInfo]] = {
+    log.debug(s"queryForAppId $id")
+    appRepository.currentVersion(id).flatMap {
+      case Some(app) if selector.matches(app) => newBaseData().appInfoFuture(app, embed).map(Some(_))
+      case None                               => Future.successful(None)
+    }
+  }
+
+  override def selectBy(selector: AppSelector, embed: Set[AppInfo.Embed]): Future[Seq[AppInfo]] = {
```
kolloch: I would prefer `selectApps` or `selectAppsBy` since you have methods relating to groups in the same class.

kolloch:
Cool! I like the selector approach.

I think that there is a bug, though, because you do not filter the group that you load itself with the `groupSelector`.

aquamatthias:
Thanks. Addressed your comments.

aquamatthias:
@kolloch added a map lookup (yes there is a getOrElseUpdate even in the standard map)
Squashed and rebased.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -0,0 +1,112 @@
...
+          else
+            None
+        def apps: Option[Seq[AppInfo]] =
+          if (groupEmbed(GroupInfo.Embed.Apps))
+            Some(ref.apps.toIndexedSeq.flatMap(a => infoById.get(a.id)).sortBy(_.app.id))
+          else
+            None
+        //if a subgroup is allowed, we also have to allow all parents implicitly
+        def groupMatches(group: Group): Boolean = {
+          alreadyMatched.getOrElseUpdate(group.id, groupSelector.matches(group) || group.groups.exists(groupMatches))
```
kolloch: didn't know `getOrElseUpdate`, nice!

kolloch:
LGTM! Thanks :rocket:

PR #3379, opened 2016-03-01
send notifications in parallel to subscribers

fr0stbyte:
Currently the http callback subscribers are notified sequentially, so if one is experiencing problems it does affect the well functioning of the rest.
From what I gather, the exponential back-off is per endpoint, so parallelizing should alleviate the described problem with no side-effects on the backoff algorithm.


aquamatthias:
Hey @fr0stbyte thanks for this PR.
I think your assumption is wrong: all listeners are informed (almost) in parallel. 
Even if one listener is slow, it should not affect any other listener:
The call to `post` is synchronous, but the pipeline spawns a `Future` for every request.
The `Future` itself lives until the request is performed (if possible), but that path of execution is detached from the `broadcast`/`post` calls.

Many slow consumers can have an impact, since the thread pool is limited.

aquamatthias:
I will close this PR. Please reopen if you disagree.

PR #3387, opened 2016-03-03
Fixes #3339 - path output format of object validation is identical to JSON structure

### LINKED ISSUE #3339: Port indices notation in validation errors is not consistent
- In some cases Marathon will use parentheses to index validation failures, in other cases it will use square brackets.
- Paths sometimes have a leading slash, but they sometimes don't.
- Paths are sometimes separated by slashes, sometimes by periods.

Some examples:

```json
{
    "details": [
        {
            "errors": [
                "Constraint definition must be an array of strings in format: <key>, <operator>[, value]"
            ],
            "path": "/constraints(0)"
        }
    ],
    "message": "Invalid JSON"
}
```

```json
{
    "details": [
        {
            "errors": [
                "URI has invalid syntax."
            ],
            "path": "fetch[0]"
        }
    ],
    "message": "Object is not valid"
}
```

```json
{
    "details": [
        {
            "errors": [
                "error.path.missing"
            ],
            "path": "/fetch(0)/uri"
        }
    ],
    "message": "Invalid JSON"
}
```

```json
{
    "details": [
        {
            "errors": [
                "got -1, expected 0 or more"
            ],
            "path": "portDefinitions[0].port"
        }
    ],
    "message": "Object is not valid"
}
```

The schema should be consistent. Until v0.15.0 we were always using parentheses, I think we should go back to that.

To find the affected validations, look for usages of the `every` validator.

cc/ @aldipower @alexanderweber

alexanderweber:
As discussed in #3339, following changes have been applied:

1. replace `[` `]` with `(` `)` respectively
2. replace `.` by `/`
3. prepend `/` at the root of each path. This would also replace self with `/` in root-only errors
4. replace `self` with `/`
5. correct docs

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -113,8 +113,10 @@ put:
               "message": "Object is not valid",
               "details": [
                 {
-                  "attribute": "upgradeStrategy.minimumHealthCapacity",
-                  "error": "is greater than 1"
+                  "path": "/upgradeStrategy/minimumHealthCapacity",
+                  "errors": [
+                    "is greater than 1"
+                  ]
                 }
               ]
             }
@@ -163,8 +165,10 @@ post:
               "message": "Object is not valid",
               "details": [
                 {
-                  "attribute": "upgradeStrategy.minimumHealthCapacity",
-                  "error": "is greater than 1"
+                  "path": "/upgradeStrategy/minimumHealthCapacity",
+                  "errors": [
+                    "is greater than 1"
+                  ]
                 }
               ]
             }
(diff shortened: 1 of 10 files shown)


alexanderweber:
@aquamatthias validation output is adjusted to be identical to JSON structure. I also did adjust the documentation.
cc/ @gkleiman I did address your concerns of #3339 in this PR.

aquamatthias:
@alexanderweber Thanks and Looks good!
@pierlo-upitup Please see the results in the tests. Do you agree on this approach?

pierlo-upitup:
cc @philipnrmn

philipnrmn:
Much neater!

pierlo-upitup:
This completely destroys our API error parsing again, obviously. Creating another issue for us to align to the new format.

PR #3396, opened 2016-03-03
Wait for taskOp acknowledgements during shutdown in AppTaskLaunchActor

kolloch:
Very difficult to test, sorry


meichstedt:
At least `AppTaskLauncherActorTest` tests the message handling on the actor side …
LGTM otherwise.

PR #3398, opened 2016-03-03
Fixes #3385 - groups with same IDs as apps should not be created.

### LINKED ISSUE #3385: Creating an empty group using an existing app ID should return 409
When sending a `POST` to `/v2/groups` with a group `{"id": "/group/existing-appname"}` I’d expect to receive a `409`, but I am getting a `200` instead 

```
{
  deploymentId: "7a8ad237-a76b-428e-8915-5dc9c98716b3"
  version: "2016-03-02T11:02:45.586Z"
}
```

alexanderweber:
Fixes #3385.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
@@ -106,14 +106,19 @@ class GroupsResource @Inject() (
                      @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
     withValid(Json.parse(body).as[GroupUpdate]) { groupUpdate =>
       val effectivePath = groupUpdate.id.map(_.canonicalPath(id.toRootPath)).getOrElse(id.toRootPath)
+      val rootGroup = result(groupManager.rootGroup())
 
-      result(groupManager.rootGroup()).findGroup(_.id == effectivePath) match {
-        case Some(currentGroup) =>
-          throw ConflictingChangeException(s"Group $effectivePath is already created. Use PUT to change this group.")
-        case None =>
-          val (deployment, path) = updateOrCreate(id.toRootPath, groupUpdate, force)
-          deploymentResult(deployment, Response.created(new URI(path.toString)))
+      def throwIfConflicting[A](conflict: Option[Any], msg: String) = {
+        conflict.map(_ => throw ConflictingChangeException(msg))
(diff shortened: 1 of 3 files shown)


alexanderweber:
@aquamatthias there was no check against identical path definition of an app with a newly created group. It was possible to create an empty group `"/group/existing-appname"` even if there was an app created as `"/group/existing-appname"` before.

I did add this check in GroupsResource:

```Scala
result(groupManager.rootGroup()).transitiveApps.find(_.id == path)
```

Please merge, if you agree with the change.

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
```diff
@@ -105,14 +105,18 @@ class GroupsResource @Inject() (
...
 
-      result(groupManager.rootGroup()).findGroup(_.id == effectivePath) match {
+      result(groupManager.rootGroup()).findGroup(_.id == path) match {
         case Some(currentGroup) =>
-          throw ConflictingChangeException(s"Group $effectivePath is already created. Use PUT to change this group.")
+          throw ConflictingChangeException(s"Group $path is already created. Use PUT to change this group.")
         case None =>
-          val (deployment, path) = updateOrCreate(id.toRootPath, groupUpdate, force)
-          deploymentResult(deployment, Response.created(new URI(path.toString)))
+          result(groupManager.rootGroup()).transitiveApps.find(_.id == path) match {
```
aquamatthias: The call to groupManager.rootGroup() is expensive and you call it twice here. Please reuse the root group from the first call.

aquamatthias: Also: nested match elements are sometimes hard to read: extract def's

aquamatthias:
@alexanderweber thanks - added some suggestions.

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
alexanderweber: Totally agree. Fixed and encapsulated, let me know if it is ok.

aquamatthias:
Looks great! Can we have a test for this?

alexanderweber:
@aquamatthias I did write two tests. Let me know if those are ok.

aquamatthias:
@thanks!

Issue #3391, opened 2016-03-03
Adjust resource matching to include disk resources for local volumes


kolloch:
Currently, that's no the case. See `TaskOpFactory`.

PR #3404, opened 2016-03-04
Fixes #1606 and #3391 - Fixed resource matching for resident tasks

### LINKED ISSUE #1606: Aggregate scalar resources across roles
Currently, all scalar resources have to be satisfied by one role. So, for example, if you specify that you need two CPUs and one is offered with role "*" and one is offered with role "public", you task will not launch.

For port resources already all resources disrespecting their roles are considered.

kolloch:


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -104,6 +104,11 @@ case class AppDefinition(
     container.fold(Seq.empty[Volume])(_.volumes).collect{ case vol: PersistentVolume => vol }
   }
 
+  /**
+    * @return the disk resources required for volumes
+    */
+  def volumeDiskResources: Double = persistentVolumes.map(_.persistent.size).sum.toDouble
```
meichstedt: since the existing resource property is called `disk`, I'm inclined to call this `diskForVolumes`.

src/main/scala/mesosphere/marathon/tasks/ResourceUtil.scala
```diff
@@ -154,33 +167,48 @@ object ResourceUtil {
...
+      val principalString = if (resource.hasReservation && resource.getReservation.hasPrincipal)
+        s", RESERVED for ${resource.getReservation.getPrincipal}"
+      else
+        ""
+      val diskString = if (resource.hasDisk && resource.getDisk.hasPersistence)
+        s", diskId ${resource.getDisk.getPersistence.getId}"
+      else
+        ""
+
+      s"${resource.getName}(${resource.getRole}$principalString$diskString)"
```
meichstedt: :+1:

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -15,66 +16,88 @@ import scala.collection.mutable
...
-    def matchingRole: Option[Role] = if (matches) Some(role) else None
+  private[this] val log = LoggerFactory.getLogger(getClass)
+
+  /**
+    * A successful match result of the [[ResourceMatcher]].matchResources method.
+    */
+  case class ResourceMatch(scalarMatches: Iterable[ScalarMatch], portsMatch: PortsMatch) {
+    lazy val hostPorts: Seq[Int] = portsMatch.hostPorts
+
+    def scalarMatch(name: String): Option[ScalarMatch] = scalarMatches.find(_.resourceName == name)
```
meichstedt: Only used in tests and I guess you meant it as a test helper – so I'd either make it package private or put it into the test as a helper function

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -15,66 +16,88 @@ import scala.collection.mutable
...
+    * Disk resources are always discarded, since we do not want to match them by
+    * accident.
+    *
+    * @param acceptedRoles contains all Mesos resource roles that are accepted
+    * @param reserved if reserved is true, only resources with a ReservationInfo
+    *                 are considered. If reserved is false, only resources without
+    *                 a ReservationInfo are considered.
+    */
+  case class ResourceSelector(acceptedRoles: Set[String], reserved: Boolean) {
+    def apply(resource: Protos.Resource): Boolean = {
```
meichstedt: I really like the application of a resourceSelector :+1:

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -15,66 +16,88 @@ import scala.collection.mutable
...
-  //scalastyle:off method.length
+  /**
+    * Checks whether the given offer contains enough resources to launch a task of the given app
+    * or to make a reservation for a task.
+    *
+    * If a task uses local volumes, this method is typically called twice for every launch. Once
+    * for the reservation on UNRESERVED resources and once for every (re-)launch on RESERVED resources.
+    *
+    * If matching on RESERVED resources as specified by the ResourceSelector, resources for volumes
+    * have to matched separately (e.g. by the [[PersistentVolumeMatcher]]). If matching on UNRESERVED
```
meichstedt: s/have to matched/have to be matched

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -15,66 +16,88 @@ import scala.collection.mutable
...
+    * resources, the disk resources for the local volumes are included since they must be part of
```
meichstedt: s/since they must be/since they must become

Thanks for this readable and helpful explanation!

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -15,66 +16,88 @@ import scala.collection.mutable
...
+    * the reservation.
+    */
   def matchResources(offer: Offer, app: AppDefinition, runningTasks: => Iterable[Task],
-                     acceptedResourceRoles: Set[String] = Set("*")): Option[ResourceMatch] = {
+                     selector: ResourceSelector = ResourceSelector.wildcard): Option[ResourceMatch] = {
```
meichstedt: Is the default value a leftover? Production code always explicitly sets the ResourceSelector and I think it would be better to not assume a default.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -92,37 +115,56 @@ object ResourceMatcher {
...
+        portsMatch <- portsMatchOpt
+        if meetsAllConstraints
+      } yield ResourceMatch(scalarMatchResults.collect { case m: ScalarMatch => m }, portsMatch)
+    }
+    else {
+      None
+    }
+  }
+
+  private[this] def matchScalarResource(
```
meichstedt: Can you add a comment here or rename the `note` param to something more explanatory? It's not really obvious here what the param is meant for since it's passed along to the `ScalarMatch`

src/test/scala/mesosphere/marathon/tasks/ResourceUtilTest.scala
```diff
@@ -15,26 +18,48 @@ class ResourceUtilTest extends FunSuite with GivenWhenThen with Assertions {
...
-      Seq(scalar("cpus", 2), scalar("cpus", 2, role = "marathon")),
-      Seq(scalar("cpus", 0.5), scalar("cpus", 1, role = "marathon"), scalar("cpus", 0.5, role = "marathon"))
+      Seq(MTH.scalarResource("cpus", 2), MTH.scalarResource("cpus", 2, role = "marathon")),
+      Seq(MTH.scalarResource("cpus", 0.5), MTH.scalarResource("cpus", 1, role = "marathon"), MTH.scalarResource("cpus", 0.5, role = "marathon"))
     )
-    assert(leftOvers == Seq(scalar("cpus", 1.5), scalar("cpus", 0.5, role = "marathon")))
+    assert(leftOvers == Seq(MTH.scalarResource("cpus", 1.5), MTH.scalarResource("cpus", 0.5, role = "marathon")))
+  }
+
+  test("resource consumption considers reservation") {
```
meichstedt: The testName does not really reflect what is tested here – shouldn't the test code call `ResourceUtil.consumeResources`?

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
```diff
@@ -30,13 +34,11 @@ class ResourceMatcherTest extends MarathonSpec with Matchers {
...
-    res.memRole should be("*")
-    res.diskRole should be("")
+    res.scalarMatch(Resource.CPUS).get.roles should be(Seq("*"))
+    res.scalarMatch(Resource.MEM).get.roles should be(Seq("*"))
+    res.scalarMatch(Resource.DISK).get.roles should be(Seq.empty)
 
-    // check if we got 2 ports
-    val range = res.ports.head.ranges.head
-    (range.end - range.begin) should be (1)
+    res.hostPorts should have size 2
```
meichstedt: the previous test code also tested that the ports are subsequent numbers – not the responsibility of this test, probably there's another test that verifies this …?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
kolloch: Thanks for the thx and the corrections!

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
kolloch: Ack

meichstedt:
Only minor comments here, @kolloch. LGTM apart from those :)

src/test/scala/mesosphere/marathon/tasks/ResourceUtilTest.scala
kolloch: Yes, I guess a copy paste confusion occurred :)

Thanks for noticing!

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
kolloch: It is not guaranteed and not a required property. Actually, the original test could have occasionally failed since the chosen ports are randomized (since there is only one range, the probability that it would choose the last port as the first port and then wrap around was small)

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
kolloch: As discussed, I'll keep it. There shouldn't be harm in production code calling this helper.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
kolloch: Ack

meichstedt:
```
               o<   o     o     o     o
o___________\_((__\-))__\-))__\-))__\-))_____________o   Thanks
  `-------------------------\-----------\-----------'            @kolloch!
                              \           \
                                \           \
```

Issue #3221, opened 2016-02-11
Resident Tasks: Acknowledgement/Timeout for Reservations/Local Volume Creates


kolloch:
Mesos does not actively acknowledge reservations or volume creation. Therefore we need to wait for offers that include the reservations and the create volumes. If we see an offer from the same host after we send Mesos the operations which does have the effects, we know that our operation succeeded. If we do see an offer afterwards without the expected effects, we do know that the operations did not succeed. If we do not see such an offer during a certain period of time, we have to assume that the operations did not succeed.

Acceptance criteria:

- [ ] we check the expected effects on offers and modify reservation state on tasks appropriately such that we retry the operations
- [ ] we enforce a timeout on this operations -- after this timeout, even if we do not see any offer by this host we will assume that the operations failed and retry

Does not include handling of `TASK_LOST`, see #3368.
Also related to garbage collect reservations that don't exist anymore, see #3455.
/cc @meichstedt @aquamatthias

PR #3411, opened 2016-03-07
Towards #3221 reservation timeouts

meichstedt:
Goal is to reflect a tasks state so we can apply specific timeouts to certain states

- changed Task to a sealed trait with Reserved, LaunchedOnReservation and LaunchedEphemeral implementations
- task state changes are triggered via TaskStateOps
- resulting TaskStateChange will be used to update tasks in TaskTracker/AppTaskLauncherActor
- reflected changes in proto

Open:
- more tests, especially regarding `Task.update()`
- removal of legacy usages (`.launched`, `.reservationWithVolumes`)

### DIFF
--- a/src/main/proto/marathon.proto
+++ b/src/main/proto/marathon.proto
@@ -92,27 +92,30 @@ message ResourceRoles {
 
 message MarathonTask {
   message Reservation {
-    repeated string local_volume_ids = 1;
-  }
-  message ReservedState {
-    enum Type {
-      New = 1;
-      Suspended = 2;
-      Garbage = 3;
-      Unknown = 4;
-    }
-    message Timeout {
-      enum Reason {
-        RelaunchEscalationTimeout = 1;
-        ReservationTimeout = 2;
+    message State {
+      enum Type {
+        New = 1;
+        Launched = 2;
+        Suspended = 3;
+        Garbage = 4;
+        Unknown = 5;
       }
+      message Timeout {
+        enum Reason {
+          RelaunchEscalationTimeout = 1;
+          ReservationTimeout = 2;
+        }
 
-      required int64 initiated = 1;
-      required int64 deadline = 2;
-      required Reason reason = 3;
+        required int64 initiated = 1;
+        required int64 deadline = 2;
+        required Reason reason = 3;
+      }
+      required Type type = 1;
+      optional Timeout timeout = 2;
     }
-    required Type type = 1;
-    optional Timeout timeout = 2;
+
+    repeated string local_volume_ids = 1;
+    required State state = 2;
   }
 
   required string id = 1;
@@ -127,7 +130,6 @@ message MarathonTask {
   optional mesos.SlaveID slaveId = 10;
   repeated mesos.NetworkInfo networks = 11;
   optional Reservation reservation = 12; // since 0.16, a list of volumes can be associated with the task ID
-  optional ReservedState reserved_state = 13;
 }
 
 message MarathonApp {
--- a/src/main/proto/marathon.proto
+++ b/src/main/proto/marathon.proto
@@ -91,8 +91,28 @@ message ResourceRoles {
 }
 
 message MarathonTask {
-  message ReservationWithVolumes {
+  message Reservation {
     repeated string local_volume_ids = 1;
+  }
+  message ReservedState {
+    enum Type {
+      New = 1;
+      Suspended = 2;
+      Garbage = 3;
+      Unknown = 4;
+    }
+    message Timeout {
+      enum Reason {
+        RelaunchEscalationTimeout = 1;
+        ReservationTimeout = 2;
+      }
+
+      required int64 initiated = 1;
+      required int64 deadline = 2;
+      required Reason reason = 3;
+    }
+    required Type type = 1;
+    optional Timeout timeout = 2;
   }
 
   required string id = 1;
@@ -106,7 +126,8 @@ message MarathonTask {
   optional mesos.TaskStatus status = 9;
   optional mesos.SlaveID slaveId = 10;
   repeated mesos.NetworkInfo networks = 11;
-  optional ReservationWithVolumes reservation_with_volumes = 12; // since 0.16, a list of volumes can be associated with the task ID
+  optional Reservation reservation = 12; // since 0.16, a list of volumes can be associated with the task ID
+  optional ReservedState reserved_state = 13;
 }
 
 message MarathonApp {
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
@@ -5,8 +5,7 @@ import mesosphere.marathon.MarathonConf
 import mesosphere.marathon.core.base.Clock
 import mesosphere.marathon.core.launcher.{ TaskOp, TaskOpFactory }
 import mesosphere.marathon.core.task.{ TaskStateChange, TaskStateOp, Task }
-import mesosphere.marathon.core.task.Task.{ LocalVolume, LocalVolumeId, ReservationWithVolumes }
-import mesosphere.marathon.state.{ AppDefinition, PersistentVolume }
+import mesosphere.marathon.state.AppDefinition
 import mesosphere.mesos.ResourceMatcher.ResourceSelector
 import mesosphere.mesos.{ PersistentVolumeMatcher, ResourceMatcher, TaskBuilder }
 import org.apache.mesos.{ Protos => Mesos }
@@ -147,8 +146,8 @@ class TaskOpFactoryImpl @Inject() (
     offer: Mesos.Offer,
     resourceMatch: ResourceMatcher.ResourceMatch): TaskOp = {
 
-    val localVolumes: Iterable[LocalVolume] = app.persistentVolumes.map { volume =>
-      LocalVolume(LocalVolumeId(app.id, volume), volume)
+    val localVolumes: Iterable[Task.LocalVolume] = app.persistentVolumes.map { volume =>
+      Task.LocalVolume(Task.LocalVolumeId(app.id, volume), volume)
     }
     val persistentVolumeIds = localVolumes.map(_.id)
     val task = Task.Reserved(
@@ -158,8 +157,7 @@ class TaskOpFactoryImpl @Inject() (
         agentId = Some(offer.getSlaveId.getValue),
         attributes = offer.getAttributesList.asScala
       ),
-      state = Task.Reserved.State.New(timeout = None),
-      reservation = ReservationWithVolumes(persistentVolumeIds)
+      reservation = Task.Reservation(persistentVolumeIds, Task.Reservation.State.New(timeout = None))
     )
     taskOperationFactory.reserveAndCreateVolumes(task, resourceMatch.resources, localVolumes)
   }
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
@@ -4,7 +4,7 @@ import com.google.inject.Inject
 import mesosphere.marathon.MarathonConf
 import mesosphere.marathon.core.base.Clock
 import mesosphere.marathon.core.launcher.{ TaskOp, TaskOpFactory }
-import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.{ TaskStateChange, TaskStateOp, Task }
 import mesosphere.marathon.core.task.Task.{ LocalVolume, LocalVolumeId, ReservationWithVolumes }
 import mesosphere.marathon.state.{ AppDefinition, PersistentVolume }
 import mesosphere.mesos.ResourceMatcher.ResourceSelector
@@ -41,22 +41,18 @@ class TaskOpFactoryImpl @Inject() (
   private[this] def inferNormalTaskOp(app: AppDefinition, offer: Mesos.Offer, tasks: Iterable[Task]): Option[TaskOp] = {
     new TaskBuilder(app, Task.Id.forApp, config).buildIfMatches(offer, tasks).map {
       case (taskInfo, ports) =>
-        val task = Task(
+        val task = Task.LaunchedEphemeral(
           taskId = Task.Id(taskInfo.getTaskId),
           agentInfo = Task.AgentInfo(
             host = offer.getHostname,
             agentId = Some(offer.getSlaveId.getValue),
             attributes = offer.getAttributesList.asScala
           ),
-          launched = Some(
-            Task.Launched(
-              appVersion = app.version,
-              status = Task.Status(
-                stagedAt = clock.now()
-              ),
-              networking = Task.HostPorts(ports)
-            )
-          )
+          appVersion = app.version,
+          status = Task.Status(
+            stagedAt = clock.now()
+          ),
+          networking = Task.HostPorts(ports)
         )
         taskOperationFactory.launch(taskInfo, task)
     }
@@ -128,23 +124,21 @@ class TaskOpFactoryImpl @Inject() (
     // create a TaskBuilder that used the id of the existing task as id for the created TaskInfo
     new TaskBuilder(app, (_) => task.taskId, config).build(offer, resourceMatch, volumeMatch) map {
       case (taskInfo, ports) =>
-        val newTask = task.copy(
-          agentInfo = Task.AgentInfo(
-            host = offer.getHostname,
-            agentId = Some(offer.getSlaveId.getValue),
-            attributes = offer.getAttributesList.asScala
+        val launch = TaskStateOp.Launch(
+          appVersion = app.version,
+          status = Task.Status(
+            stagedAt = clock.now()
           ),
-          launched = Some(
-            Task.Launched(
-              appVersion = app.version,
-              status = Task.Status(
-                stagedAt = clock.now()
-              ),
-              networking = Task.HostPorts(ports)
-            )
-          )
-        )
-        taskOperationFactory.launch(taskInfo, newTask)
+          networking = Task.HostPorts(ports))
+
+        // FIXME (3221): something like reserved.launch(...): LaunchedOnReservation so we don't need to match?
+        task.update(launch) match {
+          case TaskStateChange.Update(updatedTask) =>
+            taskOperationFactory.launch(taskInfo, updatedTask)
+
+          case unexpected: TaskStateChange =>
+            throw new scala.RuntimeException(s"Expected TaskStateChange.Update but got $unexpected")
+        }
     }
   }
 
@@ -157,15 +151,15 @@ class TaskOpFactoryImpl @Inject() (
       LocalVolume(LocalVolumeId(app.id, volume), volume)
     }
     val persistentVolumeIds = localVolumes.map(_.id)
-    val task = Task(
+    val task = Task.Reserved(
       taskId = Task.Id.forApp(app.id),
       agentInfo = Task.AgentInfo(
         host = offer.getHostname,
         agentId = Some(offer.getSlaveId.getValue),
         attributes = offer.getAttributesList.asScala
       ),
-      launched = None,
-      reservationWithVolumes = Some(ReservationWithVolumes(persistentVolumeIds))
+      state = Task.Reserved.State.New(timeout = None),
+      reservation = ReservationWithVolumes(persistentVolumeIds)
     )
     taskOperationFactory.reserveAndCreateVolumes(task, resourceMatch.resources, localVolumes)
   }
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -17,9 +17,10 @@ import org.slf4j.LoggerFactory
 sealed trait Task {
   def taskId: Task.Id
   def agentInfo: Task.AgentInfo
-  def reservationWithVolumes: Option[Task.ReservationWithVolumes]
+  def reservationWithVolumes: Option[Task.Reservation]
   def launched: Option[Task.Launched]
 
+  /** update the task based on the given trigger - depending on its state the task will decide what should happen */
   def update(update: TaskStateOp): TaskStateChange
 
   def appId: PathId = taskId.appId
@@ -64,20 +65,6 @@ sealed trait Task {
 }
 
 object Task {
-  private[this] def newHealthStatus(
-    current: MesosProtos.TaskStatus,
-    update: MesosProtos.TaskStatus): Option[MesosProtos.TaskStatus] = {
-
-    val healthy = update.hasHealthy && (!current.hasHealthy || current.getHealthy != update.getHealthy)
-    val changed = healthy || current.getState != update.getState
-    if (changed) {
-      Some(update)
-    }
-    else {
-      None
-    }
-  }
-
   /**
     * A LaunchedEphemeral task is a stateless task that does not consume reserved resources or persistent volumes.
     */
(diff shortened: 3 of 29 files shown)


src/main/proto/marathon.proto
```diff
@@ -91,9 +91,29 @@ message ResourceRoles {
...
+      Unknown = 4;
+    }
+    message Timeout {
+      enum Reason {
+        RelaunchEscalationTimeout = 1;
+        ReservationTimeout = 2;
+      }
+
+      required int64 initiated = 1;
+      required int64 deadline = 2;
```
kolloch: thanks for not using a string here :+1:

src/main/proto/marathon.proto
```diff
@@ -106,7 +126,8 @@ message MarathonTask {
   optional mesos.TaskStatus status = 9;
   optional mesos.SlaveID slaveId = 10;
   repeated mesos.NetworkInfo networks = 11;
-  optional ReservationWithVolumes reservation_with_volumes = 12; // since 0.16, a list of volumes can be associated with the task ID
+  optional Reservation reservation = 12; // since 0.16, a list of volumes can be associated with the task ID
+  optional ReservedState reserved_state = 13;
```
kolloch: Why is `ReservedState` not part of `Reservation`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -2,23 +2,25 @@ package mesosphere.marathon.core.task
...
-    agentInfo: Task.AgentInfo,
-    reservationWithVolumes: Option[Task.ReservationWithVolumes] = None,
-    launched: Option[Task.Launched] = None) {
+sealed trait Task {
+  def taskId: Task.Id
+  def agentInfo: Task.AgentInfo
+  def reservationWithVolumes: Option[Task.ReservationWithVolumes]
+  def launched: Option[Task.Launched]
+
+  def update(update: TaskStateOp): TaskStateChange
```
kolloch: Maybe a short description?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
 }
 
 object Task {
+  private[this] def newHealthStatus(
```
kolloch: Nit-pick: This method really seems to be a detail implementation.

I would move it further down so that the more important things come first. Nevertheless a short description would be nice.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+    * A LaunchedEphemeral task is a stateless task that does not consume reserved resources or persistent volumes.
+    */
+  case class LaunchedEphemeral(
+      taskId: Task.Id,
+      agentInfo: AgentInfo,
+      appVersion: Timestamp,
+      status: Status,
+      networking: Networking) extends Task {
+
+    private[this] val log = LoggerFactory.getLogger(getClass)
```
kolloch: Put the logger into the companion object. No need to resolve the logger for every task instance.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+
+    override val reservationWithVolumes: Option[ReservationWithVolumes] = None
```
kolloch: why note `def`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+
+    override val launched: Option[Launched] = Some(Task.Launched(appVersion, status, networking))
```
kolloch: why not `def`? Saves space.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+            startedAt = Some(now),
+            mesosStatus = mesosStatus))
+        TaskStateChange.Update(updated)
+
+      // case 2: terminal
+      case TaskStateOp.MesosUpdate(MarathonTaskStatus.Terminal(_), now) =>
+        TaskStateChange.Expunge
+
+      // case 3: health updated
+      case TaskStateOp.MesosUpdate(taskStatus, now) =>
```
kolloch: What if the `MesosState` changed? I think that the assumption here is that the task is running and stays running. That might or might not be always the case. Can we make it more obvious? At least the assumption?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+
+        healthChange.getOrElse{
+          log.debug("Ignoring status update for {}. Status did not change.", taskId)
+          TaskStateChange.NoChange
+        }
+
+      // failure case: Launch
+      case _: TaskStateOp.Launch =>
+        val msg = s"Unable to handle Launch op on LaunchedEphemeral $taskId"
+        log.error(msg, taskId)
```
kolloch: Do not log error here. Caller should handle failure.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+        TaskStateChange.Failure(msg)
+
+      // failure case: Timeout
+      case TaskStateOp.Timeout =>
+        val msg = s"Unable to handle Timeout op on LaunchedEphemeral $taskId"
+        log.error(msg)
```
kolloch: Do not log error here. Caller should handle failure.

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
```diff
@@ -0,0 +1,23 @@
...
+import mesosphere.marathon.state.Timestamp
+
+sealed trait TaskStateOp
+
+object TaskStateOp {
+  case class Launch(appVersion: Timestamp, status: Task.Status, networking: Task.Networking) extends TaskStateOp
+  // FIXME (3221): MarathonTaskStatus was introduced for the message bus – is it ok to use it here?
+  case class MesosUpdate(status: MarathonTaskStatus, now: Timestamp) extends TaskStateOp
+  case object Timeout extends TaskStateOp
+  //  Garbage?
```
kolloch: Garbage would not be a transition, I think. It would affect the state. If you want to destroy a task including the volume, you would need to mark the `LaunchOnReservation` as `destroyMe` and then transition to `Task.Garbage` when receiving the kill confirmation. Or something similar.

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import mesosphere.marathon.core.task.bus.MarathonTaskStatus
+import mesosphere.marathon.state.Timestamp
+
+sealed trait TaskStateOp
+
+object TaskStateOp {
+  case class Launch(appVersion: Timestamp, status: Task.Status, networking: Task.Networking) extends TaskStateOp
+  // FIXME (3221): MarathonTaskStatus was introduced for the message bus – is it ok to use it here?
+  case class MesosUpdate(status: MarathonTaskStatus, now: Timestamp) extends TaskStateOp
```
kolloch: The original plan was to get rid of `MarathonTaskStatus` again. If it proves useful we can keep it despite getting rid of the message bus.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+        TaskStateChange.Update(LaunchedOnReservation(
+          taskId, agentInfo, appVersion, status, networking, reservation))
+
+      case TaskStateOp.Timeout =>
+        TaskStateChange.Expunge
+
+      // failure case: MesosUpdate
+      case _: TaskStateOp.MesosUpdate =>
+        val msg = s"Unable to handle MesosUpdate op on Reserved task $taskId"
+        log.error(msg)
```
kolloch: caller should handle error... Will not repeat below...

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
```diff
@@ -0,0 +1,23 @@
...
+  case object Timeout extends TaskStateOp
```
kolloch: We need to clarify whether the Timeout only applies to reservations or everything. If it only applies to Reservations, please rename to "ReservationTimeout" or something more specific.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+
+  case class LaunchedOnReservation(
+      taskId: Task.Id,
+      agentInfo: AgentInfo,
+      appVersion: Timestamp,
+      status: Status,
+      networking: Networking,
+      reservation: ReservationWithVolumes) extends Task {
+
+    private[this] val log = LoggerFactory.getLogger(getClass)
```
kolloch: Put the logger into the companion object. No need to resolve the logger for every task instance.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+
+    override lazy val reservationWithVolumes: Option[ReservationWithVolumes] = Some(reservation)
```
kolloch: `def`

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+
+    override lazy val launched: Option[Launched] = Some(Task.Launched(appVersion, status, networking))
```
kolloch: `def`

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -67,6 +64,211 @@ case class Task(
...
+      case TaskStateOp.MesosUpdate(MarathonTaskStatus.Terminal(_), now) =>
+        TaskStateChange.Update(Task.Reserved(
+          taskId = taskId,
+          agentInfo = agentInfo,
+          state = Task.Reserved.State.Suspended(timeout = None),
+          reservation = reservation
+        ))
+
+      // case 3: health updated
+      case TaskStateOp.MesosUpdate(taskStatus, now) =>
```
kolloch: make it more obvious that we assume RUNNING state here...

kolloch: We could also get other states here when reconciling. The code than ignores the updates, which should be mostly fine but I am not certain. Not sure if we should change something here...

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -47,82 +45,12 @@ private[tracker] object TaskOpProcessorImpl {
...
-        case Some(updated) => Action.Update(updated)
-        case None =>
-          log.debug(s"Ignoring status update for ${taskState.taskId}. Status did not change.")
-          Action.Noop
+      val change = task.update(TaskStateOp.MesosUpdate(MarathonTaskStatus(statusUpdate), clock.now()))
+      change match {
+        case TaskStateChange.Update(updatedTask) => Action.Update(updatedTask)
+        case TaskStateChange.Expunge             => Action.Expunge
+        case TaskStateChange.NoChange            => Action.Noop
+        case TaskStateChange.Failure(_)          => Action.Noop
```
kolloch: Why not map to Action.Failure? And why mapping this at all and not just use Action. ...

kolloch: Ok, we should disallow `Action.UpdateStatus` on this level -- kind of get it.

Let's keep Action/TaskStateChange for now to not bloat the PR.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
```diff
@@ -113,23 +109,21 @@ class TaskOpFactoryImpl @Inject() (
...
-              ),
-              networking = Task.HostPorts(ports)
-            )
-          )
-        )
-        taskOperationFactory.launch(taskInfo, newTask)
+          networking = Task.HostPorts(ports))
+
+        // FIXME (3221): something like reserved.launch(...): LaunchedOnReservation so we don't need to match?
+        task.update(launch) match {
```
kolloch: your suggestion sounds good ;) Part of your refactoring was actually supposed to get rid of these cases.

But you can defer that.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -77,50 +90,154 @@ object TaskSerializer {
...
-    builder.setId(taskState.taskId.idString)
+    (taskId, agentInfo, reservationOpt, reservedStateOpt, launchedOpt) match {
 
-    import taskState.agentInfo
-    builder.setHost(agentInfo.host)
-    agentInfo.agentId.foreach { agentId =>
-      builder.setSlaveId(MesosProtos.SlaveID.newBuilder().setValue(agentId))
-    }
-    builder.addAllAttributes(agentInfo.attributes.asJava)
+      case (_, _, Some(reservation), None, Some(launched)) =>
```
kolloch: why match against `taskId` and `agentInfo` if you don't use it then in the match?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -77,50 +90,154 @@ object TaskSerializer {
...
-    val builder = Protos.MarathonTask.newBuilder()
+  private[this] def constructTask(
+    taskId: Task.Id,
+    agentInfo: Task.AgentInfo,
+    reservationOpt: Option[ReservationWithVolumes],
+    reservedStateOpt: Option[Task.Reserved.State],
+    launchedOpt: Option[Task.Launched]): Task = {
 
-    builder.setId(taskState.taskId.idString)
+    (taskId, agentInfo, reservationOpt, reservedStateOpt, launchedOpt) match {
```
kolloch: also here I do not like the splitting into reservationOpt and reservedStateOpt.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -77,50 +90,154 @@ object TaskSerializer {
...
+      case launchedOnR: Task.LaunchedOnReservation =>
+        setLaunched(launchedOnR.appVersion, launchedOnR.status, launchedOnR.networking)
+        setReservation(launchedOnR.reservation)
     }
 
     builder.build()
   }
+}
+
+object ReservedStateSerializer {
```
kolloch: I like this. It can be `private[impl]` though.

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
```diff
@@ -263,12 +262,39 @@ object MarathonTestHelper {
 
   def mininimalTask(appId: PathId): Task = mininimalTask(Task.Id.forApp(appId).idString)
   def mininimalTask(taskId: Task.Id): Task = mininimalTask(taskId.idString)
-  def mininimalTask(taskId: String): Task = {
-    Task(
+  def mininimalTask(taskId: String, now: Timestamp = clock.now()): Task = {
```
kolloch: What's the difference to `minimalLaunchedTask` now?

src/test/scala/mesosphere/marathon/api/v2/json/MarathonTaskFormatTest.scala
```diff
@@ -17,23 +17,27 @@ class MarathonTaskFormatTest extends MarathonSpec {
...
       taskId = Task.Id("/foo/bar"),
       agentInfo = Task.AgentInfo("agent1.mesos", Some("abcd-1234"), Iterable.empty),
-      reservationWithVolumes = None,
-      launched = Some(Task.Launched(time, Task.Status(time, None), NetworkInfoList(network))))
+      appVersion = time,
+      status = Task.Status(time, None),
+      networking = NetworkInfoList(network))
 
-    val taskWithLocalVolumes = new Task(
+    val taskWithLocalVolumes = new LaunchedOnReservation(
```
kolloch: `Task.`

src/test/scala/mesosphere/marathon/tasks/TaskOpFactoryImplTest.scala
```diff
@@ -159,11 +155,9 @@ class TaskOpFactoryImplTest extends MarathonSpec with GivenWhenThen with Mockito
 
     def normalApp = MTH.makeBasicApp()
     def residentApp = MTH.appWithPersistentVolume()
-    def normalLaunchedTask(appId: PathId, volumeIds: LocalVolumeId*) = MTH.mininimalTask(appId).copy(launched = Some(MTH.taskLaunched))
-    def residentReservedTask(appId: PathId, volumeIds: LocalVolumeId*) =
-      MTH.residentReservedTask(appId, volumeIds: _*)
-    def residentLaunchedTask(appId: PathId, volumeIds: LocalVolumeId*) =
-      MTH.residentReservedTask(appId, volumeIds: _*).copy(launched = Some(MTH.taskLaunched))
+    def normalLaunchedTask(appId: PathId) = MTH.mininimalLaunchedTask(appId.toString)
```
kolloch: BTW, I do not think that these forwarding methods help. Just added noise, in my opinion.

src/test/scala/mesosphere/mesos/PersistentVolumeMatcherTest.scala
```diff
@@ -108,6 +102,7 @@ class PersistentVolumeMatcherTest extends MarathonSpec with GivenWhenThen with M
 
   class Fixture {
     def makeTask(appId: PathId) = MarathonTestHelper.mininimalTask(appId)
+    def makeTask(appId: PathId, reservation: ReservationWithVolumes) = MarathonTestHelper.minimalReservedTask(appId, reservation)
```
kolloch: I would get rid of this forwarding methods... I just need to hold more info in my short term memory to understand the tests (`makeTask(appId) = MarathonTestHelper.minimalTask(appId)`)

kolloch:
Thanks! I like it. Some comments.

src/main/proto/marathon.proto
meichstedt: Fair point – The `ReservedState` is only a means of expressing the state of a `Reserved` task.

The way it is modeled now, a resident task is either `LaunchedOnReservation`, or `Reserved` - the latter one can be in one of the states `New` (has never been launched), `Suspended` (the mesos task terminated), `Garbage` (unreserve/destroy, unused) and `Unknown` (unused; could be created if we receive unknown reservations/volumes and want to timeout before destroying them).

It's not easy to find a proper representation of the scala model here, but I'm open to discuss this!

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: Sure thing!

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: makes sense, I moved it below the Task implementations

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: Absolutely, fixed that

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: ack

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: ack

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: ack

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: ack

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
meichstedt: right, removed the comment here

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
meichstedt: Yes, it was just nice to use the wrapper and its unapply method in order to match terminal or running statuses

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
meichstedt: Yes, the problem here is that I have a `Task` and not a reserved, so it comes back to either matching or fixing the call hierarchy. I will tackle this during task update serialization. (all `FIXME(3221)` should have vanished when finishing 3221 …)

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
meichstedt: Ok, I'll rename it to ReservationTimeout for the time being and we'll see whether we apply this kind of timeout to the rest as well.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: Yes, I'll make this part of the task update serialization changes

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
meichstedt: Good catch, I removed those from the match

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
meichstedt: removed it!

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: 1. I realized that I moved the old functionality into the `newHealthStatus` function. That included changes in state. I renamed it to `updatedHealthOrState` in order to reflect that
2. The behavior now is that for each task, regardless of its mesos `state`, the `TaskStatus` will be updated if it changed. So we don't assume the task is running, we'll update the status anyhow.

meichstedt:
Thanks @kolloch for your comments – I hope I've addressed them all either by a reply or a fix.

The two points I'm deferring to the upcoming task update serialization changes are
- consolidation of `TaskStateOp` and `TaskOpProcessor.Action`
- the related FIXME in `TaskOpFactoryImpl.launchOnReservation`

kolloch:
```
            <____|.----||
                        .---''---,
                         ;..__..'    _...
                       ,'/  ;|/..--''    \
                     ,'_/.-/':            :
                _..-'''/  /  |  \    \   _|/|
               \      /-./_ \;   \    \,;'   \
               ,\    / \:  `:\    \   //    `:`.
             ,'  \  /-._;   | :    : ::    ,.   .
           ,'     ::   /`-._| |    | || ' :  `.`.)
        _,'       |;._:: |  | |    | `|   :    `'
      ,'   `.     /   |`-:_ ; |    |  |  : \
      `--.   )   /|-._:    :          |   \ \
         /  /   :_|   ;`-._;   __..--';    : :
        /  (    ;|;-./_  _/.-:'o |   /     ' |
       /  , \._/_/_./--''/_|:|___|_,'        |
      :  /   `'-'--'----'---------'          |
      | :     O ._O   O_. O ._O   O_.      ; ;
      : `.      //    //    //    //     ,' /
    ~~~`.______//____//____//____//_______,'~
              //    //~   //    //
       ~~   _//   _//   _// ~ _//     ~
     ~     / /   / /   / /   / /  ~      ~~
          ~~~   ~~~   ~~~   ~~~
```

Thanks to http://www.chris.com/ascii/index.php?art=transportation/nautical

PR #3415, opened 2016-03-07
Fixes #3413 - replaces predicate validation with simpler representation.

### LINKED ISSUE #3413: Validation of predicates
There is an Accord method for wrapping simple predicate function validation, yielding a RuleViolation. Should be used where appropriate.

alexanderweber:
@aquamatthias you had your thoughts on simpler representation of if statements, as we use it quite often in our validation. Where is a nice way by using `BaseValidator[T]` and `ViolationBuilder` in order to achieve that. There is no need for writing `Failure(Set(RuleValidation(v, ..., None)))` and `Success` using this method.

Have a look at a couple of changed files and let me know what you think.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -14,6 +14,7 @@ import scala.util.Try
 object Validation {
 
   def validate[T](t: T)(implicit validator: Validator[T]): Result = validator.apply(t)
+
   def validateOrThrow[T](t: T)(implicit validator: Validator[T]): T = validate(t) match {
     case Success    => t
     case f: Failure => throw new ValidationFailedException(t, f)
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -201,12 +203,9 @@ object Validation {
...
-    new Validator[T] {
-      override def apply(t: T): Result = {
-        if (config.forall(AllConf.suppliedOptionNames)) Success else Failure(Set(RuleViolation(t,
-          s"""You have to supply ${config.mkString(", ")} on the command line.""", None)))
-      }
-    }
-  }
+  def configValueSet[T <: AnyRef](config: String*): Validator[T] =
+    new BaseValidator[T](
+      t => config.forall(AllConf.suppliedOptionNames),
```
aquamatthias: do you need `t` can this be `_`?

alexanderweber: you are right, I did change this

alexanderweber:
@aquamatthias I did write a class called `RuleValidator`, which only takes a string as a second parameter in contrast to Accord's `BaseValidator`.

aquamatthias:
Hey @alexanderweber like your approach! Made a suggestion directly in your branch.

alexanderweber:
@aquamatthias let's go with your last commit. Though it differs from the Accord's suggestion, I find it a lot easier to read. Should I squash and merge with master?

aquamatthias:
@alexanderweber can you say why this is against Accords suggestion?
Please rebase and squash.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -201,12 +202,21 @@ object Validation {
...
+    }
+
+  def isTrue[T](constraint: String)(test: T => Boolean): Validator[T] = new Validator[T] {
+    import ViolationBuilder._
+    override def apply(value: T): Result = {
+      if (test(value)) Success else RuleViolation(value, constraint, None)
     }
   }
+
+  class RuleValidator[T](val test: T => Boolean, val constraint: String) extends Validator[T] {
```
aquamatthias: I think you can also remove this class?

alexanderweber: Sure, your `isTrue` method is identical in terms of creating a `Vlidator` instance with modified apply method. RuleValidator did the same thing. I did delete it and am squashing the whole thing right now.

alexanderweber:
@aquamatthias rebased and squashed. Waiting for the test to turn green.

aquamatthias:
Thanks!

PR #3437, opened 2016-03-09
Fixes #3438 - Remove spurious reservations/volumes

### LINKED ISSUE #3223: [DEFER] Resident Tasks: Conservatively remove spurious volumes/reservations
We need a conservative housekeeping mechanism that prevents unwanted destroy/unreserve operations. State of discussion is
- Nothing is removed implicitly to prevent faulty removal and resulting data loss
- When a task terminates (due to whatever reason), reservations and volumes are retained
- When an app is deleted, reservations and volumes are removed.
- The deployment action for stopping the app terminates only after all reservations and volumes and tasks have been removed.
- Whenever an offer is received for a task that's marked, Marathon will schedule unreserve/destroy operations **and remove the task state**
- Whenever **unknown** reservations/volumes are received, Marathon will either
  - record these and provide info about them, **or**
  - log a warning that unknown reservations have been received

We might need either an extra endpoint to wipe tasks by ID.

### LINKED ISSUE #3438: Resident Tasks: Aggresively remove spurious volumes/reservations
While we probably want to be more conservative in the future with deleting state (see #3223), we first implement a rather simple strategy:

All reservations for unknown tasks or destroyed apps are cleaned up.

kolloch:
WIP, Missing:

* Integration Test(s)
* Test for StopOnFirstMatchingOfferMatcher

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -10,13 +10,15 @@ import mesosphere.marathon.core.flow.FlowModule
 import mesosphere.marathon.core.launcher.{ TaskOpFactory, LauncherModule }
 import mesosphere.marathon.core.launchqueue.LaunchQueueModule
 import mesosphere.marathon.core.leadership.LeadershipModule
+import mesosphere.marathon.core.matcher.base.util.StopOnFirstMatchingOfferMatcher
 import mesosphere.marathon.core.matcher.manager.OfferMatcherManagerModule
+import mesosphere.marathon.core.matcher.reconcile.OfferMatcherReconciliationModule
 import mesosphere.marathon.core.plugin.PluginModule
 import mesosphere.marathon.core.task.bus.TaskBusModule
 import mesosphere.marathon.core.task.jobs.TaskJobsModule
 import mesosphere.marathon.core.task.tracker.TaskTrackerModule
 import mesosphere.marathon.metrics.Metrics
-import mesosphere.marathon.state.{ AppRepository, TaskRepository }
+import mesosphere.marathon.state.{ GroupRepository, AppRepository, TaskRepository }
 import mesosphere.marathon.{ LeadershipAbdication, MarathonConf, MarathonSchedulerDriverHolder }
 
 import scala.util.Random
@@ -36,6 +38,7 @@ class CoreModuleImpl @Inject() (
     actorSystem: ActorSystem,
     marathonSchedulerDriverHolder: MarathonSchedulerDriverHolder,
     appRepository: AppRepository,
+    groupRepository: GroupRepository,
     taskRepository: TaskRepository,
     taskOpFactory: TaskOpFactory,
     leaderInfo: LeaderInfo,
@@ -64,6 +67,17 @@ class CoreModuleImpl @Inject() (
     leadershipModule
   )
 
+  private[this] lazy val offerMatcherReconcilerModule =
+    new OfferMatcherReconciliationModule(
+      marathonConf,
+      clock,
+      actorSystem.eventStream,
+      taskTrackerModule.taskTracker,
+      groupRepository,
+      offerMatcherManagerModule.subOfferMatcherManager,
+      leadershipModule
+    )
+
   override lazy val launcherModule = new LauncherModule(
     // infrastructure
     clock, metrics, marathonConf,
@@ -73,7 +87,11 @@ class CoreModuleImpl @Inject() (
     marathonSchedulerDriverHolder,
 
     // internal core dependencies
-    offerMatcherManagerModule.globalOfferMatcher)
+    StopOnFirstMatchingOfferMatcher(
+      offerMatcherReconcilerModule.offerMatcherReconciler,
+      offerMatcherManagerModule.globalOfferMatcher
+    )
+  )
 
   override lazy val appOfferMatcherModule = new LaunchQueueModule(
     marathonConf,
@@ -102,10 +120,17 @@ class CoreModuleImpl @Inject() (
   flowActors.refillOfferMatcherManagerLaunchTokens(
     marathonConf, taskBusModule.taskStatusObservables, offerMatcherManagerModule.subOfferMatcherManager)
 
+  /** Combine offersWanted state from multiple sources. */
+  private[this] lazy val offersWanted =
+    offerMatcherManagerModule.globalOfferMatcherWantsOffers
+      .combineLatest(offerMatcherReconcilerModule.offersWantedObservable)
+      .map { case (managerWantsOffers, reconciliationWantsOffers) => managerWantsOffers || reconciliationWantsOffers }
+
   lazy val maybeOfferReviver = flowActors.maybeOfferReviver(
     clock, marathonConf,
     actorSystem.eventStream,
-    offerMatcherManagerModule.globalOfferMatcherWantsOffers, marathonSchedulerDriverHolder)
+    offersWanted,
+    marathonSchedulerDriverHolder)
 
   // GREEDY INSTANTIATION
   //
@@ -121,4 +146,5 @@ class CoreModuleImpl @Inject() (
   maybeOfferReviver
   offerMatcherManagerModule
   launcherModule
+  offerMatcherReconcilerModule.start()
 }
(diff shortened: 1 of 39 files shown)


src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -73,7 +87,11 @@ class CoreModuleImpl @Inject() (
     marathonSchedulerDriverHolder,
 
     // internal core dependencies
-    offerMatcherManagerModule.globalOfferMatcher)
+    StopOnFirstMatchingOfferMatcher(
+      offerMatcherReconcilerModule.offerMatcherReconciler,
+      offerMatcherManagerModule.globalOfferMatcher
```
aquamatthias: I understand that you want to stop on the first match for the reconciler.
Not 100% sure, why the same logic applies to the global offer matcher?

src/main/scala/mesosphere/marathon/core/launcher/TaskOp.scala
```diff
@@ -37,13 +40,65 @@ object TaskOp {
...
   }
 
+  case class UnreserveAndDestroyVolumes(
+      taskId: Task.Id,
+      maybeNewTask: Option[Task] = None,
+      resources: Iterable[MesosProtos.Resource],
+      oldTask: Option[Task] = None) extends TaskOp {
+
+    override lazy val offerOperations: Iterable[MesosProtos.Offer.Operation] = {
+      val (withDisk, withoutDisk) = resources.partition(_.hasDisk)
```
aquamatthias: Do we need to filter resources that belong to us, or is this list already filtered?

src/main/scala/mesosphere/marathon/core/launcher/TaskOp.scala
```diff
@@ -3,20 +3,20 @@ package mesosphere.marathon.core.launcher
...
   */
 sealed trait TaskOp {
   /** The ID of the affected task. */
-  def taskId: Task.Id = newTask.taskId
+  def taskId: Task.Id
   /** The MarathonTask state before this operation has been applied. */
   def oldTask: Option[Task]
   /** The MarathonTask state after this operation has been applied. */
-  def newTask: Task
+  def maybeNewTask: Option[Task]
```
aquamatthias: Please add a comment, why this is an option

src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
```diff
@@ -121,10 +121,15 @@ private[launcher] class OfferProcessorImpl(
   private[this] def saveTasks(ops: Seq[TaskOpWithSource], savingDeadline: Timestamp): Future[Seq[TaskOpWithSource]] = {
     def saveTask(taskOpWithSource: TaskOpWithSource): Future[Option[TaskOpWithSource]] = {
       val taskId = taskOpWithSource.taskId
-      log.info("Save task [{}]", taskOpWithSource.taskId)
-      taskCreationHandler
-        .created(taskOpWithSource.op.newTask)
-        .map(_ => Some(taskOpWithSource))
+      val persistedOp = taskOpWithSource.op.maybeNewTask match {
+        case Some(newTask) =>
```
aquamatthias: Please add comments for the 2 cases you handle here

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
```diff
@@ -1,17 +1,27 @@
...
 import mesosphere.marathon.core.task.Task
+import org.apache.mesos.Protos.Label
 import org.apache.mesos.{ Protos => MesosProtos }
 
+import scala.collection.mutable
+
 object TaskLabels {
   private[this] final val TASK_ID_LABEL = "marathon_task_id"
 
+  def taskIdForResource(resource: MesosProtos.Resource): Option[Task.Id] = {
```
aquamatthias: Perhaps it make sense to have a separate object for resource labels?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -350,8 +350,15 @@ private class AppTaskLauncherActor(
       }
 
       val taskId = taskOp.taskId
-      tasksMap += taskId -> taskOp.newTask
-      inFlightTaskOperations += taskId -> None
+      taskOp.maybeNewTask match {
+        case Some(newTask) =>
+          tasksMap += taskId -> newTask
+          // Why?
```
aquamatthias: And if yes: how many?

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
kolloch: Oh, that's basically nesting. So if the reconciler matches, take that result. Otherwise just proceed to the globalMatcher as before. The internal logic of the globalOfferMatcher is not effected -- it encapsulates all the other matchers.

src/main/scala/mesosphere/marathon/core/launcher/TaskOp.scala
kolloch: yes, the list is already filtered. I will add a "require" to make that self-evident.

src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
kolloch: I would prefer to extend the logging statements if that's ok.

What information is missing for you?

What do you think of

```scala
      val persistedOp = taskOpWithSource.op.maybeNewTask match {
        case Some(newTask) =>
          log.info(
            "Save {} after applying the effects of {}",
            taskOpWithSource.taskId, taskOpWithSource.op.getClass.getSimpleName
          )
          taskCreationHandler.created(newTask)
        case None =>
          log.info("Remove {} because of {}", taskOpWithSource.taskId, taskOpWithSource.op.getClass.getSimpleName)
          taskCreationHandler.terminated(taskId)
      }
```

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
kolloch: I don't get "how many" but I also do not get the code below the "Why?" but I didn't change it.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
kolloch: yeah

kolloch:
Tried to reply to your comments in an extra commit.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
aquamatthias: Do you think it makes sense to also filter by framework name?
If people start more than 1 marathon on the same role, we would fight each other.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
```diff
@@ -1,24 +1,18 @@
...
 object TaskLabels {
   private[this] final val TASK_ID_LABEL = "marathon_task_id"
 
-  def labelsForTask(task: Task): Map[String, String] = labelsForTask(task.taskId)
-  def labelsForTask(taskId: Task.Id): Map[String, String] = Map(TASK_ID_LABEL -> taskId.idString)
-
-  def mesosLabelsForTask(taskId: Task.Id): MesosProtos.Labels = {
-    mesosLabels(labelsForTask(taskId))
+  def taskIdForResource(resource: MesosProtos.Resource): Option[Task.Id] = {
+    ResourceLabels(resource).labels.get(TASK_ID_LABEL).map(Task.Id(_))
```
aquamatthias: Are we sure that this framework made the reservation?
If people start more than 1 marathon on the same role, we would fight each other.

aquamatthias:
Looks very good. Nice tests. You wanted to add an integration tests.

kolloch:
(the first commit corresponds to what you already reviewed)

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
kolloch: That is a very good point that I haven't considered yet. Let's talk about it.

kolloch: I created #3471

aquamatthias: Thanks

PR #3450, opened 2016-03-10
Fixes #3439 - relative paths in dependencies validation.

### LINKED ISSUE #3439: Relative paths in dependencies
@aquamatthias I did a comparison of model validation and came to following conclusion.

~~The old dependency validation approach could only have worked for relative paths, which where relative to the app's branch **only**. That means an app defined in `/group1/group1-1/group1-1-3/app` could only be validated for dependencies defined for one of these paths: `/group1/group1-1/group1-1-3`, `/group1/group1-1`, `/group1`. That is due to the missing information of neighbourhood branches. Meaning having no information about `/group1/group1-1/group1-1-2`, /group1/group1-1/group1-1-1` and s.o.~~

~~I am about to fix that issue by gathering the information about transitive app paths at root group. What I am concerned about is the already existing structure, that the root group has no clue about. E.g. a newly created group `/existingGroup/newGroup/` has no information about other children lying in `/existingGroup/`. So it has no chance to check if newly defined relative dependencies in apps really have their correspondence in already existing structures. That concern also applies to validation of cyclic dependencies. Am I missing something obvious in order to solve that inconvenience?~~

Relative paths should be resolvable as it was before. There is no need in checking for the existence of the apps or groups though.

alexanderweber:
Dependencies have to be checked concerning all transitive IDs. This PR adds required validation.

@aquamatthias I am having troubles passing the integration test [GroupDeployIntegrationTest.scala#L35](https://github.com/mesosphere/marathon/blob/6f26274d4b16c97218c383f2deb7d1e9987cbc35/src/test/scala/mesosphere/marathon/integration/GroupDeployIntegrationTest.scala#L35), getting following error:
```Json
{
  "message" : "Object is not valid",
  "details" : [ {
    "path" : "/groups(0)/groups(0)/dependencies(0)",
    "errors" : [ "dependency definition does not exist in group's tree" ]
  } ]
}
```
which means, that the dependency `/test` defined in group `/test2` can not find the created `/test` group. I did try the same constellation through the API and it works fine there. I might come up with a solution tomorrow, but if you have a quick hint, I would not mind as well.

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -210,17 +210,26 @@ object Group {
   def defaultDependencies: Set[PathId] = Set.empty
   def defaultVersion: Timestamp = Timestamp.now()
 
-  implicit val groupValidator: Validator[Group] = validator[Group] { group =>
-    group.id is valid
-    group.apps is valid
-    group.groups is valid
-    group is noAppsAndGroupsWithSameName
-    (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
+  private def validNestedGroup(base: PathId): Validator[Group] =
+    validator[Group] { group =>
+      group.id is validPathWithBase(base)
+      group.apps is every(AppDefinition.validNestedAppDefinition(base))
+      group is noAppsAndGroupsWithSameName
+      (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
+      group is validPorts
 
-    group is validPorts
+      group.dependencies is every(validPathWithBase(base))
+      group.groups is every(valid(validNestedGroup(base)))
+    }
+
+  implicit val validRootGroup: Validator[Group] = new Validator[Group] {
+    override def apply(group: Group): Result = {
+      validate(group)(validator =
+        validNestedGroup(PathId.empty))
+    }
   }
 
-  def validWithConfig(maxApps: Option[Int])(implicit validator: Validator[Group]): Validator[Group] = {
+  def validGroupWithConfig(maxApps: Option[Int])(implicit validator: Validator[Group]): Validator[Group] = {
     new Validator[Group] {
       override def apply(group: Group): Result = {
         maxApps.filter(group.transitiveApps.size > _).map { num =>
--- a/src/main/scala/mesosphere/marathon/state/PathId.scala
+++ b/src/main/scala/mesosphere/marathon/state/PathId.scala
@@ -1,5 +1,6 @@
 package mesosphere.marathon.state
 
+import mesosphere.marathon.api.v2.Validation.isTrue
 import mesosphere.marathon.plugin
 
 import scala.language.implicitConversions
@@ -101,41 +102,40 @@ object PathId {
   private[this] val ID_PATH_SEGMENT_PATTERN =
     "^(([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])\\.)*([a-z0-9]|[a-z0-9][a-z0-9\\-]*[a-z0-9])|(\\.|\\.\\.)$".r
 
-  /**
-    * For external usage. Needed to overwrite the whole description, e.g. id.path -> id.
-    * @return
-    */
-  implicit def pathIdValidator: Validator[PathId] = {
-    new Validator[PathId] {
-      override def apply(pathId: PathId): Result = {
-        validate(pathId.path)(validator = pathId.path.each should matchRegexFully(ID_PATH_SEGMENT_PATTERN.pattern)) and
-          validChild(pathId)
-      }
+  private val validPathChars = new Validator[PathId] {
+    override def apply(pathId: PathId): Result = {
+      validate(pathId.path)(validator = pathId.path.each should matchRegexFully(ID_PATH_SEGMENT_PATTERN.pattern))
     }
   }
 
   /**
-    * Check if path's parent, if it exists, is a valid parent indeed.
-    * @return
+    * For external usage. Needed to overwrite the whole description, e.g. id.path -> id.
     */
-  private def validChild: Validator[PathId] = {
-    new Validator[PathId] {
-      override def apply(pathId: PathId): Result = {
-        if (pathId.parent == "".toPath) Success
-        else if (pathId.parent.absolute) {
-          val p = pathId.canonicalPath(pathId.parent)
-          if (pathId.parent != PathId.empty && p.parent != pathId.parent) Failure(Set(
-            RuleViolation(pathId,
-              s"""Identifier $pathId is not child of ${pathId.parent}.
-                    |Actual parent: ${p.parent}.
-                    |Hint: use relative paths.""".
-                stripMargin, None)
-          )
-          )
-          else Success
-        }
-        else Failure(Set(RuleViolation(pathId.parent, "Path of parent should be absolute.", None)))
-      }
+  implicit val pathIdValidator = validator[PathId] { path =>
+    path is childOf(path.parent)
+    path is validPathChars
+  }
+
+  /**
+    * Validate path with regards to some parent path.
+    * @param base Path of parent.
+    */
+  def validPathWithBase(base: PathId): Validator[PathId] = validator[PathId] { path =>
+    path is childOf(base)
+    path is validPathChars
+  }
+
+  private def childOf(parent: PathId): Validator[PathId] = {
+    isTrue[PathId](s"Identifier is not child of $parent. Hint: use relative paths.") { child =>
+      parent == PathId.empty || !parent.absolute ||
+        (parent.absolute && child.canonicalPath(parent).parent == parent)
     }
+  }
+
+  /**
+    * Needed for AppDefinitionValidatorTest.testSchemaLessStrictForId.
+    */
+  val absolutePathValidator = isTrue[PathId]("Path needs to be absolute") { path =>
+    path.absolute
   }
 }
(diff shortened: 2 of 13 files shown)


aquamatthias:
@alexanderweber one point to note: we allow the definition of dependencies to groups/apps that are not existent. Debatable if this is good. The validation only need to check, if it can resolve the path. The circular dependency checker will do the rest.

alexanderweber:
@aquamatthias looks like I went too far with the validation here. I did change it to resolve-only solution.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -210,14 +210,23 @@ object Group {
...
+      group is validPorts
+
+      group.dependencies is every(pathAndBaseValidator(base))
+      group.groups is every(valid(groupValidatorWithBase(base)))
+    }
+
+  implicit val groupValidator: Validator[Group] = new Validator[Group] {
+    override def apply(group: Group): Result = {
+      com.wix.accord.validate(group)(validator =
+        groupValidatorWithBase(PathId.empty))
```
aquamatthias: This is correct only for the root group - correct? 
Should probably also `group.id.parent`?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -210,14 +210,23 @@ object Group {
...
-    group is noAppsAndGroupsWithSameName
-    (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
-
-    group is validPorts
+  private def groupValidatorWithBase(base: PathId): Validator[Group] =
+    validator[Group] { group =>
+      group.id is pathAndBaseValidator(base)
+      group.apps is every(AppDefinition.nestedAppDefinitionValidator(base))
+      group is noAppsAndGroupsWithSameName
+      (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
```
aquamatthias: should be defined in `groupValidator`?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -210,14 +210,23 @@ object Group {
...
+      group is validPorts
```
aquamatthias: should be defined in `groupValidator`?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -210,14 +210,23 @@ object Group {
...
 
-  implicit val groupValidator: Validator[Group] = validator[Group] { group =>
-    group.id is valid
-    group.apps is valid
-    group.groups is valid
-    group is noAppsAndGroupsWithSameName
-    (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
-
-    group is validPorts
+  private def groupValidatorWithBase(base: PathId): Validator[Group] =
```
aquamatthias: Perhaps rename according to AppDefinition: `nestedGroupValidator`?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -210,14 +210,23 @@ object Group {
...
+      (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
+      group is validPorts
+
+      group.dependencies is every(pathAndBaseValidator(base))
+      group.groups is every(valid(groupValidatorWithBase(base)))
+    }
+
+  implicit val groupValidator: Validator[Group] = new Validator[Group] {
+    override def apply(group: Group): Result = {
+      com.wix.accord.validate(group)(validator =
```
aquamatthias: Why the qualified call to validate?
Our initial idea to not import `com.wix.accord` did not work out, so we have 2 definitions of `validate`. We probably should either rename or remove the one in `Validation`.

src/main/scala/mesosphere/marathon/state/Group.scala
alexanderweber: Agreed and changed

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -101,41 +102,40 @@ object PathId {
...
-    new Validator[PathId] {
-      override def apply(pathId: PathId): Result = {
-        validate(pathId.path)(validator = pathId.path.each should matchRegexFully(ID_PATH_SEGMENT_PATTERN.pattern)) and
-          validChild(pathId)
-      }
-    }
-  }
+  implicit val pathIdValidator = validator[PathId] { path =>
+    path is pathParentChildValidator(path.parent)
+  } and validChars
```
aquamatthias: Just curious: would that be the same as `path has validChars` inside the validator?

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -101,41 +102,40 @@ object PathId {
...
 
   /**
-    * Check if path's parent, if it exists, is a valid parent indeed.
+    * Validate path with regards to some parent path.
+    * @param base Path of parent.
     * @return
```
aquamatthias: Can you remove empty doc params?

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -101,41 +102,40 @@ object PathId {
...
-          )
-          else Success
-        }
-        else Failure(Set(RuleViolation(pathId.parent, "Path of parent should be absolute.", None)))
-      }
+  def pathAndBaseValidator(base: PathId): Validator[PathId] = validator[PathId] { path =>
+    path is pathParentChildValidator(base)
+  } and validChars
+
+  private def pathParentChildValidator(parent: PathId): Validator[PathId] = {
```
aquamatthias: Name suggestion: `childOf`(so the use is `app.id is childOf(parent))

src/main/scala/mesosphere/marathon/state/Group.scala
alexanderweber: You are right. How about explicit imports:
```
import com.wix.accord.{Validator, Result, RuleViolation, GroupViolation, Failure, Success}
import mesosphere.marathon.api.v2.Validation.{validate, every, isTrue}
```
Alternatively I would agree to removing the duplicate from `Validation` as we need to import `com.wix.accord` in almost any validation class.

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -101,41 +102,40 @@ object PathId {
...
-                    |Actual parent: ${p.parent}.
-                    |Hint: use relative paths.""".
-                stripMargin, None)
-          )
-          )
-          else Success
-        }
-        else Failure(Set(RuleViolation(pathId.parent, "Path of parent should be absolute.", None)))
-      }
+  def pathAndBaseValidator(base: PathId): Validator[PathId] = validator[PathId] { path =>
```
aquamatthias: Name suggestion: `pathWithBase`

src/main/scala/mesosphere/marathon/state/Group.scala
alexanderweber: So port validation is not needed in subgroups?

src/main/scala/mesosphere/marathon/state/Group.scala
alexanderweber: Cyclic dependencies validation is not needed in subgroups?

src/main/scala/mesosphere/marathon/state/Group.scala
aquamatthias: Please ignore. You are right - it is needed.

src/main/scala/mesosphere/marathon/state/Group.scala
alexanderweber: Yes, it is only for root group calls.
I had the same thought about that one too. According to old validation routines, `checkGroup` is only called with `PathId.empty` as the base for root. I could try it out with `group.id.parent` though.

src/main/scala/mesosphere/marathon/state/Group.scala
aquamatthias: I think removing makes sense.

src/main/scala/mesosphere/marathon/state/PathId.scala
alexanderweber: You are right, it would be the same result here. There are other examples, where a direct call will shadow the resolving process of the properties in the error message, but it is not the case here. I'll change it to the 'being-inside' version.

src/main/scala/mesosphere/marathon/state/PathId.scala
alexanderweber: Sure thing.

src/main/scala/mesosphere/marathon/state/PathId.scala
alexanderweber: Was never satisfied with the name either. Good call

src/main/scala/mesosphere/marathon/state/PathId.scala
alexanderweber: Agreed!

aquamatthias:
@alexanderweber Thanks for this PR. 
I added some suggestions (mainly naming). Perhaps we can remove the word validator from our validators (xxxValidator -> xxx) since it is of type Validator. Also xxx should say what it checks, rather than the type it is defined on.

alexanderweber:
@aquamatthias had same thoughts on xxxValidator -> xxx. How should bigger ones be called though? E.g. `basicAppDefinitionValidator` with 15 checks in total. `basicAppDefinition` or rather `validBasicAppDefinition`, or just keep it `basicAppDefinitionValidator`?

alexanderweber:
@aquamatthias I did address all of your points. Let me know if there is still room for improvement.

aquamatthias:
Thanks!

PR #3453, opened 2016-03-10
Adding test for negative resources app creation

jimenez:
adding test for #3193

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
@@ -60,6 +60,44 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
     JsonTestHelper.assertThatJsonString(response.getEntity.asInstanceOf[String]).correspondsToJsonOf(expected)
   }
 
+  test("Create a new app fails with Validation errors for negative resources") {
+    Given("An app with negative resources")
+    var app = AppDefinition(id = PathId("/app"), cmd = Some("cmd"),
+      versionInfo = OnlyVersion(Timestamp.zero), mem = -128)
+    var group = Group(PathId("/"), Set(app))
+    var plan = DeploymentPlan(group, group)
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -60,6 +60,20 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
     JsonTestHelper.assertThatJsonString(response.getEntity.asInstanceOf[String]).correspondsToJsonOf(expected)
   }
 
+  test("Create a new app fails with Validation errors for negative resources") {
+    Given("An app with negative resources")
+    val app = AppDefinition(id = PathId("/app"), cmd = Some("cmd"), versionInfo = OnlyVersion(Timestamp.zero), mem = -128, cpus = -1, instances = -1)
```
kolloch: Nit-pick: 

* not only the resources are negative, it would be great to test one thing at a time.
* it would be nice to break the line here because e.g. in github you cannot see the whole thing without scrolling.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -60,6 +60,20 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+    val group = Group(PathId("/"), Set(app))
+    val plan = DeploymentPlan(group, group)
+    val body = Json.stringify(Json.toJson(app)).getBytes("UTF-8")
+    groupManager.updateApp(any, any, any, any, any) returns Future.successful(plan)
+    groupManager.rootGroup() returns Future.successful(group)
+
+    Then("A constraint violation exception is thrown")
+    val response = appsResource.create(body, false, auth.request)
+    response.getStatus should be(422)
```
kolloch: Is it possible to make the validation more specific? E.g. test for the specific error reported?

aquamatthias:
Thanks!

PR #3454, opened 2016-03-10
Based on application definition a default residency is added if not a…

aquamatthias:
…lready defined.


kolloch:
strange error on teamcity. Otherwise looks good.

aquamatthias:
Don't call your method 'default' fixed the problem. Not exactly sure why this works locally but not on TC.

kolloch:
LGTM!

```

     __/\__
  ~~~\____/~~~~~~
    ~  ~~~   ~.  
```

PR #3475, opened 2016-03-11
Fixes #3392 and #3477 - Residents Task: Use count from LaunchQueue to determine how many tasks are still needed

### LINKED ISSUE #3392: Residents Task: Use count from LaunchQueue to determine how many tasks are still needed
Use the launch queue count instead of app.instances (see `TaskOpFactoryImpl`). The AppTaskLauncherActor does not verify whether it has to launch a new task in case of a normal app – it was designed without the need for it. For resident apps, however, the taskOptFactory checks whether it should launch a task, reserve one, or do nothing. Since this diverging functionality can be confusing, it should be aligned.

kolloch:
* combine several parameters into one request object (which will soon be extended)
* use a taskMap in the case class to make comparision cheap and correct


aquamatthias:
LGTM. Thanks.

PR #3476, opened 2016-03-11
Towards #3221 serialize task updates WIP

meichstedt:
- TaskOp.newTask is now defined by a TaskStateOp
- TaskOpProcessorImpl will apply taskStateOps to the current taskState
- The TaskTracker will broadcast stateChanges on TaskUpdated and TaskRemoved
- The AppTaskLauncherActor subscribes to the TaskTracker for task updates

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/TaskOp.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/TaskOp.scala
@@ -1,8 +1,8 @@
 package mesosphere.marathon.core.launcher
 
 import mesosphere.marathon.core.launcher.impl.TaskLabels
-import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.Task.LocalVolume
+import mesosphere.marathon.core.task.{ Task, TaskStateOp }
 import mesosphere.marathon.tasks.ResourceUtil
 import org.apache.mesos.{ Protos => MesosProtos }
 
@@ -11,14 +11,11 @@ import org.apache.mesos.{ Protos => MesosProtos }
   */
 sealed trait TaskOp {
   /** The ID of the affected task. */
-  def taskId: Task.Id
+  def taskId: Task.Id = newTask.taskId
   /** The MarathonTask state before this operation has been applied. */
   def oldTask: Option[Task]
-  /**
-    * The MarathonTask state after this operation has been applied.
-    * `None` means that the associated task should be expunged.
-    */
-  def maybeNewTask: Option[Task]
+  /** The TaskStateOp that will lead to the new state after this operation has been applied. */
+  def newTask: TaskStateOp
   /** How would the offer change when Mesos executes this op? */
   def applyToOffer(offer: MesosProtos.Offer): MesosProtos.Offer
   /** To which Offer.Operations does this task op relate? */
@@ -29,12 +26,9 @@ object TaskOp {
   /** Launch a task on the offer. */
   case class Launch(
       taskInfo: MesosProtos.TaskInfo,
-      newTask: Task,
+      newTask: TaskStateOp,
       oldTask: Option[Task] = None,
       offerOperations: Iterable[MesosProtos.Offer.Operation]) extends TaskOp {
-
-    override def taskId: Task.Id = newTask.taskId
-    override def maybeNewTask: Option[Task] = Some(newTask)
 
     def applyToOffer(offer: MesosProtos.Offer): MesosProtos.Offer = {
       import scala.collection.JavaConverters._
@@ -43,22 +37,20 @@ object TaskOp {
   }
 
   case class ReserveAndCreateVolumes(
-      newTask: Task,
+      newTask: TaskStateOp.Reserve,
       resources: Iterable[MesosProtos.Resource],
       localVolumes: Iterable[LocalVolume],
-      oldTask: Option[Task] = None,
       offerOperations: Iterable[MesosProtos.Offer.Operation]) extends TaskOp {
 
-    override def taskId: Task.Id = newTask.taskId
-    override def maybeNewTask: Option[Task] = Some(newTask)
+    // if the TaskOp is reverted, there should be no old state
+    override def oldTask: Option[Task] = None
 
     override def applyToOffer(offer: MesosProtos.Offer): MesosProtos.Offer =
       ResourceUtil.consumeResourcesFromOffer(offer, resources)
   }
 
   case class UnreserveAndDestroyVolumes(
-      taskId: Task.Id,
-      maybeNewTask: Option[Task] = None,
+      newTask: TaskStateOp,
       resources: Iterable[MesosProtos.Resource],
       oldTask: Option[Task] = None) extends TaskOp {
 
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
@@ -2,7 +2,7 @@ package mesosphere.marathon.core.launcher.impl
 
 import mesosphere.marathon.core.launcher.TaskOp
 import mesosphere.marathon.core.matcher.base.util.OfferOperationFactory
-import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.{ TaskStateOp, Task }
 import mesosphere.marathon.core.task.Task.LocalVolume
 import org.apache.mesos.{ Protos => Mesos }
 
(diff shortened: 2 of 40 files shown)


meichstedt:
Hey @kolloch would be awesome to get a high level feedback on this. I can still try to break it down into smaller commits if you want, I just didn't want to spend any more time on it upfront.

(I'm open to bribe you into reviewing it with whatever :chocolate_bar: you might need)

src/main/scala/mesosphere/marathon/core/launcher/TaskOp.scala
```diff
@@ -11,14 +11,11 @@ import org.apache.mesos.{ Protos => MesosProtos }
...
+  def taskId: Task.Id = newTask.taskId
   /** The MarathonTask state before this operation has been applied. */
   def oldTask: Option[Task]
-  /**
-    * The MarathonTask state after this operation has been applied.
-    * `None` means that the associated task should be expunged.
-    */
-  def maybeNewTask: Option[Task]
+  /** The TaskStateOp that will lead to the new state after this operation has been applied. */
+  def newTask: TaskStateOp
```
kolloch: shouldn't be called "newTask" in my opinion but "taskChange"/"taskStateOp" or something like that.

meichstedt: right, I didn't want to rename it to have "less" changes and avoid conflicts.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
```diff
@@ -12,20 +12,30 @@ class TaskOpFactoryHelper(
...
-    oldTask: Option[Task] = None): TaskOp.Launch = {
+    newTask: Task.LaunchedEphemeral): TaskOp.Launch = {
 
     assume(newTask.taskId.mesosTaskId == taskInfo.getTaskId, "marathon task id and mesos task id must be equal")
 
     def createOperations = Seq(offerOperationFactory.launch(taskInfo))
 
-    TaskOp.Launch(taskInfo, newTask, oldTask, createOperations)
+    val stateOp = TaskStateOp.Create(newTask)
+    TaskOp.Launch(taskInfo, stateOp, None, createOperations)
```
kolloch: please use a named argument for `None`, otherwise it's hard to read without looking up the argument list.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -221,41 +225,39 @@ private class AppTaskLauncherActor(
...
+            maybeOfferReviver.foreach(_.reviveOffers())
+          }
+
+        case _ =>
+          // FIXME (3221): Only TaskUpdated and TaskRemoved are propagated to the taskTracker
+          // right now, so we'll only receive Update/Expunge here. It would be good to receive
+          // Failures as well along with an actualState: Option[Task] so we can update/remove
+          // the internal tasksMap. besides, it's ugly to only partly receive stateChanges
+          // (although Failure and NoChange are no changes in that sense)
+          log.debug("receiveTaskTrackerUpdate: ignoring unexpected stateChange {}", stateChange)
```
kolloch: I don't see that receiving failures/noops here makes sense. Let's discuss why you think that it does.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -340,20 +342,21 @@ private class AppTaskLauncherActor(
...
-        case None =>
-          removeTask(taskId)
+      // We will receive the updated task once it's been persisted. Before that,
+      // we can only store the possible state, as we don't have the updated task
+      // yet.
+      taskOp.newTask.possibleNewState.foreach { newState =>
+        tasksMap += taskId -> newState
+        scheduleTaskOpTimeout(taskOp)
       }
+      // FIXME (3221): remove from taskMap if none?
```
kolloch: our you can use a `require`. But do one of these things.

src/main/scala/mesosphere/marathon/core/matcher/reconcile/impl/OfferMatcherReconciler.scala
```diff
@@ -52,7 +52,10 @@ private[reconcile] class OfferMatcherReconciler(taskTracker: TaskTracker, groupR
           case (taskId, spuriousResources) if spurious(taskId) =>
             val unreserveAndDestroy =
               TaskOp.UnreserveAndDestroyVolumes(
-                taskId = taskId, oldTask = tasksByApp.task(taskId), resources = spuriousResources.to[Seq]
+                // FIXME (3221): is ForceExpunge correct here?
+                newTask = TaskStateOp.ForceExpunge(taskId),
```
kolloch: yes, for now `ForceExpunge` should be correct. Once we really use `Garbage` states and similar things, it will no be correct.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -216,16 +243,24 @@ object Task {
...
+            val changed = healthy || current.getState != update.getState
+            if (changed) {
+              Some(update)
+            }
+            else {
+              None
+            }
+          case None => Some(update)
+        }
+      case None => None
```
kolloch: It smells like a `flatMap` for `maybeUpdate`

src/main/scala/mesosphere/marathon/core/task/tracker/TaskCreationHandler.scala
```diff
@@ -1,12 +1,13 @@
...
 
-import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.TaskStateOp
 
 import scala.concurrent.Future
 
 /**
   * Notifies the [[TaskTracker]] of task creation and termination.
   */
+// FIXME (3221): The 2 functions are now basically the same and can be consolidated
```
kolloch: Do it :)

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
```diff
@@ -205,7 +205,7 @@ class LaunchQueueModuleTest
 
   after {
     verifyNoMoreInteractions(appRepository)
-    verifyNoMoreInteractions(taskTracker)
+    //    verifyNoMoreInteractions(taskTracker)
```
kolloch: accidental removal?

src/test/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActorTest.scala
```diff
@@ -409,7 +413,8 @@ class AppTaskLauncherActorTest extends MarathonSpec with GivenWhenThen {
     // we are not interested in these. We check for these in LaunchQueueModuleTest
     // Mockito.verifyNoMoreInteractions(offerMatcherManager)
     Mockito.verifyNoMoreInteractions(taskOpFactory)
-    Mockito.verifyNoMoreInteractions(taskTracker)
```
kolloch: left-over?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/ActorTaskTrackerUpdateSubscriber.scala
```diff
@@ -0,0 +1,25 @@
+package mesosphere.marathon.core.launchqueue.impl
```
kolloch: should be part of the tracker module. Everyone using an actor as a `TaskTrackerSubscriber` should use this code.

kolloch:
Looks mostly good. I have some concerns, though, and therefore didn't really go through all the tests yet.

**Order of Task Update Processing**: As discussed before, the design decision of allowing subscriptions to the `TaskTracker` is risky since you have to ensure the order of certain operations. In your current PR it is not guaranteed (though likely) that a task termination reaches the `AppTaskLauncherActor` before determining the scale change. This was the type of race condition that eventually led to the introduction of the TaskUpdate steps instead of the task status bus in the 0.11 release.

**Ensuring serialization of TaskStateOps**: Maybe I missed it but you do not guard against applying ops in the wrong order or on the incorrect starting state.

**TaskStateOp vs TaskStateChange**: I think that the `TaskStateOp` should express the intent and `TaskStateChange` the resulting change. Therefore, I would prefer `TaskStateOp.Revert` instead of `Create` if that's really intent. Similarly, I would prefer `TaskStateOp.Launch` to `Create` when we launch an ephemeral task.

Alternatively: Instead of `TaskStateOp.Revert` make all taskState ops revertable. The easiest way would be to include the prior task state in the op. This would also make "ensuring the serialization" possible. You could then either have a `TaskStateOp.revert` method returning a revert operation or have a `TaskUpdater.revert` method. Furthermore, one can argue if the revert operation in case of a persistence failure is actually the right thing to do.

I am not certain if the related code stays relevant but you have indicated various cleanup tasks in the PR. I think you should pursue them if we keep the related code, e.g.:

* consolidate `created/updated` for `TaskUpdate`
* remove tasksStatus update handling from `AppTaskLauncherActor`

Further cleanup:

* `TaskOpProcessor.Operation` does not need to include a `taskId` directly because the `TaskStateOp` already specifies that.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/core/matcher/reconcile/impl/OfferMatcherReconciler.scala
meichstedt: Ok.

PR #3479, opened 2016-03-11
Documentation for persistent volumes

sascala:
Note: screen shots to come

### DIFF
--- a/docs/docs/persistent-volumes.md
+++ b/docs/docs/persistent-volumes.md
@@ -1,0 +1,90 @@
+---
+title: Stateful Applications Using Local Persistent Volumes
+---
+
+Marathon applications lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can create a stateful application by specifying a local persistent volume.
+
+When you specify a local volume or volumes, tasks and their associated data are “pinned” to the node they are first launched on and will be relaunched on that node if they terminate. Local volumes enable stateful tasks because the task can be restarted without data loss.
+
+## Scaling your app
+
+When you scale your app instances/tasks down, the volumes associated with the destroyed tasks/app instances are detached. At this point, you may destroy the volume from the Marathon web interface. You may wish to leave it in the Detached state, however, if you think you will be scaling up your app again; the data on the volume will still be there. **Note:** If your app is deleted, any associated volumes will also be deleted.
+
+## Create an Application with Persistent Local Volumes
+
+1. Create a new Marathon application via the web interface.
+1. Click the Volumes tab.
+1. Choose the size of the volume or volumes you will use. Be sure that you choose a volume size that will fit the needs of your application; you will not be able to modify this size after you launch your application. <!-- note, I need to actually do this and see how you can choose more than one volume -->
+1. Specify the container path, from which your application will read and write data. The container path must be non-nested, e.g. `/var/lib/mysql`, but not  `../../../etc/opt`.
+1. Click Create.
+
+**Note:** After you launch your application, you will not be able to change the specifications for your local persistent volumes.
+
+<!--
+<screen shot>
+-->
+
+Alternatively, you can configure your application to use persistent local volumes in your application definition `config.json` file:
+
+```
+{
+  "id": "/sesame/samson",
+  "residency": {
+    "relaunchEscalationTimeoutSeconds": 30,
+    "taskLostBehavior": "relaunchAfterTimeout"
+  },
+  "cpus": 0.5,
+  "mem": 512.0,
+  "disk": 1024.0, // sandbox disk space
+  "instances": 2,
+  "container": {
+    "type": "MESOS",
+    "volumes": [
+      { // --> This will create a persistent volume
+        "containerPath": "/local/a",
+        "mode": "RW",
+        "persistent": {
+          "size": 1234
+        }
+      }
+    ]
+  }
+}
+```
+
+## View the Status of Your Application with Persistent Local Volumes
+
+After you have created your Marathon application, click on the “Volumes” tab of the application detail view [check that vocab] to get detailed information about your tasks/app instances and associated volumes.
+
+The Status column tells you if your task/app instance is attached to the volume or not. The task/app instance will read as “detached” if you have scaled down your application. Currently the only Operation Type available is read/write (RW).
+
+<!--
+<screen shot>
+-->
+
+Clicking on the individual volume brings you to the Volume Detail Page, where you can see information about the individual volume:
+
+<!--
+<screen shot>
+-->
+
+## The Mesos Sandbox
+
+While persistent local volumes allow your application to read and write data to a particular node or nodes, the temporary Mesos sandbox is still the target for the `stout` and `standerr` logs. To view these logs, go to the Marathon pane of the DCOS web interface.
+
+## Potential Pitfalls
+
+### Replication and Backups
+
+Because tasks are pinned to nodes, your data can be lost if the node is lost. If you need to guard against data loss, you must set up replication and backups. 
+
+### Non-unique Roles
+
+In Mesos, volumes are created for *roles,* not for frameworks. If more than one framework is registered with a particular role, the resource offer meant for Marathon may be accepted by another framework. This can result in volume creation failure. For more information, see the [Apache Mesos documentation on persistent volumes](http://mesos.apache.org/documentation/latest/persistent-volume/).
+
+
+
+
+
+
+
--- a/docs/docs/persistent-volumes.md
+++ b/docs/docs/persistent-volumes.md
@@ -86,6 +86,6 @@ Because tasks are pinned to nodes, your data can be lost if the node is lost. If
 
 ### Non-unique Roles
 
-In Mesos, volumes are created for *roles*, not for frameworks. If more than one framework is registered with a particular role, the resource offer meant for Marathon may be accepted by another framework. This can result in volume creation failure. Make sure you register Marathon with a unique role that is not used by another framework instance. This also holds true for any other Marathon instance you are running Multiple Marathon instances in HA mode do not need to have unique roles, however, because they are considered one instance.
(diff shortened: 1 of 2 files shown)


docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Stateful Applications Using Local Persistent Volumes
+---
+
+Marathon applications lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can create a stateful application by specifying a local persistent volume.
+
+When you specify a local volume or volumes, tasks and their associated data are “pinned” to the node they are first launched on and will be relaunched on that node if they terminate. Local volumes enable stateful tasks because the task can be restarted without data loss.
```
meichstedt: Add (possibly after `if they terminate.`): `Resources required to run this application instance are reserved in order to make sure that the application can actually be relaunched on a node.` or something similar. This background knowledge is imo important, not only for the following section.

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Scaling your app
+
+When you scale your app instances/tasks down, the volumes associated with the destroyed tasks/app instances are detached. At this point, you may destroy the volume from the Marathon web interface. You may wish to leave it in the Detached state, however, if you think you will be scaling up your app again; the data on the volume will still be there. **Note:** If your app is deleted, any associated volumes will also be deleted.
```
meichstedt: I'd go for `instances` instead of `instances/tasks`. Do we have a settled terminology on this? afaik the UI says `task` while the config is about `instances`.

meichstedt: It's important the user understands that not only `detached` volumes are consumed and reserved, but also all the other resources that are required to run that application instance. So if Alice scales an application that requires `{cpus: 1, mem: 1000}` and uses 1GB persistent disk up to 100 instances, 100 CPUs, 100GB mem and 100GB persistent disk are reserved throughout the cluster in order to launch these instances. If Alice then scales down to 1 instance without explicitly destroying the other 99 instances and their volumes, these resources will remain reserved and prevent other applications and frameworks from using them.

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Create an Application with Persistent Local Volumes
+
+1. Create a new Marathon application via the web interface.
+1. Click the Volumes tab.
+1. Choose the size of the volume or volumes you will use. Be sure that you choose a volume size that will fit the needs of your application; you will not be able to modify this size after you launch your application. <!-- note, I need to actually do this and see how you can choose more than one volume -->
+1. Specify the container path, from which your application will read and write data. The container path must be non-nested, e.g. `/var/lib/mysql`, but not  `../../../etc/opt`.
```
meichstedt: These are both nested. The current Mesos restriction does not allow any `/`, so your path will end up being `data` or `foo` or even `foo#*()&`*, but nothing with a slash. See related [MESOS ticket](https://mesosphere.atlassian.net/browse/CORE-274) for updates on this.

\* We were thinking about meaningful restrictions on last week's testing day

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+1. Click Create.
+
+**Note:** After you launch your application, you will not be able to change the specifications for your local persistent volumes.
+
+<!--
+<screen shot>
+-->
+
+Alternatively, you can configure your application to use persistent local volumes in your application definition `config.json` file:
```
meichstedt: It's not _the_ `config.json`. May just write `in an application definition JSON file`?

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+```
+{
+  "id": "/sesame/samson",
+  "residency": {
+    "relaunchEscalationTimeoutSeconds": 30,
+    "taskLostBehavior": "relaunchAfterTimeout"
+  },
+  "cpus": 0.5,
+  "mem": 512.0,
+  "disk": 1024.0, // sandbox disk space
```
meichstedt: Please remove the comment. Otherwise it's invalid JSON and won't work when copy/pasting and POST'ing against the API.

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+  "instances": 2,
+  "container": {
+    "type": "MESOS",
+    "volumes": [
+      { // --> This will create a persistent volume
```
meichstedt: remove comment

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+        "containerPath": "/local/a",
```
meichstedt: For the time being, please change containerPath to just `"local"` or `"data"` since we can't have nested/absolute paths

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+        }
+      }
+    ]
+  }
+}
+```
+
+## View the Status of Your Application with Persistent Local Volumes
+
+After you have created your Marathon application, click on the “Volumes” tab of the application detail view [check that vocab] to get detailed information about your tasks/app instances and associated volumes.
```
meichstedt: @pierlo-upitup Do you guys have some sort of official terminology for the UI part? Is `application detail view` or something similar used elsewhere?

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+
+The Status column tells you if your task/app instance is attached to the volume or not. The task/app instance will read as “detached” if you have scaled down your application. Currently the only Operation Type available is read/write (RW).
```
meichstedt: Nit-picking: To me, `task/app instances` reads like `task instances OR app instances` (where task instances doesn't make sense). Do you favor either one of `tasks` or `app instances`? Then I'd be happy if you can use that one everywhere :)

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+
+Clicking on the individual volume brings you to the Volume Detail Page, where you can see information about the individual volume:
+
+<!--
+<screen shot>
+-->
+
+## The Mesos Sandbox
+
+While persistent local volumes allow your application to read and write data to a particular node or nodes, the temporary Mesos sandbox is still the target for the `stout` and `standerr` logs. To view these logs, go to the Marathon pane of the DCOS web interface.
```
meichstedt: `stderr`

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Potential Pitfalls
+
+### Replication and Backups
+
+Because tasks are pinned to nodes, your data can be lost if the node is lost. If you need to guard against data loss, you must set up replication and backups. 
+
+### Non-unique Roles
+
+In Mesos, volumes are created for *roles,* not for frameworks. If more than one framework is registered with a particular role, the resource offer meant for Marathon may be accepted by another framework. This can result in volume creation failure. For more information, see the [Apache Mesos documentation on persistent volumes](http://mesos.apache.org/documentation/latest/persistent-volume/).
```
meichstedt: Please be more specific here. If Alice wants to use Local Persistent Volumes, she needs to make sure her Marathon setup registers with a distinct role that is not used by another framework instance (not even another Marathon instance if she or her company is running more than one).

Multiple Marathon instances in `ha mode` do not fall under this restriction, as they are consideres being `one`.

meichstedt:
Thanks, @sascala, please see my comments. It's still not completely clear (because not implemented)
* how the housekeeping will work/how and when volumes are destroyed
* whether absolute and/or nested containerPaths will be allowed.

I'd go for merging this after you have addressed my comments and polishing the text as soon as we have more information. Are you ok with this?

docs/docs/persistent-volumes.md
sascala: I'd actually been writing it that way because I wasn't sure which term was most appropriate.

sascala:
@meichstedt Thanks for your careful comments. As for merging: if the page won't appear on the live site before the persistent volume functionality actually exists for users, then please do if everything looks good to you. Otherwise, I'd wait on that so as not to promise functionality that doesn't exist yet.

docs/docs/persistent-volumes.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Potential Pitfalls
+
+### Replication and Backups
+
+Because tasks are pinned to nodes, your data can be lost if the node is lost. If you need to guard against data loss, you must set up replication and backups. 
+
+### Non-unique Roles
+
+In Mesos, volumes are created for *roles*, not for frameworks. If more than one framework is registered with a particular role, the resource offer meant for Marathon may be accepted by another framework. This can result in volume creation failure. Make sure you register Marathon with a unique role that is not used by another framework instance. This also holds true for any other Marathon instance you are running Multiple Marathon instances in HA mode do not need to have unique roles, however, because they are considered one instance.
```
meichstedt: I think you're missing a period between `... you are running` and `Multiple Marathon ...`.

Also, better change `they are considered one instance` to `they are meant to use the same role`, as the former may be confusing. Sorry in case I proposed that in an earlier comment.

sascala:
Thanks again, @meichstedt. I fixed the typo and made the change you suggested.

meichstedt:
Thanks, @sascala!
```
                                                       /\  /\
                                                      /  ''  \
                                                     / .o\  ..\
                                                    /.'  |\ | '.
              __                                    '    |_\|
              \ \___     .__                             |_|<o>
            .--""___\..--"/                              |_| |
        .__.|-"""..... ' /                          _____|_|_|\__
________\_______________/______________________..-'::::::::::::::::-..___
                                                 '''""""""""""""""'''
               ~       ~           ~        ~                  ~
       ~                     ~                   ~       ~
  ~                 ~                  ~                         ~
```

PR #3483, opened 2016-03-14
Raise UI version to 0.16.0

pierlo-upitup:


meichstedt:
Thanks @pierlo-upitup

Issue #3388, opened 2016-03-03
Validation on enumerations


zunder:
Validation dependent on values of enumeration unnecessarily bloat validation code. E.g.:

``` Scala
  implicit val containerValidator: Validator[Container] = {
    val volumeValidator: Validator[Volume] = new Validator[Volume] {
      override def apply(volume: Volume): Result = volume match {
        case pv: PersistentVolume => valid[PersistentVolume](PersistentVolume.persistentVolumeValidator).apply(pv)
        case dv: DockerVolume     => valid[DockerVolume](DockerVolume.dockerVolumeValidator).apply(dv)
      }
    }

    val dockerContainerValidator: Validator[Container] = validator[Container] { container =>
      container.docker is notEmpty
      container.docker.each is valid
      container.volumes is every(valid(volumeValidator))
    }

    val mesosContainerValidator: Validator[Container] = validator[Container] { container =>
      container.docker is empty
      container.volumes is every(valid(volumeValidator))
    }

    new Validator[Container] {
      override def apply(c: Container): Result = c.`type` match {
        case Mesos.ContainerInfo.Type.MESOS  => validate(c)(mesosContainerValidator)
        case Mesos.ContainerInfo.Type.DOCKER => validate(c)(dockerContainerValidator)
        case _                               => Failure(Set(RuleViolation(c.`type`, "unknown", None)))
      }
    }
```

According to [accord/issues/58](https://github.com/wix/accord/issues/58) this might be compressed like in this example:

``` Scala
  val vt = validator[Test] { t =>
    ((t.p should be == 0) and (t.n is empty)) or
    ((t.p should be == 1) and (t.n should startWith("test")))
  }
```

It is still a workaround though. A better implementation might apply with the next accord milestone.

@meichstedt regarding container validation. There is an alternative to pattern matching described in [accord/issues/58](https://github.com/wix/accord/issues/58). I tried it out and this does compile:

``` Scala
  private val volumeValidator: Validator[Volume] = new Validator[Volume] {
    override def apply(volume: Volume): Result = volume match {
      case pv: PersistentVolume => valid[PersistentVolume](PersistentVolume.persistentVolumeValidator).apply(pv)
      case dv: DockerVolume     => valid[DockerVolume](DockerVolume.dockerVolumeValidator).apply(dv)
    }
  }

  implicit val containerValidator = validator[Container] { container =>
[...]

PR #3484, opened 2016-03-14
Fixes #3412 - nicer container validation.

### LINKED ISSUE #3412: Nicer container validation
Container validation need to have a nicer structure.

alexanderweber:
@meichstedt I did try to address your concerns in terms of a nicer looking container validation. Basically `validContainer` now contains `validGeneralContainer`, `validDockerContainer` and `validMesosContainer`. Each of those can be extended if appropriate, depending on the level of validation. The match is still needed, but should be addressed in #3388 at some point. The validation for Volume is moved to its companion object. That all reduced some redundant code. Let me know if you agree with those changes.


aquamatthias:
LGTM

PR #3507, opened 2016-03-16
Opt in for task killing state (pure signalling in the moment)

aquamatthias:
Service Discovery (minuteman) want to see tasks in killing state.
This is an opt-in feature.
The state itself is already handled (ignored).

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
@@ -5,6 +5,8 @@ import java.io.{ IOException, FileInputStream }
 import com.google.protobuf.ByteString
 import mesosphere.chaos.http.HttpConf
 import org.apache.mesos.Protos.{ Credential, FrameworkInfo, FrameworkID }
+import FrameworkInfo.Capability
+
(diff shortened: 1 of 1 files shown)


kolloch:
This PR will not make the task state visible in the Marathon API since it is (how you wrote yourself), ignored. Does the capability flag have side effects besides sending TASK_KILLING status updates to Marathon?

aquamatthias:
This PR will have the effect, that the task state `TASK_KILLING` is set in mesos and available via state.json. 
Marathon should not be affected. 

>
Today: It informs Minuteman that the task is dying and not to send it traffic
In Mesos 0.29: It gives us a period for the task to die.

See https://mesosphere.atlassian.net/browse/MARATHON-850

kolloch:
So only if you enable this capability, mesos will expose that TASK_KILLING state in the state.json?

Can you explain the effect in a comment for enabling the capability, please?

aquamatthias:
Added a comment.

kolloch:
Thanks!
```
             /|~~~
           ///|
         /////|
       ///////|
     /////////|
   \==========|===/
~~~~~~~~~~~~~~~~~~~~~
```

PR #3512, opened 2016-03-16
Fixes #3423 | Report kills due to failed healthcheck.

### LINKED ISSUE #3423: Report kills due to failed health check in "Last Task Failure"
Currently (v0.15.3) Marathon can kill task due to failed health check, but it doesn't show up in "Last Task Failure" in debug tab. I propose reporting kills as task failures and showing them in debug tab.

janisz:
Health check kills are published as an event and stored as task failure.
They appear in UI showing cause of last failed health check and host
just as normal Mesos task failures.

screenshot from 2016-03-16 16 05 55

### DIFF
--- a/src/main/scala/mesosphere/marathon/event/Events.scala
+++ b/src/main/scala/mesosphere/marathon/event/Events.scala
@@ -152,6 +152,16 @@ case class HealthStatusChanged(
   timestamp: String = Timestamp.now().toString)
     extends MarathonHealthCheckEvent
 
+case class UnhealthyTaskKillEvent(
+  appId: PathId,
+  taskId: Task.Id,
+  version: Timestamp,
+  reason: String,
+  host: String,
+  slaveId: Option[String],
+  eventType: String = "unhealthy_task_kill_event",
+  timestamp: String = Timestamp.now().toString) extends MarathonHealthCheckEvent
+
 // upgrade messages
 
 sealed trait UpgradeEvent extends MarathonEvent
--- a/src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
@@ -108,6 +108,17 @@ class HealthCheckActor(
       // kill the task
       marathonSchedulerDriverHolder.driver.foreach { driver =>
         log.info(s"Send kill request for ${task.taskId} on host ${task.agentInfo.host} to driver")
+        eventBus.publish(
+          UnhealthyTaskKillEvent(
+            appId = task.appId,
+            taskId = task.taskId,
+            version = app.version,
+            reason = health.lastFailureCause.getOrElse("unknown"),
+            host = task.agentInfo.host,
+            slaveId = task.agentInfo.agentId,
+            timestamp = health.lastFailure.getOrElse(Timestamp.now()).toString
+          )
+        )
         driver.killTask(task.taskId.mesosTaskId)
       }
     }
--- a/src/main/scala/mesosphere/marathon/state/TaskFailure.scala
+++ b/src/main/scala/mesosphere/marathon/state/TaskFailure.scala
@@ -1,6 +1,7 @@
 package mesosphere.marathon.state
 
 import mesosphere.marathon.Protos
+import mesosphere.marathon.event.UnhealthyTaskKillEvent
 import mesosphere.marathon.state.PathId._
 import mesosphere.mesos.protos.Implicits.slaveIDToProto
 import mesosphere.mesos.protos.SlaveID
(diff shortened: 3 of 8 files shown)


src/main/scala/mesosphere/marathon/state/TaskFailure.scala
```diff
@@ -65,6 +66,26 @@ object TaskFailure {
       slaveId = if (proto.hasSlaveId) Some(proto.getSlaveId) else None
     )
 
+  object FromUnhealthyTaskKillEvent {
+    def unapply(event: UnhealthyTaskKillEvent): Option[TaskFailure] =
+      apply(event)
+
+    def apply(event: UnhealthyTaskKillEvent): Option[TaskFailure] = {
```
aquamatthias: I guess you use the same method signature as `FromMesosStatusUpdateEvent`.
Since there can be different status update this makes sense there, but we want to treat every kill event as task failure. 
Please remove the option from the apply method.

src/main/scala/mesosphere/marathon/state/TaskFailure.scala
```diff
@@ -65,6 +66,26 @@ object TaskFailure {
...
+      val UnhealthyTaskKillEvent(appId, taskId, version, health, host, slaveID, _, _) = event
+
+      Some(TaskFailure(
+        appId,
+        taskId.mesosTaskId,
+        mesos.TaskState.TASK_KILLED,
+        health.lastFailureCause.getOrElse(""),
```
aquamatthias: Perhaps we should always prepend something like:
`"Task was killed since health check failed. Reason: $reason"`
Also: if the failureCause is unset: `getOrElse("unknown")` would be better than an empty string.

src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
```diff
@@ -3,7 +3,7 @@ package mesosphere.marathon.health
 import akka.actor.{ Actor, ActorLogging, ActorRef, Cancellable, Props }
 import akka.event.EventStream
 import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
-import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.{ Task }
```
aquamatthias: Remove Parentheses.

src/main/scala/mesosphere/marathon/event/Events.scala
```diff
@@ -206,6 +207,16 @@ case class AppTerminatedEvent(
   eventType: String = "app_terminated_event",
   timestamp: String = Timestamp.now().toString) extends MarathonEvent
 
+case class UnhealthyTaskKillEvent(
```
aquamatthias: Please define this event in the health check section.

src/main/scala/mesosphere/marathon/event/Events.scala
```diff
@@ -206,6 +207,16 @@ case class AppTerminatedEvent(
...
+  appId: PathId,
+  taskId: Task.Id,
+  version: Timestamp,
+  health: Health,
```
aquamatthias: Why the complete health object? 
This event can only be thrown if there were consecutive failures, that are reported via `FailedHealthCheck`.

aquamatthias:
@janisz Thanks for this contribution!
I added some comments and questions.

src/main/scala/mesosphere/marathon/event/Events.scala
janisz: I thougt it will be nice to have more details with event. I'll reduce it to keep only `lastFailureCause`

src/main/scala/mesosphere/marathon/event/Events.scala
janisz: I added a note in 
* docs/docs/event-bus.md
* docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_create.md

src/main/scala/mesosphere/marathon/state/TaskFailure.scala
janisz: Good point. Fixed

src/main/scala/mesosphere/marathon/event/Events.scala
aquamatthias: 1) Cool. Please add to this PR (I do not see that change)
2) What I originally meant was even simpler: can you define that class under https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/event/Events.scala#L118

janisz: Fixed

aquamatthias:
@janisz Looks good to me. Thanks!

PR #3523, opened 2016-03-17
small edits

sascala:
@gkleiman I made some small edits to the Ports page as I read it over. Do you think it would make sense to discuss VIPs here?


sascala:
@gkleiman vip.md is a draft of a page describing the new VIP functionality. It's pretty short! Is there anything else a user should know?

PR #3525, opened 2016-03-18
WIP: external volume support

jdef:
as the title says, WORK IN PROGRESS.

changes:
- add optional name, providerName, and options fields to PersistentVolume; and make size optional
- add InjectorContext impls to allow vol providers to inject things into protos at TaskInfo build time
- add validators to vol providers: per Volume, per Container, per Group
- rename VolumeMatcher for clarity (the match is actually for reserved resources)

todo:
- [x] ? support snapshotID or snapshotName dvdcli options? (we forward all user-defined dvdi/* options)
- [x] enforce limited upgrade strategy for apps using DVDI volumes
- [x] explicitly enforce that every vol must have a provider
- [ ] user documentation (early docs here, will need an update #3607)
  - prohibited chars for option values (as enforced by EMC mesos isolator)
- [x] split provider impls in to separate files?
- [x] ensure 1:1 mapping between task and external volume (via group validator)
- [ ] [[STARTED]] add some tests for the new validators associated with name, providerName, and options
- [x] more validation (for "optional" field contents, like driverName)
- [x] refactor lots of Volumes.scala stuff into an impl/ package
- [ ] support for remaining DVDCLI options
- [ ] unit tests for all of the new code
- [ ] integration tests for all of the new code
- [ ] ... probably other things (i'll flesh out this list later)

related:
- #3645 containerPath should not container slashes for agent local volumes


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -107,7 +107,13 @@ case class AppDefinition(
   /**
     * @return the disk resources required for volumes
     */
-  def diskForVolumes: Double = persistentVolumes.map(_.persistent.size).sum.toDouble
+  def diskForVolumes: Double = persistentVolumes.map(
+    //TODO(jdef) probably only want to consider "agent" providerName here
```
jdef: this looks ripe for refactoring into a provider-specific implementation

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -85,10 +85,14 @@ class OfferOperationFactory(
           .setLabels(TaskLabels.labelsForTask(taskId).mesosLabels)
           .build()
 
+      val size = vol.persistentVolume.persistent.size match {
```
jdef: probably want to coordinate with storage provider here

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -67,10 +67,28 @@ object VolumeSerializer {
...
 
 object PersistentVolumeInfoSerializer {
-  def toProto(info: PersistentVolumeInfo): Protos.Volume.PersistentVolumeInfo =
-    Protos.Volume.PersistentVolumeInfo.newBuilder()
-      .setSize(info.size)
-      .build()
+  def toProto(info: PersistentVolumeInfo): Protos.Volume.PersistentVolumeInfo = {
+    val builder = Protos.Volume.PersistentVolumeInfo.newBuilder()
+    info.size match {
+      case Some(size) => builder.setSize(size)
```
meichstedt: `info.size.foreach(builder.setSize)`

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
meichstedt: Yes, as for we only need to reserve disk for volumes space that's provided by the agent itself

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -96,11 +105,61 @@ object DockerVolume {
...
+  *
+  * @param size absolute size of the volume (MB)
+  * @param name identifies the volume within the context of the storage provider.
+  * @param providerName identifies the storage provider responsible for volume lifecycle operations.
+  * @param options contains storage provider-specific configuration configuration
+  */
+case class PersistentVolumeInfo(
+  size: Option[Long] = None,
+  name: Option[String] = None,
+  providerName: Option[String] = None,
```
meichstedt: Not yet sure which approach is good here – going for provider-specific implementation is probably out best bet as handling an object with multiple options and prohibited combinations might become ugly quite soon.
In case of not doing that I'd at least suggest to declare name as String and set the default ("agent") when deserializing it.

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -96,11 +105,61 @@ object DockerVolume {
...
+  options: Option[Map[String, String]] = None) // = Map.empty[String, String])
```
meichstedt: (My personal choice would be the empty Map as of now)

src/main/scala/mesosphere/marathon/state/Volume.scala
jdef: Understood. I was trying to avoid littering the code with "agent" defaults (protobuf defaults, here, elsewhere...). Ideally the code would only mention "agent" in one place. I'm going to push another commit with some additional refactoring, perhaps it will influence your thinking?

src/main/scala/mesosphere/marathon/state/Volume.scala
jdef: What are the trade-offs?

src/main/scala/mesosphere/marathon/state/Volume.scala
jdef: I toyed with the idea of an "object Volumes" in "src/main/scala/mesosphere/marathon/core/volume/Volumes.scala" that would contain a registry of volume providers. a provider *could* supply a validator. so in this func we could do something like `Volumes.for(providerName).validate(volumeInfo)`. that seemed a bit complicated for what I needed in the last commit, but as soon as I add a DVDI volume provider we might want it.

jdef: for example, I removed this just prior to commit because I wasn't using it:
```scala
protected object Volumes { 
  val registry = Map[String, Volumes]()
  
  def register(vp: Volumes) = {
    registry(vp.name) = vp
  }

  def apply(name: String): Volumes = registry(name)

  // all volume providers should be registered here, statically
  register(AgentVolumes)
}
```

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -214,6 +214,78 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+
+    When("The request is processed")
+    val response = appsResource.create(body, false, auth.request)
+
+    Then("The return code indicates create error")
+    response.getStatus should be(422)
+    response.getEntity.toString should include("supply mesos_authentication_principal, mesos_role, mesos_authentication_secret_file")
+    response.getEntity.toString should not include ("/container/volumes(0)/persistent.providerName")
+  }
+
```
jdef: add a test that validates size is required when providerName == "agent" (or empty)?

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
```diff
@@ -0,0 +1,97 @@
+package mesosphere.marathon.core.volume
```
aquamatthias: All packages under core are components that contain:
- interface definitions (traits)
- impl that holds private implementation
- Module that defines the required/provided interfaces 

We are moving the code from `*` to `core/module` thats why it is probably confusing.
But moving Volumes here feels wrong.

src/main/scala/mesosphere/marathon/state/Volume.scala
aquamatthias: Can we handle mesos persistent volumes and external volumes with the same type?
We have specific use cases and validations for each case - so also having a specific type could make life easier.

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
```diff
@@ -0,0 +1,97 @@
...
+
+object AgentVolumes extends Volumes[PersistentVolume] {
+
+  val name = "agent"
+
+  val validation = validator[Volume] { v =>
+    // don't invoke validator on v because that's circular, just check the additional
+    // things that we need for agent local volumes.
+    // see implicit validator in the PersistentVolume class for reference.
+    v.asInstanceOf[PersistentVolume].persistent.size.isDefined is true
```
aquamatthias: `persistent.size.isDefined` ==> means persistent volume (or no external volume).
Make a specific type for external volumes.

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
```diff
@@ -0,0 +1,97 @@
...
+import mesosphere.marathon.WrongConfigurationException
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.state.AppDefinition
+import mesosphere.marathon.state.Volume
+import mesosphere.marathon.state.PersistentVolume
+
+/**
+  * Volumes is an interface implemented by storage volume providers
+  */
+sealed trait Volumes[T <: Volume] {
```
aquamatthias: Feels like there is a strong (1:1 ?) relation between specific `Volume` and specific `Volumes`.
Would it make sense to make the logic here part of `Volume`?

aquamatthias:
@jdef made some comments. Your idea of defining PersistentVolume (in Mesos persistent volume is local) as either local or external could lead to confusion. My suggestion would be to model the two cases specifically.

jdef:
Thanks for the feedback! I have concerns about modeling local vs external
w/ different types - I think it takes us down a slippery slope. Different
types of external volumes will have their own requirements. Do we then fork
external volumes into extA and extB types? What about local-disk vs. (some
future) local-ramdisk types? Do we fork a "local" type to represent those?

I'd much rather let a volume provider handle the differences between volume
types. A volume record spec represents some lowest common denominator of
fields that we expect to be most common across all volume types - and then
we let a validator supplied by specific provider implementations deal with
the differences. Otherwise you get an ugly propagation of types at a level
of the system where it doesn't make sense.

Do you think these concerns are unfounded?

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
jdef: I don't understand this comment. Can you elaborate?

aquamatthias:
@jdef we definitively have different requirements for local vs external volumes. From my point of view that should be reflected also in the type system (having different properties, validations, mesos proto generation etc.).

I understand your concern, if we start to think about different external volume types.
I don't have the answer to that right now - as of today we are not sure, if that will happen (and we could tackle that still in all possible ways).
Given todays requirements I would suggest handling external and local volumes different.

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
aquamatthias: Sorry for being unclear. Please see the Conventions section here: https://mesosphere.github.io/marathon/docs/core-architecture.html

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
jdef: The intent was to implement a DVDVolumes[PersistentVolume] object to handle volumes w/ persistentName="dvdi". However, I didn't want to assume that all volume providers would *always* use PersistentVolume so I used Volume as an upper-limit instead.

jdef:
Understood. Type-per-volume statically bakes the current requirements into the type system which feels fragile.I'd much rather base decisions off of providerName. That said, I'll put together a commit wherein I refactor some of this to use an ExternalVolume type and we can see how that feels.

jdef:
@aquamatthias I started hacking on a branch wherein I created an ExternalVolume and ExternalVolumeInfo types in state/Volumes.scala, updated the Volume protobuf to support an ExternalVolumeInfo field. It's smells (between the code dup and additional case statements).

But let's say that we go this way; are you advocating to break out "external" as an independent field in the JSON schema too? I think *that* is confusing: "Marathon supports persistent storage. To use it you can populate 'persistent'. Or 'external'" -- huh? Why can't I just fill out 'persistent' and provide the details on how to use the storage that I want? But let's say that we do this. The same type of questions are going to come up again when we decide to add other types of external volumes. Only we can hide some of those differences in the "options" field. So ... if we're planning to do that, then why not keep the "providerName" field as a member of "persistent" and use it the same way? At least then we're consistent about the approach (key'ing off of field contents vs. special-field-per-volume-type .. which also smells).

I'm still not sold on this type of approach. It feels like we're going to have special case treatment in a lot of places in the code and I'd rather isolate all the special casing into some (Volumes?) abstraction. And types will propagate in an undesirable way.

src/main/scala/mesosphere/marathon/core/volume/Volumes.scala
jdef: I'll move implementation specifics to an impl package. There's lots of cleanup to do here

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,78 @@
...
+    if (volumes.isEmpty) None
+    else {
+      var cc = initialContext()
+      volumes.foreach { vol => cc = apply(cc, vol).getOrElse(cc) }
+      Some(cc)
+    }
+  }
+}
+
+sealed trait Context
```
jdef: document this

jdef:
rebased

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+/**
+  * VolumeProvider is an interface implemented by storage volume providers
+  */
+trait VolumeProvider[+T <: Volume] {
+  /** name uniquely identifies this volume provider */
+  val name: String
+  /** validation implements this provider's specific validation rules */
+  val validation: Validator[Volume]
```
jdef: we could make this a func that accepts a Container spec, but how to we determine the default containerizer? I don't think the logic for that is centralized...

src/main/scala/mesosphere/marathon/core/volume/impl/Volumes.scala
```diff
@@ -0,0 +1,209 @@
...
+  /** DockerVolumes can be serialized into a Mesos Protobuf */
+  def toMesosVolume(volume: DockerVolume): MesosVolume =
+    MesosVolume.newBuilder
+      .setContainerPath(volume.containerPath)
+      .setHostPath(volume.hostPath)
+      .setMode(volume.mode)
+      .build
+
+  /** @return a possibly modified builder if `v` is a DockerVolume */
+  override protected def updated[C <: Context](context: C, v: Volume): Option[C] = {
```
jdef: who validates that DockerVolumes are only valid for docker containers?

src/main/scala/mesosphere/marathon/core/volume/impl/Volumes.scala
```diff
@@ -0,0 +1,228 @@
...
+
+import com.wix.accord._
+import com.wix.accord.combinators.{ Fail, NilValidator }
+import com.wix.accord.dsl._
+import com.wix.accord.Validator
+import mesosphere.marathon.core.volume._
+import mesosphere.marathon.state.Container
+import mesosphere.marathon.state.{ DockerVolume, PersistentVolume, Volume }
+import org.apache.mesos.Protos.{ CommandInfo, ContainerInfo, Volume => MesosVolume, Environment }
+import scala.collection.JavaConversions._
```
jdef: don't use conversions, use asScala w/ converters

src/main/scala/mesosphere/marathon/core/volume/claims/VolumeClaimsDelegate.scala
```diff
@@ -0,0 +1,139 @@
...
+  private[this] var claimsByToken: Map[ClaimToken, ClaimEntry] = Map.empty
+
+  override def receive: Receive = LoggingReceive {
+    Seq[Receive](
+      receiveClaimRequest
+    ).reduceLeft(_.orElse[Any, Unit](_))
+  }
+
+  private[this] def receiveClaimRequest: Receive = {
+    // TODO(jdef) consider an update to an existing claim (app config changed..)
```
jdef: nothing to consider along these lines here. AppTaskLauncherActor is responsible for sync'ing claims according to app config changes. this actor just processes requests that are generated from that actor.

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,104 @@
...
+    * the default VolumeProvider implementation is returned. None is returned if Some name is given
+    * but no volume provider is registered for that name.
+    */
+  def apply(name: Option[String]): Option[VolumeProvider[Volume]];
+
+  /** @return a validator that checks the validity of a volume provider name */
+  def known(): Validator[Option[String]];
+
+  /** @return a validator that checks the validity of a volume given the volume provider name */
+  def approved[T <: Volume](name: Option[String]): Validator[T];
```
jdef: this looks redundant w/ `known` - why did I ever think I needed both?

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -33,33 +36,21 @@ object Volume {
...
-          mode = proto.getMode
-        )
-      case None =>
-        DockerVolume(
-          containerPath = proto.getContainerPath,
-          hostPath = proto.getHostPath,
-          mode = proto.getMode
-        )
-    }
+  def fromProto(proto: Protos.Volume): Volume = {
```
jdef: would rather move this into the volume module and let provider implementations deal with this

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -78,14 +69,23 @@ object Volume {
 
 /**
   * A volume mapping either from host to container or vice versa.
-  * Both paths can either refer to a directory or a file.  Paths must be
+  * Both paths can either refer to a directory or a file. Paths must be
   * absolute.
+  *
+  * TODO(jdef) rename this as DockerHostVolume for clarity
```
jdef: TODO

src/main/scala/mesosphere/marathon/core/volume/providers/DockerHost.scala
```diff
@@ -0,0 +1,47 @@
...
+/**
+  * DockerHostVolumeProvider handles Docker volumes that a user would like to mount at
+  * predetermined host and container paths. Docker host volumes are not intended to be used
+  * with "non-local" docker volume drivers. If you want to use a docker volume driver then
+  * use a PersistentVolume instead.
+  */
+protected case object DockerHostVolumeProvider
+    extends ContextUpdateHelper[DockerVolume]
+    with VolumeProvider[DockerVolume] {
+  val name = "docker" // only because we should have a non-empty name
```
jdef: or maybe make `name` an Option field? but require that all "registered" providers have a name?

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
```diff
@@ -192,6 +192,18 @@
                     "type": "integer",
                     "description": "The size of the persistent volume in MiB",
                     "minimum": 0
+                  },
+                  "name": {
+                    "type": "string",
+                    "description": "The name the volume"
```
sttts: ... **of** the volume

src/main/proto/marathon.proto
```diff
@@ -190,7 +190,10 @@ message Volume {
 
   // Defining properties of persistent volumes
   message PersistentVolumeInfo {
-    required uint64 size = 1;
+    optional uint64 size = 1;
```
sttts: Why is `size` optional "suddenly" ?

src/main/scala/mesosphere/marathon/core/volume/Context.scala
```diff
@@ -0,0 +1,58 @@
+package mesosphere.marathon.core.volume
```
sttts: mv Context.scala BuilderContext.scala

src/main/scala/mesosphere/marathon/core/volume/Context.scala
```diff
@@ -0,0 +1,58 @@
...
+final case class ContainerContext(ci: ContainerInfo.Builder) extends BuilderContext
+/**
+  * CommandContext captures the builder that generates a Task's CommandInfo
+  */
+final case class CommandContext(ct: ContainerInfo.Type, ci: CommandInfo.Builder) extends BuilderContext
+
+/**
+  * ContextUpdate implementations decorate Context with additional metadata for a given Volume.
+  */
+trait ContextUpdate {
```
sttts: `ContextUpdater` would be a better name. This is not the `Update` itself.

src/main/scala/mesosphere/marathon/core/volume/Context.scala
```diff
@@ -0,0 +1,58 @@
...
+  /**
+    * Generates an updated BuilderContext for the given input context `c` and volume `v`.
+    * @param c is the initial input BuilderContext
+    * @param v is the volume that provides metadata with which the initial context is decorated
+    * @return a BuilderContext decorated with metadata from volume `v`; None if the context is not
+    *  supported by the storage provider associated with the given volume.
+    */
+  protected def updated[C <: BuilderContext](c: C, v: Volume): Option[C] = None
```
sttts: Why this complicated signature? The `Option[C]` type is actually not used below. Just make it `C => C` with the identify if nothing is done.

src/main/scala/mesosphere/marathon/core/volume/Context.scala
```diff
@@ -0,0 +1,58 @@
...
+    * @param volumes are the storage volumes that provide metadata to be captured in BuilderContext
+    * @param initialContext generates the initial BuilderContext to be decorated with volume metadata, it
+    *  will not be invoked if there are no volumes.
+    * @return None if there are no volumes, otherwise returns Some(BuilderContext).
+    */
+  final def apply[C <: BuilderContext](volumes: Iterable[Volume])(initialContext: () => C): Option[C] = {
+    if (volumes.isEmpty) None
+    else {
+      var cc = initialContext()
+      volumes.foreach { vol => cc = updated(cc, vol).getOrElse(cc) }
```
sttts: If the default case for `updated` was the identity, you could just use `fold` here.

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,75 @@
...
+  */
+trait VolumeProvider[+T <: Volume] {
+  /** name uniquely identifies this volume provider */
+  val name: String
+  /** validation implements a provider's volume validation rules */
+  val validation: Validator[Volume]
+  /** containerValidation implements a provider's container validation rules */
+  val containerValidation: Validator[Container]
+
+  /** apply scrapes volumes from an application definition that are supported this volume provider */
```
sttts: s/supported/supporting/

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,75 @@
...
+  def apply(container: Option[Container]): Iterable[T]
```
sttts: Am not a big fan of this signature:
- Options in negative positions, i.e. as arguments, are usually an anti-pattern
- `apply` does not really make sense here. Why not call it `collect`? Then it's clear what it does.

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,75 @@
...
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.state.{ AppDefinition, Container }
+import mesosphere.marathon.state.{ PersistentVolume, Volume }
+import mesosphere.marathon.core.volume.impl._
+
+trait LocalVolumes {
+  self: VolumeProvider[PersistentVolume] =>
+
+  /** @return a stream of task local volumes, extrapolating them from the app spec */
+  def local(app: AppDefinition): Iterable[Task.LocalVolume] = {
```
sttts: `local` is a bad name for what it does. Better: `volumes`.

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+/**
+  * API facade for callers interested in storage volumes
+  */
+object VolumesModule {
+  lazy val localVolumes: LocalVolumes = AgentVolumeProvider
+  lazy val providers: VolumeProviderRegistry = StaticRegistry
+  lazy val updates: ContextUpdate = ContextUpdate
```
sttts: s/updates/contextUpdaters/

src/main/scala/mesosphere/marathon/core/volume/impl/StaticRegistry.scala
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  val registry = make(
+    // list supported providers here
+    AgentVolumeProvider,
+    DockerHostVolumeProvider,
+    DVDIProvider
+  )
+
+  def providerForName(name: Option[String]): Option[VolumeProvider[Volume]] =
```
sttts: anti-pattern: Option in argument.

src/main/scala/mesosphere/marathon/core/volume/impl/StaticRegistry.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  override def apply[T <: Volume](v: T): Option[VolumeProvider[T]] =
+    v match {
+      case dv: DockerVolume     => Some(DockerHostVolumeProvider.asInstanceOf[VolumeProvider[T]])
+      case pv: PersistentVolume => providerForName(pv.persistent.providerName).map(_.asInstanceOf[VolumeProvider[T]])
+    }
+
+  override def apply(name: Option[String]): Option[VolumeProvider[Volume]] = providerForName(name)
+
+  override def known(): Validator[Option[String]] =
```
sttts: anti-pattern: Option in argument.

src/main/scala/mesosphere/marathon/core/volume/impl/StaticRegistry.scala
```diff
@@ -0,0 +1,45 @@
...
+    new NullSafeValidator[Option[String]](
+      test = { !apply(_).isEmpty },
+      failure = _ -> s"is not one of (${registry.keys.mkString(",")})"
+    )
+
+  override def approved[T <: Volume](name: Option[String]): Validator[T] =
```
sttts: anti-pattern: Option in argument.

src/main/scala/mesosphere/marathon/core/volume/impl/Volumes.scala
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  /**
+    * @return true if volume has a provider name that matches ours exactly
+    */
+  def accepts(volume: PersistentVolume): Boolean = {
+    volume.persistent.providerName.isDefined && volume.persistent.providerName.get == name
+  }
+
+  override def apply(container: Option[Container]): Iterable[PersistentVolume] =
```
sttts: collect is a better name than overriding apply

src/main/scala/mesosphere/marathon/core/volume/impl/Volumes.scala
```diff
@@ -0,0 +1,230 @@
...
+    v.persistent.providerName.each is equalTo(name) // sanity check
+    v.persistent.options is notEmpty
+    v.persistent.options.each is valid(validOptions)
+  }
+
+  def driversInUse(ct: Container): Set[String] =
+    ct.volumes.collect{
+      case pv: PersistentVolume if accepts(pv) && !pv.persistent.options.isEmpty =>
+        pv.persistent.options.get.get(optionDriver)
+    }.flatten.foldLeft(Set.empty[String])(_ + _)
```
sttts: ....flatten.toSet

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,75 @@
...
+trait LocalVolumes {
+  self: VolumeProvider[PersistentVolume] =>
+
+  /** @return a stream of task local volumes, extrapolating them from the app spec */
+  def local(app: AppDefinition): Iterable[Task.LocalVolume] = {
+    self.apply(app.container).map{ volume => Task.LocalVolume(Task.LocalVolumeId(app.id, volume), volume) }
+  }
+  /** @return the aggregate mesos disk resources required for volumes */
+  def diskSize(container: Option[Container]): Double = {
+    self.apply(container).map(_.persistent.size).flatten.sum.toDouble
```
sttts: map(...).flatten => flatMap

src/main/scala/mesosphere/marathon/core/volume/providers/Agent.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.wix.accord.dsl._
+import com.wix.accord.Validator
+import mesosphere.marathon.core.volume._
+import mesosphere.marathon.state._
+
+/**
+  * AgentVolumeProvider handles persistent volumes allocated from agent resources.
+  */
+protected[volume] case object AgentVolumeProvider extends PersistentVolumeProvider with LocalVolumes {
+  import org.apache.mesos.Protos.Volume.Mode
```
sttts: Is this Marathon style to have imports here?

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,166 @@
...
+      pv.persistent.options.get.get(optionDriver)
+    }.foldLeft(Set.empty[String])(_ + _)
+
+  /** @return a count of volume references-by-name within an app spec */
+  def volumeNameCounts(app: AppDefinition): Map[String, Int] =
+    DVDIProvider.this.apply(app.container).flatMap{ pv => nameOf(pv.persistent) }.
+      groupBy(identity).mapValues(_.size)
+
+  protected[providers] def modes(ct: Container): Set[Mode] =
+    DVDIProvider.this.apply(Some(ct)).map(v => Set(v.mode)).reduceLeft[Set[Mode]]{ (z, v) => z ++ v }
```
sttts: ...toSet

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,166 @@
...
+        else Some(GroupViolation(app, "app contains conflicting volumes", None,
+          internalNameViolations.toSet ++ instancesViolation.toSet ++ ruleViolations.toSet))
+      }
+      if (groupViolations.isEmpty) Success
+      else Failure(groupViolations.toSet)
+    }
+  }
+
+  def driversInUse(ct: Container): Set[String] =
+    DVDIProvider.this.apply(Some(ct)).filter(!_.persistent.options.isEmpty).flatMap{ pv =>
```
sttts: !isEmpty => isDefined

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  override def updatedContainer(cc: ContainerContext, pv: PersistentVolume): Option[ContainerContext] = {
+    // special behavior for docker vs. mesos containers
+    // - docker containerizer: serialize volumes into mesos proto
+    // - docker containerizer: specify "volumeDriver" for the container
+    val ci = cc.ci // TODO(jdef) clone?
+    if (ci.getType == ContainerInfo.Type.DOCKER && ci.hasDocker) {
+      val driverName = pv.persistent.options.get(optionDriver)
+      if (ci.getDocker.getVolumeDriver != driverName) {
+        ci.setDocker(ci.getDocker.toBuilder.setVolumeDriver(driverName).build)
```
sttts: this looks fishy. Only if the name differs, you override it?

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,166 @@
...
+        val instancesViolation: Option[RuleViolation] =
+          if (app.instances > 1) Some(RuleViolation(app.id,
+            s"Number of instances is limited to 1 when declaring external volumes in app ${app.id}", None))
+          else None
+        val ruleViolations = DVDIProvider.this.apply(app.container).toSeq.flatMap{ vol =>
+          val name = nameOf(vol.persistent)
+          if (name.isDefined) {
+            for {
+              otherApp <- g.transitiveApps.toList
+              if otherApp != app.id // do not compare to self
```
sttts: comparison of different types

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,166 @@
...
+              otherVol <- DVDIProvider.this.apply(otherApp.container)
+              otherName <- nameOf(otherVol.persistent)
+              if name == otherName
```
sttts: wrong types: otherName is String, not Option[String]

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,89 @@
...
+      }.flatten.map{ p => validate(grp)(p.groupValidation) }.fold(Success)(_ and _)
+    }
+  }
+}
+
+/**
+  * API facade for callers interested in storage volumes
+  */
+object VolumesModule {
+  lazy val localVolumes: LocalVolumes = AgentVolumeProvider
```
sttts: I would completely get rid of this class. Its methods are only used once each. Moreover, the methods are trivial.

src/main/proto/marathon.proto
jdef: not required for all use cases

src/main/scala/mesosphere/marathon/core/volume/providers/Agent.scala
jdef: yep, i've seen it elsewhere, specifically for this import

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -122,23 +123,40 @@ class TaskBuilder(app: AppDefinition,
...
-        val command =
-          TaskBuilder.commandInfo(app, Some(taskId), host, resourceMatch.hostPorts, envPrefix).toBuilder.setValue(shell)
 
         val info = ExecutorInfo.newBuilder()
           .setExecutorId(ExecutorID.newBuilder().setValue(executorId))
-          .setCommand(command)
+
         containerProto.foreach(info.setContainer)
+
+        var command =
```
jdef: s/var/val/

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -122,23 +123,40 @@ class TaskBuilder(app: AppDefinition,
 
     val containerProto = computeContainerInfo(resourceMatch.hostPorts)
     val envPrefix: Option[String] = config.envVarsPrefix.get
+
+    def decorateCommandInfo(initialCi: CommandInfo.Builder): CommandInfo.Builder = {
+      // we may not actually have a container specified, which means that we're using mesos native
+      // containerization
+      var ct: ContainerInfo.Type = containerProto.fold(ContainerInfo.Type.MESOS)(_.getType)
```
jdef: s/var/val/

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+  def volumeNameCounts(app: AppDefinition): Map[String, Int] =
+    app.container.toSet[Container].flatMap(DVDIProvider.this.collect).flatMap{ pv => nameOf(pv.persistent) }.
+      groupBy(identity).mapValues(_.size)
+
+  protected[providers] def modes(ct: Container): Set[Mode] =
+    DVDIProvider.this.collect(ct).map(_.mode).toSet
+
+  /** Only allow a single docker volume driver to be specified w/ the docker containerizer. */
+  val containerValidation: Validator[Container] = validator[Container] { ct =>
+    (ct.`type` is equalTo(ContainerInfo.Type.MESOS) and (modes(ct).each is equalTo(Mode.RW))) or (
```
jdef: we may need to enforce `RW` for all containerizers; need to test docker containerizer support for RO

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+  val optionType = name.get + "/volumeType"
+
+  val validOptions: Validator[Map[String, String]] = validator[Map[String, String]] { opt =>
+    opt.get(optionDriver) as "driverName option" is notEmpty
+    // TODO(jdef) stronger validation for contents of driver name
+    opt.get(optionDriver).each as "driverName option" is notEmpty
+    // TODO(jdef) validate contents of iops and volume type options
+  }
+
+  val validPersistentVolume = validator[PersistentVolume] { v =>
```
jdef: maybe validate mode RW here if we determine that docker doesn't support `:ro` suffix for driver-backed volumes

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+  * by a specific docker volume driver plugin. This works for both docker and mesos containerizers,
+  * albeit with some limitations:
+  *   - only a single volume driver per container is allowed when using the docker containerizer
+  *   - docker containerizer requires that referenced volumes be created prior to application launch
+  *   - mesos containerizer only supports volumes mounted in RW mode
+  */
+protected case object DVDIProvider extends InjectionHelper[PersistentVolume] with PersistentVolumeProvider {
+  import org.apache.mesos.Protos.Volume.Mode
+
+  override val name = Some("dvdi")
```
sttts: How about calling this `docker`? `dvdi` sounds very cryptic.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+    }
+    else None
+  }
+
+  // group-level validation for DVDI volumes: the same volume name may only be referenced by a single
+  // task instance across the entire cluster.
+  val groupValidation: Validator[Group] = new Validator[Group] {
+    override def apply(g: Group): Result = {
+      val groupViolations = g.apps.flatMap { app =>
+        val nameCounts = volumeNameCounts(app)
```
sttts: We need a map from `driverName, name` to `volume count` here. Names of different drivers can overlap.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+        val internalNameViolations = {
+          nameCounts.filter(_._2 > 1).map{ e =>
+            RuleViolation(app.id, s"Requested volume ${e._1} is declared more than once within app ${app.id}", None)
+          }
+        }
+        val instancesViolation: Option[RuleViolation] =
+          if (app.instances > 1) Some(RuleViolation(app.id,
```
sttts: This must be restricted to DVDI apps. At the moment it applies to all apps, even with local docker volumes. It will always fail if there is one app running with more than one volume.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+            s"Number of instances is limited to 1 when declaring external volumes in app ${app.id}", None))
+          else None
+        val ruleViolations = app.container.toSet[Container].flatMap(DVDIProvider.this.collect).flatMap{ vol =>
+          val name = nameOf(vol.persistent)
+          if (name.isDefined) {
+            for {
+              otherApp <- g.transitiveApps.toList
```
sttts: This algorithm becomes quadratic in the number of apps. Better extend the `volumeNameCounts` mapping above to return lists of app ids, not the count. Then you can produce nice error messages in O(n*log(n)).

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,163 @@
...
+              if otherApp.id != app.id // do not compare to self
+              otherVol <- otherApp.container.toSet[Container].flatMap(DVDIProvider.this.collect)
```
sttts: why container.toSet? Isn't that just an option?

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
jdef: we don't actually require the docker daemon for this feature (with the mesos containerizer) so I think "docker" is confusing. i didn't want a long name like "docker-volume-driver" because that gets very tedious when creating options prefixed with the provider name. i settled on `dvdi`. i like to think that i'm open to other options though :) is this something that we can mitigate with good documentation?

src/main/scala/mesosphere/marathon/core/volume/providers/DVDI.scala
```diff
@@ -0,0 +1,220 @@
...
+  */
+protected case object DVDIProvider extends InjectionHelper[PersistentVolume]
+    with PersistentVolumeProvider with OptionSupport {
+
+  import org.apache.mesos.Protos.Volume.Mode
+
+  val NAME = "dvdi"
+  override val name = Some(NAME)
+
+  sealed trait OptNS { val namespace: String = NAME }
```
sttts: Any reason why you use traits as base class of case classes? Overriding name below feels strange.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -196,7 +216,7 @@ class TaskBuilder(app: AppDefinition,
   protected def computeContainerInfo(ports: Seq[Int]): Option[ContainerInfo] = {
     if (app.container.isEmpty && app.ipAddress.isEmpty) None
     else {
-      val builder = ContainerInfo.newBuilder
+      var builder = ContainerInfo.newBuilder
```
jdef: builder used to be assigned to the result of the injector, looks like that was lost in the refactor. not a big deal as long as we want to acknowledge (officially) that injection has side effects and mutates the context parameter.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,229 @@
...
+    volumesForApp(app).flatMap{ pv => nameOf(pv.persistent) }.groupBy(identity).mapValues(_.size)
+
+  protected[providers] def modes(ct: Container): Set[Mode] =
+    collect(ct).map(_.mode).toSet
+
+  /** Only allow a single docker volume driver to be specified w/ the docker containerizer. */
+  val containerValidation: Validator[Container] = validator[Container] { ct =>
+    (ct.`type` is equalTo(ContainerInfo.Type.MESOS) and (modes(ct).each is equalTo(Mode.RW))) or (
+      (ct.`type` is equalTo(ContainerInfo.Type.DOCKER)) and (driversInUse(ct).size should equalTo(1))
+    // TODO(jdef) fail validation if/when user specifies "size" when using docker containerizer?
```
jdef: TODO

src/main/scala/mesosphere/marathon/core/volume/providers/DockerHostVolumeProvider.scala
```diff
@@ -0,0 +1,44 @@
...
+      .setContainerPath(volume.containerPath)
+      .setHostPath(volume.hostPath)
+      .setMode(volume.mode)
+      .build
+
+  val containerInjector = new ContainerInjector[Volume] {
+    override def inject(ctx: ContainerContext, v: Volume): ContainerContext =
+      if (ctx.container.getType == ContainerInfo.Type.DOCKER) v match {
+        case dv: DockerVolume => ContainerContext(ctx.container.addVolumes(toMesosVolume(dv)))
+        case _                => ctx // TODO(jdef) log a warning here
```
jdef: TODO

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -245,6 +245,180 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
+        cmd = Some("foo"),
+        cpus = 1.0,
+        mem = 32.0,
+        executor = "//cmd",
+        portDefinitions = Nil,
+        container = Some(Container(
+          docker = Some(Docker()), // must have this to force docker container serialization
+          `type` = MesosProtos.ContainerInfo.Type.DOCKER,
+          volumes = Seq[Volume](
+            // TODO(jdef) probably need to validate uniquess of hostPaths's that reference volume
```
jdef: TODO

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -245,6 +245,180 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
+      else None
+    }
+
+    assert(taskInfo.getContainer.getVolumesList.size > 0, "check that container has volumes declared")
+    assert(!taskInfo.getContainer.getDocker.hasVolumeDriver, "docker spec should not define a volume driver")
+    assert(vol("namedFoo").isDefined,
+      s"missing expected volume namedFoo, got instead: ${taskInfo.getContainer.getVolumesList}")
+  }
+
+  // TODO(jdef) test both dockerhostvol and persistent extvol in the same docker container
```
jdef: TODO

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -22,57 +22,18 @@ object ContainerSerializer {
...
-    val builder = mesos.Protos.ContainerInfo.newBuilder
-      .setType(container.`type`)
-      .addAllVolumes(serializedVolumes.asJava)
+    val builder = mesos.Protos.ContainerInfo.newBuilder.setType(container.`type`)
     container.docker.foreach { d => builder.setDocker(DockerSerializer.toMesos(d)) }
     builder.build
   }
 }
 
-object VolumeSerializer {
```
kolloch: You cannot know that but we want to separate the serialization logic from the model object. Serialization has nothing to do with the rest of the "business logic". Therefore, deleting this separate object is a step backwards.

I know that our code is currently inconsistent regarding this. The reason is that our `MarathonStore` currently still requires objects convertible to `MarathonState` objects that provide the serialization logic. It would be much better if it required a `MarathonConverter` instead which provided the serialization logic.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
```diff
@@ -129,7 +130,7 @@ class TaskOpFactoryImpl @Inject() (
     offer: Mesos.Offer,
     task: Task.Reserved,
     resourceMatch: Option[ResourceMatcher.ResourceMatch],
-    volumeMatch: Option[PersistentVolumeMatcher.VolumeMatch]): Option[TaskOp] = {
+    volumeMatch: Option[PersistentVolumeMatcher.VolumeResourceMatch]): Option[TaskOp] = {
```
kolloch: :+1: for also changing the name when the semantics change (or at least the context)

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -93,7 +93,7 @@ class OfferOperationFactory(
       Mesos.Resource.newBuilder()
         .setName("disk")
         .setType(Mesos.Value.Type.SCALAR)
-        .setScalar(Mesos.Value.Scalar.newBuilder().setValue(vol.persistentVolume.persistent.size.toDouble).build())
+        .setScalar(Mesos.Value.Scalar.newBuilder().setValue(vol.persistentVolume.persistent.size.get.toDouble).build())
```
kolloch: Don't use `get` but preferable `foreach` around adding the disk resource. It is optional, correct?

If it is not optional, make sure we get a more meaningful error message by throwing an appropriate exception.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -350,6 +351,13 @@ object Task {
 
   }
 
+  /** @return a stream of task local volumes, extrapolating them from the app spec */
+  def localVolumes(app: AppDefinition): Iterable[LocalVolume] = {
+    app.container.toSet[Container].flatMap(VolumesModule.localVolumes.collect).map{ volume =>
```
kolloch: formatting nit-pick: We use a space between map (or another method) and a block. E.g. `.map {` instead of `.map{`.

A better place for this method would be in the companion object of `LocalVolume`. Or maybe even in the `VolumesModule`.

src/main/scala/mesosphere/marathon/core/volume/Injection.scala
```diff
@@ -0,0 +1,66 @@
...
+  def inject[V <: Volume, C <: InjectionContext](c: C, v: V)(implicit injector: VolumeInjector[V, C]): C =
+    injector.inject(c, v)
+
+  override def apply[V <: Volume, C <: InjectionContext](c: C, vols: Iterable[V])(
+    implicit injector: VolumeInjector[V, C]): C = {
+    var ctx: C = c
+    vols.foreach{ v => ctx = inject(ctx, v) }
+    ctx
+  }
+}
```
kolloch: This whole file feels like it just enables a restricted fold over two "contexts". I think it can be simplified significantly.

It is a bit difficult to do that over github PR features. Maybe we should schedule a hangout?

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,94 @@
...
+trait VolumeProvider[+T <: Volume] {
+  /** appValidation implements a provider's app validation rules */
+  val appValidation: Validator[AppDefinition]
+  /** groupValidation implements a provider's group validation rules */
+  val groupValidation: Validator[Group]
+
+  /** apply scrapes volumes from an application definition that are supported by this volume provider */
+  def collect(container: Container): Iterable[T]
+
+  val containerInjector: ContainerInjector[Volume]
```
kolloch: maybe I am wrong but could this be just a method that takes a `T` and a `CommandInfo.Builder` and applies the appropriate modifications to the `CommandInfo.Builder`?

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,94 @@
+package mesosphere.marathon.core.volume
+
+import com.wix.accord._
+import mesosphere.marathon.core.volume.providers._
+import mesosphere.marathon.state._
+
+/**
+  * VolumeProvider is an interface implemented by storage volume providers
+  */
+trait VolumeProvider[+T <: Volume] {
```
kolloch: :+1: It is nice that you create an explicit interface to different pluggable VolumeProviders!

kolloch: Sometimes `sealed traits` + pattern matching provide an alternative that is similar to the visitor pattern. Maybe we can also use that.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,245 @@
...
+  * DVDIProvider (Docker Volume Driver Interface provider) handles persistent volumes allocated
+  * by a specific docker volume driver plugin. This works for both docker and mesos containerizers,
+  * albeit with some limitations:
+  *   - only a single volume driver per container is allowed when using the docker containerizer
+  *   - docker containerizer requires that referenced volumes be created prior to application launch
+  *   - mesos containerizer only supports volumes mounted in RW mode
+  */
+protected[volume] case object DVDIProvider
+    extends AbstractPersistentVolumeProvider("dvdi")
+    with OptionSupport {
```
kolloch: I don't like to mix in implementation with traits. (there is a reason why languages like golang/rust don't allow that anymore)

You can make OptionSupport simply an object and use an import. I even prefer referring to the types it provides with a prefix, e.g. `OptionSupport.NamedOption` which makes it easy to determine where it is coming from.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,245 @@
...
+
+  import org.apache.mesos.Protos.Volume.Mode
+
+  abstract class ProviderOption(
+    override val name: String,
```
kolloch: you do not need to override the fields with just the same thing. Just get rid of `override val`. If possible, even get rid of the complete inheritance here -- only the validation is really polymorphic.

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
jdef: discussed this via hangout: quite possibly going to introduce plugins that can implement the VolumeProvider interface. in that case, a sealed trait would be limiting.

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
jdef: @sttts why didn't we do this? sure seems simpler.

src/main/scala/mesosphere/marathon/core/task/Task.scala
jdef: can formatting stuff like that be automated? dislike spending human cycles on such

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
jdef: in this case validation would have failed if size wasn't specified. but we could probably also fold or reduce here instead of using get

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
sttts: Good question. Makes sense for sure.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
jdef: I was trying to be consistent w/ respect to how and where VolumesModule.inject was being invoked. I'll think about a better way to approach this. Moving the serialization to the objects felt natural because validation is also closely coupled w/ the business objects. In the volumes case, both aspects (serialization and validation) depend on the provider so ultimately I'd like to pull everything out of the "state" layer and into somewhere else.

src/main/scala/mesosphere/marathon/core/volume/Injection.scala
jdef: discussed via hangout: we should simplify

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
jdef: done

src/main/scala/mesosphere/marathon/core/task/Task.scala
sttts: idea can format code.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
jdef: done

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -252,6 +252,232 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+    When("The request is processed")
+    val response = appsResource.create(body, false, auth.request)
+
+    Then("The return code indicates create success")
+    response.getStatus should be(201)
+  }
+
+  test("Creating an app with a duplicate external volume name (unfortunately) passes validation") {
+    // we'll need to mitigate this with documentation: probably deprecating support for using
```
jdef: TODO

src/main/scala/mesosphere/marathon/core/volume/VolumesModule.scala
```diff
@@ -0,0 +1,94 @@
...
+    * the default PersistenVolumeProvider implementation is returned. None is returned if Some name is given
+    * but no volume provider is registered for that name.
+    */
+  def apply(name: Option[String]): Option[PersistentVolumeProvider[PersistentVolume]]
+}
+
+/**
+  * API facade for callers interested in storage volumes
+  */
+object VolumesModule {
```
jdef: TODO: rename to VolumesService

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -17,20 +17,22 @@ import scala.collection.JavaConverters._
...
-    with OptionSupport {
+    extends AbstractPersistentVolumeProvider("external") {
 
   import org.apache.mesos.Protos.Volume.Mode
+  import OptionSupport._
 
   abstract class ProviderOption(
-    override val name: String,
-    override val required: Boolean = false) extends NamedOption(DVDIProvider.this.name, name, required)
+    name: String,
```
sttts: :+1: for getting rid of override.

src/main/scala/mesosphere/marathon/core/volume/providers/ResidentVolumeProvider.scala
```diff
@@ -0,0 +1,40 @@
...
+  val appValidation: Validator[AppDefinition] = new NilValidator[AppDefinition]
+
+  // no provider-specific rules at the group level
+  val groupValidation: Validator[Group] = new NilValidator[Group]
+
+  val volumeValidation = validator[PersistentVolume] { v =>
+    v.persistent.size is notEmpty
+    v.mode is equalTo(Mode.RW)
+    //persistent volumes require those CLI parameters provided
+    v is configValueSet("mesos_authentication_principal", "mesos_role", "mesos_authentication_secret_file")
```
sttts: This does not really belong here IMO.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -217,6 +218,7 @@ object Group {
       group is noAppsAndGroupsWithSameName
       (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
       group is validPorts
+      group is VolumesModule.validGroup
```
kolloch: is this the correct place to call this?

kolloch: I am concerned with the algorithmic complexity of this.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,247 @@
...
+  import org.apache.mesos.Protos.Volume.Mode
+  import OptionSupport._
+
+  abstract class ProviderOption(
+    name: String,
+    required: Boolean = false,
+    validValue: Validator[String] = labelValidator) extends NamedOption(
+    DVDIProvider.this.name, name, required, validValue)
+
+  case object OptionDriverName extends ProviderOption("driver", true)
```
sttts: The `driver` option is required. This means you always have to pass `rex-ray`. Is it what we want?

sttts: Moreover, if the driver is not given as an option, we get very strange violations:
```
{"message":"Object is not valid","details":[{"path":"/value","errors":["does not equal 1"]},{"path":"/container/volumes(0)/persistent.options","errors":["is a required option, but is not present"]},{"path":"/value/container.type","errors":["does not equal MESOS"]}]}
```

jdef: yes, you're required to always pass something like "rexray"

src/main/scala/mesosphere/marathon/state/Group.scala
jdef: could you be more specific (w/ respect to both comments)? @aquamatthias recommended that I look at service port validation for inspiration, which I did.

src/main/scala/mesosphere/marathon/core/volume/providers/ResidentVolumeProvider.scala
jdef: it used to be in state/Volume. I don't like it in either place. do you have a suggestion for relocating this?

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -17,106 +23,251 @@ object Volume {
...
-          persistent = persistentVolumeInfo,
-          mode = proto.getMode
-        )
-      case None =>
-        DockerVolume(
-          containerPath = proto.getContainerPath,
-          hostPath = proto.getHostPath,
-          mode = proto.getMode
-        )
+    if (persistent.isDefined) {
```
jdef: this feels weak! what stops someone from defining both persistent and external in the same volume spec?

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -17,106 +23,251 @@ object Volume {
...
+  * than 31 characters. YMMV. Although `name` is optional, some storage providers may require it.
+  *
+  * `name` uniqueness:
+  *  <li> A volume name MUST be unique within the scope of a volume provider.
+  *  <li> A fully-qualified volume name is expected to be unique across the cluster and may formed, for example,
+  *       by concatenating the volume provider name with the volume name. E.g “dvdi.volume123”
+  *
+  * `providerName` is optional; if specified it indicates which storage provider will implement volume
+  * lifecycle management operations for the external volume. if unspecified, “agent” is assumed.
+  * the provider names “dcos”, “agent”, and "docker" are currently reserved. The contents of providerName
```
jdef: "agent" and "docker" are no longer reserved

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -122,23 +123,38 @@ class TaskBuilder(app: AppDefinition,
 
     val containerProto = computeContainerInfo(resourceMatch.hostPorts)
     val envPrefix: Option[String] = config.envVarsPrefix.get
+
+    def decorateCommandInfo(builder: CommandInfo.Builder) = containerProto.foreach { cp =>
+      app.container.foreach { container =>
+        container.volumes.foreach {
+          case pv: ExternalVolume => VolumesModule.build(cp.getType, builder, pv)
+          case _                  =>
```
jdef: what does this even mean?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -252,6 +264,163 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+      """.stripMargin
+    )
+
+    Then("The return code indicates that the hostPath of volumes[0] is missing") // although the wrong field should fail
+    response.getStatus should be(422)
+    response.getEntity.toString should include("/container/volumes(0)/external.providerName")
+    response.getEntity.toString should include("is not a external volume provider name")
+  }
+
+  /* TODO(jdef) this test contains JSON that's not even parsable, so JsResultException is thrown
```
jdef: TODO

src/test/scala/mesosphere/marathon/state/ContainerTest.scala
```diff
@@ -152,7 +151,7 @@ class ContainerTest extends MarathonSpec with Matchers {
     val proto = ContainerSerializer.toProto(f.container)
     assert(mesos.ContainerInfo.Type.DOCKER == proto.getType)
     assert("group/image" == proto.getDocker.getImage)
-    assert(f.container.volumes == proto.getVolumesList.asScala.map(Volume(_)))
+    assert(f.container.volumes == proto.getVolumesList.asScala.map(Volume.fromProto(_)))
```
jdef: @kolloch serialization was already split, I just brought it into state vs. moved it all to api.serialization. I DID get an OK from a matthias via slack or something before I did this.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
jdef: discussed this with @sttts and identified possible workarounds. they all suck. the best thing we can do is to deprecate the use of volume names with hostPath and tell people to use external volumes if that's what they want.

src/main/scala/mesosphere/marathon/state/Volume.scala
jdef: I added checks for this using IllegalArgumentException. It's probably not kosher but the json format/play stuff is completely insane

src/main/scala/mesosphere/marathon/state/Group.scala
kolloch: I discussed it with @sttts. This should be called by the validation in the root group -- not once per every group. Make sure that validation is really cheap. In Marathon, everything is validated on every change.

src/test/scala/mesosphere/marathon/state/ContainerTest.scala
kolloch: I am not sure that I understand your comment completely. Anyway, having it separate and/or in state.serialization or something like that sounds great (it has nothing to do with the api).

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
sttts: done

aquamatthias:
Closed in favor of #3675

Issue #3091, opened 2016-01-25
Problem with upgradeStrategy and Scale


ch3lo:
Hi,

I have a problem deploying a Docker Service with Marathon **0.14** and Mesos **0.26.0**

I made a `PUT` to `localhost:8081/v2/apps` with the next payload and I have no problem. Marathon does exactly what I want.

``` json
[{
  "id": "/www/services/test-http/test-http-v0",
  "cpus": 0.1,
  "mem": 32,
  "instances": 5,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "xxx/test-http:0-svn479-test-http",
      "network": "BRIDGE",
      "portMappings": [
        { "containerPort": 80, "hostPort": 0, "protocol": "tcp"}
      ]
    }
  },
  "healthChecks": [
    {
      "protocol": "HTTP",
      "portIndex": 0,
      "path": "/v0/healthy",
      "gracePeriodSeconds": 30,
      "intervalSeconds": 20,
      "maxConsecutiveFailures": 3
    }
  ],
  "upgradeStrategy": {
    "minimumHealthCapacity": 1,
    "maximumOverCapacity": 0.2
  }
}]
```

Then I made another `PUT` to `localhost:8081/v2/apps` with the objective to upgrade my service to `0-svn480-test-http` with this:

``` json
[{
  "id": "/www/services/test-http/test-http-v0",
  "cpus": 0.1,
  "mem": 32,
  "instances": 3,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "xxx/test-http:0-svn480-test-http",
      "network": "BRIDGE",
      "portMappings": [
        { "containerPort": 80, "hostPort": 0, "protocol": "tcp"}
      ]
    }
  },
  "healthChecks": [
    {
      "protocol": "HTTP",
      "portIndex": 0,
      "path": "/v0/healthy",
      "gracePeriodSeconds": 30,
      "intervalSeconds": 20,
      "maxConsecutiveFailures": 3
    }
  ],
  "upgradeStrategy": {
    "minimumHealthCapacity": 1,
    "maximumOverCapacity": 0.2
  }
}]
```

With `minimumHealthCapacity` **1** Marathon scales `0-svn479-test-http` from **5** to **3** (I think it is ok, because my scale parameter is 3). The next step should be deploy the new service (one instance at time, because my maximumOverCapacity is 0.2), however Marathon **hangs** in **RestartApplication**.

[...]

PR #3531, opened 2016-03-18
Fixes 3091 - App updates hanging on downscales

pgkelley4:
Fixed bug where app updates would hang on a downscale
and maximumOverCapacity < (1 / app.instances).

One line fix: Increment maxCapacity on downscales to get update
started.

Added unit test to demonstrate the issue.

### DIFF
--- a/src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
+++ b/src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
@@ -45,7 +45,7 @@ class TaskReplaceActor(
     val nrToKillImmediately = math.max(0, toKill.size - minHealthy)
 
     // make sure at least one task can be started to get the ball rolling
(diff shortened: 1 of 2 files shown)


pgkelley4:
As discussed in https://github.com/mesosphere/marathon/issues/3091.

The issue is maxCapacity needs to be incremented for the update to start. But on a downscale where `nrToKillImmediately == 0` is always false and `maxCapacity == app.instances` is true when `maximumOverCapacity * instances < 1`, this won't happen.

maxCapacity is set as:
`var maxCapacity = (app.instances * (1 + app.upgradeStrategy.maximumOverCapacity)).toInt`

And this evaluates to false:
```
if (nrToKillImmediately == 0 && maxCapacity == app.instances)
     maxCapacity += 1
```

Then the update hangs.

Tests are passing for me locally and all I did was remove code and add a unit test so I'm not sure why the coverage decreased.

aquamatthias:
Thanks!

Issue #3311, opened 2016-02-23
Resident Tasks: Renaming and cleanup


meichstedt:
Renaming after other big changes (to avoid conflicts):
- [ ] #3456 adds the possibility to wipe task state in order to free reserved resources. The underlying implementation in not fully tested – write a decent IntegrationSuite for that!
- [ ] Adjust package structure and names in general. Update the [Docs](https://mesosphere.github.io/marathon/docs/core-architecture.html) on Architecture (names, flow)
- [ ] Maybe subdivide TaskStateOp into two sub sealed traits: NewTaskStateOp, ChangeTaskStateOp. Then implement Task.update only for ChangeTaskStateOp?
- [ ] `TaskStateOp.LaunchEphemeral` contains a `Task` while it should expect a `Task.launchedEphemeral`.
- [ ] rename QueuedTaskCount.taskLaunchesInFlight to taskOpsInFlight --> 1h
- [ ] Rename `LocalVolume*` to `PersistentVolume*` where applicable. We decided to speak of _Local Volumes_ and _Network Volumes_ instead of _Persistent_ and _External_ volumes, respectively. However, we should use the naming of the underlying Mesos features throughout the code and only use _official_ naming in the API layer – otherwise it's hard to draw the line somewhere else. --> 1h
- [ ] Rename `TaskOpFactory` to `TaskOpMatcher`. `TaskOpFactory.buildTaskOp` to `matchRequest`.
- [ ] Remove traits `TaskCreationHandler` and `TaskReservationTimeoutHandler`
  - Rename `TaskCreationHandlerAndUpdaterDelegate` to `TaskStateOpProcessorImpl`
  - adjust the tests (this is the main effort here)
- [ ] Get rid of `TAskStateOp.possibleNewState` – it's only used in the AppTaskLauncherActor and `None` in case of a `LaunchedOnReservation`
- [ ] Clearly distinguish between `TaskOp/TaskOpProcessor` and `TaskStateOp/TaskStateOpProcessor`. Maybe the former can be renamed to something related to `Offer` since it's output is related to OfferOperations (mostly everything in Marathon is related to tasks anyway)

meichstedt:
Note: This issue has been migrated to https://jira.mesosphere.com/browse/MARATHON-2319. For more information see https://groups.google.com/forum/#!topic/marathon-framework/khtvf-ifnp8.

PR #3533, opened 2016-03-21
Towards #3221 - serialized task updates

meichstedt:


src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
```diff
@@ -103,9 +104,9 @@ private[launcher] class OfferProcessorImpl(
       terminatedFuture.flatMap { _ =>
         nextOp.oldTask match {
           case Some(existingTask) =>
-            taskCreationHandler.created(existingTask).map(_ => ())
+            taskCreationHandler.created(TaskStateOp.Create(existingTask)).map(_ => ())
```
kolloch: I would still prefer an explicit `TaskStateOp.Revert` operation.

src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
```diff
@@ -121,20 +122,10 @@ private[launcher] class OfferProcessorImpl(
...
-              s"after applying the effects of ${taskOpWithSource.op.getClass.getSimpleName}"
-          )
-          taskCreationHandler.created(newTask)
-        case None =>
-          log.info(s"Remove ${taskOpWithSource.taskId} because of ${taskOpWithSource.op.getClass.getSimpleName}")
-          taskCreationHandler.terminated(taskId)
-      }
-
-      persistedOp.map(_ => Some(taskOpWithSource))
+      log.info("Save task [{}]", taskOpWithSource.taskId)
```
kolloch: you can already change the logging "Persisting TaskStateOp"

In fact, you can already rename the `newTask` field -- nobody is working on this concurrently. At least it confused me...

Have you documented that you want to change taskCreationHandler.created somewhere? Do you keep track of planned renames somewhere? We have this issue for that: https://github.com/mesosphere/marathon/issues/3311

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLauncherImpl.scala
```diff
@@ -37,8 +37,8 @@ private[launcher] class TaskLauncherImpl(
     if (accepted) {
       usedOffersMeter.mark()
       val launchCount = taskOps.count {
-        case TaskOp.Launch(_, _, _, _) => true
-        case _                         => false
+        case _: TaskOp.Launch => true
```
kolloch: better :+1:

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
```diff
@@ -13,21 +13,31 @@ class TaskOpFactoryHelper(
...
     def createOperations = Seq(offerOperationFactory.launch(taskInfo))
 
-    TaskOp.Launch(taskInfo, newTask, oldTask, createOperations)
+    val stateOp = TaskStateOp.Create(newTask)
+    TaskOp.Launch(taskInfo, stateOp, oldTask = None, createOperations)
+  }
+
+  def launchOnReservation(
+    taskInfo: Mesos.TaskInfo,
+    newTask: TaskStateOp,
```
kolloch: could the type be more specific? LaunchOnReservation?

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
```diff
@@ -36,6 +46,6 @@ class TaskOpFactoryHelper(
       offerOperationFactory.reserve(frameworkId, newTask.taskId, resources),
       offerOperationFactory.createVolumes(frameworkId, newTask.taskId, localVolumes))
 
-    TaskOp.ReserveAndCreateVolumes(newTask, resources, localVolumes, oldTask, createOperations)
+    TaskOp.ReserveAndCreateVolumes(newTask, resources, localVolumes, createOperations)
```
kolloch: jepp, you should be right, no old task should exist for this. Good to make that explicit :+1:

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
```diff
@@ -13,21 +13,31 @@ class TaskOpFactoryHelper(
 
   private[this] val offerOperationFactory = new OfferOperationFactory(principalOpt, roleOpt)
 
-  def launch(
+  def launchEphemeral(
```
kolloch: I like the new more explicit "launch*" names, thanks! :+1:

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
```diff
@@ -9,13 +9,17 @@ object TaskStatusObservables {
   case class TaskStatusUpdate(timestamp: Timestamp, taskId: Task.Id, status: MarathonTaskStatus) {
```
kolloch: You can remove the `TaskStatusUpdate` class.

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
```diff
@@ -9,13 +9,17 @@ object TaskStatusObservables {
...
+  }
 }
 
 /**
   * Allows subscription to TaskStatus updates.
   */
 trait TaskStatusObservables {
-  def forAll: Observable[TaskStatusUpdate]
-  def forAppId(appId: PathId): Observable[TaskStatusUpdate]
+  def forAll: Observable[TaskUpdate]
```
kolloch: With this change you actually already moved pretty close to remove these TaskStatusObervables alltogether. Nice!

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -221,41 +221,31 @@ private class AppTaskLauncherActor(
...
-
-          case TaskStateChange.NoChange =>
-            log.debug("no change for {}", taskId)
-
-          case TaskStateChange.Failure(cause) =>
-            log.warning("Task state update failed for {}. Reason: {}", taskId, cause.getMessage)
-        }
+  private[this] def receiveTaskUpdate: Receive = {
+    case TaskUpdate(stateOp, stateChange) =>
+      log.info("received TaskUpdate")
```
kolloch: I think that I would get rid of this info log statement and promote the more specific log statements below to info if you want more visibility.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -341,19 +331,23 @@ private class AppTaskLauncherActor(
...
+      // We will receive the updated task once it's been persisted. Before that,
+      // we can only store the possible state, as we don't have the updated task
+      // yet.
+      taskOp.newTask.possibleNewState match {
+        case Some(newState) =>
+          tasksMap += taskId -> newState
           scheduleTaskOpTimeout(taskOp)
+
         case None =>
-          removeTask(taskId)
```
kolloch: hmmm. "removeTask" is not really needed -- it was mostly defensive and maybe misleading. It was different in some historical state, I think. But then, we can actually just not do anything because the task should not be in the tasksMap. This branch is unnecessary, isn't it?

src/main/scala/mesosphere/marathon/core/leadership/LeadershipModule.scala
```diff
@@ -36,11 +37,13 @@ object LeadershipModule {
...
     actorRefFactory: ActorRefFactory, zk: ZooKeeperClient, leader: LeadershipAbdication) extends LeadershipModule {
 
+  private[this] val log = LoggerFactory.getLogger(getClass)
   private[this] var whenLeaderRefs = Set.empty[ActorRef]
   private[this] var started: Boolean = false
 
   override def startWhenLeader(props: Props, name: String): ActorRef = {
-    require(!started, "already started")
+    log.info("startWhenLeader: {}", name + props.toString)
+    require(!started, s"already started: $name ${props.toString}")
```
kolloch: the `toString` is implicit...

src/main/scala/mesosphere/marathon/core/leadership/LeadershipModule.scala
```diff
@@ -51,6 +54,9 @@ private[leadership] class LeadershipModuleImpl(
 
   private[this] lazy val coordinator_ = {
     require(!started, "already started")
+    log.info("LeadershipModule is now started")
+    //    throw new scala.RuntimeException("Sorry!")
```
kolloch: Apology accepted. Now you can just delete that line.

src/main/scala/mesosphere/marathon/core/matcher/reconcile/impl/OfferMatcherReconciler.scala
```diff
@@ -62,9 +62,14 @@ private[reconcile] class OfferMatcherReconciler(taskTracker: TaskTracker, groupR
             case (taskId, spuriousResources) if spurious(taskId) =>
               val unreserveAndDestroy =
                 TaskOp.UnreserveAndDestroyVolumes(
-                  taskId = taskId, oldTask = tasksByApp.task(taskId), resources = spuriousResources.to[Seq]
+                  newTask = TaskStateOp.ForceExpunge(taskId),
+                  oldTask = tasksByApp.task(taskId),
+                  resources = spuriousResources.to[Seq]
                 )
               TaskOpWithSource(source(offer.getId), unreserveAndDestroy)
+            //          case (taskId, spuriousResources) if shouldRefreshLastSeen(taskId) =>
```
kolloch: this can go, I guess

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -83,39 +83,46 @@ object Task {
...
     override def update(update: TaskStateOp): TaskStateChange = update match {
       // case 1: now running
-      case TaskStateOp.MesosUpdate(MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
+      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
         val updated = copy(
           status = status.copy(
             startedAt = Some(now),
             mesosStatus = mesosStatus))
-        TaskStateChange.Update(updated)
+        TaskStateChange.Update(updated, Some(this))
```
kolloch: I think that named parameters would make this a lot clearer.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -83,39 +83,46 @@ object Task {
...
-      case TaskStateOp.MesosUpdate(taskStatus, now) =>
-        val healthOrStateChange = for {
-          current <- status.mesosStatus
-          update <- taskStatus.mesosStatus
-          newStatus <- updatedHealthOrState(current, update)
-        } yield TaskStateChange.Update(copy(status = status.copy(mesosStatus = Some(newStatus))))
-
-        healthOrStateChange.getOrElse{
+      case TaskStateOp.MesosUpdate(_, taskStatus, now) =>
+        updatedHealthOrState(status.mesosStatus, taskStatus.mesosStatus).map { newStatus =>
```
kolloch: Style (ignore me if you want)

if you handle both cases of an option, I prefer a `match`.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -83,39 +83,46 @@ object Task {
...
-        TaskStateChange.Failure("Unable to handle Launch op on LaunchedEphemeral")
+      case TaskStateOp.ForceExpunge(_) =>
+        TaskStateChange.Expunge(this)
 
-      // failure case: Timeout
-      case TaskStateOp.ReservationTimeout =>
-        TaskStateChange.Failure("Unable to handle Timeout op on LaunchedEphemeral")
+      // FIXME (3221): If the task needs to recreated after a not-accepted taskOp, Create is used.
+      // that's neither nice nor obvious, so this should be improved
+      case TaskStateOp.Create(task) =>
```
kolloch: I would like this to be renamed to "Revert"

kolloch: And it could take an Option of a task such that it translates to an `Update` or an `Expunge`

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -169,44 +188,55 @@ object Task {
...
     override def update(update: TaskStateOp): TaskStateChange = update match {
       // case 1: now running
-      case TaskStateOp.MesosUpdate(MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
+      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
         val updated = copy(
           status = status.copy(
             startedAt = Some(now),
             mesosStatus = mesosStatus))
-        TaskStateChange.Update(updated)
+        TaskStateChange.Update(updated, Some(this))
```
kolloch: named parameters?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskCreationHandlerAndUpdaterDelegate.scala
```diff
@@ -3,50 +3,54 @@ package mesosphere.marathon.core.task.tracker.impl
...
+  TaskTrackerConfig
+}
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.control.NonFatal
 
 /**
-  * Implements the [[TaskCreationHandler]]/[[TaskUpdater]] traits by sending messages to the TaskTracker actor.
+  * Implements the [[TaskStateOpProcessor]] trait by sending messages to the TaskTracker actor.
```
kolloch: `TaskOpProcessor` and `TaskStateOpProcessor` are very confusingly named since they do very similar things.

Is there a cleanup/renaming issue that already tracks this?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskCreationHandlerAndUpdaterDelegate.scala
```diff
@@ -3,50 +3,54 @@ package mesosphere.marathon.core.task.tracker.impl
...
+  override def process(stateOp: TaskStateOp): Future[TaskStateChange] = {
+    taskUpdate(stateOp.taskId, stateOp)
+  }
+
+  override def created(taskStateOp: TaskStateOp): Future[Unit] = {
+    process(taskStateOp).map(_ => ())
   }
-  override def terminated(taskId: Task.Id): Future[_] = {
-    taskUpdate(taskId, TaskOpProcessor.Action.Expunge)
+  override def terminated(stateOp: TaskStateOp.ForceExpunge): Future[_] = {
```
kolloch: Covered by cleanup issue to remove the specialized methods?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -18,43 +19,55 @@ import scala.util.control.NonFatal
...
+      * * a TaskStateChange.Failure if the task does not exist OR ELSE
+      * * delegates the TaskStateOp to the existing task that will then determine the state change
       */
-    def resolve(taskId: Task.Id, status: TaskStatus)(
-      implicit ec: ExecutionContext): Future[Action] = {
+    def resolve(op: TaskStateOp)(implicit ec: ExecutionContext): Future[TaskStateChange] = {
+      // FIXME (3221): Create and Expunge are mostly to restore states.
+      op match {
+        case op: TaskStateOp.Create              => updateIfNotExists(op.taskId, op.task)
+        case op: TaskStateOp.ForceExpunge        => updateExistingTask(op)
```
kolloch: I am not sure if it is much better (especially with our code formatter) but you do now that you can group clauses, right? You basically only distinguish between updating and creating here.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -18,43 +19,55 @@ import scala.util.control.NonFatal
...
+          val taskId = op.taskId
+          TaskStateChange.Failure(new IllegalStateException(s"$taskId of app [${taskId.appId}] does not exist"))
       }
     }
   }
 }
+
 /**
   * Processes durable operations on tasks by
   *
```
kolloch: Extend comment to cover new functionality.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -64,51 +77,96 @@ private[tracker] object TaskOpProcessorImpl {
...
+    }
+  }
+
+  private[this] def notifyOthers(stateOp: TaskStateOp, stateChange: TaskStateChange): Future[Unit] = {
+    // compatibility: if it's a statusUpdate that resulted in a Update or Expunge, do as before
+
+    (stateOp, stateChange) match {
+      case (TaskStateOp.MesosUpdate(taskId, MarathonTaskStatus.WithMesosStatus(mesosStatus), now),
+        update: TaskStateChange.Update) =>
+        log.info("StateChange.Update -> notifying taskStatusUpdateSteps")
```
kolloch: Unify this block with the next by also logging the class name of the taskStateOp?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -64,51 +77,96 @@ private[tracker] object TaskOpProcessorImpl {
...
+          taskTrackerRef ! TaskTrackerActor.TaskUpdated(stateChange.task, TaskTrackerActor.Ack(op.sender, stateChange))
+        }.flatMap { _ =>
+          notifyOthers(op.stateOp, stateChange)
+        }.recoverWith(tryToRecover(op)(
+          expectedTaskState = Some(stateChange.task), oldTaskState = stateChange.oldTask))
+    }
+  }
+
+  private[this] def notifyOthers(stateOp: TaskStateOp, stateChange: TaskStateChange): Future[Unit] = {
+    // compatibility: if it's a statusUpdate that resulted in a Update or Expunge, do as before
```
kolloch: Am I missing something or are you only propagating mesos status updates?

We also need to propagate other updates -- at least to the launch queue (which will forward them to the `AppTaskLauncherActor`).

kolloch: That should then also be somehow covered in unit/integration tests.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -129,20 +185,20 @@ private[tracker] class TaskOpProcessorImpl(
         TaskTrackerActor.Ack(op.sender, msg)
       }
 
-      log.warn(
-        s"${op.taskId} of app [${op.taskId.appId}]: try to recover from failed ${op.action.toString}", cause
-      )
+      log.warn(s"${op.taskId} of app [${op.taskId.appId}]: try to recover from failed ${op.stateOp.toString}", cause)
```
kolloch: `toString` is implicit and it is customary to not write it explicitly.

src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
meichstedt: changed.

src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
meichstedt: now renamed – there was a fixme which I removed, I've added it to #3311.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
meichstedt: Oh, yes. I also changed the oldTask type and the related type in `VolumeMatcher`

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
meichstedt: Yay, removed. Just kept that in there while migrating.

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
meichstedt: Ha, I wasn't even aware of that :grin:

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Totally, I like the idea. Changed it!

src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
meichstedt: Right, removed it.

src/main/scala/mesosphere/marathon/core/leadership/LeadershipModule.scala
meichstedt: thanks for finding this debugging-leakage :+1:

src/main/scala/mesosphere/marathon/core/leadership/LeadershipModule.scala
meichstedt: Ha ha ^_^

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: ack – renamed to `newState` and `oldState`, respectively

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: I agree it is more readable

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: In order to keep the changes small I chose to go for
- `TaskStateOp.Revert(task: Task)` reverts to the given state
- `TaskStateOp.ForceExpunge(taskId: Task.Id)` expunges the given taskId

These 2 can be consolidated later on: at least no there are these 2 that explicitly handle recovery – I hope you're ok with that for now.

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: everywhere!

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskCreationHandlerAndUpdaterDelegate.scala
meichstedt: Agree. I've added a bullet point to #3311

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskCreationHandlerAndUpdaterDelegate.scala
meichstedt: yes

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: At least IDEA by default put annoying (yet sometimes helpful) warnings in there when I did that in the past, therefore I moved away from doing it completely.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: Seems like I accidentally reverted that – this was the main point in all this refactoring …

I agree that this should be tested, I will add some tests.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: removed

meichstedt:
Thanks @kolloch for your in-depth comments! I hope I've addressed them all. As noted above I have added tasks to #3311 – I will try to resolve those asap. I'd like to pull at least the TaskOp/TaskStateOp renaming into this PR, maybe we find some nice names tomorrow.

meichstedt:
Oh, in case you have more comments I will of course address them as well. Just realized the PR was still assigned to you :grinning:

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -18,97 +19,149 @@ import scala.util.control.NonFatal
...
-
-      case Action.UpdateStatus(status) =>
-        statusUpdateActionResolver.resolve(op.taskId, status).flatMap { action: Action =>
-          // Since this action is mapped to another action, we delegate the responsibility to inform
-          // the sender to that other action.
-          process(op.copy(action = action))
-        }
+          taskTrackerRef ! TaskTrackerActor.TaskRemoved(op.taskId, TaskTrackerActor.Ack(op.sender, change))
+        }.flatMap { _ =>
+          notifyOthers(op.stateOp, change)
```
kolloch: It is not (strictly) guaranteed that the message to the `TaskTracker` is processed before the steps for this change are processed. It would be nice, to guarantee that. The `TaskTrackerActor.Ack` construct was introduced to do exactly that for the other changes -- the `TaskTracker` would notify the requestor of this change of the completion AFTER it had applied the change to its in-memory task list.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
```diff
@@ -18,97 +19,149 @@ import scala.util.control.NonFatal
...
+        }.recoverWith(tryToRecover(op)(
+          expectedTaskState = Some(change.newState), oldTaskState = change.oldState))
+    }
+  }
+
+  private[this] def notifyOthers(stateOp: TaskStateOp, stateChange: TaskStateChange): Future[Unit] = {
+    // compatibility: if it's a statusUpdate that resulted in a Update or Expunge, do as before
+
+    stateChange match {
+      case _: TaskStateChange.Update | _: TaskStateChange.Expunge =>
```
kolloch: This logic is redundant. You only call the `notifyOthers` method for these cases so you don't really check for the relevant ops again.

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
```diff
@@ -1,13 +1,14 @@
...
-import mesosphere.marathon.state.{ PathId, Timestamp }
+import mesosphere.marathon.core.task.bus.TaskStatusObservables.TaskUpdate
+import mesosphere.marathon.core.task.{ Task, TaskStateChange, TaskStateOp }
+import mesosphere.marathon.state.PathId
 import rx.lang.scala.Observable
 
 object TaskStatusObservables {
-  case class TaskStatusUpdate(timestamp: Timestamp, taskId: Task.Id, status: MarathonTaskStatus) {
-    def appId: PathId = taskId.appId
+  case class TaskUpdate(stateOp: TaskStateOp, stateChange: TaskStateChange) {
```
kolloch: I would prefer `TaskChanged`. 

* `TaskUpdate` is very close to `TaskStateOp.Update` and not restricted to that.
* The past tense suggests that it is a notification of something that happened already.

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
```diff
@@ -1,13 +1,14 @@
...
 
-import mesosphere.marathon.core.task.Task
-import mesosphere.marathon.core.task.bus.TaskStatusObservables.TaskStatusUpdate
-import mesosphere.marathon.state.{ PathId, Timestamp }
+import mesosphere.marathon.core.task.bus.TaskStatusObservables.TaskUpdate
+import mesosphere.marathon.core.task.{ Task, TaskStateChange, TaskStateOp }
+import mesosphere.marathon.state.PathId
 import rx.lang.scala.Observable
 
 object TaskStatusObservables {
```
kolloch: Suggestion: => TaskChangeObservables

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
kolloch: As discussed, if someone outside of this package wants to update a task, the related future should only finish after the task tracker was updated and all steps executed sequentially for this update.

The steps themselves should only be executed after the TaskTracker has processed the related change. Therefore, it probably makes sense to call the step logic from the TaskTracker after it gets a change.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -83,39 +83,48 @@ object Task {
...
-      // failure case: Launch
-      case _: TaskStateOp.Launch =>
-        TaskStateChange.Failure("Unable to handle Launch op on LaunchedEphemeral")
+      case TaskStateOp.ForceExpunge(_) =>
+        TaskStateChange.Expunge(this)
 
-      // failure case: Timeout
-      case TaskStateOp.ReservationTimeout =>
-        TaskStateChange.Failure("Unable to handle Timeout op on LaunchedEphemeral")
+      case TaskStateOp.LaunchEphemeral(task) =>
```
kolloch: Should not be called.

I wonder if we could make it more clear that LaunchEphemeral/Reserve/Revert should never be called on existing tasks.

Maybe subdivide TaskStateOp into two sub sealed traits: NewTaskStateOp, ChangeTaskStateOp. Then implement Task.update only for ChangeTaskStateOp?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -140,16 +149,30 @@ object Task {
...
 
-      // failure case: MesosUpdate
+      // failure case
+      case TaskStateOp.LaunchEphemeral(task) =>
+        TaskStateChange.Failure("Launch should not be handed over to a task instance")
+
+      // failure case
       case _: TaskStateOp.MesosUpdate =>
-        TaskStateChange.Failure("Unable to handle MesosUpdate op on Reserved task")
+        TaskStateChange.Failure("MesosUpdate on Reserved is not allowed")
```
kolloch: In this case I would prefer the wording "unexpected" instead of "not allowed"

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
```diff
@@ -1,22 +1,59 @@
...
+    */
+  def possibleNewState: Option[Task] = None
+}
 
 object TaskStateOp {
-  case class Launch(appVersion: Timestamp, status: Task.Status, networking: Task.Networking) extends TaskStateOp
-  case class MesosUpdate(status: MarathonTaskStatus, now: Timestamp) extends TaskStateOp
-  case object ReservationTimeout extends TaskStateOp
+  /** Launch (aka create) an ephemeral task*/
+  // FIXME (3221): The type should be LaunchedEphemeral but that needs a lot of test adjustments
```
kolloch: Is this tracked in a cleanup issue?

src/main/scala/mesosphere/marathon/core/task/tracker/TaskCreationHandler.scala
```diff
@@ -14,12 +14,12 @@ trait TaskCreationHandler {
     * If the task exists already, the existing task will be overwritten so make sure
     * that you generate unique IDs.
     */
-  def created(task: Task): Future[Task]
+  def created(taskStateOp: TaskStateOp): Future[Unit]
```
kolloch: Is there a cleanup issue for consolidating this usages?

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
```diff
@@ -3,37 +3,43 @@ package mesosphere.marathon.core.task.update.impl.steps
...
-
-    status.getState match {
-      case Terminated(_) =>
-        // Remove from our internal list
-        log.info(s"initiating a scale check for app [${taskId.appId}] after $taskId terminated")
-        schedulerActor ! ScaleApp(taskId.appId)
+  override def processUpdate(update: TaskUpdate): Future[_] = {
+
+    val terminalOrExpungedTask: Option[Task] = {
+      (update.stateOp, update.stateChange) match {
```
kolloch: The `TaskUpdate` does not contain the oldTask option. Otherwise I would say that we need to scale if the oldTask contained a launch, and the new task option does not.

You should have covered all current cases, though.

src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
```diff
@@ -182,6 +181,21 @@ class HealthCheckActor(
 }
 
 object HealthCheckActor {
+  def props(
+    app: AppDefinition,
+    marathonSchedulerDriverHolder: MarathonSchedulerDriverHolder,
+    healthCheck: HealthCheck,
+    taskTracker: TaskTracker,
+    eventBus: EventStream): Props = {
```
kolloch: :+1: for a props method

src/main/scala/mesosphere/marathon/health/MarathonHealthCheckManager.scala
```diff
@@ -63,9 +66,7 @@ class MarathonHealthCheckManager @Inject() (
         Await.result(appRepository.app(appId, appVersion), zkConf.zkTimeoutDuration) match {
           case Some(app: AppDefinition) =>
             val ref = system.actorOf(
-              Props(
-                classOf[HealthCheckActor],
-                app, driverHolder, scheduler, healthCheck, taskTracker, eventBus))
+              HealthCheckActor.props(app, driverHolder, healthCheck, taskTracker, eventBus))
```
kolloch: thanks for making this type safe :+1: (or lightbend??)

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
```diff
@@ -205,7 +205,7 @@ class LaunchQueueModuleTest
 
   after {
     verifyNoMoreInteractions(appRepository)
-    verifyNoMoreInteractions(taskTracker)
+    //    verifyNoMoreInteractions(taskTracker)
```
kolloch: uncomment again?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImplTest.scala
```diff
@@ -22,45 +37,6 @@ class TaskOpProcessorImplTest
   // ignored by the TaskOpProcessorImpl
   val deadline = Timestamp.zero
 
-  test("process noop") {
```
kolloch: why did you delete these tests?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/StatusUpdateActionResolverTest.scala
```diff
@@ -1,57 +0,0 @@
...
-/**
-  * Some specialized tests for statusUpdate action resolving.
-  *
-  * More tests are in [[mesosphere.marathon.tasks.TaskTrackerImplTest]]
-  */
-class StatusUpdateActionResolverTest
-    extends FunSuite with Mockito with GivenWhenThen with ScalaFutures with Matchers {
-  import scala.concurrent.ExecutionContext.Implicits.global
-
-  test("an update for a non-existing tasks is mapped to fail") {
```
kolloch: why did you delete this test or is it covered elsewhere?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: Thanks for the suggestion, totally makes sense to put this one level deeper. I've bundled step processing into an own class and adjusted tha `TaskOpProcessorImpl`.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImpl.scala
meichstedt: fixed.

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
meichstedt: Agree, it's easer to understand. My original renaming just removed the `Status` in between. To keep the naming consistent, I've also renamed vals containing a `TaskUpdate` to `taskUpdate`

src/main/scala/mesosphere/marathon/core/task/bus/TaskStatusObservables.scala
meichstedt: Makes sense. Also: `InternalTaskChangeEventStream`

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: I'm absolutely with you. It bothered me several times that these case classes are handled differently, so yes, nice idea. Yet I'd like to defer this to a later clean up step as I fear that there are a lot of tedious test suite adjustments needed for this. I have added a sub task to #3311 for this.

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: done.

src/main/scala/mesosphere/marathon/core/task/TaskStateOp.scala
meichstedt: Yes, see #3311

src/main/scala/mesosphere/marathon/core/task/tracker/TaskCreationHandler.scala
meichstedt: #3311

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
meichstedt: done.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImplTest.scala
meichstedt: There is no `TaskStateOp.Noop` or adequate. The case when NoChange is needed is handled in _process statusUpdate with NoChange_

src/test/scala/mesosphere/marathon/core/task/tracker/impl/StatusUpdateActionResolverTest.scala
meichstedt: That's now located in `TaskStateOpResolverTest`

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
meichstedt: Yes, the oldTask is _hidden_ inside of `TaskChanged.stateChange` and is set if the stateChange is an Expunge. The match is needed because there's no Expunge in case it's a resident task. So basically we have either
- some StateOp that led to an Expunge OR
- a MesosStatusUpdate that led to either an Expunge or an Update(from LaunchedOnReservation to Reserved)

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
```diff
@@ -23,15 +25,129 @@ class TaskStateOpResolverTest
...
-    val appId = PathId("/app")
-    val taskId = Task.Id.forApp(appId)
-    f.taskTracker.task(taskId) returns Future.successful(None)
+    f.taskTracker.task(f.notExistingTaskId) returns Future.successful(None)
 
     When("resolve is called")
-    val stateChange = f.stateOpResolver.resolve(TaskStateOp.ForceExpunge(taskId)).futureValue
+    val stateChange = f.stateOpResolver.resolve(TaskStateOp.ForceExpunge(f.notExistingTaskId)).futureValue
+
+    Then("getTaskAsync is called")
```
kolloch: consistent with the rest of the tests -- but the method is not called getTaskAsync anymore.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
```diff
@@ -23,15 +25,129 @@ class TaskStateOpResolverTest
...
+    verify(f.taskTracker).task(f.notExistingTaskId)
+
+    And("the future fails")
+    stateChange shouldBe a[TaskStateChange.Failure]
+
+    And("there are no more interactions")
+    f.verifyNoMoreInteractions()
+  }
+
+  test("ForceExpunge fails if task does not exist") {
```
kolloch: What is the difference to the last test?

Also, if we use ForceExpunge as a revert and we detect that the task does not exist. Is that really an error? (Corner-case, ignore if you want for now)

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
```diff
@@ -23,15 +25,129 @@ class TaskStateOpResolverTest
...
+    And("there are no more interactions")
+    f.verifyNoMoreInteractions()
+  }
+
+  test("LaunchOnReservation fails if task does not exist") {
+    val f = new Fixture
+    Given("a non existing taskId")
+    f.taskTracker.task(f.notExistingTaskId) returns Future.successful(None)
+
+    When("A ForceExpunge is scheduled with that taskId")
```
kolloch: Text incorrect

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
```diff
@@ -23,15 +25,129 @@ class TaskStateOpResolverTest
...
+    And("there are no more interactions")
+    f.verifyNoMoreInteractions()
+  }
+
+  test("Reserve fails if task already exists") {
+    val f = new Fixture
+    Given("an existing task")
+    f.taskTracker.task(f.existingReservedTask.taskId) returns Future.successful(Some(f.existingReservedTask))
+
+    When("A create is scheduled with that taskId")
```
kolloch: incorrect text

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
```diff
@@ -21,14 +21,14 @@ class TaskOpFactoryHelper(
...
-    val stateOp = TaskStateOp.Create(newTask)
+    val stateOp = TaskStateOp.LaunchEphemeral(newTask)
     TaskOp.Launch(taskInfo, stateOp, oldTask = None, createOperations)
   }
 
   def launchOnReservation(
     taskInfo: Mesos.TaskInfo,
-    newTask: TaskStateOp,
-    oldTask: Task): TaskOp.Launch = {
+    newTask: TaskStateOp.LaunchOnReservation,
```
kolloch: the types read very well, I like it! The parameter newTask should be renamed to taskOp.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImplTest.scala
```diff
@@ -411,6 +413,114 @@ class TaskOpProcessorImplTest
     f.verifyNoMoreInteractions()
   }
 
+  test("LaunchEphemeral -> Update results processing update steps") {
+    val f = new Fixture
+    verifyStepsAreProcessed(
+      f = new Fixture,
```
kolloch: `f = f`? or why are you creating a new fixture here?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskOpProcessorImplTest.scala
```diff
@@ -411,6 +413,114 @@ class TaskOpProcessorImplTest
...
+      createTaskState = (appId) => MarathonTestHelper.residentReservedTask(appId),
+      createStateOp = (taskState) => f.stateOpReserve(taskState),
+      createStateChange = (taskState) => TaskStateChange.Update(taskState, oldState = None))
+  }
+
+  def verifyStepsAreProcessed(
+    f: Fixture,
+    createTaskState: (PathId) => Task,
+    createStateOp: (Task) => TaskStateOp,
+    createStateChange: (Task) => TaskStateChange): Unit = {
```
kolloch: Subjective:

I think that I (personally) would have understood the test better if the parameters were called

* initialTaskState
* stateOp
* expectedStateChange

The "create" prefix would not be necessary for me. A diferrentiation between expected and given state would help.

kolloch: And/Or maybe documentation for this method?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
meichstedt: right, fixed

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
meichstedt: There's no difference, so I've removed it. As for the Force-Expunge, you're right. I think a ForceExpunge on a non-existent task should result in a NoChange

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
meichstedt: fixed.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/TaskStateOpResolverTest.scala
meichstedt: fixed

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
```diff
@@ -23,15 +28,15 @@ object TaskTrackerActor {
   /** Forward an update operation to the child [[TaskUpdateActor]]. */
   private[impl] case class ForwardTaskOp(deadline: Timestamp, taskId: Task.Id, taskStateOp: TaskStateOp)
 
-  /** Describes where and what to send after an update event has beend processed by the [[TaskTrackerActor]]. */
+  /** Describes where and what to send after an update event has been processed by the [[TaskTrackerActor]]. */
   private[impl] case class Ack(initiator: ActorRef, msg: Any = ()) {
```
kolloch: Move to the side where it is used

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
```diff
@@ -125,13 +131,18 @@ private class TaskTrackerActor(
...
-        becomeWithUpdatedApp(task.appId)(task.taskId, newTask = Some(task))
-        ack.sendAck()
-
-      case msg @ TaskTrackerActor.TaskRemoved(taskId, ack) =>
-        becomeWithUpdatedApp(taskId.appId)(taskId, newTask = None)
-        ack.sendAck()
+      case msg @ TaskTrackerActor.StateChanged(change) =>
+        val taskAfterChange = change.stateChange match {
+          case TaskStateChange.Update(task, _) => Some(task)
+          case _                               => None
```
kolloch: This catch-all case is very dangerous. What about "Fail" And "Noop"?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
```diff
@@ -125,13 +131,18 @@ private class TaskTrackerActor(
...
+        }
+        becomeWithUpdatedApp(change.appId)(change.taskId, newTask = taskAfterChange)
+        val originalSender = sender()
+
+        import context.dispatcher
+        updateStepProcessor.process(change).map { _ =>
+          originalSender ! (())
```
kolloch: you need to handle the failure case here. (and I'd prefer foreach for side effects)

I am not sure how to deal with errors here. I think for now, we can log errors and proceed anyway, since we use "ContinueOnError" on the steps anyway and are apparently fine with that strategy.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerUpdateStepProcessorImpl.scala
```diff
@@ -0,0 +1,41 @@
...
+import mesosphere.marathon.core.task.bus.TaskChangeObservables.TaskChanged
+import mesosphere.marathon.core.task.tracker.TaskTrackerUpdateStepProcessor
+import mesosphere.marathon.core.task.update.TaskUpdateStep
+import mesosphere.marathon.metrics.{ Metrics, MetricPrefixes }
+import mesosphere.marathon.metrics.Metrics.Timer
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.{ ExecutionContext, Future }
+
+private[tracker] class TaskTrackerUpdateStepProcessorImpl(
```
kolloch: I like this separate thanks. 

A documentation oneliner, please?

kolloch:
Strange, I looked at the specific commits and commented there. I hope that the comments are now not lost. Important stuff that I remember:

* TaskTrackerActor: in handling a StateChanged, the match on TaskStateChange has a default clause. That's dangerous, since it would cause a task removal for Failure/Noop.

Style issue:

* TaskTrackerActor.Ack is not used in TaskTrackerActor and can be made more specific. Then you can avoid using a redundant tuple in TaskOpProcessorImpl.

kolloch:
Nearly there...

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
meichstedt: re-using it it in the TaskTracker again, so it's fine

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
meichstedt: nice one. I've changed this to only `become` in case of Update/Expunge

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
meichstedt: foreaching now and recovering with a `log.warn`

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerUpdateStepProcessorImpl.scala
meichstedt: A two-liner!

meichstedt:
The comments are still visible, and I double-checked against the notification emails. Some comments have become irrelevant because the related (test) code has been changed by removing them (e.g. `verifyStepsAreProcessed` in `TaskOpProcessorImpl` has been removed because the steps are now processed via the Stepprocessor and I verify that that is called in `TaskTrackerActorTest`)

meichstedt:
Thanks @kolloch, I addressed your comments!

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
```diff
@@ -131,16 +132,29 @@ private class TaskTrackerActor(
...
+      case msg @ TaskTrackerActor.StateChanged(change, ack) =>
+        change.stateChange match {
+          case TaskStateChange.Update(task, _) =>
+            becomeWithUpdatedApp(task.appId)(task.taskId, newTask = Some(task))
+
+          case TaskStateChange.Expunge(task) =>
+            becomeWithUpdatedApp(task.appId)(task.taskId, newTask = None)
+
+          case _: TaskStateChange.NoChange |
+            _: TaskStateChange.Failure =>
```
kolloch: :+1 for listing all choices explicitly

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskTrackerActor.scala
```diff
@@ -29,14 +30,14 @@ object TaskTrackerActor {
   private[impl] case class ForwardTaskOp(deadline: Timestamp, taskId: Task.Id, taskStateOp: TaskStateOp)
 
   /** Describes where and what to send after an update event has been processed by the [[TaskTrackerActor]]. */
-  private[impl] case class Ack(initiator: ActorRef, msg: Any = ()) {
+  private[impl] case class Ack(initiator: ActorRef, stateChange: TaskStateChange) {
     def sendAck(): Unit = {
-      initiator ! msg
+      initiator ! stateChange
```
kolloch: I wonder if a "Failed" should be mapped to Status.Failed. Combined with an ask, the Future with then fail instead of just returning a change that includes the failure. That would be closer to what happened before.

What do you think?

meichstedt: sgtm – changed it.

kolloch:
Test failures -- otherwise looks good.

Can you investigate? Test failures might relate to timing issues.

aquamatthias:
Thanks!

```




                             ________
                            `---.     `.
                                 \      `.
                                  )_______`.    
                                .'        //`---...___
                               /         ||    // ||  `-._
                            )`-|   =//=  ||   ||  ||  / ).`.
       _............_       ).-|         ||   ||  `........'`-._   (o)
    .-'         `.----`.   _...'.....__  ||   ||  _____      ||-\__.'
  .' | Inter- |  ).---.)  /_______..--'  ||   ||  -----    _ ||_/
.'_  |Plantery|  ||   ||  `-------'      ||   ||   =\\=    \_.'
|  | | Travel |  |'---'|    )`-|         ||   ||       _..-'
'--'_____________|_____|    ).-| =\\=    ||    \\  _.-' || 
|[]--------------/     /   __==\          \\   _.-'    /o'\
 \     .--.     /    _...--''   '-.__......_.-'        \__/
  `-._//'o\\___.'---''             \     .'
  LGB  \__/  -'                    /    /
                              ___.'    /
                              `-------'


```

PR #3543, opened 2016-03-22
Resolves #3456 add wipe flag to task delete endpoints

### LINKED ISSUE #3456: Resident Tasks: Manually kill/wipe task including reservation
Normal kills on resident tasks only cause a transition to `Task.Reserved`. The reservation/volumes are not affected. That means that the number of reservations typically corresponds to the largest instance count over the life time of the app. It makes sense to have a parameter for a kill operation (e.g. wipe) that results in the appropriate unreserve/destroy TaskOps for this task.

meichstedt:
Adds `wipe` (default=false) to the tasks deletion endpoints. Either `scale` or `wipe` may be true, never both. In case `wipe=true`, running tasks will be killed, and all given Ids will be expunged from the repo.

That means we will eventually receive `TASK_KILLED` updates, to which the code will react with a `ForceExpunge` TaskStateOp. This way, the expunge will be propagated to all relevant parties even though the task is already removed from the repo.

For later reference: it would be cleaner to mark these tasks as `Garbage` instead of expunging them immediately, and react to `TASK_KILLED` on a `Garbage` task with an Expunge operation.


src/test/scala/mesosphere/marathon/api/v2/TasksResourceTest.scala
```diff
@@ -91,6 +91,51 @@ class TasksResourceTest extends MarathonSpec with GivenWhenThen with Matchers wi
...
+    val exception = intercept[BadRequestException] {
+      taskResource.killTasks(scale = true, force = false, wipe = true, body = bodyBytes, auth.request)
+    }
+
+    Then("an exception should occur")
+    exception.getMessage shouldEqual "You cannot use scale and wipe at the same time."
+  }
+
+  // FIXME (3221): breaks – why?
+  ignore("killTasks with wipe delegates to taskKiller with wipe value") {
```
aquamatthias: Did you test this on a real Marathon instance?
Is that "just" a problem of the test setup?

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -20,17 +21,27 @@ class TaskKiller @Inject() (
     val authorizer: Authorizer) extends AuthResource {
 
   def kill(appId: PathId,
-           findToKill: (Iterable[Task] => Iterable[Task]))(implicit identity: Identity): Future[Iterable[Task]] = {
+           findToKill: (Iterable[Task] => Iterable[Task]),
+           wipe: Boolean = false)(implicit identity: Identity): Future[Iterable[Task]] = {
```
kolloch: add documentation. wipe = removing volumes/reservations

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -20,17 +21,27 @@ class TaskKiller @Inject() (
...
     result(groupManager.app(appId)) match {
       case Some(app) =>
         checkAuthorization(UpdateApp, app)
 
-        val tasks = taskTracker.appTasksLaunchedSync(appId)
-        val toKill = findToKill(tasks)
-        if (toKill.nonEmpty)
-          service.killTasks(appId, toKill)
+        import scala.concurrent.ExecutionContext.Implicits.global
+        taskTracker.appTasks(appId).map { allTasks =>
```
kolloch: :+1: for using an async method now

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -20,17 +21,27 @@ class TaskKiller @Inject() (
...
+          val allFound = findToKill(allTasks)
+
+          if (wipe) allFound.foreach { task =>
+            stateOpProcessor.process(TaskStateOp.ForceExpunge(task.taskId))
```
kolloch: you need to wait for the future, i.e. chain the processing step so that you only acknowledge the API call once the task op has been processed.

kolloch:
TODO: 

- [ ] address comments above
- [ ] rebase on master after merging #3533 
- [ ] add REST API documentation, mention that tasks will appear gone immediately, even if they are still running or the reservation still exists
- [ ] create an issue that improves the restriction, e.g. tasks will change from running to waiting and then only be gone from the API once the reservation got unreserved.

src/test/scala/mesosphere/marathon/api/v2/TasksResourceTest.scala
kolloch: I would be interested as well

PR #3547, opened 2016-03-22
Fixes #3530 by using the canonical path of a nested group or app.

### LINKED ISSUE #3530: apps shown last identifier to id, rather than entire path
If I add app "/a/b/c", when I get apps it only shows "/c"

aquamatthias:


src/main/scala/mesosphere/marathon/api/v2/json/GroupUpdate.scala
```diff
@@ -64,14 +64,16 @@ object GroupUpdate {
...
+  def validNestedGroupUpdateWithBase(base: PathId): Validator[GroupUpdate] = validator[GroupUpdate] { group =>
     group is notNull
 
     group.version is theOnlyDefinedOptionIn(group)
     group.scaleBy is theOnlyDefinedOptionIn(group)
 
     group.id is valid
-    group.apps is valid
-    group.groups is valid
+    group.apps is optional(every(AppDefinition.validNestedAppDefinition(group.id.fold(base)(_.canonicalPath(base)))))
```
gkleiman: Adding `val canonicalGroupId = group.id.fold(base)(_.canonicalPath(base))` and then using it would increase readability and remove code duplication.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -517,7 +517,6 @@ object AppDefinition {
     */
   def validNestedAppDefinition(base: PathId): Validator[AppDefinition] = validator[AppDefinition] { app =>
     app.id is PathId.validPathWithBase(base)
-    app.dependencies is every(PathId.validPathWithBase(base))
```
gkleiman: `validBasicAppDefinition` doesn't validate the app dependencies. If you remove this line, who validates the paths of the dependencies?

src/main/scala/mesosphere/marathon/api/v2/json/GroupUpdate.scala
aquamatthias: Unfortunately: since this is a macro, it does not close over external references.
So it has to be part of the validator.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -213,13 +213,11 @@ object Group {
   private def validNestedGroup(base: PathId): Validator[Group] =
     validator[Group] { group =>
       group.id is validPathWithBase(base)
-      group.apps is every(AppDefinition.validNestedAppDefinition(base))
+      group.apps is every(AppDefinition.validNestedAppDefinition(group.id.canonicalPath(base)))
       group is noAppsAndGroupsWithSameName
       (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
       group is validPorts
-
-      group.dependencies is every(validPathWithBase(base))
```
gkleiman: Same as before, where are dependencies validated?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
aquamatthias: What can we validate? I removed the validation, because I don't see a reason.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -210,4 +211,15 @@ object Validation {
       if (test(value)) Success else RuleViolation(value, constraint, None)
     }
   }
+
+  /**
+    * For debugging purposes only.
+    * Since the macro removes all logging statements in the validator itself.
```
gkleiman: A usage example would be very nice =).

aquamatthias:
@gkleiman added a test for nested group and an example for the `info` method.

gkleiman:
:ship:

Issue #1310, opened 2015-03-18
Remove deprecated ZooKeeper leader election library


ConnorDoyle:
This twitter library is deprecated and should be removed: https://github.com/twitter-archive/scala-zookeeper-client

kolloch:
+1

Most build failures currently are because of resolve failures involving the twitter maven repository.

air:
Are there other impacts caused by this? Trying to understand for prioritization. Thanks.

janisz:
Closed by https://github.com/mesosphere/marathon/pull/3777

jeschkies:
Old Twitter leader election has been removed with https://github.com/mesosphere/marathon/commit/b476f32a9dae14674bd758ff650c7fd2fe6f778e.

PR #3553, opened 2016-03-22
WIP: 2419 - implement leader election with Apache Curator

sttts:
Fixing https://github.com/mesosphere/marathon/issues/1310.

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -268,6 +268,7 @@ object Dependencies {
     datadog % "compile",
     marathonApiConsole % "compile",
     wixAccord % "compile",
+    curator % "compile",
 
     // test
     Test.diffson % "test",
@@ -306,6 +307,7 @@ object Dependency {
     val DataDog = "1.1.3"
     val Logback = "1.1.3"
     val WixAccord = "0.5"
+    val Curator = "2.10.0"
 
     // test deps versions
     val Mockito = "1.9.5"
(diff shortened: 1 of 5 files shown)


sttts:
@aquamatthias @gkleiman ptal

This implements the twitter common Candidate interface using the curator leader election, i.e. it's a drop-in for the old algorithm. Further refactoring of the whole leader election code makes probably sense, but that's not (yet) part of this PR.

Of course, the old and the new leader election are not compatible. Do you think it's worth to invest in a more graceful switch? At least, we could implement something along these lines:

- check whether leader-v2 Zookeeper key exists
  (a)
    - if not, switch to **old** leader election
    - listen for leader-v2 key creation. If created, switch to **new** leader election
  (b) if yes, use **new** leader election directly
- if a **new** marathon becomes leader (using the old leader election)
  - check whether there are **old** marathons in the Zookeeper group (we could mark the new ones with a special postifix in the supplier key)
  - if not, create leader-v2 key and switch to the new election code

aquamatthias:
@sttts 

1) Regarding backward compatibility: very good point.Learning from the past, leader election has been a big problem and can lead to even bigger problems if done correctly. To complicate things, we also have something like an AbdicateOnConnectionLossActor, which probably only makes sense in combination with twitters library (because it does not handle that case).
Pros of this approach:
- migration to a newer version is supported
Cons:
- we have to maintain 2 versions
- added complexity, because we have to check with every leader election, which protocol to use
- it will only solve the forward problem (a rollback to an older Marathon version would not work)
My gut feeling is: if we can make the leader election implementation a drop in replacement (curator/twitter), it is probably worth the effort. What do you think?

2) I really would encourage to refactor the structure we have. 

3) Can you make some tests, if the new leader election library is resistant against zk connection problems or timeouts. This is the main reason why we do this.

sttts:
@aquamatthias as far as I see `AbdicateOnConnectionLossActor` does not harm. I would leave it in as long as we maintain compatibility with the old leader election.

sttts:
Superseeded by #3777

PR #3556, opened 2016-03-23
Make the defaults consistent with the code.

aquamatthias:


meichstedt:
Thanks @aquamatthias!

Issue #3568, opened 2016-03-23
Resident Tasks: Improve #3456 Wipe Endpoint


meichstedt:
#3456 added a `wipe` flag to allow to forcefully expunge (resident) tasks from the repo before killing them. Several sub tasks have not been resolved due to time contraints:
- [ ] properly document the flag and its side effects
- [ ] improve the task life cycle when wiping reservations and persistent volumes. The tasks should change state to something like `Garbage` to allow for explicit Unreserve/Destroy operations and to allow for the UI displaying them as long as they're actually existent (and their resources reserved).

meichstedt:
Note: This issue has been migrated to https://jira.mesosphere.com/browse/MARATHON-1931. For more information see https://groups.google.com/forum/#!topic/marathon-framework/khtvf-ifnp8.

PR #3570, opened 2016-03-23
Resolves #3456 added endpoint for manual deletion of residents

### LINKED ISSUE #3456: Resident Tasks: Manually kill/wipe task including reservation

meichstedt:
Resolves #3456

Adds `wipe` (default=false) to the tasks deletion endpoints. Either `scale` or `wipe` may be true, never both. In case `wipe=true`, running tasks will be killed, and all given Ids will be expunged from the repo.

That means we will eventually receive `TASK_KILLED` updates, to which the code will react with a `ForceExpunge` TaskStateOp. This way, the expunge will be propagated to all relevant parties even though the task is already removed from the repo.

For later reference: it would be cleaner to mark these tasks as `Garbage` instead of expunging them immediately, and react to `TASK_KILLED` on a `Garbage` task with an Expunge operation. Follow-up subtasks are tracked in #3568

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
+++ b/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
@@ -2,14 +2,16 @@ package mesosphere.marathon.api
 
 import javax.inject.Inject
 
+import com.twitter.util.NonFatal
 import mesosphere.marathon.core.task.{ TaskStateOp, Task }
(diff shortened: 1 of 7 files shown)


aquamatthias:
Thanks!

Issue #3367, opened 2016-03-01
Relative paths for dependencies not working anymore


schneidexe:
With marathon 0.15.* relative dependencies for deployed groups seem not to work anymore. Re. to https://mesosphere.github.io/marathon/docs/application-groups.html#dependencies the following should work:

```
{
  "id": "/db-web-group",
  "apps": [
    {
      "id": "db",
      ...
    },
    {
      "id": "web",
      "dependencies": ["../db"],
    }
  ]
}
```

But I get:

```
{"message":"Object is not valid","details":[{"attribute":"apps[1].dependencies[0]","error":"path of parent should be absolute"}]}
```

With absolute paths it works:

```
{
  "id": "/db-web-group",
  "apps": [
    {
      "id": "db",
      ...
    },
    {
      "id": "web",
      "dependencies": ["/db-web-group/db"],
    }
  ]
}
```

The relative paths are a really helpful feature cause I can easily chnage my group name w/o touching the rest of the json, so IMHO it should be fixed. If this was changed intentionally please update at the docs accordingly.

alexanderweber:
@aquamatthias according to https://github.com/mesosphere/marathon/blob/96f685571a17f7525fa9e600d59fff21d3f56851/src/main/scala/mesosphere/marathon/state/PathId.scala#L137 pathIds cannot be relative, if and only if the path contains a parent. This rule is applied to application dependencies. So `"dependencies": ["mainApp"]` is allowed, while `"dependencies": ["../group2/mainApp"]` is not. Should we expand the relative path behaviour here? I would take out the `if (pathId.parent.absolute)` check out of dependecy related pathId validation then. It would become a special case.

@schneidexe thank you for pointing out the issue. The documentation you are mentioning provides examples for group dependencies, not application dependencies. The `"dependencies"` property is defined within a group there, not within the app.

Your example defines dependencies on app level though. Additionally, both apps are on the same folder level, namely `/db-web-group`. That is why the correct definition should be:
```Json
{
  "id": "/db-web-group",
  "apps": [
    {
      "id": "db",
      "cmd": "foo"
    },
    {
      "id": "web",
      "cmd": "bar",
      "dependencies": ["db"]
    }
  ]
}
```
I recall #2543 being similar to your use case. We did address that issue back then, meaning relative paths should work, if apps are _on the same folder level_.

I did create a more complex example though:

```Scala
{
  "id": "/product",
  "groups": [
    {
      "id": "database",
      "apps": [
[...]

PR #3578, opened 2016-03-24
Add v0.15.3 changelog

gkleiman:
The following issues in v1.0.0 look very similar and might be dupes:

 - #3367 - Relative paths for dependencies not working anymore
 - #3439 - Relative paths in dependencies should be resolvable.


aquamatthias:
Thanks.

PR #3580, opened 2016-03-24
Fixes #3500 - Add ReadinessChecks to AppDefinition

### LINKED ISSUE #3500: Scheduler Migrations: Add "readinessCheck" configuration to AppDefinition
The app definition will include an additional list of readiness checks.

```javascript
{  // app definition
  “id”: “/dcos-magicdb”,
  // ...
  “readinessChecks”: [
    {
      "protocol": "HTTP", // only HTTP/HTTPS are supported 
      "path": "/v1/plan",
      "portName": “httpApi”, // corresponds to the port name in the
                                         // port definitions
     "intervalSeconds": 20,
     "timeoutSeconds": 20,
     “httpStatusCodesForReady”: [200], // all other response or
                                                             // errors are considered
                                                            // not-ready
     “preserveLastResponse”: true // only provide last response 
                                                     // if this is true, the default
                                                    // is false.
    }
  ]
}
```

To make the API format general, a list of readiness checks is suggested. It is ok for an MVP to only allow a single readiness check if it adds the corresponding validation.

Context: We want to be able to drive scheduler upgrades manually. See [MARATHON-856](https://mesosphere.atlassian.net/browse/MARATHON-856) (private, sorry).

kolloch:


src/main/proto/marathon.proto
```diff
@@ -46,6 +46,23 @@ message HealthCheckDefinition {
   optional uint32 port = 10;
 }
 
+message ReadinessCheckDefinition {
+  enum Protocol {
+    HTTP = 0;
+    HTTPS = 3;
```
gkleiman: why 3 and not 1?

src/main/proto/marathon.proto
```diff
@@ -46,6 +46,23 @@ message HealthCheckDefinition {
...
+  }
+  optional string name = 8;
```
gkleiman: Nit: I know this was a late addition, but it might be nice to make it 1 or to move it to the end.

aquamatthias:
Looks good from my point of view. Any comments to Gastons questions? Please address and I will merge.

src/main/proto/marathon.proto
kolloch: Same numbers as in the health check. I don't care, though.

kolloch:
Ok, I addressed @gkleiman's comments.

aquamatthias:
Thanks!

PR #3581, opened 2016-03-24
Add ReadinessCheckResults classes

gkleiman:


src/main/scala/mesosphere/marathon/core/readiness/ReadinessCheckResult.scala
```diff
@@ -0,0 +1,9 @@
+package mesosphere.marathon.core.readiness
+
+case class ReadinessCheckResult(name: String,
+                                taskId: String,
```
kolloch: taskId should be of type `Task.ID`

gkleiman: Thanks, fixed!

kolloch:
Looks permanent `(marathon/compile:scalastyle) errors exist`

Issue #3562, opened 2016-03-23
Scheduler Migrations: Expose Readiness Check Results


kolloch:
Deployments as returned by the API have to be extended to include the last responses of all readiness checks. This should work at least for the “/v2/deployments” endpoint and the “/v2/apps” endpoint when embedded deployments are requested.

```javascript
 {
    "affectedApps": [
        "/foo"
    ], 
    "currentActions": [
        {
            "action": "RestartApplication", 
            "app": "/foo",
            "readinessResults": [
                {
                    "lastResponse": {
                        "body": "{}", 
                        "contentType": "application/json", 
                        "status": 500
                    }, 
                    "name": "myReadyCheck", 
                    "ready": false, 
                    "taskId": "foo.00b77"
                }
            ]
        }
    ], 
    "currentStep": 2, 
    "id": "41270cfc-edf9-44e5-a224-7fd48f8b440a", 
    "steps": [
        [
            {
                "action": "RestartApplication", 
                "app": "/foo"
            }
        ]
    ], 
    "totalSteps": 2, 
    "version": "2016-03-31T09:00:23.260Z"
}
```

Note: It is not sufficient to query the “/v2/deployments” endpoint alone for determining if user interaction is needed. It is also necessary to inspect the labels of an app to determine whether the app conforms to the DCOS Service Migration API.

Depends on #3502.

Context: We want to be able to drive scheduler upgrades manually. See [MARATHON-856](https://mesosphere.atlassian.net/browse/MARATHON-856) (private, sorry).

PR #3584, opened 2016-03-24
Towards #3562 - Readiness Checks formats

gkleiman:


src/main/scala/mesosphere/marathon/core/readiness/ReadinessCheckResult.scala
```diff
@@ -7,5 +7,4 @@ case class ReadinessCheckResult(name: String,
                                 ready: Boolean,
                                 httpResponse: Option[HttpResponse])
```
kolloch: I don't see why we shouldn't use the same name as in the JSON here. "lastResponse"

aquamatthias: And if we rename this, do we need the explicit Format? 
Can't we just use `Json.format[ReadnissCheckResult]`?

gkleiman:
Updated the PR. The tests do pass locally, but both TeamCity and Travis report build errors, so there must be something wrong. I will look into this tomorrow, but hints are more than appreciated.

cc/ @kolloch

aquamatthias:
Thanks.

Issue #3502, opened 2016-03-15
Scheduler Migrations: Integrate Readiness Checks into Restart Deployment Action


kolloch:
The deployment action Restart has to be adjusted appropriately. 
* For deciding whether old tasks can be killed, readiness checks need to be treated similarly to health checks. For rolling upgrades, old tasks can only be killed after upgraded tasks have become healthy and ready.
* Tasks which are not ready but healthy should never be killed automatically.

Depends on #3500.

Context: We want to be able to drive scheduler upgrades manually. See [MARATHON-856](https://mesosphere.atlassian.net/browse/MARATHON-856) (private, sorry).

PR #3600, opened 2016-03-29
Towards #3502 - Define ReadinessCheckExecutor

kolloch:
* Also make Task.(host)Ports a Seq since their order is important
* Implement deriving the target of readiness checks
* Provide a dummy implementation such that integration is already possible

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/PortDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/PortDefinition.scala
@@ -9,9 +9,7 @@ import scala.collection.immutable.Seq
 case class PortDefinition(port: Int,
                           protocol: String = "tcp",
                           name: Option[String] = None,
-                          labels: Map[String, String] = Map.empty[String, String]) {
-  def fixedHostPort: Option[Int] = if (port != AppDefinition.RandomPortValue) Some(port) else None
-}
+                          labels: Map[String, String] = Map.empty[String, String])
 
 object PortDefinition {
   implicit val portDefinitionValidator = validator[PortDefinition] { portDefinition =>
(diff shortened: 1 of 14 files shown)


src/test/scala/mesosphere/marathon/core/readiness/ReadinessCheckSpecTest.scala
```diff
@@ -0,0 +1,142 @@
...
+        preserveLastResponse = app.readinessChecks.head.preserveLastResponse
+      )
+    )
+  }
+
+  test("multiple readiness check specs") {
+    val f = new Fixture
+
+    Given("an app with two readiness checks and randomly assigned ports")
+    val app = f.appWithMultipleReadinessChecks
```
aquamatthias: Do we want to support multiple readiness checks (in step 1)?

aquamatthias:
LGTM.

src/main/scala/mesosphere/marathon/state/PortDefinition.scala
```diff
@@ -9,7 +9,9 @@ import scala.collection.immutable.Seq
 case class PortDefinition(port: Int,
                           protocol: String = "tcp",
                           name: Option[String] = None,
-                          labels: Map[String, String] = Map.empty[String, String])
+                          labels: Map[String, String] = Map.empty[String, String]) {
+  def fixedHostPort: Option[Int] = if (port != AppDefinition.RandomPortValue) Some(port) else None
```
aquamatthias: If we store the AppDefinition, the random port value is replaced, right?

src/test/scala/mesosphere/marathon/core/readiness/ReadinessCheckSpecTest.scala
kolloch: at least in this code, I chose to do that. I don't mind if other code doesn't and we prevent this by validation, though.

src/main/scala/mesosphere/marathon/state/PortDefinition.scala
kolloch: good point, I hate this API. Let me think about this...

kolloch:
Ok, I hopefully fixed the host port handling now.

aquamatthias:
Thanks!

aquamatthias:
Waiting for the tests.

PR #3604, opened 2016-03-29
Fixes #3575 - Get NetworkInfo from Mesos task updates

### LINKED ISSUE #3575: Container IP not correctly display in 0.15.0 and above
I'm using a Docker network plugin that assigns a routable IP address per container on a Docker container.

In Marathon 0.14.2 release the IP address of the container was correctly displayed

```
Host 172.24.197.102
IP Addresses 192.168.0.0
Ports [31830]
Endpoints 172.24.197.102:31830
Service Discovery n/a
Status Started
```

From Marathon 0.15.0 and above, the IP address of the container is no longer displayed:

```
Host 172.24.197.103
Ports [31781]
Endpoints 172.24.197.103:31781
Service Discovery n/a
Status Started
```

Why is the actual container IP no longer available in Marathon?

I'm running Mesos 0.28 on CentOS 7.

gkleiman:
**NOTE**: IP-per-Task is also broken in Marathon v0.15.x. We might need to backport this commit.

- Recent versions of Mesos send the NetworkInfo messages for all tasks,
  even if no ip address was requested.
- The NetworkInfo messages are wrapped inside of the ContainerStatus message.
- Marathon should always store the NetworkInfo messages and expose the
  IpAddresses through the tasks API. Clients (including the UI) can then decide
  if they want to use the agent info or one of the ip addresses.

Changes in order to always return the IpAddreses:

- Refactored the Task.Networking traits. It is now a case class that stores both
  host ports and NetworkInfo messages.
- Changed Task.update in order to always extract the NetworkInfo messages from
  the update messages and store them using new Task.Networking case class.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -119,13 +119,16 @@ trait Formats
     )
 
     val launched = task.launched.map { launched =>
-      base ++ Json.obj (
-        "startedAt" -> launched.status.startedAt,
-        "stagedAt" -> launched.status.stagedAt,
-        "ports" -> launched.ports,
-        "ipAddresses" -> launched.ipAddresses,
-        "version" -> launched.appVersion
-      )
+      launched.ipAddresses.foldLeft(
+        base ++ Json.obj (
+          "startedAt" -> launched.status.startedAt,
+          "stagedAt" -> launched.status.stagedAt,
+          "ports" -> launched.hostPorts,
+          "version" -> launched.appVersion
+        )
+      ){
+        case (launchedJs, ipAddresses) => launchedJs ++ Json.obj("ipAddresses" -> ipAddresses)
+      }
     }.getOrElse(base)
 
     val reservation = task.reservationWithVolumes.map { reservation =>
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -5,10 +5,12 @@ import mesosphere.marathon.Protos.MarathonTask
 import mesosphere.marathon.core.task.bus.MarathonTaskStatus
 import mesosphere.marathon.core.task.tracker.impl.TaskSerializer
 import mesosphere.marathon.state.{ AppDefinition, PathId, PersistentVolume, Timestamp }
-import org.apache.mesos.Protos.TaskState
+import org.apache.mesos.Protos.{ NetworkInfo, TaskState }
 import org.apache.mesos.Protos.TaskState._
 import org.apache.mesos.{ Protos => MesosProtos }
 import org.slf4j.LoggerFactory
+
+import scala.collection.immutable.Seq
 
 //scalastyle:off number.of.types
 /**
@@ -52,15 +54,11 @@ sealed trait Task {
     }
   }
 
-  def ipAddresses: Iterable[MesosProtos.NetworkInfo.IPAddress] = launched.map(_.ipAddresses).getOrElse(Iterable.empty)
-
-  def effectiveIpAddress(app: AppDefinition): String = {
-    val maybeContainerIp: Option[String] = ipAddresses.map(_.getIpAddress).headOption
-
-    maybeContainerIp match {
-      case Some(ipAddress) if app.ipAddress.isDefined => ipAddress
-      case _ => agentInfo.host
-    }
+  def effectiveIpAddress(app: AppDefinition): Option[String] = {
+    if (app.ipAddress.isDefined)
+      launched.flatMap(_.ipAddresses).flatMap(_.headOption).map(_.getIpAddress)
+    else
+      Some(agentInfo.host)
   }
 }
 
@@ -73,13 +71,13 @@ object Task {
       agentInfo: AgentInfo,
       appVersion: Timestamp,
       status: Status,
-      networking: Networking) extends Task {
+      hostPorts: Seq[Int]) extends Task {
 
     import LaunchedEphemeral.log
 
     override def reservationWithVolumes: Option[Reservation] = None
 
-    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, networking))
+    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, hostPorts))
 
     private[this] def hasStartedRunning: Boolean = status.startedAt.isDefined
 
@@ -149,8 +147,8 @@ object Task {
     override def launched: Option[Launched] = None
 
     override def update(update: TaskStateOp): TaskStateChange = update match {
-      case TaskStateOp.LaunchOnReservation(_, appVersion, status, networking) =>
-        val updatedTask = LaunchedOnReservation(taskId, agentInfo, appVersion, status, networking, reservation)
+      case TaskStateOp.LaunchOnReservation(_, appVersion, status, hostPorts) =>
+        val updatedTask = LaunchedOnReservation(taskId, agentInfo, appVersion, status, hostPorts, reservation)
         TaskStateChange.Update(newState = updatedTask, oldState = Some(this))
 
       case _: TaskStateOp.ReservationTimeout =>
@@ -181,14 +179,14 @@ object Task {
       agentInfo: AgentInfo,
       appVersion: Timestamp,
       status: Status,
-      networking: Networking,
+      hostPorts: Seq[Int],
       reservation: Reservation) extends Task {
 
     import LaunchedOnReservation.log
 
     override def reservationWithVolumes: Option[Reservation] = Some(reservation)
 
-    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, networking))
+    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, hostPorts))
 
     private[this] def hasStartedRunning: Boolean = status.startedAt.isDefined
 
@@ -376,20 +374,18 @@ object Task {
 
   /**
     * Represents a task which has been launched (i.e. sent to Mesos for launching).
+    *
+    * @param hostPorts sequence of ports in the Mesos Agent allocated to the task
     */
   case class Launched(
       appVersion: Timestamp,
       status: Status,
-      networking: Networking) {
+      hostPorts: Seq[Int]) {
 
     def hasStartedRunning: Boolean = status.startedAt.isDefined
 
-    def ports: Iterable[Int] = networking.ports
-
-    def ipAddresses: Iterable[MesosProtos.NetworkInfo.IPAddress] = networking match {
-      case list: NetworkInfoList => list.addresses
-      case _                     => Iterable.empty
-    }
+    def ipAddresses: Option[Seq[MesosProtos.NetworkInfo.IPAddress]] =
+      status.mesosStatus.flatMap(MesosStatus.ipAddresses)
   }
 
   /**
@@ -413,34 +409,6 @@ object Task {
     startedAt: Option[Timestamp] = None,
     mesosStatus: Option[MesosProtos.TaskStatus] = None)
 
-  /** Info on how to reach the task in the network. */
-  sealed trait Networking {
-    def ports: Iterable[Int]
-  }
-
-  /** The task is reachable via host ports which are bound to [[AgentInfo#host]]. */
-  case class HostPorts(ports: Iterable[Int]) extends Networking
-  object HostPorts {
-    def apply(ports: Int*): HostPorts = HostPorts(ports)
-  }
-
-  /**
-    * The task has been launched with one-IP-per-task settings. The ports can be discovered
-    * by inspecting the [[mesosphere.marathon.state.DiscoveryInfo]] in the [[mesosphere.marathon.state.AppDefinition]].
-    */
-  case class NetworkInfoList(networkInfoList: Iterable[MesosProtos.NetworkInfo]) extends Networking {
-    import scala.collection.JavaConverters._
-    def addresses: Iterable[MesosProtos.NetworkInfo.IPAddress] = networkInfoList.flatMap(_.getIpAddressesList.asScala)
-    override def ports: Iterable[Int] = Iterable.empty
-  }
-  object NetworkInfoList {
-    def apply(networkInfoList: MesosProtos.NetworkInfo*): NetworkInfoList = NetworkInfoList(networkInfoList)
-  }
-
-  case object NoNetworking extends Networking {
-    override def ports: Iterable[Int] = Iterable.empty
-  }
-
   object Terminated {
     def isTerminated(state: TaskState): Boolean = state match {
       case TASK_ERROR | TASK_FAILED | TASK_FINISHED | TASK_KILLED | TASK_LOST => true
@@ -448,5 +416,16 @@ object Task {
     }
 
     def unapply(state: TaskState): Option[TaskState] = if (isTerminated(state)) Some(state) else None
+  }
+
+  object MesosStatus {
+    def ipAddresses(mesosStatus: MesosProtos.TaskStatus): Option[Seq[MesosProtos.NetworkInfo.IPAddress]] = {
+      import scala.collection.JavaConverters._
+      if (mesosStatus.hasContainerStatus && mesosStatus.getContainerStatus.getNetworkInfosCount > 0)
+        Some(
+          mesosStatus.getContainerStatus.getNetworkInfosList.asScala.flatMap(_.getIpAddressesList.asScala).toList
+        )
+      else None
+    }
   }
 }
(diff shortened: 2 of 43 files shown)


gkleiman:
I tested this using the `net-modules` repo and also added a checks to an integration test, to make sure that we always return ip addresses.

src/main/scala/mesosphere/marathon/event/Events.scala
```diff
@@ -223,7 +223,7 @@ case class MesosStatusUpdateEvent(
   message: String,
   appId: PathId,
   host: String,
-  ipAddresses: Seq[org.apache.mesos.Protos.NetworkInfo.IPAddress],
```
kolloch: Isn't the order relevant?

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
```diff
@@ -257,7 +257,7 @@ object MarathonTestHelper {
         status = Task.Status(
           stagedAt = now
         ),
-        networking = Task.HostPorts(Seq(1, 2, 3))
+        networking = Task.Networking(Seq(1, 2, 3))
```
kolloch: I'd prefer named arguments here. (ports = )

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -86,11 +86,13 @@ object Task {
...
     override def update(update: TaskStateOp): TaskStateChange = update match {
       // case 1: now running
-      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
+      case TaskStateOp.MesosUpdate(_, taskStatus: MarathonTaskStatus.Running, now) if !hasStartedRunning =>
         val updated = copy(
+          networking = networking.copy(networkInfos = taskStatus.networkInfos),
           status = status.copy(
             startedAt = Some(now),
-            mesosStatus = mesosStatus))
+            mesosStatus = taskStatus.mesosStatus)
```
kolloch: aren't we essentially storing the network infos twice in our task? Once as part of the status and once as part of the networking? same potentially in the protobuf?

Or am I misunderstanding this?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -101,7 +103,10 @@ object Task {
       case TaskStateOp.MesosUpdate(_, taskStatus, now) =>
         updatedHealthOrState(status.mesosStatus, taskStatus.mesosStatus) match {
           case Some(newStatus) =>
-            val updatedTask = copy(status = status.copy(mesosStatus = Some(newStatus)))
+            val updatedTask = copy(
+              networking = networking.copy(networkInfos = taskStatus.networkInfos),
```
kolloch: Just to be sure, 
* it is not possible that these networkInfos change without a Mesos State change, correct?
* all networkInfos are sent with every statusUpdate? Not only if these change?

src/main/scala/mesosphere/marathon/core/task/bus/MarathonTaskStatus.scala
```diff
@@ -1,13 +1,18 @@
...
+import org.apache.mesos.Protos.{ NetworkInfo, TaskStatus }
 
 sealed trait MarathonTaskStatus {
   def terminal: Boolean = false
-
   def mesosStatus: Option[TaskStatus]
-  def mesosHealth: Option[Boolean] = mesosStatus.flatMap { status =>
-    if (status.hasHealthy) Some(status.getHealthy) else None
+
+  def networkInfos: Seq[NetworkInfo] = {
```
kolloch: In principal, this could also be part of `Task.Status`

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
```diff
@@ -110,7 +110,7 @@ object MarathonTestHelper {
       .addResources(memResource)
       .addResources(diskResource)
 
-    portsResource.foreach(offerBuilder.addResources(_))
+    portsResource.foreach(offerBuilder.addResources)
```
kolloch: why? It's not related to this patch and I don't think that we have a team agreement what we like better.

kolloch:
Maybe we should discuss whether to really duplicate the network info information in our task state. All other issues are just minor.

gkleiman:
@kolloch thanks for the review!

Updated the PR to address your feedback.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -115,15 +115,15 @@ trait Formats
     val base = Json.obj(
       "id" -> task.taskId,
       "slaveId" -> task.agentInfo.agentId,
-      "host" -> task.agentInfo.host
+      "host" -> task.agentInfo.host,
+      "ipAddresses" -> task.ipAddresses
```
kolloch: So it is important to you to have an empty list there disregarding the task state instead of not having the ipAddresses field in case the task is not "launched"?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -73,24 +75,25 @@ object Task {
...
-    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, networking))
+    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, hostPorts))
 
     private[this] def hasStartedRunning: Boolean = status.startedAt.isDefined
 
     //scalastyle:off cyclomatic.complexity method.length
     override def update(update: TaskStateOp): TaskStateChange = update match {
       // case 1: now running
-      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
+      case TaskStateOp.MesosUpdate(_, taskStatus: MarathonTaskStatus.Running, now) if !hasStartedRunning =>
```
kolloch: unnecessary change

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -101,7 +104,9 @@ object Task {
       case TaskStateOp.MesosUpdate(_, taskStatus, now) =>
         updatedHealthOrState(status.mesosStatus, taskStatus.mesosStatus) match {
           case Some(newStatus) =>
-            val updatedTask = copy(status = status.copy(mesosStatus = Some(newStatus)))
+            val updatedTask = copy(
```
kolloch: unnecessary change

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -181,25 +186,27 @@ object Task {
...
-    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, networking))
+    override def launched: Option[Launched] = Some(Task.Launched(appVersion, status, hostPorts))
 
     private[this] def hasStartedRunning: Boolean = status.startedAt.isDefined
 
     //scalastyle:off cyclomatic.complexity method.length
     override def update(update: TaskStateOp): TaskStateChange = update match {
       // case 1: now running
-      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
+      case TaskStateOp.MesosUpdate(_, taskStatus: MarathonTaskStatus.Running, now) if !hasStartedRunning =>
```
kolloch: unnecessary change

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -215,7 +222,9 @@ object Task {
       // case 3: health or state updated
       case TaskStateOp.MesosUpdate(_, taskStatus, _) =>
         updatedHealthOrState(status.mesosStatus, taskStatus.mesosStatus).map { newStatus =>
-          val updatedTask = copy(status = status.copy(mesosStatus = Some(newStatus)))
+          val updatedTask = copy(
+            status = status.copy(mesosStatus = Some(newStatus))
```
kolloch: variation: no change necessary ;)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -122,18 +114,12 @@ object TaskSerializer {
     def setReservation(reservation: Task.Reservation): Unit = {
       builder.setReservation(ReservationSerializer.toProto(reservation))
     }
-    def setLaunched(appVersion: Timestamp, status: Task.Status, networking: Task.Networking): Unit = {
+    def setLaunched(appVersion: Timestamp, status: Task.Status, ports: Seq[Int]): Unit = {
```
kolloch: complete nit-pick: hostPorts would be even more consistent

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -122,18 +114,12 @@ object TaskSerializer {
...
       builder.setVersion(appVersion.toString)
       builder.setStagedAt(status.stagedAt.toDateTime.getMillis)
       status.startedAt.foreach(startedAt => builder.setStartedAt(startedAt.toDateTime.getMillis))
       status.mesosStatus.foreach(status => builder.setStatus(status))
-      networking match {
-        case Task.HostPorts(hostPorts) =>
-          builder.addAllPorts(hostPorts.view.map(Integer.valueOf(_)).asJava)
-        case Task.NetworkInfoList(networkInfoList) =>
-          builder.addAllNetworks(networkInfoList.asJava)
```
kolloch: I would love a comment in the marathon.proto which explains that the networks info in the protocol buffers is redundant and we got rid of that. Maybe even rename the field to "DEPRECATED_..." or comment it out.

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
```diff
@@ -361,12 +362,12 @@ object MarathonTestHelper {
...
   def taskLaunched: Task.Launched = {
     val now = Timestamp.now()
-    Task.Launched(now, status = Task.Status(now), networking = Task.NoNetworking)
+    Task.Launched(now, status = Task.Status(now), hostPorts = Seq.empty)
   }
 
   def taskLaunchedOp(taskId: Task.Id): TaskStateOp.LaunchOnReservation = {
     val now = Timestamp.now()
-    TaskStateOp.LaunchOnReservation(taskId, now, Task.Status(now), Task.NoNetworking)
+    TaskStateOp.LaunchOnReservation(taskId, now, Task.Status(now), Seq.empty)
```
kolloch: nit-pick, named argument would be nice

src/test/scala/mesosphere/marathon/MarathonTestHelper.scala
```diff
@@ -406,7 +407,7 @@ object MarathonTestHelper {
         startedAt = None,
         mesosStatus = Some(statusForState(taskId, Mesos.TaskState.TASK_STAGING))
       ),
-      Task.NoNetworking
+      Seq.empty
```
kolloch: named argument

src/test/scala/mesosphere/marathon/core/appinfo/TaskCountsTest.scala
```diff
@@ -182,7 +184,7 @@ class Fixture {
       startedAt = None,
       mesosStatus = None
     ),
-    Task.NoNetworking
+    Seq.empty
```
kolloch: as you have realized by now, I really like named arguments in these cases (Seq.empty doesn't give you a hint what the parameter is).

src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -256,9 +256,12 @@ class AppDeployIntegrationTest
...
 
-    val tasks = marathon.tasks(appId)
-    tasks.code should be(200)
-    tasks.value should have size 2
+    val tasksResult: RestResult[List[ITEnrichedTask]] = marathon.tasks(appId)
+    tasksResult.code should be(200)
+
+    val tasks = tasksResult.value
+    tasks should have size 2
+    tasks.foreach(_.ipAddresses.get should not be empty)
```
kolloch: :+1:

kolloch:
Mostly good to go!

Only major item: Clarify the deliberate non-usage of "networks" in the protocol buffers for future generations of Marathon developers ;)

gkleiman:
@kolloch added an extra commit with the required changes, looking forward to the next round of comments =)

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -123,9 +123,10 @@ trait Formats
         "startedAt" -> launched.status.startedAt,
         "stagedAt" -> launched.status.stagedAt,
         "ports" -> launched.hostPorts,
-        "version" -> launched.appVersion,
-        "ipAddresses" -> launched.ipAddresses
-      )
+        "version" -> launched.appVersion
+      ) ++ launched.ipAddresses.map { ipAddresses =>
+          Json.obj("ipAddresses" -> ipAddresses)
+        }.getOrElse(Json.obj())
```
kolloch: I think that you do not have to make it that complicated. `"ipAddresses" -> launched.ipAddresses` should suffice.

```scala
scala> case class WithOption(test: Option[String])
defined class WithOption

scala> implicit val format = play.api.libs.json.Json.writes[WithOption]
format: play.api.libs.json.OWrites[WithOption] = play.api.libs.json.OWrites$$anon$2@47fc5ae4

scala> play.api.libs.json.Json.toJson(WithOption(None))
res1: play.api.libs.json.JsValue = {}

scala> play.api.libs.json.Json.toJson(WithOption(Some("test")))
res2: play.api.libs.json.JsValue = {"test":"test"}
```

gkleiman: The output of this complicated block and the simpler version are not the same.

The simple version can result in `"ipAddresses": null` if there's no ip address. The more complicated version will skip the property in that case.

I'll update the PR with an intermediate version using foldLeft.

src/main/scala/mesosphere/marathon/health/HealthCheckWorkerActor.scala
```diff
@@ -43,35 +43,42 @@ class HealthCheckWorkerActor extends Actor with ActorLogging {
...
-        case HTTP  => http(app, task, launched, check, port)
-        case TCP   => tcp(app, task, launched, check, port)
-        case HTTPS => https(app, task, launched, check, port)
-        case COMMAND =>
-          Future.failed {
-            val message = s"COMMAND health checks can only be performed " +
-              "by the Mesos executor."
-            log.warning(message)
-            new UnsupportedOperationException(message)
+    task.effectiveIpAddress(app) match {
```
kolloch: This double nesting is not so nice. I would be nicer to kind of chain the stuff without nesting -- e.g. with `Try`.

kolloch:
Looks great, let's get that shipped. Squash and I'll merge.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
kolloch: That sucks but it is not so important. Just ignore my comment then ;)

gkleiman:
Squashed, this includes the following change:

```diff
w/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
index bd8b2da..312c6d8 100644
--- c/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ w/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -119,14 +119,16 @@ trait Formats
     )

     val launched = task.launched.map { launched =>
-      base ++ Json.obj (
-        "startedAt" -> launched.status.startedAt,
-        "stagedAt" -> launched.status.stagedAt,
-        "ports" -> launched.hostPorts,
-        "version" -> launched.appVersion
-      ) ++ launched.ipAddresses.map { ipAddresses =>
-          Json.obj("ipAddresses" -> ipAddresses)
-        }.getOrElse(Json.obj())
+      launched.ipAddresses.foldLeft(
+        base ++ Json.obj (
+          "startedAt" -> launched.status.startedAt,
+          "stagedAt" -> launched.status.stagedAt,
+          "ports" -> launched.hostPorts,
+          "version" -> launched.appVersion
+        )
+      ){
+        case (launchedJs, ipAddresses) => launchedJs ++ Json.obj("ipAddresses" -> ipAddresses)
+      }
     }.getOrElse(base)

     val reservation = task.reservationWithVolumes.map { reservation =>
```

gkleiman:
@kolloch confirmed it works with Calico's `net-modules`, feel free to hit merge =).

kolloch:
```
       _~
    _~ )_)_~
    )_))_))_)
    _!__!__!_
    \______t/
  ~~~~~~~~~~~~~
```

PR #3605, opened 2016-03-29
remove Defaults.defaultSettings mix in

timcharper:
This isn't necessary and leads to an invalid configuration scenario which prevents ensime-sbt from working.

For more information, see https://github.com/ensime/ensime-sbt/issues/145


aquamatthias:
Thanks!

PR #3607, opened 2016-03-29
Documentation for External volumes

sascala:
@jdef Here's a draft of documentation for external volumes. I've been working with @branden, so I assigned this to him. But I'd also like your comments/corrections.

### DIFF
--- a/docs/_layouts/docs.html
+++ b/docs/_layouts/docs.html
@@ -155,6 +155,11 @@ layout: default
         </a>
       </li>
       <li>
+        <a href="{{ site.baseurl }}/docs/external-volumes.html">
+          Stateful Applications Using External Volumes
+        </a>
+      </li>
+      <li>
         <a href="{{ site.baseurl }}/docs/task-environment-vars.html">
           Task Environment Variables
         </a>
--- a/docs/_layouts/docs.html
+++ b/docs/_layouts/docs.html
@@ -156,7 +156,7 @@ layout: default
       </li>
       <li>
         <a href="{{ site.baseurl }}/docs/external-volumes.html">
-          Stateful Applications Using External Volumes
+          Stateful Applications Using External Persistent Volumes
         </a>
       </li>
       <li>
--- a/docs/docs/external-volumes
+++ b/docs/docs/external-volumes
@@ -11,7 +11,7 @@ title: Stateful Applications Using External Persistent Volumes
 
 Marathon applications normally lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can use an external storage service, such as Amazon's Elastic Block Store (EBS), to create a persistent volume that follows your application instance.
 
-Using an external storage service allows your apps to be more fault-tolerant. If a host fails, Marathon reschedules it on another host, along with its associated data, without user intervention.
+Using an external storage service allows your apps to be more fault-tolerant. If a host fails, Marathon reschedules your app on another host, along with its associated data, without user intervention.
 
 # Specifying an External Volume
 
@@ -68,7 +68,7 @@ In the app definition above:
 
 - The `external.driver["dvdi/driver"]` option specifies which Docker volume driver to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-Ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
 
-- You can specify additional options with `container.volumes[x].external.options[optionName]`.
+- You can specify additional options with `container.volumes[x].external.options[optionName]`. The dvdi provider for Mesos containers uses `dvdcli`, which offers the options [documented here[(https://github.com/emccode/dvdcli#extra-options). The availability of any given option depends on your volume driver, however.
 
 - `name` is the name by which your volume driver looks up your volume. When your task is staged on an agent, the volume driver queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used. **Note:** Implicit volume creation only works when using volumes with a Mesos container.
 
--- a/docs/docs/external-volumes
+++ b/docs/docs/external-volumes
@@ -1,6 +1,6 @@
---------------------------------------------------------------
+---
 title: Stateful Applications Using External Persistent Volumes
---------------------------------------------------------------
+---
 
 # Stateful Applications Using External Persistent Volumes
 
@@ -39,17 +39,17 @@ You specify an external volume in the app definition of your Marathon app. [Lear
   "instances": 1,
   "cpus": 0.1,
   "mem": 32,
-  "cmd": "while [ true ] ; do touch /tmp/test-rexray-volume/hello ; sleep 5 ; done",
+  "cmd": "/usr/bin/tail -f /dev/null",
   "container": {
     "type": "MESOS",
     "volumes": [
       {
         "containerPath": "/tmp/test-rexray-volume",
-        "persistent": {
+        "external": {
           "size": 100,
-          "name": "jdef-test-vol",
-          "provider": "external",
-          "options": { "external/driver": "rexray" }
+          "name": "my-test-vol",
+          "provider": "dvdi",
+          "options": { "dvdi/driver": "rexray" }
           },
         "mode": "RW"
       }
@@ -62,15 +62,21 @@ You specify an external volume in the app definition of your Marathon app. [Lear
 }
 ```
 
-In the `volumes` section of app definition above:
+In the app definition above:
 
-- `containerPath` specifies the root path inside the volume for storing data. This path is configurable, but defaults to `data/`. See [the REX-Ray documentation on root paths](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#volume-root-path-031) for more information.
+- `containerPath` specifies where the volume is mounted inside the container. This path is configurable, but defaults to `data/`. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
 
-- The `external/driver` options specifies which Docker volume driver we want to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
+- The `external.driver["dvdi/driver"]` option specifies which Docker volume driver to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-Ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
 
-- `name` is the name by which your Docker volume driver looks up your volume. When your task is staged on an agent, the volume driver queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used.
+- You can specify additional options with `container.volumes[x].external.options[optionName]`.
 
-- Volumes will always be available under `$REXRAY_HOME/var/lib/rexray/volumes/$DVDI_VOLUME_NAME`. **Note:** This applies only to non-Docker apps; Docker apps can specify the mount point inside the app definition. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
+- `name` is the name by which your volume driver looks up your volume. When your task is staged on an agent, the volume driver queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used. **Note:** Implicit volume creation only works when using volumes with a Mesos container.
+
+- Create multiple volumes by adding additional items in the `container.volumes` array.
+
+- Volume parameters cannot be changed after you create the application.
+
+- **Note:** Marathon will not launch apps with external volumes if  `upgradeStrategy.minimusHealthCapacity` is less than 0.5, or if `upgradeStrategy.maximumOverCapacity` does not equal 0.
 
 ### Using a Docker Container
 
@@ -93,11 +99,11 @@ Below is a sample app definition that uses a Docker container and specifies an e
     "volumes": [
       {
         "containerPath": "/data/test-rexray-volume",
-        "persistent": {
+        "external": {
           "size": 100,
-          "name": "test",
-          "provider": "external",
-          "options": { "external/driver": "rexray" } 
+          "name": "my-test-vol",
+          "provider": "dvdi",
+          "options": { "dvdi/driver": "rexray" }
         },
         "mode": "RW"
       }
@@ -114,9 +120,9 @@ For more information, refer to the [REX-Ray documentation](https://rexray.readth
 
 ## Potential Pitfalls
 
-- Your storage provider may have particular limitations. For instance, on Amazon's EBS, you can only assign one task per volume.
+- You can only assign one task per volume. Your storage provider may other particular limitations.
 
-- The volumes you create are not automatically cleaned up. If you you delete your cluster, go to your storage provider and delete the volumes you no longer need. Find them by searching by `DVDI_VOLUME_NAME`.
+- The volumes you create are not automatically cleaned up. If you you delete your cluster, go to your storage provider and delete the volumes you no longer need. If you using EBS, find them by searching by the `container.volume.external.name` that you set in your Marathon app definition. This name corresponds to an EBS volume `Name` tag.
 
 - Volumes are not isolated from other clusters. Choose a unique volume name to avoid conflicts.
 
--- a/docs/docs/external-volumes
+++ b/docs/docs/external-volumes
@@ -36,7 +36,7 @@ You specify an external volume in the app definition of your Marathon app. [Lear
 ```json
 { 
   "id": "hello",
-     "cmd": "while [ true ] ; do touch/var/lib/rexray/volumes/test12345/data/hello ; sleep 5 ; done",   
+     "cmd": "while [ true ] ; do touch /var/lib/rexray/volumes/test12345/data/hello ; sleep 5 ; done",   
      "mem": 32,   
      "cpus": 0.1,   
      "instances": 1,
--- a/docs/docs/external-volumes
+++ b/docs/docs/external-volumes
@@ -1,0 +1,81 @@
+--------------------------------------------------------------
+title: Stateful Applications Using Persistent External Volumes
+--------------------------------------------------------------
+
+# Stateful Applications Using Persistent External Volumes
+
+<div class="alert alert-danger" role="alert">
+  <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span> Adapted in Marathon Version 1.0 <br/>
+  The Persistent Storage functionality is considered beta, so use this feature at your own risk. We might add, change, or delete any functionality described in this document.
+</div>
+
+Marathon applications lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can use an external storage service, such as Amazon's Elastic Block Store (EBS), to create a persistent volume that follows your application instance.
+
+Using an external storage service allows your apps to be more fault-tolerant. If a host fails, Marathon reschedules it on another host, along with its associated data, without user intervention.
+
+# Specifying an External Volume
+
+When you install DCOS, add the following to your `genconf.yml` file:
+
+- `rexray_config_method: file`
+
+- `rexray_config_filename: /path/to/rexray.yaml`
+
+<!-- [Learn more](URL). link to installer docs that mention this when they are live-->
+
+## Scaling your App
+
+Apps that use external volumes should only be scaled to a single instance.
+
+If you scale your app down to 0 instances, the volume is unattached from the agent where it was mounted, but it is not deleted. If you scale your app up again, the data that had been associated with it will still be available.
+
+## Create an Application with External Volumes
+
+You specify an external volume in the app definition of your Marathon app. [Learn more about Marathon application definitions](application-basics.html).
+
+```json
+{ 
+  "id": "hello",
+     "cmd": "while [ true ] ; do touch/var/lib/rexray/volumes/test12345/data/hello ; sleep 5 ; done",   
+     "mem": 32,   
+     "cpus": 0.1,   
+     "instances": 1,
+     "env": {
+          "DVDI_VOLUME_DRIVER": "rexray",
+          "DVDI_VOLUME_NAME": "test12345",
+          "DVDI_VOLUME_OPTS":"size=1,volumetype=standard,newfstype=ext4,overwritefs=false"
+     }
+}
+```
+
+In the app definition above, the environment variables (`env`) specify a 1GB volume that will be formatted ext4 once it is created.
+
+- Volumes will always be available under `/var/lib/rexray/volumes/$DVDI_VOLUME_NAME/`. **Note:** This applies only to non-Docker apps; Docker apps can specify the mount point inside the app definition.
+
+- The root path inside the volume for storing data is configurable, but defaults to `data/`. See [the REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#volume-root-path-031) for more information.
+
+- `DVDI_VOLUME_DRIVER` tells `DVDCLI` which Docker volume driver we want to use for storage. The only driver we currently provide is REX-Ray, so this value should always be `rexray`. [Learn more about REX-ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
+
+- `DVDI_VOLUME_NAME` is the name by which REX-Ray looks up your volume. When your task is staged on an agent, REX-Ray queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used.
+
+- `DVDI_VOLUME_OPTS` is the set of volume options passed to DVDCLI. [See the full list of options](https://github.com/emccode/dvdcli/blob/3c3dcbe1490c2c4a150555978f2b4915dafaddf2/README.md#extra-options).
+
+You can define multiple volumes per app by repeating the environment variables with 0, 1 ,2, etc. appended to them. For example, `DVDI_VOLUME_NAME0`.
+
+**Note:** Most of the configuration of your external volume is cannot be changed after you create it. The exceptions are `overwriteFs` and `newFsType`. If `overwriteFs` is true, REX-Ray will reformat the volume every time it is mounted on an agent with the filesystem specified by `newFsType`. If `overwriteFs` is false, howeverm REX-Ray will only format the volume when it is created.
+
+### Docker apps
+
+Refer to the [REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/#docker-containerizer-with-marathon) for an example of an app definition of a Docker app that uses external volumes.
+
+## Potential Pitfalls
+
+- Your storage provider may have particular limitations. For instance, on Amazon's EBS, you can only assign one task per volume.
+
+- The volumes you create are not automatically cleaned up. If you you delete your cluster, go to your storage provider and delete the volumes you no longer need. Find them by searching by `DVDI_VOLUME_NAME`.
+
+- Volumes are not isolated from other clusters. Choose a unique `DVDI_VOLUME_NAME` to avoid conflicts.
+
+- Docker apps do not support external volumes on DCOS installations running Docker older than 1.8. Currently, this means that DCOS Community Edition users cannot create Docker apps with external volumes.
+
+For more information, see the [Apache Mesos documentation on persistent volumes](http://mesos.apache.org/documentation/latest/persistent-volume/).
--- a/docs/docs/external-volumes
+++ b/docs/docs/external-volumes
@@ -1,21 +1,21 @@
 --------------------------------------------------------------
-title: Stateful Applications Using Persistent External Volumes
+title: Stateful Applications Using External Persistent Volumes
 --------------------------------------------------------------
 
-# Stateful Applications Using Persistent External Volumes
+# Stateful Applications Using External Persistent Volumes
 
 <div class="alert alert-danger" role="alert">
   <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span> Adapted in Marathon Version 1.0 <br/>
   The Persistent Storage functionality is considered beta, so use this feature at your own risk. We might add, change, or delete any functionality described in this document.
 </div>
 
-Marathon applications lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can use an external storage service, such as Amazon's Elastic Block Store (EBS), to create a persistent volume that follows your application instance.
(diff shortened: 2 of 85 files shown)


docs/docs/external-volumes
```diff
@@ -0,0 +1,81 @@
...
+
+```json
+{ 
+  "id": "hello",
+     "cmd": "while [ true ] ; do touch/var/lib/rexray/volumes/test12345/data/hello ; sleep 5 ; done",   
+     "mem": 32,   
+     "cpus": 0.1,   
+     "instances": 1,
+     "env": {
+          "DVDI_VOLUME_DRIVER": "rexray",
```
jdef: for now this will work. however, specifying via envvar will be deprecated with a PR that I'm working on to enhance the persistent vol API in Marathon.

docs/docs/external-volumes
```diff
@@ -0,0 +1,81 @@
...
+
+- The root path inside the volume for storing data is configurable, but defaults to `data/`. See [the REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#volume-root-path-031) for more information.
+
+- `DVDI_VOLUME_DRIVER` tells `DVDCLI` which Docker volume driver we want to use for storage. The only driver we currently provide is REX-Ray, so this value should always be `rexray`. [Learn more about REX-ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
+
+- `DVDI_VOLUME_NAME` is the name by which REX-Ray looks up your volume. When your task is staged on an agent, REX-Ray queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used.
+
+- `DVDI_VOLUME_OPTS` is the set of volume options passed to DVDCLI. [See the full list of options](https://github.com/emccode/dvdcli/blob/3c3dcbe1490c2c4a150555978f2b4915dafaddf2/README.md#extra-options).
+
+You can define multiple volumes per app by repeating the environment variables with 0, 1 ,2, etc. appended to them. For example, `DVDI_VOLUME_NAME0`.
```
jdef: I didn't think that `0` was a valid suffix. If you specify a single volume you don't number it - but then for additional volumes, you append 1, 2, 3, and so on.

docs/docs/external-volumes
```diff
@@ -0,0 +1,81 @@
...
+  The Persistent Storage functionality is considered beta, so use this feature at your own risk. We might add, change, or delete any functionality described in this document.
+</div>
+
+Marathon applications lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can use an external storage service, such as Amazon's Elastic Block Store (EBS), to create a persistent volume that follows your application instance.
+
+Using an external storage service allows your apps to be more fault-tolerant. If a host fails, Marathon reschedules it on another host, along with its associated data, without user intervention.
+
+# Specifying an External Volume
+
+When you install DCOS, add the following to your `genconf.yml` file:
```
jdef: Are there other places in the marathon docs where we reference DCOS? If not, then this seems inconsistent

jdef:
left some comments - thanks for putting this together

docs/docs/external-volumes
sascala: Thanks for the heads up! If the enhancement will be ready for 1.7, can you tell me what the new process should be?

docs/docs/external-volumes
branden: Any single digit is an acceptable suffix. No suffix is equivalent to a suffix of `0`. https://github.com/emccode/mesos-module-dvdi/blob/9f4ab1cf187fb7b8ba6b0ec5b5518860cdb8ee44/isolator/isolator/docker_volume_driver_isolator.cpp#L491

docs/docs/external-volumes
sascala: When we merge the Marathon docs with the rest at docs.mesosphere.com, that won't be an issue. For now, I'll check with @air. And, if I understand right, you need to be using DCOS to use external volumes?

docs/docs/external-volumes
```diff
@@ -0,0 +1,81 @@
...
+If you scale your app down to 0 instances, the volume is unattached from the agent where it was mounted, but it is not deleted. If you scale your app up again, the data that had been associated with it will still be available.
+
+## Create an Application with External Volumes
+
+You specify an external volume in the app definition of your Marathon app. [Learn more about Marathon application definitions](application-basics.html).
+
+```json
+{ 
+  "id": "hello",
+     "cmd": "while [ true ] ; do touch/var/lib/rexray/volumes/test12345/data/hello ; sleep 5 ; done",
```
branden: There should be a space after `touch` here.

docs/docs/external-volumes
```diff
@@ -0,0 +1,81 @@
...
+          "DVDI_VOLUME_DRIVER": "rexray",
+          "DVDI_VOLUME_NAME": "test12345",
+          "DVDI_VOLUME_OPTS":"size=1,volumetype=standard,newfstype=ext4,overwritefs=false"
+     }
+}
+```
+
+In the app definition above, the environment variables (`env`) specify a 1GB volume that will be formatted ext4 once it is created.
+
+- Volumes will always be available under `/var/lib/rexray/volumes/$DVDI_VOLUME_NAME/`. **Note:** This applies only to non-Docker apps; Docker apps can specify the mount point inside the app definition.
```
branden: I gave you some DCOS-specific information here, my mistake. The volumes will always be available at `$REXRAY_HOME/var/lib/rexray/volumes/$DVDI_VOLUME_NAME` (DCOS sets `REXRAY_HOME=/`). See https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories

docs/docs/external-volumes
```diff
@@ -0,0 +1,81 @@
...
+
+- The root path inside the volume for storing data is configurable, but defaults to `data/`. See [the REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#volume-root-path-031) for more information.
+
+- `DVDI_VOLUME_DRIVER` tells `DVDCLI` which Docker volume driver we want to use for storage. The only driver we currently provide is REX-Ray, so this value should always be `rexray`. [Learn more about REX-ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
```
branden: This second sentence is DCOS-specific, too. Someone running Marathon may install other Docker Volume Drivers that DVDCLI could use.

docs/docs/external-volumes
branden: The pieces of software that together provide external volume support (DVDI, DVDCLI, REX-Ray) are all open source, so a user running their own Marathon/Mesos could install them and then start using external volumes, no DCOS required.

docs/docs/external-volumes
jdef: Thanks for clarifying!

docs/docs/external-volumes
sascala: What would a non-DCOS user have to do to use external volumes, apart from specify them in the Marathon app definition?

branden: They'd have to install DVDCLI, install the DVDI Mesos isolator module, configure Mesos to use that isolator module, configure the isolator with the location of DVDCLI and Mesos' work dir, and install REX-Ray and run its daemon service.

docs/docs/external-volumes
```diff
@@ -15,14 +15,12 @@ Using an external storage service allows your apps to be more fault-tolerant. If
 
 # Specifying an External Volume
 
-When you install DCOS, add the following to your `genconf.yml` file:
+If you are running Marathon on DCOS, add the following to the `genconf.yml` file you use during DCOS installation. [Learn more](https://docs.mesosphere.com/concepts/installing/installing-enterprise-edition/configuration-parameters/).
```
branden: This should be `genconf/config.yaml`.

docs/_layouts/docs.html
```diff
@@ -155,6 +155,11 @@
         </a>
       </li>
       <li>
+        <a href="{{ site.baseurl }}/docs/external-volumes.html">
+          Stateful Applications Using External Volumes
```
sttts: Are those called "external" or "persistent" volumes? The json schema in https://github.com/mesosphere/marathon/pull/3525 suggests the second. IMO "persistent" is the better term.

docs/docs/external-volumes
```diff
@@ -0,0 +1,79 @@
...
+--------------------------------------------------------------
+
+# Stateful Applications Using Persistent External Volumes
+
+<div class="alert alert-danger" role="alert">
+  <span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true"></span> Adapted in Marathon Version 1.0 <br/>
+  The Persistent Storage functionality is considered beta, so use this feature at your own risk. We might add, change, or delete any functionality described in this document.
+</div>
+
+Marathon applications lose their state when they terminate and are relaunched. In some contexts, for instance, if your application uses MySQL, you’ll want your application to preserve its state. You can use an external storage service, such as Amazon's Elastic Block Store (EBS), to create a persistent volume that follows your application instance.
```
sttts: ... normally ...

docs/docs/external-volumes
```diff
@@ -0,0 +1,79 @@
...
+
+Using an external storage service allows your apps to be more fault-tolerant. If a host fails, Marathon reschedules it on another host, along with its associated data, without user intervention.
+
+# Specifying an External Volume
+
+If you are running Marathon on DCOS, add the following to the `genconf.yml` file you use during DCOS installation. [Learn more](https://docs.mesosphere.com/concepts/installing/installing-enterprise-edition/configuration-parameters/).
```
sttts: A bit strange to find DCOS documentation in the public Marathon repo.

docs/docs/external-volumes
```diff
@@ -0,0 +1,79 @@
...
+          "DVDI_VOLUME_DRIVER": "rexray",
+          "DVDI_VOLUME_NAME": "test12345",
+          "DVDI_VOLUME_OPTS":"size=1,volumetype=standard,newfstype=ext4,overwritefs=false"
+     }
+}
+```
+
+In the app definition above, the environment variables (`env`) specify a 1GB volume that will be formatted ext4 once it is created.
+
+- Volumes will always be available under `$REXRAY_HOME/var/lib/rexray/volumes/$DVDI_VOLUME_NAME`. **Note:** This applies only to non-Docker apps; Docker apps can specify the mount point inside the app definition. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
```
sttts: All this text is very rexray oriented. Rexray is only one Docker Volume Driver out of many.

docs/_layouts/docs.html
sttts: In fact, "external persistent volumes" is the right term (in contrast to "local persistent volumes" described below).

Our json schema is a bit confusing IMO:

```json
volumes: { containerPath: "/data", persistent: {
  name: "v1",
  size: 10
}
```

is a _local persistent volume_.

```json
volumes: { containerPath: "/data", persistent: {
  name: "v1", 
  size: 10,
  providerName: "dvdi",
  options: {...}
}
```

is an _external persistent volume_. Wondering whether the latter shouldn't be
```json
volumes: { containerPath: "/data", persistent: {
  name: "v1", 
  size: 10,
  kind: "external",
  options: {
    provider: "dvdi"
    ...
  }
}
```

with `kind` defaulting to `local`.

sascala:
Thanks for your comments, @sttts. @jdef and @branden -- I've made the changes you've all suggested. I'll mark this as complete unless you feel differently.

docs/docs/external-volumes
```diff
@@ -0,0 +1,79 @@
...
+
+### Docker apps
+
+Refer to the [REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/#docker-containerizer-with-marathon) for an example of an app definition of a Docker app that uses external volumes.
+
+## Potential Pitfalls
+
+- Your storage provider may have particular limitations. For instance, on Amazon's EBS, you can only assign one task per volume.
+
+- The volumes you create are not automatically cleaned up. If you you delete your cluster, go to your storage provider and delete the volumes you no longer need. Find them by searching by `DVDI_VOLUME_NAME`.
```
jdef: What, exactly, should the user search? the agent logs?

docs/docs/external-volumes
jdef: there's also ways to mount container volumes under /tmp

docs/docs/external-volumes
branden: The user should search whatever interface/UI the storage provider offers, which is why it's hard to be exact here. Using EBS as an example, you'd go to the AWS EC2 console, click on Volumes, and search by the `DVDI_VOLUME_NAME` that was set in the Marathon app, which will be the value of the `Name` tag on the volume.

docs/docs/external-volumes
jdef: MESOS container example:
```json
{
  "id": "hello",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "cmd": "while [ true ] ; do touch /tmp/test-rexray-volume/hello ; sleep 5 ; done",
  "container": {
    "type": "MESOS",
    "volumes": [
      {
        "containerPath": "/tmp/test-rexray-volume",
        "persistent": {
          "size": 100,
          "name": "jdef-test-vol",
          "provider": "external",
          "options": { "external/driverName": "rexray" }
          },
        "mode": "RW"
      }
    ]
  },
  "upgradeStrategy": {
    "minimumHealthCapacity": 0,
    "maximumOverCapacity": 0
  }
}
```

jdef: DOCKER container example (xref #3525)
```json
{
  "id": "/test-docker",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "cmd": "/usr/bin/tail -f /dev/null",
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "alpine:3.1",
      "network": "HOST",
      "forcePullImage": true
    },
    "volumes": [
      {
        "containerPath": "/data/test-rexray-volume",
        "persistent": {
          "size": 100,
          "name": "test",
          "provider": "external",
          "options": { "external/driver": "rexray" } 
        },
        "mode": "RW"
      }
    ]
  },
  "upgradeStrategy": {
    "minimumHealthCapacity": 0,
    "maximumOverCapacity": 0
  }
}
```

docs/docs/external-volumes
branden: @jdef can you clarify? AFAIK the only way to get REX-Ray to mount volumes under `/tmp` is to set `REXRAY_HOME=/tmp`.

jdef: The MESOS container example I pasted in a comment shows how to do this

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+  "upgradeStrategy": {
+    "minimumHealthCapacity": 0,
+    "maximumOverCapacity": 0
+  }
+}
+```
+
+In the `volumes` section of app definition above:
+
+- `containerPath` specifies the root path inside the volume for storing data. This path is configurable, but defaults to `data/`. See [the REX-Ray documentation on root paths](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#volume-root-path-031) for more information.
```
sascala: @jdef Is this what Brandon had called the "root path"?

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
```
sascala: Is this what Brandon had called the "root path"?

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+
+- The `external/driver` options specifies which Docker volume driver we want to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
+
+- `name` is the name by which your Docker volume driver looks up your volume. When your task is staged on an agent, the volume driver queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used.
```
sascala: I reworded this to make it less REX-Ray-specific, though not sure if this is accurate for every volume driver.

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+
+- Volumes will always be available under `$REXRAY_HOME/var/lib/rexray/volumes/$DVDI_VOLUME_NAME`. **Note:** This applies only to non-Docker apps; Docker apps can specify the mount point inside the app definition. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
```
sascala: Is this still true?

sascala: Can we still specify things that used to be specified in the environment variable with `DVDI_VOLUME_OPTS`?

sascala: How do we define multiple volumes per app?

sascala: What can you change after volume creation, and how do you do it?

docs/docs/external-volumes
sascala: @jdef, should a user still search by `DVDI_VOLUME_NAME`?

jdef: Once #3675 is merged, using EBS as the example, the user would search by `container.volume.external.name` that was set in the Marathon app, which corresponds to an EBS volume `Name` tag

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+  "instances": 1,
+  "cpus": 0.1,
+  "mem": 32,
+  "cmd": "while [ true ] ; do touch /tmp/test-rexray-volume/hello ; sleep 5 ; done",
+  "container": {
+    "type": "MESOS",
+    "volumes": [
+      {
+        "containerPath": "/tmp/test-rexray-volume",
+        "persistent": {
```
jdef: we changed the API, sorry. new format is:
```json
{
  "id": "hello",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "cmd": "/usr/bin/tail -f /dev/null",
  "container": {
    "type": "MESOS",
    "volumes": [
      {
        "containerPath": "/tmp/test-rexray-volume",
        "external": {
          "size": 100,
          "name": "jdef-test-vol",
          "provider": "dvdi",
          "options": { "dvdi/driver": "rexray" }
          },
        "mode": "RW"
      }
    ]
  },
  "upgradeStrategy": {
    "minimumHealthCapacity": 0,
    "maximumOverCapacity": 0
  }
}
```

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+- `name` is the name by which your Docker volume driver looks up your volume. When your task is staged on an agent, the volume driver queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used.
+
+- Volumes will always be available under `$REXRAY_HOME/var/lib/rexray/volumes/$DVDI_VOLUME_NAME`. **Note:** This applies only to non-Docker apps; Docker apps can specify the mount point inside the app definition. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
+
+### Using a Docker Container
+
+Below is a sample app definition that uses a Docker container and specifies an external volume:
+
+```json
+{
```
jdef: similarly, the API change for DOCKER containers:
```json
{
  "id": "/test-docker",
  "instances": 1,
  "cpus": 0.1,
  "mem": 32,
  "cmd": "/usr/bin/tail -f /dev/null",
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "alpine:3.1",
      "network": "HOST",
      "forcePullImage": true
    },
    "volumes": [
      {
        "containerPath": "/data/test-rexray-volume",
        "external": {
          "size": 100,
          "name": "test",
          "provider": "dvdi",
          "options": { "dvdi/driver": "rexray" }
        },
        "mode": "RW"
      }
    ]
  },
  "upgradeStrategy": {
    "minimumHealthCapacity": 0,
    "maximumOverCapacity": 0
  }
}
```

docs/docs/external-volumes
jdef: "containerPath" where the volume is mounted inside either the DOCKER or MESOS container. for example, if you launch a DOCKER container, SSH to the owning slave node and then run `docker exec -ti $CONTAINER_ID cat /proc/mounts` you'd see the volume mounted to the value of `containerPath` inside the volume

docs/docs/external-volumes
jdef: implicit volume creation works when using volumes with a MESOS container.

when using volumes with a DOCKER container they should be pre-created (it's not possible to specify `volume.external.size` or most other `volume.external.options.[optionName]` values with DOCKER containers, with the exception of `optionName` == `dvdi/driver`)

docs/docs/external-volumes
jdef: 1. I've been able to create volumes under `$REXRAY_HOME/tmp`.
1. options can be specified via `container.volumes[x].external.options[optionName]`
1. if you want multiple volumes, declare them as items in the `container.volumes` array
1. I don't think we allow changing parameters of volumes for an app

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+          "size": 100,
+          "name": "jdef-test-vol",
+          "provider": "external",
+          "options": { "external/driver": "rexray" }
+          },
+        "mode": "RW"
+      }
+    ]
+  },
+  "upgradeStrategy": {
```
jdef: this piece cannot be understated. Marathon will fail to launch apps using external volumes if either `minimumHealthCapacity > 0.5` or `maximumOverCapacity` > 0

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+    "maximumOverCapacity": 0
+  }
+}
+```
+
+In the `volumes` section of app definition above:
+
+- `containerPath` specifies the root path inside the volume for storing data. This path is configurable, but defaults to `data/`. See [the REX-Ray documentation on root paths](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#volume-root-path-031) for more information.
+
+- The `external/driver` options specifies which Docker volume driver we want to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
```
jdef: `external.options["dvdi/driver"]`

docs/docs/external-volumes
```diff
@@ -0,0 +1,125 @@
...
+    "maximumOverCapacity": 0
+  }
+}
+```
+
+For more information, refer to the [REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/#docker-containerizer-with-marathon).
+
+## Potential Pitfalls
+
+- Your storage provider may have particular limitations. For instance, on Amazon's EBS, you can only assign one task per volume.
```
jdef: currently we're enforcing this for all external volume providers, not just DVDI using rexray

docs/docs/external-volumes
```diff
@@ -62,15 +62,21 @@ You specify an external volume in the app definition of your Marathon app. [Lear
...
 ```
 
-In the `volumes` section of app definition above:
+In the app definition above:
+
+- `containerPath` specifies where the volume is mounted inside the container. This path is configurable, but defaults to `data/`. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
+
+- The `external.driver["dvdi/driver"]` option specifies which Docker volume driver to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-Ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
+
+- You can specify additional options with `container.volumes[x].external.options[optionName]`.
```
sascala: @jdef One final question: how can the user learn what other options are available?

jdef: the "dvdi" provider, for MESOS containers, uses "dvdcli" which documents
the following (rexray-focused) options:
https://github.com/emccode/dvdcli#extra-options. Marathon provides some
validation for some of these. We've also enabled a general "pass-through"
so that an option named "dvdi/optionABC" is passed through to the dvdcli as
"optionABC".

the "dvdi" provider, for DOCKER containers, only supports the "dvdi/driver"
option.

options are generally intended to target specific provider implementations.
the "dvdi" implementation is somewhat complex in that it supports multiple
drivers below it (and those drivers have arbitrary options).

sascala: thanks, @jdef.

sascala:
@jdef and @branden - thanks for working with me on this. If everything looks OK and nothing is missing, I'll close this out and publish tomorrow.

docs/docs/external-volumes
jdef: we're aiming to provide a default upgrade strategy that's compat with external volumes if the user doesn't provide one. see #3699

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+  "upgradeStrategy": {
+    "minimumHealthCapacity": 0,
+    "maximumOverCapacity": 0
+  }
+}
+```
+
+In the app definition above:
+
+- `containerPath` specifies where the volume is mounted inside the container. This path is configurable, but defaults to `data/`. See [the REX-Ray documentation on data directories](https://rexray.readthedocs.org/en/v0.3.2/user-guide/config/#data-directories) for more information.
```
jdef: `containerPath` is actually a required parameter for ext vols in marathon. there is no default - the user must provide a location.

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+
+- The `external.driver["dvdi/driver"]` option specifies which Docker volume driver to use for storage. If you are running Marathon on DCOS, this value should always be `rexray`. [Learn more about REX-Ray](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/).
```
jdef: s/always/likely/

it's possible that the user has configured other docker driver backends besides rexray (though we haven't done much testing of any others)

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+
+- You can specify additional options with `container.volumes[x].external.options[optionName]`. The dvdi provider for Mesos containers uses `dvdcli`, which offers the options [documented here[(https://github.com/emccode/dvdcli#extra-options). The availability of any given option depends on your volume driver, however.
+
+- `name` is the name by which your volume driver looks up your volume. When your task is staged on an agent, the volume driver queries the storage service for a volume with this name. If one does not exist, it's created. Otherwise, the existing volume is re-used. **Note:** Implicit volume creation only works when using volumes with a Mesos container.
```
jdef: `volumes[x].external.size` is required for implicit creation

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+    "docker": {
+      "image": "alpine:3.1",
+      "network": "HOST",
+      "forcePullImage": true
+    },
+    "volumes": [
+      {
+        "containerPath": "/data/test-rexray-volume",
+        "external": {
+          "size": 100,
```
jdef: I goofed - `size` should not be defined for external vols in DOCKER containers

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+          "name": "my-test-vol",
+          "provider": "dvdi",
+          "options": { "dvdi/driver": "rexray" }
+        },
+        "mode": "RW"
+      }
+    ]
+  },
+  "upgradeStrategy": {
```
jdef: the rules are correct. whether we require that the user define this block is related to #3699

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+
+For more information, refer to the [REX-Ray documentation](https://rexray.readthedocs.org/en/v0.3.2/user-guide/schedulers/#docker-containerizer-with-marathon).
+
+## Potential Pitfalls
+
+- You can only assign one task per volume. Your storage provider may other particular limitations.
+
+- The volumes you create are not automatically cleaned up. If you you delete your cluster, go to your storage provider and delete the volumes you no longer need. If you using EBS, find them by searching by the `container.volume.external.name` that you set in your Marathon app definition. This name corresponds to an EBS volume `Name` tag.
```
jdef: s/volume/volumes/

docs/docs/external-volumes
```diff
@@ -0,0 +1,131 @@
...
+
+- Volumes are not isolated from other clusters. Choose a unique volume name to avoid conflicts.
```
jdef: we also noticed that, when using CCM, clusters can be spun up in different AWS availability zones. for example I could create a cluster and use some external vols -- and CCM picked AZ1 for this cluster. I then destroy the cluster. I then launch a new cluster. CCM picks AZ2. I try to use the external vol from my first cluster but I can't because my new cluster is in a different AZ.

This may be a cloud-provider-specific "gotcha" but it was something that bit me during testing. I'm sure I won't be the only one.

jdef: ... and CCM doesn't provide a way to pick an AZ for new clusters

jdef:
What's also intuitive (but perhaps not immediately obvious) is that
application launch time may increase for tasks that implicitly create new
external volumes. The amount of the increase depends on several factors
(including, but not limited to) size and type of the volume, as well as the
expediency of the storage provider with respect to volume operations (e.g.
EBS if using rexray on EC2 clusters).

What's less intuitive and much less obvious is that troubleshooting errors
is strikingly poor experience. There's some information in the slave/agent
log. But the gory details (e.g. rexray operation logs) are in the system
journal (this may vary per provider). If a task fails to launch due to a
problem relating to a **runtime** error related to external volume
configuration it's likely not immediately apparent in the UI what the cause
of the problem actually is. And it may not be apparent when reviewing the
slave logs (which are avaliable in the mesos UI -- not sure about our new
marathon UI?)

sascala:
Thanks for the troubleshooting comments, @jdef. I put your first paragraph in the troubleshooting section. Your second... I'm not sure what to tell a user: that it's hard and confusing? If we have some positive suggestions, I'd put those...

jdef:
Perhaps some kind of pointer to the agent and/or system logs if tasks using
external vols don't appear to be working?

sascala:
@jdef, @branden : if this document looks ok, could you please publish?

docs/docs/external-volumes
```diff
@@ -132,4 +132,6 @@ For more information, refer to the [REX-Ray documentation](https://rexray.readth
 - Launch time may increase for applications that create volumes implicitly. The amount of the increase depends on several factors
 (including, but not limited to) the size and type of the volume. Your storage provider's method of handling volumes can also influence launch time for implicitly created volumes.
 
+- If tasks using external volumes are not working, or not working the way you expect, consult the agent or system logs to troubleshoot. If you are using REX-Ray, you can also consult the system journal.
```
jdef: If you are using REX-Ray on DCOS, you can ...

jdef:
Also, there's something wrong with this PR -- the number of changed files is way too high

sascala:
Oh dear. My attempt to rebase... did not do what it was supposed to.

jdef:
closed in favor of #3715

PR #3615, opened 2016-03-30
Experiment to simplify ContextUpdate

sttts:
- rename ContextUpdate -> Decorator
- make Decorator.decorated functional, i.e. the identity if volumes are incompatible.

  If we want to detect when volumes are applied by any provider, we should insert a unprocessedVolumes set into the pipeline. When it's empty, all volumes have been applied to the context.


jdef:
incorporated into #3525

PR #1267, opened 2015-03-03
allow for disabling output to syslog with USE_SYSLOG=0

naelyn:
This makes marathon wrapper script more compatible with supervisor and systemd, or anyone else who doesn't want all of their application logs sent to syslog.

Basically the same change as: https://github.com/mesosphere/mesos-deb-packaging/pull/35

PR #3620, opened 2016-03-31
need to disable output to syslog via environment variable, anything help.

pingz:
Just as same as pull request #1267 , and I added a prefix `MARATHON_', and I also will submit an issue for that https://github.com/mesosphere/marathon/pull/1267/files

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -114,10 +114,11 @@ function run_jar {
 }
 
 function logged {
-  local token="$1[$$]" ; shift
-  exec 1> >(exec logger -p user.info   -t "$token")
-  exec 2> >(exec logger -p user.notice -t "$token")
-  "$@"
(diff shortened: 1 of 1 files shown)


aquamatthias:
The PR #1267 was closed with this comment:

>
Thanks for your contribution, but this functionality has already been added to the script and will be part of 0.8.1. If you start the script with --no-logger it will start without the logger.

This functionality is still available and can be used. 
What is missing?

pingz:
@aquamatthias I used MARATHON_NO_LOGGER before, I installed the office mesosphere rpm(repos.mesosphere.com) in an centos based docker image, and using some thing like `docker run -d --net=host MARATHON_NO_LOGGER=1 {{ registry_endpoint }}/marathon:{{ marathon_version }}` . But weired thing is that marathon keeping failing to start with `--no-logger`, so I just test another approach. I will giving some time to find out why that didn't work.

pingz:
Close this, please see my comment at issue #3621.

PR #3632, opened 2016-04-01
Fixes #3624 | Respect constraints for same version

### LINKED ISSUE #3624: Constraints are not working for updating
Consider following example. I have one application launched with 2 tasks. I run it with following constraint  `rackid:LIKE:rack-1`. Task was deployed correctly. Then I updated config to spread this two tasks across different racks `rackid:GROUP_BY:2`. But my two tasks are still on same rack.

I think constraints should be matched for specific app version.

janisz:
This PR changes behavior of deploying/scaling application with constraints described in #3624. Currently constraints are matched for all instances of application. This PR introduce matching constraints only with tasks ~~in same version~~ deployed after last configuration change. ~~What does it mean? If application has constraints and it's configuration has changed (e.g. constraint was changed), after deployment we could end up with instances that does not match desired constraints. On the other hand after scaling application we (usually) have instances that match constraints. After my changes, deployments always end up with desired distribution of instances although after scaling, newest instances could not match that constraints, because constraints will be checked across instances with same version.~~

### DIFF
--- a/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
+++ b/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
@@ -116,7 +116,7 @@ object ResourceMatcher {
     def portsMatchOpt: Option[PortsMatch] = new PortsMatcher(app, offer, selector).portsMatch
 
     def meetsAllConstraints: Boolean = {
-      lazy val tasks = runningTasks
+      lazy val tasks = runningTasks.filter(_.launched.exists(_.appVersion >= app.versionInfo.lastConfigChangeVersion))
       val badConstraints = app.constraints.filterNot { constraint =>
         Constraints.meetsConstraint(tasks, offer, constraint)
       }
--- a/src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
+++ b/src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
@@ -1,14 +1,20 @@
 package mesosphere.mesos
 
-import mesosphere.marathon.Protos.Constraint
+import mesosphere.marathon.MarathonTestHelper.Implicits._
+import mesosphere.marathon.Protos.{ Constraint }
 import mesosphere.marathon.Protos.Constraint.Operator
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.state.AppDefinition.VersionInfo.{ FullVersionInfo, OnlyVersion }
 import mesosphere.marathon.state.PathId._
-import mesosphere.marathon.state.{ AppDefinition, PortDefinitions }
+import mesosphere.marathon.state.{ Timestamp, AppDefinition, PortDefinitions }
 import mesosphere.marathon.tasks.PortsMatcher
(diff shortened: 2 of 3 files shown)


src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -116,7 +116,7 @@ object ResourceMatcher {
     def portsMatchOpt: Option[PortsMatch] = new PortsMatcher(app, offer, selector).portsMatch
 
     def meetsAllConstraints: Boolean = {
-      lazy val tasks = runningTasks
+      lazy val tasks = runningTasks.filter(t => t.launched.map(_.appVersion == app.version).getOrElse(true))
```
aquamatthias: Every scaling change will change the appVersion, so a strict equals is wrong.
Also if the task is not launched, the default should be `false` not `true`.
Perhaps something like: `runningTasks.filter(_.launched.exists(_.appVersion >= app.versionInfo.lastConfigChangeVersion))` is better.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -127,8 +127,8 @@ trait Formats
           "version" -> launched.appVersion
         )
       ){
-        case (launchedJs, ipAddresses) => launchedJs ++ Json.obj("ipAddresses" -> ipAddresses)
-      }
+          case (launchedJs, ipAddresses) => launchedJs ++ Json.obj("ipAddresses" -> ipAddresses)
```
aquamatthias: whitespace?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
janisz: Good point, I'll add test for this case.

aquamatthias:
@janisz Thanks for this PR. Added a suggestion.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
janisz: I think sbt formatted this file. But surely it's unrelated with my change and only introduce noise. Will revert this change

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
janisz: Fixed

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
janisz: Fixed

aquamatthias:
Thanks! LGTM.

aquamatthias:
@janisz can you please squash your commits - and I will merge.

janisz:
@aquamatthias Squashed

wendigo:
Github has new feature @aquamatthias and you can squash PR during merge :)

aquamatthias:
@janisz Thanks!

PR #3633, opened 2016-04-01
Remove an illegal comma from JSON example

plombardi89:
JSON does not allow trailing commas after the last key in an object. Simple fix :)


aquamatthias:
Thanks!

PR #3634, opened 2016-04-01
Render deployment check results into the app listing.

aquamatthias:
Now with the right base.

### DIFF
--- a/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
+++ b/docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
@@ -110,10 +110,9 @@ The objects in the list only contain the id of the deployment, e.g.:
 ##### readiness (Array of Objects)
 
 A list of readiness check results.
-Only available if the app has readiness checks defined and also only available during a deployment 
-in the start phase.
+Only available during a deployment in the start phase.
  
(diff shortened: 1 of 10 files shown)


docs/docs/rest-api/mesosphere/marathon/api/v2/AppsResource_show.md
```diff
@@ -107,6 +107,27 @@ The objects in the list only contain the id of the deployment, e.g.:
 }
 ```
 
+##### readiness (Array of Objects)
+
+A list of readiness check results.
+Only available if the app has readiness checks defined and also only available during a deployment
```
gkleiman: I'd change this to `only available during a deployment in the start phase`.

gkleiman:
This looks good, but it would be awesome if it included more tests, as we have for the rest of the data we embed.

For example:

- Test that exercise the JSON serialization in `AppDefinitionAppInfoTest`.
- Test that the new embed option can be resolved (`InfoEmbedResolverTest`).

gkleiman:
```
     |     | |
    / \    | |
   |--o|===|-|
   |---|   | |
  /     \  | |
 | D     | | |
 | E     |=| |
 | U     | | |
 |_______| |_|
  |@| |@|  | |
___________|_|_
```

PR #3637, opened 2016-04-03
change word delimiter for `--no-logger`

pingz:
I suggest we using `--no_logger` instead of `--no-logger`, so it will be a consistent with other command options.

But this change would render the old scripts still using `--no-logger` useless ...

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -77,11 +77,11 @@ function load_options_and_log {
   fi
   echo "${cmd[@]}"
 
-  if [[ "${cmd[@]} $@" == *'--no-logger'* ]]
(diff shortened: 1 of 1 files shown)


aquamatthias:
@pingz This will break existing clients. Perhaps it makes more sense to introduce an env var that handles this case?
@lloesche do you have a better idea to solve this problem?

pingz:
If that is the case, an easy fix is insert `if [[ -z ${NO_LOGGER} ]];then cmd+=( --no-logger ); fi` before https://github.com/mesosphere/marathon/blob/master/bin/marathon-framework#L78

aquamatthias:
Close in favor of #3643

PR #3640, opened 2016-04-04
Cancel subscriptions if the task terminates.

aquamatthias:


src/main/scala/mesosphere/marathon/upgrade/ReadinessBehavior.scala
```diff
@@ -35,16 +35,35 @@ trait ReadinessBehavior { this: Actor with ActorLogging =>
...
 
   /**
     * Hook method which is called, whenever a task becomes ready according to the given app definition.
+    *
     * @param taskId the id of the task that has become ready.
     */
   def taskIsReady(taskId: Task.Id): Unit
 
+  /**
+    * Call this method, if a task terminates.
```
gkleiman: "Actors extending this trait should call this method when they detect that a task is terminated"

I think we also need an empty line after the "brief description" and maybe also between the details and the params.

gkleiman:
:+1:

Issue #3627, opened 2016-04-01
Improve documentation for persistent volumes


aquamatthias:
The documentation we have is not sufficient:
- how to setup mesos/marathon in order to use this feature (flags)
- explain principal
- does the example work? (should it be RELAUNCH_AFTER_TIMEOUT)

aquamatthias:
@joerg84 can you try an example as described in the docs to see what is missing? Thanks!

diatmpravin:
Potential Pitfalls[https://mesosphere.github.io/marathon/docs/persistent-volumes.html]
Replication and Backups

Because tasks are pinned to nodes, your data can be lost if the node is lost = What's it mean? 

Use Case: My Jenkins is running on Slave1 and it's having local persistence volume. If some how, jenkins killed and bring back by marathon on same slave then will I lost all task? Please explain..

PR #3641, opened 2016-04-04
Towards #3627 tiny improvements of the Resident Tasks doc

meichstedt:
- replace the wrong `taskLostBehavior` value with a correct value (which is the only working behavior right now)
- add a tiny clarification on what a _lost agent_ refers to


aquamatthias:
Thanks!
If I remember correctly, the task lost behavior is not effective anyway? 
Should be clear from the docs.

PR #3642, opened 2016-04-04
Fix TaskBuilderTest: Test would fail when

kolloch:
port range was randomly chosen as [32000-31000] in 1/1000 cases.


gkleiman:
:+1:

Issue #3621, opened 2016-03-31
allow for disabling output to syslog via environment variable


pingz:
Some guys (hi @rboyer)  files a same PR #1267, I opened another pull request #3620 for this.

I run marathon in docker with official rpm package, and I can't find a way to disable marathon process started logger, I really need this feature, please.

pingz:
OK, after reading related codes, now I know why environment variables like MARATHON_NO_LOGGER didn't work, `--no-logger` using normal Unix-like parameters using dash `-` separate the words, but the environment variables using  underscore `_`, and I can't set an environment variable such like `MARATHON_NO-LOGGER`.

I created another PR #3637, that could do the job .

aquamatthias:
Specifying env vars like `MARATHON_XXX` will result in options feed to the Marathon process. The option you are looking for is handled in the starting bash script.

meichstedt:
Note: This issue has been migrated to https://jira.mesosphere.com/browse/MARATHON-1905. For more information see https://groups.google.com/forum/#!topic/marathon-framework/khtvf-ifnp8.

PR #3643, opened 2016-04-04
introduce env var `NO_LOGGER` to disable syslog output

pingz:
Another approach for issue #3621.


aquamatthias:
Hey @pingz Have you tested this?
I guess you have to add this to the if clause?

lloesche:
https://github.com/mesosphere/marathon/pull/3651 should fix the issue without breaking backwards compatibility.

PR #3650, opened 2016-04-05
WIP: Simplifications

sttts:


src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -22,33 +21,25 @@ protected[volume] case object DVDIProvider
...
     v.persistent.name.each is notEmpty
+
     v.persistent.providerName is notEmpty
-    v.persistent.providerName.each is notEmpty
-    v.persistent.providerName.each is equalTo(name) // sanity check
-    v.persistent.options is valid(validOptions)
+    v.persistent.providerName.each is equalTo(name)
+
+    v.persistent.options.get(driverOption) is definedAnd(labelValidator)
+    (v.persistent.options.get(driverOption) is definedAnd(notEqualTo("rex-ray"))) or
```
jdef: s/rex-ray/rexray/

src/main/scala/mesosphere/marathon/core/volume/providers/Helpers.scala
```diff
@@ -23,21 +23,17 @@ protected[volume] abstract class AbstractPersistentVolumeProvider(
...
-  /** NamedOption represents a (named) configurable item type that provides validation rules */
-  abstract class NamedOption(
-      val namespace: String,
-      val name: String,
-      val required: Boolean,
-      val validValue: Validator[String]) {
-
-    def fullName: String = namespace + OptionNamespaceSeparator + name
-    def from(m: Map[String, String]): Option[String] = m.get(fullName)
+  def validIfDefined[T](implicit validator: Validator[T]): Validator[Option[T]] = new Validator[Option[T]] {
```
jdef: possibly refactored into our collection of reusable validators

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -218,12 +218,12 @@ object Group {
       group is noAppsAndGroupsWithSameName
       (group.id.isRoot is false) or (group.dependencies is noCyclicDependencies(group))
       group is validPorts
-      group is VolumesModule.validGroup
       group.groups is every(valid(validNestedGroup(group.id.canonicalPath(base))))
     }
 
   implicit val validRootGroup: Validator[Group] = new Validator[Group] {
     override def apply(group: Group): Result = {
+      group is VolumesModule.validGroup
```
jdef: why is this here instead of next to where port validation is ("group is validPorts")

jdef:
@sttts it looks like validation unit tests are failing for this PR

sttts:
There is one test intentionally breaking: empty `options`.

src/main/scala/mesosphere/marathon/state/Group.scala
sttts: /cc @kolloch

src/main/scala/mesosphere/marathon/core/volume/providers/Helpers.scala
sttts: /cc @kolloch any good place to put these? Or any other way to elegantly express the same?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -384,13 +341,27 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+      """.stripMargin
+    )
 
-    When("The request is processed")
-    val response = appsResource.create(body, false, auth.request)
+    Then("The return code indicates create success")
+    response.getStatus should be(201)
+  }
+
+  test("Creating an app with an external volume without options and DOCKER containerizer should pass validation") {
```
jdef: I'd rather have this test case fail. If the user wants to avoid specifying a driver then don't declare a persistent volume - use a docker volume w/ host-path set to volume-name

jdef: the above suggestion probably sounds ugly, but it's actually very docker-esque

sttts: We should at least fix the RuleViolation then.

PR #3672, opened 2016-04-06
Continued refactoring

sttts:


src/main/scala/mesosphere/marathon/core/volume/providers/Helpers.scala
```diff
@@ -23,7 +23,7 @@ protected[volume] abstract class AbstractExternalVolumeProvider(
 protected[providers] object OptionSupport {
   import OptionLabelPatterns._
 
-  def validIfDefined[T](implicit validator: Validator[T]): Validator[Option[T]] = new Validator[Option[T]] {
+  def ifDefined[T](implicit validator: Validator[T]): Validator[Option[T]] = new Validator[Option[T]] {
```
gkleiman: This is the same as https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/v2/Validation.scala#L21

src/main/scala/mesosphere/marathon/core/volume/providers/Helpers.scala
```diff
@@ -53,7 +57,7 @@ protected[providers] object OptionSupport {
   }
 
   /** a validator to enforce that values parse to booleans */
-  lazy val booleanValidator: Validator[String] = new Validator[String] {
+  lazy val validBoolean: Validator[String] = new Validator[String] {
```
gkleiman: could be written as `Validation.isTrue(s"Expected a valid boolean", Try(_.toBoolean).isSuccess)`

PR #3675, opened 2016-04-06
Fixes #3698 - External Volumes: DVDI

### LINKED ISSUE #3698: External Volumes: DVDI/rexray
The API will be extended such that users may request different types of "external" volumes distinguished by provider. 

Initial testing will focus on the Docker Volume Driver Interface provider and the "rexray" driver. For this use case, we also provide specific validation. Other provider/driver combinations should work in principal but are not tested.

Sample application definition JSON fragment:

```
      {
        "id": "/foo",
        "container": {
	   ...
          "volumes": [
            {
              "containerPath": "/docker_storage",
              "mode": "RW",
              // also supported for non-docker containers!
              "external": {
		    // Uniquely identifies the volume inside of the cluster for some provider driver
                 // combinations which provide cluster-wide volumes.
                 // For other providers this may only be unique per node, e.g. secret providers.
                "name": "my-external-volume",
                “provider”: “dvdi”,
                “options”: { “dvdi/iops”: 1000, “dvdi/driver”: “rexray” },
                “size”: 1234
              }
            },
            ...
          ]
        },
        ...
      }
```

There still seems to be a rather big bug in combination with rexray:

* [#88 mesos-module-dvdi: hung on unmount operation](https://github.com/emccode/mesos-module-dvdi/issues/88): Workaround, use the filesystem isolator of Mesos

Related epic (sorry, internal): [MARATHON-749](https://mesosphere.atlassian.net/browse/MARATHON-749)

sttts:
Based on https://github.com/mesosphere/marathon/pull/3525, squashed and rebased.

TODO:

* [x] "providerName" to "provider" in the protobufs
* [ ] docs in #3607 
  - upgradeStrategy requirements should be doc'd
  - prohibited chars for option values (as enforced by EMC mesos isolator)
* [x] restructuring to impl/providers, etc as per @kolloch
  - left top-level traits in top-level package because it doesn't make sense to expose lower level interfaces in a higher level package
* [x] wrap complex validators with `valid(...)` to get proper error messages 
* [x] use `container.volumes is every(valid(someValidator))` to get `volume(x)` indices in failure msg

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
+++ b/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
@@ -1,0 +1,264 @@
+package mesosphere.marathon.core.externalvolume.impl.providers
+
+import com.wix.accord._
+import com.wix.accord.dsl._
+import mesosphere.marathon.core.externalvolume.{ ExternalVolumeValidations, ExternalVolumeProvider }
+import mesosphere.marathon.state._
+import org.apache.mesos.Protos.Volume.Mode
+import org.apache.mesos.Protos.{ Volume => MesosVolume, CommandInfo, ContainerInfo, Environment }
+
+import OptionSupport._
+import scala.collection.JavaConverters._
+
+/**
+  * DVDIProvider (Docker Volume Driver Interface provider) handles external volumes allocated
+  * by a specific docker volume driver plugin. This works for both docker and mesos containerizers,
+  * albeit with some limitations:
+  *   - only a single volume driver per container is allowed when using the docker containerizer
+  *   - docker containerizer requires that referenced volumes be created prior to application launch
+  *   - mesos containerizer only supports volumes mounted in RW mode
+  */
+protected[externalvolume] case object DVDIProvider extends ExternalVolumeProvider {
+  override val name: String = "dvdi"
+
+  override def validations: ExternalVolumeValidations = DVDIProviderValidations
+
+  override def build(builder: ContainerInfo.Builder, ev: ExternalVolume): Unit = {
+    def toMesosVolume(volume: ExternalVolume): MesosVolume =
+      MesosVolume.newBuilder
+        .setContainerPath(volume.containerPath)
+        .setHostPath(volume.external.name)
+        .setMode(volume.mode)
+        .build
+
+    // special behavior for docker vs. mesos containers
+    // - docker containerizer: serialize volumes into mesos proto
+    // - docker containerizer: specify "volumeDriver" for the container
+    if (builder.getType == ContainerInfo.Type.DOCKER && builder.hasDocker) {
+      val driverName = ev.external.options(driverOption)
+      builder.setDocker(builder.getDocker.toBuilder.setVolumeDriver(driverName).build)
+      builder.addVolumes(toMesosVolume(ev))
+    }
+  }
+
+  override def build(containerType: ContainerInfo.Type, builder: CommandInfo.Builder, ev: ExternalVolume): Unit = {
+    // special behavior for docker vs. mesos containers
+    // - mesos containerizer: serialize volumes into envvar sets
+    if (containerType == ContainerInfo.Type.MESOS) {
+      val env = if (builder.hasEnvironment) builder.getEnvironment.toBuilder else Environment.newBuilder
+      val toAdd = volumeToEnv(ev, env.getVariablesList.asScala)
+      env.addAllVariables(toAdd.asJava)
+      builder.setEnvironment(env.build)
+    }
+  }
+
+  val driverOption = "dvdi/driver"
+  val dvdiVolumeContainerPath = "DVDI_VOLUME_CONTAINERPATH"
+  val dvdiVolumeName = "DVDI_VOLUME_NAME"
+  val dvdiVolumeDriver = "DVDI_VOLUME_DRIVER"
+  val dvdiVolumeOpts = "DVDI_VOLUME_OPTS"
+
+  private[providers] def volumeToEnv(
+    vol: ExternalVolume,
+    i: Iterable[Environment.Variable]): Iterable[Environment.Variable] = {
+
+    import OptionLabelPatterns._
+
+    val suffix = {
+      val offset = i.filter(_.getName.startsWith(dvdiVolumeName)).map{ s =>
+        val ss = s.getName.substring(dvdiVolumeName.length)
+        if (ss.length > 0) ss.toInt else 0
+      }.foldLeft(-1)((z, i) => if (i > z) i else z)
+
+      if (offset >= 0) (offset + 1).toString else ""
+    }
+
+    def mkVar(name: String, value: String): Environment.Variable =
+      Environment.Variable.newBuilder.setName(name).setValue(value).build
+
+    val vars = Seq[Environment.Variable](
+      mkVar(dvdiVolumeContainerPath + suffix, vol.containerPath),
+      mkVar(dvdiVolumeName + suffix, vol.external.name),
+      mkVar(dvdiVolumeDriver + suffix, vol.external.options(driverOption))
+    )
+
+    val optsVar = {
+      val prefix: String = name + OptionNamespaceSeparator
+      // don't let the user override these
+      val ignore = Set(driverOption)
+      // external.size trumps any user-specified dvdi/size option
+      val opts = vol.external.options ++ Map[String, String](
+        vol.external.size.map(prefix + "size" -> _.toString).toList: _*
+      )
+
+      // forward all dvdi/* options to the dvdcli driver, stripping the dvdi/ prefix
+      // and trimming the values
+      opts.filterKeys{ k =>
+        k.startsWith(prefix) && !ignore.contains(k.toLowerCase)
+      }.map{
+        case (k, v) => k.substring(prefix.length) + "=" + v.trim()
+      }.mkString(",")
+    }
+
+    if (optsVar.isEmpty) vars
+    else { vars :+ mkVar(dvdiVolumeOpts + suffix, optsVar) }
+  }
+}
+
+private[externalvolume] object DVDIProviderValidations extends ExternalVolumeValidations {
+  import mesosphere.marathon.api.v2.Validation._
+  import DVDIProvider._
+  import ViolationBuilder._
+
+  // group-level validation for DVDI volumes: the same volume name may only be referenced by a single
+  // task instance across the entire cluster.
+  override lazy val rootGroup = new Validator[Group] {
+    override def apply(g: Group): Result = {
+      val appsByVolume = g.transitiveApps.flatMap { app =>
+        app.externalVolumes.flatMap{ vol => nameOf(vol.external).map(_ -> app.id) }
+      }.groupBy[String](_._1).mapValues(_.map(_._2))
+
+      val groupViolations = g.apps.flatMap { app =>
+        val ruleViolations = app.externalVolumes.flatMap{ vol => nameOf(vol.external) }.flatMap{ name =>
+          for {
+            otherApp <- appsByVolume(name)
+            if otherApp != app.id // do not compare to self
+          } yield RuleViolation(app.id,
+            s"Requested volume $name conflicts with a volume in app $otherApp", None)
+        }
+        if (ruleViolations.isEmpty) None
+        else Some(GroupViolation(app, "app contains conflicting volumes", None, ruleViolations.toSet))
+      }
+      if (groupViolations.isEmpty) Success
+      else Failure(groupViolations.toSet)
+    }
+  }
+
+  override lazy val app = validator[AppDefinition] { app =>
+    //app should haveOnlyOneProvider // TODO(jdef) see comments on validator decl below
+    app should haveUniqueExternalVolumeNames
+    app should haveOnlyOneReplica
+    app.container is valid(optional(validContainer))
+    app.upgradeStrategy is valid(validUpgradeStrategy)
+  }
+
+  override lazy val volume = validator[ExternalVolume] { v =>
+    v.external.name is notEmpty
+    v.external.provider is equalTo(name)
+    v.external.options.get(driverOption) as s"external/options($driverOption)" is definedAnd(validLabel)
+    v.external.options is valid(
+      conditional[Map[String, String]](_.get(driverOption).contains("rexray"))(validRexRayOptions))
+  }
+
+  /**
+    * @return true if volume has a provider name that matches ours exactly
+    */
+  private[this] def accepts(volume: ExternalVolume): Boolean = {
+    volume.external.provider == name
+  }
+
+  private[this] def optionalOption(opts: Map[String, String], v: Validator[String]): Validator[String] =
+    new Validator[String] {
+      override def apply(optionName: String): Result = {
+        validate(opts.get(optionName))(optional(v)) match {
+          case Success     => Success
+          // we do this to get sane error messages
+          case Failure(vs) => Failure(vs.map(_.withDescription(optionName)))
+        }
+      }
+    }
+
+  private[this] lazy val validRexRayOptions: Validator[Map[String, String]] = validator[Map[String, String]] { opts =>
+    // there's probaby a better way to do this; this worked and didn't require me to duplicate option keys
+    "dvdi/volumetype" as "" is valid(optionalOption(opts, validLabel))
+    "dvdi/newfstype" as "" is valid(optionalOption(opts, validLabel))
+    "dvdi/iops" as "" is valid(optionalOption(opts, validNaturalNumber))
+    "dvdi/overwritefs" as "" is valid(optionalOption(opts, validBoolean))
+  }
+
+  private[this] def nameOf(vol: ExternalVolumeInfo): Option[String] = {
+    Some(vol.provider + "::" + vol.name)
+  }
+
+  private[this] case object haveOnlyOneReplica extends Validator[AppDefinition] {
+    override def apply(app: AppDefinition): Result =
+      if (app.externalVolumes.nonEmpty && app.instances > 1)
+        Failure(Set(RuleViolation(app.id,
+          s"Number of instances is limited to 1 when declaring DVDI volumes in app ${app.id}", None
+        )))
+      else Success
+  }
+
+  private[this] case object haveUniqueExternalVolumeNames extends Validator[AppDefinition] {
+    override def apply(app: AppDefinition): Result = {
+      val conflicts = volumeNameCounts(app).filter(_._2 > 1).keys
+      if (conflicts.isEmpty) Success
+      else Failure(conflicts.toSet[String].map { e =>
+        RuleViolation(app.id, s"Requested DVDI volume '$e' is declared more than once within app ${app.id}", None)
+      })
+    }
+  }
+
+  // for now this matches the validation for resident tasks, but probably won't be as
+  // restrictive in the future.
+  private[this] lazy val validUpgradeStrategy = validator[UpgradeStrategy] { strategy =>
+    strategy.minimumHealthCapacity should be <= 0.5
+    strategy.maximumOverCapacity should equalTo(0.0)
+  }
+
+  /** @return a count of volume references-by-name within an app spec */
+  private[this] def volumeNameCounts(app: AppDefinition): Map[String, Int] =
+    app.externalVolumes.flatMap{ v => nameOf(v.external) }.groupBy(identity).mapValues(_.size)
+
+  private[this] lazy val validMesosVolume = validator[ExternalVolume] { v =>
+    v.mode is equalTo(Mode.RW)
+  }
+
+  private[this] lazy val validMesosContainer = validator[Container] { ct =>
+    ct.volumes is every(valid(ifDVDIVolume(validMesosVolume)))
+  }
+
+  /*
+  // TODO(jdef) Unclear where this requirement came from. Seems super-useful to mix and match
+  // volumes from different providers within the same app (we only have 1 right now anyway, but
+  // the limitation imposed by this validator doesn't make sense to me).
+  private[this] case object haveOnlyOneProvider extends Validator[AppDefinition] {
+    override def apply(app: AppDefinition): Result = {
+      val uniqueProviderNames = app.externalVolumes.iterator.map(_.external.provider).toSet
+      val n = uniqueProviderNames.size
+      if (n <= 1) Success
+      else Failure(Set(RuleViolation(n, "one external volume provider per app", None)))
+    }
+  }
+  */
+
+  private[this] case object haveOnlyDriverOption extends Validator[ExternalVolume] {
+    override def apply(v: ExternalVolume): Result =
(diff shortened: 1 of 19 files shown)


src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,243 @@
+package mesosphere.marathon.core.volume.providers
+
+import com.wix.accord.{ Validator, _ }
+import com.wix.accord.dsl._
+import mesosphere.marathon.core.volume.providers.DVDIProvider._
+import mesosphere.marathon.core.volume.providers.OptionSupport._
+import mesosphere.marathon.state._
+import org.apache.mesos.Protos.Volume.Mode
+import org.apache.mesos.Protos.{ Volume => MesosVolume, CommandInfo, ContainerInfo, Environment }
```
kolloch: This is also fine but we started the convention of 

`import org.apache.mesos.{Protos => MesosProtos}` 

elsewhere. Then use `MesosProtos.Volume`, `MesosProtos.CommandInfo` etc to show clearly what is from the Mesos protos.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,243 @@
...
+  *   - only a single volume driver per container is allowed when using the docker containerizer
+  *   - docker containerizer requires that referenced volumes be created prior to application launch
+  *   - mesos containerizer only supports volumes mounted in RW mode
+  */
+protected[volume] case object DVDIProvider
+    extends AbstractExternalVolumeProvider("dvdi") with DVDIProviderValidations {
+
+  val driverOption = "dvdi/driver"
+
+  def toMesosVolume(volume: ExternalVolume): MesosVolume =
```
kolloch: make private or even embed it into the method below.

We prefer moving private methods below the public methods that use them.

kolloch: (Order from important to detail)

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,243 @@
...
+    MesosVolume.newBuilder
+      .setContainerPath(volume.containerPath)
+      .setHostPath(volume.external.name)
+      .setMode(volume.mode)
+      .build
+
+  def build(builder: ContainerInfo.Builder, v: Volume): Unit = v match {
```
kolloch: change signature to "v: T" in super trait and use `v: ExternalVolume` here then you can skip the match.

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,243 @@
...
+    case pv: ExternalVolume =>
+      // special behavior for docker vs. mesos containers
+      // - docker containerizer: serialize volumes into mesos proto
+      // - docker containerizer: specify "volumeDriver" for the container
+      if (builder.getType == ContainerInfo.Type.DOCKER && builder.hasDocker) {
+        val driverName = pv.external.options(driverOption)
```
kolloch: some debug-level log statement would be nice

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,243 @@
...
+        builder.addVolumes(toMesosVolume(pv))
+      }
+    case _ =>
+  }
+
+  def build(containerType: ContainerInfo.Type, builder: CommandInfo.Builder, pv: ExternalVolume): Unit = {
+    // special behavior for docker vs. mesos containers
+    // - mesos containerizer: serialize volumes into envvar sets
+    if (containerType == ContainerInfo.Type.MESOS) {
+      val env = if (builder.hasEnvironment) builder.getEnvironment.toBuilder else Environment.newBuilder
```
kolloch: some debug-level log statement would be nice

src/main/scala/mesosphere/marathon/core/volume/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,243 @@
...
+  val appValidation = validator[AppDefinition] { app =>
+    app should haveUniqueExternalVolumeNames
+    app should haveOnlyOneReplica
+    app.container is ifDefined(validContainer)
+    app.upgradeStrategy is validUpgradeStrategy
+  }
+
+  // group-level validation for DVDI volumes: the same volume name may only be referenced by a single
+  // task instance across the entire cluster.
+  val groupValidation = new Validator[Group] {
```
kolloch: I would probably rename this (and super and dependent names) to "rootGroup" validation and only call it on a rootGroup.

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -17,68 +20,81 @@ object Volume {
     containerPath: String,
     hostPath: Option[String],
     mode: Mesos.Volume.Mode,
-    persistent: Option[PersistentVolumeInfo]): Volume =
+    persistent: Option[PersistentVolumeInfo],
+    external: Option[ExternalVolumeInfo]): Volume = {
+
     persistent match {
       case Some(persistentVolumeInfo) =>
+        if (hostPath.isDefined) throw new IllegalArgumentException("hostPath may not be set with persistent")
```
kolloch: shouldn't that be caught by validation?

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -125,3 +143,82 @@ object PersistentVolume {
...
+  * @param size absolute size of the volume (MB)
+  * @param name identifies the volume within the context of the storage provider.
+  * @param providerName identifies the storage provider responsible for volume lifecycle operations.
+  * @param options contains storage provider-specific configuration configuration
+  */
+case class ExternalVolumeInfo(
+  size: Option[Long] = None,
+  name: String,
+  providerName: String,
+  options: Map[String, String] = Map.empty[String, String])
```
kolloch: Just an idea: Use an `OptionKey` as key here `case class OptionKey(providerPrefix: String, name: String)`

src/main/scala/mesosphere/marathon/state/Volume.scala
jdef: json formats calls this. maybe there's a way to implement a "Reads" (or whatever that arcane library uses) to validate this instead of doing it this way

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -122,23 +123,38 @@ class TaskBuilder(app: AppDefinition,
 
     val containerProto = computeContainerInfo(resourceMatch.hostPorts)
     val envPrefix: Option[String] = config.envVarsPrefix.get
+
+    def decorateForExternalVolumes(builder: CommandInfo.Builder) = containerProto.foreach { cp =>
+      app.container.foreach { container =>
```
jdef: app.externalVolumes.foreach?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -252,6 +264,191 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+
+    Then("The return code indicates that the hostPath of volumes[0] is missing") // although the wrong field should fail
+    response.getStatus should be(422)
+    response.getEntity.toString should include("/container/volumes(0)/external/providerName")
+    response.getEntity.toString should include("is unknown provider")
+  }
+
+  test("Creating an app with an external volume with no name provider name specified should FAIL provider validation") {
+    Given("An app with an unnamed volume provider")
+    val e = intercept[JsResultException]{
```
jdef: why do we need to special case this, vs dealing with error codes?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -252,6 +264,191 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+      createAppWithVolumes("MESOS",
+        """
+          |    "volumes": [{
+          |      "containerPath": "/var",
+          |      "hostPath": "/var",
```
jdef: hostPath is invalid here

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -246,6 +246,180 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
+      else None
+    }
+
+    assert(taskInfo.getContainer.getVolumesList.size > 0, "check that container has volumes declared")
+    assert(!taskInfo.getContainer.getDocker.hasVolumeDriver, "docker spec should not define a volume driver")
+    assert(vol("namedFoo").isDefined,
+      s"missing expected volume namedFoo, got instead: ${taskInfo.getContainer.getVolumesList}")
+  }
+
+  // TODO(jdef) test both dockerhostvol and persistent extvol in the same docker container
```
jdef: TODO

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -246,6 +246,180 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
+        cmd = Some("foo"),
+        cpus = 1.0,
+        mem = 32.0,
+        executor = "//cmd",
+        portDefinitions = Nil,
+        container = Some(Container(
+          docker = Some(Docker()), // must have this to force docker container serialization
+          `type` = MesosProtos.ContainerInfo.Type.DOCKER,
+          volumes = Seq[Volume](
+            // TODO(jdef) probably need to validate uniquess of hostPaths's that reference volume
```
jdef: TODO

src/main/scala/mesosphere/marathon/core/externalvolume/VolumesModule.scala
```diff
@@ -36,30 +41,34 @@ trait ExternalVolumeProviderRegistry {
...
+      case Some(p) => p.validations.volume(ev)
       case None    => Failure(Set(RuleViolation(None, "is unknown provider", Some("external/providerName"))))
     }
   }
 
   /** @return a validator that checks the validity of a container given the related volume providers */
   def validApp(): Validator[AppDefinition] = new Validator[AppDefinition] {
-    def apply(app: AppDefinition) = providers.all.map(_.appValidation).map(validate(app)(_)).fold(Success)(_ and _)
+    def apply(app: AppDefinition) = {
+      val appProviders = app.externalVolumes.iterator.flatMap(ev => providers.get(ev.external.provider)).toSet
```
jdef: why iterator here?

jdef: @kolloch

src/main/scala/mesosphere/marathon/core/externalvolume/providers/DVDIProvider.scala
```diff
@@ -18,19 +18,19 @@ import scala.collection.JavaConverters._
   *   - docker containerizer requires that referenced volumes be created prior to application launch
   *   - mesos containerizer only supports volumes mounted in RW mode
   */
-protected[externalvolume] case object DVDIProvider
-    extends AbstractExternalVolumeProvider("dvdi") with DVDIProviderValidations {
+protected[externalvolume] case object DVDIProvider extends ExternalVolumeProvider {
+  override val name: String = "dvdi"
```
jdef: wasn't this previously declared as a def?

src/main/scala/mesosphere/marathon/core/externalvolume/providers/DVDIProvider.scala
```diff
@@ -101,143 +105,142 @@ protected[externalvolume] case object DVDIProvider
...
+    override def apply(app: AppDefinition): Result = {
+      val uniqueProviderNames = app.externalVolumes.iterator.map(_.external.provider).toSet
+      val n = uniqueProviderNames.size
       if (n <= 1) Success
       else Failure(Set(RuleViolation(n, "one external volume driver per app", None)))
     }
   }
 
-  private val haveOnlyDriverOption = new Validator[ExternalVolume] {
+  private[this] case object haveOnlyDriverOption extends Validator[ExternalVolume] {
```
jdef: why make this a case object? we're already scoped by a singleton?

src/main/scala/mesosphere/marathon/core/externalvolume/providers/DVDIProvider.scala
kolloch: you may override defs with vals

kolloch: In this case it does not matter. A val takes up memory for every instance, a def "calculates" the value every time. If it is a static string and there is only one instance, it does not matter.

src/main/scala/mesosphere/marathon/core/externalvolume/providers/DVDIProvider.scala
kolloch: basically using vals in "global" scope in Scala is dangerous. If you access them at the wrong time (during initialization), they can be null. Both `lazy val` and `case object` avoid this pitfall.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
jdef: @kolloch what's a better way to do this?

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,265 @@
...
+  private[this] lazy val validMesosVolume = validator[ExternalVolume] { v =>
+    v.mode is equalTo(Mode.RW)
+  }
+
+  private[this] lazy val validMesosContainer = validator[Container] { ct =>
+    ct.volumes is every(valid(ifDVDIVolume(validMesosVolume)))
+  }
+
+  /*
+  // TODO(jdef) Unclear where this requirement came from. Seems super-useful to mix and match
```
jdef: @sttts where did this come from?

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
```diff
@@ -0,0 +1,265 @@
...
+  private[this] lazy val validDockerContainer = validator[Container] { ct =>
+    ct.volumes is every(valid(ifDVDIVolume(haveOnlyDriverOption)))
+    ct.volumes is every(valid(ifDVDIVolume(haveNoSize)))
+  }
+
+  private[this] case object validContainer extends Validator[Container] {
+    override def apply(ct: Container): Result = ct.`type` match {
+      case ContainerInfo.Type.MESOS  => validate(ct)(validMesosContainer)
+      case ContainerInfo.Type.DOCKER => validate(ct)(validDockerContainer)
+      // TODO(jdef) handler for unexpected container types?
```
jdef: TODO

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
sttts: That was there before my refactoring. We actually have no idea how this will be with more than one provider. Docker only supports one driver at a time at least.

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
sttts: The match is exhaustive and the compiler will check that.

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
```diff
@@ -138,14 +138,15 @@ private[externalvolume] object DVDIProviderValidations extends ExternalVolumeVal
     app should haveUniqueExternalVolumeNames
     app should haveOnlyOneReplica
     app.container is optional(validContainer)
-    app.upgradeStrategy is validUpgradeStrategy
+    app.upgradeStrategy is valid(validUpgradeStrategy)
```
sttts: I wonder whether the `valid` shouldn't be inside `validUpgradeStrategy`.

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
```diff
@@ -138,14 +138,15 @@ private[externalvolume] object DVDIProviderValidations extends ExternalVolumeVal
...
   }
 
   override lazy val volume = validator[ExternalVolume] { v =>
     v.external.name is notEmpty
     v.external.provider is equalTo(name)
     v.external.options.get(driverOption) as s"external/options($driverOption)" is definedAnd(validLabel)
-    v.external.options is conditional[Map[String, String]](_.get(driverOption).contains("rexray"))(validRexRayOptions)
+    v.external.options is valid(
```
sttts: Same here.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -218,40 +219,256 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+         |    $docker
+         |    $volumes
+         |  }
+         |}
+      """.stripMargin
+
+    groupManager.updateApp(any, any, any, any, any) returns Future.successful(plan)
+
+    When("The request is processed")
+    appsResource.create(body.getBytes("UTF-8"), false, auth.request)
```
kolloch: you should probably check the response, right?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
kolloch: Well, this is good enough for now.

Basically, we have two ways to indicate errors: One is throwing an exception and one is returning an error response. Ultimately, in a running Marathon instance, all exceptions are also mapped to a response (see `MarathonExceptionMapper`). This is done in Chaos-Framework-Code.

If you wanted to emulate that behavior, you could do something like (probably factor that out into a `MarathonExceptionMapperTestHelper` or similar):

```scala
    val response = Try {
      createAppWithVolumes("MESOS",
        """
          |    "volumes": [{
          |      "containerPath": "/var",
          |      "external": {
          |        "size": 10
          |      },
          |      "mode": "RW"
          |    }]
        """.stripMargin
      )
    }.recover { case NonFatal(e: Exception) => (new MarathonExceptionMapper).toResponse(e) }.get
```

It is part of your testing philosophy if you want thus to extend the scope of your testing to include other components or not.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -218,40 +219,256 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+  test("Creating an app with an external volume and DOCKER containerizer should pass validation") {
+    Given("An app with a named, non-'agent' volume provider")
+    val response = createAppWithVolumes("DOCKER",
+      """
+        |    "volumes": [{
+        |      "containerPath": "/var",
+        |      "external": {
+        |        "provider": "dvdi",
+        |        "name": "namedfoo",
+        |        "options": {"dvdi/driver": "bar"}
```
kolloch: I question whether this should pass validation because the options are currently ignored. @jdef @sttts

src/test/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDITest.scala
```diff
@@ -0,0 +1,182 @@
...
+import mesosphere.marathon.state._
+import org.apache.mesos.Protos.Volume.Mode
+import org.scalatest.Matchers
+
+sealed trait TCHelpers {
+  val EVI = ExternalVolumeInfo.apply _
+  val EV = ExternalVolume.apply _
+}
+
+class DVDIProvider_VolumeValidationTest extends MarathonSpec with Matchers with TCHelpers {
```
kolloch: Remove the "_"

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
kolloch: No, we shouldn't. Missed the context.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
sttts: Don't understand "because the options are ignored".

kolloch: In the docker case, the options do not have an effect, correct? They are only passed to mesos as environment variables if the container type is Mesos, correct?

sttts: The driver is the docker volume driver. It's the only valid (and required) option.

kolloch:
````                       $o
                       $                     .........
                      $$$      .oo..     'oooo'oooo'ooooooooo....
                       $       $$$$$$$
                   .ooooooo.   $$!!!!!
                 .'.........'. $$!!!!!      o$$oo.   ...oo,oooo,oooo'ooo''
    $          .o'  oooooo   '.$$!!!!!      $$!!!!!       'oo''oooo''
 ..o$ooo...    $                '!!''!.     $$!!!!!
 $    ..  '''oo$$$$$$$$$$$$$.    '    'oo.  $$!!!!!
 !.......      '''..$$ $$ $$$   ..        '.$$!!''!
 !!$$$!!!!!!!!oooo......   '''  $$ $$ :o           'oo.
 !!$$$!!!$$!$$!!!!!!!!!!oo.....     ' ''  o$$o .      ''oo..
 !!!$$!!!!!!!!!!!!!!!!!!!!!!!!!!!!ooooo..      'o  oo..    $
  '!!$$!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!oooooo..  ''   ,$
   '!!$!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!oooo..$$
    !!$!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!$'
    '$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$!!!!!!!!!!!!!!!!!!,
.....$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$.....
```

(Thanks http://www.asciiworld.com/-Boats-.html)

PR #3677, opened 2016-04-06
Fixes #3654 - Extract port labels from port mappings

### LINKED ISSUE #3654: portMapping labels are not being set
When creating a `BRIDGE` mode container with `portMappings` the app config returned by Marathon  shows the label `VIP_0` to be there, but when checking Mesos `master/state.json` the labels are not there.

gkleiman:


aquamatthias:
LGTM - Thanks!
Will wait for the tests.

PR #3679, opened 2016-04-06
Added fixed issues in RC2

aquamatthias:


gkleiman:
We never add fixes included in a RC to the changelog, unless they fix an issue affecting the current stable version.

Fixes to issues in the previous RC belong to the release notes in GitHub and to the announcement e-mail.

PR #3680, opened 2016-04-06
External Volumes refactorings

kolloch:
* s/provider/providerName/
* extracted ExternalVolumeValidations, refactored them somewhat but
  they can obviously be simplified a lot further
* made it more obvious what the API of an object is and what
  is implementation
* s/validGroup/validRootGroup/
* filter providers for app validation

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
+++ b/src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
@@ -89,7 +89,7 @@ object ExternalVolumeInfoSerializer {
   def toProto(info: ExternalVolumeInfo): Protos.Volume.ExternalVolumeInfo = {
     val builder = Protos.Volume.ExternalVolumeInfo.newBuilder()
       .setName(info.name)
-      .setProviderName(info.providerName)
+      .setProviderName(info.provider)
 
     info.size.foreach(builder.setSize)
     info.options.map{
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -89,7 +89,7 @@ object ExternalVolumeInfoSerializer {
   def toProto(info: ExternalVolumeInfo): Protos.Volume.ExternalVolumeInfo = {
     val builder = Protos.Volume.ExternalVolumeInfo.newBuilder()
       .setName(info.name)
-      .setProviderName(info.providerName)
+      .setProviderName(info.provider)
```
sttts: We should change the protobuf as well (in a later PR)

src/main/scala/mesosphere/marathon/core/externalvolume/VolumesModule.scala
```diff
@@ -36,30 +41,34 @@ trait ExternalVolumeProviderRegistry {
...
+      appProviders.map(_.validations.app).map(validate(app)(_)).fold(Success)(_ and _)
+    }
   }
 
   /** @return a validator that checks the validity of a group given the related volume providers */
-  def validGroup(): Validator[Group] = new Validator[Group] {
-    def apply(grp: Group) = providers.all.map(_.groupValidation).map(validate(grp)(_)).fold(Success)(_ and _)
+  def validRootGroup(): Validator[Group] = new Validator[Group] {
+    def apply(grp: Group) =
+      providers.all.map(_.validations.rootGroup).map(validate(grp)(_)).fold(Success)(_ and _)
```
sttts: Here we need the same fix as above.

kolloch: actually, I do not think so. The validation has also to work for nested groups/apps which do not need the specific provider.

sttts: The validation takes the transitive app list and check all of those. Isn't that enough?

PR #3681, opened 2016-04-06
Limit group validation to the used providers

sttts:


src/main/scala/mesosphere/marathon/core/externalvolume/VolumesModule.scala
```diff
@@ -68,7 +68,11 @@ object VolumesModule {
 
   /** @return a validator that checks the validity of a group given the related volume providers */
   def validRootGroup(): Validator[Group] = new Validator[Group] {
-    def apply(grp: Group) =
-      providers.all.map(_.validations.rootGroup).map(validate(grp)(_)).fold(Success)(_ and _)
+    def apply(grp: Group) = {
+      val groupProvider = grp.transitiveApps.flatMap(_.externalVolumes).flatMap {
```
jdef: does it make sense to `toSet` the providers?

jdef: nevermind, it's already a set

jdef:
lgtm

src/main/scala/mesosphere/marathon/core/externalvolume/VolumesModule.scala
kolloch: and BTW, that it is a `Set` is a really bad idea ;)

kolloch:
I think that this PR is not necessary...

Issue #3616, opened 2016-03-30
Scheduler Migrations: Fix host port detection for readiness checks


kolloch:
Unfortunately, in case of a defined `ipAddress` we have to inspect the `DiscoveryInfo` instead of the `PortDefinition`s.

Also make sure that we correctly inspect the `portMappings` in case a docker image is used.

#3600 was incorrect.

PR #3682, opened 2016-04-06
Towards #3616 - Introduce AppDefinition.portAssignments

gkleiman:


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -365,6 +366,38 @@ case class AppDefinition(
...
+      }.toList
+
+    def fromPortDefinitions: Option[Seq[PortAssignment]] = task.launched.map { launched =>
+      portDefinitions.zip(launched.hostPorts).zipWithIndex.map {
+        case ((portDefinition, hostPort), portIndex) =>
+          PortAssignment(portDefinition.name, portIndex, task.agentInfo.host, hostPort)
+      }
+    }
+
+    if (ipAddress.isDefined) fromIpAddress // If an IP-per-Task was requested, don't fall back
```
kolloch: could also be achieved by returning "Some(Nil)" if there is an ipAddress but not assignments (e.g. because the task is not launched)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -365,6 +366,38 @@ case class AppDefinition(
...
+    else fromPortMappings.orElse(fromPortDefinitions)
```
kolloch: Is it really ok to fall-back to portDefinitions in this case? What about a BRIDGE-networking Task without any portMappings?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
kolloch: but maybe something like `if (ipAddress.isDefined)` is good. Maybe it would even be nice if the top-level thing here was kind of self-explanatory, e.g. we use discoveryInfos for ip-per-container, we use portMappings for docker bridge mode and we use portDefinitions for docker HOST mode and non-docker apps.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
gkleiman: Correct me if I'm wrong, but in that case `task.launched.hostPorts` should be empty, so it is ok to for the method to return `Some(Seq.empty)`.

gkleiman: Answering myself: it doesn't make any sense, but Marathon allows you to create an app with `bridged` networking, no port mappings and multiple host ports, like this:

```json
{
  "id": "foo",
  "cmd": null,
  "cpus": 1,
  "mem": 128,
  "disk": 0,
  "instances": 1,
  "container": {
    "docker": {
      "image": "python:2.7-alpine",
      "network": "BRIDGE"
    },
    "type": "DOCKER"
  },
  "ports": [1,2,3]
}
```

kolloch: Well, then the app is not reachable from the outside via any host ports, therefore we do not need to return them.

Or what do you think?

gkleiman:
Superseded by #3702

PR #3685, opened 2016-04-06
Resolves #3673 by only considering unreserved resources

### LINKED ISSUE #3673: Resident tasks: Disallow "acceptedResourceRoles" != ["*"]
Mesos treats resources which already have a role assignment != "*" as "reserved". Therefore, they cannot be reserved again. To change that, the reserve op would need to specify the resources to relabel as well as the goal reservation state. Since we cannot change that short term, we should disallow resident apps which target roles != "*".

/cc @aquamatthias @meichstedt

meichstedt:
for reserve/create operations.

Resources that are already associated with a role (are already reserved) cannot be reserved again. Marathon will now only consider unreserved resources when trying to reserve them for resident tasks. As a result, it makes no sense to define acceptedResourceRoles for resident apps, which is now prevented using a validator.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
@@ -110,12 +110,12 @@ class TaskOpFactoryImpl @Inject() (
 
     def maybeReserveAndCreateVolumes = if (needToReserve) {
       val configuredRoles = app.acceptedResourceRoles.getOrElse(config.defaultAcceptedResourceRolesSet)
-      if (configuredRoles != Set(ResourceRole.Unreserved)) {
-        log.warn("Ignoring acceptedResourcesRoles for {}. Only unreserved resources can be reserved", app.id)
+      // We can only reserve unreserved resources
+      val rolesToConsider = Set(ResourceRole.Unreserved).intersect(configuredRoles)
+      if (configuredRoles.isEmpty) {
+        log.warn(s"Will never match for ${app.id} as the app is configured to only accept $configuredRoles")
       }
 
-      // We can only reserve resources that are unreserved, which means they are applicable only for role(*)
-      val rolesToConsider = Set(ResourceRole.Unreserved)
       val matchingResourcesForReservation =
         ResourceMatcher.matchResources(
           offer, app, tasks.values,
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
@@ -5,7 +5,7 @@ import mesosphere.marathon.MarathonConf
 import mesosphere.marathon.core.base.Clock
 import mesosphere.marathon.core.launcher.{ TaskOp, TaskOpFactory }
 import mesosphere.marathon.core.task.{ Task, TaskStateOp }
-import mesosphere.marathon.state.AppDefinition
+import mesosphere.marathon.state.{ ResourceRole, AppDefinition }
 import mesosphere.mesos.ResourceMatcher.ResourceSelector
 import mesosphere.mesos.{ PersistentVolumeMatcher, ResourceMatcher, TaskBuilder }
 import mesosphere.util.state.FrameworkId
@@ -109,8 +109,13 @@ class TaskOpFactoryImpl @Inject() (
     else None
 
     def maybeReserveAndCreateVolumes = if (needToReserve) {
+      val configuredRoles = app.acceptedResourceRoles.getOrElse(config.defaultAcceptedResourceRolesSet)
+      if (configuredRoles != Set(ResourceRole.Unreserved)) {
+        log.warn("Ignoring acceptedResourcesRoles for {}. Only unreserved resources can be reserved", app.id)
+      }
+
       // We can only reserve resources that are unreserved, which means they are applicable only for role(*)
-      val rolesToConsider = ResourceSelector.NoRole
+      val rolesToConsider = Set(ResourceRole.Unreserved)
       val matchingResourcesForReservation =
         ResourceMatcher.matchResources(
           offer, app, tasks.values,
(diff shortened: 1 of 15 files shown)


src/test/scala/mesosphere/marathon/api/validation/AppDefinitionValidatorTest.scala
```diff
@@ -452,6 +452,23 @@ class AppDefinitionValidatorTest extends MarathonSpec with Matchers with GivenWh
...
+    val from = f.validResident
+    AllConf.SuppliedOptionNames = Set("mesos_authentication_principal", "mesos_role", "mesos_authentication_secret_file")
+
+    When("validating with acceptedResourceRoles")
+    val to1 = from.copy(acceptedResourceRoles = Some(Set("foo")))
+    Then("Should be invalid")
+    AppDefinition.validAppDefinition(to1).isSuccess shouldBe false
+
+    When("validating without acceptedResourceRoles")
+    val to2 = from.copy(acceptedResourceRoles = None)
```
kolloch: you should also test with Set(NoRole)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -495,6 +495,7 @@ object AppDefinition {
     appDef.disk should be >= 0.0
     appDef must definesCorrectResidencyCombination
     (appDef.isResident is false) or (appDef.upgradeStrategy is UpgradeStrategy.validForResidentTasks)
+    (appDef.isResident is false) or (appDef.acceptedResourceRoles is empty)
```
kolloch: Or `Set("*")`

kolloch:
Looks nearly perfect. But you also have to allow `"acceptedResourceRoles": ["*"]`

src/test/scala/mesosphere/marathon/api/validation/AppDefinitionValidatorTest.scala
meichstedt: The thing is that I explicitly **do not** use the applications acceptedResourceRoles for matching for resources to reserve, because there's no way to consider anything other than unreserved resources (unless we want to add support for statically reserved resources, which is not recommended by Mesos). So my initial inclination was to not allow setting it at all and adding a meaningful validation error. (I didn't add the latter because there are changes in the scheduler feature branch that would most probably have conflicted then).

Does that make sense to you?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
meichstedt: See my comment above

meichstedt:
Thanks for that quick review, @kolloch 

As noted above, tell me whether you'd be OK with a clear validation error, something like `resident apps cannot define acceptedResourceRoles`. Reservations can only be processed based on unreserved resources, and I feel better with the `TaskOpFactory` explicitly using `ResourceSelector.NoRole` instead of the apps `acceptedResourceRoles` or the default ones.

src/test/scala/mesosphere/marathon/api/validation/AppDefinitionValidatorTest.scala
kolloch: Maybe, I misunderstand something. I completely understand that only checking whether `acceptedResourceRoles` is set, is not sufficient. You basically have to check the "effective" acceptedResourceRoles. And the effective `acceptedResourceRoles` needs to be `Set("*")`.

Your code ignores this user setting. That means that the user said it that the task should only be run on a specific role (by the global setting since you disallow the other) and you explicitly ignore this. That is not good either.

Furthermore, we should allow `Set("*")` since it pretty much expresses what we do.

Since the global default may change, you can have an additional safe-guard at the place of making the reservation which checks whether we attempt to match resources other than "*".

Does that make sense?

kolloch:
I think that your proposal is a bit inconsistent. Our usual logic is that _all_ apps have an effective acceptResourcesRole. Either an explicit override or a global default. Your code disallows a correct override of this setting and allows a scenario where the global default is incorrect without override.

To play devils advocate, we already chose defaults differently depending on context elsewhere, so I guess it is OK to treat no overridden `acceptedResourceRoles` smartly and use an effective `acceptedResourceRoles` that is compatible with reservations. Therefore my suggestion was to also allow the override which would have the same effect for normal apps, at least.

Does that make sense? Or another question: why is it important to you to disallow setting `acceptedResourceRoles` to `["*"]`?

meichstedt:
Hey @kolloch thanks for clarifying this. I'm now allowing correct acceptedResourceRoles and logging a warning when a reservation is actually tried but the effective acceptedResourceRoles do not contain "*"

I've added #3704 to improve on this; changing the validation to allow config aware checks is a bit more workload than I'd like to have in here (given the time we have ..)

kolloch:
please rebase

meichstedt:
rebased

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
```diff
@@ -113,10 +109,17 @@ class TaskOpFactoryImpl @Inject() (
     else None
 
     def maybeReserveAndCreateVolumes = if (needToReserve) {
+      val configuredRoles = app.acceptedResourceRoles.getOrElse(config.defaultAcceptedResourceRolesSet)
+      if (configuredRoles != Set(ResourceRole.Unreserved)) {
+        log.warn("Ignoring acceptedResourcesRoles for {}. Only unreserved resources can be reserved", app.id)
```
kolloch: You should FAIL if configuredRoles does not contain `ResourceRole.Unreserved`.

kolloch: And you should maybe not even warn if `ResourceRole.Unreserved` is contained in the configured roles... or?

meichstedt: as suggested, the code will now never match if `ResourceRole.Unreserved` is not contained in the configured roles, and log a warning in this case

aquamatthias:
Perhaps we can even improve this logic to 

- set `acceptedResourceRoles` automatically to `*` if not set
- make sure `*` is configured as resource role

Looks good for now.

Thanks!
```
            .
            |\
            | \
            |  \
            |   \
            |    \
            |     \
            |      \
            |       \       ____O
            |        \     .' ./
            |   _.,-~"\  .',/~'
            <-~"   _.,-~" ~ |
^"~-,._.,-~"^"~-,._\       /,._.,-~"^"~-,._.,-~"^"~-,._
~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._
^"~-,._.,-~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._
~-,._.,-~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._.,-~"^"~-,._
```

PR #3687, opened 2016-04-06
Fixes #3682 - Use portAssignments for readiness checks

gkleiman:
**WARNING**: this PR targets gk/add-port-assignments, DO NOT merge.


src/main/scala/mesosphere/marathon/core/readiness/ReadinessCheckExecutor.scala
```diff
@@ -56,18 +52,18 @@ object ReadinessCheckExecutor {
...
-              checkDef.portName,
-              throw new IllegalArgumentException(s"no port definition for port name '${checkDef.portName}' was found")
-            )
 
-            if (app.hasFixedHostPorts) app.portDefinitions(portIndex).port
-            else launched.hostPorts(portIndex)
-          }
+          val portAssignmentsByName = app.portAssignments(task).getOrElse(
+            throw new IllegalStateException(s"no ports assignments for AppDefinition: [$app] - Task: [$task]")
+          ).groupBy(_.portName)
```
kolloch: if the key is unique, I prefer `.map(pa => p.portName -> pa).toMap`

My plan that @aquamatthias forwarded to you is actually to introduce a `PortAssignments` container that has a method `byName(portName: String): Option[PortAssignment]`. That would result in `app.portAssignments(task).flatMap(_.byName(checkDef.portName))` below

I still like that better but this is also ok.

kolloch:
Nice job! Only nit-picks, would also merge as-is.

gkleiman:
Superseded by #3712

PR #3702, opened 2016-04-07
Towards #3616 - Introduce AppDefinition.portAssignments

gkleiman:
See #3682 for a previous version based on the feature branch.

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/PortAssignment.scala
+++ b/src/main/scala/mesosphere/marathon/state/PortAssignment.scala
@@ -1,5 +1,7 @@
 package mesosphere.marathon.state
 
+case class PortAssignment(portName: Option[String], portIndex: Int, effectiveIpAddress: String, effectiveHostPort: Int)
+
 object PortAssignment {
   /**
     * If you change this, please also update AppDefinition.json.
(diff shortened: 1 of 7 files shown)


src/main/scala/mesosphere/marathon/state/PortAssignment.scala
```diff
@@ -1,5 +1,7 @@
 package mesosphere.marathon.state
 
+case class PortAssignment(portName: Option[String], portIndex: Int, effectiveIpAddress: String, effectiveHostPort: Int)
```
kolloch: Some documentation would be nice that explains "effective".

src/main/scala/mesosphere/marathon/state/PortAssignment.scala
```diff
@@ -1,5 +1,13 @@
 package mesosphere.marathon.state
 
+/**
+  * @param portName name of the port
+  * @param portIndex index of the port, used for example to define health checks
+  * @param effectiveIpAddress ip address on which the port can be reached (can be an agent's IP or an IP-per-Task)
```
kolloch: IP address

src/main/scala/mesosphere/marathon/state/PortAssignment.scala
```diff
@@ -1,5 +1,13 @@
...
+  * @param effectivePort resolved non-dynamic port. The task is reachable under effectiveIpAddress:effectivePort.
```
kolloch: The port which is reachable from the outside?? 

I like "The task is reachable under effectiveIpAddress:effectivePort".

kolloch:
Beautiful tests!

```

           .--------------.
           |     _._ _     `.,______
           |    (__((_(       ___(_()
           |  \'-:--:-.    ,'`
           '-|-'-----'--|-'
             b          d   
```

PR #3713, opened 2016-04-07
Fixes #3699 - external volumes with update strategy

### LINKED ISSUE #3699: External Volumes: Choose a matching upgrade strategy by default
External volumes require specific upgrade strategy settings. These should be chosen by default if the user does not specify anything explicitly similarly to what we do for "persistent" volumes.

Depends on #3698.

Related epic (sorry, internal): [MARATHON-749](https://mesosphere.atlassian.net/browse/MARATHON-749)

lexwbr:
According to #3699 defined external volumes should also create a standard update strategy.


lexwbr:
@kolloch tests are green. Please let me know if the result conforms to what you expected it to be.

aquamatthias:
Looks good - Thanks!

PR #3718, opened 2016-04-08
Task State Diagram

kolloch:


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -15,6 +15,37 @@ import scala.collection.immutable.Seq
...
+                                    |                             |
+                                    |   +---------------------+   |
+                        match offer |   |LaunchedOnReservation|   | terminal
+                        & launch    |   |                     |   | status
+                                    |   | Started / Staged    |   | update
+                                    |   | Running             |   |
+                                    +-> |                     +---+
+                                        +---------------------+
+
+  * ```
```
aquamatthias: How cool is that?

meichstedt: Shouldn't there be a _powered by ASCIFlow_ note? ;)

meichstedt:
👍

Issue #2419, opened 2015-10-13
ZK CandidateImpl instances leaked after leader elections


gkleiman:
Steps to reproduce:

1. Run: `while sleep 2; do http DELETE :8080/v2/leader; done`
2. Look at the logs and see that there are multiple log messages from `com.twitter.common.zookeeper.CandidateImpl` referencing ZK Member IDs that no longer exist:

```
[...]
marathon [2015-10-13 14:33:48,928] ERROR Current member ID member_0000000014 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,929] ERROR Current member ID member_0000000007 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,929] ERROR Current member ID member_0000000010 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,929] ERROR Current member ID member_0000000003 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,930] ERROR Current member ID member_0000000015 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,930] ERROR Current member ID member_0000000018 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,930] ERROR Current member ID member_0000000011 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,931] ERROR Current member ID member_0000000006 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
marathon [2015-10-13 14:33:48,931] ERROR Current member ID member_0000000008 is not a candidate for leader, current voting: [member_0000000019] (com.twitter.common.zookeeper.CandidateImpl:144)
[...]

PR #3732, opened 2016-04-11
Fix Zookeeper group member listener leak in leader election

sttts:
Fixes https://github.com/mesosphere/marathon/issues/2419#issuecomment-208214595.

Changes:
- fork Twitter CandidateImpl.java
- watch only once per CandidateImpl life time

TODO:
- [ ] ~~try to add tests which capture the leak problem~~

### DIFF
--- a/src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
+++ b/src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
@@ -1,0 +1,180 @@
+package mesosphere.marathon.core.leadership;
+
+// =================================================================================================
+// Copyright 2011 Twitter, Inc.
+// -------------------------------------------------------------------------------------------------
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this work except in compliance with the License.
+// You may obtain a copy of the License in the LICENSE file, or at:
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+// =================================================================================================
+
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Ordering;
+import com.twitter.common.base.Command;
+import com.twitter.common.base.ExceptionalCommand;
+import com.twitter.common.zookeeper.Candidate;
+import com.twitter.common.zookeeper.Group;
+import com.twitter.common.zookeeper.ZooKeeperClient;
+import org.apache.zookeeper.KeeperException;
+
+import javax.annotation.Nullable;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.google.common.base.Charsets;
+import com.google.common.base.Function;
+import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Supplier;
+
+/**
+ * Implements leader election for small groups of candidates.  This implementation is subject to the
+ * <a href="http://hadoop.apache.org/zookeeper/docs/r3.2.1/recipes.html#sc_leaderElection">
+ * herd effect</a> for a given group and should only be used for small (~10 member) candidate pools.
+ */
+public class CandidateImpl implements Candidate {
+    private static final Logger LOG = Logger.getLogger(CandidateImpl.class.getName());
+
+    private static final byte[] UNKNOWN_CANDIDATE_DATA = "<unknown>".getBytes(Charsets.UTF_8);
+
+    private static final Supplier<byte[]> IP_ADDRESS_DATA_SUPPLIER = new Supplier<byte[]>() {
+        @Override public byte[] get() {
+            try {
+                return InetAddress.getLocalHost().getHostAddress().getBytes();
+            } catch (UnknownHostException e) {
+                LOG.log(Level.WARNING, "Failed to determine local address!", e);
+                return UNKNOWN_CANDIDATE_DATA;
+            }
+        }
+    };
+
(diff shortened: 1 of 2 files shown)


src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
```diff
@@ -69,6 +70,8 @@
     private final Group group;
     private final Function<Iterable<String>, String> judge;
     private final Supplier<byte[]> dataSupplier;
+    private final AtomicReference<Group.GroupChangeListener> groupChangeListener;
+    private  Command cancelWatch;
```
gkleiman: nit: there's an extra space after `private`

src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
```diff
@@ -121,17 +126,38 @@ public CandidateImpl(
...
-        group.watch(new Group.GroupChangeListener() {
-            @Override public void onGroupChange(Iterable<String> memberIds) {
+        final AtomicReference<Group.Membership> membershipRef = new AtomicReference<Group.Membership>(null);
+        final AtomicReference<Iterable<String>> pendingChange = new AtomicReference<Iterable<String>>(null);
+        this.groupChangeListener.set(new Group.GroupChangeListener() {
+            @Override
+            public void onGroupChange(Iterable<String> memberIds) {
                 boolean noCandidates = Iterables.isEmpty(memberIds);
+                Group.Membership membership;
+                synchronized (this) {
```
gkleiman: Does this really need to be synchronized?

Quote from [ZK's doc](https://zookeeper.apache.org/doc/trunk/zookeeperProgrammers.html#Java+Binding):

> When a ZooKeeper object is created, two threads are created as well: an IO thread and an event thread. All IO happens on the IO thread (using Java NIO). All event callbacks happen on the event thread. Session maintenance such as reconnecting to ZooKeeper servers and maintaining heartbeat is done on the IO thread. Responses for synchronous methods are also processed in the IO thread. All responses to asynchronous methods and watch events are processed on the event thread. There are a few things to notice that result from this design:
>
> - All completions for asynchronous calls and watcher callbacks will be made in order, one at a time. The caller can do any processing they wish, but no other callbacks will be processed during that time.

As far as I could see, `onGroupChange` is called only when a watcher callback is triggered, but I might be missing something.

sttts: It*s not about the ZK objects. It's about synchronizing membershipRef.get and pendingChange.set. I want to avoid that the call to onGroupChange outside of the listener calls the handler with an old value of memberIds.

src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
```diff
@@ -149,8 +175,9 @@ public CandidateImpl(
                                 membership.getMemberPath(), memberIds));
 
                         leader.onElected(new ExceptionalCommand<Group.JoinException>() {
-                            @Override public void execute() throws Group.JoinException {
-                                membership.cancel();
+                            @Override
+                            public void execute() throws Group.JoinException {
+                                membershipRef.get().cancel(); // here we know that membershipRef != null
```
gkleiman: Can't [L199](https://github.com/mesosphere/marathon/pull/3732/commits/5c1d11e1d13e4646b2b4bfa48a8afa9e080ed19e?diff=split#diff-2b73a7298e29c2c5b15b8f7aae5827bfR199) be executed before this line?

For example if leader loses leadership and the abdication command is called at roughly the same time?

sttts: You are right. That's a race.

src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
```diff
@@ -166,6 +193,23 @@ public CandidateImpl(
...
 
+        // join the group
+        membershipRef.set(group.join(dataSupplier, new Command() {
+            @Override public void execute() {
+                membershipRef.set(null);
+                leader.onDefeated();
+            }
+        }));
+
+        // possibly the upper membershipRef.set is not finished yet when the groupChangeHandler
```
gkleiman: nit: s/groupChangeHandler/groupChangeListener/

src/main/java/mesosphere/marathon/core/leadership/CandidateImpl.java
gkleiman: Ohh, I see.

I think that then you need to put `synchronized (CandidateImpl.this)` here, otherwise the synchronized statements on lines 152 and 206 use different objects.

If `pendingChange` is only accessed from within synchronized blocks, does it have to be an `AtomicReference`?

sttts: `pendingChange` doesn't have to be atomic actually. But then it must be final to be usable in the closure. What's the idiomatic way to do that in Java?

gkleiman: You're right, then an atomic reference is fine, sorry for that.

gkleiman:
LGTM. Even though twitter/commons seems to be abandoned, it might be a good idea to open a pull request.

sttts:
@gkleiman yes, will make a PR in parallel.

sttts:
While trying to write a test, I noticed that actually the Zookeeper library only registers one watch per client path, but stores a list of them (compare https://github.com/apache/zookeeper/blob/trunk/src/java/main/org/apache/zookeeper/ZooKeeper.java#L528). I checked the Zookeeper server and its stats which confirm this theory.

In other words, the leak is purely Java internal. With each abdication one gets another watch callback call when the group changes.

Still, the solution in this PR stays valid. But I have no good idea how to observe the new, good behavior from outside of `CandiateImpl` without checking log output. Log output filtering is unavailable in the current setup.

Hence, I will not continue to implement an integration test for this PR.

gkleiman:
Twitter Commons' tests passed, I also verified that Marathon doesn't leak watchers by manually doing `while sleep 2; do http DELETE :8080/v2/leader; done` and looking at the logs.

gkleiman:
:+1: thanks a lot!

PR #3741, opened 2016-04-11
Add id's to the validation paths instead of index

aquamatthias:
Introduce everyWithId that adds the identifier of the object in the hierarchy instead of the index.
@kolloch Is that better?

Attention: this PR is based on your validation fix.


src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -66,6 +66,15 @@ case class PathId(path: List[String], absolute: Boolean = true) extends Ordered[
 
   def toHostname: String = path.reverse.mkString(".")
 
+  def relativePath: String = {
+    def in(currentPath: List[String]): String = currentPath match {
+      case Nil          => "/"
```
aquamatthias: `.`?

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
```diff
@@ -103,7 +103,7 @@ class ModelValidationTest
       case f: Failure =>
         val errors = (Json.toJson(f) \ "details").as[Seq[JsObject]]
         errors should have size 1
-        (errors.head \ "path").as[String] should be("/groups(0)/apps(1)")
+        (errors.head \ "path").as[String] should be("/groups(group1)/apps(invalid)")
```
kolloch: I like it but we have to include the UI guys in this decision.

What do you think? This is the path in the JSON error response. We face the problem that the indices are unreliable because they relate to unordered sets and not the order specified in the JSON.

@Poltergeist @pierlo-upitup

pierlo-upitup: Thanks for the heads up @kolloch . Yeah,  the UI currently expects an index in the form of `(N)`, for example https://github.com/mesosphere/marathon-ui/blob/master/src/js/stores/AppFormStore.js#L135 . What kind of string values are you proposing here, @aquamatthias ? 
What do `groupN`, `invalid` etc. relate to?

aquamatthias: @pierlo-upitup If we have a sequence of things, than we usually reference this item by index like `apps(0)`. We have some logic that does not work very well with the index e.g. the apps of a group is currently a `Set` so the order is not maintained. Using an index on such structure does not help. Thats why the proposal is to identify this element by id. The proposal you see here is a relative id: `/groups(group1)/apps(invalid)` means that in group `group1` there is an app `invalid`. The full path of the app would be `/group1/invalid`. We could also use the full id instead of the relative one.

kolloch: @pierlo-upitup would that be good for you?

src/main/scala/mesosphere/marathon/state/PathId.scala
kolloch: well, it's more consistent to what we have right now...

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
pierlo-upitup: I think the full path would be a lot more informative than the relative one. 

In the example above, I'd expect to receive an error pointing me to `/group1/invalid` so that I can understand immediately what the error refers to. 

@aquamatthias will this ship in a 1.2 version? If so we'll have to schedule some time to adjust the UI.

aquamatthias: We can make the path absolute.
Yes it will not ship in 1.1

aquamatthias: I still wonder if the syntax is not confusing: `/groups(/nested)/apps(/nested/app1)/externalVolumes(0)`

kolloch: /cc @pierlo-upitup @aquamatthias Absolute paths in the parenthesis seem redundant and hard to read that way. That said if you have a validation error for a single app, they give you context:

`/app(/nested/group/app1)/cmd`

instead of

`/app(app1)/cmd`

aquamatthias:
Close for now.

PR #3715, opened 2016-04-08
Documentation for external volumes

jdef:
supersedes #3607

/cc @sascala

PR #3749, opened 2016-04-12
ext vol docs: s/external.drivers/external.options/

jdef:
xref https://github.com/mesosphere/marathon/pull/3715/files#r59402203


aquamatthias:
Thanks.

PR #3737, opened 2016-04-11
Fixes #3723 - Fix validation of duplicate volume names

kolloch:

Issue #3723, opened 2016-04-08
external volumes: allows user to create multiple zero-instance apps that ref the same volume, and then scale them both to 1


jdef:
* It also lets me create an app (instances = 1) that points to the same ext vol while another, running app is already referencing that external volume.

the group-validation isn't working

kolloch:
Actually, it doesn't depend on instances being 0.

```http
PUT /v2/groups/ext HTTP/1.1
Accept: application/json
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 1171
Content-Type: application/json
Host: localhost:8080
User-Agent: HTTPie/0.9.3

{
    "apps": [
        {
            "cmd": "sleep 3000",
            "container": {
                "type": "MESOS",
                "volumes": [
                    {
                        "containerPath": "/docker_storage",
                        "external": {
                            "name": "my-external-volume",
                            "options": {
                                "dvdi/driver": "rexray",
                                "dvdi/iops": "1000"
                            },
                            "provider": "dvdi",
                            "size": 1234
                        },
                        "mode": "RW"
                    }
                ]
            },
            "id": "foo1",
            "instances": 1
        },
        {
            "cmd": "sleep 3000",
            "container": {
                "type": "MESOS",
                "volumes": [
                    {
                        "containerPath": "/docker_storage",
                        "external": {
                            "name": "my-external-volume",
                            "options": {
                                "dvdi/driver": "rexray",
                                "dvdi/iops": "1000"
                            },
                            "provider": "dvdi",
                            "size": 1234
                        },
                        "mode": "RW"
                    }
                ]
            },
            "id": "foo2",
            "instances": 1
        }
    ]
}

HTTP/1.1 200 OK
Cache-Control: no-cache, no-store, must-revalidate
Content-Type: application/json; qs=2
Date: Mon, 11 Apr 2016 10:22:58 GMT
Expires: 0
Pragma: no-cache
Server: Jetty(9.3.6.v20151106)
Transfer-Encoding: chunked
X-Marathon-Leader: http://macptr.local:8080

{
    "deploymentId": "00e32fcb-90e7-4f19-b564-406f80cc6a51",
    "version": "2016-04-11T10:22:58.125Z"
}
```

kolloch:
The same request with the new code results in to following error messages:

```http
PUT /v2/groups/ext HTTP/1.1
Accept: application/json
Content-Type: application/json
Host: localhost:8080

{
    "apps": [
        {
[...]

Issue #3563, opened 2016-03-23
Resident Tasks: Flaky test: persistent volume will be re-attached and keep state


gkleiman:
```
[14:24:16][Step 3/3] - persistent volume will be re-attached and keep state *** FAILED *** (30 seconds, 378 milliseconds)
[14:24:16][Step 3/3]   java.lang.AssertionError: Waiting for event deployment_success to arrive took longer than 30 seconds. Give up.
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.next$1(WaitTestSupport.scala:30)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.waitFor(WaitTestSupport.scala:36)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.setup.MarathonCallbackTestSupport$class.waitForEventMatching(MarathonCallbackTestSupport.scala:48)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.ResidentTaskIntegrationTest.waitForEventMatching(ResidentTaskIntegrationTest.scala:16)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.setup.MarathonCallbackTestSupport$class.waitForEventWith(MarathonCallbackTestSupport.scala:52)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.ResidentTaskIntegrationTest.waitForEventWith(ResidentTaskIntegrationTest.scala:16)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.setup.MarathonCallbackTestSupport$class.waitForEvent(MarathonCallbackTestSupport.scala:32)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.ResidentTaskIntegrationTest.waitForEvent(ResidentTaskIntegrationTest.scala:16)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.ResidentTaskIntegrationTest$$anonfun$3.apply(ResidentTaskIntegrationTest.scala:56)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.ResidentTaskIntegrationTest$$anonfun$3.apply(ResidentTaskIntegrationTest.scala:44)
[14:24:16][Step 3/3]   at mesosphere.marathon.integration.ResidentTaskIntegrationTest$$anonfun$test$1.apply$mcV$sp(ResidentTaskIntegrationTest.scala:225)
[...]

PR #3751, opened 2016-04-12
take app instances into account when detecting volume conflicts

jdef:
xref https://github.com/mesosphere/marathon/pull/3737/files#r59401585
xref #3723

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
+++ b/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
@@ -117,24 +117,26 @@ private[impl] object DVDIProviderValidations extends ExternalVolumeValidations {
     override def apply(g: Group): Result = {
       val appsByVolume: Map[String, Set[PathId]] =
         g.transitiveApps
+          .filter { app => app.instances > 0 }
           .flatMap { app => namesOfMatchingVolumes(app).map(_ -> app.id) }
           .groupBy { case (volumeName, _) => volumeName }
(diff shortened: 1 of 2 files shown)


jdef:
xref #3563

sttts:
If you scale up the suspended app again, tasks will fail. Or do we get a proper error message then from the scaling API endpoint?

kolloch:
What is the use case of this? You have multiple app definitions that access the same volume and sometimes you suspend A and start B, sometimes you suspend B and start A?

I would not allow this for simplicity. You can still destroy A and create B or the other way around.

There are also race-conditions involving this that are more likely to occur if you encourage this kind of pattern. Removing an app does not remove all tasks instantaneously, therefore temporarily there could exist tasks accessing the same volume.

jdef:
@kolloch I don't think I had a specific use case in mind, it just seemed intuitive that I'd be able to define multiple apps that reference the same vol as long as the number of task instances that reference the same volume is limited to 1.

Yes, I can see how this could be a bit racy. In a race we'd depend on the underlying provider to generate an error on task-launch if multiple running tasks referenced the same volume instance. Since tasks are respawned indefinitely in Marathon the system should eventually settle into a stable state since the "desired end state" only allows 1 task instance per volume.

@sttts no, the error messages sent back from the isolator are not very informative; as we say in the docs users are basically expected to grep the agent/system logs to troubleshoot.

sttts:
@jdef I meant the error message from the Marathon REST API when scaling a 0-instance app to 1 instance.

aquamatthias:
I would also recommend to make sure to check the App<-> Volume relationship, even if there are currently no running tasks. 
Close this for now, if we don't see any specific use case for this.

PR #3753, opened 2016-04-12
Add clarifying statement about portMappings

ssk2:
Added a line to hopefully make it clearer that these are used instead of the`ports` or `portDefinitions` array.

### DIFF
--- a/docs/docs/ports.md
+++ b/docs/docs/ports.md
@@ -16,11 +16,11 @@ Port configuration for applications in Marathon can be confusing and there is [a
 
 *HOST networking*: `HOST` networking is used by non-Docker Marathon applications and Docker applications that use `HOST` mode networking. In this mode, applications bind directly to one or more ports on the host machine.
 
-*portMapping*: A _port mapping_ is used to define ports that should be considered as part of a resource offer. It is necessary for Docker applications that use `BRIDGE` mode networking and is a tuple containing a host port, container port, service port and protocol. Multiple _port mappings_ may be specified for a Marathon application.
+*portMapping*: In Docker `BRIDGE` mode, a port mapping is necessary for every port that should be reachable from outside of your container. A port mapping is a tuple containing a host port, container port, service port and protocol. Multiple _port mappings_ may be specified for a Marathon application.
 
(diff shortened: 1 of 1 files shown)


docs/docs/ports.md
```diff
@@ -16,7 +16,7 @@ Port configuration for applications in Marathon can be confusing and there is [a
 
 *HOST networking*: `HOST` networking is used by non-Docker Marathon applications and Docker applications that use `HOST` mode networking. In this mode, applications bind directly to one or more ports on the host machine.
 
-*portMapping*: A _port mapping_ is necessary for Docker applications that use `BRIDGE` mode networking and is a tuple containing a host port, container port, service port and protocol. Multiple _port mappings_ may be specified for a Marathon application.
+*portMapping*: A _port mapping_ is used to define ports that should be considered as part of a resource offer. It is necessary for Docker applications that use `BRIDGE` mode networking and is a tuple containing a host port, container port, service port and protocol. Multiple _port mappings_ may be specified for a Marathon application.
```
sttts: Even clearer:

A _port mapping_ is used to define ports that should be considered as part of a resource offer in `BRIDGE` mode. It is a tuple...
.
.
.
*ports*: The _ports_ array is used to define ports that should be considered as part of a resource offer in `HOST` mode....

I think most of the confusion comes from the use of the word `necessary`, sounds like `optional`.

kolloch: If you declare a port mapping, Marathon will make sure that there is also a port on the host interface that forwards to it. I am not sure if the mentioning of the "resource offer" is helpful.

Maybe

In Docker `BRIDGE` mode, a _port mapping_ is necessary for every port that should be reachable from outside of your container.

ssk2: @kolloch, @sttts: updated, thanks!

sttts: @ssk2 sorry for being picky, but the logic is wrong: it must be "... is used to define ports ... in BRIGDE mode". That's different from "... is used to define ports ... <FULLSTOP>. It is necessary ... for BRIDGE mode".

meichstedt: I agree with @sttts here. Seems like a matrix might be more helpful here? ;)

aquamatthias:
@ssk2 Thanks.

PR #3755, opened 2016-04-13
Introduce feature_flag external_volumes.

aquamatthias:
-Make sure this feature is set during validation of external volumes
-Remove the SuppliedOptionNames from AllConf
-Introduce the option to override the config for test purposes

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -62,6 +62,12 @@ object Validation {
     }
   }
 
+  def featureEnabled[T](feature: String): Validator[T] = {
+    isTrue(s"Feature $feature is not enabled. Enable with --enable_features $feature)") { _ =>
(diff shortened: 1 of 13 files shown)


src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -62,6 +62,12 @@ object Validation {
     }
   }
 
+  def featureEnabled[T](feature: String): Validator[T] = {
+    isTrue(s"Feature $feature is not enabled. Enable with --enable_features $feature)") { _ =>
```
kolloch: 👍 for the hint

kolloch:
Some comments:

* Missing documentation
* You only prevent creating/changing apps with external volumes -- not using them. That said, any change of app definition would be prevented until you removed apps with external volumes. What do thing of a rootGroup validation on startup that at least logs an ERROR if there are any problems?
* Ignore me: I am not a fan of global state but hey, at least it does not change frequently.

PR #3758, opened 2016-04-13
Fix offersWantedSubjects

kolloch:
Make sure that the offersWantedSubjects always submit a
value on subscription. Otherwise, combineLatest in the
CoreModuleImpl will potentially subscribe after the
offerReconcilerActor emitted its signal.


kolloch:
Related: https://mesosphere.atlassian.net/browse/MARATHON-926

aquamatthias:
Awesome! Thanks!

PR #3759, opened 2016-04-13
Update License Copyright

kolloch:


aquamatthias:
I am not a license expert: does this mean we defined this license in 2013?
If we say 2013-2016: what happens in 2017?

kolloch:
I am not a license expert either but I have seen people updating their copyrights. You cannot just extend the copyright into the future, I guess.

/cc @florianleibert

nklovejoy:
The date should reflect the creation year.  Really, though, you could put 2041 and it would have no practical legal significance in the US or EU.

aquamatthias:
Than I will just leave it as is `2013`

PR #3777, opened 2016-04-14
Factor out leader election code into core module and add curator support

sttts:
- create core module "election"
- move all generic leadership logic from `MarathonSchedulerService` into `ElectionService` implementation
- add curator support which is active by default. A "always wins the election" tombstone is created in the twitter common `/leader` group such that old Marathon seamlessly agree on the new master of the new election code
- add a burn-in integration test which re-elects 10 times with ~~half a seconds~~ a random value between 0 to 1000ms in between

TODO:
- [x] ~~deduplicate/reuse ExponentialBackoff logic~~ – haven't found anything in the code base
- [x] document traits
- [x] ~~analyze and possibly fix burn-in tests with `Thread.sleep(500L)` (probably they fail without the sleep because the `DELETE` call on the leader endpoints in the tests fails)~~ randomized the sleep now, with a 10% chance that the wait is actually zero. This should make us aware of any flakyness
- [x] decide whether we switch on the curator leader election immediately or avoid two big changes at once: the refactoring + the curator lib switch.


src/main/scala/mesosphere/marathon/core/election/ElectionServiceBase.scala
```diff
@@ -0,0 +1,115 @@
...
+  protected def offerLeadershipImpl(): Unit
+
+  override def offerLeadership(): Unit = synchronized {
+    log.info(s"Will offer leadership after ${backoff.value()} backoff")
+    after(backoff.value(), system.scheduler)(Future {
+      offerLeadershipImpl()
+    })
+  }
+
+  protected def stopLeadershop(): Unit = synchronized {
```
aquamatthias: rename to `stopLeadership`

src/main/scala/mesosphere/marathon/core/election/ElectionServiceBase.scala
```diff
@@ -0,0 +1,115 @@
+package mesosphere.marathon.core.election
```
aquamatthias: Ca we move this class to `impl`

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -53,7 +63,31 @@ class CoreModuleImpl @Inject() (
   private[this] lazy val shutdownHookModule = ShutdownHooks()
   private[this] lazy val actorsModule = new ActorsModule(shutdownHookModule, actorSystem)
 
-  override lazy val leadershipModule = LeadershipModule(actorsModule.actorRefFactory, zk, leader)
+  override lazy val leadershipModule = LeadershipModule(actorsModule.actorRefFactory, zk, electionModule)
+  override lazy val electionModule = if (marathonConf.highlyAvailable()) {
```
aquamatthias: Can we move this logic into ElectionModule?

src/main/scala/mesosphere/marathon/core/election/ExponentialBackoff.scala
```diff
@@ -0,0 +1,27 @@
+package mesosphere.marathon.core.election
+
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.duration._
+
+private class ExponentialBackoff(
```
aquamatthias: We have this functionality elsewhere (e.g. RateLimiter).
Perhaps we can reuse logic?

src/main/scala/mesosphere/marathon/core/election/PseudoElectionService.scala
```diff
@@ -0,0 +1,30 @@
+package mesosphere.marathon.core.election
```
aquamatthias: move to impl

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -53,7 +63,31 @@ class CoreModuleImpl @Inject() (
...
+    new TwitterCommonsElectionService(
```
aquamatthias: where do we put the logic for choosing either twitter commons or curator?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -0,0 +1,30 @@
...
+  def abdicateLeadership(error: Boolean = false): Unit
+}
+
+trait ElectionService extends LeadershipAbdication {
+  def isLeader: Boolean
+  def leader: Option[String]
+  def offerLeadership(): Unit
+}
+
+trait ElectionDelegate {
```
aquamatthias: Just looking into trait definitions, it is not clear what an ElectionDelegate is.
1) Perhaps we need a better name?
2) Please add some class level documentation.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -0,0 +1,30 @@
+package mesosphere.marathon.core.election
+
+import scala.concurrent.Future
+
+trait LeadershipAbdication {
```
aquamatthias: I like small traits, but perhaps this method could just move into ElectionService?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -0,0 +1,30 @@
...
+  def abdicateLeadership(error: Boolean = false): Unit
+}
+
+trait ElectionService extends LeadershipAbdication {
```
aquamatthias: Please add some lines of class level documentation.

src/main/scala/mesosphere/marathon/core/election/TwitterCommonsElectionService.scala
```diff
@@ -0,0 +1,78 @@
+package mesosphere.marathon.core.election
```
aquamatthias: move to impl.

aquamatthias:
@sttts did a first pass. I think this is the right direction. Added a couple of comments.

src/main/scala/mesosphere/marathon/core/election/ExponentialBackoff.scala
sttts: That was the plan.

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
sttts: I guess the ElectionModule is a good place for that. In my prototype I have a wrapper around the TwitterCommon and the Curator class which switches over and passes through the API.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
sttts: Historical. Maybe we can merge them, yes.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
sttts: The delegate is similar to a callback, kind of the primary callback. Not sure yet whether we are able to merge these two.

sttts:
PTAL @gkleiman

gkleiman:
@sttts I haven't started the review yet, but `mesosphere.marathon.integration.LeaderIntegrationTest` failed, it might be worth taking a look into that.

sttts:
@gkleiman I am on that now

src/main/scala/mesosphere/marathon/MarathonLeaderInfo.scala
```diff
@@ -1,43 +1,32 @@
...
-          None
-      }
-      maybeLeaderData.map { data =>
-        new String(data, "UTF-8")
-      }
+    try {
+      electionService.leaderHostPort
+    }
+    catch {
+      case e: java.lang.Exception => None
```
gkleiman: Shouldn't we catch only NonFatal exceptions?

I think it would also be nice to keep the previous behaviour and log something.

sttts: Resolved.

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
```diff
@@ -53,6 +58,110 @@ class LeaderIntegrationTest extends IntegrationFunSuite
     WaitTestSupport.waitUntil("the leader changes", 30.seconds) { marathon.leader().value != leader }
   }
 
+  test("it survives a small burn-in reelection test") {
```
gkleiman: I like the idea of a burn-in reelection test. What about also checking that once an election happens, all all nodes return the same leader? There's an already existing test that does that, so you could re-use it.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -250,140 +171,72 @@ class MarathonSchedulerService @Inject() (
...
-  //Begin Leader interface, which is required for CandidateImpl.
-  override def onDefeated(): Unit = synchronized {
-    log.info("Defeated (Leader Interface)")
+  //Begin ElectionCandidate interface
 
-    log.info(s"Call onDefeated leadership callbacks on ${leadershipCallbacks.mkString(", ")}")
-    Await.result(Future.sequence(leadershipCallbacks.map(_.onDefeated)), zkTimeout)
-    log.info(s"Finished onDefeated leadership callbacks")
+  def startLeadership(): Unit = synchronized {
+    log.info("Elect leadership, running driver")
```
gkleiman: nit: maybe s/Elect leadership/Elected as leader/ ?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -0,0 +1,68 @@
...
+  */
+trait ElectionCandidate {
+  /**
+    * stopLeadership is called when the candidate was leader, but was defeated. It is guaranteed
+    * that before startLeadership has been called.
+    */
+  def stopLeadership(): Unit
+
+  /**
+    * startLeadership is called when the candidate has become leader. It is guaranteed that
```
gkleiman: Which is the guarantee? That `startLeadership` will always be called before `stopLeader`?

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
```diff
@@ -0,0 +1,209 @@
...
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.duration._
+import scala.concurrent.{ Await, Future }
+import scala.util.control.NonFatal
+
+private[impl] object ElectionServiceBase {
+  protected type Abdicator = /* error: */ Boolean => Unit
+
+  abstract class State {
```
gkleiman: nit: I think this could be a sealed abstract class.

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
```diff
@@ -0,0 +1,209 @@
...
+      case Leading(c, _)       => Some(c)
+      case Abdicating(c, _, _) => Some(c)
+      case Offering(c)         => Some(c)
+      case Offered(c)          => Some(c)
+    }
+  }
+
+  case class Idle(candidate: Option[ElectionCandidate]) extends State
+  case class Leading(candidate: ElectionCandidate, abdicate: Abdicator) extends State
+  case class Abdicating(candidate: ElectionCandidate, reoffer: Boolean, shortcut: Boolean = false) extends State
```
gkleiman: what does shortcut mean?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -0,0 +1,85 @@
...
+    * @param candidate is called back once elected or defeated
+    */
+  def offerLeadership(candidate: ElectionCandidate): Unit
+
+  /**
+    * abdicateLeadership is called to resign from leadership. If this instance is no leader, this
+    * call does nothing for reoffer=false. It will call offerLeadership for reoffer=true..
+    *
+    * @param error is true if the abdication is due to some error.
+    * @param reoffer is true if leadership should be offered again after abdiction
```
gkleiman: s/abdiction/abdication/

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
```diff
@@ -0,0 +1,254 @@
...
+      Await.result(Future.sequence(electionCallbacks.map(_.onDefeated)), config.zkTimeoutDuration)
+      log.info(s"Finished onDefeated leadership callbacks")
+
+      // Our leadership has been defeated. Tell the candidate and the world
+      candidate.stopLeadership()
+      eventStream.publish(LocalLeadershipEvent.Standby)
+      stopMetrics()
+    }
+
+    // call abdiction continuations
```
gkleiman: s/abdiction/abdication/

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
sttts: I don't like this word.

Fixed.

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
sttts: Fixed

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
sttts: fixed

sttts: The alternate, i.e. start, stop, start, stop. Not: start, start, stop, stop, start.

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
sttts: Replace with `candidateWasStarted`. Should be more clear.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
sttts: Fixed: `As new leader running the driver`

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
sttts: Fixed.

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+
+  protected def stopLeadership(): Unit = synchronized {
+    val (candidate, reoffer, candidateWasStarted) = state match {
+      case Leading(c, a)         => (c, false, false)
+      case Abdicating(c, ro, cws) => (c, ro, cws)
+      case Offered(c)            => (c, false, false)
+      case Offering(c)           => (c, false, false)
+      case Idle(c)               => (c.get, false, false)
```
gkleiman: You can use `case Idle(Some(c)) => ` to avoid the dreaded `.get`.

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
```diff
@@ -0,0 +1,258 @@
...
+  // scalastyle:on
+
+  protected def offerLeadershipImpl(): Unit
+
+  private def setOfferState(offeringCase: => Unit, idleCase: => Unit): Unit = synchronized {
+    state match {
+      case Abdicating(candidate, reoffer, candidateWasStarted) =>
+        log.error("Will reoffer leadership after abdicating")
+        state = Abdicating(candidate, reoffer = true, candidateWasStarted)
+      case Leading(candidate, abdicate) =>
```
gkleiman: I don't think we have a convention about this in our style guide, but you can use `_` if you won't use the values.

src/test/scala/mesosphere/marathon/integration/LeaderIntegrationTest.scala
sttts: Done.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -250,140 +188,84 @@ class MarathonSchedulerService @Inject() (
...
-  override def onElected(abdicateCmd: ExceptionalCommand[JoinException]): Unit = synchronized {
-    var driverHandlesAbdication = false
-    try {
-      log.info("Elected (Leader Interface)")
+    // execute tasks, only the leader is allowed to
+    migration.migrate()
 
-      //execute tasks, only the leader is allowed to
-      migration.migrate()
+    // start all leadership coordination actors
```
gkleiman: In the original version the actors are started after having called the callbacks. Did you change the order on purpose? If so, why?

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -250,140 +188,84 @@ class MarathonSchedulerService @Inject() (
...
 
-      //run all leadership callbacks
-      log.info(s"""Call onElected leadership callbacks on ${leadershipCallbacks.mkString(", ")}""")
-      Await.result(Future.sequence(leadershipCallbacks.map(_.onElected)), config.onElectedPrepareTimeout().millis)
-      log.info(s"Finished onElected leadership callbacks")
-
-      //start all leadership coordination actors
-      Await.result(leadershipCoordinator.prepareForStart(), config.maxActorStartupTime().milliseconds)
+    // start timers
+    schedulePeriodicOperations()
```
gkleiman: In the original version the actors are started after creating/starting the driver. Did you change the order on purpose? If so, why?

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -250,140 +188,84 @@ class MarathonSchedulerService @Inject() (
...
+        log.info(s"Driver future completed with result=$result.")
+        result match {
+          case Failure(t) => log.error("Exception while running driver", t)
+          case _          =>
+        }
 
-      // We successfully took over leadership. Time to reset backoff
-      resetOfferLeadershipBackOff()
+        // tell leader election that we step back, but want to be re-elected if isRunning is true.
+        electionService.abdicateLeadership(error = result.isFailure, reoffer = latch.getCount > 0)
```
gkleiman: Wouldn't `isRunning` or `isRunningLatch` be a better name for the latch?

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -66,19 +53,18 @@ trait LeadershipAbdication {
...
   frameworkIdUtil: FrameworkIdUtil,
-  @Named(ModuleNames.LEADER_ATOMIC_BOOLEAN) leader: AtomicBoolean,
+  electionService: ElectionService,
+  prePostDriverCallbacks: Seq[PrePostDriverCallback],
   appRepository: AppRepository,
   driverFactory: SchedulerDriverFactory,
   system: ActorSystem,
   migration: Migration,
   @Named("schedulerActor") schedulerActor: ActorRef,
   @Named(EventModule.busName) eventStream: EventStream,
```
gkleiman: The `MarathonSechedulerService` doesn't use the eventbus anymore , so we can get rid of this dependency.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -25,7 +26,7 @@ import scala.util.control.NonFatal
   */
 trait LeaderInfo {
```
gkleiman: This trait is not used after this PR, can we get rid of it?

src/main/scala/mesosphere/marathon/core/election/ElectionModule.scala
```diff
@@ -0,0 +1,64 @@
...
+    config: MarathonConf,
+    system: ActorSystem,
+    eventStream: EventStream,
+    http: HttpConf,
+    metrics: Metrics = new Metrics(new MetricRegistry),
+    hostPort: String,
+    shutdownHooks: ShutdownHooks) {
+  private lazy val backoff = new ExponentialBackoff(name = "offerLeadership")
+  lazy val service: ElectionService = if (config.highlyAvailable()) {
+    if (config.isFeatureSet(Features.TWITTER_COMMONS)) {
```
gkleiman: I'd add a new `--election_service <curator/twitter_commons>` flag and make it hidden.

We don't want to have to support both implementations in the future.

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
```diff
@@ -0,0 +1,274 @@
...
+
+import scala.concurrent.duration._
+
+object ElectionServiceBaseTest {
+  import Mockito.mock
+
+  val MaxActorStartupTime = 5000L
+  val OnElectedPrepareTimeout = 3 * 60 * 1000L
+
+  def mockConfig: MarathonConf = {
```
gkleiman: Any reasons for not using `MarathonTestHelper.defaultConfig`?

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
```diff
@@ -0,0 +1,274 @@
...
+  private[this] var config: MarathonConf = _
+  private[this] var httpConfig: HttpConf = _
+  private[this] var electionService: ElectionService = _
+  private[this] var events: EventStream = _
+  private[this] var candidate: ElectionCandidate = _
+  private[this] var metrics: Metrics = _
+  private[this] var backoff: Backoff = _
+  private[this] var shutdownHooks: ShutdownHooks = _
+
+  before {
```
gkleiman: We normally put all of this in a Fixture class instead of using before/after.

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
```diff
@@ -0,0 +1,274 @@
...
+
+  test("state is Idle initially") {
+    val electionService = new ElectionServiceBase(
+      config, system, events, metrics, backoff, shutdownHooks
+    ) {
+      override protected def offerLeadershipImpl(): Unit = ???
+      override def leaderHostPortImpl: Option[String] = ???
+    }
+
+    awaitAssert(electionService.state should equal(Idle(candidate = None)))
```
gkleiman: can't this be a normal `should` statement without `awaitAssert`?

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
```diff
@@ -0,0 +1,195 @@
...
+  config, system, eventStream, metrics, backoff, shutdownHooks
+) {
+  private lazy val log = LoggerFactory.getLogger(getClass.getName)
+
+  private lazy val client = provideCuratorClient()
+  private var latch: Option[LeaderLatch] = None
+
+  override def leaderHostPortImpl: Option[String] = synchronized {
+    try {
+      latch match {
```
gkleiman: This is not very idiomatic, you can do:

```scala
maybeLatch.map { latch =>
  val participant = latch.getLeader
  if (participant.isLeader) Some(participant.getId) else None
}
```

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
```diff
@@ -0,0 +1,195 @@
...
+        log.error("error while getting current leader", e)
+        None
+    }
+  }
+
+  override def offerLeadershipImpl(): Unit = synchronized {
+    log.info("Using HA and therefore offering leadership")
+    latch match {
+      case Some(l) =>
+        log.error("Offering leadership while being candidate")
```
gkleiman: This looks more like a warning than an error.

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
```diff
@@ -0,0 +1,195 @@
...
+        l.close()
+      case _ =>
+    }
+    latch = Some(new LeaderLatch(
+      client, config.zooKeeperLeaderPath + "-curator", hostPort, LeaderLatch.CloseMode.NOTIFY_LEADER
+    ))
+    latch.get.addListener(Listener) // idem-potent
```
gkleiman: `maybeLatch.foreach(_.addListener(Listener))` is a bit more idiomatic.

We set latch to a new value in the line before, so the comment doesn't seem to be necessary.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
sttts: done

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
sttts: Actually, it was called after creating the driver, not running it (the run call is asynchronously). But to avoid regressions, I have moved the `schedulePeriodicOperations` call behind the driver creation again.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
sttts: fixed.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
sttts: Renamed to `isRunningLatch`.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
sttts: fixed

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
sttts: `flatMap` actually. Fixed.

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
sttts: fixed

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
sttts: I prefer `get` because "fail early" is good.

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
sttts: That code is stolen from `MarathonSchedulerServiceTest`.

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
sttts: Also stolen from `MarathonSchedulerServiceTest`.

sttts: fixed

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
sttts: fixed

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
sttts: Yes, for Idle it can be `should`. Fixed.

src/test/scala/mesosphere/marathon/core/election/impl/ElectionServiceBaseTest.scala
```diff
@@ -0,0 +1,269 @@
...
+    val config: MarathonConf = mockConfig
+    val httpConfig: HttpConf = mock[HttpConf]
+    val electionService: ElectionService = mock[ElectionService]
+    val events: EventStream = new EventStream()
+    val candidate: ElectionCandidate = mock[ElectionCandidate]
+    val metrics: Metrics = new Metrics(new MetricRegistry)
+    val backoff: Backoff = new ExponentialBackoff(0.01.seconds, 0.1.seconds)
+    val shutdownHooks: ShutdownHooks = mock[ShutdownHooks]
+
+    def mockConfig: MarathonConf = {
```
gkleiman: I understand that you took this from already existing tests, but we're trying to move to a new cleaner style, and since this is a new test suite, there's no reason for adopting the old one.

There are now two helper methods to create configuration objects in tests: `MarathonTestHelper.makeConfig` and `MarathonTestHelper.defaultConfig`. You can see example usages here: https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/MarathonConfTest.scala

sttts: `MarathonTestHelper.makeConfig` is really much nicer 👍  Fixed.

src/main/scala/mesosphere/marathon/core/election/impl/ElectionServiceBase.scala
sttts: fixed.

PR #3785, opened 2016-04-15
Remove dependency of HealthCheckManager.add to the AppRepository

sttts:
This fixes a race of the HelathCheckManager expecting a given AppDefinition in the
AppRepository which is not always true (yet) when an app is launched.

This PR removes the dependency by passing in the known AppDefinition from addForAll,
not only the id and version.

### DIFF
--- a/src/main/scala/mesosphere/marathon/health/HealthCheckManager.scala
+++ b/src/main/scala/mesosphere/marathon/health/HealthCheckManager.scala
@@ -19,9 +19,9 @@ trait HealthCheckManager {
   def list(appId: PathId): Set[HealthCheck]
 
   /**
-    * Adds a health check for the app with the supplied id.
+    * Adds a health check of the supplied app.
     */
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/health/HealthCheckManager.scala
```diff
@@ -19,9 +19,9 @@ trait HealthCheckManager {
   def list(appId: PathId): Set[HealthCheck]
 
   /**
-    * Adds a health check for the app with the supplied id.
+    * Adds a health check of a the given app.
```
sttts: -a

aquamatthias:
Thanks! Looks good. Waiting for the tests to pass.

sttts:
@aquamatthias let me add a test case before merging.

sttts:
Added test case and squashed.

PR #3799, opened 2016-04-18
Potential task count fixes

kolloch:


src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -331,12 +336,16 @@ private class AppTaskLauncherActor(
...
   private[this] def replyWithQueuedTaskCount(): Unit = {
+    val taskLaunchesInFlight =
+      inFlightTaskOperations
+        .keys
+        .count(taskId => tasksMap.get(taskId).exists(_.launched.isDefined))
     sender() ! QueuedTaskInfo(
       app,
       tasksLeftToLaunch = tasksToLaunch,
-      taskLaunchesInFlight = inFlightTaskOperations.size,
+      taskLaunchesInFlight = taskLaunchesInFlight,
```
aquamatthias: We need to add taskReservationsInFlight.
QueuedTaskInfo.inProgress should be adapted accordingly.

kolloch: Make also sure that you are not scanning all tasks each time your calculate this number.

Issue #3261, opened 2016-02-18
http_event_callback_slow_consumer_timeout does not influence the HTTP timeout


flosell:
In recent releases, `http_event_callback_slow_consumer_timeout` was added to configure a timeout when marathon should treat event callbacks as responding too slow (#3047). 

It seems like this doesn't affect the actual HTTP timeout but instead uses the hardcoded 10sec timeout from `HTTPEventModule`: https://github.com/mesosphere/marathon/blob/000a1cc66ab9ad96fb686e7a1350e652f205c85c/src/main/scala/mesosphere/marathon/event/http/HttpEventActor.scala#L63

We are currently dealing with an event callback that can take more than 10sec to respond but increasing the `http_event_callback_slow_consumer_timeout` doesn't have any effect, leading to lost events. 

Shouldn't marathon just use the configured timeout for event callbacks?

philipnrmn:
I wonder whether we missed this TODO:
https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/event/http/HttpEventModule.scala#L94-L95

tmmi:
Maybe https://github.com/mesosphere/marathon/pull/3805 ?

meichstedt:
@flosell I'd vote for a different timeout because I'd actually expect the consumerTimeout to be lower than the requestTimeout (which is currently not configurable), otherwise the logic makes not much sense.

tmmi:
Ok, requestTimeout included in the new PR https://github.com/mesosphere/marathon/pull/3827

meichstedt:
Resolved in #3827

PR #3805, opened 2016-04-19
#3261 | http_event_callback_slow_consumer_timeout influence the HTTP …

tmmi:
…timeout


meichstedt:
Hey @tmmi, these two timeouts are related but not the same. You definitely want to configure the request timeout separately from the `http_event_callback_slow_consumer_timeout` in terms of, e.g.
requestTimeout: 30s
consumerTimeout: 10s

That way, the consumer would be considered _slow_ if requests take longer than 10s to complete, but the actual request timeout would allow for requests to take up to 30s.

PR #3808, opened 2016-04-19
Fixes #3806 - Do not silently fail on reservations

### LINKED ISSUE #3806: Not setting the Mesos principal leads to silently failed reservations
If you try to use local volumes without setting a principal, it will not work without even logging errors. The reservation will be created and ignored by Mesos. On the Marathon side, the reservation will be timed out and a new reservation will be attempted.

/cc @aquamatthias

kolloch:
if Mesos principal is not given.


aquamatthias:
Thanks!

PR #3823, opened 2016-04-24
escape underscore. Otherwise jekyl will do a <em> around it

fatz:
https://mesosphere.github.io/marathon/docs/command-line-flags.html

it seems like the markdown interpreter of jekyll is doing this differently as github. Not sure if this is really wrong. 

```html
<li>"task<em>killing" can be used to enable the TASK</em>KILLING state in Mesos (0.28 or later)</li>
```

Hower if the normal text flows underscores are escaped we get the expected result.


aquamatthias:
Thanks!

PR #3828, opened 2016-04-25
maxConsecutiveFailures==0 will not kill only for http and tcp.

aquamatthias:


gkleiman:
https://github.com/apache/mesos/blob/master/src/health-check/main.cpp#L106 confirms this.

It doesn't feel nice that Marathon and Mesos are not consistent. Do you think we should update Marathon or file a Mesos JIRA?

PR #3833, opened 2016-04-25
Makes headers case-insensitive again

rhruiz:
The Jersey thread restriction circumvention changes the behavior on Servlets, making headers case sensitive. This enforces lower casing when mapping the original headers and accessing them.


gkleiman:
Thanks!

PR #3837, opened 2016-04-26
Increment tasksToLaunch only for TaskOp.Launch rejections

aquamatthias:
Part 1: only the fix for task launch increments. Will make the change to the QueuedTaskInfo as a separate PR.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
@@ -231,9 +231,15 @@ private class AppTaskLauncherActor(
   private[this] def receiveTaskLaunchNotification: Receive = {
     case TaskOpSourceDelegate.TaskOpRejected(op, reason) if inFlight(op) =>
       removeTask(op.taskId)
-      tasksToLaunch += 1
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -231,9 +231,15 @@ private class AppTaskLauncherActor(
   private[this] def receiveTaskLaunchNotification: Receive = {
     case TaskOpSourceDelegate.TaskOpRejected(op, reason) if inFlight(op) =>
       removeTask(op.taskId)
-      tasksToLaunch += 1
       log.info("Task op '{}' for {} was REJECTED, reason '{}', rescheduling. {}",
         op.getClass.getSimpleName, op.taskId, reason, status)
+
+      op match {
```
meichstedt: While testing this I realized that the above line `removeTask(op.taskId)` is also wrong in case the op is a TaskOp.Launch – only the launch was rejected, the Task.Reserved should stay in the tasksMap …

meichstedt:
Hey Matthias, LGTM so far regarding the actual change. As commented, there's another flaw while handling the `TaskOpRejected`. Can you have a look please?

PR #3840, opened 2016-04-28
[WIP] AppDefinition -> RunSpec

aquamatthias:
Proposal to use a trait in the core launcher module instead of an AppDefinition:
- RunSpec from AppDefinition created
- All occurrences of AppDefinition replaced by RunSpec (1. commit)
- All properties names appXXX renamed to runSpecXXX (3.commit)


aquamatthias:
@gkleiman @meichstedt proposal for AppDefinition -> RunSpec

PR #3841, opened 2016-04-28
Add --leader_proxy_ssl_ignore_hostname option

sttts:
When proxying by default the certificate CN attribute is matched against the leader hostname. If they differ, the proxying fails. The `--leader_proxy_ssl_ignore_hostname` disables this check.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
+++ b/src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
@@ -2,8 +2,9 @@ package mesosphere.marathon.api
 
 import java.io.{ IOException, InputStream, OutputStream }
 import java.net._
+import java.security.cert.X509Certificate
 import javax.inject.Named
-import javax.net.ssl.{ HttpsURLConnection, SSLContext }
+import javax.net.ssl._
 import javax.servlet._
 import javax.servlet.http.{ HttpServletRequest, HttpServletResponse }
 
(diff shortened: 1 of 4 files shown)


jdef:
lgtm

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -186,6 +187,10 @@ class JavaUrlConnectionRequestForwarder @Inject() (
 
   private[this] val viaValue: String = s"1.1 $myHostPort"
 
+  private lazy val ignoreHostnameVerifier = new javax.net.ssl.HostnameVerifier {
+    def verify(hostname: String, sslSession: SSLSession): Boolean = true
```
jdef: is there a way to avoid naming parameters that are unused in the func?

jdef: should there be an `override` here?

jdef: is it worth tracking a set of the last N hostnames and logging a warning whenever we get a hostname that's not already in the set?

sttts: Nice idea. But out of scope for this PR I think.

sttts: AFAIK using _ in function declarations are not that common in Scala.

jdef:
:ship:

PR #3849, opened 2016-04-29
correct default for zk flag

air:
As per https://mesosphere.slack.com/archives/marathon-private/p1461961924000043


gkleiman:
Thanks!

PR #3852, opened 2016-05-02
Fixes #3848 - Size check for deployments

### LINKED ISSUE #3848: Size check for deployments
When a group deployment is initiated, the size of the znode created is `old group + deployment metadata + updated group`. This can become too large to transact with ZooKeeper if it exceeds approximately 500kb (the jute.maxbuffer defaults to 1mb). Can we put a line in the log in this case that can advise the user to adjust their app and group definition sizes?

gkleiman:
cc/ @jgarcia-mesosphere @mcclurmc

### DIFF
--- a/src/main/scala/mesosphere/marathon/ZookeeperConf.scala
+++ b/src/main/scala/mesosphere/marathon/ZookeeperConf.scala
@@ -51,6 +51,13 @@ trait ZookeeperConf extends ScallopConf {
     default = Some(64 * 1024)
   )
 
+  lazy val zooKeeperMaxNodeSize = opt[Long]("zk_max_node_size",
+    descr = "(Default: 1 MiB) Maximum allowed ZooKeeper node size (in bytes).",
+    noshort = true,
+    validate = _ >= 0,
+    default = Some(1024 * 1000)
+  )
+
   def zooKeeperStatePath: String = "%s/state".format(zkPath)
   def zooKeeperLeaderPath: String = "%s/leader".format(zkPath)
   def zooKeeperServerSetPath: String = "%s/apps".format(zkPath)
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/ZookeeperConf.scala
```diff
@@ -51,6 +51,13 @@ trait ZookeeperConf extends ScallopConf {
     default = Some(64 * 1024)
   )
 
+  lazy val zooKeeperMaxNodeSize = opt[Long]("zk_max_node_size",
+    descr = "(Default: 1 MiB) Maximum allowed ZooKeeper node size (in bytes).",
+    noshort = true,
+    validate = _ >= 0,
+    default = Some(1024 * 1000000)
```
aquamatthias: Shouldn't that be 1024*1000?

gkleiman: Yes, fixed.

src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala
```diff
@@ -305,7 +307,18 @@ object DeploymentPlan {
     result
   }
 
-  implicit lazy val deploymentPlanIsValid: Validator[DeploymentPlan] = validator[DeploymentPlan] { plan =>
-    plan.createdOrUpdatedApps as "app" is every(valid(AppDefinition.updateIsValid(plan.original)))
+  def deploymentPlanValidator(conf: MarathonConf): Validator[DeploymentPlan] = {
+    val maxSize = conf.zooKeeperMaxNodeSize()
+    val notBeTooBig = isTrue[DeploymentPlan](s"The deployment plan size can't be larger than $maxSize bytes.") { plan =>
```
aquamatthias: Perhaps something like:

>
The way we persist data in ZooKeeper would lead to a ZK node size greater than $maxSize.
You can adjust this value via `--zk_max_node_size`, but make sure this value is supported! 
See: http://zookeeper.apache.org/doc/r3.3.1/zookeeperAdmin.html#sc_configuration

aquamatthias:
Thanks! LGTM

mcclurmc:
Thanks @gkleiman! I'm really glad this was a feasible check to make.

aquamatthias:
Thanks. Need to cherry pick this.

PR #3855, opened 2016-05-03
Resolves #3851 allow dynamically reserved resources for normal task matching

### LINKED ISSUE #3851: marathon can not consume dynamic reserved resources
currently, Marathon can not use dynamic reserved resources:

1. https://github.com/mesosphere/marathon/blob/master/src%2Fmain%2Fscala%2Fmesosphere%2Fmesos%2FTaskBuilder.scala#L83

will we support this in the near future?

also, what is the `reserved` param means here, want to filter out `reserved resources`? if that's case, we may missuse `ReservationInfo`:

1. https://github.com/apache/mesos/blob/master/include%2Fmesos%2Fmesos.proto#L681-L684
2. https://github.com/mesosphere/marathon/blob/master/src%2Fmain%2Fscala%2Fmesosphere%2Fmesos%2FResourceMatcher.scala#L41-L43

meichstedt:
A normal (non-resident) task can now use dynamically reserved resources as long as these are not labeled with known labels which are used for resident tasks. That way, dynamically reserving resources for role foo and using those is possible like with static reservations.

### DIFF
--- a/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
+++ b/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
@@ -1,6 +1,6 @@
 package mesosphere.mesos
 
-import mesosphere.marathon.core.launcher.impl.ResourceLabels
+import mesosphere.marathon.core.launcher.impl.{ ReservationSelector, TaskLabels }
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.state.{ AppDefinition, ResourceRole }
 import mesosphere.marathon.tasks.{ PortsMatch, PortsMatcher }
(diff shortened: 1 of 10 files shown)


src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -38,17 +38,15 @@ object ResourceMatcher {
...
-    *                 are considered. If reserved is false, only resources without
-    *                 a ReservationInfo are considered.
-    * @param requiredLabels only resources with the given keys/values are matched.
+    * @param reservation if given, only resources with a ReservationInfo are
+    *                    considered and will only match if their labels match
+    *                    the specified labels.
     */
-  case class ResourceSelector(
-      acceptedRoles: Set[String], reserved: Boolean, requiredLabels: ResourceLabels = ResourceLabels.empty) {
+  case class ResourceSelector(acceptedRoles: Set[String], reservation: Option[ReservationSelector] = None) {
```
aquamatthias: Instead of having an optional ReservationSelector, we could implement 2 selectors:
1) that selects based on the existing labels
2) that selects based on non existing labels
But it is probably minor.

aquamatthias:
@meichstedt Thanks LGTM.

meichstedt:
Thanks @aquamatthias for the suggestion. I've refactored this a little – POI is the `object ResourceSelector` that now has functions
- `reservedWithLabels(acceptedRoles: Set[String], labels: Map[String, String])` which returns a ResourceSelector matching resources that are (dynamically) reserved for one of the given roles and have the given reservation labels
- `reservable` which returns a ResourceSelector that matches unreserved resources that do **not** have known (resident) reservation labels (they may have other labels)
- `any(acceptedRoles: Set[String])` which returns a ResourceSelector that matches either unreserved or reserved resources as long as they do **not** have known (resident) reservation labels

I hope the semantics are now clearer. It is still irritating that we treat static reservations (_roles_) and dynamic reservations differently, but that's another scope 😬

aquamatthias:
LGTM. Thanks!

PR #3856, opened 2016-05-03
WIP: add secrets

sttts:


jdef:
TODO: AppUpdate has an env map that needs to be updated

jdef:
TODO: unit tests for envvar secret proto ser/deser

src/main/proto/marathon.proto
```diff
@@ -316,4 +317,19 @@ message ResidencyDefinition {
 message Secret {
   required string id = 1;
   required string source = 2;
-}
\ No newline at end of file
+}
+
+message EnvVarReference {
+  enum Type {
+    UNKNOWN = 0;
```
sttts: Why this UNKNOWN?

src/main/proto/marathon.proto
```diff
@@ -103,6 +103,7 @@ message ServiceDefinition {
   repeated mesos.Port portDefinitions = 27;
   repeated ReadinessCheckDefinition readinessCheckDefinition = 28;
   repeated Secret secrets = 29; // since 1.2
+  repeated EnvVarReference envVarReferences = 30;
```
sttts: Maybe add: `// store env var reference separately as long as the Mesos CommandInfo cannot store these natively`

src/main/proto/marathon.proto
jdef: pattern adopted from the core team. useful when the enum changes in a backwards incompatible way; you want a sane default value for enum values that don't exist any more. in this case, the sane value is UNKNOWN.

src/main/proto/marathon.proto
jdef: good idea

src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueueModule.scala
```diff
@@ -43,6 +50,18 @@ class LaunchQueueModule(
 
   val launchQueue: LaunchQueue = new LaunchQueueDelegate(config, launchQueueActorRef, rateLimiterActor)
 
+  // TODO(jdef) copied this from auth.AuthModule; seems useful to have this as a reusable func
```
jdef: refactor this as PluginManager.onePlugin?

jdef: arguably we'd want to support multiple plugins here, so nevermind

src/main/scala/mesosphere/marathon/Features.scala
```diff
@@ -11,10 +11,14 @@ object Features {
   //enable external volumes
   val EXTERNAL_VOLUMES = "external_volumes"
 
+  //enable secrets
+  val SECRETS = "secrets"
```
jdef: should this (and the other feature name constants) also be declared as `lazy` to avoid init race conditions? seeing strange CI errors...

jdef:
strange errors in CI:
```
[info]   java.lang.NullPointerException:
[info]   at org.rogach.scallop.Scallop.verify(Scallop.scala:624)
[info]   at org.rogach.scallop.ScallopConf.verifyBuilder(ScallopConf.scala:343)
[info]   at org.rogach.scallop.ScallopConf.verify(ScallopConf.scala:587)
[info]   at mesosphere.marathon.AllConf.<init>(AllConf.scala:22)
[info]   at mesosphere.marathon.MarathonApp.conf$lzycompute(Main.scala:81)
[info]   at mesosphere.marathon.MarathonApp.conf(Main.scala:81)- A group update can be applied to an empty group *** FAILED *** (61 milliseconds)
  java.lang.NullPointerException:
  at org.rogach.scallop.Scallop.verify(Scallop.scala:624)
  at org.rogach.scallop.ScallopConf.verifyBuilder(ScallopConf.scala:343)
  at org.rogach.scallop.ScallopConf.verify(ScallopConf.scala:587)
  at mesosphere.marathon.AllConf.<init>(AllConf.scala:22)
  at mesosphere.marathon.MarathonApp.conf$lzycompute(Main.scala:81)
  at mesosphere.marathon.MarathonApp.conf(Main.scala:81)
  at mesosphere.marathon.AllConf$$anonfun$1.apply(AllConf.scala:28)
  at mesosphere.marathon.AllConf$$anonfun$1.apply(AllConf.scala:28)
  at scala.Option.getOrElse(Option.scala:121)
  at mesosphere.marathon.AllConf$.option(AllConf.scala:28)
  at mesosphere.marathon.AllConf$.enabledFeatures(AllConf.scala:33)
  at mesosphere.marathon.AllConf$.isFeatureSet(AllConf.scala:37)
```

plugin-interface/src/main/scala/mesosphere/marathon/plugin/optfactory/AppOptFactory.scala
```diff
@@ -0,0 +1,7 @@
+package mesosphere.marathon.plugin.optfactory
+
+import mesosphere.marathon.plugin.AppDefinition
+import mesosphere.marathon.plugin.plugin._
+
+/** AppOptFactory generates functional options that configure objects of type T */
+trait AppOptFactory[T] extends Opt.Factory[AppDefinition, T] with Plugin
```
jdef: Is it worth declaring something so specific? or should we instead declare an `OptFactoryPlugin[P,T] extends Opt.Factory[P,T] with Plugin` ?

src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueueModule.scala
jdef: so it'd be really cool if this func combined all the plugins that satisfied an Opt.Factory[_,_] interface into a single factory via `combine`

plugin-interface/src/main/scala/mesosphere/marathon/plugin/EnvVarValue.scala
```diff
@@ -0,0 +1,11 @@
+package mesosphere.marathon.plugin
+
+trait EnvVarValue
+
+trait EnvVarString extends EnvVarValue {
```
sttts: Why no case class? That would give you static case checking.

jdef: maybe that's what gaston was suggesting? the plugin API looked like mostly
traits so I was trying to stick with that. also, case classes can't be
extended. so if we made this a case class here, then we'd delete the
EnvVarString case class in mesosphere/marathon/state .. does that sound OK?

sttts: Why do we need these types here anyway? Also case classes can be extended if some plugin needs to.

jdef: Well, we want other plugins to use these types. For example, a the plugin
that generates DCOS_SECRETS_DIRECTIVE. Plugins don't depend on the core
marathon APIs -- it's the other way around.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/task/AppTaskProcessor.scala
```diff
@@ -1,13 +1,30 @@
 package mesosphere.marathon.plugin.task
 
 import mesosphere.marathon.plugin.AppDefinition
-import mesosphere.marathon.plugin.plugin
-import org.apache.mesos.Protos
+import mesosphere.marathon.plugin.plugin.Plugin
+import org.apache.mesos.Protos.TaskInfo
 
 /**
   * AppTaskProcessor is a factory func that generates functional options that mutate Mesos
```
sttts: Why "options"? Is this up-to-date?

jdef: doh!

sttts:
Followed up by https://github.com/dcos/marathon/pull/3

SEJeff:
Is this functionality only going to be available in the dcos version of Marathon? That link isn't available to the general public.

Issue #3857, opened 2016-05-03
GROUP_BY constraint - specifying number of values causes deployment to stall


kellyaa:
Deployed a new application with 5 instances.  Specified the following constraint:

 "constraints": [["rack", "GROUP_BY", 5]]

Where there are exactly 5 different values of the "rack" attribute amongst all of the mesos agents.

I would have expected that each of the 5 instances would be deployed on 5 different mesos agents that each have a unique "rack" value.  The result seen is that only one instance is deployed. The other 4 stay in a waiting state indefinitely. If I remove the number of values  and redeploy so that it looks like this:

 "constraints": [["rack", "GROUP_BY"]]

The all 5 instances are deployed, however not spread across all 5 racks - which is expected behavior.  I've tried this against a few different clusters for sanity sake, ensuring sufficient resources were there across all 5 racks - with the same exact result.  

Any ideas what may cause this?

gkleiman:
@kellyaa according to https://mesosphere.github.io/marathon/docs/constraints.html, constraint values have to be specified as a string ("5") instead as an integer (5).

@brndnmtthws created a PR to change that behaviour, but changing your app definition might be a good workaround until the next Marathon release.

brndnmtthws:
This should be resolved by #3865.

kellyaa:
@gkleiman Thanks! Yeah, that was a typo on my part.  Putting quotes around "5" still displayed the same behavior but after help from @brndnmtthws we've realized that the root cause will most definitely be solved by the referenced PR.  Our constraints that we were grouping on were scalar values.  Thanks for your help!!

Issue #1711, opened 2015-06-26
Make Scalar-valued attributes work in constraints


alberts:
Mesos 0.22.1 attributes that have integer values don't work if you try to use them as constraints in Marathon 0.8.2. Jobs never get scheduled.

Not quite sure where things go wrong.

pierlo-upitup:
Hey @alberts,

are you sure the attribute is defined on the slave?

> If the specified attribute is not defined on the slave, then most operators will refuse to run tasks on this slave. In fact, only the UNLIKE operator will (and always will) accept this offer for now, while other operators will always refuse it.

See :
https://github.com/mesosphere/marathon/blob/master/docs/docs/constraints.md#attribute-field

Could you post the relevant parts of your JSON definition for the app you are trying to deploy?

Thanks.

bobrik:
This is how mesos slave reports in `/state` of mesos master api:

```
    {
      "active": true,
      "attributes": {
        "rack": 109,
        "type": "compute"
      },
      "hostname": "myhost",
```

This is how constraint is specified:

```json
[["rack","CLUSTER","109"]]
```

Marathon fails to schedule tasks in this case.

Mesos 0.25.0, Marathon 0.11.1.

kolloch:
This is a known limitation that is documented here (https://mesosphere.github.io/marathon/docs/constraints.html):

"""
Note that currently Marathon only supports text values for attributes and will not be able to process any numberic values. For example the attribute pair foo:bar will be recognised, but marathon will not be able to match cpu:4.
"""

That doesn't make it pretty, though.

I would like to keep this issue as a reminder to extend the functionality.

bobrik:
@kolloch do you have any plans for using numeric constraints differently? I thought that attributes are opaque strings, looks like I could be wrong.

kolloch:
Attributes have the same types as resources but they are not consumed. Thus if the value has the "shape" of an integer, Mesos will use a different type for them. See [attributes/resources in Mesos](http://mesos.apache.org/documentation/attributes-resources/)

bobrik:
Any chance to at least pretend that numeric attributes are strings before implementing full support for numbers and ranges?

This way you can have your `rack:123` attribute now and group by it. Once numbers land, you'll be able to use ranges, if you wish. All of that without changing attributes to hacks like `rack:r123` and wiping mesos slave state twice.

kolloch:
[...]

PR #3865, opened 2016-05-04
Support all attribute types with constraints.

brndnmtthws:
If attributes are some non-text type (such as scalar or range),
constraints will not work because we assume all attributes are text.

To address:
 * https://github.com/mesosphere/marathon/issues/3857
 * https://github.com/mesosphere/marathon/issues/1711

### DIFF
--- a/src/main/scala/mesosphere/mesos/Constraints.scala
+++ b/src/main/scala/mesosphere/mesos/Constraints.scala
@@ -4,7 +4,7 @@ import mesosphere.marathon.Protos.Constraint
 import mesosphere.marathon.Protos.Constraint.Operator
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.state.AppDefinition
-import org.apache.mesos.Protos.Offer
+import org.apache.mesos.Protos.{ Attribute, Offer, Value }
 import org.slf4j.LoggerFactory
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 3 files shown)


air:
hi @aquamatthias, just a heads up we need this in the next available Marathon release. Thanks.

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -25,6 +25,21 @@ object Constraints {
...
+      java.text.NumberFormat.getInstance.format(attribute.getScalar.getValue)
+    case Value.Type.TEXT =>
+      attribute.getText.getValue
+    case Value.Type.RANGES =>
+      val s = attribute.getRanges.getRangeList.asScala
+        .map(r => s"${r.getBegin.toString}-${r.getEnd.toString}")
+        .mkString(",")
+      s"[$s]"
+    case Value.Type.SET =>
+      val s = attribute.getSet.getItemList.asScala.mkString(",")
```
aquamatthias: I think we should sort the set. `"a,b" != "b,a"`
With sorting the use at least know how to specify this.
This should also be part of the documentation.

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -25,6 +25,21 @@ object Constraints {
...
   }
 
+  private def getValueString(attribute: Attribute): String = attribute.getType match {
+    case Value.Type.SCALAR =>
+      java.text.NumberFormat.getInstance.format(attribute.getScalar.getValue)
+    case Value.Type.TEXT =>
+      attribute.getText.getValue
+    case Value.Type.RANGES =>
+      val s = attribute.getRanges.getRangeList.asScala
+        .map(r => s"${r.getBegin.toString}-${r.getEnd.toString}")
```
aquamatthias: I think we should sort the ranges. `"1-2,2-3" != "2-3,1-2"`
With sorting the use at least know how to specify this.
This should also be part of the documentation.

aquamatthias:
@brndnmtthws Thanks for this PR. Added 2 small suggestions.

aquamatthias:
@brndnmtthws thanks

bobrik:
This is great, @brndnmtthws. Can't wait to try the next release.

PR #3873, opened 2016-05-07
Add documentation for MARATHON_SINGLE_APP_INSTANCE label.

ssk2:


ssk2:
cc @sascala

docs/docs/recipes.md
```diff
@@ -241,3 +241,26 @@ You can use constraints to restrict where to run the tasks for your apps. See
...
+
+In your application definition, set the label `MARATHON_SINGLE_INSTANCE_APP` to true:
+
+```json
+  "labels":{
+    "MARATHON_SINGLE_INSTANCE_APP": "true",
+  }
+```
+
+It's also necessary to specify an `upgradeStrategy` - this permits Marathon to kill all instances of this application when upgrading (otherwise it would be impossible to upgrade), and prevents Marathon from launching more than one instance:
```
meichstedt: I'd get rid of the dash. It's not only necessary to specify an upgradeStrategy; it's necessary to specify _an upgradeStrategy that permits Marathon to kill all instances_. Apps with label `SINGLE_INSTANCE_APP` will fail validation (and thus any deployment) if their upgradeStrategy is unequal to `UpgradeStrategy(minimumHealthCapacity = 0, maximumOverCapacity = 0))`.

Moreover, the part _otherwise it would be impossible to upgrade_ is not correct. In general, Marathon will spawn new instances if it cannot start the upgrade by killing at least one task – it handles the upgradeStrategy as a proposal, not as a command.

aquamatthias:
Thanks!

ssk2:
Thanks @sascala!

PR #3887, opened 2016-05-11
External volumes docs

sascala:
This should not appear on the site until 1.7 GAs (currently slated for Monday, 5/16).


sascala:
Does this look OK to you too, @branden ?

jdef:
lgtm

branden:
👍

jdef:
docs say that relative-paths are allowed for mesos container, but actually a "no-slashes" rule is enforced by Marathon when using MESOS containers. For example, when using MESOS containers the path "foo" is acceptable, but "foo/bar" is not.

PR #3890, opened 2016-05-11
Revert "External volumes docs"

sascala:
Reverts mesosphere/marathon#3887

I merged this to Master, then @jdef made the point that having this doc accessible before the external volume code arrives in Marathon/Master could be confusing to users. How would you like to handle this?


aquamatthias:
Makes sense.

PR #3897, opened 2016-05-14
Alauda is powered by Marathon

oilbeater:
Alauda build public CaaS on top of mesos and marathon in multiple datacenters around the world running tens of thousands of apps in each datacenter.


gkleiman:
@oilbeater glad to hear this! Thanks for the PR.

PR #3920, opened 2016-05-23
Add ExpungeOverdueLostTaskActor

aquamatthias:
Add TaskJobsConfig
Add new Actor to module.
Add Test case for overdue lost task expunging.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
@@ -1,0 +1,77 @@
+package mesosphere.marathon.core.task.jobs.impl
+
+import akka.actor.{ ActorLogging, Props, Cancellable, Actor }
+import akka.pattern.pipe
+import com.google.inject.Provider
+import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.core.base.Clock
+import mesosphere.marathon.core.task.bus.MesosTaskStatus.TemporarilyUnreachable
+import mesosphere.marathon.core.task.jobs.TaskJobsConfig
+import mesosphere.marathon.core.task.tracker.TaskTracker
+import mesosphere.marathon.core.task.tracker.TaskTracker.AppTasks
+import mesosphere.marathon.core.task.update.TaskStatusUpdateProcessor
+import mesosphere.marathon.state.PathId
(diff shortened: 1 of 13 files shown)


src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
```diff
@@ -0,0 +1,77 @@
...
+    log.info("ExpungeOverdueLostTasksActor has stopped")
+  }
+
+  override def receive: Receive = {
+    case Tick                             => taskTracker.tasksByApp() pipeTo self
+    case TaskTracker.TasksByApp(appTasks) => filterLostGCTasks(appTasks).foreach(expungeLostGCTask)
+  }
+
+  def expungeLostGCTask(task: MarathonTask): Unit = {
+    log.warning(s"Task ${task.getId} is lost longer than ${config.taskLostExpungeGC} and will be expunged.")
```
meichstedt: I think `... is lost since <timestamp> ...` would be more helpful.

src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
```diff
@@ -0,0 +1,77 @@
...
+    val update = task.getStatus.toBuilder
+      .setState(TaskState.TASK_ERROR)
+      .setReason(mesos.TaskStatus.Reason.REASON_TASK_UNKNOWN)
+      .setTimestamp((clock.now().toDateTime.getMillis / 1000).toDouble)
+      .build()
+    updateProcessor.publish(update, ack = false)
```
meichstedt: Can you add a note explaining why we publish a status update here? I'm afraid people might find this disturbing.

src/main/scala/mesosphere/marathon/core/task/update/TaskStatusUpdateProcessor.scala
```diff
@@ -5,5 +5,5 @@ import org.apache.mesos.Protos.TaskStatus
 import scala.concurrent.Future
 
 trait TaskStatusUpdateProcessor {
-  def publish(status: TaskStatus): Future[Unit]
+  def publish(status: TaskStatus, ack: Boolean = true): Future[Unit]
```
meichstedt: I think a comment on why we decided to add an ack flag might clarify this as well

PR #3941, opened 2016-05-27
HealthChecks: do not perform health checks and task kills, if the tas…

aquamatthias:
…k is not reachable.

### DIFF
--- a/src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
@@ -4,6 +4,7 @@ import akka.actor.{ Actor, ActorLogging, ActorRef, Cancellable, Props }
 import akka.event.EventStream
 import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
 import mesosphere.marathon.Protos.MarathonTask
+import mesosphere.marathon.core.task.bus.MesosTaskStatus.TemporarilyUnreachable
 import mesosphere.marathon.core.task.tracker.TaskTracker
 import mesosphere.marathon.event._
 import mesosphere.marathon.state.AppDefinition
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
```diff
@@ -104,10 +105,15 @@ class HealthCheckActor(
...
       log.info(f"Detected unhealthy task ${task.getId} of app [$app.id] version [$app.version] on host ${task.getHost}")
 
-      // kill the task
-      marathonSchedulerDriverHolder.driver.foreach { driver =>
-        log.info(s"Send kill request for task ${task.getId} on host ${task.getHost} to driver")
-        driver.killTask(TaskID(task.getId))
+      // kill the task, if it is reachable
+      task match {
+        case TemporarilyUnreachable(_) =>
+          log.info(s"Task ${task.getId} on host ${task.getHost} is temporarily unreachable. Performing no task kill.")
```
gkleiman: I think this should be a WARN.

src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
```diff
@@ -104,10 +105,15 @@ class HealthCheckActor(
...
+        case _ =>
+          marathonSchedulerDriverHolder.driver.foreach { driver =>
+            log.info(s"Send kill request for task ${task.getId} on host ${task.getHost} to driver")
```
gkleiman: I think this should be a WARN.
 Add a line note

src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
```diff
@@ -104,10 +105,15 @@ class HealthCheckActor(
     if (consecutiveFailures >= maxFailures && maxFailures > 0) {
       log.info(f"Detected unhealthy task ${task.getId} of app [$app.id] version [$app.version] on host ${task.getHost}")
```
gkleiman: Since we're already changing this method... this could be a WARN =).

gkleiman:
LGTM, but I think that changing the logging level would be handy for future troubleshooting sessions..

gkleiman:
Thanks!

PR #3960, opened 2016-06-02
Fixed tomcat distro uri in deploy example

golonzovsky:
Currently link points to 404 page, therefore this change


kensipe:
LGTM  thx!!

Issue #3967, opened 2016-06-08
change build.scala to build.sbt


shijinkui:
move build.scala to root build.sbt

jasongilanfarr:
As stated in the PR, we do prefer the .scala edition over the .sbt edition of the build. If there is a highly compelling reason to move over to the .sbt edition, please reopen.

shijinkui:
hey @jasongilanfarr ,thank for review. At the newer sbt version, build.sbt is the recommended style, support multi-project.

http://www.scala-sbt.org/0.13/docs/Full-Def.html
>This page describes an old style of .scala build definition. In the previous versions of sbt, .scala was the only way to create multi-project build definition, but sbt 0.13 added multi-project .sbt build definition, which is the recommended style.

PR #3968, opened 2016-06-08
[3967] change build.scala to build.sbt #3967

shijinkui:
this is the first step.

Then I will upgrade the plugin, add sbt-native-packager


shijinkui:
@timcharper @burke @cbuben @stem @rboyer who have time to review this? thx :)

timcharper:
This looks fantastic. :+1: from me.

jasongilanfarr:
Could you explain a bit more about why we should migrate to build.sbt over Build.scala?
Adding sbt-native-packager should be relatively easy to do in Build.scala as well.

timcharper:
@jasongilanfarr at one point there were things that you could do in `Build.scala` but not `build.sbt`; that's no longer true.

There's not a particularly *STRONG* reason to switch, as what you say is true; In general Scala sbt projects in general are converging to use `build.sbt`.

jasongilanfarr:
@timcharper I do agree and have observed that as well (with the notable exception of Akka). I tend to lean toward "don't fix what's not broken" but if there's a compelling use case I'm not aware of.

timcharper:
¯\_(ツ)_/¯ 

It's a low risk change. I'm inclined to say merge it but won't throw a fit if it isnt

jasongilanfarr:
I spoke with some of the other team members and we do prefer the .scala approach over the .sbt approach - even though it is now fully supported.

shijinkui:
hey @jasongilanfarr ,thank for review. At the newer sbt version, build.sbt is the recommended style, support multi-project.

http://www.scala-sbt.org/0.13/docs/Full-Def.html
>This page describes an old style of .scala build definition. In the previous versions of sbt, .scala was the only way to create multi-project build definition, but sbt 0.13 added multi-project .sbt build definition, which is the recommended style.

jasongilanfarr:
Hey @shijinkui, I'm aware of the more recent changes in SBT that enabled multi-project builds to be expressed in Build.scala; however, as stated above, the .scala definition is not longer recommended but has not been deprecated either. Akka, for example, still uses the older .scala edition. Is there a compelling need to change over other than matching up with the new SBT recommendations?

shijinkui:
@jasongilanfarr Indeed SBT does't force to replace build.scala definition. It's only recommend to use build.sbt。
There are two benefit:    
* build.sbt is sbt config spec，not full support scala style, maybe limited spec have less misticks
* most of new SBT project follow new definitation, follow this can make user easy adapt this project（没有违和感）

Go forward in my mind :)

PR #3992, opened 2016-06-16
Fixes #3991 - Use suppressOffers

### LINKED ISSUE #3991: Allow Marathon to suppress offers when there is nothing to launch
Mesos introduces the `suppressOffers` call in 0.25.0. This call tells Mesos to stop sending further offers to the framework until it calls `revive`. 
This alleviates pressure on the allocator, and acts as an infinite filter on all agents until `revive` is called, reducing the amount of work Marathon has to do when it wants to be in an idle mode.

jmlvanre:
When we don't want any more offers, we ask mesos to suppress
them. This alleviates load on the allocator, and acts as an
infinite duration filter for all agents until the next time
we call `Revive`.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
@@ -85,6 +85,11 @@ private[impl] class ReviveOffersActor(
     }
   }
 
+  private[this] def suppressOffers(): Unit = {
+    log.info("=> Suppress offers NOW")
+    driverHolder.driver.foreach(_.suppressOffers())
+  }
+
   override def receive: Receive = LoggingReceive {
     Seq(
       receiveOffersWantedNotifications,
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
```diff
@@ -85,6 +85,11 @@ private[impl] class ReviveOffersActor(
     }
   }
 
+  private[this] def suppressOffers(): Unit = {
+    log.info("=> Suppress offers NOW")
```
aquamatthias: Please remove `=> `.

aquamatthias:
Integration Test failed: ResidentTaskIntegrationTest: persistent volume will be re-attached and keep state 
I rerun to see if this is shaky

src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
aquamatthias: Log statements are in line with the other ones. 
I will create a ticket which goes over all log lines to remove arrows.

aquamatthias:
Thanks!

PR #4001, opened 2016-06-20
Updated Mesos to 0.28.2

nfnt:
Mesos suffered from a memory leak ([MESOS-5449](https://issues.apache.org/jira/browse/MESOS-5449)) that could affect large clusters
and has been fixed in 0.28.2.


aquamatthias:
Already used on master. Close.

Issue #4005, opened 2016-06-20
Also create 'PORT_<NAME>' env variables for ports defined in port mappings


gkleiman:
Right now, `PORT_<NAME>` env variables are only added for ports defined in `portDefinitions`, but not for ports defined in `portMappings`.

cc/ @jdef @jasongilanfarr

kensipe:
It looks like https://github.com/mesosphere/marathon/pull/4008 fixes this.  can we close this @gkleiman ?

gkleiman:
There's still a TODO in the code, so we might not be generating the named env vars in all cases:

https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/TaskBuilder.scala#L420

aquamatthias:
@jdef can you fix this TODO?

jeschkies:
It seems the case @gkleiman mentions is related to #4183. Can you verify this @jdef?

jdef:
I guess I wouldn't consider them closely related. But they should both be
addressed

PR #3998, opened 2016-06-17
WIP: virtual networking for Docker containers

jdef:
- docker.portMappings.hostPort is optional
- support for USER network mode in Marathon scala API
- add USER as a valid networking mode to docker.network JSON schema
- portMappings require hostPort in docker BRIDGE network mode
- move container portMappings, hostPorts, servicePorts from AppDefinition to Container
- marathon serialization proto tests for docker networking
- disallow ipaddress/discoveryInfo for docker containers using BRIDGE/USER networking
- updated docs for USER mode and portMapping
- generate PORT{x} and PORT_{y} variables more consistently
- removed unused portIndex from PortAssignment

Conflicts:
	src/main/java/mesosphere/marathon/Protos.java
	src/main/scala/mesosphere/mesos/TaskBuilder.scala
	src/test/scala/mesosphere/mesos/TaskBuilderTest.scala

PR #4008, opened 2016-06-20
WIP: DO NOT MERGE: fix problem with docker USER networking w/ portMappings and ipAddress

jdef:
#3998 was merged prematurely. this PR attempts to fix additional bugs found in integration testing w/ development builds of dc/os

- TODO: unit test for PORT envvar generation that tests error condition reported (and fixed) below
  - scenarios: (a) no ports are defined; (b) portMappings = `[{}]`
- TODO: see UGLY https://github.com/mesosphere/marathon/pull/4008#issuecomment-227853766
  - this probably is not a blocker for this PR, but should have a follow-up PR to fix

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -279,7 +279,7 @@ case class AppDefinition(
 
   def servicePorts: Seq[Int] = container.flatMap(_.servicePorts).getOrElse(portNumbers)
 
-  def hasDynamicPort: Boolean = servicePorts.contains(AppDefinition.RandomPortValue)
+  def hasDynamicServicePorts: Boolean = servicePorts.contains(AppDefinition.RandomPortValue)
 
   def networkModeBridge: Boolean =
     container.exists(_.docker.exists(_.network.exists(_ == mesos.ContainerInfo.DockerInfo.Network.BRIDGE)))
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -100,7 +100,8 @@ case class AppDefinition(
 
   import mesosphere.mesos.protos.Implicits._
 
-  require(ipAddress.isEmpty || portDefinitions.isEmpty, "IP address and ports are not allowed at the same time")
+  require(ipAddress.isEmpty || portDefinitions.isEmpty,
+    s"IP address ($ipAddress) and ports ($portDefinitions) are not allowed at the same time")
 
   lazy val portNumbers: Seq[Int] = portDefinitions.map(_.port)
 
@@ -273,6 +274,8 @@ case class AppDefinition(
   }
 
   private def portIndices: Range = container.flatMap(_.hostPorts).getOrElse(portNumbers).indices
+
+  def hostPorts: Seq[Option[Int]] = container.flatMap(_.hostPorts).getOrElse(portNumbers.map(Some(_)))
 
   def servicePorts: Seq[Int] = container.flatMap(_.servicePorts).getOrElse(portNumbers)
 
--- a/src/main/scala/mesosphere/marathon/state/GroupManager.scala
+++ b/src/main/scala/mesosphere/marathon/state/GroupManager.scala
@@ -202,10 +202,10 @@ class GroupManager @Inject() (
       }
   }
 
-  //scalastyle:off method.length
+  //scalastyle:off method.length cyclomatic.complexity
   private[state] def assignDynamicServicePorts(from: Group, to: Group): Group = {
     val portRange = Range(config.localPortMin(), config.localPortMax())
-    var taken = from.transitiveApps.flatMap(_.portNumbers) ++ to.transitiveApps.flatMap(_.portNumbers)
+    var taken = from.transitiveApps.flatMap(_.hostPorts.flatten) ++ to.transitiveApps.flatMap(_.hostPorts.flatten)
 
     def nextGlobalFreePort: Int = synchronized {
       val port = portRange.find(!taken.contains(_))
@@ -218,11 +218,14 @@ class GroupManager @Inject() (
       port
     }
 
-    def mergeServicePortsAndPortDefinitions(portDefinitions: Seq[PortDefinition], servicePorts: Seq[Int]) = {
-      portDefinitions.zipAll(servicePorts, AppDefinition.RandomPortDefinition, AppDefinition.RandomPortValue).map {
-        case (portDefinition, servicePort) => portDefinition.copy(port = servicePort)
-      }
-    }
+    def mergeServicePortsAndPortDefinitions(
+      portDefinitions: Seq[PortDefinition],
+      servicePorts: Seq[Int]): Seq[PortDefinition] =
+      if (portDefinitions.nonEmpty)
+        portDefinitions.zipAll(servicePorts, AppDefinition.RandomPortDefinition, AppDefinition.RandomPortValue).map {
+          case (portDefinition, servicePort) => portDefinition.copy(port = servicePort)
+        }
+      else Seq.empty
 
     def assignPorts(app: AppDefinition): AppDefinition = {
 
@@ -230,7 +233,7 @@ class GroupManager @Inject() (
       //if the app uses dynamic ports (0), it will get always the same ports assigned
       val assignedAndAvailable = mutable.Queue(
         from.app(app.id)
-          .map(_.portNumbers.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))
+          .map(_.hostPorts.flatten.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))
           .getOrElse(Nil): _*
       )
 
@@ -249,7 +252,15 @@ class GroupManager @Inject() (
         pms <- d.portMappings
       } yield {
         val mappings = pms.zip(servicePorts).map {
-          case (pm, sp) => pm.copy(servicePort = sp)
+          case (pm, sp) => pm match {
+            // merge service ports and container port mappings:
+            // always assign service port; if hostPort == 0, assign that to service port too
+            // (same affect as mergeServicePortsDefinitions, except on portMappings)
+            case Container.Docker.PortMapping(_, Some(0), 0, _, _, _) =>
+              pm.copy(hostPort = Some(sp), servicePort = sp)
+            case _ =>
+              pm.copy(servicePort = sp)
+          }
         }
         c.copy(
           docker = Some(d.copy(
--- a/src/main/scala/mesosphere/marathon/state/GroupManager.scala
+++ b/src/main/scala/mesosphere/marathon/state/GroupManager.scala
@@ -205,8 +205,7 @@ class GroupManager @Inject() (
   //scalastyle:off method.length cyclomatic.complexity
   private[state] def assignDynamicServicePorts(from: Group, to: Group): Group = {
     val portRange = Range(config.localPortMin(), config.localPortMax())
-    var taken = (from.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts) ++
-      to.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts)).toSet
+    var taken = from.transitiveApps.flatMap(_.servicePorts) ++ to.transitiveApps.flatMap(_.servicePorts)
 
     def nextGlobalFreePort: Int = synchronized {
       val port = portRange.find(!taken.contains(_))
@@ -234,16 +233,16 @@ class GroupManager @Inject() (
       //if the app uses dynamic ports (0), it will get always the same ports assigned
       val assignedAndAvailable = mutable.Queue(
         from.app(app.id)
-          .map(_.hostPorts.flatten.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))
+          .map(_.servicePorts.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))
           .getOrElse(Nil): _*
       )
 
-      def nextFreeAppPort: Int =
+      def nextFreeServicePort: Int =
         if (assignedAndAvailable.nonEmpty) assignedAndAvailable.dequeue()
         else nextGlobalFreePort
 
       val servicePorts: Seq[Int] = app.servicePorts.map { port =>
-        if (port == 0) nextFreeAppPort else port
+        if (port == 0) nextFreeServicePort else port
       }
 
       // defined only if there are port mappings
@@ -252,20 +251,13 @@ class GroupManager @Inject() (
         d <- c.docker
         pms <- d.portMappings
       } yield {
-        val mappings = pms.zip(servicePorts).map {
-          case (pm, sp) => pm match {
-            // merge service ports and container port mappings:
-            // always assign service port; if hostPort == 0, assign that to service port too
-            // (same affect as mergeServicePortsDefinitions, except on portMappings)
-            case Container.Docker.PortMapping(_, Some(0), 0, _, _, _) =>
-              pm.copy(hostPort = Some(sp), servicePort = sp)
-            case _ =>
-              pm.copy(servicePort = sp)
-          }
+        val portMappings = pms.zip(servicePorts).map {
+          case (portMapping, servicePort) =>
+            portMapping.copy(servicePort = servicePort)
         }
+
         c.copy(
-          docker = Some(d.copy(
-            portMappings = Some(mappings)))
+          docker = Some(d.copy(portMappings = Some(portMappings)))
         )
       }
 
@@ -278,12 +270,9 @@ class GroupManager @Inject() (
     val dynamicApps: Set[AppDefinition] =
       to.transitiveApps.map {
         // assign values for service ports that the user has left "blank" (set to zero)
-        case app: AppDefinition if app.hasDynamicPort => assignPorts(app)
+        case app: AppDefinition if app.hasDynamicServicePorts => assignPorts(app)
         case app: AppDefinition =>
           // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
-          // TODO(jdef) not sure why we do this:
-          // (1) HOST mode, servicePorts returns portNumbers; we know all ports != 0 so this is an identity assignment
-          // (2) !HOST mode, servicePorts = some set(mappings.servicePort); we shouldn't generate portDefinitions anyway
           app.copy(
             portDefinitions = mergeServicePortsAndPortDefinitions(app.portDefinitions, app.servicePorts)
           )
--- a/src/main/scala/mesosphere/marathon/state/GroupManager.scala
+++ b/src/main/scala/mesosphere/marathon/state/GroupManager.scala
@@ -205,7 +205,8 @@ class GroupManager @Inject() (
   //scalastyle:off method.length cyclomatic.complexity
   private[state] def assignDynamicServicePorts(from: Group, to: Group): Group = {
     val portRange = Range(config.localPortMin(), config.localPortMax())
-    var taken = from.transitiveApps.flatMap(_.hostPorts.flatten) ++ to.transitiveApps.flatMap(_.hostPorts.flatten)
+    var taken = (from.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts) ++
+      to.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts)).toSet
 
     def nextGlobalFreePort: Int = synchronized {
       val port = portRange.find(!taken.contains(_))
@@ -276,9 +277,13 @@ class GroupManager @Inject() (
 
     val dynamicApps: Set[AppDefinition] =
       to.transitiveApps.map {
+        // assign values for service ports that the user has left "blank" (set to zero)
         case app: AppDefinition if app.hasDynamicPort => assignPorts(app)
         case app: AppDefinition =>
           // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
+          // TODO(jdef) not sure why we do this:
+          // (1) HOST mode, servicePorts returns portNumbers; we know all ports != 0 so this is an identity assignment
+          // (2) !HOST mode, servicePorts = some set(mappings.servicePort); we shouldn't generate portDefinitions anyway
           app.copy(
             portDefinitions = mergeServicePortsAndPortDefinitions(app.portDefinitions, app.servicePorts)
           )
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -289,13 +289,23 @@ object TaskBuilder {
                   host: Option[String],
                   hostPorts: Seq[Option[Int]],
                   envPrefix: Option[String]): CommandInfo.Builder = {
-    val containerPorts = for {
-      c <- runSpec.container
-      pms <- c.portMappings
-    } yield pms.map(_.containerPort)
-    val declaredPorts = containerPorts.getOrElse(runSpec.portNumbers)
-    // TODO(jdef) why doesn't this pull port names from container.portMappings?
-    val portNames = runSpec.portDefinitions.map(_.name)
+
+    val declaredPorts = {
+      val containerPorts = for {
+        c <- runSpec.container
+        pms <- c.portMappings
+      } yield pms.map(_.containerPort)
+
+      containerPorts.getOrElse(runSpec.portNumbers)
+    }
+    val portNames = {
+      val containerPortNames = for {
+        c <- runSpec.container
+        pms <- c.portMappings
+      } yield pms.map(_.name)
+
+      containerPortNames.getOrElse(runSpec.portDefinitions.map(_.name))
+    }
 
     val envMap: Map[String, String] =
       taskContextEnv(runSpec, taskId) ++
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -415,7 +415,7 @@ object TaskBuilder {
       }
 
       val allAssigned = effectivePorts.flatten ++ generatedPorts.values
-      env += ("PORT" -> allAssigned.head.toString)
+      allAssigned.headOption.foreach { port => env += ("PORT" -> port.toString) }
       env += ("PORTS" -> allAssigned.mkString(","))
       env.result()
     }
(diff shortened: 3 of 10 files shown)


src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -218,19 +218,22 @@ class GroupManager @Inject() (
       port
     }
 
-    def mergeServicePortsAndPortDefinitions(portDefinitions: Seq[PortDefinition], servicePorts: Seq[Int]) = {
-      portDefinitions.zipAll(servicePorts, AppDefinition.RandomPortDefinition, AppDefinition.RandomPortValue).map {
```
jdef: This is inserting service ports even when `portDefinitions` is empty! So I fixed it, and then service ports weren't merging properly with container port mappings. So I changed those bits too.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -249,7 +252,15 @@ class GroupManager @Inject() (
         pms <- d.portMappings
       } yield {
         val mappings = pms.zip(servicePorts).map {
-          case (pm, sp) => pm.copy(servicePort = sp)
+          case (pm, sp) => pm match {
+            // merge service ports and container port mappings:
+            // always assign service port; if hostPort == 0, assign that to service port too
+            // (same affect as mergeServicePortsDefinitions, except on portMappings)
+            case Container.Docker.PortMapping(_, Some(0), 0, _, _, _) =>
```
jdef: This is basically what mergeServicePortsDefinitions is doing, except this works on container port mappings instead of port definitions.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -304,7 +305,58 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
   }
 
   test("Create a new app with IP/CT with virtual network foo w/ Docker") {
+    useRealGroupManager()
+    object Config extends ScallopConf() {
```
jdef: this seems silly. there must be a better way

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -274,6 +275,8 @@ case class AppDefinition(
 
   private def portIndices: Range = container.flatMap(_.hostPorts).getOrElse(portNumbers).indices
 
+  def hostPorts: Seq[Option[Int]] = container.flatMap(_.hostPorts).getOrElse(portNumbers.map(Some(_)))
```
jdef: there's probably other places this is useful too, so I preserved the arity instead of flattening

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -341,6 +342,89 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
     JsonTestHelper.assertThatJsonString(response.getEntity.asInstanceOf[String]).correspondsToJsonOf(expected)
   }
 
+  /*
+  test("Create a new app with IP/CT with virtual network foo w/ Docker") {
```
jdef: I really want this version of the unit test to work but I get a validation error with ClassCastExceptions that follow. no comprendo.

jdef:
seeing a new error on a hot-patched cluster during integration testing:
```
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: [2016-06-21 03:46:50,183] ERROR empty.head (akka.actor.OneForOneStrategy:marathon-akka.actor.default-dispatcher-9)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: java.lang.UnsupportedOperationException: empty.head
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.collection.immutable.Vector.head(Vector.scala:193)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.mesos.TaskBuilder$.portsEnv(TaskBuilder.scala:418)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.mesos.TaskBuilder$.commandInfo(TaskBuilder.scala:301)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.mesos.TaskBuilder.build(TaskBuilder.scala:131)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.mesos.TaskBuilder.build(TaskBuilder.scala:66)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.mesos.TaskBuilder.buildIfMatches(TaskBuilder.scala:85)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.marathon.core.launcher.impl.TaskOpFactoryImpl.inferNormalTaskOp(TaskOpFactoryImpl.scala:50)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.marathon.core.launcher.impl.TaskOpFactoryImpl.buildTaskOp(TaskOpFactoryImpl.scala:43)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.marathon.core.launchqueue.impl.TaskLauncherActor$$anonfun$receiveProcessOffers$1.applyOrElse(TaskLauncherActor.scala:363)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:170)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.PartialFunction$OrElse.applyOrElse(PartialFunction.scala:171)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at akka.actor.Actor$class.aroundReceive(Actor.scala:467)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at mesosphere.marathon.core.launchqueue.impl.TaskLauncherActor.aroundReceive(TaskLauncherActor.scala:76)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at akka.actor.ActorCell.receiveMessage(ActorCell.scala:516)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at akka.actor.ActorCell.invoke(ActorCell.scala:487)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at akka.dispatch.Mailbox.processMailbox(Mailbox.scala:238)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at akka.dispatch.Mailbox.run(Mailbox.scala:220)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(AbstractDispatcher.scala:397)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
Jun 21 03:46:50 ip-10-10-0-199 java[20164]: at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
```

jdef:
previous reported error caused with this app JSON (note: hostport == None; only one port mapping)
```json
{
  "id": "jdef-networking",
  "cmd": "ip -o addr; sleep 30",
  "cpus": 0.10,
  "mem": 64,
  "instances": 1,
  "backoffFactor": 1.14472988585,
  "backoffSeconds": 5,
  "ipAddress": { "networkName": "overlay-1" },
  "container": {
    "type": "DOCKER",
    "docker": {
      "network": "USER",
      "image": "busybox",
      "portMappings": [{
        "containerPort": 123, "servicePort": 80, "name": "foo"
      }]
    }
  }
}
```

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -202,10 +202,11 @@ class GroupManager @Inject() (
       }
   }
 
-  //scalastyle:off method.length
+  //scalastyle:off method.length cyclomatic.complexity
   private[state] def assignDynamicServicePorts(from: Group, to: Group): Group = {
     val portRange = Range(config.localPortMin(), config.localPortMax())
-    var taken = from.transitiveApps.flatMap(_.portNumbers) ++ to.transitiveApps.flatMap(_.portNumbers)
+    var taken = (from.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts) ++
+      to.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts)).toSet
```
jdef: TODO: write unit test for this change.
Symptom: multiple apps using port mappings that didn't specify service ports would result in Marathon attempting to reassign the same service port (that was already assigned to app1) to app2. (e.g. `portMappings: [{}]`)

jdef: done

jdef:
`PORTS` still seems inconsistent:
```
PORTS=23409,10003
PORT0=123
PORT1=23409
PORT2=10003
PORT=23409
PORT_23409=23409
PORT_123=123
PORT_10003=10003
```
given app def:
```
{
  "id": "/jdef-networking",
  "cmd": "env; ip -o addr; sleep 30",
  "cpus": 0.1, "mem": 64, "instances": 1,
  "container": {
    "type": "DOCKER",
    "docker": {
      "image": "busybox",
      "network": "USER",
      "portMappings": [
        { "containerPort": 123 },
        { "hostPort": 23409 },
        {}
      ]
    }
  }
}
```

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
jdef: @gkleiman @aquamatthias suggestions welcome

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jdef: @gkleiman @aquamatthias I believe this is consistent with prior (pre-vnetworking) behavior. Please say something now if you think this will break the world.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -415,7 +415,7 @@ object TaskBuilder {
       }
 
       val allAssigned = effectivePorts.flatten ++ generatedPorts.values
-      env += ("PORT" -> allAssigned.head.toString)
+      allAssigned.headOption.foreach { port => env += ("PORT" -> port.toString) }
```
gkleiman: Should we add a test in which no ports are requested?

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
...
+      AppDefinition("/app1".toPath, portDefinitions = Seq(), container = c1),
+      AppDefinition("/app2".toPath, portDefinitions = Seq(), container = c2)
+    ))
+    val update = manager(minServicePort = 10, maxServicePort = 20).assignDynamicServicePorts(Group.empty, group)
+    update.transitiveApps.filter(_.hasDynamicPort) should be ('empty)
+    update.transitiveApps.flatMap(_.hostPorts.flatten.filter(x => x >= 10 && x <= 20)).toSet should have size 0
+    update.transitiveApps.flatMap(_.servicePorts.filter(x => x >= 10 && x <= 20)).toSet should have size 2
+  }
+
+  test("Assign dynamic service ports w/ anonymous host-ports specified in multiple containers") {
```
gkleiman: what's an anonymous host-port?

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
...
+        portMappings = Some(Seq(
+          PortMapping(containerPort = 8081)
+        ))
+      ))
+    ))
+    val group = Group(PathId.empty, Set(
+      AppDefinition("/app1".toPath, portDefinitions = Seq(), container = c1),
+      AppDefinition("/app2".toPath, portDefinitions = Seq(), container = c2)
+    ))
+    val update = manager(minServicePort = 10, maxServicePort = 20).assignDynamicServicePorts(Group.empty, group)
```
gkleiman: I think that if this test used the range [10-12), then it would be more likely to fail if the service ports becomes random (with duplicates).

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
...
+      AppDefinition("/app1".toPath, portDefinitions = Seq(), container = c1)
+    ))
+    val toGroup = Group(PathId.empty, Set(
+      AppDefinition("/app1".toPath, portDefinitions = Seq(), container = c1),
+      AppDefinition("/app2".toPath, portDefinitions = Seq(), container = c2)
+    ))
+    // the port range 10 - 12 actually excludes 12
+    val mgr = manager(minServicePort = 10, maxServicePort = 12)
+    val groupsV1 = mgr.assignDynamicServicePorts(Group.empty, fromGroup)
+    groupsV1.transitiveApps.filter(_.hasDynamicPort) should be ('empty)
```
gkleiman: `hasDynamicPort` is confusing. `c1` requests a "dynamic port" on the agent, but after assigning service ports, this method will return false.

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
...
+          PortMapping(containerPort = 8081)
+        ))
+      ))
+    ))
+    val group = Group(PathId.empty, Set(
+      AppDefinition("/app1".toPath, portDefinitions = Seq(), container = c1),
+      AppDefinition("/app2".toPath, portDefinitions = Seq(), container = c2)
+    ))
+    val update = manager(minServicePort = 10, maxServicePort = 20).assignDynamicServicePorts(Group.empty, group)
+    update.transitiveApps.filter(_.hasDynamicPort) should be ('empty)
```
gkleiman: You can use `empty` instead of `'empty`

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
gkleiman: You can use `empty` instead of `'empty`

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
...
+    val toGroup = Group(PathId.empty, Set(
+      AppDefinition("/app1".toPath, portDefinitions = Seq(), container = c1),
+      AppDefinition("/app2".toPath, portDefinitions = Seq(), container = c2)
+    ))
+    // the port range 10 - 12 actually excludes 12
+    val mgr = manager(minServicePort = 10, maxServicePort = 12)
+    val groupsV1 = mgr.assignDynamicServicePorts(Group.empty, fromGroup)
+    groupsV1.transitiveApps.filter(_.hasDynamicPort) should be ('empty)
+    val groupsV2 = mgr.assignDynamicServicePorts(groupsV1, toGroup)
+    groupsV2.transitiveApps.filter(_.hasDynamicPort) should be ('empty)
```
gkleiman: You can use `empty` instead of `'empty`

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
...
+    val assignedHostPorts = groupsV2.transitiveApps.flatMap(_.hostPorts.flatten.filter(x => x >= 10 && x <= 12)).toSet
```
gkleiman: what do you think about adding a check ensuring that there's one "dynamicHostPort" (a `Some(0)`)?

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
jdef: what's the difference? I just copy/pasted

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
gkleiman: The idea of the tests in this class is to exercise `AppsResource`'s methods with mocked dependencies, not with a proper `GroupManager`.

Do we already have a test in `GroupManager` that checks that it assign the right ports with an app like the one in this test? If the answer is yes, then we don't need this one. If it is not, then maybe we should add one there =).

I tried to make this test pass, but there are multiple problems with it, and I am not sure if we really need it.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -202,10 +202,11 @@ class GroupManager @Inject() (
       }
   }
 
-  //scalastyle:off method.length
+  //scalastyle:off method.length cyclomatic.complexity
```
gkleiman: :cold_sweat:

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -202,10 +202,11 @@ class GroupManager @Inject() (
...
   private[state] def assignDynamicServicePorts(from: Group, to: Group): Group = {
     val portRange = Range(config.localPortMin(), config.localPortMax())
-    var taken = from.transitiveApps.flatMap(_.portNumbers) ++ to.transitiveApps.flatMap(_.portNumbers)
+    var taken = (from.transitiveApps.flatMap(app => app.hostPorts.flatten ++ app.servicePorts) ++
```
gkleiman: Would the following also work?

```
    var taken = from.transitiveApps.flatMap(_.servicePorts) ++ to.transitiveApps.flatMap(_.servicePorts)
```

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -218,19 +219,22 @@ class GroupManager @Inject() (
...
+      else Seq.empty
 
     def assignPorts(app: AppDefinition): AppDefinition = {
 
       //all ports that are already assigned in old app definition, but not used in the new definition
       //if the app uses dynamic ports (0), it will get always the same ports assigned
       val assignedAndAvailable = mutable.Queue(
         from.app(app.id)
-          .map(_.portNumbers.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))
+          .map(_.hostPorts.flatten.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))
```
gkleiman: Would `.map(_.servicePorts.filter(p => portRange.contains(p) && !app.servicePorts.contains(p)))` be correct?

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
gkleiman: IMHO the only difference is that it is easier to read.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
gkleiman: I don't think we really need this change, and it makes the code harder to read.

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
```diff
@@ -87,7 +87,80 @@ class GroupManagerTest extends MarathonActorSupport with MockitoSugar with Match
     ))
     val update = manager(minServicePort = 10, maxServicePort = 20).assignDynamicServicePorts(Group.empty, group)
     update.transitiveApps.filter(_.hasDynamicPort) should be ('empty)
-    update.transitiveApps.flatMap(_.portNumbers.filter(x => x >= 10 && x <= 20)) should have size 2
+    update.transitiveApps.flatMap(_.hostPorts.flatten.filter(x => x >= 10 && x <= 20)) should have size 2
```
gkleiman: s/hostPorts.flatten/servicePorts/

jdef:
UGLY: when in docker USER network mode generated container ports (when containerPort == 0 and hostPort == None) do indeed result in PORT0 being populated. so that's useful from an app perspective but it's not aligned with the service port and the appdef isn't updated with the generated container port -- which means that nothing outside of the app can associate the PORT0 value with the service port. not very helpful.

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jdef: Actually I did need this change, otherwise things broke

gkleiman: What broke? I tried undoing this changes, together with using `servicePorts` instead of `hostPorts` in the new tests, and the tests were green.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
jdef: I started hacking on the group manager tests because this seemed futile. Was hoping that I was just missing something simple but from your explanation it seems I'm pushing the limits of what this suite of unit tests is aimed at.

There's probably a case to make for tests that live somewhere in between simpler unit testing and full-fledged integration testing. For example I want to round-trip test from a JSON create-app request down through an actual group manager, without the hassle of setting up a mesos environment that supports overlay networking.

We can probably remove this commented test for now since I've tried to augment via group manager tests, even though it's less ideal IMHO.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
jdef: yes. I've elaborated on the TODO in the top-level desc of this PR

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
jdef: originally was testing both apps w/ hostport == none, but that wasn't quite right. this test case desc needs an update now

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
jdef: other tests use this same assertion. is it more confusing here vs. elsewhere?

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
jdef: once service ports have been assigned, any prior mapping w/ `hostport==Some(0),servicePort==0` will now be set to `hostport=Some(assignedServicePort),servicePort=assignedServicePort`

jdef: ah. I see, because hostPort==None we'll still have a servicePort == 0 and you want to test for that?

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
jdef: why? your suggestion is not consistent with prior behavior. I changed `portNumbers` to `hostPorts.flatten` here to also consider the host ports specified for containers (vs the host ports specified in `ports`). using `servicePorts` instead is a much different semantic.

we could **add** a check for `servicePorts` but they would not all fall into a range of `[10,20)` since there's a manually specified servicePort of `10555`

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jdef: I guess I'd have to see your changes to understand why it worked for you. Without this change I saw broken tests

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jdef: that depends: how wrong was the original code? substituting `hostPorts` for `portNumbers` uses host ports from portmappings when there's a container. using `servicePorts` instead sounds like a different semantic

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jdef: agreed. this ports stuff is way too complex

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jdef: I'm seeing a theme here where you want to use service ports instead of host ports everywhere. I don't understand it. I'm probably missing something obvious. My understanding of `portNumbers` was that they were cousins of `hostPorts`. `servicePorts` in my mind are completely different

src/test/scala/mesosphere/marathon/state/GroupManagerTest.scala
jdef: sorry, misspoke. don't understand what you're saying here

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -289,12 +289,23 @@ object TaskBuilder {
...
+
+    val declaredPorts = {
+      val containerPorts = for {
+        c <- runSpec.container
+        pms <- c.portMappings
+      } yield pms.map(_.containerPort)
+
+      containerPorts.getOrElse(runSpec.portNumbers)
+    }
+    val portNames = {
```
jdef: fixes #4005

aquamatthias:
Thanks!

PR #4016, opened 2016-06-22
Make Framework authentication explicit by using a command line flag

aquamatthias:
Only if the flag is set to true, we do framework authentication (default false)
The secret needs to be optional.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -223,7 +223,7 @@ trait MarathonConf
   lazy val mesosAuthentication = toggle("mesos_authentication",
     default = Some(false),
     noshort = true,
-    descrYes = "Will enforce framework authentication while registering with Mesos with principal and optional secret.",
+    descrYes = "Will use framework authentication while registering with Mesos with principal and optional secret.",
     descrNo = "(Default) will not use framework authentication while registering with Mesos.",
     prefix = "disable_"
   )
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -220,15 +220,29 @@ trait MarathonConf
     noshort = true
   )
 
+  lazy val mesosAuthentication = toggle("mesos_authentication",
+    default = Some(false),
+    noshort = true,
+    descrYes = "Will enforce framework authentication while registering with Mesos with principal and optional secret.",
```
gkleiman: s/enforce/use/ ?

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -220,15 +220,29 @@ trait MarathonConf
...
+  )
+  dependsOnAll(mesosAuthentication, List(mesosAuthenticationPrincipal))
+
   lazy val mesosAuthenticationPrincipal = opt[String]("mesos_authentication_principal",
     descr = "Mesos Authentication Principal.",
     noshort = true
   )
 
   lazy val mesosAuthenticationSecretFile = opt[String]("mesos_authentication_secret_file",
+    descr = "Mesos Authentication Secret written to a file.",
```
gkleiman: s/Mesos Authentication Secret written to a file./Path to a file containing the Mesos Authentication Secret/ ?

adam-mesos:
When you merge this into sec18, please verify that Marathon is correctly authenticating with Mesos master when it registers. You'll need to change some of the build/config in dcos-image

gkleiman:
Thanks!

PR #4018, opened 2016-06-22
Scala friendly wrapper around Apache Curator

jasongilanfarr:
- Future based version of the curator client.
- UnitTest/WordSpec mixin with common classes
  available.
- PortAllocator which should be used to allocate ports
  for testing and once adopted, it should give
  us the ability to run a lot of tests in parallel.
- Add annotation to tag classes as IntegrationTests.


titosand:
retest this please

jasongilanfarr:
This is based on a slightly out of date branch (with all of the library upgrades).

PR #4020, opened 2016-06-22
report buildref in logs and info resource

jdef:


project/build.scala
```diff
@@ -48,7 +48,12 @@ object MarathonBuild extends Build {
         libraryDependencies ++= Dependencies.root,
         parallelExecution in Test := false,
         sourceGenerators in Compile <+= buildInfo,
-        buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion),
+        buildInfoKeys := Seq[BuildInfoKey](
+          name, version, scalaVersion,
+          BuildInfoKey.action("buildref") {
+            val suffix = Process("git diff --shortstat").lines.headOption.map(_ => "-dev").getOrElse("")
+            Process("git rev-parse HEAD").lines.headOption.getOrElse("unknown") + suffix
```
aquamatthias: This will fail if git is not in path. Please call in a `Try`
```
Try(Process("git rev-parse HEAD").lineStream).toOption.flatMap(_.headOption).getOrElse("unknown") + suffix
```
The same for applies for shortstat.

aquamatthias:
@jdef A small suggestion + rebase please and I will merge.

jdef:
rebased

jdef:
@aquamatthias PTAL

aquamatthias:
Nice. Thanks!

jdef:
the builds on TC generate `Starting Marathon 1.2.0-SNAPSHOT/unknown` which is less than desirable. Pretty sure that `git` is available on TC so it must be something screwy with the repo? I had a similar problem with the k8s-mesos TC build configuration. Haven't been able to remember how I fixed it.

aquamatthias:
@jdef The integration tests on TC are build/run inside a docker container:
The git version information is not part of the docker context, see:
https://github.com/mesosphere/marathon/blob/master/.dockerignore#L6

This leads to this message during compile:
```
[info] Formatting 316 Scala sources {file:/marathon/}marathon(compile) ...
Not a git repository
To compare two paths outside a working tree:
usage: git diff [--no-index] <path> <path>
fatal: Not a git repository (or any of the parent directories): .git
[info] Compiling 316 Scala sources and 3 Java sources to /marathon/target/scala-2.11/classes...
```

In order to have the version information for integration tests, we need to remove this line.

jdef:
Hm .. either that or else we add a host-volume mount for the `.git`
directory to the `docker run` command

PR #4021, opened 2016-06-23
While taking over leadership, Marathon need to handle fatal errors.

aquamatthias:
Native code handling is a fatal error.
If for example the scheduler driver can not be loaded (native code), those exception were not handled at all. 
The only good reason to handle fatal errors I can think of, is to exit the process.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/base/CurrentRuntime.scala
+++ b/src/main/scala/mesosphere/marathon/core/base/CurrentRuntime.scala
@@ -5,6 +5,7 @@ import org.slf4j.LoggerFactory
 import scala.concurrent.{ ExecutionContext, TimeoutException, Await, Future }
 import scala.util.control.NonFatal
 import scala.concurrent.duration._
+import scala.concurrent.blocking
 
 object CurrentRuntime {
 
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/core/base/CurrentRuntime.scala
```diff
@@ -0,0 +1,37 @@
...
+    *
+    * @param exitCode the exit code to signal.
+    * @param waitForExit the time to wait for a normal exit.
+    * @return the Future of this operation.
+    */
+  //scalastyle:off magic.number
+  def asyncExit(exitCode: Int = 9,
+                waitForExit: FiniteDuration = 10.seconds)(implicit ec: ExecutionContext): Future[Unit] = {
+    Future({
+      try {
```
jasongilanfarr: Add blocking{} around block Future blocks to inform ExecutionContext. (scala.concurrent.blocking).

Otherwise seems reasonable. I think it'd be even better to use a timer of sorts, eg:
'def timeout[T](in: Duration)(f: => T): Future[T]' and 'def timeout[T](in: Duration)(f: => Future[T]): Future[T]'

That way we can quickly express "operation X should complete within N" for any arbitrary method.

The method taking in f: => T should wrap 'f' in blocking()

aquamatthias:
We need to cherry pick this on 1.2

src/main/scala/mesosphere/marathon/core/base/CurrentRuntime.scala
aquamatthias: Thanks for the hint

jasongilanfarr:
LGTM

meichstedt:
LGTM too, waiting for integration tests

PR #4022, opened 2016-06-23
We need Mesos 1.0 for the docker image to work.

aquamatthias:
Since there will be no release soon, we need to cut the RC against a Mesos RC.


gkleiman:
Thanks

PR #4023, opened 2016-06-23
Restructured the Container class towards UC

bernd-mesos:
Restructured the Container class in anticipation of Universal Containerizer support.

In the course of this, changed all occurences of "{Protos => Mesos}"
to "{Protos => mesos}" for uniformity across the project.

(rebased to mesosphere/marathon - does not compile yet - Scala compiler crashes)

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/RestResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/RestResource.scala
@@ -4,13 +4,12 @@ import java.net.URI
 import javax.ws.rs.core.Response
 import javax.ws.rs.core.Response.{ ResponseBuilder, Status }
 
-import mesosphere.marathon.MarathonConf
+import mesosphere.marathon.{ MarathonConf, ValidationFailedException }
 import mesosphere.marathon.api.v2.json.Formats._
 import mesosphere.marathon.state.{ PathId, Timestamp }
 import mesosphere.marathon.upgrade.DeploymentPlan
 import play.api.libs.json.Json.JsValueWrapper
-import play.api.libs.json.{ Writes, Json }
-
+import play.api.libs.json.{ Json, Writes }
 import com.wix.accord._
 import mesosphere.marathon.api.v2.Validation._
 
@@ -55,22 +54,32 @@ trait RestResource {
   //scalastyle:off
   /**
     * Checks if the implicit validator yields a valid result.
-    * @param t object to validate
+    *
+    * @param ft          function that returns the object to validate
     * @param description optional description which might be injected into the failure message
-    * @param fn function to execute after successful validation
-    * @param validator validator to use
+    * @param fn          function to execute after successful validation
+    * @param validator   validator to use
     * @tparam T type of object
     * @return returns a 422 response if there is a failure due to validation. Executes fn function if successful.
     */
-  protected def withValid[T](t: T, description: Option[String] = None)(fn: T => Response)(implicit validator: Validator[T]): Response = {
+  protected def withValid[T](ft: => T, description: Option[String] = None)(fn: T => Response)(implicit validator: Validator[T]): Response = {
     //scalastyle:on
-    validator(t) match {
-      case f: Failure =>
-        val entity = Json.toJson(description.map(f.withDescription).getOrElse(f)).toString
+    try {
+      val t: T = ft
+      validator(t) match {
+        case f: Failure =>
+          val entity = Json.toJson(description.map(f.withDescription).getOrElse(f)).toString
+          //scalastyle:off magic.number
+          Response.status(422).entity(entity).build()
+        //scalastyle:on
+        case Success => fn(t)
+      }
+    } catch {
+      case ValidationFailedException(validator, failure) =>
+        val entity = Json.toJson(description.map(failure.withDescription).getOrElse(failure)).toString
         //scalastyle:off magic.number
         Response.status(422).entity(entity).build()
       //scalastyle:on
-      case Success => fn(t)
     }
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
+++ b/src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
@@ -2,7 +2,6 @@ package mesosphere.marathon.api.serialization
 
 import mesosphere.marathon.Protos
 import mesosphere.marathon.core.externalvolume.ExternalVolumes
-import mesosphere.marathon.state.Container.Docker
 import mesosphere.marathon.state.Container.Docker.PortMapping
 import mesosphere.marathon.state._
 import org.apache.mesos
@@ -13,27 +12,51 @@ import scala.collection.immutable.Seq
 object ContainerSerializer {
   def toProto(container: Container): Protos.ExtendedContainerInfo = {
     val builder = Protos.ExtendedContainerInfo.newBuilder
-      .setType(container.`type`)
       .addAllVolumes(container.volumes.map(VolumeSerializer.toProto).asJava)
-    container.docker.foreach { d => builder.setDocker(DockerSerializer.toProto(d)) }
+
+    container match {
+      case _: Container.Mesos =>
+        builder.setType(mesos.Protos.ContainerInfo.Type.MESOS)
+      case docker: Container.Docker =>
+        builder.setType(mesos.Protos.ContainerInfo.Type.DOCKER)
+        builder.setDocker(DockerSerializer.toProto(docker))
+    }
+
     builder.build
   }
 
   def fromProto(proto: Protos.ExtendedContainerInfo): Container = {
-    val maybeDocker = if (proto.hasDocker) Some(DockerSerializer.fromProto(proto.getDocker)) else None
-    Container(
-      `type` = proto.getType,
-      volumes = proto.getVolumesList.asScala.map(Volume(_)).to[Seq],
-      docker = maybeDocker
-    )
+    if (proto.hasDocker) {
+      val d = proto.getDocker
+      val pms = d.getPortMappingsList.asScala
+      Container.Docker(
+        volumes = proto.getVolumesList.asScala.map(Volume(_)).to[Seq],
+        image = d.getImage,
+        network = if (d.hasNetwork) Some(d.getNetwork) else None,
+        portMappings = if (pms.isEmpty) None else Some(pms.map(PortMappingSerializer.fromProto).to[Seq]),
+        privileged = d.getPrivileged,
+        parameters = d.getParametersList.asScala.map(Parameter(_)).to[Seq],
+        forcePullImage = if (d.hasForcePullImage) d.getForcePullImage else false
+      )
+    } else {
+      Container.Mesos(
+        volumes = proto.getVolumesList.asScala.map(Volume(_)).to[Seq]
+      )
+    }
   }
 
   def toMesos(container: Container): mesos.Protos.ContainerInfo = {
-    val builder = mesos.Protos.ContainerInfo.newBuilder.setType(container.`type`)
+    val builder = mesos.Protos.ContainerInfo.newBuilder
 
-    // first set container.docker because the external volume provider might depend on it
-    // to set further values.
-    container.docker.foreach { d => builder.setDocker(DockerSerializer.toMesos(d)) }
+    // First set type-specific values (for Docker) because the external volume provider
+    // might depend on it to set further values (builder is passed down as arg below).
+    container match {
+      case _: Container.Mesos =>
+        builder.setType(mesos.Protos.ContainerInfo.Type.MESOS)
+      case docker: Container.Docker =>
+        builder.setType(mesos.Protos.ContainerInfo.Type.DOCKER)
+        builder.setDocker(DockerSerializer.toMesos(docker))
+    }
 
     container.volumes.foreach {
       case pv: PersistentVolume => // PersistentVolumes are handled differently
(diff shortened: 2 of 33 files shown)


src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,20 +3,31 @@ package mesosphere.marathon.state
...
+
 import scala.collection.immutable.Seq
 
-// TODO: trait Container and specializations?
-// Current implementation with type defaulting to DOCKER and docker to NONE makes no sense
-case class Container(
-    `type`: Mesos.ContainerInfo.Type = Mesos.ContainerInfo.Type.DOCKER,
-    volumes: Seq[Volume] = Nil,
-    docker: Option[Container.Docker] = None) {
+case class Container(`type`: mesos.ContainerInfo.Type,
```
meichstedt: The original idea to refactor the Container model was more like this (simplified):
```
sealed trait Container {
  def volumes: Seq[Volume]
}

case class Docker(volumes: Seq[Volume]) extends Container
case class Mesos(volumes: Seq[Volume]) extends Container
case class Universal(volumes: Seq[Volume]) extends Container
```

This would let us get rid of things like _hasConsistentType_, lots of overloaded constructors, and having a mesos.Protos.ContainerInfo.Type as a member of the Container. Plus, it would make constructing a container based on a proto much easier, and type safe, as each specific case class implementing the Container trait would require a specific set of properties.

Did you consider this approach?

bernd-mesos: Yes, I tried this first. But with a tendency towards the least commotion in other files, I could not use this approach. I googled and tinkered a lot to get Formats working with this and failed. Play documentation is mute about this natural pattern. :-(

The type field is still needed for backwards compatibility. If it is not present, ContainerFormats won't work similar to the way they used to.

All that said: we could write a custom reader and writer to make it as you want. But it's more code and not the prevailing pattern in Formats. If you want this, I can go for it.

bernd-mesos: As just discussed verbally, I'll now go for a clean Container class and work around this with custom reader/writer as needed.

bernd-mesos:
Ready for review!

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,75 +3,64 @@ package mesosphere.marathon.state
...
+  }
+
+  def foldDocker[R](defaultResult: R)(f: Container.Docker => R): R = {
+    this match {
+      case docker: Container.Docker => f(docker)
+      case _ => defaultResult
+    }
+  }
+
+  def mapDocker(f: Container.Docker => Container.Docker): Container = {
```
jdef: this appears to be unused. it's also confusing because map functions typically take a different form (`A => Option[B]`)

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,75 +3,64 @@ package mesosphere.marathon.state
...
+    foldDocker(this)(f(_))
+  }
 
-  import Container._
+  def docker(): Option[Container.Docker] = foldDocker[Option[Container.Docker]](None)(Some(_))
 
-  def portMappings: Option[Seq[Docker.PortMapping]] = {
-    import Mesos.ContainerInfo.DockerInfo.Network
+  def dockerPortMappings: Option[Seq[Container.Docker.PortMapping]] = {
```
jdef: i'm not crazy about the `dockerPortMappings`, `dockerHostPorts`, and `dockerServicePorts` renames. going forward the network team is going to implement support for port mappings w/ mesos containers. so at this level of the API I'd rather keep the container-type-agnostic function names to expose port information.

bernd-mesos:
The follow-up PR which introduces UC support can be looked at here:
https://github.com/bernd-mesos/marathon-1/pull/1

src/main/scala/mesosphere/marathon/state/Container.scala
bernd-mesos: Oops, leftovers from the other solution. Removing the two unused functions (keeping `foldDocker` only. (BTW: this kind of folding of one select case, not just for Option, should be a general feature of the language. Any tips where I can find such a thing?)

src/main/scala/mesosphere/marathon/state/Container.scala
bernd-mesos: Aha. OK. But the function name portMappings conflicts with the field name after this refactoring. One of them has to give. Changing the field to `dockerPortMappings` now.

bernd-mesos: Hm. That breaks a lot of tests, respectively if I change the field name then runtime error messages will look different. This could be avoided by refactoring a lot of code to use the parent's function instead of the field. I suggest to leave this for another PR. There are other cleanups that can be bundled with that then. For now, renaming the parent's function to getPortMappings. No conflicts with anything this way.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -2,38 +2,61 @@ package mesosphere.marathon.api.serialization
...
   }
 
   def fromProto(proto: Protos.ExtendedContainerInfo): Container = {
-    val maybeDocker = if (proto.hasDocker) Some(DockerSerializer.fromProto(proto.getDocker)) else None
-    Container(
-      `type` = proto.getType,
-      volumes = proto.getVolumesList.asScala.map(Volume(_)).to[Seq],
-      docker = maybeDocker
-    )
+    if (proto.hasDocker) {
```
meichstedt: I understand that the docker value makes most sense to evaluate here as it is the only value serialized into the proto. Yet, we might as well deserialize a proto that this code did not serialize; the proto.getType is a required field and is not evaluated at all now. Since the proto has no meaninful validation, I think we should check the type here as well and throw an Exception if type and docker do not match.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -2,38 +2,61 @@ package mesosphere.marathon.api.serialization
 
 import mesosphere.marathon.Protos
 import mesosphere.marathon.core.externalvolume.ExternalVolumes
-import mesosphere.marathon.state.Container.Docker
 import mesosphere.marathon.state.Container.Docker.PortMapping
 import mesosphere.marathon.state._
-import org.apache.mesos
+import org.apache.mesos.{ Protos => Mesos }
```
meichstedt: I really would have liked all these import renames in a separate commit, as I am reluctant to merge those. Renaming an import is something scala allows us to, in order to avoid namespace conflicts, but we shouldn't do that if we don't really need to.

1. imo simply importing `org.apache.mesos` and using `mesos.Protos` is the obvious choice and clearer than renaming `mesos.Protos => Mesos`.
2. Splitting up the marshaller layers (`model <-> mesos.Protos` for communication to Mesos; `model <-> marathon.Protos` for persisting objects in Marathon) would have made import renames completely unneeded as each Protos object is unambiguous in its own scope.

I know reverting the renames or changing them is tedious as the IDE doesn't help you, but they're not needed to merge the required changes and I'd rather keep the change set small – unless the extra changes add value by separating concerns as described in 2).

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,75 +3,53 @@ package mesosphere.marathon.state
...
+
+  def foldDocker[R](defaultResult: R)(f: Container.Docker => R): R = {
+    this match {
+      case docker: Container.Docker => f(docker)
+      case _ => defaultResult
+    }
+  }
 
-  import Container._
+  def docker(): Option[Container.Docker] = foldDocker[Option[Container.Docker]](None)(Some(_))
```
meichstedt: Do we really need this? Will the code where this is used not be more readable if we match against the type?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -237,12 +237,12 @@ class TaskBuilder(
 
         val containerWithPortMappings = portMappings match {
           case None => c
-          case Some(newMappings) => c.copy(
-            docker = c.docker.map {
-              _.copy(portMappings = newMappings)
-            }
-          )
+          case Some(newMappings) =>
+            c.foldDocker(c)(_.copy(
```
meichstedt: Instead of mapping against the `container.docker(): Option[...]` above, it would be much more obvious to match against the container type here (then you don't need `foldDocker` btw), and provide a `def portMappings(docker: Container.Docker)` that computes the portMappings.

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,75 +3,53 @@ package mesosphere.marathon.state
...
-// TODO: trait Container and specializations?
-// Current implementation with type defaulting to DOCKER and docker to NONE makes no sense
-case class Container(
-    `type`: Mesos.ContainerInfo.Type = Mesos.ContainerInfo.Type.DOCKER,
-    volumes: Seq[Volume] = Nil,
-    docker: Option[Container.Docker] = None) {
+trait Container {
+  val volumes: Seq[Volume]
+
+  def foldDocker[R](defaultResult: R)(f: Container.Docker => R): R = {
```
meichstedt: I don't think there's a use case important enough to need this (there are 2 usages and both are better off without this).

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -108,54 +84,150 @@ object Container {
...
+  }
+
+  /**
+    * An intermediate structure that parallels the external JSON API for containers.
+    * This allows for validation of all JSON input combinations,
+    * before they have been reduced to a single outcome of type Container
+    * which would cover up too many possible error conditions.
+    * Furthrmore, we can provide trivial Formats implementation via apply/unapply.
+    */
+  case class Mask(
```
meichstedt: Did you try a solution similar to the example provided [here](http://stackoverflow.com/questions/17021847/noise-free-json-format-for-sealed-traits-with-play-2-2-library) and come up with a solution that doesn't require conversion? What's the exact problem you were facing?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -373,11 +367,11 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
     val expected = AppInfo(
       app.copy(
         versionInfo = AppDefinition.VersionInfo.OnlyVersion(clock.now()),
-        container = Some(app.container.get.copy(docker = Some(app.container.get.docker.get.copy(
+        container = Some(app.container.get.asInstanceOf[Container.Docker].copy(
```
meichstedt: No `asInstanceOf`, please. I know it's only a test and the test case is about docker, but there is a type safe solution 😬

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
```diff
@@ -142,64 +140,75 @@ class AppDefinitionTest extends MarathonSpec with Matchers {
...
     shouldNotViolate(
       app,
-      "/container/docker/portMappings",
+      "/container/portMappings",
       "Port names must be unique."
     )
 
     app = correct.copy(
-      container = Some(app.container.get.copy(docker = Some(app.container.get.docker.get.copy(
+      container = Some(app.container.get.asInstanceOf[Docker].copy(
```
meichstedt: Again, not fond of casting this

meichstedt:
@bernd-mesos Thanks for cleaning up the Container model!

I added several comments and suggestions. Please address changes in (one or multiple) additional commits to make reviewing them easier.

src/main/scala/mesosphere/marathon/api/RestResource.scala
```diff
@@ -55,22 +54,32 @@ trait RestResource {
...
+      validator(t) match {
+        case f: Failure =>
+          val entity = Json.toJson(description.map(f.withDescription).getOrElse(f)).toString
+          //scalastyle:off magic.number
+          Response.status(422).entity(entity).build()
+        //scalastyle:on
+        case Success => fn(t)
+      }
+    } catch {
+      case ValidationFailedException(validator, failure) =>
```
bernd-mesos: The reason for the try/catch is that we now have 2 validation phases and the results of the first one land here by means of the ValidationFailedException. This has nothing to do with the description issue, with AFAIU is unchanged.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
bernd-mesos: OK, I will add the extra checks

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
bernd-mesos: OK, I will spilt this into an extra PR and go with option 1.

src/main/scala/mesosphere/marathon/state/Container.scala
bernd-mesos: We don't need this, but without it you have a lot more code changes!

src/main/scala/mesosphere/marathon/state/Container.scala
bernd-mesos: Matter of taste. I will remove it.

src/main/scala/mesosphere/marathon/state/Container.scala
bernd-mesos: I don't understand how your example code relates to this. It's about Format, not Validator.

I basically copied the above code from existing code, except that we dispatch over the case class type now and not over the `type` field contents.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
bernd-mesos: Alright, but then I need to write extra code to make the test fail. I think that asInstanceOf is great for testing!

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
bernd-mesos: To carry this out, given the urgency of UC, I will eliminate all changes related to the imports and leave the pure UC-related changes. Then we can create an additional PR for the import cleanups later.

bernd-mesos:
Addressed Matthias' comments above. Thanks for the review, @meichstedt!

bernd-mesos:
Closing this request and opening a new one on branch (bernd-uc1).

PR #4027, opened 2016-06-24
And Benchmarking (and a ZK Benchmark for storage comparisons)

jasongilanfarr:
- add sbt:bench which runs benchmarks.
- a history/html graph is available at
  target/benchmarks/report/index.html

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -39,6 +39,7 @@ object MarathonBuild extends Build {
       scalaStyleSettings ++
       testSettings ++
       integrationTestSettings ++
+      benchmarkSettings ++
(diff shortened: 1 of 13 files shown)


aquamatthias:
Nice! 

The setup seems wrong: 
```
Uncaught exception when running mesosphere.marathon.benchmarks.storage.zk.ZkStorageBenchmark: java.lang.RuntimeException: failed to parse args: `false' expected but `-' found
sbt.ForkMain$ForkError: java.lang.RuntimeException: failed to parse args: `false' expected but `-' found
```

Otherwise looks good. Thanks.

PR #4029, opened 2016-06-27
Rework health checks/readiness checks effective ip:port resolution

gkleiman:
- If IP-per-Task is used, then use the task's IP address. If port mappings are
  defined used the container ports, otherwise the discovery ports.
- If IP-per-Task is not used, then use the Mesos Agent IP address. If a host
  port is requested, use it, otherwise use the container port.

cc/ @sargun


aquamatthias:
LGTM. Wait for the integration tests to pass. 
Please cherry pick in 1.2 branch

PR #4035, opened 2016-06-28
Velocity: explicitly run test and integration:test

jasongilanfarr:


jasongilanfarr:
Only major issue is this- java.lang.RuntimeException: No mesos library found. Candidates: /usr/java/packages/lib/amd64/libmesos.dylib, /usr/java/packages/lib/amd64/libmesos.so, /usr/lib64/libmesos.dylib, /usr/lib64/libmesos.so, /lib64/libmesos.dylib, /lib64/libmesos.so, /lib/libmesos.dylib, /lib/libmesos.so, /usr/lib/libmesos.dylib, /usr/lib/libmesos.so, /usr/local/lib/libmesos.dylib, /usr/local/lib/libmesos.so, /usr/lib/libmesos.dylib, /usr/lib/libmesos.so, /usr/lib64/libmesos.dylib, /usr/lib64/libmesos.so

kensipe:
simple change that LGTM however the build is failing... It seems all PRs are failing tho.  Perhaps not an issue with this PR.

jasongilanfarr:
Oh it doesn't pass. I'm not sure which mesos lib to use... It could be nice
to include the native libraries in the mesos-util jar and self-extract as
needed.

PR #4036, opened 2016-06-28
Return a 503 if the authenticator fails

gkleiman:


src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -16,15 +19,23 @@ trait AuthResource extends RestResource {
...
-      } catch {
-        case e: AccessDeniedException => withResponseFacade(authorizer.handleNotAuthorized(identity, _))
-      }
-    }.getOrElse {
-      withResponseFacade(authenticator.handleNotAuthenticated(requestWrapper, _))
+    val authenticationRequest = authenticator.authenticate(requestWrapper)
+
+    Await.ready(authenticationRequest, config.zkTimeoutDuration)
+
+    authenticationRequest.value.get match {
```
gkleiman: I am not super happy with the `.get` here, but my scala-fu failed me here... any suggestions?

jasongilanfarr:
```
try {
   Await.result(...)
} catch {
   case e: ...
}
```

?

gkleiman: I used `scala.utils.Try` instead, to avoid catching the exceptions thrown by `fn`.

jasongilanfarr: I know... If you do an Await.result instead (and do the try/catch) you can
avoid the .value.get. It's unfortunate that there's no Await.maybe that
returns a Try

jasongilanfarr:
LGTM

aquamatthias:
LGTM. Wait for the integration tests to pass.
Please cherry pick on 1.2 branch

PR #4041, opened 2016-06-28
Add Benchmarking (and a ZK Benchmark for storage comparisons)

jasongilanfarr:
- add sbt:bench which runs benchmarks.
- a history/html graph is available at
  target/benchmarks/report/index.html


aquamatthias:
LVGTM (V=very)

PR #4042, opened 2016-06-28
Generic Timeout and Retry functional transforms.

jasongilanfarr:
Future: Go find places where we use this kind of logic (there are a few)
and use the generic form (there are some exponential backoff with
Thread.sleep and no jitter, for example)


aquamatthias:
Cool. Thanks!

PR #4045, opened 2016-06-29
verizon integration test for partitioned networks

kensipe:
this integration test confirms that a test stays running after the loss of a master and ZK, then their return.  If there additional ideas on testing this, I would love the feedback.

This test runs successfully on master.  I ran this test on `releases/1.1` and it fails, confirming the marathon task_lost fix.


jasongilanfarr:
Curious, should there actually be 3 of all of the services so we can do full partition testing? (perhaps that's a different test case) - otherwise looks pretty good, pending TeamCity passing.

kensipe:
@jasongilanfarr thanks for the review!  I will add some additional tests to this on a separate PR.  In addition I plan to have system integration testing with true network partitioning.

kensipe:
All tests pass!   Merging!

PR #4048, opened 2016-06-30
Use a custom TrustManager when proxying requests

gkleiman:
Use a custom SSL trust manager when proxying requests to the current
leader. Until now Marathon was using a trust manager that contained only
Marathon's SSL certificate, which would work for self-signed
certificates, but not certificates signed by a CA.

cc/ @lloesche @jgehrcke


src/main/scala/mesosphere/marathon/io/SSLContextUtil.scala
```diff
@@ -20,19 +21,53 @@ object SSLContextUtil {
...
+      systemTmf.init(null.asInstanceOf[KeyStore])
+      getX509TrustManager(systemTmf)
+    }
+
+    val customTrustManager: X509TrustManager = new X509TrustManager() {
+      lazy val getAcceptedIssuers = keyStoreTrustManager.getAcceptedIssuers ++ systemTrustManager.getAcceptedIssuers
+
+      override def checkClientTrusted(chain: Array[X509Certificate], authType: String): Unit = {
+        try {
+          keyStoreTrustManager.checkClientTrusted(chain, authType);
```
meichstedt: do I see a semicolon?

src/main/scala/mesosphere/marathon/io/SSLContextUtil.scala
```diff
@@ -20,19 +21,53 @@ object SSLContextUtil {
...
+        } catch {
+          case e: CertificateException => systemTrustManager.checkClientTrusted(chain, authType);
+        }
+      }
+
+      override def checkServerTrusted(chain: Array[X509Certificate], authType: String): Unit = {
+        try {
+          keyStoreTrustManager.checkServerTrusted(chain, authType);
```
meichstedt: I do!

kensipe:
Way to pull this off so quickly!   I have 2 comments beyond the semicolons :)

1. it would be great to have a test with this
2. I would add a reference in the comments to: http://stackoverflow.com/questions/24555890/using-a-custom-truststore-in-java-as-well-as-the-default-one
so that the next developer to as WTH will have a reference and context.   

Nice work!

src/main/scala/mesosphere/marathon/io/SSLContextUtil.scala
gkleiman: Copying and pasting Java code can be bad ;-)

meichstedt:
👍  Thanks @gkleiman!

jgehrcke:
:rocket:

lloesche:
Thanks @gkleiman 🚀 
Always a pleasure working with the Marathon team!

PR #4049, opened 2016-06-30
Pedantic doc fix

lloesche:
Starting marathon with `--http_credentials "cptPicard:topSecretPa$$word"` will replace $$ with the current PID which is likely not the author's intention. Even though it's unlikely users will copy'paste our documentation example 1:1 they should still be correct.


gkleiman:
Thanks!

PR #4050, opened 2016-06-30
Remove unecessary match in deployments resource.

janisz:


jasongilanfarr:
Could you run this through an `sbt test` please? I believe the formatting changed such that 'else' is on the same line as the '}'

janisz:
Fixed.

gkleiman:
Thanks!

PR #4052, opened 2016-06-30
Fix style issue and enable scalastyle check

janisz:


src/main/scala/mesosphere/marathon/upgrade/DeploymentPlanReverter.scala
```diff
@@ -55,85 +55,44 @@ private[upgrade] object DeploymentPlanReverter {
...
-          result.update(
-            newGroup.id,
-            group => group.copy(dependencies = group.dependencies -- newGroup.dependencies),
-            version)
-        case _ =>
-          // still contains apps/groups, so we keep it
-          result
-      }
-    }
+sealed trait DeploymentPlanGroupChangesReverter {
```
aquamatthias: What is the reasoning behind this trait? Could we define the methods directly in the object?

aquamatthias:
@janisz Sorry for getting back so late - please rebase. One question if we can define all methods in the object directly? Otherwise looks good.

src/main/scala/mesosphere/marathon/upgrade/DeploymentPlanReverter.scala
janisz: I was thinking about separating group from app revert but yes, this trait is redundant.

ssk2:
ok to tests

aquamatthias:
@janisz Our test infrastructure is not stable, so I was not able to merge. Can you rebase again please?

jeschkies:
@janisz could you rebase once more?

jeschkies:
It seems stylecheck has been remove with 31426bd1f7b091b78748ea7fd83c497db3dfea9e

PR #4062, opened 2016-07-02
Preview: New Persistence Store using Akka Marshalling infrastructure.

jasongilanfarr:
Notes:
- May change ZkSerialized to use Proto Messages
- Need to implement compression (or force it in the client) and check
  the serialized length is less than 1MB and consider
  what _should_ happen if it takes more than 1MB, e.g. splitting
  the serialized data across two nodes.
- Add more comments to the commit (and likely docs)


project/plugins.sbt
```diff
@@ -30,3 +30,4 @@ addSbtPlugin("org.scoverage" % "sbt-coveralls" % "1.1.0")
 
 addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.1.10")
 
+addSbtPlugin("com.thoughtworks.sbt-api-mappings" % "sbt-api-mappings" % "latest.release")
```
aquamatthias: Did not know that one. Cool.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+  *       def fromStorageId(id: String): String = id
+  *   }
+  * }}}
+  *
+  * @tparam Id The marathon type of the ID for the given Value type
+  * @tparam K The persisted type of the ID
+  * @tparam V The value type being stored
+  * @tparam Serialized The serialized type of 'V' for the given [[PersistenceStore]].
+  */
+trait IdResolver[Id, K, +V, +Serialized] {
```
aquamatthias: I see your solution regarding different storage engines: the resolver is defined with respect to a specific key and serialized type - nice!

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+  *  - Provide an [[IdResolver]] for your class for the supported [[PersistenceStore]]s
+  *  - Provide a Marshaller and Unmarshaller for
+  *    your class and the Serialized form for the supported [[PersistenceStore]]s.
+  *  - For example, given a class 'A', a K set to [[mesosphere.marathon.core.storage.impl.zk.ZkId]]
+  *    and Serialized as [[mesosphere.marathon.core.storage.impl.zk.ZkSerialized]],
+  *    the following implicits should be sufficient
+  * {{{
+  *   case class A(id: Int, name: String)
+  *   object A {
+  *     implicit val zkIdResolver = new IdResolver[Int, ZkId, A, ZkSerialized] {
```
aquamatthias: nit: the example says, we should implement the IdResolver as part of the companion object. In order to use different resolver for different storage engines, the id resolver should be implemented with respect to the storage engine, or?

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+  *       def toStorageId(id: Int): ZkId =
+  *         ZkId("/A/" + id.toString) // note: scaladoc bug where string interpolation fails
+  *       def fromStorageId(key: ZkId): Int = key.value.replaceAll("/A/", "").toInt
+  *     }
+  *     implicit val zkMarshaller = Marshaller[A, ZkSerialized] =
+  *       Marshaller.opaque { (a: A) =>
+  *         val bytes = ByteString.newBuilder
```
aquamatthias: We used protobufs in the past to marshal objects to byte strings. Are you suggesting to use plain akka primitives here, or is this just an example? 
I think it would be good to use the Marshaller/Unmarshaller traits from akka but use protobuf as implementation.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+    ctx: ExecutionContext = ExecutionContext.global): Future[Option[V]] = async {
+    val path = ir.toStorageId(id)
+    val old = await(get(id))
+    change(old) match {
+      case Success(newValue) =>
+        val serialized = await(Marshal(newValue).to[Serialized])
+        old match {
+          case Some(_) =>
+            await(rawSet(path, serialized))
+            old
```
aquamatthias: Not 100% sure about the expected return value, but I would have guessed it should be `newValue` not `old`.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+          case None =>
+            await(rawCreate(path, serialized))
+            old
```
aquamatthias: `newValue` ?

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+        }
+      case Failure(error: StoreCommandFailedException) =>
+        throw error
+      case Failure(error) =>
+        old
```
aquamatthias: What is the reasoning here to ignore all errors silently?

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+    pathAsKey: String => K)(
+    implicit
+    override val mat: Materializer
+) extends PersistenceStore[K, Serialized] with StrictLogging with PrePostDriverCallback {
+
+  private[storage] val cache = Promise[TrieMap[String, Either[Serialized, Any]]]()
+
+  override def preDriverStarts: Future[Unit] = {
+    val storage = TrieMap[String, Either[Serialized, Any]]()
+    val future = store.keys().mapAsync(8) { key =>
```
aquamatthias: Hard coded value should be made configurable.
Also keep in mind all caches are created in parallel - this will result in high load on zk.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+      store.rawGet(key).map(v => key -> v)
+    }.runForeach {
+      case (key, maybeSerialized) =>
+        maybeSerialized match {
+          case Some(v) =>
+            storage.put(keyAsPath(key), Left(v))
+          case None =>
+        }
+    }
+    cache.completeWith(future.map(_ => storage))
```
aquamatthias: `preDriverStart` will be called every time we got elected (multiple times). The promise here can be completed only once. Perhaps create a new Promise for every call on `preDriverStarts`?

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+      val result = cached.keySet.withFilter(_.startsWith(storageId)).map { key =>
+        val path = if (storageId == "") storageId else key.replaceAll(s"$storageId/", "")
+        path.split("/").head
+      }
+      result
+    }
+    Source.fromFuture(future).mapConcat(_.map(k => ir.fromStorageId(pathAsKey(k)))(collection.breakOut))
+  }
+
+  override def get[Id, V](id: Id)(implicit
```
aquamatthias: In our current implementation we have special handling for versioned entries (generally only the current version is kept in memory). This implementation will also cache versioned entries. This could eventually cause problems with a lot of apps/groups and a high history number.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  override def update[Id, V](id: Id)(change: (V) => Try[V])(implicit
+    ir: IdResolver[Id, K, V, Serialized],
+    um: Unmarshaller[Serialized, V],
+    m: Marshaller[V, Serialized]): Future[V] = {
+    async {
+      await(get(id)) match {
+        case Some(old) =>
+          change(old) match {
```
aquamatthias: This change is never persisted in the store.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,254 @@
...
+  def delete[Id, V](k: Id)(implicit ir: IdResolver[Id, K, V, Serialized]): Future[Done] = {
+    rawDelete(ir.toStorageId(k))
+  }
+
+  /**
+    * @return A source of _all_ keys in the Persistence Store (which can be used by a
+    *         [[mesosphere.marathon.core.storage.impl.CachingPersistenceStore]] to populate the
+    *         cache completely on startup.
+    */
+  protected[storage] def keys(): Source[K, NotUsed]
```
aquamatthias: Perhaps divide this trait into:
- an abstract with public methods trait without any implementation
- an abstract class
This will also help with implementations (like cache) that does not need the raw* methods.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  override def postDriverTerminates: Future[Unit] = Future.successful(())
+
+  override def ids[Id, V](parent: Id)(implicit ir: IdResolver[Id, K, V, Serialized]): Source[Id, NotUsed] = {
+    val future = async {
+      val cached = await(cache.future)
+
+      val storageId = keyAsPath(ir.toStorageId(parent))
+      val result = cached.keySet.withFilter(_.startsWith(storageId)).map { key =>
+        val path = if (storageId == "") storageId else key.replaceAll(s"$storageId/", "")
```
aquamatthias: This will fail e.g. for 
``` scala
"/a/b/c/a/b/c".replaceAll("/a", "")
```
Use `replaceFirst("^($storageId/)", "")` or `substring`

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+    Source.fromFuture(future).mapConcat(_.map(k => ir.fromStorageId(pathAsKey(k)))(collection.breakOut))
+  }
+
+  override def get[Id, V](id: Id)(implicit
+    ir: IdResolver[Id, K, V, Serialized],
+    um: Unmarshaller[Serialized, V]): Future[Option[V]] = async {
+    val cached = await(cache.future)
+    val stored = cached.get(keyAsPath(ir.toStorageId(id)))
+    stored match {
+      case Some(Left(serialized)) =>
```
aquamatthias: What is the reasoning behind on demand serialization instead of doing it in preDriverStart?

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+          throw e
+      }
+    }
+  }
+
+  override def toString: String = s"CachingPersistenceStore($store)"
+
+  // None of the raw methods need to be implemented.
+
+  override protected[storage] def rawGet(id: K): Future[Option[Serialized]] = ???
```
aquamatthias: See my comment regarding splitting the PersistentStore trait.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
```diff
@@ -0,0 +1,158 @@
...
+  * the cached data is actually stale.
+  *
+  * @param store The store to cache
+  * @param keyAsPath Conversion from the Persistence layer's key to a stringified path.
+  * @param mat a materializer for akka streaming
+  * @tparam Serialized The serialized format for the persistence store.
+  */
+class CachingPersistenceStore[K, Serialized](
+    store: PersistenceStore[K, Serialized],
+    keyAsPath: K => String,
```
aquamatthias: Just thinking: If we have the assumption, that K is (or could be) a path, perhaps we should make this functionality part of IdResolver 
``` scala
isChild(parent: Id, path: Id): Boolean
```

src/main/scala/mesosphere/marathon/core/storage/impl/InMemoryPersistenceStore.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import scala.annotation.tailrec
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.Future
+
+case class RamId(id: String)
+case class Identity(value: Any)
+
+trait InMemoryStoreSerialization {
+  implicit def idResolver[V]: IdResolver[String, RamId, V, Identity] =
```
aquamatthias: Most of our Ids are not Strings. E.g. PathId, Task.Id...

src/test/scala/mesosphere/marathon/core/storage/CachingPersistenceStoreTest.scala
```diff
@@ -0,0 +1,44 @@
+package mesosphere.marathon.core.storage
```
aquamatthias: Please move this test class next to `CachingPersistenceStore` (inside impl)

src/test/scala/mesosphere/marathon/core/storage/InMemoryPersistenceStoreTest.scala
```diff
@@ -0,0 +1,17 @@
+package mesosphere.marathon.core.storage
```
aquamatthias: Please move this test class next to CachingPersistenceStore (inside impl)

src/test/scala/mesosphere/marathon/core/storage/ZkPersistenceStoreTest.scala
```diff
@@ -0,0 +1,63 @@
+package mesosphere.marathon.core.storage
```
aquamatthias: Please move this test class next to CachingPersistenceStore (inside impl/zk)

aquamatthias:
Hey @jasongilanfarr looks promising. I like the approach. I made a couple of comments, plus:
- I assume we will replace MarathonStore with PersistentStore? 
- we currently serialize changes made to one entry: https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/state/MarathonStore.scala#L43 and I fear there is code relying on this
- please also remove `mesosphere.util.state` with this PR

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
jasongilanfarr: Hmm, thought I suggested that the implicits may be defined in the companion or elsewhere. I'll make this more clear. I actually like the setup for Zk's test where its a mixin trait. I imagine the trait(s) would be mixed into something like `AppRepository`.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
jasongilanfarr: This is purely an example. I wish GitHub would actually show more of the commit body - I have a TODO: consider using `Message` for ZkPersistence

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
jasongilanfarr: Was thinking the exact same thing once I wrote CachingPersistenceStore.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
jasongilanfarr: `create` returns a `Future[Done]`. `update` returns a `Future[V]` where `V` is the old value. I probably need to comment a bit more in this method as it handles both cases.

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
jasongilanfarr: That's not _quite_ true here, and only here specifically - though I do want to add it to the log. 

If `change` fails with anything other than a `StoreCommandFailedException` (which itself doesn't exactly make a lot of sense), we don't actually want to fail the future.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
jasongilanfarr: Will move to a configuration parameter. My intent is that there is _exactly_ one PersistenceStore in all of Marathon.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
jasongilanfarr: Right, didn't think of that. I'll add comments to that method as well and do as you suggest.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
jasongilanfarr: Good catch.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
jasongilanfarr: Good catch here too. I didn't notice that we currently only cache the latest version. I'm wondering if maybe we should be Versioning into the PersistenceStore itself?

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
jasongilanfarr: Another tricky gotcha... in `preDriverStart` we have none of the implicits available to us! You'll note that we cache based on the underlying store's 'K' not 'Id'. So when we load the cache, we don't actually know how to deserialize anything until something asked for it to be deserialized. Once it is deserialized successfully, we swap it out for the real value.

src/main/scala/mesosphere/marathon/core/storage/impl/CachingPersistenceStore.scala
jasongilanfarr: good catch.

src/main/scala/mesosphere/marathon/core/storage/impl/InMemoryPersistenceStore.scala
jasongilanfarr: This was just a default resolver.

jasongilanfarr:
Thanks for all the comments on the preview!

- I intend on replacing everything up to `XYZRepository` more or less. Each repository will take in the PersistenceStore.
- I'll add per-key serialization.
- Will remove the state package in a later PR/commit
  - My intention is to have a series of commits that add the new functionality, use it, then delete the old.

jasongilanfarr:
While I've updated it, I realized versioning added a lot of complexity (and its something I'd like the persistence layer to handle itself), so its now baked into the interface.

src/main/scala/mesosphere/marathon/core/storage/IdResolver.scala
```diff
@@ -0,0 +1,41 @@
...
+  *
+  * This IdResolver must be provided for all storage engines. See [[PersistenceStore]]
+  *
+  * @tparam Id The marathon type of the ID for the given Value type
+  * @tparam K The persisted type of the ID
+  * @tparam Category The category that 'V' belongs to.
+  * @tparam V The value type being stored
+  * @tparam Serialized The serialized type of 'V' for the given [[PersistenceStore]].
+  */
+trait IdResolver[Id, K, Category, V, +Serialized] {
```
jdef: curious: what's the benefit of `+Serialized` in the type signature here?

jasongilanfarr: It shouldn't have any benefit, I've removed it.

project/build.scala
```diff
@@ -345,6 +348,7 @@ object Dependency {
   val akkaActor = "com.typesafe.akka" %% "akka-actor" % V.Akka
   val akkaSlf4j = "com.typesafe.akka" %% "akka-slf4j" % V.Akka
   val akkaStream = "com.typesafe.akka" %% "akka-stream" % V.Akka
+  val akkaHttp = "com.typesafe.akka" %% "akka-http-experimental" % V.Akka
```
jasongilanfarr: http-core is marked stable, but the higher API (which happens to be where
marshaling is) is still undergoing changes. Marshaling is quite close to
what it was in spray, so I'm not that concerned.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,145 @@
...
+  * Persistence Store that handles all marshalling and unmarshalling, allowing
+  * subclasses to focus on the raw formatted data.
+  *
+  * @tparam K The persistence store's primary key type
+  * @tparam Serialized The serialized format for the persistence store.
+  */
+abstract class BasePersistenceStore[K, Category, Serialized](implicit
+  ctx: ExecutionContext,
+  mat: Materializer) extends PersistenceStore[K, Category, Serialized]
+    with TimedPersistenceStore[K, Category, Serialized] {
```
kensipe: Metrics is always present then?  or should this be configurable?

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.zookeeper.data.Stat
+
+import scala.async.Async.{ async, await }
+import scala.collection.immutable.Seq
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.control.NonFatal
+import scala.util.{ Failure, Success }
+
+case class ZkId(category: String, id: String, version: Option[OffsetDateTime]) {
+  private val bucket = id.hashCode % 16
```
kensipe: simply beautiful.   I don't see a way to reverse this for debugging / support.  It appears we save the full ID in the bucket.  I just checked and Exhibitor doesn't provide a way of searching.   Any thoughts here?

jasongilanfarr: All objects can be found in one of their categorized buckets (there are
only 16). This facilitates even distribution of the persisted data, but
yes, it makes it a bit harder if you're looking to debug.

In order to facilitate support, we agreed that a tool should be provided to
decode/search ZK for marathon data - plus, it can decode proto into a human
readable form.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
jasongilanfarr: That was the intent yes. Operations at the lead persistence layer should be
timed. I'd like yo eventually add error tracking as well.

jasongilanfarr:
Update: I started going through and propagating PersistenceStore into the *Repository classes (as a feature flag, so they actually support both EntityStore and PersistenceStore). 

Marshaller and Unmarshaller seem to resolve correctly, but I've run into issues with IdResolver not resolving the implicits even when they are in scope. I feel like this is a variance issue - a lot of the type class implicits (e.g. JsonWriter) tend to look like this:

```
trait Writer[-T] {
  def toJson(t: t): JsValue
}
```
As such, I'm carefully going through the IdResolution type-classing to make them resolve correctly. I hope to have this a bit further along tonight and into tomorrow - I spent almost the entire day trying to figure this out. 

Finally, the one other big change I know we're going to want is to make Group serialization to  symbolically reference apps instead of serializing the entire app. I think we should do this as part of the feature flag...

kensipe:
Currently fails compilation:

```
Compiling 330 Scala sources and 3 Java sources to /Users/kensipe/projects/mesos/marathon/target/scala-2.11/classes...
[error] 
ZkPersistenceStore.scala:98: not found: value maxDelay
[error]     maxDelay
[error]     ^
[error] 
ZkPersistenceStore.scala:90: could not find implicit value for parameter mat: akka.stream.Materializer
[error]     maxDelays: FiniteDuration = Retry.DefaultMaxDelay) = this(
[error]                                                          ^
[error] two errors found
```

jasongilanfarr:
Replacing with the full branch.

PR #4067, opened 2016-07-06
Updated the validation logic for acceptedResourceRoles in AppDefinition

mpark:


mpark:
Note that this patch only updates the validation logic for `acceptedResourceRoles` for `AppDefinition`. I think this also needs to be updated for the one in `AppUpdate`. Is this correct?

gkleiman:
@mpark yes, the `AppUpdate`'s and `AppDefinition`'s validations should stay consistent.

Can you also add a test to make sure that the new validation returns user-friendly messages? See https://github.com/mesosphere/marathon/blob/d39289e3663eae0876541b0d748057b0625c753e/src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala#L710 for inspiration.

mpark:
@gkleiman Thanks, in which case I think I would much prefer to factor this out and do
`app.acceptedResourceRoles is valid(validAcceptedResourceRoles)`.

Where would be a good place to define `validAcceptedResourceRoles` so that I can access it from both `AppUpdate` and `AppDefinition`?

gkleiman:
@mpark one possibility would be to create a RoleId class and to add the validation there.

If it is a validation that applies to more fields (should we allow empty volume paths? what about paths with those blacklisted chars?), then you could add a String validator in `mesosphere.marathon.api.v2.Validation`

src/main/scala/mesosphere/marathon/state/ResourceRole.scala
```diff
@@ -1,5 +1,34 @@
...
+import com.wix.accord._
+import com.wix.accord.dsl._
+import mesosphere.marathon.api.v2.Validation._
+
 object ResourceRole {
   val Unreserved = "*"
+
+  val validResourceRole: Validator[String] = {
+    val message = "A role name must not be %s."
+    isTrue[String](message.format("\"\"")) { role => role != ""; } and
```
mpark: Not sure why the formatting is a bit funky here, but this is what `sbt test` seems to like...

src/main/scala/mesosphere/marathon/state/ResourceRole.scala
```diff
@@ -1,5 +1,34 @@
...
+      isTrue[String](message.format("\"..\"")) { role => role != ".."; } and
+      isTrue[String]("A role name must not start with a '-'.") { role => !role.startsWith("-") } and
+      validator[String] { role =>
+        role.each is valid(validResourceRoleChar)
+      }
+  }
+
+  val validResourceRoleChar: Validator[Char] = {
+    val message = "A role name must not include a %s character."
+    isTrue[Char](message.format("horizontal tab (\\x09)")) { char => char != '\u0009'; } and
```
mpark: Not sure why the formatting is a bit funky here, but this is what `sbt test` seems to like...

mpark:
Summary of the approach taken:

1. Added validation logic of a resource role to `ResourceRole` (i.e., `validResourceRole`)
     The pattern there of using `isTrue` is used to produce better error messages for the user.
     e.g., `char is notEqualTo('\u0020')` would print something like `"value is equal to  "` as opposed to
             `"A role name must not include a space (\x20) character"`.
2. Narrowed the scope of `acceptedResourceRoles` validation.
     Rather than `complyWithResourceRoleRules: Validator[AppDefinition]`,
     I've introduced `validAcceptedResourceRoles(isResident: Boolean): `Validator[Set[String]]`.
     I think the most important benefit here is that we don't need to look at `AppDefinition` as a whole,
     and only requires that `isResident` be passed. This leads to (3).
3. Since `validAcceptedResourceRoles` does not look at `AppDefinition` specifically,
    `AppUpdate` can leverage it to validate its `acceptedResourceRoles`.
    
    In general, `AppUpdate` mostly consists of fields of `AppDefinition`, but wrapped in an `Option`.
    To keep these in sync, I think it would be nice to introduce the validation logic to `AppDefinition`,
    then have `AppUpdate` use the `optional` version of it.

    For example, given a field `AppDefinition.foo`, we can introduce: `AppDefinition.validFoo`,
    then perform `appDef.foo is valid(validFoo)` within `AppDefinition`.
    Then `AppUpdate` would simply perform `appUp.foo is optional(AppDefinition.validFoo)`

@jasongilanfarr @gkleiman could you guys take another look at this patch?

src/main/scala/mesosphere/marathon/state/ResourceRole.scala
```diff
@@ -1,5 +1,38 @@
...
+    isTrue[String](message.format("\"\"")) { role => role != ""; } and
+      isTrue[String](message.format("\".\"")) { role => role != "."; } and
+      isTrue[String](message.format("\"..\"")) { role => role != ".."; } and
+      isTrue[String]("A role name must not start with a '-'.") { role => !role.startsWith("-") } and
+      validator[String] { role =>
+        role.each is valid(validResourceRoleChar)
+      }
+  }
+
+  val validResourceRoleChar: Validator[Char] = {
```
aquamatthias: Looks like you are a big fan of `isTrue` - which is ok.
Just want to show, that you can also define the validator like this:
```
  val validResourceRoleChar = validator[Char] { char =>
    char is notEqualTo('\u0009')
    char is notEqualTo('xxx`)
    char is notEqualTo('yyy`)
  }
```

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -630,10 +630,12 @@ object AppDefinition extends GeneralPurposeCombinators {
       !(app.residency.isDefined ^ app.persistentVolumes.nonEmpty)
     }
 
-  private val complyWithResourceRoleRules: Validator[AppDefinition] =
-    isTrue("""Resident apps may not define acceptedResourceRoles other than "*" (unreserved resources)""") { app =>
-      def hasResidencyCompatibleRoles = app.acceptedResourceRoles.fold(true)(_ == Set(ResourceRole.Unreserved))
-      !app.isResident || hasResidencyCompatibleRoles
+  def validAcceptedResourceRoles(isResident: Boolean): Validator[Set[String]] =
```
aquamatthias: Please move this validator also to `ResourceRole`

aquamatthias:
@mpark Thanks. LGTM. Please address one comment and I will merge.

src/main/scala/mesosphere/marathon/state/ResourceRole.scala
mpark: Yep, I appreciate suggestions like this! This particular one I already knew about, and intentionally opted for `isTrue`, as I described in the `NOTE`:

> // NOTE: The validators below use conjunction over `isTrue` in order to provide more user-friendly error messages.
> //       For example, `char is notEqualTo('\u0020')` would print something like "char is equal to  " as opposed to
> //       "A role name must not include a space (\x20) character".

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
mpark: @aquamatthias: I've moved this to `ResourceRole`. Thanks!

mpark: @aquamatthias: I've moved this validator to `ResourceRole`, and also moved the tests to a `ResourceRoleTest`. Does that look good to you?

src/main/scala/mesosphere/marathon/state/ResourceRole.scala
aquamatthias: A cool - missed the comment.

aquamatthias:
Started the integration tests 2 times now, always with this failure:
```
 ResidentTaskIntegrationTest:
[07:58:39][Step 3/3]  *** ABORTED *** (40 seconds, 5 milliseconds)
[07:58:39][Step 3/3]   java.lang.AssertionError: Waiting for clean slate in Mesos took longer than 30 seconds. Give up.
```
Don't think this has to do with this PR - just tracking.

aquamatthias:
LGTM. Thanks!

Issue #4063, opened 2016-07-04
Unsuccessful leader election - [500 Internal Server on everything Marathon related -- DCOS 1.7 on Azure]


ryankennedyio:
Hi guys,

Ran into this fun one about half an hour ago. Wasn't doing anything I hadn't done before -- I think I was changing the ports in one of the apps loaded before scaling it up.

All the docker images that were running from Marathon still seem to be up.  DCOS is showing the correct CPU/Memory allocated. Mesos was giving me 503 errors for a few minutes but has fixed itself -- showing all slaves correctly but no active tasks or frameworks shown.

Anyway, have forwarded my DCOS Master http to Localhost to view the UI.
### DC/OS Dashboard

http://localhost:8000/#/dashboard/ is fine
DCOS Dashboard -> Services shows this:
![screen shot 2016-07-04 at 3 10 08 pm](https://cloud.githubusercontent.com/assets/11158006/16551627/8edfff14-41f9-11e6-96a2-73aa8102f783.png)
### Marathon Dashboard

http://localhost:8000/service/marathon/ui/# shows this:
![screen shot 2016-07-04 at 3 10 52 pm](https://cloud.githubusercontent.com/assets/11158006/16551624/88d14a6a-41f9-11e6-9f77-8a8ca017faef.png)
### Mesos UI

Showing this -- no active tasks/frameowkrs and apparently all resources are idle.
![screen shot 2016-07-04 at 3 12 32 pm](https://cloud.githubusercontent.com/assets/11158006/16551665/f0bac11a-41f9-11e6-993d-96ee2289c48c.png)

However if I go over to `Slaves`, we see:
![screen shot 2016-07-04 at 3 14 05 pm](https://cloud.githubusercontent.com/assets/11158006/16551676/00fbb2aa-41fa-11e6-89a1-796401bdb215.png)
The "Registered" column is roughly when this 500 issue started.
### Command  Line from Master

`user@dcos-master-0:~$ dcos marathon app list`
`Error decoding response from [https://localhost/service/marathon/v2/info]: HTTP 500: Internal Server Error`

`user@dcos-master-0:~$ dcos marathon about`
`Error decoding response from [https://localhost/service/marathon/v2/info]: HTTP 500: Internal Server Error`

`user@dcos-master-0:~$ dcos package list`
`Unexpected down stream http error: GET /v2/apps 503`
#### Logs

`$ journalctl --follow -u dcos-marathon` isn't giving me anything helpful:

```
[...]

PR #4069, opened 2016-07-07
Virtual pacemaker using task reconciliation w/ fake task and agent

jdef:
Some context:
* mesos clusters may suffer from periodic network partition conditions
* scheduler.disconnected is not guaranteed to be fired under network partition conditions
* marathon wants to be able to **reliably** detect network partitions (in particular, wants to detect disconnection from the leading mesos master) and enter a disconnected state until the partition is resolved
* mesos v1 http API supports a heartbeat event to mitigate this
  * this is the preferred, long term approach
  * this approach requires a lot of work
* we need a short term solution for this problem, one that is easily backported to prior releases w/ the intent to support customers using older versions

Short Term Solution (this PR)
* treat most scheduler callbacks as heartbeat "pulse" events
* expect that a pulse event should occur within some timeout period/interval
  * missing 1 pulse event is a **skip**
  * missing X subsequent pulse events is a **failure**
* upon **skip** generate a fake explicit task reconciliation request for a bogus task/agent; the leading master should respond immediately with a TASK_LOST. we'll see this as a pulse
* upon **failure** disconnect from the master and await reconnection
* it would be nice if some of this code could be reused for the long term solution (e.g. `HeartbeatActor`)

Current design:
* `MesosHeartbeatMonitor` is a `Scheduler` **decorator**
  - generates `MessagePulse` for most callbacks
  - reacts to **skip** and **failure** conditions
* `HeartbeatActor`
  - generic implementation of a monitor that watches for pulses and reacts when they don't happen as expected (by calling back to a `Reactor` impl)
* `MarathonModules`
  - Guice wiring for instantiating the decorated `Scheduler` impl and `HeartbeatActor`
* `MarathonSchedulerService`
  - deactivates the mesos heartbeat monitor upon `stopLeadership`

New command line options:
* `--mesos_heartbeat_interval` :: in the absence of receiving a message from the mesos master during a time window of this duration, attempt to coerce mesos into communicating with marathon
* `--mesos_heartbeat_failure_threshold` :: after missing this number of expected communications from the mesos master, infer that marathon has become disconnected from the master

Task List
- [x] initial implementation of virtual heartbeat monitor
- [x] implement command line parameters that configure the heartbeat machinery
- [x] command line parameter documentation
- [x] unit testing
- integration testing
  - [x] manual
  - [ ] ?automated (shakedown?) -- planning to tackle this in a follow-up PR

Testing
* manual integration testing
to force a network partition add an `iptables` rule (on the marathon node) to drop traffic to port 5050.
``` sh
$ sudo iptables -A OUTPUT -m comment --comment "remove_me" -p tcp --destination-port 5050 -j DROP
```
to resolve the network partition remove the aforementioned `iptables` rule.
```sh
$ sudo iptables-save | grep -v "remove_me" | sudo iptables-restore
```


src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -86,6 +91,18 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
...
   }
 
+  @Named(ModuleNames.SCHEDULER_HEARTBEAT_MONITOR)
+  @Provides
+  @Singleton
+  def provideHeartbeatActor(system: ActorSystem): ActorRef = {
+    system.actorOf(HeartbeatActor.props(HeartbeatActor.Config(
+      "scheduler-hbm",
+      system,
+      FiniteDuration(15, TimeUnit.SECONDS), // TODO(jdef) conf.heartbeatTimeout.get.getOrElse(15 second),
```
jdef: TODO

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
```diff
@@ -0,0 +1,96 @@
...
+    heartbeatTimeout: FiniteDuration,
+    missedHeartbeatsThreshold: Int)
+
+  sealed trait Event
+  case object EventPulse extends Event
+  case object EventSkipped extends Event
+  case object EventDeactivate extends Event
+  case class EventActivate(heartbeatSkippedAction: Runnable, heartbeatFailureAction: Runnable) extends Event
+
+  trait StateFn extends Function2[ActorRef, Event, StateFn]
```
jdef: drop ActorRef from the parameter list here. the stateFn impls already live in the context of the actor

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -29,32 +32,66 @@ class MarathonScheduler @Inject() (
...
+      log.warn("Too many subsequent heartbeats missed; assuming disconnected from mesos master")
+
+      // for now, same logic as in disconnect() below
+      heartbeatMonitor ! HeartbeatActor.EventDeactivate
+      eventBus.publish(SchedulerDisconnectedEvent())
+      schedulerCallbacks.disconnected()
+    }
+  }
+
+  protected def heartbeatSkipped(driver: SchedulerDriver): Runnable = new Runnable {
```
jdef: make this private too?

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -86,6 +90,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
     bind(classOf[HttpEventStreamActorMetrics]).in(Scopes.SINGLETON)
   }
 
+  @Named(ModuleNames.SCHEDULER_HEARTBEAT_MONITOR)
+  @Provides
+  @Singleton
+  def provideHeartbeatActor(system: ActorSystem): ActorRef = {
+    system.actorOf(HeartbeatActor.props(HeartbeatActor.Config(
+      system,
+      FiniteDuration(15, TimeUnit.SECONDS), // TODO(jdef) conf.heartbeatTimeout.get.getOrElse(15 second),
```
jdef: TODO

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -335,4 +335,19 @@ trait MarathonConf
...
+    "mesos_heartbeat_interval",
+    descr = "(milliseconds) in the absence of receiving a message from the mesos master " +
+      "during a time window of this duration, attempt to coerce mesos into communicating with marathon.",
+    noshort = true,
+    hidden = true,
+    default = Some(15000L)) // 15 seconds
+
+  lazy val mesosHeartbeatFailureThreshold = opt[Int](
+    "mesos_heartbeat_failure_threshold",
+    descr = "(milliseconds) after missing this number of expected communications from the " +
```
jdef: remove `(milliseconds)`

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -86,6 +90,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
     bind(classOf[HttpEventStreamActorMetrics]).in(Scopes.SINGLETON)
   }
 
+  @Named(ModuleNames.SCHEDULER_HEARTBEAT_MONITOR)
```
meichstedt: In order to align with the direction of code structures: PTAL e.g. at the `mesosphere.marathon.core.launchqueue` package:
- all the newer implementation is trying to prevent wiring in Guice: you could add a `HeartbeatModule` and use a distinct config similar to `LaunchQueueConfig` to separate wiring from config (and have a nice place to specify the default config values)
- you can/should use the `LeadershipModule.startWhenLeader` to make sure this actor is started/suspended depending on leadership status.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -86,6 +90,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
...
+  @Provides
+  @Singleton
+  def provideHeartbeatActor(system: ActorSystem): ActorRef = {
+    system.actorOf(HeartbeatActor.props(HeartbeatActor.Config(
+      system,
+      FiniteDuration(conf.mesosHeartbeatInterval.get.getOrElse(15000L), TimeUnit.MILLISECONDS),
+      conf.mesosHeartbeatFailureThreshold.get.getOrElse(5)
+    )), "SchedulerHeartbeatMonitor")
```
meichstedt: `mesosHeartbeatInterval` vs `SchedulerHeartbeatMonitor` – please pick one prefix, I'm fine with both.

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
```diff
@@ -0,0 +1,81 @@
...
+}
+
+object HeartbeatActor {
+
+  case class Config(
+    system: ActorSystem,
+    heartbeatTimeout: FiniteDuration,
+    missedHeartbeatsThreshold: Int)
+
+  sealed trait Event
```
meichstedt: In Akka tongue, _messages_ are sent to/from actors

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
```diff
@@ -0,0 +1,81 @@
...
+  case object EventPulse extends Event
```
meichstedt: In Marathon codebase, messages are usually not pre- or suffixed. Following that, you could name the message `Pulse` – yet I like that it extends a trait (although I'd rather name that `Message`).

That said, we have no convention on the naming of messages an actor _handles_ vs messages it _sends_ (probably to itself). I could think of traits `Incoming` and `Outgoing`.

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
```diff
@@ -0,0 +1,81 @@
...
+import akka.event.LoggingReceive
+import scala.concurrent.duration._
+
+// HeartbeatActor monitors the heartbeat of some process and executes handlers for various conditions.
+// If an expected heartbeat is missed then execute an `onSkipped` handler.
+// If X number of subsequent heartbeats are missed then execute an `onFailure` handler.
+// Upon creation the actor is in an inactive state and must be sent an EventActivate message to activate.
+// Once activated the actor will monitor for EventPulse messages (these are the heartbeats).
+// The actor may be deactivated by sending it an EventDeactivate message.
+class HeartbeatActor @Inject() (
```
meichstedt: What do you need the `@Inject()` for? The `def props` should be sufficient.

jdef:
wondering if activation of the heartbeat monitor should really happen somewhere else, perhaps attached to driver start/stop vs. scheduler callback (re)registered/disconnected? as it stands now, the driver could be started, the network could become partitioned before registration message is sent from mesos, and the heartbeat monitor won't detect that. it's probably not be an issue in practice because as long as marathon isn't initially registered it won't act as if it's connected (so its expectations are already correctly set).

src/main/scala/mesosphere/marathon/MarathonModule.scala
jdef: yes exactly. the existing scheduler/guice wiring is complicated and i didn't want to unravel all of that in this PR, especially since one of the goals is to make this easy to backport to earlier releases.

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
jdef: pretty sure that the first commit's integration test failed before I added this `Inject`. the logs should still be on TC and I haven't yet squashed the commits, so you should be able to get to it pretty easily.

src/main/scala/mesosphere/marathon/MarathonModule.scala
jdef: ack

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -29,32 +32,66 @@ class MarathonScheduler @Inject() (
...
+  // to use the new mesos v1 http API.
+  private[this] val virtualHeartbeatTasks: java.util.Collection[TaskStatus] = Seq(
+    TaskStatus.newBuilder
+    .setTaskId(TaskID.newBuilder.setValue("fake-marathon-pacemaker-task-" + UUID.randomUUID().toString))
+    .setState(TaskState.TASK_LOST) // required, so we just need to set something
+    .setSlaveId(SlaveID.newBuilder.setValue("fake-marathon-pacemaker-agent-" + UUID.randomUUID().toString))
+    .build
+  ).asJava
+
+  private[this] val heartbeatFailure = new Runnable {
```
jasongilanfarr: I still think this should just be `() => { body }`, likewise for `heartbeatSkipped`

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
```diff
@@ -0,0 +1,81 @@
+package mesosphere.util.monitor
+
+import javax.inject.Inject
+
+import akka.actor._
+import akka.event.LoggingReceive
+import scala.concurrent.duration._
+
+// HeartbeatActor monitors the heartbeat of some process and executes handlers for various conditions.
```
jasongilanfarr: Please use scaladocs here (also add a unit test for this as well). Perhaps this class should be in the scheduler package, not just "util"? Its pretty specific to marathon and the scheduler/monitoring.

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -29,32 +32,66 @@ class MarathonScheduler @Inject() (
...
   import scala.concurrent.ExecutionContext.Implicits.global
 
   implicit val zkTimeout = config.zkTimeoutDuration
 
+  // virtualHeartbeatTasks is sent in a reconciliation message to mesos in order to force a
+  // predictable response: the master (if we're connected) will send back a TASK_LOST because
+  // the fake task ID and agent ID that we use will never actually exist in the cluster.
+  // this is part of a short-term workaround: will no longer be needed once marathon is ported
+  // to use the new mesos v1 http API.
+  private[this] val virtualHeartbeatTasks: java.util.Collection[TaskStatus] = Seq(
```
meichstedt: I'd suggest to put all the specific logic into the HeartbeatActor. If I'm correct, the SchedulerCallbacks actually only need to send pulses to the Actor. See my notes on the actor.

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
```diff
@@ -0,0 +1,81 @@
...
+// The actor may be deactivated by sending it an EventDeactivate message.
+class HeartbeatActor @Inject() (
+    config: HeartbeatActor.Config
+) extends Actor {
+
+  import HeartbeatActor._
+
+  def receive: Receive = stateInactive
+
+  def stateInactive: Receive = LoggingReceive.withLabel("inactive") {
```
meichstedt: I'd change this slightly: the actor could internally
- become active when receiving a pulse
- react to each subsequent pulse with staying in the active state
- become flatlined after the configured duration and trigger a heartbeat reconciliation during that transition after each interval (increasing a counter)
- become active if it receives a pulse while flatlining
- die after having reached the max count and call `driver.disconnected()`

I think the separation would be a little cleaner as the scheduler only sends messages to some ActorRef, while the Actor internally has all the state and heartbeat logic. Does that make sense to you? The implementation would be a nice fit for the [Akka FSM](http://doc.akka.io/docs/akka/2.4.7/scala/fsm.html) dsl as that makes defining state timeouts and state transitions really obvious. I'd be happy to explain that in detail if you want.

meichstedt:
Thanks for the poc, @jdef! I think the approach is well suited as a short term solution. Please see my comments above.

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
meichstedt: ack, the case class should just use higher order functions – yet I'm not sure whether the callbacks are actually needed (see my take on the Actor driving this on its own by calling the scheduler driver)

jdef: I think you're suggesting this, but it doesn't work:
```
scala> case class A(b: => Unit)
<console>:1: error: `val' parameters may not be call-by-name
case class A(b: => Unit)
                ^
```

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
jdef: ack, unit test is already on the TODO list (see top of the PR).

we don't actually have a scheduler package. we have a bunch of scheduler-related stuff in mesosphere/marathon but i'd like to avoid polluting that more than necessary.

@jasongilanfarr you and @meichstedt are making similar points re: consolidating/moving the handler action implementations into this actor. I still don't follow the rationale for that, I'm wondering if I'm missing something obvious or part of a scala philosophy that's just not clicking for me. To me that would seem to take something that's mesos-scheduler-agnostic (a generic heartbeat monitor) and tightly couple it to our current scheduler/scheduler-driver callback implementation. I'd like to better understand the motivation for doing such. As it currently stands the heartbeat actor is a reusable component that does one generic thing. It's also testable outside of any scheduler abstraction. Why complicate it further?

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
jdef: see my comment https://github.com/mesosphere/marathon/pull/4069#discussion_r69972382

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
kensipe: my preference would be in the marathon package structure vs util

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
jdef: I actually found the akka FSM over complicated for this. But now that I've restructured things I'll take another look at it. I sure do miss the simplicity of other langs...

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
jasongilanfarr: There's kind of two pieces here: one is to put it in the marathon package structure, the second is that it seems to be pretty specific to mesos scheduler's requirements for heartbeating, no? If not, maybe it should be used in the health check infrastructure, for example? If the former, I support the tight coupling.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -335,4 +335,19 @@ trait MarathonConf
     default = Some(500)
   )
 
+  lazy val mesosHeartbeatInterval = opt[Long](
+    "mesos_heartbeat_interval",
+    descr = "(milliseconds) in the absence of receiving a message from the mesos master " +
+      "during a time window of this duration, attempt to coerce mesos into communicating with marathon.",
+    noshort = true,
+    hidden = true,
+    default = Some(15000L)) // 15 seconds
```
jdef: extract constants

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -86,6 +98,17 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
     bind(classOf[HttpEventStreamActorMetrics]).in(Scopes.SINGLETON)
   }
 
+  @Named(ModuleNames.MESOS_HEARTBEAT_MONITOR)
+  @Provides
+  @Singleton
+  def provideHeartbeatActor(system: ActorSystem): ActorRef = {
+    system.actorOf(HeartbeatActor.props(HeartbeatActor.Config(
+      system,
+      FiniteDuration(conf.mesosHeartbeatInterval.get.getOrElse(15000L), TimeUnit.MILLISECONDS),
```
jdef: extract constants

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
```diff
@@ -0,0 +1,115 @@
...
+import mesosphere.util.monitor._
+import org.apache.mesos.{ Scheduler, SchedulerDriver }
+import org.apache.mesos.Protos._
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConverters._
+
+class MesosHeartbeatMonitor @Inject() (
+    @Named(MesosHeartbeatMonitor.BASE) scheduler: Scheduler,
+    @Named(ModuleNames.MESOS_HEARTBEAT_MONITOR) heartbeatMonitor: ActorRef
```
jdef: s/heartbeatMonitor/heartbeatActor/

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
```diff
@@ -0,0 +1,115 @@
...
+    driver: SchedulerDriver,
+    executor: ExecutorID,
+    slave: SlaveID,
+    message: Array[Byte]): Unit = {
+    heartbeatMonitor ! HeartbeatActor.MessagePulse
+    scheduler.frameworkMessage(driver, executor, slave, message)
+  }
+
+  override def disconnected(driver: SchedulerDriver): Unit = {
+    // heartbeat monitor may have (transiently) triggered this, but that's ok because if it did then
```
jdef: s/heartbeat monitor/reactor/ (and no so transiently)

src/main/scala/mesosphere/util/monitor/HeartbeatMonitor.scala
jdef: ack. it's designed with scheduler heartbeat req's in mind. but there's no scheduler specific glue in here ATM. we'll want some kind of heartbeat monitoring once we switch to the v1 API as well. the bits that integrate with the mesos scheduler may look much different. these more generic monitor bits (as they currently stand) probably won't. which, to me, is an argument to keep things separate.

moving to a separate package is fine. i really don't like util either. i cry a little bit inside thinking about all the boilerplate/overhead required to use the service/actor/impl/etc/etc "core" packaging structure that we've established for bigger modules -- I only want a simple actor after all.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -54,6 +57,8 @@ object ModuleNames {
   final val STORE_GROUP = "GroupStore"
   final val STORE_TASK = "TaskStore"
   final val STORE_EVENT_SUBSCRIBERS = "EventSubscriberStore"
+
+  final val MESOS_HEARTBEAT_MONITOR = "MesosHeartbeatActor"
```
jdef: naming is hard. should probably call this MESOS_HEARTBEAT_ACTOR

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
```diff
@@ -0,0 +1,129 @@
...
+    private[this] val virtualHeartbeatTasks: java.util.Collection[TaskStatus] = Seq(
+      TaskStatus.newBuilder
+      .setTaskId(TaskID.newBuilder.setValue("fake-marathon-pacemaker-task-" + UUID.randomUUID().toString))
+      .setState(TaskState.TASK_LOST) // required, so we just need to set something
+      .setSlaveId(SlaveID.newBuilder.setValue("fake-marathon-pacemaker-agent-" + UUID.randomUUID().toString))
+      .build
+    ).asJava
+
+    override def onSkip(): Unit = {
+      driver.reconcileTasks(virtualHeartbeatTasks)
```
jdef: log.debug here?

jdef:
rebased to latest master

src/test/scala/mesosphere/marathon/util/heartbeat/HeartbeatActorTest.scala
```diff
@@ -0,0 +1,121 @@
...
+      val fsm = TestFSMRef(new HeartbeatActor(fakeConfig))
+      fsm.setState(stateName = StateActive, stateData = DataActive(FakeReactor, FakeSessionToken, missed = 1))
+      fsm ! MessagePulse
+      fsm.stateName should be (StateActive)
+      fsm.stateData should be (DataActive(FakeReactor, FakeSessionToken))
+    }
+
+    "reactivate when active, upon receipt of a valid activation message" in {
+      val fsm = TestFSMRef(new HeartbeatActor(fakeConfig))
+      fsm.setState(stateName = StateActive, stateData = DataActive(FakeReactor, FakeSessionToken))
```
jdef: set missed = 1

src/test/scala/mesosphere/marathon/util/heartbeat/HeartbeatActorTest.scala
```diff
@@ -0,0 +1,121 @@
...
+      fsm ! MessageDeactivate(FakeSessionToken)
+      fsm.stateName should be (StateInactive)
+      fsm.stateData should be (DataNone)
+    }
+
+    "do NOT deactivate when active, upon receipt of an invalid deactivation message" in {
+      val fsm = TestFSMRef(new HeartbeatActor(fakeConfig))
+      fsm.setState(stateName = StateActive, stateData = DataActive(FakeReactor, FakeSessionToken))
```
jdef: set missed = 1

jdef:
manual integration testing yields troubling results. upon invoking `disconnect` for missing too many heartbeats a single marathon instance (HA or standalone mode) fails to reobtain leadership after the network partition is resolved.

logs:
```
[2016-07-12 18:39:00,344] WARN Too many subsequent heartbeats missed; inferring disconnected from mesos master (mesosphere.marathon.MesosHeartbeatMonitor$$EnhancerByGuice$$c191120b:marathon-akka.actor.default-disp
atcher-7)
[2016-07-12 18:39:00,345] WARN Disconnected (mesosphere.marathon.MarathonScheduler$$EnhancerByGuice$$9539aab2:marathon-akka.actor.default-dispatcher-7)
[2016-07-12 18:39:00,347] INFO Abdicating leadership while leading (reoffer=false) (mesosphere.marathon.core.election.impl.CuratorElectionService:marathon-akka.actor.default-dispatcher-7)
[2016-07-12 18:39:00,353] INFO Defeated (LeaderLatchListener Interface). New leader: - (mesosphere.marathon.core.election.impl.CuratorElectionService:marathon-akka.actor.default-dispatcher-7)
[2016-07-12 18:39:00,357] INFO Lost leadership (mesosphere.marathon.MarathonSchedulerService$$EnhancerByGuice$$89824ffd:marathon-akka.actor.default-dispatcher-7)
[2016-07-12 18:39:00,359] INFO All actors suspended:
* Actor[akka://marathon/user/killOverdueStagedTasks#-1899497586]
* Actor[akka://marathon/user/rateLimiter#1457533469]
* Actor[akka://marathon/user/offerMatcherLaunchTokens#1758026608]
* Actor[akka://marathon/user/offersWantedForReconciliation#-1880030536]
* Actor[akka://marathon/user/reviveOffersWhenWanted#-518171829]
* Actor[akka://marathon/user/taskTracker#1576535952]
* Actor[akka://marathon/user/offerMatcherManager#2035485922]
* Actor[akka://marathon/user/expungeOverdueLostTasks#-1404564564]
* Actor[akka://marathon/user/launchQueue#1501413190] (mesosphere.marathon.core.leadership.impl.LeadershipCoordinatorActor:marathon-akka.actor.default-dispatcher-16)
I0712 18:39:00.361474 31487 sched.cpp:1949] Asked to stop the driver
[2016-07-12 18:39:00,360] INFO Stopping driver (mesosphere.marathon.MarathonSchedulerService$$EnhancerByGuice$$89824ffd:marathon-akka.actor.default-dispatcher-7)
I0712 18:39:00.363247 31532 sched.cpp:1152] Stopping framework 'ab0b8f17-be54-4894-9dd2-4c772a4a5203-0001'
[2016-07-12 18:39:00,363] INFO Will offer leadership after 500 milliseconds backoff (mesosphere.marathon.core.election.impl.CuratorElectionService:marathon-akka.actor.default-dispatcher-7)
[2016-07-12 18:39:00,365] INFO Driver future completed with result=Success(()). (mesosphere.marathon.MarathonSchedulerService$$EnhancerByGuice$$89824ffd:ForkJoinPool-2-worker-5)
[2016-07-12 18:39:00,365] INFO ExpungeOverdueLostTasksActor has stopped (mesosphere.marathon.core.task.jobs.impl.ExpungeOverdueLostTasksActor:marathon-akka.actor.default-dispatcher-18)
[2016-07-12 18:39:00,365] INFO Canceling leadership offer waiting for backoff (reoffer=true) (mesosphere.marathon.core.election.impl.CuratorElectionService:ForkJoinPool-2-worker-5)
[2016-07-12 18:39:00,365] INFO Now standing by. Closing existing handles and rejecting new. (mesosphere.marathon.event.http.HttpEventStreamActor:marathon-akka.actor.default-dispatcher-18)
[2016-07-12 18:39:00,365] INFO Call postDriverRuns callbacks on EntityStoreCache(MarathonStore(app:)), EntityStoreCache(MarathonStore(group:)), EntityStoreCache(MarathonStore(deployment:)), EntityStoreCache(Marath
onStore(framework:)), EntityStoreCache(MarathonStore(taskFailure:)), EntityStoreCache(MarathonStore(events:)) (mesosphere.marathon.MarathonSchedulerService$$EnhancerByGuice$$89824ffd:ForkJoinPool-2-worker-5)
[2016-07-12 18:39:00,366] INFO Suspending scheduler actor (mesosphere.marathon.MarathonSchedulerActor:marathon-akka.actor.default-dispatcher-16)
[2016-07-12 18:39:00,370] INFO Finished postDriverRuns callbacks (mesosphere.marathon.MarathonSchedulerService$$EnhancerByGuice$$89824ffd:ForkJoinPool-2-worker-5)
[2016-07-12 18:39:00,874] ERROR Will reoffer leadership after abdicating (mesosphere.marathon.core.election.impl.CuratorElectionService:ForkJoinPool-2-worker-5)
```

followed by repeating blocks of (approx every minute):
```
[2016-07-12 18:39:01,221] INFO Do not proxy to myself. Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:01,221] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:01,472] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:01,722] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:01,973] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:02,223] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:02,473] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:02,724] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:02,974] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:03,224] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:03,475] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:03,725] INFO Waiting for consistent leadership state. Are we leader?: false, leader: None (mesosphere.marathon.api.LeaderProxyFilter$:qtp1391446933-70)
[2016-07-12 18:39:03,979] INFO 0:0:0:0:0:0:0:1 - - [12/Jul/2016:18:39:01 +0000] "GET //localhost:8080/v2/tasks HTTP/1.1" 503 353 "-" "curl/7.35.0" 275 (mesosphere.chaos.http.ChaosRequestLog$$EnhancerByGuice$$f945c131:qtp1391446933-70)
```

jdef:
#4063

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
```diff
@@ -0,0 +1,160 @@
...
+    // virtualHeartbeatTasks is sent in a reconciliation message to mesos in order to force a
+    // predictable response: the master (if we're connected) will send back a TASK_LOST because
+    // the fake task ID and agent ID that we use will never actually exist in the cluster.
+    // this is part of a short-term workaround: will no longer be needed once marathon is ported
+    // to use the new mesos v1 http API.
+    private[this] val virtualHeartbeatTasks: java.util.Collection[TaskStatus] = Seq(fakeHeartbeatStatus).asJava
+
+    override def onSkip(): Unit = {
+      log.debug("Prompting mesos for a heartbeat via explicit task reconciliation")
+      driver.reconcileTasks(virtualHeartbeatTasks)
```
aquamatthias: I remember situations, where sending an unknown task to Mesos will trigger an implicit reconciliation for all tasks of the framework - not sure if this is the case here. 
@neilconway can you comment on that? Do you have another suggestion to "force" a response?

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
```diff
@@ -0,0 +1,160 @@
...
+      disconnected(driver)
+    }
+  }
+
+  override def registered(
+    driver: SchedulerDriver,
+    frameworkId: FrameworkID,
+    master: MasterInfo): Unit = {
+    log.debug("registered heartbeat monitor")
+    heartbeatActor ! Heartbeat.MessageActivate(heartbeatReactor(driver), sessionOf(driver))
```
aquamatthias: Perhaps it is easier to start a new HeartbeatActor here (stopped in reregister), than to reuse a pre-wired one?

src/main/scala/mesosphere/marathon/util/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
...
+// Upon creation the actor is in an inactive state and must be sent a MessageActivate message to activate.
+// Once activated the actor will monitor for MessagePulse messages (these are the heartbeats).
+// The actor may be forcefully deactivated by sending it an MessageDeactivate message.
+class HeartbeatActor(config: Heartbeat.Config) extends LoggingFSM[HeartbeatInternal.State, HeartbeatInternal.Data] {
+  import Heartbeat._
+  import HeartbeatInternal._
+
+  startWith(StateInactive, DataNone)
+
+  when(StateInactive) {
```
aquamatthias: If you create the actor when the driver is created, you don't need this state.

src/main/scala/mesosphere/marathon/util/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
+package mesosphere.marathon.util.heartbeat
```
aquamatthias: Feels like `MesosHeartbeatMonitor` and `HeartbeatActor` have a close relationship and should live in the same package. We try to make components under `core` perhaps `core/heartbeat`?

src/main/scala/mesosphere/marathon/util/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  when(StateInactive) {
+    case Event(MessageActivate(reactor, token), DataNone) =>
+      log.debug("heartbeat activated")
+      goto(StateActive) using DataActive(reactor, token)
+    case _ =>
+      stay // swallow all other event types
+  }
+
+  when(StateActive, stateTimeout = config.heartbeatTimeout) {
```
aquamatthias: I like FSM. But if you create the actor when the driver is created, this is the only state left. Then you could make this a simple actor which maintains a counter.

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
neilconway: @aquamatthias If a framework tries to kill an unknown task, the master will automatically trigger an implicit reconciliation for all of the framework's tasks. That seems like weird behavior to me, not sure why we do that.

If you want to trigger a response from Mesos, doing explicit reconciliation for an unknown agent/task ID seems fine to me -- you will get TASK_LOST (or TASK_UNKNOWN in the future), but won't trigger any automatic reconciliation.

src/main/scala/mesosphere/marathon/util/heartbeat/HeartbeatActor.scala
jdef: The heartbeat actor really doesn't have any scheduler integration code, it's completely generic - so that's why it lives where it currently does. `MesosHeartbeatMonitor` is a `Scheduler` implementation and so .. since we don't have a proper `core` package for scheduler stuff yet, I just put it next to the other scheduler stuff.

You're the 4th person to suggest that I couple these classes more closely. Since it seems intuitive to the majority ... I'll create `core/heartbeat` to house these.

src/main/scala/mesosphere/marathon/util/heartbeat/HeartbeatActor.scala
jdef: Will consider this approach

src/main/scala/mesosphere/marathon/MesosHeartbeatMonitor.scala
jdef: In terms of "ease" -- this is pretty easy. Controlling multiple heartbeat actors seems more complicated vs. interacting with a single one.

src/main/scala/mesosphere/marathon/util/heartbeat/HeartbeatActor.scala
jdef: so moved

src/main/scala/mesosphere/marathon/core/heartbeat/MesosHeartbeatMonitor.scala
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  override def statusUpdate(driver: SchedulerDriver, status: TaskStatus): Unit = {
+    heartbeatActor ! Heartbeat.MessagePulse
+
+    // filter TASK-LOST updates for our fake tasks to avoid cluttering the logs
+    if (!isFakeHeartbeatUpdate(status)) {
+      scheduler.statusUpdate(driver, status)
+    } else {
+      log.info("received fake heartbeat task-status update")
```
jdef: is this too noisy? should we reduce the verbosity here to `debug`?

jdef:
@meichstedt PTAL

jdef:
Reproduced failing integration test locally, via `sbt 'integration:testOnly mesosphere.marathon.integration.NetworkPartitionIntegrationTest'`

Relevant logs indicate something still amiss with marathon leader election:
```
2016-07-14 17:45:30,815] INFO [] zookeeper: [2016-07-14 17:45:30,815] INFO [] Accepted socket connection from /127.0.0.1:60268
[2016-07-14 17:45:30,815] INFO [] marathon_33896: [2016-07-14 17:45:30,814] INFO [] Opening socket connection to server 127.0.0.1/127.0.0.1:37725. Will not attempt to authenticate using SASL (unknown error)
[2016-07-14 17:45:30,815] INFO [] marathon_33896: [2016-07-14 17:45:30,815] INFO [] Socket connection established to 127.0.0.1/127.0.0.1:37725, initiating session
[2016-07-14 17:45:30,817] INFO [] zookeeper: [2016-07-14 17:45:30,817] INFO [] Client attempting to renew session 0x155ea8335b50006 at /127.0.0.1:60268
[2016-07-14 17:45:30,820] INFO [] zookeeper: [2016-07-14 17:45:30,820] INFO [] Established session 0x155ea8335b50006 with negotiated timeout 10000 for client /127.0.0.1:60268
[2016-07-14 17:45:30,823] INFO [] marathon_33896: [2016-07-14 17:45:30,823] INFO [] Session establishment complete on server 127.0.0.1/127.0.0.1:37725, sessionid = 0x155ea8335b50006, negotiated timeout = 10000
[2016-07-14 17:45:30,823] INFO [] marathon_33896: [2016-07-14 17:45:30,823] INFO [] State change: RECONNECTED
[2016-07-14 17:45:30,841] INFO [] marathon_33896: [2016-07-14 17:45:30,841] INFO [] Defeated (LeaderLatchListener Interface). New leader: node-1:33896
[2016-07-14 17:45:30,847] INFO [] marathon_33896: [2016-07-14 17:45:30,846] INFO [] Lost leadership
[2016-07-14 17:45:30,848] INFO [] marathon_33896: [2016-07-14 17:45:30,848] INFO [akka://marathon/user/leaderShipCoordinator] All actors suspended:
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/taskTracker#-889198712]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/expungeOverdueLostTasks#719056488]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/offerMatcherManager#974787931]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/killOverdueStagedTasks#1142051975]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/reviveOffersWhenWanted#-447111798]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/launchQueue#-155891661]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/offerMatcherLaunchTokens#1878562387]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/rateLimiter#-1156584663]
[2016-07-14 17:45:30,848] INFO [] marathon_33896: * Actor[akka://marathon/user/offersWantedForReconciliation#-714912465]
[2016-07-14 17:45:30,850] INFO [] marathon_33896: [2016-07-14 17:45:30,850] INFO [] Stopping driver
[2016-07-14 17:45:30,850] INFO [] marathon_33896: I0714 17:45:30.850613 22220 sched.cpp:1949] Asked to stop the driver
[2016-07-14 17:45:30,851] INFO [] marathon_33896: I0714 17:45:30.850899 22243 sched.cpp:1152] Stopping framework '25c82df8-c356-4bee-a1fe-dd7e092cf6d3-0000'
[2016-07-14 17:45:30,851] INFO [] marathon_33896: [2016-07-14 17:45:30,851] INFO [] Will offer leadership after 500 milliseconds backoff
[2016-07-14 17:45:30,851] INFO [] marathon_33896: [2016-07-14 17:45:30,851] INFO [] Driver future completed with result=Success(()).
[2016-07-14 17:45:30,852] INFO [] marathon_33896: [2016-07-14 17:45:30,851] INFO [] Now standing by. Closing existing handles and rejecting new.
[2016-07-14 17:45:30,853] INFO [] marathon_33896: [2016-07-14 17:45:30,852] INFO [] Canceling leadership offer waiting for backoff (reoffer=true)
[2016-07-14 17:45:30,853] INFO [] marathon_33896: [2016-07-14 17:45:30,852] INFO [akka://marathon/user/MarathonScheduler/$a] Suspending scheduler actor
[2016-07-14 17:45:30,855] INFO [] marathon_33896: [2016-07-14 17:45:30,853] INFO [] Call postDriverRuns callbacks on EntityStoreCache(MarathonStore(app:)), EntityStoreCache(MarathonStore(group:)), EntityStoreCache(MarathonStore(deployment:)), EntityStoreCache(MarathonStore(framework:)), EntityStoreCache(MarathonStore(taskFailure:)), EntityStoreCache(MarathonStore(events:))
[2016-07-14 17:45:30,855] INFO [] marathon_33896: [2016-07-14 17:45:30,853] INFO [] Not leader therefore not scaling apps
[2016-07-14 17:45:30,860] INFO [] marathon_33896: [2016-07-14 17:45:30,859] INFO [] Finished postDriverRuns callbacks
[2016-07-14 17:45:30,860] INFO [] zookeeper: [2016-07-14 17:45:30,858] INFO [] Creating new log file: log.26
[2016-07-14 17:45:30,866] INFO [] marathon_33896: [2016-07-14 17:45:30,865] INFO [akka://marathon/user/expungeOverdueLostTasks/1] ExpungeOverdueLostTasksActor has stopped
[2016-07-14 17:45:30,866] INFO [] marathon_33896: [2016-07-14 17:45:30,866] INFO [akka://marathon/user/launchQueue/1/0-marathonintegrationtest_app] Deregister as matcher.
[2016-07-14 17:45:30,866] INFO [] marathon_33896: [2016-07-14 17:45:30,866] INFO [akka://marathon/user/launchQueue/1/0-marathonintegrationtest_app] Stopped appTaskLaunchActor for /marathonintegrationtest/app version 2016-07-14T17:45:23.869Z
[2016-07-14 17:45:30,872] INFO [] marathon_33896: [2016-07-14 17:45:30,872] INFO [] Elected (LeaderLatchListener Interface)
[2016-07-14 17:45:30,872] INFO [] marathon_33896: [2016-07-14 17:45:30,872] INFO [] Became leader and abdicating immediately
[2016-07-14 17:45:31,095] INFO [] marathon_33896: [2016-07-14 17:45:31,095] INFO [] Opening socket connection to server 127.0.0.1/127.0.0.1:37725. Will not attempt to authenticate using SASL (unknown error)
[2016-07-14 17:45:31,095] INFO [] marathon_33896: [2016-07-14 17:45:31,095] INFO [] Socket connection established to 127.0.0.1/127.0.0.1:37725, initiating session
[2016-07-14 17:45:31,096] INFO [] zookeeper: [2016-07-14 17:45:31,096] INFO [] Accepted socket connection from /127.0.0.1:60269
[2016-07-14 17:45:31,096] INFO [] zookeeper: [2016-07-14 17:45:31,096] INFO [] Client attempting to renew session 0x155ea8335b50007 at /127.0.0.1:60269
[2016-07-14 17:45:31,097] INFO [] zookeeper: [2016-07-14 17:45:31,096] INFO [] Established session 0x155ea8335b50007 with negotiated timeout 10000 for client /127.0.0.1:60269
[2016-07-14 17:45:31,097] INFO [] marathon_33896: [2016-07-14 17:45:31,097] INFO [] Session establishment complete on server 127.0.0.1/127.0.0.1:37725, sessionid = 0x155ea8335b50007, negotiated timeout = 10000
[2016-07-14 17:45:31,363] INFO [] marathon_33896: [2016-07-14 17:45:31,362] ERROR [] Will reoffer leadership after abdicating
[2016-07-14 17:45:32,745] INFO [] marathon_33896: 2016-07-14 17:45:32,745:22168(0x7efba9da9700):ZOO_WARN@zookeeper_interest@1570: Exceeded deadline by 3051ms
[2016-07-14 17:45:32,745] INFO [] marathon_33896: 2016-07-14 17:45:32,745:22168(0x7efba9da9700):ZOO_INFO@check_events@1728: initiated connection to server [127.0.0.1:37725]
```

src/main/scala/mesosphere/marathon/core/heartbeat/MesosHeartbeatMonitor.scala
meichstedt: imo debug would be enough. for long running services we expect only few updates (unless there are mesos health checks reporting stuff anyways)

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
+package mesosphere.marathon.core.heartbeat
+
+import akka.actor._
+import scala.concurrent.duration._
+
+// HeartbeatActor monitors the heartbeat of some process and executes handlers for various conditions.
```
meichstedt: Is there a reason why you didn't use scaladoc syntax? not that we're consistent with that ..

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
...
+      stay // swallow all other event types
+  }
+
+  when(StateActive, stateTimeout = config.heartbeatTimeout) {
+    case Event(MessagePulse, data: DataActive) =>
+      stay using data.copy(missed = 0)
+
+    case Event(StateTimeout, data: DataActive) =>
+      if (data.missed + 1 >= config.missedHeartbeatsThreshold) {
+        data.reactor.onFailure
```
meichstedt: I'd put a `log.warning` here.

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
...
+        goto(StateInactive) using DataNone
+      } else {
+        data.reactor.onSkip
```
meichstedt: And probably a `log.info` here?

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  trait Reactor {
+    def onSkip(): Unit
+    def onFailure(): Unit
+  }
+
+  def props(config: Config): Props = Props(classOf[HeartbeatActor], config)
+}
+
+protected[heartbeat] object HeartbeatInternal {
```
meichstedt: A little unusual to add an extra _private_ object here, but I actually like it: you put _protocol_ and _internal_ stuff into separate scopes without the need to add ugly `private[this]`'s on the internals.

Yet you can make this package private (`private[heartbeat]` instead of `protected`) – it'll still be visible in the test.

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -0,0 +1,97 @@
...
+  import Heartbeat._
+
+  sealed trait State
+  case object StateInactive extends State
```
meichstedt: Iirc I've added a note about this before which probably wasn't obvious: imho `Inactive` would be sufficient to denote a state, while `Deactivate` would be sufficient to denote a message – I don't think we need a prefix, just as `Starting` seems like a catchier task state to me than `TASK_STARTING`.

We don't have a clear convention here and I don't feel too strong about this. I'd be ok to leave it like that and discuss mid of August ;)

src/test/scala/mesosphere/marathon/MarathonSchedulerTest.scala
```diff
@@ -126,6 +122,12 @@ class MarathonSchedulerTest extends MarathonActorSupport with MarathonSpec with
     } finally {
       eventBus.unsubscribe(probe.ref)
     }
+
+    // we **heavily** rely on driver.stop to delegate enforcement of leadership abdication,
+    // so it's worth testing that this behavior isn't lost.
```
meichstedt: 👍

meichstedt:
@jdef Great, LGTM – how confident are you in those leadership election changes? I'd need a little more time to wrap my head around those. Maybe @gkleiman and @aquamatthias can have a look at those, too? I'm afraid none of us is a leadership election expert ...

jdef:
the leadership election changes are probably what I'm least confident about
w/ respect to this PR. but without them Marathon was unstable in the face
of various partition/error scenarios. particularly because of
double-reactions to leadership change: e.g. the candidate implementation of
stopLeadership (that's reactivating to "i'm no longer leader" event) was
redundantly abdicating leadership and wreaking havoc. I *think* I've
considered the error conditions properly in my refactoring, but .. yes,
possible that I've missing something/assumed too much. Ideally this code
would be completely rewritten into a more easily understood FSM. I was
trying to keep this PR small AND solve the problems at hand. If we decide
that an FSM for this code must land in this PR .. I'm fine heading that
direction, it just seemed more appropriate for a follow-up PR

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
jdef: So I put some logging in the heartbeat monitor. I'm happy to put it here
instead. If it lands both places then there will be somewhat redundant logs
... or is that what you're aiming for?

Even better, I could create a LoggingReactor that decorates what the caller
provides and generates appropriate logging before delegating. Thoughts?

src/main/scala/mesosphere/marathon/core/heartbeat/MesosHeartbeatMonitor.scala
jdef: ack

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
jdef: ack

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
jdef: yes, you did comment on this prior. I tried it and found the code harder to
read. The convention I used carries over from golang and IMHO makes it
easier to differentiate among enum'd types. So, I left it as is. I'm happy
to revise if people find this confusing.

TASK_STARTING is not a good name, it's already encapsulated by TaskState.
so TaskState.STARTING would be sufficient for context and to reduce
stutter. Oh well, that's inherited from the Mesos API.

FWIW in this PR StateInactive and MessageDeactivate do not stutter.

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
jdef: because I haven't memorized scaladoc conventions yet. I can take a stab at
this. Last time I tried to do this sbt complained about something I didn't
understand, and .. time...

meichstedt: I see. No big deal. We don't generate any docs so far, so that's that.

In general, the way to go would be to
```
  /**
    * [docs here]
    */
```

(IntelliJ happens to generate annotations for arguments/return type and such if you type the `/**` and hit enter fwiw.)

jdef: right, well.. VI doesn't :)

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
meichstedt: Ah I didn't follow the lines then. Seemed appropriate to have them where the HeartbeatActor _decides_ to call `onFailure`. Not generally fond of redundant logs, although, in this case, it would just log _hey I'm calling onFailure_, then _hey, onFailure was called_ and then abdicate, so not really polluting the logs.

Long story short: I have a slight inclination towards logging here, and I wouldn't mind redundancy in this case.

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
meichstedt: ack, fine with that.

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
meichstedt: that's the true spirit, man 😎

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -245,12 +260,14 @@ class MarathonSchedulerService @Inject() (
...
+      mesosHeartbeatActor ! Heartbeat.MessageDeactivate(MesosHeartbeatMonitor.sessionOf(driverInstance))
       // Our leadership has been defeated. Thus, stop the driver.
-      // Note that abdication command will be ran upon driver shutdown which
-      // will then offer leadership again.
       stopDriver()
-    } else {
+    }
+    // Abdication will have already happened if the driver terminated abnormally.
+    // Otherwise we've either been terminated or have lost leadership for some other reason (network part?)
+    if (driver.isEmpty && isRunningLatch.getCount > 0) {
```
jdef: the check for `driver.isEmpty` here is useless since `stopDriver` is guaranteed to set `driver` to `None`

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
jdef: I've addressed with a logging decorator

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
jdef: I've addressed this with a logging decorator

jdef:
@meichstedt I've addressed your review comments PTAL

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -218,16 +225,23 @@ class MarathonSchedulerService @Inject() (
...
           case _ =>
         }
 
-        // tell leader election that we step back, but want to be re-elected if isRunning is true.
-        electionService.abdicateLeadership(error = result.isFailure, reoffer = isRunningLatch.getCount > 0)
+        // ONLY do this if there's some sort of driver crash: avoid invoking abdication logic if
+        // the driver was stopped via stopDriver. stopDriver only happens when
+        //   1. we're being terminated (and have already abdicated)
+        //   2. we've lost leadership (no need to abdicate if we've already lost)
+        driver.foreach { _ =>
```
aquamatthias: So in the case we get elected, and fail during driver creation, we would never abdicate?

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -245,12 +260,14 @@ class MarathonSchedulerService @Inject() (
     timer = newTimer()
     oldTimer.cancel()
 
-    if (driver.isDefined) {
+    driver.map { driverInstance =>
```
aquamatthias: use foreach

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
jdef: the driver is created above via `driverFactory.createDriver`. so this `onComplete` block will always execute once `driver.run` completes. if "we get elected, and fail during driver creation" means what? a runtime exception? if so, then the code responsible for calling `startLeadership` better handle that case.

jdef:
rebased and addressed comments from @aquamatthias

aquamatthias:
@jdef Thanks!

aquamatthias:
@jdef please cherry pick in 1.2, 1.1 and 0.15

PR #4070, opened 2016-07-07
Add warning that Marathon doesn't enforce service ports uniqueness

gkleiman:


gkleiman:
Actually... Marathon is inconsistent when it comes to validating service ports uniqueness =(.

It is possible to specify a duplicated service port through the `/v2/apps` endpoints, but not through the `/v2/groups` endpoints.

jdef:
@gkleiman should we create an issue to fix the aforementioned inconsistency problem?

PR #4081, opened 2016-07-11
Bump Mesos dep to 1.0.0-rc2

gkleiman:


aquamatthias:
Some methods have been deprecated.

gkleiman:
@aquamatthias I'll look into this with @jdef's help.

gkleiman:
Mesos v1.1 will ship with https://reviews.apache.org/r/49958/, which should let us use the new non-deprecated method without a workaround

In the meantime I updated the build definition in order to allow calls to deprecated methods.

project/build.scala
```diff
@@ -306,7 +306,7 @@ object Dependency {
     // runtime deps versions
     val Chaos = "0.8.7"
     val Guava = "19.0"
-    val MesosUtils = "1.0.0-rc1"
+    val MesosUtils = "1.0.0-rc2"
```
aquamatthias: Please add a comment here (regarding deprecation) as well, so we get reminded if we change the version here.

aquamatthias:
LGTM

gkleiman:
@aquamatthias done

aquamatthias:
Waiting for tests to pass

PR #4088, opened 2016-07-12
Unified Mesos Containerizer support

bernd-mesos:
Formerly https://github.com/mesosphere/marathon/pull/4023


aquamatthias:
@nfnt Can you work on this based on the comments from @meichstedt

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -291,15 +286,15 @@ object TaskBuilder {
     val declaredPorts = {
       val containerPorts = for {
         c <- runSpec.container
-        pms <- c.portMappings
+        pms <- c.getPortMappings
```
aquamatthias: Please revert

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -33,43 +34,43 @@ case class Container(
...
-  /**
-    * Docker-specific container parameters.
-    */
   case class Docker(
+    volumes: Seq[Volume] = Seq.empty,
     image: String = "",
-    network: Option[Mesos.ContainerInfo.DockerInfo.Network] = None,
-    portMappings: Option[Seq[Docker.PortMapping]] = None,
+    network: Option[ContainerInfo.DockerInfo.Network] = None,
+    pms: Seq[Docker.PortMapping] = Seq.empty,
```
aquamatthias: `pms` is bad idea. Validation will report `pms`

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -765,7 +902,10 @@ trait AppAndGroupFormats {
     * made it optional (also with ip-per-container in mind) and we have to re-add it in cases where it makes sense.
     */
   private[this] def addHealthCheckPortIndexIfNecessary(app: AppDefinition): AppDefinition = {
-    val hasPortMappings = app.container.exists(_.docker.exists(_.portMappings.exists(_.nonEmpty)))
+    val hasPortMappings = app.container.exists(_ match {
+      case docker: Container.Docker => docker.pms.nonEmpty
```
aquamatthias: This should be delegated to the container. E.g. MesosDocker could support this in the future.

src/main/scala/mesosphere/marathon/state/Container.scala
aquamatthias: I guess this should be `Option[Seq[PortMapping]]` here as well.  The TaskBuilder differentiates between None and Seq.empty.

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,25 +3,26 @@ package mesosphere.marathon.state
...
+  def docker(): Option[Container.Docker] = {
+    this match {
+      case docker: Container.Docker => Some(docker)
+      case _ => None
+    }
+  }
 
-  def portMappings: Option[Seq[Docker.PortMapping]] = {
-    import Mesos.ContainerInfo.DockerInfo.Network
+  def portMappings: Option[Seq[Container.Docker.PortMapping]] = {
```
aquamatthias: Should be overriden in Docker. If you want to provide a default implementation for all other Containers, use `None` here.

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -33,43 +34,43 @@ case class Container(
...
 
     def withDefaultPortMappings(
+      volumes: Seq[Volume],
       image: String = "",
-      network: Option[Mesos.ContainerInfo.DockerInfo.Network] = None,
+      network: Option[ContainerInfo.DockerInfo.Network] = None,
       portMappings: Option[Seq[Docker.PortMapping]] = None,
-      privileged: Boolean = false,
-      parameters: Seq[Parameter] = Nil,
+      privileged: Option[Boolean] = None,
```
aquamatthias: why is this an Option? (options as arguments are usually an anti pattern)

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -33,43 +34,43 @@ case class Container(
...
+      parameters: Option[Seq[Parameter]] = None,
```
aquamatthias: why is this an Option? (options as arguments are usually an anti pattern)

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -33,43 +34,43 @@ case class Container(
...
+        case Some(networkMode) if networkMode == ContainerInfo.DockerInfo.Network.BRIDGE =>
           portMappings.map(_.map { m =>
             m match {
               // backwards compat: when in BRIDGE mode, missing host ports default to zero
               case PortMapping(x, None, y, z, w, a) => PortMapping(x, Some(PortMapping.HostPortDefault), y, z, w, a)
               case _ => m
             }
-          })
-        case _ => portMappings
+          }).getOrElse(Seq.empty)
```
aquamatthias: I think the getOrElse is just wrong here.

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -108,54 +109,82 @@ object Container {
...
+      def validForDocker(docker: Docker): Validator[Seq[PortMapping]] = validator[Seq[PortMapping]] { pm =>
+        pm is every(valid(PortMapping.networkHostPortValidator(docker)))
       }
     }
 
-    implicit val dockerValidator = validator[Docker] { docker =>
+    val validDockerContainer = validator[Docker] { docker =>
       docker.image is notEmpty
-      docker.portMappings is optional(PortMappings.portMappingsValidator and PortMappings.validForDocker(docker))
+      docker.pms is PortMapping.portMappingsValidator and PortMapping.validForDocker(docker)
```
aquamatthias: The user provides `portMappings` but the validation will return `pms`...

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -44,9 +44,8 @@ class PortsMatcher(
     val portMappings: Option[Seq[Container.Docker.PortMapping]] =
       for {
         c <- runSpec.container
-        d <- c.docker
-        pms <- d.portMappings if pms.nonEmpty
-      } yield pms
+        d <- c.docker if !d.pms.isEmpty
```
aquamatthias: `c.portMappings`?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -208,15 +208,16 @@ class TaskBuilder(
...
+    } else {
       val builder = ContainerInfo.newBuilder
 
       // Fill in Docker container details if necessary
       runSpec.container.foreach { c =>
-        val portMappings = c.docker.map { d =>
-          d.portMappings.map { pms =>
-            pms.zip(hostPorts).collect {
+        val containerWithPortMappings = c match {
+          case docker: Container.Docker => docker.copy(pms =
```
aquamatthias: same here. Should we handle portMappings more generic?

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -3,25 +3,26 @@ package mesosphere.marathon.state
...
-// Current implementation with type defaulting to DOCKER and docker to NONE makes no sense
-case class Container(
-    `type`: Mesos.ContainerInfo.Type = Mesos.ContainerInfo.Type.DOCKER,
-    volumes: Seq[Volume] = Nil,
-    docker: Option[Container.Docker] = None) {
+sealed trait Container {
+  val volumes: Seq[Volume]
 
-  import Container._
+  def docker(): Option[Container.Docker] = {
```
aquamatthias: Please add a todo to remove this method here.
Also: review usages of this method and see if we can remove a hard dependency to docker (or if this is even wrong e.g. portMappings).

src/main/scala/mesosphere/marathon/state/Container.scala
aquamatthias: Add validation, that makes sure that portMappings are only valid if network mode is bridge or user

src/main/scala/mesosphere/marathon/state/Container.scala
aquamatthias: You can override the default implementation in Container here.

src/main/scala/mesosphere/marathon/state/Container.scala
nfnt: Good point! Looks like the containerizer is assumed to be Docker, when the docker field is `Some`. This may lead to problems with the Mesos containerizer, that now also can set this field.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
nfnt: In general: Yes, please! In this particular case, it's hard to change. Here, a new Container is created (by copying) that may have different port mappings. Because `Container` has been changed to be a trait, the copy has to be created from the actual implementation. Right now, only `Container.Docker` supports port mappings, hence only this case is considered here. If/When other containers implementations get support for port mappings, these cases have to be implemented here as well.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -208,41 +208,37 @@ class TaskBuilder(
...
-          }
-        }
-
-        val containerWithPortMappings = portMappings match {
-          case None => c
-          case Some(newMappings) => c.copy(
-            docker = c.docker.map {
-              _.copy(portMappings = newMappings)
+        val containerWithPortMappings = c match {
+          case docker: Container.Docker => docker.copy(
```
aquamatthias: Why copy here? And why special docker handling? Could be `c.portMappings`?

nfnt: We need a container that zips its `portMappings` with the host's ports but otherwise not mutating the container. Is there a nicer way to do this in Scala? IMO copying is the best way to do here, but as we cannot copy the trait, we need to match the type first.

aquamatthias: I think the problem here is: the change of the portMappings should be done elsewhere, since this logic is tied to docker. Please add a TODO on top of this code - we have to solve this separately.

aquamatthias:
Integration tests failed for: 
```
DockerAppIntegrationTest:
 - deploy a simple Docker app using the Mesos containerizer *** FAILED *** (30 seconds, 94 milliseconds)
```
might be shaky. I started it again.

docs/docs/native-docker.md
```diff
@@ -280,6 +280,58 @@ the future, as Mesos may not always interact with Docker via the CLI.
 }
 ```
 
+#### Docker container support without Docker Engine
+
+Starting with version 1.2.0, Marathon supports docker container images without having the Docker Containerizer
```
aquamatthias: It will be part of 1.3.0

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -248,17 +248,14 @@ private[impl] class GroupManagerActor(
       // defined only if there are port mappings
       val newContainer: Option[Container] = for {
         c <- app.container
-        d <- c.docker
-        pms <- d.portMappings
+        d <- c.docker if !d.portMappings.isEmpty
```
aquamatthias: Do we have to make sure docker here?
```
     c <- app.container
   pms <- c.portMappings
```

nfnt:
No, the integration test will always fail. The Mesos environment is lacking support for running docker apps using the Mesos containerizer, so it can't work. I'll fix it.

PR #4089, opened 2016-07-12
Universal Mesos Containerizer support

bernd-mesos:


aquamatthias:
@nfnt Can you work on this based on the comments from @meichstedt

src/main/scala/mesosphere/marathon/state/Container.scala
```diff
@@ -37,16 +36,16 @@ object Container {
 
   case class Mesos(volumes: Seq[Volume] = Seq.empty) extends Container
 
-  case class Docker(
+  case class DockerDocker(
```
meichstedt: Imo we can go with `Docker` vs `MesosDocker`. I'm getting gosebumps when I read _DockerDocker_

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -2,49 +2,49 @@ package mesosphere.marathon.api.serialization
...
-import mesosphere.marathon.state.Container.Docker.PortMapping
+import mesosphere.marathon.state.Container.DockerDocker.PortMapping
 import mesosphere.marathon.state._
 import org.apache.mesos
 
 import scala.collection.JavaConverters._
 import scala.collection.immutable.Seq
 
 object ContainerSerializer {
+  def fromProto(proto: Protos.ExtendedContainerInfo): Container = {
```
meichstedt: The specific `fromProto` to which we delegate from here do not check whether the respective proto field is set. Instead, they directly access `getDocker`, `getMesosDocker`, `getMesosAppC`, which will lead to runtime exceptions when the caller of these specific `fromProto` functions does not correctly verify whether the fields are set, or when the functions are called from another location.

I'd strongly suggest to make this type- and nullsafe. One solution would be to provide partial functions that only match if the given field is set

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -2,49 +2,49 @@ package mesosphere.marathon.api.serialization
...
+
   def toProto(container: Container): Protos.ExtendedContainerInfo = {
     val builder = Protos.ExtendedContainerInfo.newBuilder
       .addAllVolumes(container.volumes.map(VolumeSerializer.toProto).asJava)
 
     container match {
       case _: Container.Mesos =>
         builder.setType(mesos.Protos.ContainerInfo.Type.MESOS)
-      case docker: Container.Docker =>
+      case dd: Container.DockerDocker =>
```
meichstedt: the `trait Container` is not sealed, therefore the compiler cannot check whether this check is exhaustive. Please change Container to be a `sealed trait`.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -53,12 +53,18 @@ object ContainerSerializer {
...
+      case md: Container.MesosDocker =>
+        builder.setType(mesos.Protos.ContainerInfo.Type.MESOS)
+        builder.setMesos(MesosDockerSerializer.toMesos(md))
+      case ma: Container.MesosAppC =>
+        builder.setType(mesos.Protos.ContainerInfo.Type.MESOS)
+        builder.setMesos(MesosAppCSerializer.toMesos(ma))
     }
 
-    container.volumes.foreach {
+    container.volumes foreach {
```
meichstedt: We don't have a preference towards dotless syntax, so I suggest to revert this change and keep the changed lines small.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -117,17 +123,31 @@ object ExternalVolumeInfoSerializer {
       .setName(info.name)
       .setProvider(info.provider)
 
-    info.size.foreach(builder.setSize)
+    info.size foreach builder.setSize
```
meichstedt: We don't have a preference towards dotless syntax, so I suggest to revert this change and keep the changed lines small.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -117,17 +123,31 @@ object ExternalVolumeInfoSerializer {
...
     info.options.map{
       case (key, value) => mesos.Protos.Label.newBuilder().setKey(key).setValue(value).build
-    }.foreach(builder.addOptions)
+    } foreach builder.addOptions
```
meichstedt: We don't have a preference towards dotless syntax, so I suggest to revert this change and keep the changed lines small.

docs/docs/native-docker.md
```diff
@@ -280,6 +280,58 @@ the future, as Mesos may not always interact with Docker via the CLI.
 }
 ```
 
+#### Docker container support without Docker Engine
+
+Starting with version 1.2.0, Marathon supports docker container images without having the Docker Containerizer
+depend on a Docker Engine. Instead the Mesos containerizer with native AppC support added in Apache Mesos version 1.0
+(released July 2016) directly uses native OS features to configure and start Docker containers and to provide isolation.
+This setup is selected by specifying a JSON combination that used to provoke an error message:
```
meichstedt: I am not sure whether extending the v2 container spec is a nice way to go. Several thoughts here:
- we will start working on v3 _very soon_ and are about to deprecate v2
- the container spec is already overloaded and this difficult to parse and validate
- adding a new json node with a specific set of available properties would not ease validation and prevent extra effort in the MarathonUI needed to cope with new combinations that were invalid before.

nfnt:
I've addressed some of the comments and made it mergeable with **bernd-uc1**. I'll merge it now and keep working on resolving the remaining comments in **bernd-uc1**.

Issue #3957, opened 2016-06-01
COMMAND HealthChecks removed on mesos-master restart/issue


davent:
COMMAND Health Checks are being removed (i.e. `INFO Removing health check for app`) and never re-created when something happens to the mesos-master. I have seen this with both a restart of the mesos-master and also under less controlled situations such as a leader re-election or network issues.

To reproduce:

1) Create an App with a COMMAND Health Check
2) Restart mesos-master
3) grep marathon log for `Removing health check`

Although `addAllFor` and subsequently `add` is called (which you can see int he logs: `INFO Adding health check for app`) No health check data is ever available either via the UI (the good old grey bar of doom) or via the JSON return via the API.

The HTTP and TCP Health Checks return successfully but the only way to re-establish the COMMAND Health Checks is to restart the Apps.

Using:
Mesos: 0.28.0
Marathon: 1.1.1

brat002:
Have this problem too.

andlaz:
Same issue.

davent:
Some additional information:

When a Mesos Slave drops from the cluster (in the case just witnessed this was due to a faulty EC2 node having to be rebooted) and then rejoins this appears to also cause this issue.

brat002:
Does any plans to fix it? Marathon is unusable in real production infrastructure without health checks.

kensipe:
@davent @brat002 @andlaz  Thanks for the report.  After some review, It looks like it may be an issue with Apache Mesos.   Command health checks are reported differently than other health checks in Mesos.  They are received via a task status update.   If you look at the implementation in Mesos:  https://github.com/apache/mesos/blob/master/src/health-check/main.cpp#L135-L137
Updates are only sent:
1) for the first success
2) after a failure
In your example there is no failure so no update of status.  It doesn't make sense for Marathon to assume a success.  When need to get the health status from mesos after a network partition or master fail over.

I'm working with the core team to get an answer in place.

brat002:
After some improvements in mesos https://github.com/apache/mesos/issues/118, looks like marathon also has a problem. 
Now Mesos shows health: true after leader re-election, but marathon still show "unknown" health.

Marathon task status:
`id: "lsau01-XXXXXXXXXXXXXX",
name: "lsau01-XXXXXXXXXXXXXX",
framework_id: "c5ad05ba-4a33-4d65-ab67-4d7c29928985-0001",
executor_id: "",
[...]

PR #4094, opened 2016-07-13
Fixed health check status on restart or mesos leader loss

tsydd:
When mesos master is lost or marathon is restarted, task status is shown as "Unknown". It happens because health check status is not handled correctly when health check is in "reconcile" state.

Prepare steps:

1. run zookeeper with default config:

   ```ini
   tickTime=2000
   initLimit=10
   syncLimit=5
   dataDir=/tmp/zookeeper
   ```
   `$ ./bin/zkServer.sh start-foreground`

2. run mesos master and mesos slaves:
   `$ mesos-master.sh --ip=127.0.0.1 --work-dir=/tmp/wd1 --quorum=2 --port=5050 --zk=zk://localhost:2181/mesos`
   `$ mesos-master.sh --ip=127.0.0.1 --work-dir=/tmp/wd2 --quorum=2 --port=5051 --zk=zk://localhost:2181/mesos`
   `$ mesos-master.sh --ip=127.0.0.1 --work-dir=/tmp/wd3 --quorum=2 --port=5052 --zk=zk://localhost:2181/mesos`

3. run mesos agent:
   `$ mesos-agent.sh --master=zk://localhost:2181/mesos --work_dir=/tmp/wda1 --port=5150`

4. run marathon with parameters
   `--master zk://localhost:2181/mesos --zk zk://localhost:2181/marathon`

5. go to http://localhost:8080 and create application with
   * command: `tail -f /dev/null`
   * health check:
      * protocol: `command`
      * command: `true`

After application is created, it's health status will be "healthy"

Steps to reproduce:
* kill mesos master
* restart marathon

In both cases task status becomes "Unknown" instead of "Healthy"

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -8,6 +8,7 @@ import mesosphere.marathon.core.base.{ CurrentRuntime, Clock }
 import mesosphere.marathon.core.launcher.OfferProcessor
 import mesosphere.marathon.core.task.update.TaskStatusUpdateProcessor
 import mesosphere.marathon.event._
+import mesosphere.marathon.health.HealthCheckManager
 import mesosphere.util.state.{ FrameworkIdUtil, MesosLeaderInfo }
 import org.apache.mesos.Protos._
 import org.apache.mesos.{ Scheduler, SchedulerDriver }
@@ -29,7 +30,8 @@ class MarathonScheduler @Inject() (
     mesosLeaderInfo: MesosLeaderInfo,
     system: ActorSystem,
     config: MarathonConf,
-    schedulerCallbacks: SchedulerCallbacks) extends Scheduler {
+    schedulerCallbacks: SchedulerCallbacks,
+    healthCheckManager: HealthCheckManager) extends Scheduler {
 
   private[this] val log = LoggerFactory.getLogger(getClass.getName)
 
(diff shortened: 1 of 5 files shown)


brat002:
Looks great! Thank you!

brat002:
Works as expected. Please, merge this.

markeijsermans:
I can agree @brat002 in #3957, this is a major issue for us as we heavily rely on health checks in production and use this data for such things as (Sensu) alerts, service discovery etc. Expediting this would be highly appreciated. At the moment we are forced to restart/deploy apps manually, thereby rendering our masters sort of non-HA. Thanks

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -71,7 +73,7 @@ class MarathonScheduler @Inject() (
   override def statusUpdate(driver: SchedulerDriver, status: TaskStatus): Unit = {
     log.info("Received status update for task %s: %s (%s)"
       .format(status.getTaskId.getValue, status.getState, status.getMessage))
-
+    healthCheckManager.reconcile(status)
```
aquamatthias: @meichstedt: Instead of hard wiring the reconcile logic here: can we make this part of `NotifyHealthCheckManagerStepImpl`?

meichstedt: Not really with the current implementation: If a statusUpdate received during a reconciliation does not change the state of a task, no StateChange is populated, and we can't use the update step processor.

meichstedt:
Imo a better solution similar to other places would be to route all tasks to the HealthCheckManagerActor after a leader election, as we're only interested in reconciling after a failover. A simple approach would be to have the `HealthCheckManagerActor` load all tasks from the TaskTracker in `preStart` (the actor is started after Marathon becomes leader) and call `update` for each launched task with the according status. That would load a potentially out of date state, but as soon as statusUpdates triggered by reconciliation are received, the health status would be updated anyways.

aquamatthias:
Close in favor of #4210

PR #4096, opened 2016-07-13
Fixes #4095 - Used Map instead of Set to store apps in Group.

### LINKED ISSUE #4095: Use a Map to store apps in Group
The `Group` class currently uses a `Set` to store the child apps. This can lead to bad performance as we can not index the `Set` by the `appid`. Instead we iterate over the whole collection doing expensive comparisons.
This breaks down when there are many apps in a single group.
Consider replacing the `Set` with a `Map` that can represent the primary key `appid` correctly.

jmlvanre:


plugin-interface/src/main/scala/mesosphere/marathon/plugin/Group.scala
```diff
@@ -5,7 +5,7 @@ package mesosphere.marathon.plugin
   */
 trait Group {
   def id: PathId
-  def apps: Iterable[RunSpec]
+  def apps: Iterable[(_, RunSpec)]
```
jasongilanfarr: whats the first part of the tuple? maybe better to be explicit/a map/a case class?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -1058,11 +1058,14 @@ trait AppAndGroupFormats {
...
     (__ \ "id").format[PathId] ~
-    (__ \ "apps").formatNullable[Set[AppDefinition]].withDefault(Group.defaultApps) ~
+    (__ \ "apps").formatNullable[Iterable[AppDefinition]].withDefault(Iterable.empty) ~
     (__ \ "groups").lazyFormatNullable(implicitly[Format[Set[Group]]]).withDefault(Group.defaultGroups) ~
     (__ \ "dependencies").formatNullable[Set[PathId]].withDefault(Group.defaultDependencies) ~
     (__ \ "version").formatNullable[Timestamp].withDefault(Group.defaultVersion)
-  ) (Group(_, _, _, _, _), unlift(Group.unapply))
+  ) (
+      (id, apps, groups, dependencies, version) =>
+        Group(id, apps.map(app => app.id -> app).toMap, groups, dependencies, version),
```
jasongilanfarr: replace ".toMap" with "(collection.breakOut)" which should build the map for you (and it avoids an intermediate List of Pairs)

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -54,7 +55,7 @@ case class Group(
   def updateApp(path: PathId, fn: Option[AppDefinition] => AppDefinition, timestamp: Timestamp): Group = {
     val groupId = path.parent
     makeGroup(groupId).update(timestamp) { group =>
-      if (group.id == groupId) group.putApplication(fn(group.apps.find(_.id == path))) else group
+      if (group.id == groupId) group.putApplication(fn(group.apps get path)) else group
```
jasongilanfarr: could you use "." notation please? e.g. group.apps.get(path) ?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -1058,11 +1058,14 @@ trait AppAndGroupFormats {
 
   implicit lazy val GroupFormat: Format[Group] = (
     (__ \ "id").format[PathId] ~
-    (__ \ "apps").formatNullable[Set[AppDefinition]].withDefault(Group.defaultApps) ~
+    (__ \ "apps").formatNullable[Iterable[AppDefinition]].withDefault(Iterable.empty) ~
```
gkleiman: We should probably make Group.defaultApps be Iterable.empty/Map.empty.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/Group.scala
gkleiman: PathId would make sense.

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -83,7 +83,7 @@ private[appinfo] class DefaultInfoService(
             None
         def apps: Option[Seq[AppInfo]] =
           if (groupEmbed(GroupInfo.Embed.Apps))
-            Some(ref.apps.toIndexedSeq.flatMap(a => infoById.get(a.id)).sortBy(_.app.id))
+            Some(ref.apps.values.toIndexedSeq.flatMap(a => infoById.get(a.id)).sortBy(_.app.id))
```
gkleiman: Looks like `ref.apps.keys.flatMap(infoById.get).sortBy(_.app.id)` could also work and be cheaper.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -42,7 +43,7 @@ case class Group(
     if (fn(this)) Some(this) else in(groups.toList)
   }
 
-  def app(appId: PathId): Option[AppDefinition] = group(appId.parent).flatMap(_.apps.find(_.id == appId))
+  def app(appId: PathId): Option[AppDefinition] = group(appId.parent).flatMap(_.apps get appId)
```
gkleiman: Please use the "." notation here as well.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -96,7 +97,7 @@ case class Group(
       // replace it. Otherwise do not replace it. Validation will catch conflicting app/group IDs later.
       groups = groups.filter { group => group.id != appDef.id || group.containsApps },
       // replace potentially existing app definition
-      apps = apps.filter(_.id != appDef.id) + appDef
+      apps = apps - appDef.id + ((appDef.id, appDef))
```
gkleiman: nit: I think that `appDef.id -> appDef` makes it clearer that you're adding a tuple to the map.

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -7,6 +7,8 @@ import mesosphere.marathon.plugin
 
 case class PathId(path: List[String], absolute: Boolean = true) extends Ordered[PathId] with plugin.PathId {
 
+  val str = toString()
```
aquamatthias: instead of creating a new member, you can redefine `toString()`
```scala
override val toString: String = toString("/")
```

plugin-interface/src/main/scala/mesosphere/marathon/plugin/Group.scala
aquamatthias: why `Iterable` and not `Map[PathId, RunSpec]`?

src/test/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProviderRootGroupValidationTest.scala
```diff
@@ -16,14 +16,16 @@ class DVDIProviderRootGroupValidationTest extends FunSuite with Matchers with Gi
...
+    val app2 = f.appWithDVDIVolume(appId = PathId("/nested/foo2"), volumeName = "vol2")
     val rootGroup = Group(
       id = PathId.empty,
       groups = Set(
         Group(
           id = PathId("/nested"),
-          apps = Set(
-            f.appWithDVDIVolume(appId = PathId("/nested/foo1"), volumeName = "vol1"),
-            f.appWithDVDIVolume(appId = PathId("/nested/foo2"), volumeName = "vol2")
+          apps = Map(
```
aquamatthias: Side note:
you can write Tuple2 as this: a->b, so you can define Maps like:
```scala
Mao(app1.id->app1, app2.id->apps2)
```

src/main/scala/mesosphere/marathon/state/RunSpec.scala
```diff
@@ -12,6 +12,9 @@ import scala.collection.immutable.Seq
 
 //scalastyle:off
 trait RunSpec extends plugin.RunSpec {
+
+  type Id = PathId
```
jasongilanfarr: This doesn't quite work as you would expect, it will if and only if you put it in "object RunSpec".

aquamatthias:
@jmlvanre Thanks! LGTM. There are a couple of comments, that should be addressed before merge.
I'm also interested in the representation problems you mentioned - not 100% sure what you mean.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -96,7 +97,7 @@ case class Group(
       // replace it. Otherwise do not replace it. Validation will catch conflicting app/group IDs later.
       groups = groups.filter { group => group.id != appDef.id || group.containsApps },
       // replace potentially existing app definition
-      apps = apps.filter(_.id != appDef.id) + appDef
+      apps = apps - appDef.id + ((appDef.id -> appDef))
```
aquamatthias: Please remove -appId and redundant braces

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -243,7 +244,7 @@ object Group {
   private def noAppsAndGroupsWithSameName: Validator[Group] =
     isTrue(s"Groups and Applications may not have the same identifier.") { group =>
       val groupIds = group.groups.map(_.id)
-      val clashingIds = group.apps.map(_.id).intersect(groupIds)
+      val clashingIds = groupIds.intersect(group.apps.keys.toSet)
```
aquamatthias: use `keySet` instead of `keys.toSet`

src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -69,14 +69,27 @@ case class PathId(path: List[String], absolute: Boolean = true) extends Ordered[
     path.zip(definition.path).forall { case (left, right) => left == right }
   }
 
-  override def toString: String = toString("/")
+  override def toString: String = {
+    lazy val str = toString("/")
```
aquamatthias: This will be executed every time `toString` is called.
Please use 
```
override val toString: String = toString("/")
```
this will be evaluated only once.

src/test/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProviderRootGroupValidationTest.scala
aquamatthias: Thanks for changing it to this syntax!

aquamatthias:
Does not think this has something to do with this PR.
```
*** 1 TEST FAILED ***
[14:06:38][Step 3/3] ResidentTaskIntegrationTest:
[14:06:38][Step 3/3] 
[14:06:38][Step 3/3] - Restart *** FAILED *** (30 seconds, 509 milliseconds)
[14:06:38][Step 3/3]   java.lang.AssertionError: Waiting for event deployment_success to arrive took longer than 30 seconds. Give up.
```

aquamatthias:
Thanks!

PR #4097, opened 2016-07-13
Preview: Persistence API with Propagation

jasongilanfarr:
- New Persistence API
- ZK (with folder structures), InMemory, Lazy and Eager Caching
- Lots of "scala.collection.{Iterable,Seq}" -> "scala.collection.immutable.{Iterable,Seq}"
- AppRepository can now store in Zk too (not wired via a FeatureFlag yet)
- Rewrote AppRepositoryTest to use InMemoryStorage and no mocks - turned out it had incorrect tests...


src/main/scala/mesosphere/marathon/core/storage/MarathonState.scala
```diff
@@ -0,0 +1,22 @@
+package mesosphere.marathon.core.storage
+
+import akka.util.ByteString
+import com.google.protobuf.{ MessageLite, Parser }
+import mesosphere.marathon.state.Timestamp
+
+trait MarathonState[Proto <: MessageLite] {
```
aquamatthias: Do we really need this? It means that all entity types have to implement this.
I like the idea more to have a type class implement the proto handling.

I guess everything we need is to add this into  MarathonProto:
```scala
trait MarathonProto[Proto <: MessageLite, T] {
  def toProto(t: T): Proto
  def version(t: T): Timestamp
  ...
```

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkStoreSerialization.scala
```diff
@@ -0,0 +1,51 @@
...
+  private val bucket = id.hashCode % 16
+  def path: String = version.fold(s"/$category/$bucket/$id") { v =>
+    s"/$category/$bucket/$id/versions/$v"
+  }
+}
+
+case class ZkSerialized(bytes: ByteString)
+
+trait ZkStoreSerialization {
+  val maxVersions = 25
```
aquamatthias: This value should be set via configuration ``--zk_max_versions``

src/main/scala/mesosphere/marathon/core/storage/repository/impl/BaseAppRepository.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  override def app(appId: PathId, version: OffsetDateTime): Future[Option[AppDefinition]] = store.get(appId, version)
+
+  override def store(appDef: AppDefinition): Future[Done] = store.store(appDef.id, appDef)
+
+  override def apps(): Source[AppDefinition, NotUsed] =
+    allPathIds().mapAsync(MaxParallelGets)(store.get(_)).filter(_.isDefined).map(_.get)
+}
+
+class AppZkRepository(store: PersistenceStore[ZkId, String, ZkSerialized], maxVersions: Int)
```
aquamatthias: Why do we need this specific classes here?
```
class AppRepository[K,C,S](store: PersistenceStore[K,C,S], maxVersions: Int)

val zkStore: PersistenceStore[ZkId, String, ZkSerialized] = _
val zkRepo = new AppRepository(zkStore, 1)

val memStore: PersistenceStore[RamId, String, Identity] = _
val memRepo = new AppRepository(memStore, 1)
```

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -348,7 +351,7 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
   def provideAppRepository(
     @Named(ModuleNames.STORE_APP) store: EntityStore[AppDefinition],
     metrics: Metrics): AppRepository = {
-    new AppRepository(store, maxVersions = conf.zooKeeperMaxVersions.get, metrics)
+    new AppEntityRepository(store, maxVersions = conf.zooKeeperMaxVersions.get, metrics)
```
aquamatthias: Ideally we get rid of all store/repository objects here and make them available via the storage module (no guice)

src/main/scala/mesosphere/marathon/core/storage/repository/AppRepository.scala
```diff
@@ -0,0 +1,61 @@
...
+  *
+  * Current state:
+  * - all applications are stored as part of the root group in the group repository for every user intended change
+  * - all applications are stored again in the app repository, if the deployment of that application starts
+  *
+  * Future plan:
+  * - the applications should be always loaded via the groupManager or groupRepository.
+  * - the app repository is used to store versions of the application
+  *
+  * Until this plan is implemented, please think carefully when to use the app repository!
```
aquamatthias: I guess this plan is not 100% valid any longer...

src/main/scala/mesosphere/marathon/core/storage/repository/AppRepository.scala
```diff
@@ -0,0 +1,61 @@
...
+  */
+trait AppRepository {
```
aquamatthias: Just thinking: the main purpose of this interface is to:
- talk to the PersistenceStore 
- take the IdResolver, Marshaller, Unmarshaller

Would it make sense to define this interface generic, so we can use it for all types?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/BaseAppRepository.scala
```diff
@@ -0,0 +1,47 @@
...
+import mesosphere.marathon.core.storage.repository.AppRepository
+import mesosphere.marathon.core.storage.{ IdResolver, PersistenceStore }
+import mesosphere.marathon.state.{ AppDefinition, PathId }
+
+import mesosphere.marathon.core.storage.impl.memory.InMemoryStoreSerialization._
+import mesosphere.marathon.core.storage.impl.zk.ZkStoreSerialization._
+
+import scala.concurrent.Future
+
+abstract class BaseAppRepository[K, C, S](store: PersistenceStore[K, C, S], maxVersions: Int)(implicit
```
aquamatthias: See my comment above: all methods simply delegate to the persistence store, while resolving the implicits.

src/main/scala/mesosphere/marathon/stream/CollectionStage.scala
```diff
@@ -0,0 +1,48 @@
...
+import akka.stream.stage.{ GraphStageLogic, GraphStageWithMaterializedValue, InHandler }
+
+import scala.collection.mutable
+import scala.concurrent.{ Future, Promise }
+
+/**
+  * Akka Streaming Graph Stage that collects a set of values into the given collection
+  * Based on akka's SeqStage
+  */
+private final class CollectionStage[T, C](buf: mutable.Builder[T, C])
```
aquamatthias: nice!

src/main/scala/mesosphere/marathon/core/storage/repository/impl/BaseAppRepository.scala
```diff
@@ -0,0 +1,47 @@
+package mesosphere.marathon.core.storage.repository.impl
```
aquamatthias: Perhaps move this file to storage/impl

src/test/scala/mesosphere/marathon/core/storage/impl/InMemoryPersistenceStoreTest.scala
```diff
@@ -0,0 +1,31 @@
+package mesosphere.marathon.core.storage.impl
```
aquamatthias: move file to storage/impl/memory

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,141 @@
...
+    version: OffsetDateTime)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized]): Future[Done] = {
+
+    val storageId = ir.toStorageId(id, Some(version))
+    val currentId = ir.toStorageId(id, None)
+    lockManager.executeSequentially(id.toString) {
+      async {
+        val serialized = await(Marshal(v).to[Serialized])
+        await(rawGet(currentId)) match {
```
aquamatthias: This will store the value as current version, if not existent.
Could be surprising - If I say store this entry in this version, I would not expect to store it also without version.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,141 @@
...
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized],
+    um: Unmarshaller[Serialized, V]): Future[Done] = {
+    val storageId = ir.toStorageId(id, None)
+    lockManager.executeSequentially(id.toString) {
+      async {
+        val serialized = await(Marshal(v).to[Serialized])
+        val (_, _) = (
+          await(rawStore(storageId, serialized)),
+          await(rawStore(ir.toStorageId(id, Some(ir.version(v))), serialized)))
```
aquamatthias: There are cases, where we don't want to store a versioned entry (e.g.DeploymentPlan).
Perhaps add a class level flag, which indicates, if versions should be written or not.
We would create a PersistentStore for DeploymentPlans without versions.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,141 @@
...
+      case Some(v) =>
+        Some(await(Unmarshal(v).to[V]))
+      case None =>
+        None
+    }
+  }
+
+  protected def deleteOld(k: K, maxVersions: Int): Future[Done] = async {
+    val versions = await(rawVersions(k).toMat(Sink.seq)(Keep.right).run()).sortBy(_.toEpochSecond)
+    // we always store the current version (twice), once as a versioned node and once as the current one.
```
aquamatthias: The current version counts into versions?

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,141 @@
...
+    val storageId = ir.toStorageId(k, Some(version))
+    lockManager.executeSequentially(k.toString) {
+      rawDelete(ir.toStorageId(k, Some(version)), version)
+    }
+  }
+
+  protected def rawDeleteAll(k: K): Future[Done]
+
+  final override def deleteAll[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
+    val storageId = ir.toStorageId(k, None)
```
aquamatthias: remove this line.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,141 @@
...
+    id: Id)(implicit ir: IdResolver[Id, V, Category, K]): Source[OffsetDateTime, NotUsed] = {
+    rawVersions(ir.toStorageId(id, None))
+  }
+
+  protected def rawDelete(k: K, version: OffsetDateTime): Future[Done]
+
+  override def delete[Id, V](
+    k: Id,
+    version: OffsetDateTime)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
+    val storageId = ir.toStorageId(k, Some(version))
```
aquamatthias: remove this line

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,147 @@
...
+class ZkPersistenceStore(val client: RichCuratorFramework)(
+    implicit
+    mat: Materializer,
+    ctx: ExecutionContext,
+    scheduler: Scheduler,
+    val metrics: Metrics
+) extends BasePersistenceStore[ZkId, String, ZkSerialized]() with StrictLogging {
+
+  private val retryOn: Retry.RetryOnFn = {
+    case _: KeeperException.ConnectionLossException => true
```
aquamatthias: We currently abdicate on connection loss, which is probably to aggressive...

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,147 @@
...
+
+  private def retry[T](name: String)(f: => Future[T]) = Retry(name, retryOn = retryOn)(f)
+
+  override protected def rawIds(category: String): Source[ZkId, NotUsed] = {
+    val childrenFuture = retry(s"ZkPersistenceStore::ids($category)") {
+      async {
+        val buckets = await(client.children(s"/$category").recover {
+          case _: NoNodeException => Children(category, new Stat(), Nil)
+        }).children
+        val children = await(Future.sequence(buckets.map(b => client.children(s"/$category/$b").map(_.children))))
```
aquamatthias: I'm always a bit frightened by using Future.sequence while hitting zk (we had some P0 exactly for that). We definitively need a specific ExecutionContext for zk with a limited thread pool (Something like n=8 ?).

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,147 @@
...
+        children.flatten.map { child =>
+          ZkId(category, child, None)
+        }
+      }
+    }
+    Source.fromFuture(childrenFuture).mapConcat(identity)
+  }
+
+  override protected def rawVersions(id: ZkId): Source[OffsetDateTime, NotUsed] = {
+    val key = id.copy(version = None)
```
aquamatthias: rename: unversioned

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,147 @@
...
+          case Failure(e) =>
+            throw e
+        }
+      }
+    }
+
+    Source.fromFuture(versions).mapConcat(identity)
+  }
+
+  override protected[storage] def rawGet(k: ZkId): Future[Option[ZkSerialized]] =
```
aquamatthias: s/k/id

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,147 @@
...
+            None
+          case Failure(e: KeeperException) =>
+            throw new StoreCommandFailedException(s"Unable to get $k", e)
+          case Failure(e) =>
+            throw e
+        }
+      }
+    }
+
+  override protected def rawDelete(k: ZkId, version: OffsetDateTime): Future[Done] =
```
aquamatthias: s/k/id

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,147 @@
...
+          case Success(_) | Failure(_: NoNodeException) => Done
+          case Failure(e: KeeperException) =>
+            throw new StoreCommandFailedException(s"Unable to delete $k", e)
+          case Failure(e) =>
+            throw e
+        }
+      }
+    }
+
+  override protected def rawStore[V](k: ZkId, v: ZkSerialized): Future[Done] =
```
aquamatthias: s/k/id (since key has a different meaning in this context)

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
```diff
@@ -0,0 +1,163 @@
...
+    *         will fail the future with [[mesosphere.marathon.StoreCommandFailedException]]
+    */
+  def deleteAll[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done]
+
+  /**
+    * @return A source of _all_ keys in the Persistence Store (which can be used by a
+    *         [[mesosphere.marathon.core.storage.impl.cache.LoadTimeCachingPersistenceStore]] to populate the
+    *         cache completely on startup.
+    */
+  protected[storage] def keys(): Source[CategorizedKey[Category, K], NotUsed]
```
aquamatthias: I think the name is misleading (I misunderstood it the first time). Perhaps better: `idsByCategory()` ?

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+class LazyCachingPersistenceStore[K, Category, Serialized](
+    store: BasePersistenceStore[K, Category, Serialized])(implicit
+  mat: ActorMaterializer,
+    ctx: ExecutionContext) extends PersistenceStore[K, Category, Serialized] with StrictLogging {
+
+  private[storage] val idCache = RwLock(TrieMap[Category, Seq[Any]]())
+  private[storage] val valueCache = RwLock(TrieMap[K, Option[Any]]())
+
+  override def ids[Id, V]()(implicit ir: IdResolver[Id, V, Category, K]): Source[Id, NotUsed] = {
+    val storeId = ir.category
```
aquamatthias: s/storeId/category

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+  * @param ctx The execution context for future chaining.
+  * @tparam K The persistence store's primary key type
+  * @tparam Serialized The serialized format for the persistence store.
+  */
+class LazyCachingPersistenceStore[K, Category, Serialized](
+    store: BasePersistenceStore[K, Category, Serialized])(implicit
+  mat: ActorMaterializer,
+    ctx: ExecutionContext) extends PersistenceStore[K, Category, Serialized] with StrictLogging {
+
+  private[storage] val idCache = RwLock(TrieMap[Category, Seq[Any]]())
```
aquamatthias: Nit: to distinguish type from value I think it easier to read `TrieMap.empty[A,B]`

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+  private[storage] val valueCache = RwLock(TrieMap[K, Option[Any]]())
+
+  override def ids[Id, V]()(implicit ir: IdResolver[Id, V, Category, K]): Source[Id, NotUsed] = {
+    val storeId = ir.category
+    idCache.read { ids =>
+      if (ids.contains(storeId)) {
+        Source(ids(storeId).map(_.asInstanceOf[Id]))
```
aquamatthias: Do you have to create a new `Seq` here?
The cheaper cast would be `ids(storeId).asInstanceOf[Seq[Id]]`
Or if you want to make it type safe, take `collect`

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+
+  override def store[Id, V](id: Id, v: V)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized],
+    um: Unmarshaller[Serialized, V]): Future[Done] = {
+    async {
+      await(store.store(id, v))
+      val storageId = ir.toStorageId(id, None)
+      val storageCategory = ir.category
+      valueCache.write { values =>
```
aquamatthias: I think this will not guarantee the correct order of updates in the cache, since the `store` operation is performed asynchronously. In the EntityStoreCache we introduced an update closure, that was called inside the future:
https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/state/EntityStoreCache.scala#L51

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+/**
+  * A Write Ahead Cache of another persistence store that lazily loads values into the cache.
+  *
+  * @param store The store to cache
+  * @param mat a materializer for Akka Streaming
+  * @param ctx The execution context for future chaining.
+  * @tparam K The persistence store's primary key type
+  * @tparam Serialized The serialized format for the persistence store.
+  */
+class LazyCachingPersistenceStore[K, Category, Serialized](
```
aquamatthias: This cache will also cache versioned entries.
I don't see a problem than higher resource demands.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized],
+    um: Unmarshaller[Serialized, V]): Future[Done] = {
+    async {
+      await(store.store(id, v))
+      val storageId = ir.toStorageId(id, None)
+      val storageCategory = ir.category
+      valueCache.write { values =>
+        idCache.write { ids =>
+          values.put(storageId, Some(v))
```
aquamatthias: The underlying store will store the object under 2 ids: 
- the versioned entry
- the unversioned entry

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -0,0 +1,142 @@
...
+      Done
+    }
+  }
+
+  override def store[Id, V](id: Id, v: V, version: OffsetDateTime)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized]): Future[Done] = async {
+    await(store.store(id, v, version))
+    valueCache.write { vc =>
+      vc.putIfAbsent(ir.toStorageId(id, None), Some(v))
```
aquamatthias: Not sure if this is intended - see my comment on zkStore

aquamatthias:
@jasongilanfarr That's a big PR - I reviewed all 11 commits. (Perhaps we can do changes like Seq->immutable.Seq etc. on a separate branch in the future).
I added a couple of suggestions and comments, but the base really looks very good!

Things to do
0) rebase and repair the tests
1) Write a `storage/StorageModule` where you do the wiring of this component
2) Wire the new persistence layer and let the integration tests run
3) Move/Wire all other repos
4) Try to use create a big number of apps, tasks and see if it performs well
5) Perhaps you can move the old implementation to the new structure (might be needed for the migration logic)
6) Implement a migration that translates the old zk layout to the new layout
7) Make the new store the default store

src/main/scala/mesosphere/marathon/core/storage/MarathonState.scala
jasongilanfarr: I actually split the existing MarathonState in half, fromProto really belongs in the companion not the instance. This is actually used by the ZK typeclasses, marshaling is free, unmarshalling just requires the companion object to be specified.

And they do implement it already ;)

src/main/scala/mesosphere/marathon/core/storage/PersistenceStore.scala
jasongilanfarr: I still need to move this one. Ids() is actually idsByCategory(). Keys() is all of the ids. That is confusing. This method is exclusively used by the load time cache.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
jasongilanfarr: This was kind of a weird oddity but it actually avoids a lot of extra gets and makes store() a bit easier - since its expected to keep the existing one if there are any.

versions is supposed to list the current version as well as the historical (I didn't originally do it this way) and that's consistent with the current MarathonStore.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
jasongilanfarr: If maxVersions is 0, only the current version will actually be stored...

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: The versioned calls pass through to the underlying store, I'll verify that versions aren't stored, but quite sure this cache only stores the current version.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Collect wont work anyways as its type erased.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Yep, not concerned. We only cache the unversioned form.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: I may be misunderstanding await() as my impression is the code that follows it is executed after. I'll check this to make sure my assumption is correct.

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkStoreSerialization.scala
jasongilanfarr: Agreed, still trying to figure out how to do that effectively. This is only the default value, I'll make that more clear.

src/main/scala/mesosphere/marathon/core/storage/repository/AppRepository.scala
jasongilanfarr: Not entirely sure. I think GroupManager just handle the repo ultimately.

src/main/scala/mesosphere/marathon/core/storage/repository/AppRepository.scala
jasongilanfarr: I was thinking the same thing and struggling with implicit not found :(

src/main/scala/mesosphere/marathon/stream/CollectionStage.scala
jasongilanfarr: This was the "coolest" change and reminded me why the collection API is so nice. Wish akka did this instead of only doing it for seq.

jasongilanfarr:
Sorry about the size, I intended on reordering the commits into smaller PRs but I've also been feeling a bit crunched for time. The immutable seq changes came from making AppRepository a common interface and changing its signatures a bit. I could have done that separately (reordering commits).

I try to use preview as "here's the entire set of changes so far, I'll reorder the commits from my working view into a merge able view later" - I didn't have that much time to do the latter and got my branches mixed together :/

Agreed on remaining steps. I'd still like to cleanup repository and make it fully reusable without all the sub classing- I think it can be done but I've struggled with the implicit resolution.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: "the main thing to understand about async/await is that it's just another way of doing Future's flatMap. there's no concurrency magic behind the scenes. all the calls inside an async block are sequential, including await which doesn't actually block the executing thread but rather wraps the rest of the async block in a closure and passes it as a callback on completion of the Future we're waiting on. so in the first piece of code the second calculation doesn't start until the first await has completed since no one started it prior to that."

So, this should work as expected no? I think I can add a little more parallelism elsewhere by having things like:

val storeCurrent = rawStore()
val storeAsVersioned = rawStore()
(await(storeCurrent), await(storeAsVersioned))

src/main/scala/mesosphere/marathon/MarathonModule.scala
jasongilanfarr: 100% agreed. That's coming.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: It is actually intended. store(id, v, version) says "if there is no current, make it current". I'll add a note to that effect.

src/main/scala/mesosphere/marathon/core/storage/impl/zk/ZkPersistenceStore.scala
jasongilanfarr: I'll add this soon. I think we also need to pass in an ACL for create()

src/main/scala/mesosphere/marathon/core/storage/repository/impl/BaseAppRepository.scala
jasongilanfarr:
```
storage {
   impl {
      zk {}
      cache {}
   }
   repository {
      impl {}
   }
} 
``` 
seems more easy to navigate then
```
storage {
   impl {
      zk {}
      cache {}
      repository {
      }
   }
   repository {
   }
}
```

src/main/scala/mesosphere/marathon/core/storage/MarathonState.scala
aquamatthias: I see it the same way as a separate Unmarshaller/Marshaller. 
Even if we implement it directly (as it is now) we should not force this.

src/main/scala/mesosphere/marathon/core/storage/impl/BasePersistenceStore.scala
aquamatthias: Perhaps execute storeVersion and deleteOld only if needed.

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
aquamatthias: Ah ok - cool

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
aquamatthias: That's true

src/main/scala/mesosphere/marathon/core/storage/impl/cache/LazyCachingPersistenceStore.scala
aquamatthias: My point is more:
Thread A: store("a", "one) => store "one" in zk async and then store in cache
Thread B: store("a", "two") => store "two" in zk async and then store in cache
The base repo will force an order (executeSequential) but this order is not maintained in the cache due to the async operation, or?
So it is possible that zk stores "one" and the cache stores "two"

jasongilanfarr: That does make sense. Perhaps we can remove the RwLock and instead use the lock manager here too...

src/main/scala/mesosphere/marathon/core/storage/MarathonState.scala
jasongilanfarr: I'll go ahead and rip this out too (and instead just implement the marshallers directly)

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -291,9 +291,9 @@ trait MarathonConf
   //Internal settings, that are not intended for external use
   lazy val internalStoreBackend = opt[String](
     "internal_store_backend",
-    descr = "The backend storage system to use. One of zk, mesos_zk, mem.",
+    descr = "The backend storage system to use. One of zk, mesos_zk, mem, new_zk, new_mem.",
```
aquamatthias: terms like new and old get old pretty quick.
Do we still have the idea to make the "what is zk" today available?
Perhaps we should rename like:
zk->twitter_zk
mem-> (remove)
new_zk -> zk
new_mem -> mem

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
```diff
@@ -0,0 +1,316 @@
...
+            ctx: ExecutionContext, scheduler: Scheduler): PersistenceStore[K, C, S] = {
+    cacheType match {
+      case NoCaching => leafStore
+      case LazyCaching => new LazyCachingPersistenceStore[K, C, S](leafStore)
+      case EagerCaching => new LoadTimeCachingPersistenceStore[K, C, S](leafStore)
+    }
+  }
+}
+
+case class NewZk(
```
aquamatthias: proposal: CuratorZK

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
```diff
@@ -0,0 +1,316 @@
...
+
+    val client = ZkClient(connector)
+      .withAcl(creationAcl)
+      .withRetries(retries)
+    val compressionConf = CompressionConf(enableCompression, compressionThreshold.toBytes)
+    new ZKStore(client, client(zkPath), compressionConf)
+  }
+}
+
+object ClassicZk {
```
aquamatthias: Proposal: TwitterZK

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
```diff
@@ -0,0 +1,316 @@
...
+  }
+}
+
+case class ClassicInMem(maxVersions: Int) extends StorageConfig {
+  def store: PersistentStore = new InMemoryStore()
+  val enableCache: Boolean = false
+}
+
+object ClassicInMem {
+  def apply(config: AllConf): ClassicInMem =
```
aquamatthias: Should take LegacyStorageConfig

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
```diff
@@ -0,0 +1,316 @@
...
+    val client = ZkClient(connector)
+      .withAcl(creationAcl)
+      .withRetries(retries)
+    val compressionConf = CompressionConf(enableCompression, compressionThreshold.toBytes)
+    new ZKStore(client, client(zkPath), compressionConf)
+  }
+}
+
+object ClassicZk {
+  def apply(config: AllConf): ClassicZk =
```
aquamatthias: Should take the specific Config?

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
```diff
@@ -0,0 +1,316 @@
...
+      timeout.toMillis,
+      TimeUnit.MILLISECONDS,
+      zkPath
+    )
+    new MesosStateStore(state, timeout)
+  }
+}
+
+object MesosZk {
+  def apply(config: AllConf): MesosZk =
```
aquamatthias: Should take the specific Config?
Idea: AllConf should never be used in the components.

src/main/scala/mesosphere/marathon/core/storage/StorageModule.scala
```diff
@@ -0,0 +1,24 @@
...
+import mesosphere.marathon.metrics.Metrics
+
+import scala.concurrent.ExecutionContext
+
+trait StorageModule {
+  def appRepository: AppRepository
+}
+
+object StorageModule {
+  def apply(conf: AllConf, metrics: Metrics, mat: Materializer, ctx: ExecutionContext,
```
aquamatthias: Should take the StorageConf

src/main/scala/mesosphere/marathon/core/storage/StorageModule.scala
```diff
@@ -0,0 +1,24 @@
...
+            scheduler: Scheduler): StorageModule = {
+
+  }
+}
+
+private[storage] class StorageModuleImpl extends StorageModule {
```
aquamatthias: Why is this needed?

src/main/scala/mesosphere/marathon/MarathonConf.scala
jasongilanfarr: I agree. I don't think we can realistically change "zk -> twitter_zk" as it is still the default store. I was tempted to use "legacy_zk". Perhaps the new one should be "zk2"?

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
jasongilanfarr: Sure.

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
jasongilanfarr: Sure.

jasongilanfarr:
I'm working now (on a separate branch) on the following:

- Migrate over GroupRepository
- Move all of the EntityRepositories into `marathon/core/storage/...`
- Cleanup all of Repository (new and old) to have a single, consistent interface, e.g.:
- Bring Migration into `marathon/core/storage/migration` and make it internal to the StorageModule (so that we can kill off the storage layers used for legacy migration once its up).

```
trait Repository[Id, T] {
  def ids(): Source[Id, NotUsed]
  def all(): Source[T, NotUsed]
  def get(id: Id): Future[Option[T]]
  def delete(id: Id): Future[Done]
  def store(v: T): Future[Done]
}

trait VersionedRepository[Id, T] extends Repository[Id, T] {
  def versions(id: Id): Source[OffsetDateTime, NotUsed]
  def getVersion(id: Id, version: OffsetDateTime): Future[Option[T]]
  def storeVersion(v: T): Future[Done]
}

trait AppRepository extends VersionedRepository[PathId, AppDefinition]

class PersistenceStoreRepository[...](...) extends Repository[...]
class PersistenceStoreVersionedRepository[...](...) extends PersistenceStoreRepository[...](...) with VersionedRepository[...]

class AppZkRepository extends PersistenceStoreVersionedRepository[...] with AppRepository
...

class EntityRepository[...](...) extends Repository[...]
class AppEntityRepository[...](...) extends EntityRepository with AppRepository
...

```

Guidance:
- The new Persistence Store and its helpers are in `marathon/core/storage`
- Anything in `marathon/state` is the old storage
- XRepository exists as a bridge to support both from the outside.
- All the signatures have changed to bring consistency between them.
  - Future[Iterable[T]] is generally replaced with Source[T, NotUsed], this will allow us to more easily support backing stores that support cursors.

src/main/scala/mesosphere/marathon/core/storage/IdResolver.scala
```diff
@@ -0,0 +1,40 @@
...
+  * @tparam Id         The marathon type of the ID for the given Value type
+  * @tparam K          The persisted type of the ID
+  * @tparam Category   The category that 'V' belongs to.
+  * @tparam V          The value type being stored
+  */
+trait IdResolver[Id, V, Category, K] {
+  /**
+    * Translate the marathon id into the given persisted format
+    */
+  def toStorageId(id: Id, version: Option[OffsetDateTime]): K
```
aquamatthias: You use OffesetDateTime instead of Timestamp. What is the reasoning behind this?

src/main/scala/mesosphere/marathon/core/storage/repository/AppRepository.scala
aquamatthias: So we replace all occurrences of `AppEntityRepository` with `AppRepository` and then we can remove `AppEntityRepository`?

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -340,42 +343,41 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
...
-    metrics: Metrics): TaskFailureRepository = {
-    new TaskFailureRepository(store, conf.zooKeeperMaxVersions.get, metrics)
+    metrics: Metrics): TaskFailureEntityRepository = {
+    new TaskFailureEntityRepository(store, conf.zooKeeperMaxVersions.get, metrics)
   }
 
   @Provides
   @Singleton
-  def provideAppRepository(
+  def provideAppEntityRepository(
```
aquamatthias: I hope we can get rid of this.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -340,42 +343,41 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
   @Singleton
   def provideTaskFailureRepository(
```
aquamatthias: I hope we can get rid of all the repositories here in this module and do the wiring in the `StorageModule`. You can reexport the objects via `CoreGuiceModule`.

jasongilanfarr: They are all coming from StorageModule. I'm still working through moving migration into storage module and writing the migration from zk -> zk2.

src/main/scala/mesosphere/marathon/MarathonModule.scala
jasongilanfarr: All of these injections are going away. Migrations will be the only consumer of the raw "XEntityRepository" - storageModule will only expose the regular repositories - and AppRepository will be read-only (GroupRepository is the only writer of AppRepository)

src/main/scala/mesosphere/marathon/core/storage/IdResolver.scala
jasongilanfarr: OffsetDateTime is part of Java8 Time and it has a particular advantage in utilizing the timezone while still being in UTC. Java8 Time is essentially joda2.

aquamatthias:
@jasongilanfarr Looks really nice!

Can you try to make a list of things that are missing:

- [ ] the usage of the `XXXRepository` trait whenever `XXXEntityRepository` is used
- [ ] the wiring of the new stores
- [ ] the migration from the old layout to the new layout
- [ ] remove the `XXXEntityRepository`

jasongilanfarr:
Definitely, I thought the "big comment" on guidance above did that, but I missed the checkboxes. Here's whats left with the status based on my group branch.

- [ ] Migrate GroupRepository
- [ ] Make AppRepository Read Only, Group Repository is allowed to write into AppRepository
- [ ] Implement StoredGroup where apps are stored in groups only by id
- [ ] Implement common base class for Repository/VersionedRepository and use that for the "old style" storage as well
- [ ] Move Migration into StorageModule and allow only it to access the specific "new/old" Repositories.
- [ ] Fix the few broken tests.
- [ ] Write a generic XRepository and GroupRepository test, replacing the existing ones.
- [ ] Remove Injections for things in the storage module, exporting only what is strictly required by Guice via CoreGuiceModule

PR #4098, opened 2016-07-14
Refactored 'GroupManager' as an actor.

nfnt:
Following the pattern in the `mesosphere.marathon.core` namespace,
'GroupManager' is refactored as a module. The existing behavior hasn't been
changed and leave room for further refactoring. While 'GroupManager' is now an
actor, it doesn't persist any state yet. Further work could focus on that.


aquamatthias:
@nfnt There is a cyclic dependency - so Marathon can not start:
```
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344: 
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344: 1) Tried proxying mesosphere.marathon.MarathonSchedulerService to support a circular dependency, but it is not an interface.
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344:   at mesosphere.marathon.MarathonModule.configure(MarathonModule.scala:77)
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344:   while locating mesosphere.marathon.MarathonSchedulerService
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344:     for parameter 13 at mesosphere.marathon.core.CoreModuleImpl.<init>(CoreModuleImpl.scala:44)
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344:   while locating mesosphere.marathon.core.CoreModuleImpl
[07:36:11][Step 3/3] [2016-07-15 07:37:27,044] INFO [] marathon_52344:   at mesosphere.marathon.core.CoreGuiceModule.configure(CoreGuiceModule.scala:145)
```
Can you take a look?

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerActor.scala
```diff
@@ -0,0 +1,262 @@
...
+  private var scheduler: DeploymentService = _
+
+  override def preStart(): Unit = {
+    super.preStart()
+    scheduler = schedulerProvider.get()
+  }
+
+  override def receive: Receive = {
+    Seq(
+      receiveApp,
```
aquamatthias: You basically translate every method into a receive function.
Multiple receive functions make sense, if you have different behaviors or complex receive functions.

- it is not easy to see which messages get handled
- you tangle messaging logic and business logic
- every receive function directly handles the logic
- it is not easy to test the domain logic

I would suggest this:

```scala
def receive: Receive = {
  case GetRootGroup => getRootGroup().pipeTp(sender())
  case GetGroupWithId(id) => getGroupWithId(id).pipeTo(sender())
}

def getRootGroup(): Future[Group] = ???

```

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerDelegate.scala
```diff
@@ -0,0 +1,145 @@
...
+    implicit val timeoutImplicit: Timeout = timeout
+
+    val answerFuture = actorRef ? message
+    answerFuture
+  }
+}
+
+private[impl] object GroupManagerDelegate {
+
+  sealed trait Request
```
aquamatthias: Please define the messages send to the actor in the actor's companion object.
So it is easy to see, which message types are handled by this actor.

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerDelegate.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  case class App(id: PathId) extends Request
```
aquamatthias: Messages mean actions. They should describe the action: ==> `GetAppWithId(id: PathId)`

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerDelegate.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  case class GroupWithId(id: PathId) extends Request
```
aquamatthias: GetGroupWithId

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerDelegate.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  case class GroupWithVersion(id: PathId, version: Timestamp) extends Request
```
aquamatthias: GetGroupWithVersion

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerDelegate.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  case object RootGroup extends Request
```
aquamatthias: GetRootGroup

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerDelegate.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  case class Upgrade(
+    gid: PathId,
+    change: Group => Group,
+    version: Timestamp = Timestamp.now(),
+    force: Boolean = false,
+    toKill: Map[PathId, Iterable[Task]] = Map.empty) extends Request
+
+  case class Versions(id: PathId) extends Request
```
aquamatthias: GetAllVersions

src/main/scala/mesosphere/marathon/core/groupmanager/package.scala
```diff
@@ -0,0 +1,6 @@
+package mesosphere.marathon.core
+
+/**
+  * This package allows to manage all group related actions.
+  */
+package object groupmanager
```
aquamatthias: why do you need this file?

aquamatthias: The service `GroupManager` now is in package `groupmanager`.
We should find a better name for the package. 
Could be as simple as `group`

aquamatthias:
@nfnt Thanks for this PR. Added a couple of comments. Please rebase.

src/main/scala/mesosphere/marathon/core/groupmanager/package.scala
nfnt: Sorry, I just followed the pattern of some other modules in `core` without giving it too much thought. Better remove that file then?

src/main/scala/mesosphere/marathon/core/groupmanager/GroupManagerConfig.scala
```diff
@@ -0,0 +1,13 @@
+package mesosphere.marathon.core.groupmanager
+
+import org.rogach.scallop.ScallopConf
+
+trait GroupManagerConfig extends ScallopConf {
+  //scalastyle:off magic.number
+
+  lazy val groupManagerRequestTimeout = opt[Int](
+    "group_manager_request_timeout",
```
jasongilanfarr: Please specify the unit, e.g. "ms"

src/main/scala/mesosphere/marathon/core/groupmanager/GroupManagerConfig.scala
```diff
@@ -0,0 +1,13 @@
...
+    descr = "Foo",
+    hidden = true,
+    default = Some(1000))
```
jasongilanfarr: 10.seconds.toMillis?

src/main/scala/mesosphere/marathon/core/groupmanager/package.scala
aquamatthias: I don't see much value - please remove.
Also: please rename the package to `group` instead of `groupmanager`

src/main/scala/mesosphere/marathon/core/groupmanager/GroupManagerConfig.scala
```diff
@@ -0,0 +1,13 @@
+package mesosphere.marathon.core.groupmanager
+
+import org.rogach.scallop.ScallopConf
+
+trait GroupManagerConfig extends ScallopConf {
+  //scalastyle:off magic.number
+
+  lazy val groupManagerRequestTimeout = opt[Int](
+    "group_manager_request_timeout",
+    descr = "Foo",
```
aquamatthias: Not very good description.

src/main/scala/mesosphere/marathon/core/groupmanager/GroupManagerConfig.scala
aquamatthias: All command line flags are configured in ms and don't have the unit in the name. But the description should have the unit.

src/main/scala/mesosphere/marathon/core/groupmanager/GroupManagerConfig.scala
nfnt: Whoops! Thanks for catching that!

src/main/scala/mesosphere/marathon/core/groupmanager/GroupManagerConfig.scala
aquamatthias: Very good point - also this is one second, but should be 10

aquamatthias:
Please rebase

src/main/scala/mesosphere/marathon/core/groupmanager/impl/GroupManagerActor.scala
```diff
@@ -0,0 +1,279 @@
...
+import mesosphere.marathon.upgrade.{ DeploymentPlan, GroupVersioningUtil, ResolveArtifacts }
+import mesosphere.util.CapConcurrentExecutions
+import org.slf4j.LoggerFactory
+
+import scala.collection.immutable.Seq
+import scala.collection.mutable
+import scala.concurrent.Future
+import scala.util.{ Failure, Success }
+
+private[groupmanager] object GroupManagerActor {
```
jasongilanfarr: While the pipeTo is nice, it seems a bit off to me- as it stands there is little reason to actually have the GroupManager as an actor since everything it does is purely future based, so the Single Threaded Apartment has almost no value.

Instead, for every `Request`, there should be a corresponding `Response` message type (unless its "Done") and the Future methods should change the actor's internal state while they are being processed. I think the latter is out of the scope of this change, but it would be good to adapt a `Response` type and use it.

This is mostly a cosmetic comment, but there is some value in it. Perhaps, simply adding "Replies with T" to all of the request types would help?

aquamatthias: @jasongilanfarr this is only the first step. With an actor in place, we can hold the rootGroup, dependencyTree etc. Then the actor will make more sense.

@nfnt Can you add the `Response` types and document which request will send which response?

aquamatthias:
@nfnt LGTM. One comment regarding response types.

aquamatthias:
Thanks!

PR #4101, opened 2016-07-15
Fixes #4092 Logging with Elasticsearch and Kibana

### LINKED ISSUE #4092: Logging with Elasticsearch and Kibana
Currently Marathon uses logback to process logs. Logs pattern could be customized with `JAVA_OPTS` that provides path to custom logback configuration. Unfortunately to use JSON layout and send logs to Elastic and Kibana user must somehow provide missing jars to classpath which could be done in three ways:

1. Wrap marathon jar with custom jar containing dependencies
2. Maintain custom Marathon fork with dependencies
3. Put dependencies inside classpath

All options need user action on upgrade. It'll be cool if Marathon allow to send JSON formatted logs directly to elastic at least choose JSON format so user don't need to parse them.

Related: #4087

janisz:
Added new debug option in command line flags `--logstash`.
User can specify edpoint `<host>[:<port>]` and logs will be formated
with JSON and sent to that endpoint over TCP.

### DIFF
--- a/src/main/scala/mesosphere/marathon/DebugConf.scala
+++ b/src/main/scala/mesosphere/marathon/DebugConf.scala
@@ -1,11 +1,15 @@
 package mesosphere.marathon
 
+import java.net.URI
 import javax.inject.Provider
 
-import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.{ AsyncAppender, Level, LoggerContext }
+import ch.qos.logback.core.net.ssl.SSLConfiguration
 import com.google.inject.AbstractModule
 import com.google.inject.matcher.{ AbstractMatcher, Matchers }
 import mesosphere.marathon.metrics.{ MetricPrefixes, Metrics }
+import net.logstash.logback.appender._
+import net.logstash.logback.composite.loggingevent.ArgumentsJsonProvider
 import org.aopalliance.intercept.{ MethodInterceptor, MethodInvocation }
 import org.rogach.scallop.ScallopConf
 import org.slf4j.{ Logger, LoggerFactory }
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/DebugConf.scala
```diff
@@ -106,4 +119,29 @@ class DebugModule(conf: DebugConf) extends AbstractModule {
...
+
+  private def configureLogstash(hostPort: String) {
+    val context = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
+
+    val encoder = new net.logstash.logback.encoder.LogstashEncoder()
+    encoder.setContext(context)
+    encoder.addProvider(new ArgumentsJsonProvider())
+    encoder.start()
+
+    val logstashAppender = new LogstashTcpSocketAppender()
```
jasongilanfarr: What if logstash is listening on SSL?

jasongilanfarr:
Looks fairly good, but I wonder if it should support syslog/udp/tcp/ssl as well? I would argue UDP is likely better for logging in most cases.

My personal view is that we should support logging into Kafka instead of directly into logstash - I believe we are considering a Kafka appender in the near future.

src/main/scala/mesosphere/marathon/DebugConf.scala
janisz: I'm sorry I forget about it. I'll add another flag `--logstash_scheme` with `udp`,`tcp` options or (I think better) modify `--logstash` to accept `[udp|tcp|ssl]://<host>[:<port>]`.

janisz:
@jasongilanfarr good point. I think only syslog/udp should be supported, TCP/SSL is nice to have. Supporting syslog/udp will enable users to setup adapter (e.g., syslog-ng, nxlog) that will enhance logs (e.g., add custom fields) and push them to final destination (e.g., kafka, elastic, graylog), providing standard way of getting structured logs from Marathon.

janisz:
@jasongilanfarr I updated PR. Added UDP/TCP/SSL support.

aquamatthias:
Hey @janisz Thanks for this PR - looks good to me!

PR #4102, opened 2016-07-15
docs: add HTTPS healthcheck option

air:


jdef:
how does our HTTPS health check handle certificate verification? or does it disable such?

aquamatthias:
The health check actor will not verify any certificate. See the `BlindFaithX509TrustManager`. That's probably something we should add to the docs.
Also: Please rebase.

aquamatthias:
Rebased. Added a note regarding ssl validation.

PR #4103, opened 2016-07-15
Made `dependencyGraph` be only called from the root group.

mpark:
There are a couple of goals met with this PR.

(1) As of today, `Group.dependencyGraph` is computed 2600 times when I deploy 50 independent apps all under the root group. This patch makes `Group.dependencyGraph` as a `lazy val`, which reduces this to 100 invocations.

(2) In order for us to store a __single__ dependency graph that we update as things change, we need to ensure that `Group.dependencyGraph` is only ever called on the root group.

(3) This patch makes (2) become true, by fixing the following validation check:
```
```
Because the `or` is not a `||`, the expression on the right is always evaluated as we recurse down the `Group` tree. This causes the dependency graph of all group subtrees to be computed. This didn't help with my benchmark since I was deploying independent apps all under the root group, but it's critical invariant for us to achieve in order to get to (2).

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -148,6 +148,7 @@ case class Group(
   }
 
   def dependencyGraph: DirectedGraph[AppDefinition, DefaultEdge] = {
+    require(id.isRoot)
     val graph = new DefaultDirectedGraph[AppDefinition, DefaultEdge](classOf[DefaultEdge])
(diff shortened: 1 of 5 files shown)


aquamatthias:
@mpark Great! Thanks!

aquamatthias:
cherry picked in 1.2 branch

aquamatthias:
@mpark We should probably report a bug for the `or` implementation in https://github.com/wix/accord/

Issue #3765, opened 2016-04-13
Allow allocation of persistent volumes on non-root mesos disk resources


timcharper:
Mesos supports multiple-disk resources of the following kinds of disk resources: "Root", "Path", and "Mount". In the present release candidate, it is impossible to allocate a persistent volume on a disk resource other than root. Marathon will see non-root disk resources offered, and completely ignore them.

I believe this could be resolved somewhat easily by:

- Adding an optional string parameter, `kind`, to [PersistentVolumeInfo](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/state/Volume.scala#L114) (and updating serializers / json-schema)
- Updating [ResourceMatcher](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/ResourceMatcher.scala#L101) to consider non-Root Resource offers (currently, my experience indicates these are ignored, but I can't exactly tell where, in the code, they are being filtered out). I would be inclined to make it such that if `kind` were unspecified, then a root volume is used, although there is no reason to not consider path disk resources as well. (the biggest concern would be allocating an entire disk to a process that doesn't need / want an entire disk).
- Updating OfferOperationFactory [createVolumes](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala#L57) and [reserve](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala#L33) to specify the appropriate disk source for the selected disk resource(s).

Am I totally off in thinking this should be somewhat easy to do? I'm an experienced Scala programmer and am willing to take a stab at it, but want to discuss / solicit any guidance before starting. This feature would drastically simplify my mesos cluster deployment.

meichstedt:
Hi @timcharper, thanks for offering your help here, we'd be happy about your contribution.

As for the current state, I am not sure why disks other than `Root` are not used; they are not filtered out explicitly.

For `Path` disks, the approach sounds good and I agree, that seems rather easy to implement. For `Mount` disks, however, I'm not sure about the simplicity or the best approach. There are things to consider:

- `Mount` disks cannot be carved up into smaller chunks. If the disk has 100GB, how can the volume specify _take all_? It will get the whole disk, but it needs to know how big that is ...
- therefore, a `Mount` disk seems somewhat special (think fast SSD)
[...]

PR #4107, opened 2016-07-16
Support for creating persistent volumes on path or mount disk resources

timcharper:
This PR is ready for further review.

- DONE
  - Update API documentation
  - when reserving mount disk, set allocation size to be disk size, not requested size. (consider setting over-allocation threshold? Don't allocate a disk 20% larger than I requested?)
  - (decision: no) Consider allowing `unspecified` to match `path` or `root` disk resources. (more complex to implement, wasn't sure if it would be a needed use case, so I didn't implement it).
  - discuss and nail down allocation priorities. Do we fit small volumes in small places like we're packing a suitcase? Or do we prefer to spread utilization?
  - allocate largest volumes first (prevent marathon from "allocating itself into a corner")
  - implement path LIKE / UNLIKE constraints.
  - allocate space across multiple resources for the same disk. (IE two resources, one prereserved, the other not, both for the same disk).
  - add tests.
- NOT DONE
  - ~~Update Marathon-UI to support the new fields.~~ (let's move to another task when this is merged)


timcharper:
This PR addresses issue https://github.com/mesosphere/marathon/issues/3765

timcharper:
The [test that is failing on CI](https://travis-ci.org/mesosphere/marathon/builds/145325104) passes when I run it locally:

```
> test-only mesosphere.marathon.core.task.tracker.impl.TaskOpProcessorImplTest
// snip ...
Run completed in 3 seconds, 562 milliseconds.
Total number of tests run: 9
Suites: completed 1, aborted 0
Tests: succeeded 9, failed 0, canceled 0, ignored 0, pending 0
All tests passed.
[info] Run completed in 3 seconds, 562 milliseconds.
[info] Total number of tests run: 9
[info] Suites: completed 1, aborted 0
[info] Tests: succeeded 9, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] Total time: 14 s, completed Jul 17, 2016 11:14:30 AM
```

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -111,14 +114,79 @@ object DockerVolume {
...
 
 object PersistentVolumeInfo {
+  import DiskType._
+  import scala.collection.JavaConversions._
   def fromProto(pvi: Protos.Volume.PersistentVolumeInfo): PersistentVolumeInfo =
-    new PersistentVolumeInfo(pvi.getSize)
+    new PersistentVolumeInfo(
+      size = pvi.getSize,
+      `type` = if (pvi.hasType)
+      pvi.getType match {
```
jdef: default match case?

timcharper: good catch.

timcharper:
I was having too many issues with `1.2.0RC4` so I rebased the patch to be applied on top of `1.1.2`

timcharper:
@jdef we've been running this patch for a week now and it works, I'd certainly be appreciative of some dialogue and feedback with the direction I've chosen to go! Specifically, are you okay with the implicit classes I've created for mesos resources? Do you feel okay about making ScalarResourceMatch a sealed trait and implementing DiskResourceMatch? Do you feel okay with the match algoirthms? Do we need more comments? etc.

Since we have a stable patched version of this patch that works for us, I can work on integrating it with 1.2.0, which we'll switch to when the blocker bugs (such as 4137 I'd just reported) is fixed. I would just like a little feedback before I commence.

Thank you!

meichstedt:
@timcharper I'm about to review this within the next 2 days, but I'd also appreciate if anyone has time to take a look.

timcharper:
@meichstedt gentle nudge on this one; there's still some further code changes to be made (for example, right now, you have to specify the exact size of the persistent mount disk you would like to use, or it fails). I'd like to get validation / guidance on my general direction before I go much further.

We been running with it in production now for several weeks and it has worked well.

meichstedt:
@timcharper thanks for the nudge ;)
I've been on vacation for two weeks and it seems like nobody wanted to take my part in this. Sorry!

I have some initial loose feedback put together hastily on the train, I think I can share at least that part.

Btw, in case you haven't noticed, we're about to plan a [v3 API](https://docs.google.com/document/d/1uPH58NWN_OuynptsqTOq8v5qlkivq2mUb2M9J5jZTMU/edit#) – feedback welcome as well.

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
```diff
@@ -35,17 +35,24 @@ class TaskOpFactoryHelper(
...
+  /**
+    * Returns a set of operations to reserve ALL resources (cpu, mem, ports, disk, etc.) and then create persistent
+    * volumes against them as needed
+    */
   def reserveAndCreateVolumes(
     frameworkId: FrameworkId,
     newTask: TaskStateOp.Reserve,
     resources: Iterable[Mesos.Resource],
-    localVolumes: Iterable[LocalVolume],
+    localVolumes: Iterable[(Option[Mesos.Resource.DiskInfo.Source], LocalVolume)],
```
meichstedt: 1) in general, I dislike tuples in interfaces. I could lie with it here, but it's but too obvious at first glance why the first one is an option

2) please try to prevent using generated Mesos proto classes when possible. In this case, the lack of an enum value ~ root makes this an Option. It would be nicer to get rid of the option and use a dedicated Marathon type (why not the existing `DiskType`?)

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -27,8 +30,14 @@ object ResourceMatcher {
...
     def scalarMatch(name: String): Option[ScalarMatch] = scalarMatches.find(_.resourceName == name)
 
-    def resources: Iterable[org.apache.mesos.Protos.Resource] =
-      scalarMatches.flatMap(_.consumedResources) ++ portsMatch.resources
+    def resources: Iterable[Protos.Resource] =
+      scalarMatches.flatMap(_.consumedResources) ++
+        portsMatch.resources
+
+    // TODO - this assumes that volume matches are one resource to one volume, which should be correct, but may not be.
+    // TODO - don't use this because it doesn't return reservation info !!!
```
meichstedt: Not sure I get the meaning of this TODO – can you elaborate please?

timcharper: I actually think I fixed this but didn't get rid of the TODO.  I was thinking that it might be possible to get multiple offers for the same disk resource (as in, half of the space already allocated to the principal, but the other was not allocated at all)

meichstedt:
Some random comments to your above TODOs:
- I wouldn't consider `unspecified` to match `path` or `root` disk resources as of now, thinking no one would create path disks explicitly if any app would be able to own them without even specifying it. Does that make sense?
- Regarding allocation priorities, I'd suggest to make this s simple as possible for now. Specifying lower and upper limits to not need to define an exact match probably makes it easier, but I think we rather need an explicit and usable volume management sooner than later
- I can imagine _requiring_ a constraint in case of path disks: why would I want _any_ path disk? (this is non-MVP though)
- For mount disks, I wonder if requiring the exact amount of disk size AND/OR requiring a root constraint would make sense. Since we cannot label resources in Mesos, it's hard to allocate the correct resources. Similar to enabling "path" constraints for path disks, you could do for mount disks. Given a resource like
`{
  "resources" : [
    {
      "name" : "disk",
      "type" : "SCALAR",
      "scalar" : { "value" : 2048 },
      "disk" : {
        "source" : {
          "type" : "MOUNT",
          "mount" : { "root" : "/mnt/postgres" }
        }
      }
    }
  ]
}`
Marathon could be strict and only launch a task on this resource if the persistent volume has diskType=Mount and size=2048 and a constraint "root" LIKE "/mnt/postgres". (We could leave out the size probably).

I didn't find the time to go over these comments again and verify if I still think they make sense – sorry if some of them are confusing.

joerg84:
Thanks!
Could you add a brief documentation on how to use this (or is it there and I missed it)?

timcharper:
@meichstedt thank you for your comments! Sorry for the latency on my end this time. I hit an intense period of focus at work will try and be more responsive on this pull request.

I agree with you on Tuples, generally, and try to be conservative with them. I think in this case it is warranted.

> I wouldn't consider unspecified to match path or root disk resources as of now, thinking no one would create path disks explicitly if any app would be able to own them without even specifying it. Does that make sense?

In my implementation, as of now, you must specify on which type of disk resource your persistent volume should be created, `root`, `mount`, or `path`. Here is an example:

```js
{
  "type": "DOCKER",
  "volumes": [
    {
      "containerPath": "data",
      "mode": "RW",
      "persistent": {
        "size": 960679,
        "type": "mount",
        "constraints": []
      }
    }
    // ...
```

Failing to provide the persistent `type` will default it to `root`. If you specify a persistent type of `path`, the `root` volume resource will not be considered.

> Regarding allocation priorities, I'd suggest to make this s simple as possible for now. Specifying lower and upper limits to not need to define an exact match probably makes it easier, but I think we rather need an explicit and usable volume management sooner than later

I think this is wise. Since I've implemented path constraints already, it is possible (in a round-about way) to tell Marathon to avoid using certain disks. Adding an optional `maxSize` parameter for `mount` persistent types would be trivial to add and probably makes sense.

> I can imagine requiring a constraint in case of path disks: why would I want any path disk? (this is non-MVP though)
> For mount disks, I wonder if requiring the exact amount of disk size AND/OR requiring a root constraint would make sense. Since we cannot label resources in Mesos, it's hard to allocate the correct resources. Similar to enabling "path" constraints for path disks, you could do for mount disks. Given a resource like `{ "resources" : [ { "name" : "disk", "type" : "SCALAR", "scalar" : { "value" : 2048 }, "disk" : { "source" : { "type" : "MOUNT", "mount" : { "root" : "/mnt/postgres" } } } } ] }` Marathon could be strict and only launch a task on this resource if the persistent volume has `diskType`=`Mount` and `size`=`2048` and a constraint `"root" LIKE "/mnt/postgres"`. (We could leave out the size probably).

Since, as you say, Mesos does not provide a way to label resources, the only signal we can use is the path of the resources. I don't think path constraints particularly clutter the API and the same general constraint mechanism can be used to select Mesos resource labels at the time they are available. I'm not particularly keen on changing the already existing `size` attribute to an optional (fear of breaking compatibility somewhere); if people truly don't have a lower range, they can set it to 0.

So, after these comments, the following should be done before this PR is considered ready:

- `[ ]` - As seen in the inline comments, avoid writing generated Mesos proto classes when possible; Use a dedicated Marathon type (IE the existing DiskType?)
- `[ ]` - add optional `maxSize` parameter for `root` persistent volumes. Make it invalid to set this field invalid for `path` or `root` resources.
- `[ ]` - Modify the code so the existing `size` attribute behaves as a minimum, and not an absolute size, when matching `mount` resources.
- `[ ]` - Make it invalid to specify `constraints` on `root` disk resources (for now).
- `[ ]` - Update documentation to reflect the API changes and behavior.

For clarity, the following remains true:

- `root` persistent volumes will not match Mesos `path` resources. `path` persistent volumes will not match Mesos `root` resources.
- When considering multiple `path` resources, prefer to spread allocations evenly. When multiple persistent `path` or `root` volumes are requested, process the largest ones, first.
- `constraints` (on the filesystem path at which a `mount` or `path` resource resides) feature continues to function as implemented.

What do you think about my rich resource helpers? I know some people frown on implicit classes and I'm happy to convert / move these as is the coding style.

timcharper:
@joerg84 I have not yet written documentation but in my above comment you can see a snippet in which it is used. You could also look at the tests. :)

timcharper:
I've updated the merge request to target 1.3 release, and rebased patches to the latest head of releases/1.3

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryHelper.scala
timcharper: Actually, there is a good reason to use `Option[Mesos.Resource.DiskInfo.Source]` here and not `DiskType`, as `DiskType` does not indicate the corresponding path for the resource.

I could change it to `Iterable[(DiskType, Option[String], LocalVolume)` or something... or create a new type? `Allocation` or something, which contains all of the signal necessary to create the appropriate mesos resource allocation request?

timcharper:
I don't think the current build failure is my fault.

```
[23:25:17][Step 3/3] *** 1 SUITE ABORTED ***
[23:25:17][Step 3/3] GracefulTaskKillIntegrationTest:
[23:25:17][Step 3/3]  *** ABORTED *** (41 seconds, 841 milliseconds)
[23:25:17][Step 3/3]   java.lang.AssertionError: Waiting for clean slate in Mesos took longer than 30 seconds. Give up.
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.next$1(WaitTestSupport.scala:30)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.waitFor(WaitTestSupport.scala:36)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.waitUntil(WaitTestSupport.scala:22)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.setup.SingleMarathonIntegrationTest$class.waitForCleanSlateInMesos(SingleMarathonIntegrationTest.scala:311)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest.waitForCleanSlateInMesos(GracefulTaskKillIntegrationTest.scala:10)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.setup.SingleMarathonIntegrationTest$class.cleanUp(SingleMarathonIntegrationTest.scala:295)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest.cleanUp(GracefulTaskKillIntegrationTest.scala:10)
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest$$anonfun$1.apply$mcV$sp(GracefulTaskKillIntegrationTest.scala:18)
[23:25:17][Step 3/3] [info] Run completed in 10 minutes, 12 seconds.
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest$$anonfun$1.apply(GracefulTaskKillIntegrationTest.scala:18)
[23:25:17][Step 3/3] [info] Total number of tests run: 140
[23:25:17][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest$$anonfun$1.apply(GracefulTaskKillIntegrationTest.scala:18)
[23:25:17][Step 3/3] [info] Suites: completed 195, aborted 1
[23:25:17][Step 3/3]   at org.scalatest.BeforeAndAfter$class.runTest(BeforeAndAfter.scala:200)
[23:25:17][Step 3/3] [info] Tests: succeeded 140, failed 0, canceled 0, ignored 0, pending 0
[23:25:17][Step 3/3] [info] *** 1 SUITE ABORTED ***
```

jasongilanfarr:
You can retest in teamcity manually (that test is flakey)

timcharper:
@jasongilanfarr it's unclear to me how to restart the test; I've created account and clicked through quite a bit but I'm unsure if I am lacking special permissions.

jasongilanfarr:
We're disabling teamcity integration tests since they mirror velocity. I'll merge when I get in.

jasongilanfarr:
@timcharper - I didn't notice this was set to target 1.3 instead of master. In general, things should land in master first then they can be considered for a cherry-pick. We talked about whether this should be in 1.3 and we don't think it should (open to argument to the contrary). I'm going to revert from 1.3 and land this in master shortly.

timcharper:
@jasongilanfarr sorry about the confusion. I can rebase to the latest master, again, if that helps.

jasongilanfarr:
Don't worry about it, I'll take care of it.

jasongilanfarr:
On top of master at #4418

Issue #3795, opened 2016-04-18
Scalar GPU Resource Support


kolloch:
From the design document referenced below:

"the focus is on supporting a narrow initial use case: homogenous machine-level GPU support (i.e. only 1 kind of GPU per machine)"

**Acceptance Criteria**

* [ ] Introduce a new field "gpus" in the app definition alongside "mem", "cpus" and "disk". Only integer values are allowed.
* [ ] Consume the appropriate amount of "gpus" when launching tasks. To be decided: Do we match a "SCALAR" "gpus" value or a "INTEGER_SCALAR" "gpus" value?
* [ ] Create the appropriate user-facing documentation.

Please note that this might not fulfill all requirements. E.g. look at [this user comment](https://github.com/mesosphere/marathon/issues/375#issuecomment-125449960) @air 

**Testing**

Can we test this by simply configuring a [scalar custom resource](http://mesos.apache.org/documentation/latest/attributes-resources/) "gpus" for now? For example:

```json
{ "name": "gpus", "type": "SCALAR", "scalar": { "value": 8 } }
```

Or is INTEGER_SCALAR support already available and we should rather use:

```json
{ "name": "gpus", "type": "INTEGER_SCALAR", "integer": { "value": 8 } }
```

I could not find it in [mesos.proto](https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L570-L575) yet.

/cc @bmahler @klueska 

**Related Issues**

* Support for arbitrary resource types #375

**References**

Internal [JIRA ticket MARATHON-209](https://mesosphere.atlassian.net/browse/MARATHON-209) and the [design document](https://docs.google.com/document/d/10GJ1A80x4nIEo8kfdeo9B11PIbS1xJrrB4Z373Ifkpo/edit?usp=sharing).

seelam:
any update on this?

vditya:
@kolloch We worked with @klueska on MESOS 4424, wondering what is the status on Marathon support for GPU. Have you started working on it? If not we can help to expedite the development for it. Let us know

kolloch:
Just FYI: Hi @vditya, I will not work on Marathon anymore. So better ask e.g. @aquamatthias

vditya:
Thanks for update, Hey Matthias @aquamatthias , any update on GPU support for Marathon effort. If it helps we can support that effort and work with Marathon engineer to enable it.

Yubo-Li:
[...]

PR #4112, opened 2016-07-19
Added GPU support for Marathon API.

liyubobj:
Mesos 1.0 will support a new first-class resource type `gpus` to manage GPU resources ([MESOS-4424](https://issues.apache.org/jira/browse/MESOS-4424)). This patch extended Marathon API to support `gpus` as an optional integer entry with default value `0`. That is, this API change is compatible to old API, but can accept new `gpus` entry to run a task with GPU. For more discussion on this patch, please refer to issue #3795 .


liyubobj:
Is there anyone help a code review on this? Thanks!

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -247,6 +252,8 @@ case class AppDefinition(
       cpus = resourcesMap.getOrElse(Resource.CPUS, this.cpus),
       mem = resourcesMap.getOrElse(Resource.MEM, this.mem),
       disk = resourcesMap.getOrElse(Resource.DISK, this.disk),
+      // TODO (Yubo): Change hard-coded "gpus" to "Resource.GPUS" after mesos-util has GPU resource type support.
```
janisz: Add link to issue/PR

liyubobj: I submitted the issue here: https://github.com/mesosphere/mesos-utils/issues/27, this is another project need to support "Resource.GPUS" resource type.

meichstedt:
@liyubobj thanks for your PR! We'll look into this short term. Imo we could get rid of mesos-utils, but I need to check back with some people about that.

meichstedt:
We might consider putting this into 1.2 if we're confident about the changes.

nfnt:
I'll take a look at how it integrates with Mesos. I'm mostly concerned with things like version upgrades and mixed-node clusters (i.e. nodes with GPUs together with nodes without GPUs).

liyubobj:
Thanks @meichstedt @nfnt . In current the patch did not rely on mesos-utils because I hard-coded "gpus" as resources name which is defined in mesos. In my implementation, "gpus" is an optional entry so that it is compatible with original API.

liyubobj:
@nfnt I'm also a team member of GPU enablement on mesos, so if you have any questions/comments on GPU on mesos, please let me know.

meichstedt:
regarding mesos-utils as to my comment above: we agreed internally to remove the library short term, I'll see how much work that'll be. Feels like it's a no brainer done in 15mins ...

nfnt:
Frameworks that want to use GPU resources need to advertise this by setting the `GPU_RESOURCES` capability in its `FrameworkInfo`. Otherwise it won't retrieve any offers from agents with GPUs. It looks like this PR doesn't set this capability at the moment, or am I missing something here?

liyubobj:
@nfnt yes, you are right. But now `mesos-1.0.0-rc2.jar` used by marathon still have no support on this feature. I plan to support this by a new flag `--enable_features gpu_resources` to marathon after the updated mesos jar supported by marathon.

nfnt:
The `mesos-1.0.0-rc2.jar` has support for the `GPU_RESOURCES` capability (rc1 didn't, though).
Please add the flag (and subsequent setting of the capability) as part of this PR, otherwise the feature wouldn't work as expected, as Mesos >=1.0rc2 wouldn't send out any offers.
It would also make sense to add some validation that if the GPU feature is disabled, one shouldn't set `gpus` values >0.

liyubobj:
OK, I'll work on this. Thanks!

liyubobj:
Added GPU_RESOURCE capability. API validation for GPU feature disabled is on-going, still need more time since I'm not an expert in scala and validation logic. Any suggestion for this?

src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
```diff
@@ -69,6 +69,14 @@ object MarathonSchedulerDriver {
     if (config.isFeatureSet(Features.TASK_KILLING))
       frameworkInfoBuilder.addCapabilities(Capability.newBuilder().setType(Capability.Type.TASK_KILLING_STATE))
 
+    // GPU Resources feature gives Marathon capability to get GPU related resources offers from Mesos. In current,
+    // key "gpus" in Marathon's API can only work with GPU Resources feature enabled. For details of GPU Resource
+    // feature, see MESOS-5634.
```
nfnt: I would write this a bit different, because no offers would get send by agents with GPUs: "GPU Resources allows Marathon to get offers from Mesos agents with GPUs. For details, see MESOS-5634."

liyubobj: The comments ammended. Thanks!

nfnt:
@liyubobj I've created https://github.com/mesosphere/marathon/tree/nfnt/gpu-resource-validation to implement some validation rules (and tests for them). It's important to know, that the GPU feature only works with the Mesos containerizer. This will be tested there as well. Please cherry-pick or merge that branch with yours.

liyubobj:
@nfnt Great! cherry-picked your commit to my branch. It's fine for working with mesos containerizer only now. After our docker containerizer work done, the rule should support docker containerizer.

nfnt:
LGTM!
I've tested this on a machine with an NVIDIA GPU. Tasks run as expected, and app definitions are validated properly, i.e. definitions wanting Docker containers and GPU resources are not valid.

liyubobj:
Also passed validation tests in my machine :)

liyubobj:
Removed hard-coded "gpus"

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -583,6 +593,7 @@ object AppDefinition extends GeneralPurposeCombinators {
     appDef must complyWithSingleInstanceLabelRules
     appDef must complyWithReadinessCheckRules
     appDef must complyWithUpgradeStrategyRules
+    appDef must complyWithGpuRules
```
aquamatthias: I think `complyWithGpuRules` can be replaced by:
```scala
appDef is conditional(app.gpus > 0)(featureEnabled(Features.GPU_RESOURCES))
```

aquamatthias:
@liyubobj @nfnt LGTM. I added a suggestion regarding the validation, which has a TODO we hopefully can remove, once the PR for unified containerizer is merged.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
aquamatthias: If the default containerizer is Mesos, than we should probably remove `app.container.isDefined` to only `app.container.exists(_.docker.isEmpty)`

liyubobj:
Ammended validation logic according to @aquamatthias 's suggestion. @nfnt

PR #4113, opened 2016-07-19
Improve servicePort validation in terms of efficiency:

aquamatthias:
- every app in a list is compared to all other apps _after_ the app (O(n log n))
- all service ports are computed only once

@jmlvanre The computation should be still valid. Can you test, if this improves the performance?


gkleiman:
I think that this approach is actually O(N^2), see: https://stackoverflow.com/questions/13674719/easy-solve-tn-tn-1n-by-iteration-method

The same validation could be done in O(N), with a single iteration through all apps, using a Map<Int, PatId> with service port as key.

gkleiman:
A little off topic: this validation can potentially leak names and ports of apps that a user might not be authorized to see.

aquamatthias:
Close in favor of #4115

PR #4116, opened 2016-07-20
Made `dependencyGraph` only be called once per `Group` instance.

mpark:


jasongilanfarr:
Best 4 character change ever. LGTM.

aquamatthias:
Thanks!

Issue #3867, opened 2016-05-06
Multiple instances of a Task with persistent volume do not dispatch to different nodes when unique hostname constraint is set


IvanTing:
I'm running Marathon 1.1.1 on Mesos 0.28.1.

I have a cluster deployed initially with 3 nodes. Originally, i have a marathon app which has set instance to 3 and "constraints" : [["hostname", "UNIQUE"]].
It works as expected that 3 tasks running on 3 different nodes in the cluster.

After that, i want to set the marathon app as persistent task such that when my cluster scale  to more than 3 nodes (e.g. scale up to 5 nodes), the marathon app can still attach to original running node if the marathon app is killed/restarted. To achieve this this, i update the marathon app json to add persistent volume as follow:
{
...
   "volumes": [
        {
                "containerPath": "data",
                "mode": "RW",
                "persistent": {
                        "size": 10
                }
       }
  ]
...
}
The issue is, when i re-deploy the marathon app with persistent volume again, the 3 tasks may dispatch to same node but not spread to 3 nodes. It violates the unique hostname constraints.

When i look at the marathon log, i find that it tries to reserve and create the volume before launching the task. Below are partial log and from last 3 line, it tries to reserve on node-1b 2 times and node-1a 1 time. It seems that it does not consider the unique hostname constraints yet and hence reserve on node-1b 2 times.

```
May  6 04:13:14 ip-10-10-3-29 marathon[777]: [2016-05-06 04:13:14,601] INFO Request ReserveAndCreateVolumes for task 'ceph-monitor.da7c1b57-1340-11e6-b6ac-0242d42d37a3', version '2016-05-06T04:13:14.226Z'. 3 tasksToLaunch, 0 in flight, 0 confirmed.  not backing off (mesosphere.marathon.core.launchqueue.impl.AppTaskLauncherActor:marathon-akka.actor.default-dispatcher-17)
May  6 04:13:14 ip-10-10-3-29 marathon[777]: [2016-05-06 04:13:14,607] INFO Request ReserveAndCreateVolumes for task 'ceph-monitor.da806119-1340-11e6-b6ac-0242d42d37a3', version '2016-05-06T04:13:14.226Z'. 3 tasksToLaunch, 1 in flight, -1 confirmed. instance count delta 1. not backing off (mesosphere.marathon.core.launchqueue.impl.AppTaskLauncherActor:marathon-akka.actor.default-dispatcher-7)

[...]

PR #4117, opened 2016-07-20
WIP – TaskKillService

### LINKED ISSUE #4039: Deployment gets stuck killing tasks that are lost
During the a deployment, Marathon should expunge lost tasks instead of trying to kill them.

Right now Marathon keeps trying to kill the lost tasks, resulting in a stuck deployment:

```
Jun 28 12:18:38 java[2129]: [2016-06-28 12:18:38,390] INFO Killing 40 instances from 40 (mesosphere.marathon.upgrade.AppStopActor:marathon-akka.actor.default-dispatcher-448)
Jun 28 12:18:38 java[2129]: [2016-06-28 12:18:38,392] INFO Received status update for task foo_bar.09e6dade-3d24-11e6-b927-867b4e0a3eb0: TASK_LOST (Reconciliation: Task is unknown) (mesosphere.marathon.MarathonScheduler$$EnhancerByGuice$$2ddc3939:Thread-11449)
[...]
```

meichstedt:
## Introducing a TaskKillService
Fixes #4039 by adding a service that handles killing tasks, throttling kill requests and retrying kills if no terminal status update was received within a certain amount of time. Replaces `StoppingBehavior` and actors using it, and reduces any `driver.kill` calls to exactly one inside the service.

### Background
#4039 revealed that, when tasks are lost but kept in state because they _might_ come back, deployments of all sorts would fail to complete. This is due to the fact that issuing a `driver.kill` on a lost task will simply return another `TASK_LOST` status update, which does not get processed and published via the event stream because it's not considered a _change_ in state – the task was lost before as well. While the latter could have been improved by publishing _any_ status update on the event bus, it wouldn't have solved all problems. Additionally, it felt about the right time to introduce an abstraction at least for `driver.kill` calls, so that not each and every place where a kill is issued needs to take care about the state of the task. This also improves upon the kill throttling of `StoppingBehavior`, because it effectively throttles _all_ pending kills, not only those of one app.

### Notes
- substantial changes are in
  - TaskTerminationModule
  - TaskKillConfig
  - TaskKillService
  - TaskKillServiceDelegate
  - TaskKillServiceActor
  - TaskKillProgressActor
  - PostToEventStreamStepImpl
- retrying kills is probably suboptimal if the retry happens within the taskKillGracePeriod (the time between the executor sending a sigterm and a sigkill). However, the service shouldn't have to load app definitions in order to consider a different retry timeout.
- There are several places where we need to infer the actual state of a task. This should be much simpler when the task state refactoring is finished – @unterstein is working on that.
- there is a number of TODO's in the code which I will review; any comments on those highly appreciated


src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
```diff
@@ -0,0 +1,41 @@
...
+    default = Some(10000L) // 10 seconds
+  )
+
+  private[this] lazy val _killRetryMax = opt[Int](
+    "kill_retry_max",
+    descr = "INTERNAL TUNING PARAMETER: " +
+      "The maximum number of kill retries before which a task will be forcibly expunged from state.",
+    noshort = true,
+    hidden = true,
+    default = Some(100)
```
meichstedt: IIRC I wanted to set 5 as a default, the 100 slipped in here.

meichstedt:
_randomly assigning people for review_

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+  private[this] lazy val _killRetryTimeout = opt[Long](
+    "kill_retry_timeout",
+    descr = "INTERNAL TUNING PARAMETER: " +
+      "The timeout after which a task kill will be retried.",
+    noshort = true,
+    hidden = true,
+    default = Some(10000L) // 10 seconds
```
jasongilanfarr: 10.seconds.toMillis? - Obviously I'm huge on using durations.

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
```diff
@@ -0,0 +1,41 @@
+package mesosphere.marathon.core.task.termination
+
+import org.rogach.scallop.ScallopConf
+import scala.concurrent.duration._
+
+import scala.concurrent.duration.FiniteDuration
+
+trait TaskKillConfig extends ScallopConf {
+  //scalastyle:off magic.number
```
jasongilanfarr: if you put the magic number comment as a line comment after the code, it disables only for the line.

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
```diff
@@ -0,0 +1,46 @@
...
+import mesosphere.marathon.core.task.Task
+
+import scala.concurrent.Future
+
+/**
+  * A service that handles killing tasks. This will take care about extra logic for lost tasks,
+  * apply a retry strategy and throttle kill requests to Mesos.
+  */
+trait TaskKillService {
+  // TODO: should we favor distinct vs overloaded functions? killTask, killTaskById, killTasks, killUnknownTask?
```
jasongilanfarr: I prefer the distinct option. It makes the code easier to read at the call sites.

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+    * Kill the given tasks and return a future that is completed when all of the tasks
+    * have been reported as terminal.
+    *
+    * @param tasks the tasks that shall be killed.
+    * @return a future that is completed when all tasks are killed.
+    */
+  def kill(tasks: Iterable[Task]): Future[Unit]
```
jasongilanfarr: Consider using `Done` instead of `Unit` - Its really the same thing, but looks nicer IMO. `()` vs `Done`

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+    * Kill the given task. The implementation should add the task onto a queue that is processed
+    * short term and will eventually kill the task.
+    *
+    * @param taskId the id of the task that shall be killed.
+    */
+  def kill(taskId: Task.Id): Unit
```
jasongilanfarr: Why not return Futures for the others?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+  * This actor watches over a given set of taskIds and completes a promise when
+  * all tasks have been reported terminal (including LOST et al)
+  *
+  * @param taskIds the taskIds that shall be watched.
+  * @param promise the promise that shall be completed when all tasks have been
+  *                reported terminal.
+  */
+class TaskKillProgressActor(taskIds: mutable.HashSet[Task.Id], promise: Promise[Unit]) extends Actor with ActorLogging {
```
jasongilanfarr: Curious why it takes in a mutable.HashSet?, maybe the class should be `private[impl]`?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
```diff
@@ -0,0 +1,70 @@
...
+  // TODO: if one of the watched task is reported terminal before this actor subscribed to the event bus,
+  //       it won't receive that event. should we reconcile tasks after a certain amount of time?
+
+  // this should be used for logging to prevent polluting the logs
+  val name = "TaskKillProgressActor" + self.hashCode()
```
jasongilanfarr: s"TaskKillProgressActor(${taskIds.mkString})"?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
```diff
@@ -0,0 +1,70 @@
...
+      if (taskIds.isEmpty) {
+        log.info("All tasks watched by {} are killed, completing promise", name)
+        val success = promise.tryComplete(Try(()))
+        if (!success) log.error("Promise has already been completed in {}", name)
+        context.stop(self)
+      } else {
+        log.info("{} still waiting for {} tasks to be killed", name, taskIds.size)
+      }
+
+    case _ => // ignore
```
jasongilanfarr: why ignore instead of letting it go through the normal actor hierarchy?

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
meichstedt: ack; copied and pasted from another ScallopConf trait and didn't even change the value there ;)

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
meichstedt: 👍

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
meichstedt: agreed. didn't want to change the code unless someone thinks the same.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+
+    case unexpected: Any =>
+      log.debug("Received unexpected {}", unexpected)
+  }
+  //scalastyle:on
+
+  def killTaskById(taskId: Task.Id): Unit = {
+    log.debug("Received KillTaskById({})", taskId)
+    import context.dispatcher
+    taskTracker.task(taskId).map {
```
jdef: why re-dispatch these calls through the actor vs. invoking the related funcs directly?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  def isLost(task: Task): Boolean = {
+    import org.apache.mesos
+    task.mesosStatus.fold(false)(_.getState == mesos.Protos.TaskState.TASK_LOST)
+  }
+}
+
+private[termination] object TaskKillServiceActor {
+
+  sealed trait Request
```
jdef: what do you think about including a `reason: String` member as part of this trait? the goal being that anyone wanting to kill a task needs to describe why they want to do so

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+    case KillTasks(tasks, promise) =>
+      killTasks(tasks, promise)
+
+    case Terminal(event) if inFlight.contains(event.taskId) || tasksToKill.contains(event.taskId) =>
+      handleTerminal(event.taskId)
+
+    case Retry =>
+      retry()
+
+    case unhandled: Request =>
```
jdef: .. or `InternalRequest` ?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+    val taskIsLost: Boolean = maybeTask.fold(false)(isLost)
+
+    if (taskIsLost) {
+      log.warning("Expunging lost {} from state because it should be killed", taskId)
+      // we will eventually be notified of a taskStatusUpdate after the task has been expunged
+      stateOpProcessor.process(TaskStateOp.ForceExpunge(taskId))
+    } else {
+      val knownOrNot = if (maybeTask.isDefined) "known" else "unknown"
+      log.warning("Killing {} {}", knownOrNot, taskId)
+      driverHolder.driver.map(_.killTask(taskId.mesosTaskId))
```
jdef: `driver.foreach`?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
```diff
@@ -0,0 +1,70 @@
...
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.event.MesosStatusUpdateEvent
+
+import scala.concurrent.Promise
+import scala.collection.mutable
+import scala.util.Try
+
+/**
+  * This actor watches over a given set of taskIds and completes a promise when
+  * all tasks have been reported terminal (including LOST et al)
```
jdef: It also subscribes to the event bus to listen for interesting mesos task status updates (this behavior is deprecated and will be removed in a future release (because MesosStatusUpdateEvent is)?).

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+
+  def killUnknownTaskById(taskId: Task.Id): Unit = {
+    log.debug("Received KillUnknownTaskById({})", taskId)
+    tasksToKill.update(taskId, None)
+    processKills()
+  }
+
+  def killTasks(tasks: Iterable[Task], promise: Promise[Unit]): Unit = {
+    log.debug("Adding {} tasks to queue; setting up child actor to track progress", tasks.size)
+    context.actorOf(TaskKillProgressActor.props(tasks.map(_.taskId), promise))
```
jdef: so this actor spawns child actors of itself? trying to understand the overall lifecycle for the actor.

jdef: nevermind, misread. I see that this is a kill-progress actor

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
```diff
@@ -0,0 +1,70 @@
...
+      promise.tryComplete(Try(()))
+      log.info("premature aborting of {} - no tasks to watch for", name)
+      context.stop(self)
+    }
+  }
+
+  override def postStop(): Unit = {
+    context.system.eventStream.unsubscribe(self)
+
+    if (!promise.isCompleted) {
```
jdef: under what conditions do we expect this to happen? worth documenting?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
jdef: And what's the point of this watcher? better logging? doesn't look like it manages state other than its own internal list of tasks to watch

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+    import context.dispatcher
+    taskTracker.task(taskId).map {
+      case Some(task) => self ! KillTask(task)
+      case _ => self ! KillUnknownTaskById(taskId)
+    }
+  }
+
+  def killTask(task: Task): Unit = {
+    log.debug("Received KillTask({})", task)
+    tasksToKill.update(task.taskId, Some(task))
```
jdef: why don't we track kills via the progress actor here?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+    processKills()
+  }
+
+  def killUnknownTaskById(taskId: Task.Id): Unit = {
+    log.debug("Received KillUnknownTaskById({})", taskId)
+    tasksToKill.update(taskId, None)
```
jdef: why don't we track kills via the progress actor here?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,181 @@
...
+import scala.collection.mutable
+import scala.concurrent.Promise
+
+/**
+  * An actor that handles killing tasks in chunks and depending on the task state.
+  * Lost tasks will simply be expunged from state, while active tasks will be killed
+  * via the scheduler driver. There is be a maximum number of kills in flight, and
+  * the service will only issue more kills when tasks are reported terminal.
+  */
+private[impl] class TaskKillServiceActor(
```
jdef: (a) what do you think about modeling this as an FSM actor?
(b) what do you think about batching together kills that arrive within some window (so that back-to-back individual kills don't have to wait for the retry period)?

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
meichstedt: Nice, I seemed to have missed the introduction of `Done`. Changed it, but the result will be mapped to Unit in places I didn't want to change now.

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
meichstedt: tl;dr: the code that kills single tasks in a fire and forget manner doesn't need to wait for results, or does that by subscribing to the eventStream.

The `killTasks(..)` replaces the former `StoppingBehavior` trait that was used by TaskKillActor/AppStopActor which were used when scaling down/stopping an app. In that case, the actors performed the _killing and watching_, and completed a Future when ready. The related implementations around that now use `killTasks` and wait for the Future, leaving waiting, throttling and retrying up to the `TaskKillProgressActor`.

Most other places where Marathon issues kill requests are fire and forget requests. E.g., the `TaskReplaceActor` (which performs app upgrades) kills tasks in that manner and subscribes to the eventBus listening for `MesosStatusUpdateEvent`'s to figure out whether it can kill another task/start a new one/has finished.

So, yes, we can return a Future for all of those, makes sense to me. But then we need to adjust the usages and rely on the futures instead of the event bus which means slightly more changes that I wanted to do here.

We have some work items scheduled around deployment logic and that change would fit just fine I guess.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
meichstedt: I obviously forgot making it package private – yes, it should be. The idea was to create the mutable data structure within the `props` of its companion object, and let the ctor be private.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
meichstedt: Hm, when we stopp an application with thousands of tasks it's a bit verbose to print all tasks in every log message. Instead, what about changing
> `Starting {} to track kill progress of {} tasks`
to include the complete list of tasks?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
meichstedt: ha, I usually have a `log.debug` in there while during implementation and tend to miss removing it .. removed.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: Good question. My intention was to have a 1:1 mapping of `Message -> Action`, that's why I chose to use the existing Messages in this case as well. Afaik we have both ways of doing it in our code base, maybe it's time for a style direction and picking one best practice/convention. Preferences, anybody?

That said, I realize the only place where we call `killTaskbyId` is via the TaskReplaceActor, which performs a lossy mapping of `Iterable[Task]` to a Set[Task.Id] (although it has all the tasks in scope all the way). I figure by changing that to kill [Task]'s explicitly, we can get rid of `killTaskbyId` completely. Which makes sense imo, because requesting a task to be killed without having its state at hand seems like killing it blindfold ..

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: +1 from me
I actually thought about that! It always bothers me if the logs don't say _why_ a task was killed, we have to infer the reason from the logs above which can be painful.

Would you be comfortable with a `trait TaskKillReason` on the delegate level of `TaskKillService`, so that every `kill()` function has a second argument that defines that reason?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: ack

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: totally

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
meichstedt: Why/since when is `MesosStatusUpdateEvent` deprecated? `mesosphere.marathon.core.task.bus` is, somehow, but that doesn't include publishing `MesosStatusUpdateEvent` on the eventStream afaik.

The point of this actor is: it waits until all the tasks it watches are terminal and then completes a promise.

Before I introduced this, there were 2 Actors (TaskKillActor, AppStopActor) that mixed in a common behavior that killed tasks and waited until they'd be reported terminal. The whole concern of killing/throttling/retrying is moved into the TaskKillServiceActor, and the watch dogs are here only to allow waiting for a future when all requested task kills have happened.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: Seems like I should add better documentation ;)

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
meichstedt: There is currently no expected condition for this to happen which is why the promise will be failed in this case. Worth documenting.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: To prevent shared mutable state and separate concerns, but most of all because there were observed kill sequences before as well as fire and forget killings (this sounds way too figurative ...). A `TaskKillProgressActor`'s data will usually be a subset of the `TaskKillServiceActor`'s set. Killing tasks does not imply the creation of a `TaskKillProgressActor`, as not all users of this service wait for a future. When, for example, 10 apps are deleted, there will be 10 progress actors watching for the respective tasks to terminate, 1 for each `killTasks(...)` call. Yet there might be an Upgrade of another app that simply chimes in several kill requests and doesn't wait for anything but listen to the eventStream. As noted in a comment above, that might be adjusted in the future.

So, overall,
* The `TaskKillServiceActor`'s concern is to kill a given set of tasks, throttle and retry (both of these _could_ be separated if necessary), while
* The `TaskKillProgressActor`'s concern is to watch a given set of tasks and complete a promise when all of them are reported terminal.

I acknowledge there is or should be a strong relation between those concerns, and I'm open to refactor this when we tackle the deployment logic (see `TaskReplaceActor` for an example).

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: see above

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: a) the current implementation of the actor has merely one state, so that might be considered too much. Yet I like the concise and readable notation of akka FSM, so yes, I consider that if others are fine with that decision.

b) Makes sense to me. I was thinking about doing it it the beginning but chose a less sophisticated approach to get the ball rolling and provide the main functionality. When batch processing the messages, an FSM would make sense, so doing b) justifies a) ;)
OTOH, batch processing would slightly defer incoming kills – in a range of hundreds of milliseconds, that would be just fine I guess.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
jdef: Sounds good to me

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
jdef: The comments for the MesosStatysUpdateEvent class suggest that it should go
away at some point, that's all.

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
```diff
@@ -0,0 +1,40 @@
...
+  private[this] lazy val _killRetryMax = opt[Int](
+    "kill_retry_max",
+    descr = "INTERNAL TUNING PARAMETER: " +
+      "The maximum number of kill retries before which a task will be forcibly expunged from state.",
+    noshort = true,
+    hidden = true,
+    default = Some(5) //scalastyle:off magic.number
+  )
+
+  def killChunkSize: Int = _killChunkSize()
```
jasongilanfarr: these could be lazy val no?

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
jasongilanfarr: That makes sense to me. I was thinking that having the futures now would make testing easier and should we later use the futures, we won't need to adjust this api.

meichstedt: Yes, @aquamatthias also liked the idea of having Futures on all functions. Will do.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -437,14 +440,18 @@ class SchedulerActions(
...
 
-  def stopApp(driver: SchedulerDriver, app: AppDefinition): Future[_] = {
+  def stopApp(app: AppDefinition): Future[_] = {
     healthCheckManager.removeAllFor(app.id)
 
     log.info(s"Stopping app ${app.id}")
     taskTracker.appTasks(app.id).map { tasks =>
+      tasks.foreach { task =>
+        if (task.launchedMesosId.isDefined) {
+          log.warn("Killing {}", task.taskId)
```
aquamatthias: I would not use warn. Also: isn't the logging done on the KillService?

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -554,9 +561,9 @@ class SchedulerActions(
         .filter(t => t.mesosStatus.fold(false)(status => runningOrStaged.get(status.getState).nonEmpty))
         .sortWith(sortByStateAndTime)
         .take(launchedCount - targetCount)
-      val taskIds: Iterable[TaskID] = toKill.flatMap(_.launchedMesosId)
-      log.info(s"Killing tasks: ${taskIds.map(_.getValue)}")
-      taskIds.foreach(driver.killTask)
+
+      log.warn("Killing tasks {}", toKill.map(_.taskId))
```
aquamatthias: I would not use warn. Also: isn't the logging done on the KillService?

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueTasksActor.scala
```diff
@@ -50,11 +51,9 @@ private[jobs] object OverdueTasksActor {
     }
 
     private[this] def killOverdueTasks(now: Timestamp, tasks: Iterable[Task]): Unit = {
-      driverHolder.driver.foreach { driver =>
-        overdueTasks(now, tasks).foreach { overdueTask =>
-          log.warn("Killing overdue {}", overdueTask.taskId)
-          driver.killTask(overdueTask.taskId.mesosTaskId)
-        }
+      overdueTasks(now, tasks).foreach { overdueTask =>
+        log.warn("Killing overdue {}", overdueTask.taskId)
```
aquamatthias: I would not use warn. Also: isn't the logging done on the KillService?

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillConfig.scala
meichstedt: right, changed that

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
meichstedt: Yes, the `TaskKillServiceDelegate` now logs kill requests and their respective reasons. changed the others to debug/info

src/main/scala/mesosphere/marathon/core/event/Events.scala
```diff
@@ -186,6 +186,13 @@ case class MesosStatusUpdateEvent(
   eventType: String = "status_update_event",
   timestamp: String = Timestamp.now().toString) extends MarathonEvent
 
+// TODO these must be removed when we have a distinct MarathonTaskStatus
```
aquamatthias: Will this be available, once the PR from Johannes has been merged?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -88,6 +88,8 @@ sealed trait Task {
     else
       Some(agentInfo.host)
   }
+
+  def version: Option[Timestamp]
```
aquamatthias: Why is this an Option?
Can you add a short scaladoc on top?

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+    * Kill a task by the given Id. The implementation should add the task onto a queue that is processed
+    * short term and will eventually kill the task.
+    *
+    * @param taskId the id of the task that shall be killed.
+    * @param reason the reason why the task shall be killed.
+    * @return a future that is completed when all tasks are killed.
+    */
+  // TODO: remove function and only allow killing unknown tasks or Tasks
```
aquamatthias: Looks like this function is still used.
Either replace the usage or remove the TODO

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillReason.scala
```diff
@@ -0,0 +1,35 @@
...
+  /** The task is killed because of an incoming http request */
+  case object KillingTasksViaApi extends TaskKillReason
+
+  /** The task is killed because a new version is being deployed */
+  case object Upgrading extends TaskKillReason
+
+  /** The task is killed because the app no longer exists */
+  case object Orphaned extends TaskKillReason
+
+  /** The task is killed because the app no longer exists */
```
aquamatthias: Wrong comment?

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillProgressActor.scala
aquamatthias: One possible condition: a new leader gets elected, and not all tasks where killed.
Is there any transient state, that does not get reconciled:
- deployments are reconciled
- healthchecks would be executed again on the new leader
- scale changes would be executed again on the new leader

But
- unsuccessful kills via the API would get lost eventually

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,183 @@
...
+  def processKills(): Unit = {
+    val killCount = config.killChunkSize - inFlight.size
+    val toKillNow = tasksToKill.take(killCount)
+
+    log.info("processing {} kills", toKillNow.size)
+    toKillNow.foreach {
+      case (taskId, maybeTask) => processKill(taskId, maybeTask)
+    }
+
+    retryTimer = Some(context.system.scheduler.scheduleOnce(config.killRetryTimeout, self, Retry))
```
aquamatthias: You call `processKills` in different methods. 
This method schedules a `Retry` message. (Also `retry` schedules a `Retry` message).
To not schedule more messages than needed, either look for already scheduled messages or cancel the already schedules message.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -0,0 +1,183 @@
...
+  import TaskKillServiceActor._
+  import context.dispatcher
+
+  val tasksToKill: mutable.HashMap[Task.Id, Option[Task]] = mutable.HashMap.empty
+  val inFlight: mutable.HashMap[Task.Id, TaskToKill] = mutable.HashMap.empty
+  var retryTimer: Option[Cancellable] = None
+
+  override def preStart(): Unit = {
+    context.system.eventStream.subscribe(self, classOf[MesosStatusUpdateEvent])
+  }
```
aquamatthias: add postStop:
- unsubscribe from the event stream.
- cancel the retryTimer
- if not all tasks were killed, log the remaining tasks

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
```diff
@@ -12,6 +12,7 @@ object ScalingProposition {
     meetConstraints: ((Iterable[Task], Int) => Iterable[Task]),
     scaleTo: Int): ScalingProposition = {
 
+    // TODO: tasks in state KILLING shouldn't be killed and should decrease the amount to kill
```
aquamatthias: Can we add this functionality and remove the TODO?

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
```diff
@@ -29,10 +30,22 @@ object ScalingProposition {
     // rest are tasks that are not sentenced and need not be killed to meet constraints
     val rest = notSentencedAndRunningMap -- killToMeetConstraints.map(_.taskId)
 
+    // TODO: this should evaluate a task's health as well
```
aquamatthias: Can we add this functionality and remove the TODO?

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
```diff
@@ -44,3 +57,22 @@ object ScalingProposition {
   }
 
 }
+
+private[this] object SortHelper {
+  import mesosphere.marathon.core.task.bus.MesosTaskStatus._
+
+  // TODO: as soon as we have a clear task state the weights should be defined there or in a map
```
aquamatthias: Same question as above: will this be addressed in Johannes PR?

aquamatthias:
@meichstedt I added some comments - mainly regarding open TODO's. 
I would like to minimize the number of TODOs, if we merge a PR - if possible.

src/main/scala/mesosphere/marathon/core/event/Events.scala
meichstedt: Exactly. I would actually like to merge his PR first because I think it is easier to rebase mine on top of his. The other way 'round shouldn't be a big problem, though.

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: It's an option because `Task.Reserved` currently has no version (although it should). I didn't want to change that and introduce a bunch more changes therefore I added this for convenience getter.

Should be fixed and removed in #3867

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillService.scala
meichstedt: I chose to follow the _right_ path, removed the function and adjusted code in the TaskReplaceActor so this now uses `killTask` instead of the removed function.

src/main/scala/mesosphere/marathon/core/task/termination/TaskKillReason.scala
meichstedt: good catch, fixed.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: True, even better catch here. I've restructured that and hid the mutable timer in a trait, yet decided not to use that as a mixin. I think a `RetryingBehavior` might make sense for various Actors, but I didn't want to optimize here now.

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
meichstedt: I realized the missing postStop when restructuring the timer ... thanks for the hint!

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
meichstedt: Yes, let's merge #4141 first then I can rebase and fix this

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
meichstedt: Yes, let's merge #4141 first then I can rebase and fix this

src/main/scala/mesosphere/marathon/upgrade/ScalingProposition.scala
meichstedt: Yes, let's merge #4141 first then I can rebase and fix this

src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -168,3 +175,31 @@ private[termination] object TaskKillServiceActor {
...
 }
+
+/**
+  * Wraps a timer into an interface that hides internal mutable state behind simple setup and cancel methods
+  */
+private[this] trait RetryTimer {
+  private[this] var retryTimer: Option[Cancellable] = None
+
+  /** Creates a new timer when setup() is called */
+  def createTimer: () => Cancellable
```
aquamatthias: createTimer is  a function without parentheses that gives you a function of arity 0.
I would argue, defining it simply as:
```scala
def createTimer(): Cancellable
```
would be more natural and easy to read here.

meichstedt: totally agree – fixed.

aquamatthias:
Thanks!

PR #4118, opened 2016-07-20
WIP - task state refactoring

unterstein:
Aim of this refactoring was to reduce the usage of mesos.Protos.TaskStatus and to merge the existing MarathonTaskStatus and the MesosTaskStatus and introduce therefore a consistent and mandatory marathonTaskStatus for each Task.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
@@ -123,10 +123,7 @@ object TaskSerializer {
       builder.addAllPorts(hostPorts.map(Integer.valueOf).asJava)
     }
     def setMarathonTaskStatus(marathonTaskStatus: MarathonTaskStatus): Unit = {
-      //scalastyle:off null
-      if (marathonTaskStatus != null) {
-        builder.setMarathonTaskStatus(MarathonTaskStatusSerializer.toProto(marathonTaskStatus))
-      }
+      builder.setMarathonTaskStatus(MarathonTaskStatusSerializer.toProto(marathonTaskStatus))
     }
 
     setId(task.taskId)


src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+  // RUNNING
+  case object Running extends MarathonTaskStatus
+  // STAGING
+  case object Staging extends MarathonTaskStatus
+  // STARTING
+  case object Starting extends MarathonTaskStatus
+
+  // Temporarily transformation states, should become mesos native in the future
+  // UNREACHABLE
+  case object Unreachable extends MarathonTaskStatus
```
meichstedt: https://docs.google.com/document/d/1AYoF5HZPRdQN2TsRpPOliGC6oHen6aHVc0FBOo30rLQ/edit#

mentions a state `TASK_DROPPED` as well (this might be a recent addition, haven't seen it so far). We should include this as well.

src/main/proto/marathon.proto
```diff
@@ -141,6 +141,22 @@ message MarathonTask {
     repeated string local_volume_ids = 1;
     required State state = 2;
   }
+  enum MarathonTaskStatus {
```
jdef: pretty sure that I know what most of these mean. that said, docs would be appreciated

jdef: for example, what's the difference between Gone and Lost. would be good to spell that out

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+    mesos.Protos.TaskStatus.Reason.REASON_RECONCILIATION,
+    mesos.Protos.TaskStatus.Reason.REASON_SLAVE_DISCONNECTED,
+    mesos.Protos.TaskStatus.Reason.REASON_SLAVE_REMOVED
+  )
+
+  val WontComeBack: Set[mesos.Protos.TaskStatus.Reason] = {
+    mesos.Protos.TaskStatus.Reason.values().toSet.diff(MightComeBack)
+  }
+
+  trait Terminal
```
jdef: ?should this be sealed

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  //scalastyle:off cyclomatic.complexity
+  def apply(taskStatus: mesos.Protos.TaskStatus): MarathonTaskStatus = {
+    taskStatus.getState match {
```
jdef: default case?

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+  import org.apache.mesos.Protos.TaskState._
+
+  def fromMesosStateName(mesosStateName: String): Option[MarathonTaskStatus] = all.collectFirst {
+    case status: MarathonTaskStatus if status.toMesosStateName == mesosStateName => status
+  }
+
+  // If we're disconnected at the time of a TASK_LOST event, we will only get the update during
+  // a reconciliation. In that case, the specific reason will be shadowed by REASON_RECONCILIATION.
+  // Since we don't know the original reason, we need to assume that the task might come back.
+  val MightComeBack: Set[mesos.Protos.TaskStatus.Reason] = Set(
```
jdef: ?why is this capitalized, vs `mightBeLost`

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+    mesos.Protos.TaskStatus.Reason.REASON_RECONCILIATION,
+    mesos.Protos.TaskStatus.Reason.REASON_SLAVE_DISCONNECTED,
+    mesos.Protos.TaskStatus.Reason.REASON_SLAVE_REMOVED
+  )
+
+  val WontComeBack: Set[mesos.Protos.TaskStatus.Reason] = {
```
jdef: ditto

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
jdef: .. and since this is a collection of "reasons" (and not statuses) would it be more clear to renamed as `reasonMightComeBack`?

src/main/proto/marathon.proto
jdef: or maybe link to src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala and put the docs there?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -138,6 +149,24 @@ object TaskSerializer {
   }
 }
 
+object MarathonTaskStatusSerializer {
+  def fromProto(proto: Protos.MarathonTask.MarathonTaskStatus): MarathonTaskStatus = {
+    val result = MarathonTaskStatus.all.collectFirst {
+      case status: MarathonTaskStatus if status.toString == "" + proto => status
+    }
+    // TODO ju<>me discuss
+    result.orNull
```
jdef: how is a null better than an `Option`?

src/main/proto/marathon.proto
unterstein: Thanks for the note. I agree that src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala should be a good place and link doc in Proto.

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
unterstein: Good point, will change this! Was a `clone and own` from the old MesosTaskStatus :)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
unterstein: Well .. this _should_ occur only during migration where no MarathonTaskStatus is present in Proto. Afterwards MarathonTaskStatus is mandatory in Task.Status and Proto, therefore the Serializer was intended to handle no `Option`. To indicate my  unhappiness with this `orNull` I added the TODO ... maybe you have an hint to enhance this.

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+
+// TODO ju handle garbage flag
+sealed trait MarathonTaskStatus {
+  def toMesosStateName: String = "TASK_" + toString.toUpperCase
+}
+
+object MarathonTaskStatus {
+  import org.apache.mesos.Protos.TaskState._
+
+  def fromMesosStateName(mesosStateName: String): Option[MarathonTaskStatus] = all.collectFirst {
```
meichstedt: Seems like this function is not used. Do we need it?

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
meichstedt: Do we need these Sets in this scope? They're just for translation from Mesos `TaskStatus -> MarathonTaskStatus`, correct? Can you put those into a layer that solves that purpose so we don't see them here?

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
meichstedt: It can and should

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+  )
+
+  val WontComeBack: Set[mesos.Protos.TaskStatus.Reason] = {
+    mesos.Protos.TaskStatus.Reason.values().toSet.diff(MightComeBack)
+  }
+
+  trait Terminal
+
+  //scalastyle:off cyclomatic.complexity
+  def apply(taskStatus: mesos.Protos.TaskStatus): MarathonTaskStatus = {
```
meichstedt: This one is also only needed for migration if I'm right – then this should be defined in that scope

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -527,7 +527,8 @@ class SchedulerActions(
   def scale(driver: SchedulerDriver, app: AppDefinition): Unit = {
     import SchedulerActions._
 
-    def launchedNotLost(t: Task) = t.launched.isDefined && t.mesosStatus.fold(false)(_.getState != TaskState.TASK_LOST)
+    def launchedNotLost(t: Task) = t.launched.isDefined && !t.isMesosLost
```
meichstedt: this should be `!t.isUnreachable` and the def should be named accordingly. Every terminal task will be expunged from state, so `Gone, Unknown, Lost` will never apply here. The background is that we only want to consider launched _reachable_ tasks here, so it makes sense to evaluate Unreachable only.

That said, you could add an implicit `isReachable` (Staging, Started, Running, Killing), and boil this down to
```
val launchedCount = taskTracker.countAppTasksSync(app.id, _.isReachable)
``` 
without the need for the extra def

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -457,4 +475,28 @@ object Task {
       else None
     }
   }
+
+  implicit class TaskStatusComparison(val task: Task) extends AnyVal {
```
meichstedt: I'd put these either into a package object or a `TaskImplicits` object, not sure we have a convention for that. I guess `TaskImplicits` would be the obvious fit.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -457,4 +475,28 @@ object Task {
...
+    def isReserved: Boolean = compareTo(MarathonTaskStatus.Reserved)
+    def isCreated: Boolean = compareTo(MarathonTaskStatus.Created)
+    def isError: Boolean = compareTo(MarathonTaskStatus.Error)
+    def isFailed: Boolean = compareTo(MarathonTaskStatus.Failed)
+    def isFinished: Boolean = compareTo(MarathonTaskStatus.Finished)
+    def isKilled: Boolean = compareTo(MarathonTaskStatus.Killed)
+    def isKilling: Boolean = compareTo(MarathonTaskStatus.Killing)
+    def isLost: Boolean = compareTo(MarathonTaskStatus.Lost)
+    def isMesosLost: Boolean = isLost || isGone || isUnreachable
```
meichstedt: We shouldn't need or have this. the _old_ Mesos `LOST` doesn't tell much about what's actually going on with that task, so I'd rather explicitly handle the new, better states (even if they're translated for now)

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -457,4 +475,28 @@ object Task {
...
+    def isRunning: Boolean = compareTo(MarathonTaskStatus.Running)
+    def isStaging: Boolean = compareTo(MarathonTaskStatus.Staging)
+    def isStarting: Boolean = compareTo(MarathonTaskStatus.Starting)
+    def isUnreachable: Boolean = compareTo(MarathonTaskStatus.Unreachable)
+    def isGone: Boolean = compareTo(MarathonTaskStatus.Gone)
+    def isUnknown: Boolean = compareTo(MarathonTaskStatus.Unknown)
+
+    def mightBeLost: Boolean = MarathonTaskStatus.mightBeLost.contains(task.status.taskStatus)
```
meichstedt: same thing as above: might be lost doesn't help us – for a transitional phase this might have value but it should be removed before merging.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -457,4 +475,28 @@ object Task {
...
+
+    private def compareTo(marathonTaskStatus: MarathonTaskStatus): Boolean = {
```
meichstedt: I'm not sure about the added value of this ;)

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
```diff
@@ -176,7 +179,11 @@ class TaskOpFactoryImpl(
         agentId = Some(offer.getSlaveId.getValue),
         attributes = offer.getAttributesList.asScala
       ),
-      reservation = Task.Reservation(persistentVolumeIds, Task.Reservation.State.New(timeout = Some(timeout)))
+      reservation = Task.Reservation(persistentVolumeIds, Task.Reservation.State.New(timeout = Some(timeout))),
+      status = Task.Status(
+        stagedAt = now,
```
meichstedt: it becomes very obvious now, that a mandatory `stagedAt` is a lie 😁 
But that can be fixed in the next step as far as in concerns me

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -345,7 +345,7 @@ private class TaskLauncherActor(
       inProgress = tasksToLaunch > 0 || inFlightTaskOperations.nonEmpty,
       tasksLeftToLaunch = tasksToLaunch,
       finalTaskCount = tasksToLaunch + taskLaunchesInFlight + tasksLaunched,
-      tasksLost = tasksMap.values.count(_.mesosStatus.fold(false)(_.getState == Mesos.TaskState.TASK_LOST)),
+      tasksLost = tasksMap.values.count(value => value.mightBeLost),
```
meichstedt: also next step: rename tasksLost to unreachable – it's actually only about them

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -112,23 +115,27 @@ object Task {
     //scalastyle:off cyclomatic.complexity method.length
     override def update(update: TaskStateOp): TaskStateChange = update match {
       // case 1: now running
-      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running(mesosStatus), now) if !hasStartedRunning =>
+      case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running, mesosStatus, now) if !hasStartedRunning =>
```
meichstedt: with a dedicated status, we no longer need to infer the state by evaluating `hasStartedRunning`.

This can be something like
`case TaskStateOp.MesosUpdate(_, MarathonTaskStatus.Running, mesosStatus, now) if this.isStaging =>`

Not sure whether `Starting` would be possible as well

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -112,23 +115,27 @@ object Task {
...
         val updated = copy(
           status = status.copy(
             startedAt = Some(now),
-            mesosStatus = mesosStatus))
+            mesosStatus = Option(mesosStatus)))
```
meichstedt: Some(mesosStatus)

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
meichstedt: The conversion from Mesos TaskStatus to MarathonTaskStatus in currently done in the `TaskStatusUpdateProcessorImpl`, called by the `MarathonScheduler`. From a boundary perspective, I can imagine putting the conversion login into that layer, i.e. convert within the `MarathonScheduler` (or whatever will be that layer in the future) and only pass the MarathonTaskStatus into the `TaskStatusUpdateProcessor`. That's currently not easily possible because the conversion is lossy (health info's are dropped) and we still depend on the mesos TaskStatus elsewhere (yet we should encompass those values in a future Status).

I don't think we can use an existing enum as a default case without affecting the scheduling logic. I'd rather make this an `Option` or `Try[MarathonTaskStatus]` and fail fast if we receive something that's not mapped. Plus, write a test that makes sure that all `mesos.Protos.TaskState.values()` are mapped to a `MarathonTaskStatus`.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -112,23 +115,27 @@ object Task {
...
+        val updated = copy(status = status.copy(
+          mesosStatus = Option(mesosStatus),
+          taskStatus = taskStatus))
         TaskStateChange.Expunge(updated)
 
       // case 3: health or state updated
-      case TaskStateOp.MesosUpdate(_, taskStatus, now) =>
-        updatedHealthOrState(status.mesosStatus, taskStatus.mesosStatus) match {
+      case TaskStateOp.MesosUpdate(_, taskStatus, mesosStatus, now) =>
+        updatedHealthOrState(status.mesosStatus, Option(mesosStatus)) match {
```
meichstedt: TODO: put all relevant information from `mesos.Protos.TaskStatus` into `Task.Status` and get rid of `mesos.Protos.TaskStatus` alltogether 😬  (probably too much for this PR though)

src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
```diff
@@ -48,16 +49,13 @@ class ExpungeOverdueLostTasksActor(
   }
 
   def filterLostGCTasks(tasks: Map[PathId, AppTasks]): Iterable[Task] = {
-    def isTimedOut(task: Task): Boolean = {
-      task.mesosStatus.fold(false) { status =>
+    def isTimedOut(taskStatus: Option[TaskStatus]): Boolean = {
+      taskStatus.fold(false) { status =>
```
meichstedt: This is the place where we actually only want to have the MarathonTaskStatus and the timestamp of when the task entered that state. Since the Task only offers `stagedAt` and `startedAt`, we need to rely on the mesos TaskStatus here, which is quite unfortunate.

In the _hopefully near_ future, this should be something like
`isTimedOut(task.status.since)`

I don't think this improvement should be part of this PR though.

src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
```diff
@@ -48,16 +49,13 @@ class ExpungeOverdueLostTasksActor(
...
         val age = clock.now().toDateTime.minus(status.getTimestamp.toLong * 1000).getMillis.millis
         age > config.taskLostExpungeGC
       }
     }
-    tasks.values.flatMap(_.tasks.filter {
-      case TemporarilyUnreachable(task) if isTimedOut(task) => true
-      case _ => false
-    })
+    tasks.values.flatMap(_.tasks.filter(task => isTimedOut(task.mesosStatus)))
```
meichstedt: This is the place where we actually only want to have the MarathonTaskStatus and the timestamp of when the task entered that state. Since the Task only offers `stagedAt` and `startedAt`, we need to rely on the mesos TaskStatus here, which is quite unfortunate.

In the _hopefully near_ future, this should be something like
`isTimedOut(task.status.since)`

I don't think this improvement should/can be part of this PR though.

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+    *
+    * @return list of possible mappings of the mesos.Protos.TaskStatus.TASK_LOST
+    */
+  lazy val mightBeLost: Set[MarathonTaskStatus] = Set(Gone, Unreachable, Lost)
+
+  lazy val all: Set[MarathonTaskStatus] = Set(Reserved, Created, Error, Failed, Finished, Killed, Killing, Lost, //
+    Running, Staging, Starting, Unreachable, Gone, Unknown)
+
+  // Marathon specific states
+  // RESERVED
```
meichstedt: I am not sure if these capitalized comments add information or disturb readability. If you can add a doc line describing what the state _means_, this would be far better imo.

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,100 @@
...
+  // tl;dr: don't use Lost
+  case object Lost extends MarathonTaskStatus with Terminal
+  // RUNNING
+  case object Running extends MarathonTaskStatus
+  // STAGING
+  case object Staging extends MarathonTaskStatus
+  // STARTING
+  case object Starting extends MarathonTaskStatus
+
+  // Temporarily transformation states, should become mesos native in the future
```
meichstedt: Don't think we need this comment – there's a conversion from TaskStatus to MarathonTaskStatus and for Marathon internal code it's only relevant which states there are.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -118,9 +122,16 @@ object TaskSerializer {
       status.mesosStatus.foreach(status => builder.setStatus(status))
       builder.addAllPorts(hostPorts.map(Integer.valueOf).asJava)
     }
+    def setMarathonTaskStatus(marathonTaskStatus: MarathonTaskStatus): Unit = {
+      //scalastyle:off null
+      if (marathonTaskStatus != null) {
```
meichstedt: why check for null? After migration, all tasks will have a MarathonTaskStatus, and before migration, they won't be serialized

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
```diff
@@ -138,6 +149,24 @@ object TaskSerializer {
...
+    // TODO ju<>me discuss
+    result.orNull
+  }
+
+  def toProto(marathonTaskStatus: MarathonTaskStatus): Protos.MarathonTask.MarathonTaskStatus = {
+    val result = Protos.MarathonTask.MarathonTaskStatus.values().collectFirst {
+      case status: Protos.MarathonTask.MarathonTaskStatus if status.toString == "" + marathonTaskStatus => status
+    }
+    // TODO ju<>me discuss
+    result.orNull
```
meichstedt: `MarathonTaskStatus` is a sealed trait, therefore it's compile safe to do a pattern match here

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
meichstedt: I'd make this more specific. The translation `MarathonTaskStatus -> Proto` can be compile safe via pattern matching; use that to construct a Mapping and the vice versa mapping. Fail if a proto value is not present in the mapping when doing `Proto -> MarathonTaskStatus`

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
```diff
@@ -45,8 +44,9 @@ class ScaleAppUpdateStepImpl @Inject() (
     terminalOrExpungedTask.foreach { task =>
       val appId = task.taskId.runSpecId
       val taskId = task.taskId
-      val state = task.mesosStatus.fold(TaskState.TASK_STAGING)(_.getState)
-      val reason = task.mesosStatus.fold("")(status => if (status.hasReason) status.getReason.toString else "")
+      val state = task.status.taskStatus.toMesosStateName
```
meichstedt: do tests fails if you don't do `toMesosStateName`? Otherwise, for the log message using the taskStatus would be fine

src/main/scala/mesosphere/marathon/health/HealthCheckActor.scala
```diff
@@ -83,12 +82,11 @@ private[health] class HealthCheckActor(
...
 
   def dispatchJobs(): Unit = {
-    import TemporarilyUnreachable.isUnreachable
-
     log.debug("Dispatching health check jobs to workers")
     taskTracker.appTasksSync(app.id).foreach { task =>
+
       task.launched.foreach { launched =>
-        if (launched.runSpecVersion == app.version && launched.hasStartedRunning && !isUnreachable(task)) {
+        if (launched.runSpecVersion == app.version && launched.hasStartedRunning && !task.isUnreachable) {
```
meichstedt: Not for this PR, but keep in mind: _Every task has a version_

meichstedt:
@unterstein thanks for your effort, overall looks good to me. I haven't yet looked over the tests but I think I've commented on anything implementation-wise that was worth commenting. I see mostly minor issues plus several points that we should address in follow-up PR's:
- TaskChanged overhaul
- encompassing relevant values from mesos `TaskStatus` into `Task.Status`
- general version information
- timestamps (`stagedAt`, `startedAt` vs `since`)
- cleaning up `Task.Reserved`, `Task.LaunchedOnReservation` and `Task.LaunchedEphemeral` (and probably getting rid of these types)

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -371,21 +374,47 @@ class MigrationTo0_16(groupRepository: GroupRepository, appRepository: AppReposi
...
       }
-    }.flatMap { Unit =>
+
+    def loadAndMigrateTasks(id: String): Future[MarathonTask] = {
+      taskRepository.task(id).flatMap {
+        case Some(entity) =>
+          val updatedEntity = entity.toBuilder
+            .setMarathonTaskStatus(MarathonTaskStatusSerializer.toProto(MarathonTaskStatus(entity.getStatus)))
+            .build()
+          taskRepository.store(updatedEntity)
```
meichstedt: Thinking about this, you should skip entities that already have a marathonTaskStatus. In case a migration attempt fails and is re-run, some of these might have been migrated already. That's not necessarily nice, but it's the way it is handled until now.

unterstein: Yep!

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -580,6 +581,7 @@ private[this] object SchedulerActions {
       case (Some(av), Some(bv)) => fn(av, bv)
       case _ => default
     }
+    // TODO ju replaceable with ```a.taskStatus, ...``` ?
```
meichstedt: Yes, definitely. We want to sort by state, then timestamp. The timestamp _should_ be the timestamp of the state, but the stagedAt will do as well as long as the compared states are equal.

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
unterstein: `TASK_LOST will be replaced with a new set of task statuses that are defined more precisely: TASK_DROPPED, TASK_UNREACHABLE, TASK_GONE, and TASK_UNKNOWN. The new task statuses will only be shown to frameworks that opt-in via a framework capability.`

Sounds like LOST will be completely replaced. Therefore MarathonTaskStatus#apply for TASK_LOST should not map to TASK_LOST anymore, right?

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
unterstein: nope

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
unterstein: I like the `isReachable` solution.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
unterstein: In ```QueuedTaskInfo``` do we really need only the number of unreachable tasks for the val ```tasksLost```? Gone and Dropped are unimportant?

src/main/scala/mesosphere/marathon/core/task/Task.scala
unterstein: How could the status be Running and Staging at the same time?

src/main/scala/mesosphere/marathon/core/task/Task.scala
unterstein: yes, we should open a follow up task for this.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
unterstein: Tests...I am simulating existing Tasks without MarathonTaskStatus in the repo and unfortunately toProto is called there...

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
unterstein: They do. I would like to remove toMesosStateName and change the integration tests waiting for this log messages...what do you think?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
unterstein: No `null` available anymore. Neither migration nor tests do need this.

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
unterstein: removed

src/main/scala/mesosphere/marathon/core/task/Task.scala
unterstein: no mightBeLost anymore

src/main/scala/mesosphere/marathon/core/task/tracker/impl/TaskSerializer.scala
unterstein: Removed. Changed test and migration usage.

unterstein:
moved to: https://github.com/mesosphere/marathon/pull/4141

Issue #1090, opened 2015-01-23
Show refused offers as part of deployment status


cmaloney:
Frequently deploys, and just spinning up new instances doesn't get any acceptable offers (or sometimes any offers at all). Marathon should give some UI for this. 

A "Time since last offer" + list of recently refused offers with reason for being refused would be really helpful for debugging why things aren't starting in marathon.

ConnorDoyle:
Possible duplicate or extension of these issues:

- #126
- #652
- #951
- #968
- #1077

sttts:
:+1:

smarthall:
+1

air:
Related to #951 and #1307, which are different cases of deployments going wrong.

kolloch:
:+1:

stuart-warren:
copied from #1728 

I've noticed that trying to work out why an application is stuck in the queue is pretty much impossible to do without accessing the marathon log files. Though the information is at least present there.

eg:
```
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]: Conflicting constraints are: [field: "hostname"
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]: operator: UNIQUE
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]: ] (mesosphere.mesos.ResourceMatcher$:39)
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]: [2015-06-30 13:44:18,497] INFO All
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]: [2015-06-30 13:44:18,498] WARN App ports are not available in the offer. (mesosphere.mesos.ResourceMatcher$:50)
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]: [2015-06-30 13:44:18,498] INFO No matching offer for /platform/services/discovery (need cpus=0.25, mem=64.0, disk=0.0, ports=List(8880)) : id {
Jun 30 13:44:18 master-eeb65358-49f6-45e2-80ab-2018dc057a50 marathon[933]:   value: "20150626-230939-3209314314-5050-1116-O259398"
[...]

PR #4119, opened 2016-07-20
#1090 Show refused offers endpoint 

tmmi:
after rebase master


src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -22,7 +21,8 @@ class TaskBuilder(
     runSpec: RunSpec,
     newTaskId: PathId => Task.Id,
     config: MarathonConf,
-    appTaskProc: Option[RunSpecTaskProcessor] = None) {
+    appTaskProc: Option[RunSpecTaskProcessor] = None,
+    rejectionCollectorOpt: Option[RejectOfferCollector]) {
```
jdef: Instead of further parameterizing `TaskBuilder` I'd rather see "rejections" be emitted as the output from a build attempt.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -73,6 +76,10 @@ class MarathonModule(conf: MarathonConf, http: HttpConf)
     bind(classOf[SchedulerDriverFactory]).to(classOf[MesosSchedulerDriverFactory]).in(Scopes.SINGLETON)
     bind(classOf[MarathonScheduler]).in(Scopes.SINGLETON)
     bind(classOf[MarathonSchedulerService]).in(Scopes.SINGLETON)
+
+    bind(classOf[RejectOfferCollector]).to(classOf[InMemRejectOfferCollector]).in(Scopes.SINGLETON)
+    bind(classOf[TaskOpFactory]).to(classOf[TaskOpFactoryImpl]).in(Scopes.SINGLETON)
```
jdef: we're trying to do away with guice-based dependencies. AFAIK the tasks module has been written to avoid guice as much as possible. can you factor guice out of this?

jdef:
Thanks so much for the PR, left some comments for you

src/main/scala/mesosphere/mesos/TaskBuilder.scala
tmmi: Do you think that return rejections collector from TaskBuilder.build is a better idea ?
This might complicate TaskOpFactoryImpl.
Why don't hide creation of RejectOfferCollector in MarathonModule ?

tmmi:
I added master changes, but I am not sure how to really improve it by changing TaskBuilder.

jasongilanfarr:
@tmmi Thanks for the PR. We have a similar change on our roadmap and expect to address it soon. Because it will also be involved with the pods work, I'd suggest we close this PR out and we will address it.

kamilchm:
@jasongilanfarr thx for info. I'm curious what you come up with.

PR #4121, opened 2016-07-20
Enhanced volume validation messages

unterstein:
If the validation of a volume failed, the validation should return a readable error message.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -5,7 +5,7 @@ import java.util.regex.Pattern
 
 import com.wix.accord._
 import com.wix.accord.ViolationBuilder._
-import mesosphere.marathon.{ AllConf, ValidationFailedException }
+import mesosphere.marathon.{AllConf, ValidationFailedException}
 import mesosphere.marathon.state.FetchUri
 import org.slf4j.LoggerFactory
 import play.api.libs.json._
@@ -13,6 +13,7 @@ import play.api.libs.json._
 import scala.collection.GenTraversableOnce
 import scala.reflect.ClassTag
 import scala.util.Try
+import scala.util.matching.Regex
 
 object Validation {
   def validateOrThrow[T](t: T)(implicit validator: Validator[T]): T = validate(t) match {
@@ -292,4 +293,7 @@ object Validation {
     extends NullSafeValidator[String](
       v => if (partialMatchAllowed) pattern.matcher(v).find() else pattern.matcher(v).matches(),
       v => v -> validationMessage)
+
+  def matchRegexFullyWithMessage(regex: Regex, message: String): Validator[String] =
+    new MatchesRegexWithMessage(regex.pattern, false, message)
 }
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -1,8 +1,10 @@
 package mesosphere.marathon.api.v2
 
 import java.net._
+import java.util.regex.Pattern
 
 import com.wix.accord._
+import com.wix.accord.ViolationBuilder._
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
```diff
@@ -182,7 +182,8 @@ private[impl] object DVDIProviderValidations extends ExternalVolumeValidations {
         volume.external.size is isTrue("must be undefined for Docker containers")(_.isEmpty)
         volume.containerPath is notOneOf(DotPaths: _*)
         // TODO(jdef) change this once docker containerizer supports relative containerPaths
-        volume.containerPath should matchRegexFully(AbsolutePathPattern)
+        volume.containerPath should
+          matchRegexFullyWithMessage(AbsolutePathPattern, "containerPath must not starts with \"/\"")
```
jdef: s/starts/start/

jdef: is `containerPath` really needed in these messages? doesn't accord add that in automatically when it parses the object hiearchy (object == the subject of validation)?

src/test/scala/mesosphere/mesos/PersistentVolumeValidationTest.scala
```diff
@@ -0,0 +1,44 @@
...
+    When("The volume is created and validation failed")
+    volume should not be null
+    volume.containerPath should be (path)
+    val validation = Volume.validVolume(volume)
+    validation.isSuccess should be (false)
+
+    Then("A validation exists with a readable error message")
+    validation match {
+      case Failure(violations) =>
+        violations should contain (RuleViolation("/path", "containerPath must not contain \"/\"", Some("containerPath")))
```
jdef: as per the above, if `containerPath` weren't part of the message there wouldn't be the redundant appearance of `containerPath` on this line

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -280,4 +283,17 @@ object Validation {
       validator(t)
     }
   }
+
+  /**
+    * A validator that succeeds only if the provided string matches the specified pattern.
+    *
+    * Initially took from com.wix.accord.combinators.StringCombinators to enhance validationMessage feature
+    */
+  class MatchesRegexWithMessage(pattern: Pattern, partialMatchAllowed: Boolean = true, validationMessage: String)
```
jdef: how about a unit test that exercises partial vs. non-partial matching?

jdef:
looking good so far, left some comments

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
unterstein: Maybe something like `value must not start with "/"` ?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -280,4 +283,17 @@ object Validation {
...
+    extends NullSafeValidator[String](
+      v => if (partialMatchAllowed) pattern.matcher(v).find() else pattern.matcher(v).matches(),
+      v => v -> validationMessage)
+
+  def matchRegexFullyWithMessage(regex: Regex, message: String): Validator[String] =
```
meichstedt: I think you can boil this down to
```
  def regexValidator(regex: Regex, failureMessage: String) = new NullSafeValidator[String](
    test = _.matches(regex.regex),
    failure = _ -> failureMessage
  )
```
withouth the need for `MatchesRegexWithMessage `.

meichstedt:
LGTM too – please remove the extra unneeded complexity in `Validation.scala` and this is ready merge.

unterstein:
done

meichstedt:
👍

Issue #4109, opened 2016-07-18
GracefulTaskKillIntegrationTest is flaky


gkleiman:
```
[12:49:48][Step 3/3] GracefulTaskKillIntegrationTest:
[12:49:48][Step 3/3]  *** ABORTED *** (43 seconds, 227 milliseconds)
[12:49:48][Step 3/3]   java.lang.AssertionError: Waiting for clean slate in Mesos took longer than 30 seconds. Give up.
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.next$1(WaitTestSupport.scala:30)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.waitFor(WaitTestSupport.scala:36)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.setup.WaitTestSupport$.waitUntil(WaitTestSupport.scala:22)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.setup.SingleMarathonIntegrationTest$class.waitForCleanSlateInMesos(SingleMarathonIntegrationTest.scala:315)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest.waitForCleanSlateInMesos(GracefulTaskKillIntegrationTest.scala:10)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.setup.SingleMarathonIntegrationTest$class.cleanUp(SingleMarathonIntegrationTest.scala:299)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest.cleanUp(GracefulTaskKillIntegrationTest.scala:10)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest$$anonfun$1.apply$mcV$sp(GracefulTaskKillIntegrationTest.scala:18)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest$$anonfun$1.apply(GracefulTaskKillIntegrationTest.scala:18)
[12:49:48][Step 3/3]   at mesosphere.marathon.integration.GracefulTaskKillIntegrationTest$$anonfun$1.apply(GracefulTaskKillIntegrationTest.scala:18)
[...]

PR #4122, opened 2016-07-20
GracefulTaskKillIntegrationTest should not be flaky

unterstein:
According to https://github.com/mesosphere/marathon/issues/4109 GracefulTaskKillIntegrationTest looks like to be flaky, therefore I reduced the waiting duration and connected appCommand with timeout.


gkleiman:
LGTM

meichstedt:
👍

PR #4124, opened 2016-07-22
wip: Persistence API phase 2 (preview)

jasongilanfarr:
This has a big chunk of moves (38fc291) which you may want to more or less ignore. I really wish github could take care of it.

- [x] Migrate GroupRepository and make it lazy load.
- [x] Make AppRepository Read Only, Group Repository is allowed to write into AppRepository
- [x] Implement StoredGroup where apps are stored in groups only by id
- [x] Implement common base class for Repository/VersionedRepository and use that for the "old style" storage as well
- [x] Move Migration into StorageModule and allow only it to access the specific "new/old" Repositories.
- [ ] Write a generic XRepository test replacing the existing ones.
- [x] Write a GroupRepositoryTest (50%)
- [x] Remove Injections for things in the storage module, exporting only what is strictly required by Guice via CoreGuiceModule
- [ ] Fix MarathonHealthCheckManagerTest
- [ ] Fix MarathonSchedulerActorTest
- [x] Fix GroupManagerTest
- [x] Ensure Marathon can run with the new storage selected.
- [x] Migrate EventSubscribers storage
- [x] Migrate FrameworkId storage 

- [ ] Some day in the future: Garbage Collect Groups


src/main/proto/marathon.proto
```diff
@@ -274,6 +274,18 @@ message GroupDefinition {
   repeated string dependencies = 5;
 }
 
+message StoredGroup {
```
aquamatthias: Minor: Naming: All entities in this file are `stored`.

aquamatthias: You could also extend GroupDefinition, instead of creating another message.
E.g.
```protobuf
message GroupDefinition {
  message AppReference {
    required string id = 1;
    required string version = 2;
  }
  required string id = 1;
  required string version = 2;
  repeated ServiceDefinition deprecatedApps = 3;
  repeated GroupDefinition groups = 4;
  repeated string dependencies = 5;
  repeated AppReference appIds = 6;
}
```

src/main/scala/mesosphere/marathon/core/storage/StorageModule.scala
```diff
@@ -22,68 +19,96 @@ import scala.concurrent.ExecutionContext
...
 
 object StorageModule {
-  def apply(conf: MarathonConf, metrics: Metrics, mat: Materializer, ctx: ExecutionContext,
-    scheduler: Scheduler, actorRefFactory: ActorRefFactory): StorageModule =
-    apply(StorageConfig(conf))(metrics, mat, ctx, scheduler, actorRefFactory)
+  def apply(conf: MarathonConf)(implicit metrics: Metrics, mat: Materializer, ctx: ExecutionContext,
+    scheduler: Scheduler, actorRefFactory: ActorRefFactory): StorageModule = {
+    val currentConfig = StorageConfig(conf)
+    val legacyConfig = conf.internalStoreBackend() match {
+      case TwitterZk.StoreName => Some(TwitterZk(cache = true, conf))
```
aquamatthias: Not 100% sure: wouldn't `currectConfig` be already `TwitterZk` in this case? 
Why would you override the cache configuration here to be always true?
As far as I can see the legacyConfig is only used in Migration for a special purpose. Perhaps legacyConfig is a pretty generic term, perhaps we can rename it to something that expresses the intent more clearly?

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
```diff
@@ -29,10 +29,11 @@ import scala.collection.immutable.Seq
 import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.{ Duration, _ }
 import scala.reflect.ClassTag
+// scalastyle:on
 
 sealed trait StorageConfig
```
aquamatthias: 1)
In other modules, the config is a trait, which defines the configuration properties of this module.
The `StorageConfig` totally makes sense, but is something different. 
The `StorageModule` takes the complete `MarathonConf` - so it is hard to see which configuration parameter will be used (and it is hard to use those classes in another context e.g. metronome).
Perhaps it makes sense to move this class into `impl`?
Can you create a trait that defines the config options of this module?

2) 
I would like to see the repository as a separate top level module.
This component could be used from other applications (like metronome) and should not contain specific classes (like GroupRepositoryImpl)

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
```diff
@@ -0,0 +1,184 @@
+package mesosphere.marathon.core.storage.repository.impl
```
aquamatthias: Please move this class into the store module. The repository module should not have any knowledge about specific types.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/legacy/GroupEntityRepository.scala
```diff
@@ -0,0 +1,53 @@
+package mesosphere.marathon.core.storage.repository.impl.legacy
```
aquamatthias: Please move this class into the store module. The repository module should not have any knowledge about specific types.

src/main/scala/mesosphere/marathon/core/event/EventConf.scala
```diff
@@ -0,0 +1,62 @@
+package mesosphere.marathon.core.event
```
aquamatthias: Can you rebase instead of merge, so we see only your changes?

src/main/scala/mesosphere/marathon/core/storage/migration/Migration.scala
```diff
@@ -0,0 +1,208 @@
...
+        new MigrationTo0_16(legacyConfig).migrate().recover {
+          case NonFatal(e) => throw new MigrationFailedException("while migrating storage to 0.16", e)
+        }
+      },
+      StorageVersions(1, 2, 0) -> { () =>
+        new MigrationTo1_2(legacyConfig).migrate().recover {
+          case NonFatal(e) => throw new MigrationFailedException("while migrating storage to 1.2", e)
+        }
+      },
+      StorageVersions(1, 3, 0) -> { () =>
```
aquamatthias: 1) If you want cherry pick this in 1.2, we need to change this to the correct version. 
2) since we now have 2 different layouts in one version (legacy and new), it is not clear from the version, which layout has been used. Perhaps we also have to store the layout in the SchemaVersion. (Otherwise it is hard to say what to do, once we have 1.3)

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,114 @@
...
+  *
+  * Does nothing unless Legacy Storage and New Storage are configured.
+  */
+class MigrationTo1_3_PersistentStore(migration: Migration)(implicit
+  executionContext: ExecutionContext,
+    mat: Materializer,
+    metrics: Metrics) extends StrictLogging {
+  def migrate(): Future[Done] = async {
+    val legacyStore = await(migration.legacyStoreFuture)
+    // This migration only goes from legacy storage -> new storage.
```
aquamatthias: all our migrations currently go only forward...

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,114 @@
...
+    (legacyStore, migration.persistenceStore, migration.legacyConfig) match {
+      case (Some(_), Some(_), Some(legacyConfig)) =>
+        val futures = legacyStore.map { _ =>
+          val legacyConfig = {
+            // if persistent store is defined, the config has to be too.
+            require(migration.legacyConfig.isDefined)
```
aquamatthias: This requirement is already made sure by the pattern match

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,114 @@
...
+  executionContext: ExecutionContext,
+    mat: Materializer,
+    metrics: Metrics) extends StrictLogging {
+  def migrate(): Future[Done] = async {
+    val legacyStore = await(migration.legacyStoreFuture)
+    // This migration only goes from legacy storage -> new storage.
+    (legacyStore, migration.persistenceStore, migration.legacyConfig) match {
+      case (Some(_), Some(_), Some(legacyConfig)) =>
+        val futures = legacyStore.map { _ =>
+          val legacyConfig = {
```
aquamatthias: The whole val statement can be removed

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,114 @@
...
+            // if persistent store is defined, the config has to be too.
+            require(migration.legacyConfig.isDefined)
+            migration.legacyConfig.get
+          }
+          Seq(
+            migrateTasks(legacyConfig, migration.taskRepo),
+            migrateDeployments(legacyConfig, migration.deploymentRepository),
+            migrateTaskFailures(legacyConfig, migration.taskFailureRepo),
+            // note: we don't actually need to migrate apps (group does it)
+            migrateGroups(legacyConfig, migration.groupRepository)
```
aquamatthias: What about:
```
EntityStore[FrameworkId]
EntityStore[EventSubscribers]
```
?

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,114 @@
...
+        logger.info(s"Migrated ${summary.mkString} to 1.3")
+        Done
+      case _ =>
+        logger.info("Skipping Curator Persistence Migration (no legacy store/persistent store in use)")
+        Done
+    }
+  }
+
+  def migrateRepo[T](oldRepo: Repository[_, T], newRepo: Repository[_, T]): Future[Int] = {
+    oldRepo.all().mapAsync(Int.MaxValue) { value =>
```
aquamatthias: I'm probably jus a burned child, which stumbles across this here over and over again.
To repeat myself: this will put a lot of pressure on zk...

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,114 @@
...
+      newRepo.store(value)
+    }.runFold(0) { case (acc, _) => acc + 1 }
```
aquamatthias: The write is performed sequentially - good!

src/main/scala/mesosphere/marathon/state/GroupManager.scala
```diff
@@ -138,17 +139,19 @@ class GroupManager @Inject() (
 
     log.info(s"Upgrade group id:$gid version:$version with force:$force")
 
-    def storeUpdatedApps(plan: DeploymentPlan): Future[Unit] = {
-      plan.affectedApplicationIds.foldLeft(Future.successful(())) { (savedFuture, currentId) =>
+    case class AppDelta(updated: Seq[AppDefinition], deleted: Seq[PathId])
```
aquamatthias: The DeploymentPlan has a function `createdOrUpdatedApps`.
Perhaps also add a `deletedApps` and use those functions instead of computing AppDelta in GroupManager?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
```diff
@@ -0,0 +1,184 @@
...
+      case Some(group) =>
+        Some(await(group))
+      case None =>
+        None
+    }
+  }
+
+  /**
+    * @inheritdoc
+    * @todo We need to eventually garbage collect as we will accumulate apps/appVersions that neither
```
aquamatthias: We already have a couple of actors that run from time to time to do all kinds of gc operations. We can easily add another one.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
```diff
@@ -0,0 +1,184 @@
...
+    * This may be as simple as "get all of the app ids" and compare them to the appIds that are in a group.
+    * If no group has the appId, delete it. We probably don't actually need to delete app versions that are no longer
+    * in use as there is some group still referring to some version of the app, so eventually it will be deleted once
+    * it is no longer in actual usage.
+    *
+    * The garbage collection does not need to run as part of storeRoot (which would likely slow it down by quite
+    * a bit), instead, we should run it out of band (triggered by a storeRoot) and use the prepare/commit/cancellable
+    * described above.
+    */
+  override def storeRoot(group: Group, updatedApps: Seq[AppDefinition], deletedApps: Seq[PathId]): Future[Done] = {
```
aquamatthias: - a group is stored with references to app+versions
- the app is stored as separate entity

We have to make sure that all app+version references are always available.
We currently limit the number of versions stored per entity hard in BaseRepository.
Perhaps it makes sense to also do this kind of operation in the gc operation.

Please note: the same is true for tasks (reference an app with version).

src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
```diff
@@ -1,6 +1,7 @@
 package mesosphere.util.state
 
-import mesosphere.marathon.state.{ Timestamp, EntityStore, MarathonState }
+import mesosphere.marathon.core.storage.repository.impl.legacy.store.EntityStore
```
aquamatthias: Importing something from `impl` is bad

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
```diff
@@ -0,0 +1,184 @@
...
+private[storage] case class StoredGroup(
+    id: PathId,
+    appIds: Map[PathId, OffsetDateTime],
+    storedGroups: Seq[StoredGroup],
+    dependencies: Set[PathId],
+    version: OffsetDateTime) {
+  def resolve(
+    groupRepository: GroupRepository,
+    appRepository: AppRepository)(implicit ctx: ExecutionContext): Future[Group] = async {
+    val appFutures = appIds.map { case (appId, appVersion) => appRepository.getVersion(appId, appVersion) }
```
aquamatthias: Perhaps add a method `AppRepository.getVersions` so we can cap concurrent access?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
```diff
@@ -0,0 +1,184 @@
...
+    * a bit), instead, we should run it out of band (triggered by a storeRoot) and use the prepare/commit/cancellable
+    * described above.
+    */
+  override def storeRoot(group: Group, updatedApps: Seq[AppDefinition], deletedApps: Seq[PathId]): Future[Done] = {
+    async {
+      val storeAppFutures = updatedApps.map(appRepository.store)
+      val deleteAppFutures = deletedApps.map(appRepository.deleteCurrent)
+      val storedGroup = StoredGroup(group)
+      await(Future.sequence(storeAppFutures))
+      await(Future.sequence(deleteAppFutures).recover { case NonFatal(e) => Done })
```
aquamatthias: Thanks for handling the exception case here.

aquamatthias:
@jasongilanfarr 
The last commit message sounds great!
I left a couple of comments and added one todo item.

Perhaps you can rebase your branches, so we only see your changes.
Please also try to make the build/test happy.

src/main/proto/marathon.proto
jasongilanfarr: that would work well too, I'll switch over

src/main/scala/mesosphere/marathon/core/storage/StorageModule.scala
jasongilanfarr: For migration purposes, I figured having caching on would be very helpful. The legacyConfig (i realize its confusing here - this area needs a lot more cleanup still) is used solely to move from pre-1.3 -> 1.3 - and the migration code closes the underlying store when its done -- needs a lot more documentation.

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
jasongilanfarr: (1) mostly agree, but I had something different in mind:

StorageConfig itself not a "Scallop Option", its a typesafe "runtime edition" of the config, so we can read from `Config` later or `Scallop` now. Having this in the impl makes sense.

The storeCache() is defined only in `MarathonConf` sadly, otherwise, this only needs ZookeeperConf.

(2) I believe I put store under "storage"

src/main/scala/mesosphere/marathon/core/event/EventConf.scala
jasongilanfarr: Definitely.

src/main/scala/mesosphere/marathon/core/storage/migration/Migration.scala
jasongilanfarr: agreed, perhaps 1.2.1?

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
jasongilanfarr: I was trying to point out that this one only makes sense if we go from `legacy->new`, it doesn't make sense in `new->new'`. I made those be strongly differentiated.

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
jasongilanfarr: oops! good catch

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
jasongilanfarr: good catch.

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
jasongilanfarr: Yeah, but... ZkPersistenceStore has capped concurrency on purpose. I kind of got sick or commenting `// the underlying stores are expected to cap their concurrency`

src/main/scala/mesosphere/marathon/state/GroupManager.scala
jasongilanfarr: great idea.

src/main/scala/mesosphere/util/state/FrameworkIdUtil.scala
jasongilanfarr: ^ TODO: Migrate FrameworkId storage.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
jasongilanfarr: The underlying store is capped already.

jasongilanfarr:
Thanks for looking through the gigantic diff. I'll rebase both phase 1 and two ontop of master as one squashed commit each (its way easier in terms of merge conflicts), then create a combined set.

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
jasongilanfarr: Umm, actually, they aren't, they will write as fast as they can (in parallel) for this one.Keep in mind "PersistenceStores are expected to cap their own concurrency" - I'll make sure this is in the PersistenceStore implementation notes in the class comment.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GroupRepositoryImpl.scala
jasongilanfarr: I was very careful when doing that to ensure this is the case. Its why _only_ `GroupRepository` can even write to `AppRepository`. I think I missed the task relation.

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
aquamatthias: To 1) I am 100% with you, that using scallop directly is a bad idea (at least it would be consistent). Let's discuss f2f how we want to proceed with configuration. 

To 2) We want to create a library, which can be reused elsewhere (e.g. metronome). The `store` module is general purpose and can be reused  (and could be part of this library), but the `storage` module is Marathon specific. That's why it would make sense to have 2 modules (perhaps rename `storage` into `repository`)

src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
aquamatthias: Cool. Sorry for being so picky on this...

src/main/scala/mesosphere/marathon/core/storage/StorageConfig.scala
jasongilanfarr: (1) Sounds good. I did move to StorageConf.
(2) Hmm. We could do that, the repository package has generic repositories too and most of the marathon repositories are just named editions of them.

PR #4126, opened 2016-07-22
Document PUT /v2/apps/{appId} is update-or-create

annaken:
The PUT /v2/apps/{appId} was documented as updating a running app; however if the app is not running then PUT will actually create a new app (ie what a POST /v2/apps would do).


gkleiman:
Thanks!

Issue #4093, opened 2016-07-13
Unit test failed: MigrationTo1_2Test


aquamatthias:
This could indicate a problem with the migration:

```
 MigrationTo1_2Test:
[10:06:32][Step 3/3] 
[10:06:32][Step 3/3] - should remove deployment version nodes, but keep deployment nodes *** FAILED *** (27 milliseconds)
[10:06:32][Step 3/3]   ArrayBuffer("foo:bar", "deployment:42c6b840-5a4b-4110-a7d9-d4835f7499b9:2016-06-13T18:47:15.862Z", "deployment:fcabfa75-7756-4bc8-94b3-c9d5b2abd38c") did not contain the same elements as List("deployment:fcabfa75-7756-4bc8-94b3-c9d5b2abd38c", "foo:bar") (MigrationTo1_2Test.scala:43)
```

gkleiman:
I ran this test in a loop for over an hour and it always succeeded :(

aquamatthias:
Saw that problem again: https://travis-ci.org/mesosphere/marathon/builds/145976834
```
*** 1 TEST FAILED ***
MigrationTo1_2Test:
- should remove deployment version nodes, but keep deployment nodes *** FAILED *** (26 milliseconds)
  ArrayBuffer("foo:bar", "deployment:265fe17c-2979-4ab6-b906-9c2b34f9c429:2016-06-23T22:16:03.880Z", "deployment:42c6b840-5a4b-4110-a7d9-d4835f7499b9:2016-06-13T18:47:15.862Z", "deployment:fcabfa75-7756-4bc8-94b3-c9d5b2abd38c") did not contain the same elements as List("deployment:fcabfa75-7756-4bc8-94b3-c9d5b2abd38c", "foo:bar") (MigrationTo1_2Test.scala:43)
  org.scalatest.exceptions.TestFailedException:
  at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
  at org.scalatest.words.ResultOfContainWord.theSameElementsAs(ResultOfContainWord.scala:115)
  at mesosphere.marathon.state.MigrationTo1_2Test$$anonfun$1.apply$mcV$sp(MigrationTo1_2Test.scala:43)
  at mesosphere.marathon.state.MigrationTo1_2Test$$anonfun$1.apply(MigrationTo1_2Test.scala:29)
[...]

PR #4128, opened 2016-07-22
Towards #4093 - Refactor MigrationTo1_2Test using async/wait

gkleiman:


src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -376,19 +377,18 @@ class MigrationTo0_16(groupRepository: GroupRepository, appRepository: AppReposi
...
-          deploymentRepository.store.expunge(versionNode).map(_ => ())
-        }
-      }
-    }.flatMap { Unit =>
-      log.info("Finished 1.2 migration")
-      Future.successful(())
+    val nodes: Seq[String] = await(deploymentRepository.store.names())
+    val deploymentVersionNodes = nodes.filter(deploymentRepository.isVersionKey)
+
+    val it = deploymentVersionNodes.iterator
```
jasongilanfarr: val futures = deploymentVersionNodes.map(deploymentRepository.store.expunge)
await(Future.sequence(futures))?

The underlying store already has concurrency limiting.

jasongilanfarr:
My change rewrites this completely... otherwise looks good to me.

src/main/scala/mesosphere/marathon/state/Migration.scala
gkleiman: Looking at `ZkStore` and `MarathonStore`, it seems that they use the default execution context.

I think we should merge this to see if it helps with the flaky unit test, even if it will be replaced by something different very soon.

jasongilanfarr:
Agreed. I missed that ZKStore doesn't cap concurrency.

gkleiman:
@aquamatthias PTAL

aquamatthias:
Thanks!

PR #4130, opened 2016-07-22
Implement and Propagate new Persistence API

jasongilanfarr:
This is all of #4097 and #4124 merged together and rebased on master.

## TODOS
- [x] Address review comments from https://github.com/mesosphere/marathon/pull/4124
- [x] Write generic XRepository test and replace all of the existing ones
- [x] Finish generic XRepository test (and make sure the underlying stores have full coverage too)
- [x] Finish GroupRepositoryTest
- [x] Fix MarathonHealthCheckManagerTest
- [x] Fix MarathonSchedulerActorTest
- [x] Fix GroupManagerTest
- [x] Migrate EventSubscribers Storage
- [x] Migrate FrameworkId Storage
- [x] Add Concurrency Limiting to legacy ZKStore.
- [x] Expand test coverage a bit more.
- [x] Split off things required from MarathonConf in StorageModule into StorageConf.
- [x] Merge proto "StoredGroup" into GroupDefinition.
- [x] Add a ton more documentation/cleanup to Migration.
- [x] Utilize DeploymentPlan.createdOrUpdatedApps instead of calculating a delta in GroupManager
- [x] Add root caching to StoredGroupRepository
- [x] Wire up "PreDriverStart" for EntityStores and LoadTimeCachingStore.
- [x] Implement Garbage Collection for StoredGroups #4156
- [x] Add GC Test #4156
- [ ] Fix broken integration tests
- [ ] Verify old marathon (pre-branch) ZK database -> new (legacy) database and old marathon (pre-branch) ZK database -> new (persistence store) database. - I'm curious how easy it would be to have ZK load a database from a file so we can have this be 100% automated.

##  Optimistic/Future TODOs (future PRs likely)
- Change ZkPersistenceStore's Category type so that we can have
  single nodes (where there is exactly one object instead of multiple)
- Simulated Benchmarks to stress the storage layer with tens of
  thousands of apps.

## Major changes 
- New Persistence API with support for ZK, Memory, and Two caching
  modes.
- Move all storage related "stuff" into marathon/core/storage
- Default the storage backend to the new ZK API.
- Lazy storage of Groups (new API only) where Apps are stored
  separately.
- Only GroupRepository is allowed write access to AppRepository
- Completely unify the persistence API between new and old, keeping
  a single, simplified chain of Repositories.
- Make versioned repositories explicit.
- Prepare for a future where typesafe config is used as opposed to a
  million command line options (storage supports it now).
- Refactor migration a bit to differentiate new API vs old API (needs a
  bit more work)
- Run repository tests against all of the backing stores.
- (in some cases) remove mocks and prefer the in memory implementation
  where it otherwise makes sense (not to simulate errors, etc).
- Add ZookeeperServerTest that will spin up/down a zk server and provide
  a client on a ephemeral port. This has the major advantage of bringing
  us closer to being able to run tests in parallel.

## Minor changes
- run scalastyle for test in compile, not test.
- add plugin to auto-resolve documentation links
- remove lots of Guice Injections for Storage.
- Use immutable.Seq across the persistence layer.
- Use OffsetDateTime (java8 time) for versioning in the new API
  as it is strictly better than timestamp.
- Make some traits have companion objects so they can be imported
  instead of mixed-in.
- Add a generic Sink stage that can drain into any collection, then wrap
  the Akka Sink API in a companion.
- Add some tooling for enriched Locks.
- Add scala friendly API around typesafe config.
- Add RichFuture that adds "asTry" which makes a future always complete,
  but with a try instead (very nice for async/await)
- Fix some oversight issues with CapConcurrentExecutions.
    - I would like to remove the actor from this and make it a
      functional transform instead.
- Add a SettableClock (the java8 time Clock) that lets you control time
  a bit easier for testing.


src/main/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_3_PersistentStore.scala
```diff
@@ -0,0 +1,157 @@
...
+    val result = oldRepo.rootVersions().mapAsync(Int.MaxValue) { version =>
+      oldRepo.rootVersion(version)
+    }.collect {
+      case Some(root) => root
+    }.concat { Source.fromFuture(oldRepo.root()) }.mapAsync(1) { root =>
+      // we store the roots one at a time with the current root last,
+      // adding a new app version for every root (for simplicity)
+      groupRepository.storeRoot(root, root.apps.values.toVector, Nil)
+    }.runFold(0) { case (acc, _) => acc + 1 }.map("root versions" -> _)
+    await(oldRepo.ids().mapAsync(Int.MaxValue)(oldRepo.delete).runWith(Sink.ignore).asTry)
```
aquamatthias: You do not wait for the delete operation in the resulting future. Surround with async and await `result` as well.

jasongilanfarr:
Remaining TODOs

- [ ] Fix GroupRepositoryImpl to not store the group if storing the apps fails (broken by the root caching, needs an additional test case that verifies the underlying store.
- [ ] Cleanup Migration more
  - [ ] Add "if storedVersion > currentVersion -> fail" and "if storedFormat is new and current format is legacy -> fail"
  - [ ] Reenable all the tests for old migrations
  - [ ] Tests for new migration.
  - [ ] ensure all old entries are deleted _after_ migrating to new.
- [ ] Fix broken test case (MarathonSchedulerActorTest) - I still don't quite understand what is happening with this one. It fails pretty unreliably and seems to be a race between persisted deployments and cancellation.

jasongilanfarr:
Benchmarks and GC are coming... I'll stack them as separate PRs.

src/test/scala/mesosphere/marathon/core/storage/migration/MigrationTo1_2_PersistenceStoreTest.scala
```diff
@@ -0,0 +1,264 @@
...
+import mesosphere.marathon.core.storage.{ LegacyInMemConfig, LegacyStorageConfig }
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.Task.{ AgentInfo, Status }
+import mesosphere.marathon.metrics.Metrics
+import mesosphere.marathon.state.{ AppDefinition, Group, MarathonTaskState, TaskFailure, Timestamp }
+import mesosphere.marathon.test.Mockito
+import mesosphere.marathon.upgrade.DeploymentPlan
+import mesosphere.util.state.FrameworkId
+
+class MigrationTo1_2_PersistenceStoreTest extends AkkaUnitTest with Mockito {
```
aquamatthias: Nice Tests! 👍

PR #4131, opened 2016-07-22
stop using deprecated protobufs for ext vols with docker apps

jdef:
* generate external vol protos using the unified container protobufs for apps using docker engine
* stop setting volume-driver at container level, instead only set volume driver per-volume

TODO
- [x] reenable the "use of deprecated code" check in `build.scala` (for Compile)
- [x] code snippet in docs that demonstrates a relative host-volume (sandbox) mount

Use Case:

AS Alice I MUST be able to declare an application specification that uses docker engine and mounts a combination of 1 or more local host-volumes AND 1 or more external volumes.

Example app definition:
```json
{
  "id": "dockervol",
  "cpus": 1,
  "mem": 2048,
  "disk": 0,
  "instances": 1,
  "container": {
    "docker": {
      "image": "gitlab/gitlab-ce:8.10.0-rc9.ce.0",
      "network": "HOST"
    },
    "type": "DOCKER",
    "volumes": [
      {
        "hostPath": "var-opt",
        "containerPath": "/var/opt/gitlab",
        "mode": "RW"
      },
      {
        "hostPath": "var-log",
        "containerPath": "/var/log/gitlab",
        "mode": "RW"
      },
      {
        "containerPath": "/etc/gitlab",
        "external": {
          "name": "jdef-etc-gitlab456",
          "provider": "dvdi",
          "options": {
            "dvdi/driver": "rexray"
          }
        },
        "mode": "RW"
      }
    ]
  }
}
```

Fix Versions:
- [x] cherry-pick this PR to support the above use case for DC/OS 1.7.x -- #4132 
- [ ] cherry-pick this PR to support the above use case for DC/OS 1.8.x

Side Effects:
1. It's no longer possible to specify a non-external **named volume** with the intent that docker will lookup the driver for you. if you want an external volume, you **must** declare it as external and pick a driver name via `dvdi/driver`. otherwise if you declare a non-external container volume with a non-absolute path MESOS will assume that it's a hostPath relative to `$MESOS_SANDBOX`

### DIFF
--- a/docs/docs/external-volumes.md
+++ b/docs/docs/external-volumes.md
@@ -172,8 +172,8 @@ The default implicit volume size is 16 GB. If you are using the Mesos containeri
 
 ### Potential Pitfalls
 
-*   If one or more external volumes are declared for a Marathon app Docker may create anonymous external volumes if the docker image specification includes one or more `VOLUME` entries. This is default docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. To prevent the creation of anonymous volumes follow these steps:
-  *  `docker inspect` the app's docker image prior to running the app in Marathon and identify the `VOLUME` entries in the specification.
+*   If one or more external volumes are declared for a Marathon app, and the Docker image specification includes one or more `VOLUME` entries, Docker may create anonymous external volumes. This is default Docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. Follow these steps to prevent Docker from creating anonymous volumes:
+  *  `docker inspect` the app's Docker image before running the app in Marathon and make a note of the `VOLUME` entries in the specification.
   *   declare non-external, host-volume mounts in your app's container specification for each `VOLUME` entry that should not map to an anonymous external volume.
   *   specify a relative `hostPath` for a host-volume to instruct Mesos to create the mount in the task's sandbox (`$MESOS_SANDBOX/$hostPath`).
 
--- a/docs/docs/external-volumes.md
+++ b/docs/docs/external-volumes.md
@@ -172,15 +172,20 @@ The default implicit volume size is 16 GB. If you are using the Mesos containeri
 
 ### Potential Pitfalls
 
+*   If one or more external volumes are declared for a Marathon app Docker may create anonymous external volumes if the docker image specification includes one or more `VOLUME` entries. This is default docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. To prevent the creation of anonymous volumes follow these steps:
+  *  `docker inspect` the app's docker image prior to running the app in Marathon and identify the `VOLUME` entries in the specification.
+  *   declare non-external, host-volume mounts in your app's container specification for each `VOLUME` entry that should not map to an anonymous external volume.
+  *   specify a relative `hostPath` for a host-volume to instruct Mesos to create the mount in the task's sandbox (`$MESOS_SANDBOX/$hostPath`).
+
 *   You can only assign one task per volume. Your storage provider might have other limitations.
 
-*   The volumes you create are not automatically cleaned up. If you delete your cluster, you must go to your storage provider and delete the volumes you no longer need. If you're using EBS, find them by searching by the `container.volumes.external.name` that you set in your Marathon app definition. This name corresponds to an EBS volume `Name` tag.
+*   The volumes you create are not automatically cleaned up. If you delete your cluster, you must go to your storage provider and delete the volumes you no longer need. If you're using EBS, find them by searching by the `container.volumes.external.name` that you set in your Marathon app definition. This name corresponds to an EBS volume `Name` tag. Any anonymous volumes that docker has created on your behalf will have been assigned a UUID for their name.
 
 *   Volumes are namespaced by their storage provider. If you're using EBS, volumes created on the same AWS account share a namespace. Choose unique volume names to avoid conflicts.
 
 *   Docker apps with external volumes on DC/OS installations must use Docker 1.8 or later.
 
-*   If you are using Amazon's EBS, it is possible to create clusters in different availability zones (AZs). If you create a cluster with an external volume in one AZ and destroy it, a new cluster may not have access to that external volume because it could be in a different AZ.
+*   If you are using Amazon's EBS, it is possible to create clusters in different availability zones (AZs). If you create a cluster with an external volume in one AZ and subsequently destroy that cluster, a new cluster may not have access to that external volume because it could be in a different AZ.
 
 *   Launch time might increase for applications that create volumes implicitly. The amount of the increase depends on several factors which include the size and type of the volume. Your storage provider's method of handling volumes can also influence launch time for implicitly created volumes.
 
--- a/docs/docs/external-volumes.md
+++ b/docs/docs/external-volumes.md
@@ -178,7 +178,7 @@ The default implicit volume size is 16 GB. If you are using the Mesos containeri
 
 ### Potential Pitfalls
 
-*   If one or more external volumes are declared for a Marathon app, and the Docker image specification includes one or more `VOLUME` entries, Docker may create anonymous external volumes. This is default Docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. Follow these steps to prevent Docker from creating anonymous volumes:
+*   If one or more external volumes are declared for a Marathon app, and the Docker image specification includes one or more `VOLUME` entries, Docker may create anonymous external volumes. This is default Docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. However, anonymous volumes are not automatically deleted and will accumulate over time unless you manually delete them. Follow these steps to prevent Docker from creating anonymous volumes:
   *  `docker inspect` the app's Docker image before running the app in Marathon and make a note of the `VOLUME` entries in the specification.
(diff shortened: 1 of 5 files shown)


jdef:
It's ALIVE: the above app spec results in a running docker container via:
```
I0722 21:20:19.955410  4814 docker.cpp:785] Running docker -H unix:///var/run/docker.sock run --cpu-shares 1024 --memory 2147483648 -e MARATHON_APP_VERSION=2016-07-22T21:19:04.604Z -e HOST=10.0.0.190 -e MARATHON_APP_RESOURCE_CPUS=1.0 -e MARATHON_APP_DOCKER_IMAGE=gitlab/gitlab-ce:8.10.0-rc9.ce.0 -e PORT_10000=24046 -e MESOS_TASK_ID=dockervol.eb1fedee-5051-11e6-a853-3270dcc6df30 -e PORT=24046 -e MARATHON_APP_RESOURCE_MEM=2048.0 -e PORTS=24046 -e MARATHON_APP_RESOURCE_DISK=0.0 -e MARATHON_APP_LABELS= -e MARATHON_APP_ID=/dockervol -e PORT0=24046 -e LIBPROCESS_IP=10.0.0.190 -e MESOS_SANDBOX=/mnt/mesos/sandbox -e MESOS_CONTAINER_NAME=mesos-0f07d0da-f02f-4f58-8cab-22bae1b60658-S0.4a367387-448c-490a-9e41-88736d1209ae -v /var/lib/mesos/slave/slaves/0f07d0da-f02f-4f58-8cab-22bae1b60658-S0/frameworks/0f07d0da-f02f-4f58-8cab-22bae1b60658-0000/executors/dockervol.eb1fedee-5051-11e6-a853-3270dcc6df30/runs/4a367387-448c-490a-9e41-88736d1209ae/var-opt:/var/opt/gitlab:rw -v /var/lib/mesos/slave/slaves/0f07d0da-f02f-4f58-8cab-22bae1b60658-S0/frameworks/0f07d0da-f02f-4f58-8cab-22bae1b60658-0000/executors/dockervol.eb1fedee-5051-11e6-a853-3270dcc6df30/runs/4a367387-448c-490a-9e41-88736d1209ae/var-log:/var/log/gitlab:rw --volume-driver=rexray -v jdef-etc-gitlab456:/etc/gitlab:rw -v /var/lib/mesos/slave/slaves/0f07d0da-f02f-4f58-8cab-22bae1b60658-S0/frameworks/0f07d0da-f02f-4f58-8cab-22bae1b60658-0000/executors/dockervol.eb1fedee-5051-11e6-a853-3270dcc6df30/runs/4a367387-448c-490a-9e41-88736d1209ae:/mnt/mesos/sandbox --net host --name mesos-0f07d0da-f02f-4f58-8cab-22bae1b60658-S0.4a367387-448c-490a-9e41-88736d1209ae gitlab/gitlab-ce:8.10.0-rc9.ce.0
```

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -306,22 +306,21 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
-    def vol(name: String): Option[MesosProtos.Volume] = {
+    def vol(path: String): Option[MesosProtos.Volume] = {
       if (taskInfo.hasContainer) {
-        taskInfo.getContainer.getVolumesList.asScala.find(_.getHostPath == name)
+        taskInfo.getContainer.getVolumesList.asScala.find(_.getHostPath == path)
       } else None
     }
 
     assert(taskInfo.getContainer.getVolumesList.size > 0, "check that container has volumes declared")
     assert(!taskInfo.getContainer.getDocker.hasVolumeDriver, "docker spec should not define a volume driver")
```
gkleiman: Can't we remove this assertion (and the other similar ones), so that we can reenable the scalac `deprecated` check for tests?

`Docker.volumeDriver` is deprecated, so the scalac `deprecated` check ensures that it is not set by the "production" code.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -365,15 +360,90 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
       } else None
     }
 
-    assert(taskInfo.getContainer.getVolumesList.size > 0, "check that container has volumes declared")
-    assert(taskInfo.getContainer.getDocker.hasVolumeDriver, "docker spec should define a volume driver")
-    assert(taskInfo.getContainer.getDocker.getVolumeDriver == "ert", "docker spec should choose ert driver")
+    assert(taskInfo.getContainer.getVolumesList.size == 2, "check that container has volumes declared")
+    assert(!taskInfo.getContainer.getDocker.hasVolumeDriver, "docker spec should not define a volume driver")
+
```
gkleiman: What do you think about doing something like the following?

```scala
taskInfo.getContainer.getVolumesList.asScala should contain theSameElementsAs Seq(volumeWith(...), volumeWith(...))
```

We are trying to deprecate asserts in favour of scalatest matchers in our tests.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
jdef: I'd rather keep it for now since this is a crucial bit of state that Mesos still uses to make decisions.

docs/docs/external-volumes.md
```diff
@@ -172,15 +172,20 @@ The default implicit volume size is 16 GB. If you are using the Mesos containeri
 
 ### Potential Pitfalls
 
+*   If one or more external volumes are declared for a Marathon app, and the Docker image specification includes one or more `VOLUME` entries, Docker may create anonymous external volumes. This is default Docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. Follow these steps to prevent Docker from creating anonymous volumes:
+  *  `docker inspect` the app's Docker image before running the app in Marathon and make a note of the `VOLUME` entries in the specification.
```
sascala: "identify" seemed ambiguous here.

docs/docs/external-volumes.md
```diff
@@ -172,15 +172,20 @@ The default implicit volume size is 16 GB. If you are using the Mesos containeri
 
 ### Potential Pitfalls
 
+*   If one or more external volumes are declared for a Marathon app, and the Docker image specification includes one or more `VOLUME` entries, Docker may create anonymous external volumes. This is default Docker behavior with respect to volume management when the `--volume-driver` flag is passed to `docker run`. Follow these steps to prevent Docker from creating anonymous volumes:
```
sascala: Why wouldn't I want Docker to create anonymous volumes?

docs/docs/external-volumes.md
```diff
@@ -172,15 +172,20 @@ The default implicit volume size is 16 GB. If you are using the Mesos containeri
...
+  *  `docker inspect` the app's Docker image before running the app in Marathon and make a note of the `VOLUME` entries in the specification.
+  *   declare non-external, host-volume mounts in your app's container specification for each `VOLUME` entry that should not map to an anonymous external volume.
```
sascala: Is it worth have a code snippet here?

docs/docs/external-volumes.md
jdef: So we're talking about external volumes in this section and the default
implementation, rexray. The reference implementation that we offer (via AWS
templates) integrates with EBS. Also realize that neither dc/os or marathon
offer volume lifecycle management features for external volumes. Meaning
that Dan (or Alice) needs keep track of all volumes used by apps and delete
old/expired volumes when they're no longer needed.

To answer your question: when Docker creates anonymous external volumes for
an app (and doesn't subsequently delete them (expected behavior)) a couple
of things happen:
* The anonymous volumes linger on in the storage system (on AWS this means
lots of EBS volumes accumulating over time)
* Dan (or Alice) need to identify these volumes and delete them; this is
relatively straight forward given the obvious UUID nature of the names but
the volumes created this way aren't labeled by a mechanism that makes them
easy to select (and delete) in bulk.
* Since it's not immediately clear that Docker is doing this, Dan (or
Alice) may be (a) overwhelmed by suddenly needing to manage a large number
of volumes, and; (b) surprised by a very large AWS bill at the end of the
month because they weren't tracking all the anonymous volumes that Docker
created and suddenly are using 30TB of space (in the form of detached,
anonymous volumes) on EBS.

docs/docs/external-volumes.md
jdef: Probably worth a code snippet. I'll put something together

jdef: done

docs/docs/external-volumes.md
jdef: ack

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
jdef: ugh. more DSL

jdef: i think rewriting the assertions should be handled for the entire file in a separate PR, not this one

gkleiman: I don't agree, we should write all new tests in the new style, regardless of the rest of the file.

gkleiman:
@jdef please add jiras/tickets to fix the issues pointed out during the review.

gkleiman:
cc/ @aquamatthias

jdef:
done

Issue #3901, opened 2016-05-16
Allow specification of >50% minimum healthy instances for persistent tasks


timcharper:
There doesn't seem to be a really good reason for this restriction. I'd prefer a scenario in which my rollout stalls because minimum healthy instances cannot be satisfied, than to have more unhealthy instances than can be tolerated by the cluster.

hotstepper13:
If i remind correctly you could set minimumHealthCapacity to 0.5 which means that at least 50% of your tasks must remain active.

In conjunction with maximumOverCapacity you can define how your deployment should work.

For instance:
All tasks are stopped and redeployt
minimumHealthCapacity: 0
maximumOverCapacity: 0

Stop 50% of tasks, redeploy them, wait until they are healthy and then stop and redeploy the remaining
minimumHealthCapacity: 0.5
maximumOverCapacity: 0

Let all your taks live until the new ones become healthy
minimumHealthCapacity: 1
maximumOverCapacity: 1

timcharper:
Hi @hotstepper13 thanks for the reply! This was all clear to me. Ultimately, https://github.com/mesosphere/marathon/issues/3900 is what I want, but I didn't want to conflate the features too much so I opened them separately.

My requirement is that at most 1 instance is unhealthy at a given time during deploy; I absolutely cannot over-provision. This requirement could be satisfied by allowing me to specify >50% healthy instances minimum, or by specifying `maximumUnhealthyQty`, or something.

meichstedt:
@timcharper I can see your point. In general we try to prevent a deployment from becoming stalled, if possible, because there's no easy way to find out why it's not proceeding. However, the upgradeStrategy is already interpreted in context of the _type_ of application: for apps using persistent volumes, Marathon won't allow any overCapacity (which it **does** for normal apps even if `maximumOverCapacity = 0`).

I'd suggest allowing any `minimumHealthCapacity < 1` but killing at least 1 task to get an upgrade rolling. Or, would you rather prefer Marathon **not** killing 1 task if that'd keep less than the specified `minimumHealthCapacity`?

timcharper:
[...]

PR #4138, opened 2016-07-25
allow 100% minimum health capacity setting for resident tasks

timcharper:
When push comes to shove, Marathon will allow one instance to be unhealthy vs. having an extra resident task or having a deployment get stuck.

Fixes https://github.com/mesosphere/marathon/issues/3901


timcharper:
We have this patch deployed and can confirm that allowing the specification of 100% minimum capacity (and 0% over capacity) provides the desired behavior of restarting one task at a time.

meichstedt:
Thanks @timcharper, LGTM! Waiting for tests to pass ..

PR #4139, opened 2016-07-25
allow 100% minimum health capacity setting for resident tasks (#4138)

meichstedt:
When push comes to shove, Marathon will allow one instance to be unhealthy vs. having an extra resident task or having a deployment get stuck.

Fixes https://github.com/mesosphere/marathon/issues/3901

Cherry-picking from releases/1.2

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/UpgradeStrategy.scala
+++ b/src/main/scala/mesosphere/marathon/state/UpgradeStrategy.scala
@@ -27,7 +27,7 @@ object UpgradeStrategy {
   }
 
   lazy val validForResidentTasks: Validator[UpgradeStrategy] = validator[UpgradeStrategy] { strategy =>
-    strategy.minimumHealthCapacity should be <= 0.5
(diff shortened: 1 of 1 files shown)


aquamatthias:
Thanks!

jdef:
why isn't there a related unit/integration test for this change? ext vols probably implements similar validation and so may need updates

meichstedt:
@jdef The functionality is tested in `TaskReplaceActorTest`. Before this change, resident apps were forced to have a `minimumHealthCapacity <= 0.5` which led to upgrade deployments where at least half the instances were killed when starting the deployment. The change to allow any `minimumHealthCapacity` merely allows to adjust the percentage of tasks being killed at once.

So far, there is no such validation for apps that use Ext Volumes.

jdef:
xref
https://github.com/mesosphere/marathon/blob/6a5cee2778c4d852a3db83e4cca15e4795b36b41/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala#L206

PR #4141, opened 2016-07-25
task state refactoring

unterstein:
Accidentally the original pull request https://github.com/mesosphere/marathon/pull/4118 was messed up, therefore this is the new one with the adapted feedback of https://github.com/mesosphere/marathon/pull/4118.

Aim of this refactoring was to reduce the usage of mesos.Protos.TaskStatus and to merge the existing MarathonTaskStatus and the MesosTaskStatus and introduce therefore a consistent and mandatory marathonTaskStatus for each Task.


src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -388,7 +393,36 @@ class MigrationTo1_2(deploymentRepository: DeploymentRepository) {
       await(deploymentRepository.store.expunge(it.next))
     }
 
-    log.info("Finished 1.2 migration")
+    val store = taskRepository.store
+    //scalastyle:off null
```
gkleiman: You can use `hasMarathonStatus()` later on and avoid nulls.

src/main/scala/mesosphere/marathon/state/Migration.scala
```diff
@@ -388,7 +393,36 @@ class MigrationTo1_2(deploymentRepository: DeploymentRepository) {
...
+        case None => Future.failed(new MigrationFailedException(s"Inconsistency in the task store detected, " +
+          s"task with id $id not found, but delivered in allIds()."))
+      }
+    }
+
+    val migratedTasks = for {
+      ids <- store.names()
+      tasks <- {
+        log.info(s"Discovered ${ids.size} tasks for status migration")
+        Future.sequence(ids.map(loadAndMigrateTasks))
```
gkleiman: The ZK stores are unbounded and in the past users have reported issues when we tried to use Future.sequence + ZK operations in migrations.

That's why I remove them serially in lines 391-394. I think we should do something similar here.

unterstein: ok, i`ll change this.

src/main/scala/mesosphere/marathon/state/Migration.scala
unterstein: thanks!

src/main/proto/marathon.proto
```diff
@@ -141,6 +141,23 @@ message MarathonTask {
...
+    Finished = 5;
+    Killed = 6;
+    Killing = 7;
+    Lost = 8;
+    Running = 9;
+    Staging = 10;
+    Starting = 11;
+    Unreachable = 12;
+    Gone = 13;
+    Unknown = 14;
```
gkleiman: This should be `Unknown = 0` to make the protos as backwards-compatible as possible.

Messages will return the first enum value (which should be `Unknown` if they don't know the enum value that they're trying to parse).

unterstein: Unknown will be an explicit mesos status, therefore I introduced a state Invalid with order number 0 which is mapped to a serialization error.

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -84,12 +83,11 @@ private[health] class HealthCheckActor(
...
 
   def dispatchJobs(): Unit = {
-    import TemporarilyUnreachable.isUnreachable
-
     log.debug("Dispatching health check jobs to workers")
     taskTracker.appTasksSync(app.id).foreach { task =>
+
       task.launched.foreach { launched =>
-        if (launched.runSpecVersion == app.version && launched.hasStartedRunning && !isUnreachable(task)) {
+        if (launched.runSpecVersion == app.version && launched.hasStartedRunning && !task.isUnreachable) {
```
meichstedt: This can simply be
`if (launched.runSpecVersion == app.version && task.isRunning)` 😁

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
```diff
@@ -57,7 +58,8 @@ class TaskOpFactoryImpl(
           ),
           runSpecVersion = runSpec.version,
           status = Task.Status(
-            stagedAt = clock.now()
+            stagedAt = clock.now(),
+            taskStatus = MarathonTaskStatus.Staging
```
meichstedt: Wouldn't this be `Created`?

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskOpFactoryImpl.scala
```diff
@@ -145,7 +147,8 @@ class TaskOpFactoryImpl(
           task.taskId,
           runSpecVersion = spec.version,
           status = Task.Status(
-            stagedAt = clock.now()
+            stagedAt = clock.now(),
+            taskStatus = MarathonTaskStatus.Staging
```
meichstedt: `Created`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -61,11 +61,14 @@ sealed trait Task {
...
   def runSpecId: PathId = taskId.runSpecId
 
+  def status: Task.Status
+
   def launchedMesosId: Option[MesosProtos.TaskID] = launched.map { _ =>
     // it doesn't make sense for an unlaunched task
     taskId.mesosTaskId
   }
 
+  // TODO ju remove
```
meichstedt: Please remove the TODO for now and create a follow up ticket containing removed TODO's

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -112,23 +115,27 @@ object Task {
...
+        val updated = copy(status = status.copy(
+          mesosStatus = Some(mesosStatus),
+          taskStatus = taskStatus))
         TaskStateChange.Expunge(updated)
 
       // case 3: health or state updated
-      case TaskStateOp.MesosUpdate(_, taskStatus, now) =>
-        updatedHealthOrState(status.mesosStatus, taskStatus.mesosStatus) match {
+      case TaskStateOp.MesosUpdate(_, taskStatus, mesosStatus, now) =>
+        updatedHealthOrState(status.mesosStatus, Some(mesosStatus)) match {
```
meichstedt: The second argument is now always `Some` – please change the signature of `updatedHealthOrState` and simplify the implementation!

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -457,4 +475,23 @@ object Task {
...
+    def isKilling: Boolean = task.status.taskStatus == MarathonTaskStatus.Killing
+    def isRunning: Boolean = task.status.taskStatus == MarathonTaskStatus.Running
+    def isStaging: Boolean = task.status.taskStatus == MarathonTaskStatus.Staging
+    def isStarting: Boolean = task.status.taskStatus == MarathonTaskStatus.Starting
+    def isUnreachable: Boolean = task.status.taskStatus == MarathonTaskStatus.Unreachable
+    def isGone: Boolean = task.status.taskStatus == MarathonTaskStatus.Gone
+    def isUnknown: Boolean = task.status.taskStatus == MarathonTaskStatus.Unknown
+    def isDropped: Boolean = task.status.taskStatus == MarathonTaskStatus.Dropped
+
+    def isReachable: Boolean = isStaging || isStarting || isRunning || isKilling
```
meichstedt: Please add a scaladoc here, as it is not clear from the name what is included. From the perspective of `!unreachable`, the given set makes sense. Yet, from the perspective of _can we reach this task_, this would probably only be `isRunning || is Killing` (we can't perform a HealthCheck against a staging or starting task).

src/main/scala/mesosphere/marathon/core/task/bus/impl/TaskStatusEmitterImpl.scala
```diff
@@ -11,7 +11,7 @@ private[bus] class TaskStatusEmitterImpl(internalTaskStatusEventStream: Internal
 
   override def publish(taskChanged: TaskChanged): Unit = {
     taskChanged.stateOp match {
-      case TaskStateOp.MesosUpdate(task, status, timestamp) =>
+      case TaskStateOp.MesosUpdate(task, marathonTaskStatus, mesosStatus, timestamp) =>
```
meichstedt: since we don't use any of the extracted vals here, this could be `case update: TaskStateOp.MesosUpdate =>`
(this avoids extracting the values and assigning them)

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.mesos
+
+/**
+  * To define the status of a Task, this trait is used and stored for each Task in Task.Status.
+  * The existing case objects are:
+  * - marathon exclusive status
+  * - representations of the mesos.Protos.TaskStatus
+  * - mapping of existing (soon-to-be deprecated) mesos.Protos.TaskStatus.TASK_LOST to the new representations
+  */
+// TODO ju handle garbage flag
```
meichstedt: remove the TODO

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatus.scala
```diff
@@ -0,0 +1,82 @@
...
+  //scalastyle:off cyclomatic.complexity
+  def apply(taskStatus: mesos.Protos.TaskStatus): MarathonTaskStatus = {
+    taskStatus.getState match {
+      case TASK_ERROR => Error
+      case TASK_FAILED => Failed
+      case TASK_FINISHED => Finished
+      case TASK_KILLED => Killed
+      case TASK_KILLING => Killing
+      case TASK_LOST => taskStatus.getReason match {
+        case state: mesos.Protos.TaskStatus.Reason if MarathonTaskStatusMapping.WontComeBack(state) => Gone
```
meichstedt: `case reason` instead of `case state` makes more sense imo

src/main/scala/mesosphere/marathon/core/task/state/MarathonTaskStatusMapping.scala
```diff
@@ -0,0 +1,32 @@
+package mesosphere.marathon.core.task.state
+
+import org.apache.mesos
+
+/**
+  * To convert `old` mesos.Protos.TaskStatus.TASK_LOST to a proper
+  * MarathonTaskStatus representation (not `Lost` existing), the TaskStatus.Reason is needed. This object
+  * provides the according Reasons for MarathonTaskStatus.Gone, MarathonTaskStatus.Unreachable,
+  * MarathonTaskStatus.Unknown and MarathonTaskStatus.Dropped.
```
meichstedt: 1. I don't see a set for `Dropped` (iirc it's the fallback when the reason doesn't match any of the defines sets in here)
2. Since the comment nicely describes the relation to `Gone, Unreachable and Unknown`, maybe rename the Sets accordingly? If this would imply naming conflicts, prefix with ReasonXyz?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -61,11 +61,14 @@ sealed trait Task {
 
   def runSpecId: PathId = taskId.runSpecId
 
+  def status: Task.Status
```
meichstedt: since the status is now a member of Task, would it be easily possible to remove it from Launched?

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/PostToEventStreamStepImpl.scala
```diff
@@ -29,8 +28,8 @@ class PostToEventStreamStepImpl @Inject() (eventBus: EventStream, clock: Clock)
     taskChanged match {
       // case 1: Mesos status update => update or expunge
       // In this case, we post the OLD state - when terminated, a persistent task no longer has a launched
-      case TaskChanged(MesosUpdate(task, WithMesosStatus(status), now), EffectiveTaskStateChange(_)) =>
-        postEvent(clock.now(), Some(status), task)
+      case TaskChanged(MesosUpdate(task, status, mesosStatus, now), EffectiveTaskStateChange(_)) =>
+        postEvent(clock.now(), Option(mesosStatus), task)
```
meichstedt: `Some(mesosStatus)`

src/main/scala/mesosphere/marathon/core/task/Task.scala
unterstein: I think we should refactor/remove the `def launched: Option[Launched]` usage in the follow up issue https://github.com/mesosphere/marathon/issues/4153

meichstedt:
@unterstein I've added my final comments above – otherwise ready to go from my side 👍

aquamatthias:
Thanks!

PR #4149, opened 2016-07-27
no code change

meichstedt:
just for testing :)


foo
```diff
@@ -0,0 +1 @@
+bar
```
gkleiman: This should be baz.

meichstedt: I tried baz, but it didn't fit well into our architecture.

meichstedt:
Currently, both the velocity _and_ TeamCity builds have failed here. As the commit message states, there was no code change.

aquamatthias:
I like the baz architecture!

PR #4154, opened 2016-07-28
Correct son for example app definition.

aquamatthias:


unterstein:
`docs/docs/rest-api/public/api/v2/examples/app.json` is now `VALID JSON (RFC 4627)`

PR #4170, opened 2016-08-02
Fixes #4100 - Use output_file option in FetchUri

### LINKED ISSUE #4100: Fetch URI with query string
When I need to fetch a URI that contains a query string the downloaded file will contain the query string in the name. Is there a way to avoid this? Is it possible to specify the file name?

mchataigner:
mesos.proto defines URI message type with output_file option.
Make FetchUri use this option in order to be able to define the name of
 files we want to add to the mesos task.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -563,8 +563,9 @@ trait FetchUriFormats {
       (__ \ "uri").format[String] ~
       (__ \ "extract").formatNullable[Boolean].withDefault(true) ~
       (__ \ "executable").formatNullable[Boolean].withDefault(false) ~
-      (__ \ "cache").formatNullable[Boolean].withDefault(false)
-    )(FetchUri(_, _, _, _), unlift(FetchUri.unapply))
+      (__ \ "cache").formatNullable[Boolean].withDefault(false) ~
+      (__ \ "outputFile").formatNullable[String]
+    )(FetchUri(_, _, _, _, _), unlift(FetchUri.unapply))
   }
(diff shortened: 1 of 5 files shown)


mchataigner:
Hello, I do not quite understand the error I got inside TeamCity Build.
It seems that it has nothing to do with my patch. Could you help me, please?
Thanks.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -563,8 +563,9 @@ trait FetchUriFormats {
       (__ \ "uri").format[String] ~
       (__ \ "extract").formatNullable[Boolean].withDefault(true) ~
       (__ \ "executable").formatNullable[Boolean].withDefault(false) ~
-      (__ \ "cache").formatNullable[Boolean].withDefault(false)
-    )(FetchUri(_, _, _, _), unlift(FetchUri.unapply))
+      (__ \ "cache").formatNullable[Boolean].withDefault(false) ~
+      (__ \ "output-file").formatNullable[String]
```
aquamatthias: Marathon uses camel-case for json properties. Please rename to `outputFile`.

aquamatthias:
@mchataigner Thanks for this PR! One small comment regarding naming, otherwise looks good.
Would be great, if you could add the new property also to the example `public/api/v2/examples/app.json` and the schema `public/api/v2/schema/AppDefinition.json`.

The failure in TeamCity does not seem related to your PR. Will try to rerun.

mchataigner:
Done.
Thanks for your help.

aquamatthias:
Thanks! LGTM. Waiting for the tests to pass.

Issue #4167, opened 2016-08-01
Usability problem with new service port checks after upgrade. 


joerg84:
After upgrading a cluster with many apps which might have the same service port allocated from before the update (bad configuration, but note that the ports might not be used),
I cannot restart any app as the check is checking all service for all apps -not just the one I upgraded...

I would expect that when upgrading a cluster with many apps, it forces me to only update the app I want to restart, but not all other apps which are running in the cluster.

gkleiman:
cc/ @aquamatthias

unterstein:
Fixed through #4172

PR #4172, opened 2016-08-02
Moved unique port validation from Group to DeploymentPlan

unterstein:
Like we discussed today:
Moved unique port validation from Group to DeploymentPlan, like requested in https://github.com/mesosphere/marathon/issues/4167

### DIFF
--- a/src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala
+++ b/src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala
@@ -343,12 +343,8 @@ object DeploymentPlan {
         // We keep track of the total number of ports to support an
         // early exit condition.
 
-        val createdOrUpdatedApps = plan.createdOrUpdatedApps.map(app => app.id).toSet
         var ports: Int = 0
-        val allPorts =
-          new mutable.HashMap[Int, mutable.Set[AppDefinition.AppKey]] with mutable.MultiMap[Int, AppDefinition.AppKey]
-
-        val createdOrUpdatedPorts =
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala
```diff
@@ -321,7 +321,65 @@ object DeploymentPlan {
...
+          Success
+        } else {
+          // Otherwise we find all ports that have more than 1 app
+          // interested in them.
+
+          // We report all the conflicting apps along with which other
+          // apps they conflict with.
+          val violations = for {
+            // only yield port as conflict if createdOrUpdatedApps are involved and therefore in the intersection set
+            (port, apps) <- merged if apps.size > 1 && apps.intersect(createdOrUpdatedApps).nonEmpty
```
unterstein: apps.intersect(createdOrUpdatedApps).nonEmpty is the interesting section here

aquamatthias: This will filter the list of `(port, apps)`. It could lead to an empty array for violations, which still would cause a Failure - or (the filter is applied in the else/failure branch)?

aquamatthias:
Thanks! LGTM. Wait for the tests to finish.

aquamatthias:
@unterstein Have you tested this patch on top of the OVH state?

aquamatthias:
And please cherry pick in 1.2 branch.

unterstein:
OVH tested positive locally, cherry pick done.

PR #4177, opened 2016-08-04
Use Mesos 1.0.0

aquamatthias:
Use mesos 1.0.0
Remove deprecated calls in test code.


unterstein:
looks awesome to me, merged :)

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -313,7 +313,6 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
     }
 
     assert(taskInfo.getContainer.getVolumesList.size > 0, "check that container has volumes declared")
-    assert(!taskInfo.getContainer.getDocker.hasVolumeDriver, "docker spec should not define a volume driver")
```
jdef: concerned that these asserts were removed. i had hoped these would linger here until mesos removes the docker.volume_driver field

PR #4178, opened 2016-08-04
New Persistence API with higher scalability limits

jasongilanfarr:
Major changes

- New Persistence API with support for ZK, Memory, and Two caching modes.
- Move all storage related "stuff" into marathon/core/storage
- Default the storage backend to the new ZK API.
- Lazy storage of Groups (new API only) where Apps are stored separately.
- Only GroupRepository is allowed write access to AppRepository
- Completely unify the persistence API between new and old, keeping a single, simplified chain of Repositories.
- Make versioned repositories explicit.
- Prepare for a future where typesafe config is used as opposed to a million command line options (storage supports it now).
- Refactor migration a bit to differentiate new API vs old API (needs a bit more work)
- Run repository tests against all of the backing stores.
- (in some cases) remove mocks and prefer the in memory implementation where it otherwise makes sense (not to simulate errors, etc).
- Add ZookeeperServerTest that will spin up/down a zk server and provide a client on a ephemeral port. This has the major advantage of bringing us closer to being able to run tests in parallel.
- Add a Garbage Collector to the new persistence store that will try to enforce max versions, but in the background.
  The garbage collector is triggered by deployment plan deletion.

Minor changes

- run scalastyle for test in compile, not test.
- add plugin to auto-resolve documentation links
- remove lots of Guice Injections for Storage.
- Use immutable.Seq across the persistence layer.
- Use OffsetDateTime (java8 time) for versioning in the new API as it is strictly better than timestamp.
- Make some traits have companion objects so they can be imported instead of mixed-in.
- Add a generic Sink stage that can drain into any collection, then wrap the Akka Sink API in a companion.
- Add some tooling for enriched Locks.
- Add scala friendly API around typesafe config.
- Add RichFuture that adds "asTry" which makes a future always complete, but with a try instead (very nice for async/await)
- Fix some oversight issues with CapConcurrentExecutions.
     I would like to remove the actor from this and make it a functional transform instead.
- Add a SettableClock (the java8 time Clock) that lets you control time a bit easier for testing.


aquamatthias:
GCActorTest is failing:
```
GcActorTest:
GcActor when compacting should 
- block plans with deleted roots until compaction completes *** FAILED *** (32 milliseconds)
  false was not true (GcActorTest.scala:340)
  org.scalatest.exceptions.TestFailedException:
  at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
  at org.scalatest.Matchers$ShouldMethodHelper$.shouldMatcher(Matchers.scala:6254)
  at org.scalatest.Matchers$AnyShouldWrapper.should(Matchers.scala:6288)
```

src/test/scala/mesosphere/marathon/core/storage/repository/GcActorTest.scala
```diff
@@ -0,0 +1,482 @@
+package mesosphere.marathon.core.storage.repository
```
aquamatthias: move to impl

src/test/scala/mesosphere/marathon/core/storage/repository/GroupRepositoryTest.scala
```diff
@@ -0,0 +1,215 @@
+package mesosphere.marathon.core.storage.repository
```
aquamatthias: move to impl

aquamatthias:
Package structure: ideally we would have separated modules:
- module 1: defines abstract how to store things, with implementations for specific storage layer. This module can be part of a library that can be used as is elsewhere (e.g. metronome)
- modules 2: uses module1  and defines specific classes used in Marathon. This module will be part of Marathon only.

I think we have the classes and traits, but the modules/namespace mixes both currently.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/GcActor.scala
```diff
@@ -0,0 +1,438 @@
...
+  * - While deleting, check any store requests to see if they _could_ conflict with the in progress deletions.
+  *   If and only if there is a conflict, 'block' the store (via a promise/future) until the deletion completes.
+  *   If there isn't a conflict, let it save anyway.
+  * - When the deletion completes, inform any attempts to store a potential conflict that it may now proceed,
+  *   then transition back to idle or scanning depending on whether or not one or more additional GC Requests
+  *   were sent to the actor.
+  * - If compact fails for any reason,  transition back to idle or scanning depending on whether or not one or
+  *   more additional GC Requests were sent to the actor.
+  */
+private[storage] class GcActor[K, C, S](
```
aquamatthias: Making this an FSM was a really good idea - very clean.

src/main/scala/mesosphere/marathon/core/event/impl/callback/SubscribersKeeperActor.scala
```diff
@@ -38,38 +41,46 @@ class SubscribersKeeperActor(val store: EntityStore[EventSubscribers]) extends A
...
 
   protected[this] def add(callbackUrl: String): Future[EventSubscribers] =
-    store.modify(Subscribers) { deserialize =>
-      val existingSubscribers = deserialize()
-      if (existingSubscribers.urls.contains(callbackUrl)) {
-        log.info("Existing callback {} resubscribed.", callbackUrl)
-        existingSubscribers
-      } else EventSubscribers(existingSubscribers.urls + callbackUrl)
-    }
+    lockManager.executeSequentially("") {
```
aquamatthias: Add a message here.

src/main/scala/mesosphere/marathon/core/event/impl/callback/SubscribersKeeperActor.scala
```diff
@@ -38,38 +41,46 @@ class SubscribersKeeperActor(val store: EntityStore[EventSubscribers]) extends A
...
-
-      if (existingSubscribers.urls.contains(callbackUrl))
-        EventSubscribers(existingSubscribers.urls - callbackUrl)
-
-      else {
-        log.warning("Attempted to unsubscribe nonexistent callback {}", callbackUrl)
-        existingSubscribers
-      }
-    }
+    lockManager.executeSequentially("") {
```
aquamatthias: add a message here

src/main/scala/mesosphere/marathon/core/storage/repository/impl/legacy/EntityRepository.scala
```diff
@@ -0,0 +1,214 @@
+package mesosphere.marathon.core.storage.repository.impl.legacy
```
aquamatthias: Rename file to LegacyEntityRepository

src/main/scala/mesosphere/marathon/core/storage/repository/impl/legacy/GroupEntityRepository.scala
```diff
@@ -0,0 +1,53 @@
+package mesosphere.marathon.core.storage.repository.impl.legacy
```
aquamatthias: Just curious, why this is not defined in LegacyEntityRepository.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/legacy/store/EntityStore.scala
```diff
@@ -1,10 +1,12 @@
-package mesosphere.marathon.state
+package mesosphere.marathon.core.storage.repository.impl.legacy.store
```
aquamatthias: Can you please also move also move all relevant marathon/state/*Test.scala classes so they live side by side?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/PersistenceStoreRepository.scala
```diff
@@ -0,0 +1,53 @@
...
+import akka.http.scaladsl.marshalling.Marshaller
+import akka.http.scaladsl.unmarshalling.Unmarshaller
+import akka.stream.scaladsl.Source
+import akka.{ Done, NotUsed }
+import mesosphere.marathon.core.storage.repository.{ Repository, VersionedRepository }
+import mesosphere.marathon.core.storage.store.{ IdResolver, PersistenceStore }
+
+import scala.concurrent.Future
+
+class PersistenceStoreRepository[Id, V, K, C, S](
```
aquamatthias: Perhaps add a little scala doc: this is the default repository...

jasongilanfarr:
I think this is still a tiny bit racey- I ran the suite in a loop for
10,000 iterations. 37 of them failed.

aquamatthias:
LGTM. @kensipe @unterstein please help reviewing the latest commits to get this PR merged

jasongilanfarr:
Retest this please

jasongilanfarr:
retest this please

jasongilanfarr:
retest this please

.sbtopts
```diff
@@ -1,5 +1,15 @@
--mem 4096
+-mem 8192
 -J-XX:MaxMetaspaceSize=512m
 -J-XX:+UseConcMarkSweepGC
 -J-XX:+CMSClassUnloadingEnabled
-
+-J-server
+-J-XX:+CMSParallelRemarkEnabled
+-J-XX:+UseCMSInitiatingOccupancyOnly
+-J-XX:+ScavengeBeforeFullGC
```
jdef: stutter

scalastyle-config.xml
```diff
@@ -115,11 +115,6 @@
  </check>
  <check level="error" class="org.scalastyle.scalariform.DeprecatedJavaChecker" enabled="true"/>
  <check level="error" class="org.scalastyle.scalariform.EmptyClassChecker" enabled="true"/>
- <check level="error" class="org.scalastyle.scalariform.ClassTypeParameterChecker" enabled="true">
```
jdef: why remove this? Aren't single-character type parameters idiomatic?

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -37,7 +38,7 @@ class MarathonScheduler @Inject() (
     frameworkId: FrameworkID,
     master: MasterInfo): Unit = {
     log.info(s"Registered as ${frameworkId.getValue} to master '${master.getId}'")
-    frameworkIdUtil.store(frameworkId)
+    Await.result(frameworkIdRepository.store(FrameworkId.fromProto(frameworkId)), zkTimeout)
```
jdef: Looking at the underlying JNI glue code (mesos project) it appears that any runtime execption in this callback will cause the mesos driver to `abort()`. Which is probably fine ... I just don't see that fact documented anywhere except the c++ code.

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -136,7 +137,7 @@ class MarathonScheduler @Inject() (
   protected def suicide(removeFrameworkId: Boolean): Unit = {
     log.error(s"Committing suicide!")
 
-    if (removeFrameworkId) Await.ready(frameworkIdUtil.expunge(), config.zkTimeoutDuration)
+    if (removeFrameworkId) Await.ready(frameworkIdRepository.delete(), config.zkTimeoutDuration)
```
jdef: If this times out we still probably want to asyncExit, right?

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -334,7 +338,7 @@ object MarathonSchedulerActor {
     createSchedulerActions: ActorRef => SchedulerActions,
     deploymentManagerProps: SchedulerActions => Props,
     historyActorProps: Props,
-    appRepository: AppRepository,
+    appRepository: ReadOnlyAppRepository,
```
jdef: :+1: I like the more narrow interface

src/main/scala/mesosphere/marathon/core/CoreGuiceModule.scala
```diff
@@ -164,11 +206,15 @@ class CoreGuiceModule extends AbstractModule {
       capMetrics,
       actorRefFactory,
       "serializeTaskStatusUpdates",
-      maxParallel = config.internalMaxParallelStatusUpdates(),
+      maxConcurrent = config.internalMaxParallelStatusUpdates(),
       maxQueued = config.internalMaxQueuedStatusUpdates()
-    )
+    )(ExecutionContext.global)
```
jdef: isn't this the default? wondering why this is passed explicitly

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
```diff
@@ -106,8 +106,18 @@ class CuratorElectionService(
       connectString(config.zkHosts).
       sessionTimeoutMs(config.zooKeeperSessionTimeout().toInt).
       aclProvider(new ACLProvider {
+        val rootAcl = {
```
jdef: does this fix some related (but not specific to this PR) bug?

src/main/scala/mesosphere/marathon/core/event/impl/callback/SubscribersKeeperActor.scala
```diff
@@ -38,38 +42,46 @@ class SubscribersKeeperActor(val store: EntityStore[EventSubscribers]) extends A
...
+          log.info("Existing callback {} resubscribed.", callbackUrl)
+          subscribers
+        } else EventSubscribers(subscribers.urls + callbackUrl)
+
+        if (updated != subscribers) {
+          await(store.store(updated))
+        }
+        updated
+      }(ExecutionContext.global)
+    }(ExecutionContext.global) // blocks a thread, don't block the actor.
```
jdef: just to be sure I understand:

CURRENT STATE: executeSequentially immediately blocks on obtaining the lock and only releases the lock once the `async {}` completes. executeSequentially may return before the lock is actually released.

FUTURE STATE: invoking executeSequentially will not block, and instead will queue an async operation that will be executed exclusive of all others sync'd to the same lock-name?

scalastyle-config.xml
jasongilanfarr: The way scalacheck does them breaks some of the common idioms, e.g.:

sealed trait X
...
assume 'y' is of type X
```
y match {
   case SomeSpecific(x) =>
   case SomeOtherSpecific(z) =>
   case catchOthers => 
}
```
catchOthers is _guaranteed_ to be of type 'X' and the scalastyle check complains because the type isn't there.

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
jasongilanfarr: Don't think anything changed there... the name of `expunge` is just `delete` now.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -306,23 +310,23 @@ class MarathonSchedulerActor private (
   }
 
   def deploy(driver: SchedulerDriver, plan: DeploymentPlan): Unit = {
-    deploymentRepository.store(plan).foreach { _ =>
+    deploymentRepository.store(plan).foreach { done =>
```
jasongilanfarr: It could, a ton of Unit could be done. I accidentally named the useless 'done' after running into some strange errors and forgot to remove it.

src/main/scala/mesosphere/marathon/core/CoreGuiceModule.scala
jasongilanfarr: There isn't an implicit in scope.

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
jasongilanfarr: accident, but harmless.

src/main/scala/mesosphere/marathon/core/event/impl/callback/SubscribersKeeperActor.scala
jasongilanfarr: Old:
PersistenceStore would let you provide a `change` method that updated the existing state.

New:
PersistenceStore doesn't provide you a method to change the existing state given the old state. So you have to read it first yourself if you want a "go grab what I stored last time, append something to it, then store it". And you probably want only a single writer allowed to do this at a time.

src/main/scala/mesosphere/marathon/core/storage/store/impl/BasePersistenceStore.scala
```diff
@@ -0,0 +1,152 @@
...
+    k: Id,
+    version: OffsetDateTime)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
+    lockManager.executeSequentially(k.toString) {
+      rawDelete(ir.toStorageId(k, Some(version)), version)
+    }
+  }
+
+  protected def rawDeleteAll(k: K): Future[Done]
+
+  final override def deleteAll[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
```
jdef: why are some of these overrides final and others not?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
```diff
@@ -0,0 +1,261 @@
...
+import scala.async.Async.{ async, await }
+import scala.collection.immutable.Seq
+import scala.concurrent.duration.Duration
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.control.NonFatal
+import scala.util.{ Failure, Success }
+
+case class ZkId(category: String, id: String, version: Option[OffsetDateTime]) {
+  private val bucket = math.abs(id.hashCode % ZkId.HashBucketSize)
+  def path: String = version.fold(f"/$category/$bucket%x/$id") { v =>
```
jdef: why is `path` a `def` instead of a `val`? Couldn't this be declared as:
```
val path: String = {
  val bucket = math.abs(...)
  version.fold(...) { ... }
}
```

PR #4179, opened 2016-08-05
Fixes #4028 health check need to be scheduled immediately after deployed

### LINKED ISSUE #4028: Marathon remains "Deploying" status when container has already deployed
When I deploy nginx with following JSON definition:

```
{
    "id": "nginx",
    "labels": {
        "HAPROXY_GROUP": "external"
    },
    "cpus": 0.2,
    "mem": 20.0,
    "instances": 2,
    "healthChecks": [{
        "path": "/"
    }],
    "container": {
        "type": "DOCKER",
        "docker": {
            "image": "nginx",
            "network": "BRIDGE",
            "portMappings": [{ "containerPort": 80, "hostPort": 0, "servicePort": 80, "protocol": "tcp" }]
        }
    }
}
```

2 instance of nginx containers start immediately because I have pulled nginx image. I can find that nginx containers are running and I can visit nginx through browser. However, the marathon UI shows that the nginx application is still "Deploying". After about **60 seconds**, its status change to "RUNNING".

I found this is related to **healthChecks**, more specifically, the  **intervalSeconds** of **healthChecks**:
- intervalSeconds (Optional. Default: 60): Number of seconds to wait between health checks

So I set **intervalSeconds** to 5:

```
{
    "id": "nginx",
    "labels": {
        "HAPROXY_GROUP": "external"
    },
    "cpus": 0.2,
    "mem": 20.0,
    "instances": 2,
    "healthChecks": [{
        "path": "/",
        "intervalSeconds": 5
    }],
    "container": {
        "type": "DOCKER",
        "docker": {
            "image": "nginx",
            "network": "BRIDGE",
            "portMappings": [{ "containerPort": 80, "hostPort": 0, "servicePort": 80, "protocol": "tcp" }]
        }
    }
}
```

Then I can find that the status of nginx application change to "RUNNING" in about 5 seconds.

However, this is not a good solution. Because when the application is running, we don't need to check the healthy in every 5 seconds. We only need to check the healthy with high frequency when we first start the application.

SiyuanHe:
tiny fix to make health check working properly

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
@@ -130,6 +130,8 @@ object HealthCheck {
   val DefaultIgnoreHttp1xx = false
   val DefaultPort = None
 
+  val FirstHealthCheckTime = 1.seconds
+
   implicit val healthCheck = validator[HealthCheck] { hc =>
     (hc.portIndex.nonEmpty is true) or (hc.port.nonEmpty is true)
     hc is validProtocol
--- a/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
@@ -76,9 +76,9 @@ private[health] class HealthCheckActor(
         healthCheck
       )
       nextScheduledCheck = Some(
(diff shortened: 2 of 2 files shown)


zen-dog:
The problem with provided fix is that scheduler.schedule(...) needs to be canceled (gives you a Cancellable() that you need to store). It would be better to just trigger the first health check immediately and keep using scheduleNextHealthCheck(...).

aquamatthias:
Close in favor of #4368

PR #4184, opened 2016-08-05
Fix portDefinition <-> Health Check Validation

jasongilanfarr:
- Critical line: ContainerSerializer.scala line 143.
  If there are no container ports, this _was_ None
  before we serialized, after serializing and deserializing
  the None becomes the empty sequence which is not treated the same
- Change a bunch of "defs" to "vals". There is absolutely no reason
  these need to be computed twice.
- Fix AppDefinition serialization to properly serialize portDefinitions
  if there are no service ports (this was a red-herring suspected
  of being the core bug)

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -994,8 +994,14 @@ trait AppAndGroupFormats {
       if (runSpec.ipAddress.isEmpty) {
         appJson = appJson ++ Json.obj(
           "ports" -> runSpec.servicePorts,
-          "portDefinitions" -> runSpec.portDefinitions.zip(runSpec.servicePorts).map {
-            case (portDefinition, servicePort) => portDefinition.copy(port = servicePort)
+          "portDefinitions" -> {
+            if (runSpec.servicePorts.nonEmpty) {
+              runSpec.portDefinitions.zip(runSpec.servicePorts).map {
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -994,8 +994,14 @@ trait AppAndGroupFormats {
       if (runSpec.ipAddress.isEmpty) {
         appJson = appJson ++ Json.obj(
           "ports" -> runSpec.servicePorts,
-          "portDefinitions" -> runSpec.portDefinitions.zip(runSpec.servicePorts).map {
-            case (portDefinition, servicePort) => portDefinition.copy(port = servicePort)
+          "portDefinitions" -> {
+            if (runSpec.servicePorts.nonEmpty) {
```
jdef: was this change actually needed? `servicePorts` defaults to `portNumbers` if they're not defined by the container. and `portNunbers` comes from `portDefinitions`...

jdef:
should the `val` really be `lazy val`? sorry if this is a noob question, but they were `def` before which means they weren't computed at all unless they were called.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
jasongilanfarr: It is. It was the red herring where portDefs weren't showing up after creating the app

jasongilanfarr:
I believe these are cheap enough that they should be just vals, lazy Val is a bit more expensive, in most cases more than computing the mapping.

PR #4193, opened 2016-08-08
Add killCount and noOp to delete tasks endpoint

ammaraskar:
# Motivation

At Yelp, for [PaaSTA](https://github.com/Yelp/paasta), we would like to be able to predict Marathon's actions during a scale down event. This gives us a chance to decide if we're really sure about killing an instance/scaling down. This noOp flag allows us to get a preview of Marathon's plan during these events.

Ideally it'd be possible to hit up `PUT /v2/apps/{app_id}` with some sort of noOp flag to probe Marathon's deploy plans for any arbitrary change but that's a far more complicated code diff, so I opted for this simpler approach.
## Implementation

This is my first time dabbing into Scala, so I apologize if I'm doing anything horribly wrong. All feedback is much appreciated. I checked the project with scalastyle and as far as I can tell I didn't any new style violations.

I added two basic tests following the pattern for the rest of AppTaskResourceTest.scala file and some tests had to be changed a little because of the type change to be just a generic `Iterable[Task]` instead of some specific collection. I'll add some github comments to mark some of the more interesting testing changes.

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -339,7 +339,7 @@ post:
       is: [ secured, deployable ]
       queryParameters:
         host:
-          description: all tasks of that application on the supplied slave are killed
+          description: all tasks of that application on the supplied slave are killed.
         scale:
           type: boolean
           description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks. Only possible if `wipe=false` or not specified.
@@ -347,7 +347,12 @@ post:
         wipe:
           type: boolean
           description: If `wipe=true` is specified and the app uses local persistent volumes, associated dynamic reservations will be unreserved, and persistent volumes will be destroyed. Only possible if `scale=false` or not specified.
-          default: false
+        killCount:
+          type: integer
+          description: The number of tasks to kill. A negative value signifies killing all running tasks.
+        noOp:
+          type: boolean
+          description: If `noOp=true` is specified then the app's tasks aren't actually killed, instead the endpoint responds with a list of tasks that would have been killed.
       responses:
         200:
           description: If scale=false, all tasks that were killed are returned.
--- a/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
+++ b/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
@@ -8,8 +8,9 @@ import mesosphere.marathon.core.task.{ Task, TaskStateOp }
 import mesosphere.marathon.core.task.tracker.{ TaskStateOpProcessor, TaskTracker }
 import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer, Identity, UpdateRunSpec }
 import mesosphere.marathon.state._
-import mesosphere.marathon.upgrade.DeploymentPlan
+import mesosphere.marathon.upgrade.{ DeploymentPlan, ScalingProposition }
 import mesosphere.marathon.{ MarathonConf, MarathonSchedulerService, UnknownAppException }
+import mesosphere.mesos.Constraints
 import org.slf4j.LoggerFactory
 
 import scala.concurrent.{ ExecutionContext, Future }
@@ -28,7 +29,8 @@ class TaskKiller @Inject() (
   def kill(
     appId: PathId,
     findToKill: (Iterable[Task] => Iterable[Task]),
-    wipe: Boolean = false)(implicit identity: Identity): Future[Iterable[Task]] = {
+    wipe: Boolean = false,
+    killCount: Integer = -1)(implicit identity: Identity): Future[Iterable[Task]] = {
 
     result(groupManager.app(appId)) match {
       case Some(app) =>
@@ -36,7 +38,7 @@ class TaskKiller @Inject() (
 
         import scala.concurrent.ExecutionContext.Implicits.global
         taskTracker.appTasks(appId).flatMap { allTasks =>
-          val foundTasks = findToKill(allTasks)
+          val foundTasks = getTasksToKill(appId, app, findToKill(allTasks), killCount)
           val expungeTasks = if (wipe) expunge(foundTasks) else Future.successful(())
 
           expungeTasks.map { _ =>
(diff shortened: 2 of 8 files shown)


src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -126,7 +136,7 @@ class AppTasksResource @Inject() (
       val deploymentF = taskKiller.killAndScale(pathId, findToKill, force)
       deploymentResult(result(deploymentF))
     } else {
-      reqToResponse(taskKiller.kill(pathId, findToKill, wipe)) {
+      reqToResponse(taskKiller.kill(pathId, findToKill, wipe, -1)) {
```
ammaraskar: For some reason, not explicitly passing the -1 argument, even though its a default causes the mockito `verifyNoMoreInteractions(taskKiller)` check to fail. I didn't look into it too much but I believe it has to do with the way scala implements default arguments internally.

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
```diff
@@ -128,7 +128,7 @@ class TasksResource @Inject() (
       affectedApps.foreach(checkAuthorization(UpdateRunSpec, _))
 
       val killed = result(Future.sequence(toKill.map {
-        case (appId, tasks) => taskKiller.kill(appId, _ => tasks, wipe)
+        case (appId, tasks) => taskKiller.kill(appId, _ => tasks, wipe, -1)
```
ammaraskar: See: https://github.com/mesosphere/marathon/pull/4193/files#r73925367

src/test/scala/mesosphere/marathon/api/TaskKillerTest.scala
```diff
@@ -84,6 +86,48 @@ class TaskKillerTest extends MarathonSpec
...
+      toKillCaptor.capture()
+    )).thenReturn(Future.successful(expectedDeploymentPlan))
+
+    val result = f.taskKiller.killAndScale(appId, (tasks) => tasksToKill, force = true, killCount = 1)
+    result.futureValue shouldEqual expectedDeploymentPlan
+    toKillCaptor.getValue.size shouldEqual 1
+    tasksToKill contains toKillCaptor.getValue.values.take(1)
+  }
+
+  class IterableEquals[A](var original: Iterable[A]) extends ArgumentMatcher[Iterable[A]] {
```
ammaraskar: This is necessary since the TaskKiller test internally passes around generic `Iterable[Task]` objects. So it's no longer possible to just use equality operators.

ammaraskar:
Can't recreate the failing travis locally, not sure what's up with that.

jasongilanfarr:
Thanks for the PR- I like the idea a lot and It'd be really helpful. My day was taken over fixing two major 1.3 regressions, I'll try to take a look as soon as I can.

jeschkies:
@ammaraskar Could you update the branch?

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -339,15 +339,20 @@ post:
...
           description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks. Only possible if `wipe=false` or not specified.
           default: false
         wipe:
           type: boolean
           description: If `wipe=true` is specified and the app uses local persistent volumes, associated dynamic reservations will be unreserved, and persistent volumes will be destroyed. Only possible if `scale=false` or not specified.
-          default: false
+        killCount:
+          type: integer
+          description: The number of tasks to kill. A negative value signifies killing all running tasks.
+        noOp:
```
aquamatthias: Marathon groups API calls this parameter `dryRun`. Please use the same name here as well.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -339,15 +339,20 @@ post:
...
           type: boolean
           description: If `scale=true` is specified, then the application is scaled down by the number of killed tasks. Only possible if `wipe=false` or not specified.
           default: false
         wipe:
           type: boolean
           description: If `wipe=true` is specified and the app uses local persistent volumes, associated dynamic reservations will be unreserved, and persistent volumes will be destroyed. Only possible if `scale=false` or not specified.
-          default: false
+        killCount:
+          type: integer
+          description: The number of tasks to kill. A negative value signifies killing all running tasks.
```
aquamatthias: Negative number is not intuitive. Let's either define a (positive) value or no value.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -28,15 +29,16 @@ class TaskKiller @Inject() (
   def kill(
     appId: PathId,
     findToKill: (Iterable[Task] => Iterable[Task]),
-    wipe: Boolean = false)(implicit identity: Identity): Future[Iterable[Task]] = {
+    wipe: Boolean = false,
+    killCount: Integer = -1)(implicit identity: Identity): Future[Iterable[Task]] = {
```
aquamatthias: The TaskKiller does not need to know anything about `killCount`.
`findToKill` is a function that chooses the tasks to kill and can be limited.
Please revert all changes to TaskKiller.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -97,11 +99,19 @@ class AppTasksResource @Inject() (
 
     if (scale && wipe) throw new BadRequestException("You cannot use scale and wipe at the same time.")
 
-    if (scale) {
-      val deploymentF = taskKiller.killAndScale(pathId, findToKill, force)
+    if (noOp) {
```
aquamatthias: This will result in wrong result for `scale=true`

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -97,11 +99,19 @@ class AppTasksResource @Inject() (
...
+      result(groupManager.app(pathId)) match {
+        case None => unknownApp(pathId)
+        case Some(appInfo) =>
+          var tasks = findToKill(taskTracker.appTasksLaunchedSync(pathId))
+          tasks = taskKiller.getTasksToKill(pathId, appInfo, tasks, killCount)
```
aquamatthias: `findToKill` and `getTasksToKill` should be one function, which is used in dry run and kill mode.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
aquamatthias: See my earlier comment: I would opt to remove this argument from TaskKiller.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -83,9 +83,11 @@ class AppTasksResource @Inject() (
   def deleteMany(
     @PathParam("appId") appId: String,
     @QueryParam("host") host: String,
+    @QueryParam("killCount")@DefaultValue("-1") killCount: Integer = -1,
```
aquamatthias: Remove default value and handle the optional case.

aquamatthias:
@ammaraskar no activity for the last month. Please incorporate the feedback and reopen the PR.

PR #4195, opened 2016-08-08
Never default servicePorts since they are globally unique!

jasongilanfarr:
In order to allocate a service port, you must have Host Networking
and explicitly specify them.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -993,7 +993,7 @@ trait AppAndGroupFormats {
       // top-level ports fields are incompatible with IP/CT
       if (runSpec.ipAddress.isEmpty) {
         appJson = appJson ++ Json.obj(
-          "ports" -> runSpec.servicePorts,
+          "ports" -> runSpec.portDefinitions.map(_.port),
           "portDefinitions" -> {
             if (runSpec.servicePorts.nonEmpty) {
               runSpec.portDefinitions.zip(runSpec.servicePorts).map {
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
@@ -224,7 +224,9 @@ private[impl] class GroupManagerActor(
       servicePorts: Seq[Int]): Seq[PortDefinition] =
       if (portDefinitions.nonEmpty)
         portDefinitions.zipAll(servicePorts, AppDefinition.RandomPortDefinition, AppDefinition.RandomPortValue).map {
-          case (portDefinition, servicePort) => portDefinition.copy(port = servicePort)
+          case (portDefinition, servicePort) =>
+
+            portDefinition.copy(port = servicePort)
         }
       else Seq.empty
 
@@ -266,14 +268,11 @@ private[impl] class GroupManagerActor(
 
     val dynamicApps: Set[AppDefinition] =
       to.transitiveApps.map {
-        // assign values for service ports that the user has left "blank" (set to zero)
-        case app: AppDefinition if app.hasDynamicServicePorts => assignPorts(app)
         case app: AppDefinition if app.servicePorts.nonEmpty =>
-          // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
-          // TODO: This is almost definitely wrong.
-          app.copy(
-            portDefinitions = mergeServicePortsAndPortDefinitions(app.portDefinitions, app.servicePorts)
-          )
+          synchronized {
+            app.servicePorts.withFilter(_ != AppDefinition.RandomPortValue).foreach { port => taken += port }
+          }
+          assignPorts(app)
         case app: AppDefinition if app.portDefinitions.exists(_.port == AppDefinition.RandomPortValue) =>
           synchronized {
             app.portDefinitions.filter(_.port != AppDefinition.RandomPortValue).foreach { portDef =>
@@ -283,7 +282,7 @@ private[impl] class GroupManagerActor(
           app.copy(
             portDefinitions = app.portDefinitions.map {
               case portDef: PortDefinition if portDef.port == AppDefinition.RandomPortValue =>
-                // we're reserving these ports _away_ from the set of reservable ports.
+                // we're reserving these ports _away_ from the set of reservable service ports.
                 portDef.copy(port = nextGlobalFreePort)
               case portDef: PortDefinition => portDef
             }
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
@@ -248,7 +248,7 @@ private[impl] class GroupManagerActor(
       // defined only if there are port mappings
       val newContainer: Option[Container] = for {
         c <- app.container
-        d <- c.docker if !d.portMappings.isEmpty
+        d <- c.docker() if d.portMappings.isDefined
       } yield {
         val newPortMappings = d.portMappings.get.zip(servicePorts).map {
           case (portMapping, servicePort) =>
@@ -268,11 +268,13 @@ private[impl] class GroupManagerActor(
       to.transitiveApps.map {
         // assign values for service ports that the user has left "blank" (set to zero)
         case app: AppDefinition if app.hasDynamicServicePorts => assignPorts(app)
-        case app: AppDefinition =>
+        case app: AppDefinition if app.servicePorts.nonEmpty =>
           // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
+          // TODO: This is almost definitely wrong.
           app.copy(
             portDefinitions = mergeServicePortsAndPortDefinitions(app.portDefinitions, app.servicePorts)
           )
+        case app: AppDefinition => app
       }
 
     dynamicApps.foldLeft(to) { (group, app) =>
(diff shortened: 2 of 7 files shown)


kensipe:
we should add a test for having an app with a port and asserting it doesn’t have a service port

kensipe:
LGTM!

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -268,11 +268,13 @@ private[impl] class GroupManagerActor(
       to.transitiveApps.map {
         // assign values for service ports that the user has left "blank" (set to zero)
         case app: AppDefinition if app.hasDynamicServicePorts => assignPorts(app)
-        case app: AppDefinition =>
+        case app: AppDefinition if app.servicePorts.nonEmpty =>
           // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
+          // TODO: This is almost definitely wrong.
```
adam-mesos: Please you add your name/github-id to the `TODO(jasongilanfarr): ...` so we can attribute these later, unless you're planning on fixing this before we merge.

adam-mesos:
Looks better than the previous mess, but I can't claim to completely understand what's going on here.
And as Jason's TODO points out, there's still more weirdness left to cleanup.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -993,7 +993,7 @@ trait AppAndGroupFormats {
       // top-level ports fields are incompatible with IP/CT
       if (runSpec.ipAddress.isEmpty) {
         appJson = appJson ++ Json.obj(
-          "ports" -> runSpec.servicePorts,
+          "ports" -> runSpec.portDefinitions.map(_.port),
```
gkleiman: This is a breaking change. Can you add more context to make sure that it makes sense?

If we decide to do this, then we should add a huge warning in our changelog and release notes.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -248,7 +250,7 @@ private[impl] class GroupManagerActor(
       // defined only if there are port mappings
       val newContainer: Option[Container] = for {
         c <- app.container
-        d <- c.docker if !d.portMappings.isEmpty
+        d <- c.docker() if d.portMappings.isDefined
```
gkleiman: why did you add the parenthesis?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -266,13 +268,30 @@ private[impl] class GroupManagerActor(
 
     val dynamicApps: Set[AppDefinition] =
       to.transitiveApps.map {
-        // assign values for service ports that the user has left "blank" (set to zero)
-        case app: AppDefinition if app.hasDynamicServicePorts => assignPorts(app)
-        case app: AppDefinition =>
-          // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
+        case app: AppDefinition if app.servicePorts.nonEmpty =>
+          synchronized {
```
gkleiman: Isn't `GroupManager` an actor now? do we really need `synchronized` here?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -266,13 +268,30 @@ private[impl] class GroupManagerActor(
...
+            app.servicePorts.withFilter(_ != AppDefinition.RandomPortValue).foreach { port => taken += port }
+          }
+          assignPorts(app)
+        case app: AppDefinition if app.portDefinitions.exists(_.port == AppDefinition.RandomPortValue) =>
```
gkleiman: This is another breaking change that contradicts what's described in https://mesosphere.github.io/marathon/docs/ports.html

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -266,13 +268,30 @@ private[impl] class GroupManagerActor(
...
-            portDefinitions = mergeServicePortsAndPortDefinitions(app.portDefinitions, app.servicePorts)
+            portDefinitions = app.portDefinitions.map {
+              case portDef: PortDefinition if portDef.port == AppDefinition.RandomPortValue =>
+                // we're reserving these ports _away_ from the set of reservable service ports.
+                portDef.copy(port = nextGlobalFreePort)
+              case portDef: PortDefinition => portDef
+            }
           )
+        case app: AppDefinition =>
+          synchronized {
```
gkleiman: Same as above.. it's strange to see a `synchronized` block inside an actor.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -231,6 +231,7 @@ object Group {
       group is noAppsAndGroupsWithSameName
       group is conditional[Group](_.id.isRoot)(noCyclicDependencies)
       group.groups is every(valid(validNestedGroup(group.id.canonicalPath(base))))
+      group is servicePortsDontOverlap
```
gkleiman: Doesn't https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/upgrade/DeploymentPlan.scala#L328 take care of this?

kensipe:
lots of changes since my original comments.  I'm concerned with the new code that @gkleiman raised issues with as well.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
jasongilanfarr: Can revert this one... keep in mind that servicePorts are explicitly opt-in.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
jasongilanfarr: container line 15, `def docker()`, e.g. `Empty paren method accessed as paramterless`

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
jasongilanfarr: just followed the pattern in the class. taken was protected by serialized.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
jasongilanfarr: How so? These are "portDefintion" style ports.

The way servicePorts are implemented today and I _still_ think its completely broken (e.g. 1.3) - is that they are _globally unique_ not "servicePort: A service port is a port used to describe the port that a service should made available at. Marathon does not bind to the service ports specified but will ensure that you cannot have multiple applications that use the same service port running on the same host. Service ports are typically only used by external applications (e.g. HAProxy) to make the application available at the specified port. See Service Discovery & Load Balancing for more information."

This is the major regression, so maybe this entire suite of logic should be killed off and the old behavior should be brought back? (e.g. no servicePort uniqueness)

src/main/scala/mesosphere/marathon/state/Group.scala
jasongilanfarr: Then deployment plan's validations should move over.

jasongilanfarr:
@gkleiman I believe you've pointed out that the entire logic in 1.3 that makes servicePorts globally unique is wrong and needs to be reverted. This massively simplifies the correct solution as well as the patch.

src/main/scala/mesosphere/marathon/state/Group.scala
gkleiman: AFAIK @unterstein moved the validation to deployment plan, in order to allow users to make changes to the root group without having to fix all existing applications, see #4167

jasongilanfarr:
Following our internal discussion, making servicePorts globally unique is the actual regression. Instead of working around logic to maintain this, I'm going to only remove the regression itself.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
gkleiman: `portDefinitions` were supposed to be a superset of `ports`. I think that we shouldn't make breaking changes of the interface, without having clearly communicated a deprecation plan.

Until recently, our docs used to say that the user was responsibly for ensuring the uniqueness of the service ports, and there was no validation.

Then the docs were updated to say that Marathon enforces the uniqueness, and a broken validation was added.

I recently fixed the validation. But I'd be ok with going back to not enforcing the uniqueness, as long as that is properly documented and maybe even checked with the community.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
gkleiman: Until every port is assigned a `servicePort`, they're not explicitly opt-in, and making them so would be a breaking change.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
jasongilanfarr: As stated, I'm reverting the change as is and reverting service port global
uniqueness.

There is absolutely no reason to have globally unique ports: even poor mans
load balancing has to consider which hosts have a given app, so if another
app is using the same port, they simply can't be collocated. I have no
intention of breaking compatibility, but this "global service port" change
is absolutely a regression and compatibility breaking. It's also why the
bug you pointed out with unrelated deployments being blocked on each other
occurs.

Issue #4165, opened 2016-08-01
Deployment locked by another completely independent deployment


gkleiman:
1) `PUT /v2/groups` with:

```json
{
  "id": "/foo",
  "apps": [
    {
      "id": "initial",
      "cpu": 999,
      "container": {
        "type": "DOCKER",
        "docker": {
          "image": "foo",
          "network": "BRIDGE",
          "portMappings": []
        }
      }
    }
  ]
}
```

2)  `PUT /v2/groups` with:

```json
{
  "id": "/bar",
  "apps": [
    {
      "id": "/bar/initial",
      "cmd": "sleep 100"
    }
  ]
}
```

This app definition is completely unrelated to the first one, so it should be possible to create a deployment. However Marathon will return an error like the following one:

```json
 {"message":"App is locked by one or more deployments. Override with the option '?force=true'. View details at '/v2/deployments/<DEPLOYMENT_ID>'."
```

Note that this doesn't happen if the first `PUT` doesn't include an empty `portMappings` array.

jasongilanfarr:
This does not happen on master and both apps are deployed correctly. I'm checking 1.2 and 1.3 now.

jasongilanfarr:
The real reason for this is that the portDefinitions is initially unassigned then reassigned on the next deployment.

PR #4199, opened 2016-08-09
Don't enforce service port uniqueness (it can't be done)

jasongilanfarr:
Also, fix an oversight about Option[Seq[]] in Container (#4165)

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -274,11 +274,12 @@ case class AppDefinition(
     )
   }
 
-  private val portIndicies: Range = container.flatMap(_.hostPorts).getOrElse(portNumbers).indices
+  private val portIndices: Range = container.flatMap(_.hostPorts.filter(_.nonEmpty)).getOrElse(portNumbers).indices
 
-  val hostPorts: Seq[Option[Int]] = container.flatMap(_.hostPorts).getOrElse(portNumbers.map(Some(_)))
+  val hostPorts: Seq[Option[Int]] =
+    container.flatMap(_.hostPorts.filter(_.nonEmpty)).getOrElse(portNumbers.map(Some(_)))
 
-  val servicePorts: Seq[Int] = container.flatMap(_.servicePorts).getOrElse(portNumbers)
+  val servicePorts: Seq[Int] = container.flatMap(_.servicePorts.filter(_.nonEmpty)).getOrElse(portNumbers)
 
   val hasDynamicServicePorts: Boolean = servicePorts.contains(AppDefinition.RandomPortValue)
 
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -274,11 +274,12 @@ case class AppDefinition(
     )
   }
 
-  private val portIndicies: Range = container.flatMap(_.hostPorts).getOrElse(portNumbers).indices
+  private val portIndicies: Range = container.flatMap(_.hostPorts.filter(_.nonEmpty)).getOrElse(portNumbers).indices
```
gkleiman: nit: it should be `portIndices` =)

jasongilanfarr: Doh!

gkleiman:
LGTM, but please update the documentation too!

unterstein:
LGTM, lets wait for the tests and get this merged :)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -274,11 +274,12 @@ case class AppDefinition(
     )
   }
 
-  private val portIndicies: Range = container.flatMap(_.hostPorts).getOrElse(portNumbers).indices
+  private val portIndices: Range = container.flatMap(_.hostPorts.filter(_.nonEmpty)).getOrElse(portNumbers).indices
```
jdef: doesn't filtering here screw up the indices?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -274,11 +274,12 @@ case class AppDefinition(
...
 
-  val hostPorts: Seq[Option[Int]] = container.flatMap(_.hostPorts).getOrElse(portNumbers.map(Some(_)))
+  val hostPorts: Seq[Option[Int]] =
+    container.flatMap(_.hostPorts.filter(_.nonEmpty)).getOrElse(portNumbers.map(Some(_)))
 
-  val servicePorts: Seq[Int] = container.flatMap(_.servicePorts).getOrElse(portNumbers)
+  val servicePorts: Seq[Int] = container.flatMap(_.servicePorts.filter(_.nonEmpty)).getOrElse(portNumbers)
```
jdef: ditto

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
jasongilanfarr: if it was correctly defined as a Seq[], yes, but its not, its an Option[Seq[]] - so we're saying "use the container's host ports if they are defined and non-empty" - we can't serialize Option[Seq[]] to proto, so there's a major inconsistency here.

jdef: sounds like we need some proto changes if there's a serialization issue

jasongilanfarr: No, it was a major regression when the original change was introduced.

jasongilanfarr: Proto has no notion of an "optional repeated field" its also very strange to have an Option[Seq[]] of anything. Option is by definition a Seq of 0 or 1 element. I tried to keep this and the other regression change as minimal as I could - keeping the Option[Seq[]] in place - but fixing the serialization/deserialization (there were a lot of regressions). I have a branch at home where I started to remove the wrapping Option[] but it didn't help at all.

PR #4203, opened 2016-08-10
fixed wrong gc calculation

unterstein:
Before the task state refactoring the class ```ExpungeOverdueLostTasksActor``` had following logic to filter unreachable tasks which received 24 hours no updates:

```
  def filterLostGCTasks(tasks: Map[PathId, AppTasks]): Iterable[Task] = {
    def isTimedOut(task: Task): Boolean = {
      task.mesosStatus.fold(false) { status =>
        val age = clock.now().toDateTime.minus(status.getTimestamp.toLong * 1000).getMillis.millis
        age > config.taskLostExpungeGC
      }
    }
    tasks.values.flatMap(_.tasks.filter {
      case TemporarilyUnreachable(task) if isTimedOut(task) => true
      case _ => false
    })
  }

```

During the refactoring the filter was changed to 

```tasks.values.flatMap(_.tasks.filter(task => isTimedOut(task.mesosStatus)))```

which misses the fact that it should filter the unreachable tasks, which I re-introduced with this PR by adding ```task => task.isUnreachable && isTimedOut(task.mesosStatus)```

### DIFF
--- a/src/test/scala/mesosphere/marathon/test/Mockito.scala
+++ b/src/test/scala/mesosphere/marathon/test/Mockito.scala
@@ -20,6 +20,7 @@ trait Mockito extends MockitoSugar {
   def times(num: Int) = M.times(num)
   def timeout(millis: Int) = M.timeout(millis.toLong)
   def atLeastOnce = M.atLeastOnce()
+  def once = M.times(1)
   def atLeast(num: Int) = M.atLeast(num)
   def atMost(num: Int) = M.atMost(num)
   def never = M.never()
--- a/src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala
@@ -55,7 +55,7 @@ class ExpungeOverdueLostTasksActor(
         age > config.taskLostExpungeGC
       }
     }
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/test/scala/mesosphere/marathon/test/Mockito.scala [hunk 1]
```diff
@@ -20,6 +20,7 @@ trait Mockito extends MockitoSugar {
   def times(num: Int) = M.times(num)
   def timeout(millis: Int) = M.timeout(millis.toLong)
   def atLeastOnce = M.atLeastOnce()
+  def once = M.times(1)
```
gkleiman: `times(1)` is the default, so this is redundant, but ok if you want to be very explicit =)

unterstein: Ah ok, did`t know that. Should I change this?

gkleiman:
LGTM, I guess that we need to cherry-pick this, so a changelog entry would be nice =)

kensipe:
LGTM after the style updates.

src/test/scala/mesosphere/marathon/test/Mockito.scala
jasongilanfarr: I actually like it personally.

`verify(mock).abc() `vs `verify(mock, once).abc()` is much more obvious.

gkleiman:
Thanks for fixing this nasty bug!

PR #4208, opened 2016-08-11
test to verify the default user on marathon is root

kensipe:
system integration test for confirming the default user is root on marathon for dcos-9039.

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/setup/SingleMarathonIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/setup/SingleMarathonIntegrationTest.scala
@@ -9,8 +9,8 @@ import mesosphere.marathon.state.{ AppDefinition, Container, DockerVolume, PathI
 import org.apache.commons.io.FileUtils
 import org.apache.mesos.Protos
 import org.apache.zookeeper.ZooDefs.Perms
-import org.apache.zookeeper.data.{ ACL, Id }
 import org.apache.zookeeper._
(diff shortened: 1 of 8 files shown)


sschneid:
👍 

I wonder if functionality like `client.add_app()` and `client.remove_app()` should exist as methods in Shakedown, but that isn't a blocker for this PR obviously.  :)

adam-mesos:
So this only checks Marathon solo, not Marathon on DC/OS? Marathon has a `--mesos_user` flag for setting the default user, and it defaults to whatever user runs the Marathon scheduler.
But DC/OS has its own configuration for `--mesos_user` for Marathon/Metronome.. shouldn't we be testing in DC/OS instead?

kensipe:
@adam-mesos while the flags you mentioned are new to me, the integration tests here are shakedown and run against a running dcos env

adam-mesos:
Cool. Sorry, I guess I need to learn more about shakedown. How does it setup its config.yaml? Default? Does it test against open or enterprise or both?
This runs in the Marathon CI, every time a Marathon PR is submitted? But what about Metronome PRs or DCOS (or dcos-enterprise) PRs? Those could break this just as easily, and Marathon PRs wouldn't know until it's merged into dcos and tested here.

kensipe:
@adam-mesos details:  shakedown simply works against a created cluster.  it is up to the CI job to setup that cluster.   currently we have a CI job that nightly runs all marathon shakedown tests against dcos-enterprise.   I will be setting up more by next week.

PR #4210, opened 2016-08-11
Fixes #4168 - Add support for Mesos HTTP/TCP health checks

### LINKED ISSUE #4168: Support Mesos HTTP/TCP health checks
The Mesos team is working hard on adding support for HTTP/TCP health checks. Their solution will initially reach parity with the currently supported `COMMAND` health checks: [MESOS-5961](https://issues.apache.org/jira/browse/MESOS-5961), [MESOS-3567](https://issues.apache.org/jira/browse/MESOS-3567), [MESOS-2533]( https://issues.apache.org/jira/browse/MESOS-2533).

In order to scale better, Marathon should support these new kinds of health checks as soon as possible.

The idea is to update the app definition, introducing these new health check protocols: `MESOS_HTTP`, `MESOS_HTTPS`, `MESOS_TCP`. They'll behave exactly like the currently supported `COMMAND` health check.

The following needs to be done in order to achieve this:

- [ ] Refactor `HealthCheck`, adding `HttpConf`,`TcpConf`, `MesosHttpConf`, etc. and moving the settings from the main class to these per-protocol config classes.
- [ ] Add JSON formats and validations. Mesos will initially support at-most one health check.
- [ ] Update TaskBuilder logic to account for the new health check types. Marathon will have to implement logic to put the right port number in the health checks proto, taking in consideration the networking specified in the app definition.
- [ ] Integration tests.

This is also related to #3957 (Marathon will currently forget the app health information if an app is using Mesos health checks).

gkleiman:
This change is [Reviewable](https://reviewable.io/reviews/mesosphere/marathon/4210)

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -1,11 +1,6 @@
 <scalastyle commentFilter="enabled">
  <name>Mesosphere Marathon configuration</name>
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"/>
- <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
-  <parameters>
-   <parameter name="maxFileLength"><![CDATA[800]]></parameter>
-  </parameters>
- </check>
  <check level="error" class="org.scalastyle.file.HeaderMatchesChecker" enabled="false">
   <parameters>
    <parameter name="header"><![CDATA[]]></parameter>
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -396,13 +396,11 @@ trait ContainerFormats {
         "appc" -> appcValues(appc)
       )
     }
-    Writes { container =>
-      container match {
-        case m: Container.Mesos => MesosContainerWrites.writes(m)
-        case d: Container.Docker => DockerContainerWrites.writes(d)
-        case c: Container.MesosDocker => MesosDockerContainerWrites.writes(c)
-        case c: Container.MesosAppC => AppCContainerWrites.writes(c)
-      }
+    Writes {
+      case m: Container.Mesos => MesosContainerWrites.writes(m)
+      case d: Container.Docker => DockerContainerWrites.writes(d)
+      case c: Container.MesosDocker => MesosDockerContainerWrites.writes(c)
+      case c: Container.MesosAppC => AppCContainerWrites.writes(c)
     }
   }
 }
@@ -733,7 +731,7 @@ trait AppAndGroupFormats {
         json.asOpt[Seq[String]] match {
           case Some(seq) if seq.size >= 2 && seq.size <= 3 =>
             if (validOperators.contains(seq(1))) {
-              val builder = Constraint.newBuilder().setField(seq(0)).setOperator(Operator.valueOf(seq(1)))
+              val builder = Constraint.newBuilder().setField(seq.head).setOperator(Operator.valueOf(seq(1)))
               if (seq.size == 3) builder.setValue(seq(2))
               JsSuccess(builder.build())
             } else {
@@ -934,12 +932,11 @@ trait AppAndGroupFormats {
 
       Option(TaskLostBehavior.valueOf(behaviorString)) match {
         case Some(taskLostBehavior) => JsSuccess(taskLostBehavior)
-        case None => {
+        case None =>
           val allowedTaskLostBehaviorString =
             TaskLostBehavior.values().toSeq.map(_.getDescriptorForType.getName).mkString(", ")
 
           JsError(s"'$behaviorString' is not a valid taskLostBehavior. Allowed values: $allowedTaskLostBehaviorString")
-        }
       }
 
     }
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -10,7 +10,7 @@ import mesosphere.marathon.core.plugin.{ PluginDefinition, PluginDefinitions }
 import mesosphere.marathon.core.readiness.ReadinessCheck
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.event._
-import mesosphere.marathon.core.health.{ Health, HealthCheck }
+import mesosphere.marathon.core.health._
 import mesosphere.marathon.state._
 import mesosphere.marathon.upgrade.DeploymentManager.DeploymentStepInfo
 import mesosphere.marathon.upgrade._
@@ -628,22 +628,61 @@ trait HealthCheckFormats {
   implicit lazy val HealthCheckProtocolFormat: Format[Protocol] =
     enumFormat(Protocol.valueOf, str => s"$str is not a valid protocol")
 
-  implicit lazy val HealthCheckFormat: Format[HealthCheck] = {
+  val BasicHealthCheckFormatBuilder = {
     import mesosphere.marathon.core.health.HealthCheck._
 
-    (
-      (__ \ "path").formatNullable[String] ~
-      (__ \ "protocol").formatNullable[Protocol].withDefault(DefaultProtocol) ~
-      (__ \ "portIndex").formatNullable[Int] ~
-      (__ \ "command").formatNullable[Command] ~
-      (__ \ "gracePeriodSeconds").formatNullable[Long].withDefault(DefaultGracePeriod.toSeconds).asSeconds ~
+    (__ \ "gracePeriodSeconds").formatNullable[Long].withDefault(DefaultGracePeriod.toSeconds).asSeconds ~
       (__ \ "intervalSeconds").formatNullable[Long].withDefault(DefaultInterval.toSeconds).asSeconds ~
       (__ \ "timeoutSeconds").formatNullable[Long].withDefault(DefaultTimeout.toSeconds).asSeconds ~
-      (__ \ "maxConsecutiveFailures").formatNullable[Int].withDefault(DefaultMaxConsecutiveFailures) ~
-      (__ \ "ignoreHttp1xx").formatNullable[Boolean].withDefault(DefaultIgnoreHttp1xx) ~
-      (__ \ "port").formatNullable[Int]
-    )(HealthCheck.apply, unlift(HealthCheck.unapply))
+      (__ \ "maxConsecutiveFailures").formatNullable[Int].withDefault(DefaultMaxConsecutiveFailures)
   }
+
+  val HealthCheckWithPortsFormatBuilder =
+    BasicHealthCheckFormatBuilder ~
+      (__ \ "portIndex").formatNullable[Int] ~
+      (__ \ "port").formatNullable[Int]
+
+  implicit val HttpHealthCheckFormat: Format[HttpHealthCheck] = {
+    import mesosphere.marathon.core.health.HttpHealthCheck._
+    (
+      HealthCheckWithPortsFormatBuilder ~
+      (__ \ "path").formatNullable[String] ~
+      (__ \ "ignoreHttp1xx").formatNullable[Boolean].withDefault(DefaultIgnoreHttp1xx) ~
+      (__ \ "protocol").formatNullable[Protocol].withDefault(Protocol.HTTP)
+    )(HttpHealthCheck.apply, unlift(HttpHealthCheck.unapply))
+  }
+
+  implicit val CommandHealthCheckFormat: Format[CommandHealthCheck] = (
+    BasicHealthCheckFormatBuilder ~
+    (__ \ "command").format[Command]
+  )(CommandHealthCheck.apply, unlift(CommandHealthCheck.unapply))
+
+  implicit val TcpHealthCheckFormat: Format[TcpHealthCheck] =
+    HealthCheckWithPortsFormatBuilder(TcpHealthCheck.apply, unlift(TcpHealthCheck.unapply))
+
+  implicit val HealthCheckFormat: Format[HealthCheck] = Format[HealthCheck] (
+    new Reads[HealthCheck] {
+      override def reads(json: JsValue): JsResult[HealthCheck] = {
+        val protocol = (json \ "protocol").validateOpt[Protocol](HealthCheckProtocolFormat)
+
+        protocol.flatMap {
+          case None => json.validate[HttpHealthCheck]
+          case Some(Protocol.COMMAND) => json.validate[CommandHealthCheck]
+          case Some(Protocol.HTTP | Protocol.HTTPS) => json.validate[HttpHealthCheck]
+          case Some(Protocol.TCP) => json.validate[TcpHealthCheck]
+          case _ => JsError("Invalid health check protocol.")
+        }
+      }
+    },
+    Writes[HealthCheck] {
+      case tcp: TcpHealthCheck =>
+        Json.toJson(tcp)(TcpHealthCheckFormat).as[JsObject] ++ Json.obj("protocol" -> "TCP")
+      case http: HttpHealthCheck =>
+        Json.toJson(http)(HttpHealthCheckFormat).as[JsObject] ++ Json.obj("protocol" -> Json.toJson(http.protocol))
+      case command: CommandHealthCheck =>
+        Json.toJson(command)(CommandHealthCheckFormat).as[JsObject] ++ Json.obj("protocol" -> "COMMAND")
+    }
+  )
 }
 
 trait ReadinessCheckFormats {
@@ -901,26 +940,29 @@ trait AppAndGroupFormats {
     * In the past, healthCheck.portIndex was required and had a default value 0. When we introduced healthCheck.port, we
     * made it optional (also with ip-per-container in mind) and we have to re-add it in cases where it makes sense.
     */
+  private[this] def addHealthCheckPortIndexIfNecessary(healthChecks: Set[_ <: HealthCheck]): Set[_ <: HealthCheck] =
+    healthChecks.map {
+      case healthCheck: TcpHealthCheck =>
+        def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty
+        if (needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
+        else healthCheck
+      case healthCheck: HttpHealthCheck =>
+        def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty
+        if (needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
+        else healthCheck
+      case healthCheck: HealthCheck => healthCheck
+    }
+
   private[this] def addHealthCheckPortIndexIfNecessary(app: AppDefinition): AppDefinition = {
     val hasPortMappings = app.container.exists(_.portMappings.nonEmpty)
     val portIndexesMakeSense = app.portDefinitions.nonEmpty || hasPortMappings
-    app.copy(healthChecks = app.healthChecks.map { healthCheck =>
-      def needsDefaultPortIndex =
-        healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty && healthCheck.protocol != Protocol.COMMAND
-      if (portIndexesMakeSense && needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
-      else healthCheck
-    })
+
+    if (portIndexesMakeSense) app.copy(healthChecks = addHealthCheckPortIndexIfNecessary(app.healthChecks))
+    else app
   }
 
   private[this] def addHealthCheckPortIndexIfNecessary(appUpdate: AppUpdate): AppUpdate = {
-    appUpdate.copy(healthChecks = appUpdate.healthChecks.map { healthChecks =>
-      healthChecks.map { healthCheck =>
-        def needsDefaultPortIndex =
-          healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty && healthCheck.protocol != Protocol.COMMAND
-        if (needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
-        else healthCheck
-      }
-    })
+    appUpdate.copy(healthChecks = appUpdate.healthChecks.map(addHealthCheckPortIndexIfNecessary))
   }
 
   implicit lazy val taskLostBehaviorWrites = Writes[TaskLostBehavior] { taskLostBehavior =>
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -6,11 +6,11 @@ import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
 import mesosphere.marathon.Protos.ResidencyDefinition.TaskLostBehavior
 import mesosphere.marathon.SerializationFailedException
 import mesosphere.marathon.core.appinfo._
+import mesosphere.marathon.core.event._
+import mesosphere.marathon.core.health._
 import mesosphere.marathon.core.plugin.{ PluginDefinition, PluginDefinitions }
 import mesosphere.marathon.core.readiness.ReadinessCheck
 import mesosphere.marathon.core.task.Task
-import mesosphere.marathon.core.event._
-import mesosphere.marathon.core.health._
 import mesosphere.marathon.state._
 import mesosphere.marathon.upgrade.DeploymentManager.DeploymentStepInfo
 import mesosphere.marathon.upgrade._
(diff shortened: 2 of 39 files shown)


scalastyle-config.xml
```diff
@@ -3,7 +3,7 @@
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"/>
  <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxFileLength"><![CDATA[800]]></parameter>
+   <parameter name="maxFileLength"><![CDATA[900]]></parameter>
```
jasongilanfarr: maybe we should just remove this?

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -10,146 +9,210 @@ import org.apache.mesos.{ Protos => MesosProtos }
...
+  protected def protoBuilder: Protos.HealthCheckDefinition.Builder =
+    Protos.HealthCheckDefinition.newBuilder
+      .setGracePeriodSeconds(this.gracePeriod.toSeconds.toInt)
+      .setIntervalSeconds(this.interval.toSeconds.toInt)
+      .setTimeoutSeconds(this.timeout.toSeconds.toInt)
+      .setMaxConsecutiveFailures(this.maxConsecutiveFailures)
+}
 
-  portIndex: Option[Int] = HealthCheck.DefaultPortIndex,
+trait HealthCheckWithPort {
```
jasongilanfarr: Seems weird to me to have "WithPort" but the port is optional!? Also, should this have a self type of HealthCheck?

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -10,146 +9,210 @@ import org.apache.mesos.{ Protos => MesosProtos }
...
+    }
 
-  ignoreHttp1xx: Boolean = HealthCheck.DefaultIgnoreHttp1xx,
+    port.orElse(portViaIndex)
+  }
+}
 
-  port: Option[Int] = HealthCheck.DefaultPort)
-    extends MarathonState[Protos.HealthCheckDefinition, HealthCheck] {
+trait MesosHealthCheck {
```
jasongilanfarr: no self type?

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -67,33 +66,36 @@ private[health] class HealthCheckActor(
...
+        hc
       )
       nextScheduledCheck = Some(
-        context.system.scheduler.scheduleOnce(healthCheck.interval) {
+        context.system.scheduler.scheduleOnce(hc.interval) {
           self ! Tick
         }
       )
-    }
+    case _ => () // Don't do anything for Mesos health checks
```
jasongilanfarr: () is unnecessary

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
gkleiman: Both fields are exclusive, and there is a validation that makes sure that at least one of them is set.

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -10,146 +9,210 @@ import org.apache.mesos.{ Protos => MesosProtos }
...
+  protected def protoBuilder: Protos.HealthCheckDefinition.Builder =
+    Protos.HealthCheckDefinition.newBuilder
+      .setGracePeriodSeconds(this.gracePeriod.toSeconds.toInt)
+      .setIntervalSeconds(this.interval.toSeconds.toInt)
+      .setTimeoutSeconds(this.timeout.toSeconds.toInt)
+      .setMaxConsecutiveFailures(this.maxConsecutiveFailures)
+}
 
-  portIndex: Option[Int] = HealthCheck.DefaultPortIndex,
+trait HealthCheckWithPort { this: HealthCheck =>
```
jasongilanfarr: should this also be sealed?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -152,17 +150,16 @@ class TaskBuilder(
...
-    // Mesos supports at most one health check, and only COMMAND checks
-    // are currently implemented in the Mesos health check helper program.
+    // Mesos supports at most one health check
     val mesosHealthChecks: Set[org.apache.mesos.Protos.HealthCheck] =
       runSpec.healthChecks.collect {
-        case healthCheck: HealthCheck if healthCheck.protocol == Protocol.COMMAND => healthCheck.toMesos
+        case mesosHealthCheck: MesosHealthCheck => mesosHealthCheck.toMesos
       }
 
     if (mesosHealthChecks.size > 1) {
```
jasongilanfarr: shouldn't we validate this instead?

jeschkies:
Do we have a story or something similar describing the health checks?

jdef:
Long term goal is to push health checks to Mesos. Apparently documentation
there is a bit lacking https://issues.apache.org/jira/browse/MESOS-5597

On Thu, Aug 25, 2016 at 9:23 PM, Karsten Jeschkies <notifications@github.com

jasongilanfarr:
Review status: 0 of 36 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[project/build.scala, line 150 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQyFAXeI7SYGA9bbCU:-KQQyFAXeI7SYGA9bbCV:bu3z297) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/project/build.scala#L150)):*
> ```Scala
>       "Mesosphere Public Repo"    at "http://downloads.mesosphere.com/maven",
>       "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/",
>       "Local Maven Repository" at "file://"+Path.userHome.absolutePath+"/.m2/repository",
> ```

Can you mark a TODO for yourself?

---

*[src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala, line 57 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQyLetOfKMofNQTQq_:-KQQyLetOfKMofNQTQqa:bwsr2ml) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala#L57)):*
> ```Scala
>     container: Option[Container] = None,
> 
>     healthChecks: Option[Set[_ <: HealthCheck]] = None,
> ```

Isn't Set already Covariant? e.g. no need for a "_ <:"?

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 13 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQyRwQDHkssQnTCmyc:-KQQyRwQDHkssQnTCmyd:bc43bm4) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L13)):*
> ```Scala
> import mesosphere.marathon.core.task.Task
> import mesosphere.marathon.core.event._
> import mesosphere.marathon.core.health._
> ```

Can you increase your import limits? I thought we were trying to avoid wildcard imports unless strictly necessary

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 978 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQywxdWwzSw2MzPMgy:-KQQywxe760l2HQlGFlH:b8mroj1) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L978)):*
> ```Scala
>     val portIndexesMakeSense = app.portDefinitions.nonEmpty || hasPortMappings
> 
>     if (portIndexesMakeSense) app.copy(healthChecks = addHealthCheckPortIndexIfNecessary(app.healthChecks))
> ```

Should we log if it doesn't make sense?

---

*[src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala, line 12 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQz2nBQibiexTg-0mT:-KQQz2nCtRknk3NsO24u:b5ks0yc) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala#L12)):*
> ```Scala
> import scala.concurrent.duration._
> 
> sealed trait HealthCheck extends MarathonState[Protos.HealthCheckDefinition, HealthCheck] {
> ```

Why does this need to extend MarathonState? My impression is we're ultimately trying to get rid of this.

---

*[src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala, line 333 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQR0Ebb3iQ6IZVrX_xD:-KQR0Ebb3iQ6IZVrX_xE:bq9c1ol) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala#L333)):*
> ```Scala
>   }
> 
>   def mergeFromProto(proto: Protos.HealthCheckDefinition): HealthCheck = {
> ```

I really don't like this method name. it really should be "fromProto" unless its going to read the current object and update only _some_ fields.

---

*[src/main/scala/mesosphere/marathon/state/RunSpec.scala, line 56 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQR0d_kxnYB3sv0szhk:-KQR0d_kxnYB3sv0szhl:b-kylm6b) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/state/RunSpec.scala#L56)):*
> ```Scala
>   def container: Option[Container]
> 
>   def healthChecks: Set[_ <: HealthCheck]
> ```

Just HealthCheck should be sufficient.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4210)*

gkleiman:
> Can you mark a TODO for yourself?

I'll actually remove this commit from the PR, as soon as I can upload a maven artifact with Mesos 1.1.0-SNAPSHOT.

>Isn't Set already Covariant? e.g. no need for a "_ <:"?

Scala immutable sets are not covariant.

> Can you increase your import limits? I thought we were trying to avoid wildcard imports unless strictly necessary

👍 

> Should we log if it doesn't make sense?

I don't think we should log, the method sets a default portIndex only if it is necessary.

> Why does this need to extend MarathonState? My impression is we're ultimately trying to get rid of this.

`HealthCheck` used to extend `MarathonState`, but you are right, this is not necessary. I cleaned this up.

> I really don't like this method name. it really should be "fromProto" unless its going to read the current object and update only some fields.

👍 

> Just HealthCheck should be sufficient.

See the second comment.

jdef:
Review status: 0 of 36 files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgIBGph16FbmSMijqA:-KQgIBGph16FbmSMijqB:buo6i3f) (no related file):*
Maybe I'm missing something, but shouldn't marathon track/report the value of mesosProto.taskStatus.healthy for tasks using mesos health checks?

---

*[project/build.scala, line 150 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQyFAXeI7SYGA9bbCU:-KQg9SaAxPBBsErGy5tr:b-m5jvnd) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/project/build.scala#L150)):*
<details><summary><i>Previously, jasongilanfarr (Jason Gilanfarr) wrote…</i></summary>
> Can you mark a TODO for yourself?
</details>
Why is this even needed? Doesn't Ivy provide a cache? If we're going to keep this then can it be envvar controlled? e.g. `MARATHON_BUILD_MVN_CACHE=1` or something?

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 978 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQywxdWwzSw2MzPMgy:-KQgC0IYdpVJUbceJhWU:b-mfq32c) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L978)):*
<details><summary><i>Previously, jasongilanfarr (Jason Gilanfarr) wrote…</i></summary>
> Should we log if it doesn't make sense?
</details>
i'd actually like this variable renamed to something that makes more sense. like `maybeAddPortIndex`

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 648 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgA5OqBFtSbQGXVxDZ:-KQgA5OqBFtSbQGXVxD_:bl43tyq) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L648)):*
> ```Scala
>     HealthCheckWithPortsFormatBuilder ~
>       (__ \ "path").formatNullable[String] ~
>       (__ \ "protocol").formatNullable[Protocol].withDefault(Protocol.HTTP)
> ```

why Protocol.HTTP instead of DefaultProtocol?

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 695 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgBUVgZVUgMEazolLP:-KQgBUVhb0tE7IHodAJ7:b9rgsi1) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L695)):*
> ```Scala
>       case http: MarathonHttpHealthCheck =>
>         Json.toJson(http)(MarathonHttpHealthCheckFormat).as[JsObject] ++
>           Json.obj("protocol" -> Json.toJson(http.protocol))
> ```

why does this need to be appended? `protocol` as a JSON field and `http.protocol` share the same name..

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 701 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgBdNoUp9JTnwYpOmH:-KQgBdNoUp9JTnwYpOmI:b-7ropln) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L701)):*
> ```Scala
>         Json.toJson(tcp)(MesosTcpHealthCheckFormat).as[JsObject] ++ Json.obj("protocol" -> "MESOS_TCP")
>       case http: MesosHttpHealthCheck =>
>         Json.toJson(http)(MesosHttpHealthCheckFormat).as[JsObject] ++ Json.obj("protocol" -> Json.toJson(http.protocol))
> ```

ditto

---

*[src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala, line 216 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgDrfdxNyBat2Rw5XV:-KQgDrfetbboRlipxp0k:b-wynalw) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala#L216)):*
> ```Scala
>     path.foreach(builder.setPath)
> 
>     portIndex.foreach { p => builder.setPortIndex(p) }
> ```

nit: style; why does that path.foreach look different than this one and the one below?

---

*[src/main/scala/mesosphere/marathon/state/AppDefinition.scala, line 747 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgFnrmT186VgQgzI0p:-KQgFnrmT186VgQgzI0q:bywzxs8) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L747)):*
> ```Scala
>   }
> 
>   private val haveAtMostOneMesosHealtCheck: Validator[AppDefinition] =
> ```

Health vs Healt

---

*[src/main/scala/mesosphere/marathon/state/AppDefinition.scala, line 749 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgG1P8slE8BzJFWdzO:-KQgG1P8slE8BzJFWdzP:b-mkb8w7) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/state/AppDefinition.scala#L749)):*
> ```Scala
>   private val haveAtMostOneMesosHealtCheck: Validator[AppDefinition] =
>     isTrue[AppDefinition]("AppDefinition can contain at most one Mesos health check") { appDef =>
>       appDef.healthChecks.filter(_.isInstanceOf[MesosHealthCheck]).size <= 1
> ```

use `count` instead of `filter(...).size` ?

---

*[src/main/scala/mesosphere/marathon/state/PortAssignment.scala, line 13 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgGM3l89dxIyineG4S:-KQgGM3l89dxIyineG4T:b-gxqst2) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/marathon/state/PortAssignment.scala#L13)):*
> ```Scala
>   effectiveIpAddress: Option[String],
>   effectivePort: Int,
>   hostPort: Option[Int] = None,
> ```

hostPort, containerPort => requestedHostPort, requestedContainerPort? if so .. some docs for these?

---

*[src/main/scala/mesosphere/mesos/TaskBuilder.scala, line 173 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgGiz7mb0jeT1yPhit:-KQgGiz7mb0jeT1yPhiu:b-riicuz) ([raw file](https://github.com/mesosphere/marathon/blob/017106e750f85b4ec3d1f64239056db0d9136855/src/main/scala/mesosphere/mesos/TaskBuilder.scala#L173)):*
> ```Scala
>     if (mesosHealthChecks.size > 1) {
>       val numUnusedChecks = mesosHealthChecks.size - 1
>       log.warn(
> ```

nit: isn't this actually an error, because validation should prevent this state from occurring?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4210)*

gkleiman:
Review status: 0 of 36 files reviewed at latest revision, 16 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgIBGph16FbmSMijqA:-KQuGCWbhiQeK8Uq8s20:bhfdie1) (no related file):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> Maybe I'm missing something, but shouldn't marathon track/report the value of mesosProto.taskStatus.healthy for tasks using mesos health checks?
</details>
It already does that for Command health checks, Mesos HTTP/TCP health checks will work in the same way.

---

*[project/build.scala, line 150 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQyFAXeI7SYGA9bbCU:-KQuGZ3L290ImigXXGOU:brqh8gk) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/project/build.scala#L150)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> Why is this even needed? Doesn't Ivy provide a cache? If we're going to keep this then can it be envvar controlled? e.g. `MARATHON_BUILD_MVN_CACHE=1` or something?
</details>
As the commit description says, this is just a temporary hack. I'll now upload the artifact of Mesos 1.1.0-SNAPSHOT to Mesosphere's mvn repo, and replace this hack with a nicer one.

---

*[src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala, line 57 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQyLetOfKMofNQTQq_:-KQuGsqtQIA5BCYKBeeV:b-wcvnn3) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala#L57)):*
<details><summary><i>Previously, jasongilanfarr (Jason Gilanfarr) wrote…</i></summary>
> Isn't Set already Covariant? e.g. no need for a "_ <:"?
</details>
Already answered in GitHub. (No, it isn't covariant)

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 978 \[r2\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQQywxdWwzSw2MzPMgy:-KQuv3i9rsyilS65BVre:bu5mv9i) ([raw file](https://github.com/mesosphere/marathon/blob/79d66ac37b21cf6afcaf22eb3b71f7234b07fa5b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L978)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> i'd actually like this variable renamed to something that makes more sense. like `maybeAddPortIndex`
</details>
we normally use the `maybe`prefix for Options, this is not an option, just a boolean to make the if statement shorter..

This method does the following:
 - Checks if an app definition specifies ports or port mappings, if that's the case, it calls another method. The other method checks if all HTTP/TCP health checks specify a portIndex or port. If that is not the case, then it will add portIndex = 0 to the corresponding HTTP/TCP health checks.

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 648 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgA5OqBFtSbQGXVxDZ:-KQuuvJgLvUvRj82Tjtb:b-896fix) ([raw file](https://github.com/mesosphere/marathon/blob/edf862824eab5c5dba50c9c6a491d9c33ae690a5/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L648)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> why Protocol.HTTP instead of DefaultProtocol?
</details>
Done.

---

*[src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala, line 695 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4210#-KQgBUVgZVUgMEazolLP:-KQuuxp_xhzg1Op1MGqP:b-rlmsgx) ([raw file](https://github.com/mesosphere/marathon/blob/edf862824eab5c5dba50c9c6a491d9c33ae690a5/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala#L695)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> why does this need to be appended? `protocol` as a JSON field and `http.protocol` share the same name..
</details>
Added extra tests to make sure we always add the protocol.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4210)*

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -903,26 +960,37 @@ trait AppAndGroupFormats {
...
+      case healthCheck: MarathonHttpHealthCheck =>
+        def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty
+        if (needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
+        else healthCheck
+      case healthCheck: MesosTcpHealthCheck =>
+        def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty
+        if (needsDefaultPortIndex) healthCheck.copy(portIndex = Some(0))
+        else healthCheck
+      case healthCheck: MesosHttpHealthCheck =>
+        def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty
```
aquamatthias: Looks like 4 times the same functionality besides the copy function. Perhaps extract this to a separate method:
```scala
def withPort[T<:HealthCheck](healthCheck:T, adapt: T=>T): T = {
  def needsDefaultPortIndex = healthCheck.port.isEmpty && healthCheck.portIndex.isEmpty
  if (needsDefaultPortIndex) adapt(healthCheck)  else healthCheck
}
withPort(hc, hc.copy(portIndex = Some(0))
```

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -10,146 +9,329 @@ import org.apache.mesos.{ Protos => MesosProtos }
 
 import scala.concurrent.duration._
 
-case class HealthCheck(
+sealed trait HealthCheck {
```
aquamatthias: Thanks for introducing a trait here.

aquamatthias:
Thanks! 
```

                 ;~
               ./|\.
             ./ /| `\.
            /  | |   `\.
           |   | |     `\.
           |    \|       `\.
         .  `----|__________\.
          \-----''----.....___
mesos      \               ""/
      ~^~^~^~^~^`~^~^`^~^~^`^~^~^
       ~^~^~`~~^~^`~^~^~`~~^~^~



```

astawiarski:
Which release is this planned to be part of ?

jdef:
will be first introduced in the v1.4.x series

PR #4218, opened 2016-08-13
remove artificial memory constraints from docker-run, TC is failing a…

jdef:
…nd I think we are running out of memory


jasongilanfarr:
retest this please

PR #4224, opened 2016-08-15
Fix GPU validation if 'container' isn't set.

nfnt:
GPU validation would fail if `container` isn't set in an app definition. Because in that case the Mesos Containerizer will be used, the validation shouldn't fail.

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -7,7 +7,7 @@ import com.wix.accord.combinators.GeneralPurposeCombinators
 import com.wix.accord.dsl._
 import mesosphere.marathon.Protos.Constraint
 import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
-import mesosphere.marathon.state.Container.{ Mesos, MesosAppC, MesosDocker }
+import mesosphere.marathon.state.Container.Docker
 // scalastyle:off
 import mesosphere.marathon.api.serialization.{ ContainerSerializer, EnvVarRefSerializer, PortDefinitionSerializer, ResidencySerializer, SecretsSerializer }
 // scalastyle:on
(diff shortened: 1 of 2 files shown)


klueska:
Do we know why two of these checks were not successful?

jdef:
@jasongilanfarr and others have been working to improve the test flake situation. I suggest that you rebase this PR to see if that helps

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -657,7 +657,9 @@ object AppDefinition extends GeneralPurposeCombinators {
 
   private val complyWithGpuRules: Validator[AppDefinition] = conditional[AppDefinition](_.gpus > 0) {
     isTrue[AppDefinition]("GPU resources only work with the Mesos containerizer") { app =>
-      app.container.exists{
+      // If 'container' is not defined, the Mesos containerizer is used.
+      // By using forall, this case will yield true.
+      app.container.forall{
```
meichstedt: This depends on how we want to formulate the rule. I'd say, GPU is allowed for _everything but Docker_, so I'd write
```
app.container match {
  case Some(_: Docker) => false
  case _ => true
}
```

I we wanted to say _we specifically allow MesosDocker, MesosAppC, Mesos but nothing else_, then the proposed solution would be ok'ish (yet it's cumbersome to have the `forall` plus a pattern match for when the option is non empty)

meichstedt:
@nfnt One little comment – please think about that and feel free to not change anything, it's semantically correct.

Apart from that, please rebase! Happy to merge this.

klueska:
I know @nfnt is still on vacation, but I'd really like to see this make it into 1.3.

nfnt:
Vacation's over, @meichstedt, your comment has been addressed.

klueska:
Hooray!

PR #4239, opened 2016-08-22
Correctly pass in the password when using new zookeeper storage.

jasongilanfarr:
I was wondering why some tests had thousands of "NoAuth" lines...


gkleiman:
@jasongilanfarr LGTM. There's a weird merge commit in the PR, can you rebase please?

PR #4241, opened 2016-08-22
Expunge lost tasks if the task update message indicates it won't come back

gkleiman:


gkleiman:
cc/ @graham-murphy @bernadinm @neilconway

jasongilanfarr:
Looks as good to me as we likely can get for a while.

neilconway:
LGTM, at least until we have better Mesos features upstream.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/StatusUpdateActionResolverTest.scala
```diff
@@ -74,6 +74,17 @@ class StatusUpdateActionResolverTest
     }
   }
 
+  test("a TASK_LOST update with a message saying that the task is unknown to the slave is mapped to an expunge") {
+    val f = new Fixture
+    val update = TaskStatusUpdateTestHelper.lost(
+      TaskStatus.Reason.REASON_SLAVE_DISCONNECTED, "Reconciliation: Task is unknown to the slave")
```
meichstedt: The combination in Mesos master is, as indicated by the message, reconciliation. The only place where this message is set, is
https://github.com/mesosphere/mesos/blob/master/src/master/master.cpp#L5557
> `"Reconciliation: Task is unknown to the agent",`
> `TaskStatus::REASON_RECONCILIATION);``

This should be reflected in this test as well, so that we don't test invalid combinations

src/main/scala/mesosphere/marathon/core/task/bus/MesosTaskStatus.scala
```diff
@@ -18,9 +18,15 @@ object MesosTaskStatus {
 
   val WontComeBack: Set[TaskStatus.Reason] = TaskStatus.Reason.values().toSet.diff(MightComeBack)
 
+  def wontComeBack(status: TaskStatus): Boolean = {
+    // FIXME (gkleiman): comparing against a string is ugly and fragile, we shouldn't do it, but there's no better way
+    // for now
+    WontComeBack(status.getReason) || status.getMessage.startsWith("Reconciliation: Task is unknown to the")
```
meichstedt: This is correct but not sufficient. The `TemporarilyUnreachable` extractor must also consider the message if the reason is `TASK_LOST`.

Besides, The `MightComeBack` set includes `REASON_RECONCILIATION`, which is not completely correct any more.

src/main/scala/mesosphere/marathon/core/task/bus/MesosTaskStatus.scala
```diff
@@ -36,7 +38,7 @@ object MesosTaskStatus {
   object TemporarilyUnreachable {
     def isUnreachable(task: MarathonTask): Boolean = isUnreachable(task.getStatus)
     def isUnreachable(taskStatus: TaskStatus): Boolean = {
-      taskStatus.getState == TASK_LOST && MightComeBack(taskStatus.getReason)
+      taskStatus.getState == TASK_LOST && MightComeBack(taskStatus.getReason) && !wontComeBack(taskStatus)
```
meichstedt: I'd have preferred a `def mightComeBack` in contrast to the double negation, but I can live with it in releases/0.15 👍

meichstedt:
Thanks @gkleiman!

I'm afraid I'll have to merge this without successful tests, as the config is not compatible with release branches ..

unterstein:
retest this please

unterstein:
@gkleiman adapted the test plan. If you (I`ll tried, but I think the hook works only for creators) post `retest this please` as comment, the CI should do another run :)

gkleiman:
retest this please

unterstein:
hmmm ... seams not to be working :(

PR #4242, opened 2016-08-23
Don't consider lost tasks when evaluating constraints.

gkleiman:
This change should let Marathon start a new task to replace a lost one,
even if an app is pinned to a slave. This could potentially lead to more
than one task running on the same agent, violating a UNIQUE constraint.
Marathon will in that case try to remediate it by killing the new
running/staging task.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
@@ -332,7 +332,11 @@ private class AppTaskLauncherActor(
       sender ! MatchedTasks(offer.getId, Seq.empty)
 
     case ActorOfferMatcher.MatchOffer(deadline, offer) =>
-      val newTaskOpt: Option[CreatedTask] = taskFactory.newTask(app, offer, tasksMap.values)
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/impl/AppTaskLauncherActor.scala
```diff
@@ -332,7 +332,11 @@ private class AppTaskLauncherActor(
       sender ! MatchedTasks(offer.getId, Seq.empty)
 
     case ActorOfferMatcher.MatchOffer(deadline, offer) =>
-      val newTaskOpt: Option[CreatedTask] = taskFactory.newTask(app, offer, tasksMap.values)
+      val newTaskOpt: Option[CreatedTask] = taskFactory.newTask(
+        app,
+        offer,
+        tasksMap.values.filterNot(_.getStatus.getState == TaskState.TASK_LOST))
```
jdef: is it worth checking the reason here too? LOST could happen for several reasons right? The goal should be to "not consider terminal tasks when evaluating constraints"

meichstedt: the task should only be in state if the reason/message indicate that it's unreachable, so I think a check for state `TASK_LOST` is sufficient

PR #4250, opened 2016-08-24
WIP: v2 raml for pods

jasongilanfarr:
Purely an API sketch still.
The new stuff is all RAML 1.0

TODO:
- [ ] add pods to v2 groups API (will report PodStatus objects, not Pod)
- [ ] strip out the `additionalProperties: false` from most things (as per @jasongilanfarr)
- [x] align command spec w/ how mesos deals with `entrypoint` in UNICON docker isolator

References:
- docker isolator CommandInfo decoder table
  - https://github.com/apache/mesos/blob/6f2a452d4876b2073ad552df88aa6e49f3e36a34/src/slave/containerizer/mesos/isolators/docker/runtime.cpp#L255
- appc isolator CommandInfo decoder table
  - https://github.com/apache/mesos/blob/6f2a452d4876b2073ad552df88aa6e49f3e36a34/src/slave/containerizer/mesos/isolators/appc/runtime.cpp#L203

### DIFF
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -23,7 +23,7 @@ traits: !include ../traits.raml
 
   get:
     is: [ secured ]
-    description:|
+    description: |
       List all the pod-based services in the system.
       TODO(jdef) do we really need this?
     responses:
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -3,3 +3,20 @@ title: v2 pod API
 mediaType: application/json
 types:
   Pod: !include types/pod.raml
+  PathId: !include types/pathId.raml
+/{id}:
+  uriParameters:
+    id:
+      type: PathId
+      description: The path of the pod
+  # TODO: traits.raml needs to be updated... is: [ secured ]
+  get:
+    description:
+      Get the pod at the given id
+    responses:
+      200:
+        body:
+          application/json:
+            type: Pod
+      404:
+        description: No pod found with this `id`
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -2,8 +2,9 @@
 title: v2 pod API
 mediaType: application/json
 types:
-  Pod: !include types/pod.raml
   PathId: !include types/pathId.raml
+  Pod: !include types/pod.raml
+  PodStatus: !include types/pod-status.raml
 traits: !include ../traits.raml
 
 /:
@@ -62,3 +63,23 @@ traits: !include ../traits.raml
         body:
           application/json:
             type: Pod
+
+'/{id}::status':
+  usage: |
+    Implementations should probably generate do-not-cache headers because
+    status should always be live.
+
+  uriParameters:
+    id:
+      type: PathId
+      description: The path of the pod
+
+  get:
+    is: [ secured, objectLocator ]
+    description: |
+      Get the status of the pod with the given id
+    responses:
+      200:
+        body:
+          application/json:
+            type: PodStatus
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -12,6 +12,8 @@ traits: !include ../traits.raml
     is: [ secured, jsonValidator, objectValidator, objectCreator ]
     description: |
       Create and start a new pod-based service.
+      TODO(jdef) would be nice to include link to deployment URI in a Location
+      (or some such) header.
     body:
       application/json:
         type: Pod
@@ -44,12 +46,16 @@ traits: !include ../traits.raml
     is: [ secured, objectLocator, deployable ]
     description: |
       Delete an existing pod-based service.
+      TODO(jdef) would be nice to include link to deployment URI in a Location
+      (or some such) header.
 
   put:
     is: [ secured, jsonValidator, objectValidator, objectLocator, deployable ]
     description: |
       Update an existing pod-based service.
       TODO(jdef) separate error code when update of an immutable field fails?
+      TODO(jdef) would be nice to include link to deployment URI in a Location
+      (or some such) header.
     body:
       application/json:
         type: Pod
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -28,13 +28,27 @@ traits: !include ../traits.raml
     is: [ secured ]
     description: |
       List all the pod-based services in the system.
-      TODO(jdef) do we really need this?
     responses:
       200:
-        description: A list of all pods in the system.
+        description: |
+          Yields a list of all pods in the system.
+          Useful to perform backups of all pods registered with Marathon.
         body:
           application/json:
             type: Pod[]
+
+  put:
+    is: [ secured, jsonValidator, objectValidator, objectCreator ]
+    description: |
+      Create or update each pod in the list.
+      Useful to restore all pods previously backed up via GET /v2/pods.
+      TODO(jdef) this is a stretch goal for the MVP.
+    responses:
+      207:
+        description: |
+          A list of all created/updated pods.
+          TODO(jdef) format for the multi-status response type, see
+          https://tools.ietf.org/html/rfc4918#section-13
 
 /{id}:
   uriParameters:
@@ -71,7 +85,7 @@ traits: !include ../traits.raml
             type: Pod
 
 '/{id}::status':
-  usage: |
+  description: |
     Implementations should probably generate do-not-cache headers because
     status should always be live.
 
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -1,0 +1,5 @@
+#%RAML 1.0
+title: v2 pod API
+mediaType: application/json
+types:
+  Pod: !include types/pod.raml
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -27,7 +27,7 @@ post:
   delete:
     is: [ secured, objectLocator, deployable ]
     description:|
-      Update an existing pod-based service.
+      Delete an existing pod-based service.
 
   put:
     is: [ secured, jsonValidator, objectValidator, objectLocator, deployable ]
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -4,13 +4,42 @@ mediaType: application/json
 types:
   Pod: !include types/pod.raml
   PathId: !include types/pathId.raml
+
+post:
+  is: [ secured, jsonValidator, objectValidator, objectCreator ]
+  description:|
+    Create and start a new pod-based service.
+  body:
+    application/json:
+      type: Pod
+  responses:
+    201:
+      description: Pod created successfully.
+      body:
+        application/json:
+          type: Pod
+
 /{id}:
   uriParameters:
     id:
       type: PathId
       description: The path of the pod
-  # TODO: traits.raml needs to be updated... is: [ secured ]
+  delete:
+    is: [ secured, objectLocator, deployable ]
+    description:|
+      Update an existing pod-based service.
+
+  put:
+    is: [ secured, jsonValidator, objectValidator, objectLocator, deployable ]
+    usage: TODO(jdef) separate error code when update of an immutable field fails?
+    description:|
+      Update an existing pod-based service.
+    body:
+      application/json:
+        type: Pod
+
   get:
+    is: [ secured, objectLocator ]
     description:
       Get the pod at the given id
     responses:
@@ -18,5 +47,3 @@ types:
         body:
           application/json:
             type: Pod
-      404:
-        description: No pod found with this `id`
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -4,20 +4,22 @@ mediaType: application/json
 types:
   Pod: !include types/pod.raml
   PathId: !include types/pathId.raml
+traits: !include ../traits.raml
 
-post:
-  is: [ secured, jsonValidator, objectValidator, objectCreator ]
-  description:|
-    Create and start a new pod-based service.
-  body:
-    application/json:
-      type: Pod
-  responses:
-    201:
-      description: Pod created successfully.
-      body:
-        application/json:
-          type: Pod
+/:
+  post:
+    is: [ secured, jsonValidator, objectValidator, objectCreator ]
+    description: |
+      Create and start a new pod-based service.
+    body:
+      application/json:
+        type: Pod
+    responses:
+      201:
+        description: Pod created successfully.
+        body:
+          application/json:
+            type: Pod
 
 /{id}:
   uriParameters:
@@ -26,21 +28,21 @@ post:
       description: The path of the pod
   delete:
     is: [ secured, objectLocator, deployable ]
-    description:|
+    description: |
       Delete an existing pod-based service.
 
   put:
     is: [ secured, jsonValidator, objectValidator, objectLocator, deployable ]
-    usage: TODO(jdef) separate error code when update of an immutable field fails?
-    description:|
+    description: |
       Update an existing pod-based service.
+      TODO(jdef) separate error code when update of an immutable field fails?
     body:
       application/json:
         type: Pod
 
   get:
     is: [ secured, objectLocator ]
-    description:
+    description: |
       Get the pod at the given id
     responses:
       200:
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -21,11 +21,24 @@ traits: !include ../traits.raml
           application/json:
             type: Pod
 
+  get:
+    is: [ secured ]
+    description:|
+      List all the pod-based services in the system.
+      TODO(jdef) do we really need this?
+    responses:
+      200:
+        description: A list of all pods in the system.
+        body:
+          application/json:
+            type: Pod[]
+
 /{id}:
   uriParameters:
     id:
       type: PathId
       description: The path of the pod
+
   delete:
     is: [ secured, objectLocator, deployable ]
     description: |
--- a/docs/docs/rest-api/public/api/v2/pods.raml
+++ b/docs/docs/rest-api/public/api/v2/pods.raml
@@ -1,6 +1,7 @@
 #%RAML 1.0
 title: v2 pod API
 mediaType: application/json
+annotationTypes:  !include ../annotations.raml
 types:
   PathId: !include types/pathId.raml
   Pod: !include types/pod.raml
--- a/docs/docs/rest-api/public/api/v2/schema/Group.json
+++ b/docs/docs/rest-api/public/api/v2/schema/Group.json
@@ -23,6 +23,13 @@
         "$ref": "./AppDefinition.json"
       }
     },
+    "pods" : {
+      "type": "array",
+      "description": "The list of PodDefinitions in this group. See PodDefinition.json for the schema.",
+      "items": {
+        "$ref": "./PodDefinition.json"
+      }
+    },
     "groups": {
       "type": "array",
       "description": "Groups can build a tree. Each group can contain sub-groups. The sub-groups are defined here.",
--- a/docs/docs/rest-api/public/api/v2/types/container-status.raml
+++ b/docs/docs/rest-api/public/api/v2/types/container-status.raml
@@ -49,7 +49,7 @@ properties:
           description: true if a health check is defined for this endpoint and is passing
   termination?:
     type: object
-    additionalProperies: false
+    additionalProperties: false
     properties:
       exitCode?:
         type: integer
--- a/docs/docs/rest-api/public/api/v2/types/container-status.raml
+++ b/docs/docs/rest-api/public/api/v2/types/container-status.raml
@@ -1,0 +1,56 @@
+#%RAML 1.0 DataType
+type: object
+additionalProperties: false
+properties:
+  name:
+    type: !include name.raml
+    description: name of the container specification (within the pod)
+  status:
+    type: string
+    enum: [ STAGING, STABLE, DEGRADED, TERMINAL ]
+    description:|
+      STAGING  - Container has been launched but is not yet running.
+      STABLE   - Container is running and, if health checks were specified, is healthy.
+      DEGRADED - Container is running but is not stable.
+      TERMINAL - Container is in the process of terminating or else has terminated.
+  message?:
+    type: string
+    description:|
+      Human-friendly explanation for the container's current status.
+  containerId?:
+    type: string
+    description:|
+      Unique ID of this container in the cluster.
+      TODO(jdef) Probably represents the Mesos task ID.
+  endpoints?:
+    type: array
+    items:
+      type: object
+      additionalProperties: false
+      properties:
+        name:
+          type: !include name.raml
+          description: name of the endpoint
+        allocatedHostPort?:
+          type: !include port.raml
+        healthy?:
+          type: boolean
+          usage: should only be present if a health check is defined for this endpoint
+          description: true if a health check is defined for this endpoint and is passing
+  termination?:
+    type: object
+    additionalProperies: false
+    properties:
+      exitCode?:
+        type: integer
+      message?:
+        type: string
+        description: Human-explanation for container termination.
+  lastUpdated:
+    type: datetime-only
+    description:|
+      Time that this status was last checked and updated (even if nothing changed)
+  lastChanged:
+    type: datetime-only
+    description:|
+      Time that this status was last modified (some aspect of status did change)
--- a/docs/docs/rest-api/public/api/v2/types/container-status.raml
+++ b/docs/docs/rest-api/public/api/v2/types/container-status.raml
@@ -1,11 +1,13 @@
 #%RAML 1.0 DataType
 type: object
+(scalaType): ContainerStatus
 additionalProperties: false
 properties:
   name:
     type: !include name.raml
     description: name of the container specification (within the pod)
   status:
+    (scalaType): ContainerState
     type: string
     usage: |
       TODO(jdef) this is not an enum because it reflects Marathon internal state that
@@ -35,6 +37,7 @@ properties:
   endpoints?:
     type: array
     items:
+      (scalaType): ContainerEndpointStatus
       type: object
       additionalProperties: false
       properties:
@@ -48,6 +51,7 @@ properties:
           usage: should only be present if a health check is defined for this endpoint
           description: true if a health check is defined for this endpoint and is passing
   termination?:
+    (scalaType): ContainerTerminationState
     type: object
     additionalProperties: false
     properties:
--- a/docs/docs/rest-api/public/api/v2/types/container-status.raml
+++ b/docs/docs/rest-api/public/api/v2/types/container-status.raml
@@ -47,10 +47,10 @@ properties:
         type: string
         description: Human-explanation for container termination.
   lastUpdated:
-    type: datetime-only
+    type: datetime
     description: |
       Time that this status was last checked and updated (even if nothing changed)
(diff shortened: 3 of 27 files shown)


docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -0,0 +1,22 @@
+#%RAML 1.0
+title: v2 pod API
+mediaType: application/json
+types:
+  Pod: !include types/pod.raml
+  PathId: !include types/pathId.raml
+/{id}:
+  uriParameters:
+    id:
+      type: PathId
```
jdef: so our `PathId` could have `/` chars in it. Is that legal in RAML? I couldn't find a section of the doc that explained if that was allowed

docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -0,0 +1,22 @@
+#%RAML 1.0
```
jdef: what happens when a user queries /v2/pods/?

docs/docs/rest-api/public/api/v2/pods.raml
jasongilanfarr: No idea

docs/docs/rest-api/public/api/v2/pods.raml
jasongilanfarr: Nothing? I really don't like the idea of listing hundreds to thousands of pods in one massive json, don't see why not though...

docs/docs/rest-api/public/api/v2/types/endpoint.raml
```diff
@@ -0,0 +1,29 @@
...
+      The container port that a task's process is actually listening on.
+      Required if the network mode is container
+  hostPort?:
+    type: !include port.raml
+    description: |
+      Mapped port on the host
+      All host ports are allocated from resource offers
+  protocol?:
+    type: string
+    enum: [UDP, TCP]
```
jdef: this should be a free-form string

jdef: it's used for discovery

docs/docs/rest-api/public/api/v2/types/healthCheck.raml
```diff
@@ -0,0 +1,33 @@
...
+        minLength: 1
+        maxLength: 1024
+      scheme?:
+        type: string
+        enum: [ HTTP, HTTPS ]
+        description: The http scheme to use
+  tcp?:
+    type: object
+    properties:
+      type: !include name.raml
```
jdef: this should be the name of the endpoint. i think you're missing a line here

docs/docs/rest-api/public/api/v2/types/label.raml
```diff
@@ -0,0 +1,10 @@
+#%RAML 1.0 DataType
+type: object
+properties:
+  name:
+    type: string
+    minLength: 1
+    maxLength: 255
+    pattern: (?!^((MARATHON_)|(DCOS_)|(MESOS_)).*)
```
jdef: i like the pattern. but it stops people from using special labels that we come up with. "reserved for internal use" was perhaps the wrong statement. maybe "reserved for use by marathon, mesos, and/or dcos" would have been better? basically don't want user's inventing arbitrary labels that they craft names for. but we may tell them to set some special MARATHON_XYZ label to accomplish something

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -0,0 +1,75 @@
...
+      The maximum number of instances of this pod that should
+      be running at a given time.
+    minimum: 0
+  constraints?:
+    type: array
+    items: !include constraint.raml
+    description:|
+      Valid constraint operators are one of UNIQUE, CLUSTER, GROUP_BY, LIKE, UNLIKE, MAX_PER
+    example: [hostname:UNIQUE]
+  user?:
```
jdef: I think having user here (instead of at the container level) is a mistake. Totally reasonable that people want to run different tasks as different users. @vinodkone will the mesos MVP support subcontainers/tasks that run as different users?

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -0,0 +1,75 @@
...
+    type: string
+    description: The user to use to run the tasks on the agent
+  version?:
+    type: datetime-only
+    description: The version of the definition, immutable
+  versionInfo?: !include versionInfo.raml
+  env?:
```
jdef: should be consistent w/ names. either container.environment and use the environment here, or both should use env

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -0,0 +1,75 @@
...
+    description:|
+      Valid constraint operators are one of UNIQUE, CLUSTER, GROUP_BY, LIKE, UNLIKE, MAX_PER
+    example: [hostname:UNIQUE]
+  user?:
+    type: string
+    description: The user to use to run the tasks on the agent
+  version?:
+    type: datetime-only
+    description: The version of the definition, immutable
+  versionInfo?: !include versionInfo.raml
```
jdef: is this legal? instead of putting `type: !include versionInfo.raml` on the following line?

docs/docs/rest-api/public/api/v2/types/port.raml
```diff
@@ -0,0 +1,4 @@
+#%RAML 1.0 DataType
+type: integer
+minimum: 1
```
jdef: hostport needs to have a value of 0 (or some other way to indicate that marathon should "just pick one" from the offered port range)

docs/docs/rest-api/public/api/v2/types/resources.raml
```diff
@@ -0,0 +1,30 @@
+#%RAML 1.0 DataType
+type: object
+description: Resource Allocations
+properties:
+  cpus:
+    displayName: cpu shares
+    description: The number of CPU shares this pod needs per instance. This number does not have to be integer, but can be a fraction.
+    type: number
+    minimum: 0
```
jdef: should probably be 0.001 (the lowest value that has meaning given fixed precision scalars in mesos)

docs/docs/rest-api/public/api/v2/types/resources.raml
```diff
@@ -0,0 +1,30 @@
...
+    format: float
+    example: 0.1
+  mem:
+    displayName: memory shares
+    type: number
+    minimum: 0
```
jdef: should probably be 32 since we're only defining resources at the pod level - this is the min memory that mesos needs to launch something

docs/docs/rest-api/public/api/v2/types/volumeMount.raml
```diff
@@ -0,0 +1,62 @@
+#%RAML 1.0 DataType
```
jdef: this type is very strange. i think too much got mushed together here. containers need to share access to the same volume. the specs for the volume should be defined ONCE at the pod level. containers can then choose how each volume is mounted into their own private mountns.

docs/docs/rest-api/public/api/v2/types/endpoint.raml
jasongilanfarr: Ahh ok

docs/docs/rest-api/public/api/v2/types/healthCheck.raml
jasongilanfarr: Whoops. Should be name (and you can inline the type of you don't add anything else)

docs/docs/rest-api/public/api/v2/types/label.raml
jasongilanfarr: Ahh. I kind of don't like that pattern, but sure.

docs/docs/rest-api/public/api/v2/types/pod.raml
jasongilanfarr: I was just thinking "shared namespace", let's verify here.

docs/docs/rest-api/public/api/v2/types/pod.raml
jasongilanfarr: Yep, you just can't add fields. It's basically the same as string.

docs/docs/rest-api/public/api/v2/types/pod.raml
jasongilanfarr: Agreed

docs/docs/rest-api/public/api/v2/types/port.raml
jasongilanfarr: Ahh, right, typo

docs/docs/rest-api/public/api/v2/types/resources.raml
jasongilanfarr: Oops

docs/docs/rest-api/public/api/v2/types/resources.raml
jasongilanfarr: Yep

docs/docs/rest-api/public/api/v2/types/volumeMount.raml
jasongilanfarr: Agreed, I may have mixed pod and container volumes together by accident. It was long and kind of confusing in v3

docs/docs/rest-api/public/api/v2/types/pod.raml
jdef: namespace is related, but the fact is that each task will be running different commands and it's highly likely that different commands will be run as different users, regardless of whether the userns is shared

docs/docs/rest-api/public/api/v2/types/resources.raml
jdef: given comments from @vinodkone we should be defining resources at the container level. AFAIK the total min resources of the pod should be enforced by Marathon: cpus(0.01) mem(32). I'm not aware of min resource requirements on a per-task level

jdef: ...so maybe we should just leave this mins as-is and document the overall pod min req's somewhere

docs/docs/rest-api/public/api/v2/types/port.raml
jdef: I used an override on the hostPort declaration

jdef:
Among other, general cleanup I've moved `resources` to mesosContainer: mesos requires that individual tasks are assigned resources so this makes sense to expose at the API level. It's not hard to sum up resources to determine the requirements for a pod so this shouldn't be a big deal.

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -0,0 +1,80 @@
...
+    required: true
+    example: /ops/audit
+  containers:
+    type: array
+    items: !include mesosContainer.raml
+  instances:
+    type: integer
+    description: The number of instances of this pod to start.
+    minimum: 0
+  maxInstances?:
```
meichstedt: Is this meant to replace an upgradeStrategy?

docs/docs/rest-api/public/api/v2/types/mesosContainer.raml
```diff
@@ -0,0 +1,43 @@
...
+      The OS user to use to run the tasks on the agent.
+      Values here override a "user" value specified in the pod definition.
+  healthCheck?:
+    type: !include healthCheck.raml
+    description: All healthchecks to perform on the container
+  volumeMounts?:
+    type: array
+    items: !include volumeMount.raml
+    description: All volume mounts
+  artifacts?:
```
meichstedt: The initial v3 api doc specified artifacts on pod level (= shared artifacts) **and** on container level.
Would artifacts on pod level only be sufficient for the MVP? Just wondering because this api spec doesn't allow artifacts on pod level.

docs/docs/rest-api/public/api/v2/types/pod.raml
jdef: no it's meant to replace the SINGLE_APP_INSTANCE label. i'm adding upgradestrategy in a follow up commit

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -0,0 +1,80 @@
...
+    description:|
+      Valid constraint operators are one of UNIQUE, CLUSTER, GROUP_BY, LIKE, UNLIKE, MAX_PER
+    example: [hostname:UNIQUE]
+  user?:
+    type: string
+    description:|
+      The OS user to use to run the tasks on the agent.
+      May be overridden by a container.
+  version?:
+    type: datetime-only
```
jdef: why is this a `datetime-only` type? shouldn't the nature of this data be opaque from an API perspective (could be a integer, timestamp, string, whatever -- the user should not care since they never set it)

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -0,0 +1,80 @@
...
+    description: The version of the definition, immutable
+  versionInfo?:
```
jdef: this looks like status, not part of a pod spec

docs/docs/rest-api/public/api/v2/types/mesosContainer.raml
jdef: as discussed via hangout, need to confirm w/ the mesos team the order of operations: are pod artifacts downloaded before or after volume creation/mounts? if after then it makes sense to include artifacts at the pod level so that tasks may share access to the same artifacts

docs/docs/rest-api/public/api/v2/types/volumeMount.raml
jdef: resolved.

docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -0,0 +1,49 @@
...
+
+/{id}:
+  uriParameters:
+    id:
+      type: PathId
+      description: The path of the pod
+  delete:
+    is: [ secured, objectLocator, deployable ]
+    description:|
+      Update an existing pod-based service.
```
jdef: s/Update/Delete/

docs/docs/rest-api/public/api/v2/schema/Group.json
```diff
@@ -23,6 +23,13 @@
         "$ref": "./AppDefinition.json"
       }
     },
+    "pods" : {
+      "type": "array",
+      "description": "The list of PodDefinitions in this group. See PodDefinition.json for the schema.",
+      "items": {
+        "$ref": "./PodDefinition.json"
```
jdef: wonder if there's a nice tool to translate from RAML to jsonschema

jasongilanfarr: its built into RAML

docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -0,0 +1,49 @@
...
+    Create and start a new pod-based service.
+  body:
+    application/json:
+      type: Pod
+  responses:
+    201:
+      description: Pod created successfully.
+      body:
+        application/json:
+          type: Pod
```
jasongilanfarr: Maybe we should return "{ deploymentPlan, pod }"?

docs/docs/rest-api/public/api/v2/schema/Group.json
jasongilanfarr: There's a tool in the raml toolbox to do it. Can't remember what its called but its referenced in their spec.

docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -0,0 +1,49 @@
...
+      Delete an existing pod-based service.
+
+  put:
+    is: [ secured, jsonValidator, objectValidator, objectLocator, deployable ]
+    usage: TODO(jdef) separate error code when update of an immutable field fails?
+    description:|
+      Update an existing pod-based service.
+    body:
+      application/json:
+        type: Pod
```
jasongilanfarr: ditto above. Maybe the plan should actually be in a Location header?

docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -21,11 +21,24 @@ traits: !include ../traits.raml
           application/json:
             type: Pod
 
+  get:
```
jasongilanfarr: I vote no.

docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -0,0 +1,64 @@
...
+        description: Pod created successfully.
+        body:
+          application/json:
+            type: Pod
+
+  get:
+    is: [ secured ]
+    description:|
+      List all the pod-based services in the system.
+      TODO(jdef) do we really need this?
```
jdef: jason voted no (GH is hiding the comment, so I'm restating this here)

docs/docs/rest-api/public/api/v2/types/pod-status.raml
```diff
@@ -0,0 +1,58 @@
+#%RAML 1.0 DataType
+type: object
+description: |
```
jasongilanfarr: Description was copied to status...

docs/docs/rest-api/public/api/v2/types/pod-status.raml
```diff
@@ -0,0 +1,58 @@
...
+  A pod allows one to launch a collection co-located (on the same agent) containers
+  that share the same network namespace and that may share the same volumes.
+  Resources are specified on a per-container basis.
+additionalProperties: false
+properties:
+  id:
+    type: !include pathId.raml
+    description:|
+      Fully qualified name of the pod that this status object is describing.
+  spec:
```
jasongilanfarr: spec?

docs/docs/rest-api/public/api/v2/types/pod-status.raml
```diff
@@ -0,0 +1,58 @@
...
+    type: !include pod.raml
+    description:|
+      The latest version of the pod specification (that has the same pod ID).
+  status:
+    type: string
+    enum: [ DEGRADED, STABLE, TERMINAL ]
```
jasongilanfarr: Missing a state for "all instances have started and health checks are specified but one or more are unhealthy"

docs/docs/rest-api/public/api/v2/types/pod-status.raml
```diff
@@ -0,0 +1,58 @@
...
+    items:
+      type: !include pod-instance-status.raml
+  terminationHistory?:
+    type: array
+    description:|
+      List of most recent instance terminations.
+      TODO(jdef) determine how many items might show up here; current thinking is .. not many
+    items:
+      type: object
+      additionalProperties: false
```
jasongilanfarr: i really think this should _only_ be used when the object needs the equivalent of "oneOf". Putting this in says "we will never add anything to this"

docs/docs/rest-api/public/api/v2/types/pod-status.raml
```diff
@@ -0,0 +1,58 @@
...
+      properties:
+        instanceID:
+          type: string
+        startedAt:
+          type: datetime-only
```
jasongilanfarr: prefer "datetime", timezone information, even if UTC is valuable.

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
```diff
@@ -0,0 +1,62 @@
...
+      Unique ID of this pod instance in the cluster.
+      TODO(jdef) Probably represents the Mesos executor ID.
+  status:
+    type: string
+    enum: [ PENDING, STAGING, STABLE, DEGRADED, TERMINAL ]
+    description:|
+      PENDING  - Instance is queued for launch.
+      STAGING  - Instance has been launched but is not yet running.
+      STABLE   - Instance is running and, if health checks were specified, is healthy.
+      DEGRADED - One or more of the instance's containers are not stable.
```
jasongilanfarr: UNHEALTHY? I don't really understand this state, especially for the instance.

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
```diff
@@ -0,0 +1,62 @@
...
+          description: name of the network
+        addresses?:
+          type: string[]
+  containers?:
+    type: array
+    description: status for each running container of this instance.
+    items:
+      type: !include container-status.raml
+  startedAt?:
+    type: datetime-only
```
jasongilanfarr: datetime

docs/docs/rest-api/public/api/v2/types/pod-status.raml
jdef: ack

docs/docs/rest-api/public/api/v2/types/pod-status.raml
jdef: i didn't realize that it worked that way. my interpretation was that "no, you inventive user, you can't add arbitrary props to an object that we're trying to parse"

docs/docs/rest-api/public/api/v2/types/pathId.raml
```diff
@@ -0,0 +1,4 @@
+#%RAML 1.0 DataType
+type: string
+pattern: ^(\/?((\.{2})|([a-z0-9][a-z0-9\-.]*[a-z0-9]+)|([a-z0-9]*))($|\/))+$
```
jdef: not sure why we allow segments that contain `..` @meichstedt is there some history here?

docs/docs/rest-api/public/api/v2/types/pod-status.raml
jasongilanfarr: It means that too. But who cares?

docs/docs/rest-api/public/api/v2/types/pod-status.raml
jdef: you want it to be optional? or you don't like the name? i'm open to naming suggetions

docs/docs/rest-api/public/api/v2/types/pod-status.raml
jdef: I tried to capture that in DEGRADED by referencing "stable pod instances" in the description. Or do you think this overloads the meaning of DEGRADED?

I also don't know what @unterstein and @meichstedt have cooked up for instance states. could influence what we have here

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
jdef: ack

jdef: actually, each container can have a health check. so a single pod instance can have multiple health checks. STABLE pod instances have containers with passing (or none at all) health checks. DEGRADED pod instances have at least one container w/ a failing health check.

i'm happy to change the name to UNHEALTHY if this is a naming thing

docs/docs/rest-api/public/api/v2/types/container-status.raml
```diff
@@ -0,0 +1,56 @@
+#%RAML 1.0 DataType
+type: object
+additionalProperties: false
+properties:
+  name:
+    type: !include name.raml
+    description: name of the container specification (within the pod)
+  status:
+    type: string
+    enum: [ STAGING, STABLE, DEGRADED, TERMINAL ]
```
meichstedt: Thinking about the enum values we can/should have here. If both tasks and podInstances are abstracted to a generic `Instance` internally, then it might be nice to have the same states on both (or a generic subset).

Yet we could as well publish distinct events on API level.

I think I like `STABLE` vs `DEGRADED` more than `Running` vs `Running` (_but unhealthy_) as we have for tasks 👍

docs/docs/rest-api/public/api/v2/types/container-status.raml
```diff
@@ -0,0 +1,56 @@
...
+    description: |
+      STAGING  - Container has been launched but is not yet running.
+      STABLE   - Container is running and, if health checks were specified, is healthy.
+      DEGRADED - Container is running but its health check is not passing.
```
meichstedt: What would the status be if the tasks of a podInstance are Unreachable (LOST for now)? I think we need an `UNREACHABLE` here as well.

docs/docs/rest-api/public/api/v2/types/container-status.raml
meichstedt: Oh I just see this is the container-status

docs/docs/rest-api/public/api/v2/types/container-status.raml
jdef: maybe we should add a "reasonCode" or something. So that we can mark a
container as DEGRADED with reasonCode=unreachable?

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
```diff
@@ -0,0 +1,62 @@
...
+additionalProperties: false
+properties:
+  id:
+    type: !include name.raml
+    description: |
+      Unique ID of this pod instance in the cluster.
+      TODO(jdef) Probably represents the Mesos executor ID.
+  status:
+    type: string
+    enum: [ PENDING, STAGING, STABLE, DEGRADED, TERMINAL ]
```
meichstedt: As noted above, the plan for the internal model is to abstract (AppDefinition) task instances and (PodDefinition) instances into one `Instance`. A task can have the following states:
```
Reserved
Created
Error
Failed
Finished
Killed
Killing
Running
Staging
Starting
Unreachable
Gone
Dropped
Unknown
```

For the sake of aligning things in v2, what about simply mapping task status to instance status? @unterstein made plan of how status are aggregated into an aggregate status of a pod instance (e.g., the instance is staging as long as any task is staging, and when all are running, the instance turns running).

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
```diff
@@ -0,0 +1,62 @@
...
+          type: string[]
+  containers?:
+    type: array
+    description: status for each running container of this instance.
+    items:
+      type: !include container-status.raml
+  startedAt?:
+    type: datetime
+    description: time at which this instance was marked PENDING
+  terminatedAt?:
```
meichstedt: for tasks in v2 and the initial draft for v2, @aquamatthias and I were thinking about having only one timestamp, which is the `since` when the instance is in the current status. Not sure how much value `startedAt` has, once the instance is running, but it definitely has value to know _since when_ it is unhealthy, unreachable, or terminating.

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
jdef: Assuming that you mean "mapping task status to [container] status" because
there's already a plan to aggregate the individual container statuses into
a pod-instance status. If that's the case ... what's the plan? Can you
point me to a PR?

unterstein: Like @meichstedt said, I would introduce a PODStatus with this 14 status and pod status change on defined task status changes. Some POD status changes are triggered if all tasks are in a defined status and some POD status changes are triggered if one task has a certain status.

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
unterstein: +1 `since`

jasongilanfarr: I think this should only be in the status and we should one have one.

jdef: what do you mean "only in the status"?

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
```diff
@@ -9,10 +10,8 @@ properties:
       TODO(jdef) Probably represents the Mesos executor ID.
   status:
     type: string
-    usage: |
-      TODO(jdef) this is not an enum because it reflects Marathon internal state that
-      could easily change going forward and I don't want breaking API changes that result
-      from a shifting enum here.
+    (scalaType): PodInstanceState
+    enum: [ PENDING, STAGING, STABLE, DEGRADED, TERMINAL ]
```
jdef: Did you intentionally add these back? I had removed them because I thought we were just going to dump the string value of the internal pod instance status here (and I didn't want to reflect the internal states as an API enum)

jdef:
merged into `feature/pods` branch see commit https://github.com/mesosphere/marathon/commit/4464a1d3ed6750fa15bf8c0f6ee6798e34ead821

PR #4254, opened 2016-08-25
Remove some more globals from test.

jasongilanfarr:


src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -455,7 +455,8 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
 
   test("Create a new app (that uses secrets) successfully") {
     Given("The secrets feature is enabled")
-    AllConf.withTestConfig(Seq("--enable_features", "secrets"))
+    configArgs = Seq("--enable_features", "secrets")
+    resetAppsResource
```
jeschkies: Hm, does `configArgs` have to be mutable? It's probably too much for this diff but it feels that `configArgs`, `appsResource` etc are fixtures more or less unique to each test case.

jasongilanfarr: Almost everything in this class is mutable.

jasongilanfarr: Refactoring it into a fixture wasn't worth the effort.

jeschkies: Kind of hat I meant. It's probably too much effort to refactor this.

jdef:
:+1: :+1: :+1:

Issue #4191, opened 2016-08-08
Kill task of app in deployment does not work when scale=false


fengyehong:
### Reproduce Steps
1. Make an app stay in deployment, in my case I have a cluster with 3 mesos slaves, I set constraint [hostname, UNIQUE], and scale to 5, so it has 3 tasks running and stays in "Waiting"
2. Kill one of the running task by `DELETE /v2/apps/tasks/<taskid>` with scale=false
### Expected Result
1. the task get killed or error returned
### Real Result
1. the API returned 200 and the task to 'killed'
2. but the task is not killed in fact
3. setting parameter force=true or force=false not make a difference

I think the scheduler actor just failed to acquire the lock and didn't not check the result:
https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala#L162

unterstein:
Hi @fengyehong ,

which version of marathon are you using? Can you post the relevant marathon logs?

Thanks
Johannes

fengyehong:
I'm using 1.1.1

When setting logging level to info, I got only one line log indicating receiving the kill request :(
```
2016-08-23 10:57:34,969  INFO 10.1.243.147 - - [23/Aug/2016:10:57:34 +0000] "DELETE //mesos-master-dev042-cqdx.qiyi.virtual:8080/v2/apps/tt/tasks/tt.d573304c-691e-11e6-b254-525400fd9fbc HTTP/1.1" 200 349 "-" "HTTPie/0.9.2" 5 [mesosphere.chaos.http.ChaosRequestLog:16]
```
All other logs are offer processing related.

unterstein:
I retested this locally and the API behave in a non expected way. This is because the TaskKillActor told to kill the tasks instead of being asked and the result is not used to calculate the HTTP result, see https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala#L154.

The reason the deletion is failing, because the task is currently locked through the other deployment (see https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala#L168).

The correct status code in this situation would be something like ```423 Locked (WebDAV; RFC 4918)```.

Thanks for reporting this issue! I`ll move this to backlog, but maybe someone wants to provide a PR :)

Cheers
Johannes

jeschkies:
[...]

PR #4257, opened 2016-08-26
Report error in `killTasks`.

jeschkies:
This relates to #4191. However, I couldn't test the reported case.

There are a few comments on code I'm unsure about.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
+++ b/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
@@ -12,6 +12,7 @@ import mesosphere.marathon.upgrade.DeploymentPlan
 import mesosphere.marathon.{ MarathonConf, MarathonSchedulerService, UnknownAppException }
 import org.slf4j.LoggerFactory
 
+import scala.async.Async.{ async, await }
 import scala.concurrent.{ ExecutionContext, Future }
 
 class TaskKiller @Inject() (
@@ -39,12 +40,13 @@ class TaskKiller @Inject() (
           val foundTasks = findToKill(allTasks)
           val expungeTasks = if (wipe) expunge(foundTasks) else Future.successful(())
 
-          expungeTasks.map { _ =>
+          async {
+            await(expungeTasks)
             val launchedTasks = foundTasks.filter(_.launched.isDefined)
-            if (launchedTasks.nonEmpty) {
-              service.killTasks(appId, launchedTasks)
-              foundTasks
-            } else foundTasks
+            if (launchedTasks.nonEmpty) await(service.killTasks(appId, launchedTasks))
+            // For some reasons tasks returned by service.killTasks !=
+            // foundTasks in some cases. What should be returned?
+            foundTasks
           }
         }
 
--- a/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
+++ b/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
@@ -46,7 +46,9 @@ class TaskKiller @Inject() (
 
           val launchedTasks = foundTasks.filter(_.launched.isDefined)
           if (launchedTasks.nonEmpty) await(service.killTasks(appId, launchedTasks))
-          // Return killed *and* expunged tasks
+          // Return killed *and* expunged tasks.
+          // The user only cares that all tasks won't exist eventually. That's why we send all tasks back and not just
+          // the killed tasks.
           foundTasks
         }
 
--- a/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
+++ b/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
@@ -2,6 +2,7 @@ package mesosphere.marathon.api
 
 import javax.inject.Inject
 
+import akka.Done
 import com.twitter.util.NonFatal
 import mesosphere.marathon.core.group.GroupManager
 import mesosphere.marathon.core.task.{ Task, TaskStateOp }
@@ -35,18 +36,18 @@ class TaskKiller @Inject() (
       case Some(app) =>
         checkAuthorization(UpdateRunSpec, app)
 
-        // We probably want to use the same execution context as the actors.
+        // TODO: We probably want to pass the execution context as an implcit.
         import scala.concurrent.ExecutionContext.Implicits.global
 
         async {
           val allTasks = await(taskTracker.appTasks(appId))
           val foundTasks = findToKill(allTasks)
 
-          if (wipe) await(expunge(foundTasks))
+          if (wipe) expunge(foundTasks)
 
           val launchedTasks = foundTasks.filter(_.launched.isDefined)
           if (launchedTasks.nonEmpty) await(service.killTasks(appId, launchedTasks))
-          // For some reasons tasks returned by service.killTasks !=
+          // TODO: For some reasons tasks returned by service.killTasks !=
           // foundTasks in some cases. What should be returned?
           foundTasks
         }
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -34,32 +36,40 @@ class TaskKiller @Inject() (
...
+
+        val processNextTask = stateOpProcessor.process(TaskStateOp.ForceExpunge(nextTask.taskId))
+          .map(_ => Done)
+          .recover {
+            case NonFatal(cause) =>
+              log.info("Failed to expunge {}, got: {}", Array[Object](nextTask.taskId, cause): _*)
+              Done // Note, this error is swallowed.
+          }
+
+        await(processNextTask)
```
jeschkies: This is supposed to make it more explicit that we purge sequentially and not in parallel.

src/test/scala/mesosphere/marathon/test/CaptureLogEvents.scala
```diff
@@ -21,6 +21,12 @@ object CaptureLogEvents {
     private[this] def rootLogger: Logger = LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME).asInstanceOf[Logger]
 
     def appendToRootLogger(): Unit = {
+      // Waiting for Logback to be ready.
+      // See https://gist.github.com/mbknor/34944ea4589a5fc6974c
+      // Should actually have been fixed with http://jira.qos.ch/browse/SLF4J-167
+      while (LoggerFactory.getILoggerFactory.isInstanceOf[org.slf4j.helpers.SubstituteLoggerFactory]) {
+        Thread.sleep(50)
```
jeschkies: Not happy about this busy waiting.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -34,32 +36,40 @@ class TaskKiller @Inject() (
       case Some(app) =>
         checkAuthorization(UpdateRunSpec, app)
 
+        // TODO: We probably want to pass the execution context as an implcit.
         import scala.concurrent.ExecutionContext.Implicits.global
-        taskTracker.appTasks(appId).flatMap { allTasks =>
+
+        async {
```
unterstein: Why using async await and not just asking the actor?

```
     taskTracker.appTasks(appId).flatMap { allTasks =>
          val foundTasks = findToKill(allTasks)

          if (wipe) expunge(foundTasks)

          val launchedTasks = foundTasks.filter(_.launched.isDefined)
          if (launchedTasks.nonEmpty) service.killTasks(appId, launchedTasks).map(killed => killed.tasks)
          else Future.successful(Seq.empty)
        }
```

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -34,32 +36,40 @@ class TaskKiller @Inject() (
...
-              foundTasks
-            } else foundTasks
-          }
+
+          if (wipe) expunge(foundTasks)
+
+          val launchedTasks = foundTasks.filter(_.launched.isDefined)
+          if (launchedTasks.nonEmpty) await(service.killTasks(appId, launchedTasks))
+          // TODO: For some reasons tasks returned by service.killTasks !=
+          // foundTasks in some cases. What should be returned?
```
unterstein: That`s correct IMO. Only killed tasks should be returned.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -34,32 +36,40 @@ class TaskKiller @Inject() (
...
       case None => Future.failed(UnknownAppException(appId))
     }
   }
 
-  private[this] def expunge(tasks: Iterable[Task])(implicit ec: ExecutionContext): Future[Unit] = {
-    tasks.foldLeft(Future.successful(())) { (resultSoFar, nextTask) =>
-      resultSoFar.flatMap { _ =>
+  private[this] def expunge(tasks: Iterable[Task])(implicit ec: ExecutionContext): Unit = {
+    tasks.foreach { nextTask =>
+      async {
```
unterstein: Why using async await?

src/test/scala/mesosphere/marathon/api/TaskKillerTest.scala
```diff
@@ -137,7 +164,8 @@ class TaskKillerTest extends MarathonSpec
     when(f.tracker.appTasks(appId)).thenReturn(Future.successful(tasksToKill))
     when(f.stateOpProcessor.process(stateOp1)).thenReturn(Future.successful(TaskStateChange.Expunge(runningTask)))
     when(f.stateOpProcessor.process(stateOp2)).thenReturn(Future.successful(TaskStateChange.Expunge(reservedTask)))
-    when(f.service.killTasks(appId, launchedTasks)).thenReturn(launchedTasks)
+    when(f.service.killTasks(appId, launchedTasks))
+      .thenReturn(Future.successful(MarathonSchedulerActor.TasksKilled(appId, launchedTasks)))
```
unterstein: Should return only launched tasks (only launched tasks will be killed)

src/test/scala/mesosphere/marathon/test/CaptureLogEvents.scala
unterstein: ouh, I understand why you are unhappy :-). Whats happening there?

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: I want to make it explicit that the tasks are not processed in parallel but rather sequentially.

unterstein: Was this a requirement?

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: Then I have to fix it.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: I talked to Jason about Scala Async and agreed that it has advantages over nested flatMaps or Scala's for comprehensions. E.g. it easier to reason about and avoids eird things like Future.sucessful(...) or the map.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
unterstein: Would suggest something like
```
if (launchedTasks.nonEmpty) await(service.killTasks(appId, launchedTasks)).map(killed => killed.tasks)
else Seq.empty
```

Maybe @meichstedt will add his opinion :)

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: The original code is and Jason said that there have been some limitations.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: This change lets `TaskKillerTest "kill with wipe will kill running and expunge all"` fail. The expected return value is all killed tasks *and* all expunged tasks.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: I reverted the change of the expunge method. My change was not completely sequential.

jeschkies:
Added a small comment. @aquamatthias @meichstedt Could please verify.

meichstedt:
@jeschkies @aquamatthias @unterstein tl;dr: LGTM.

Long version: Killing a resident task will kill the running Mesos task, but not expunge its metadata (because we might eventually want to launch a new task using the same reservations/persistent volumes). That's what the `wipe` flag is there for: if `wipe==true`, the metadata will be expunged _and_ the task killed if it's launched (a workaround because there's no StorageManagement providing this functionality). It is definitely possible that a task is found, expunged but currently not launched, so there won't be a kill request to Mesos. Yet, since the user passed its ID and asked to wipe the metadata, she should receive the taskId as a kind of ACK.

meichstedt:
@jeschkies please rebase

jeschkies:
@meichstedt I guess we can merge now : )

meichstedt:
👍

PR #4261, opened 2016-08-26
Used a Map to store sub-groups as opposed to a Set.

jmlvanre:


jasongilanfarr:
LGTM. :+1:

jdef:
lgtm

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -17,7 +17,7 @@ import scala.collection.JavaConversions._
 case class Group(
     id: PathId,
     apps: Map[AppDefinition.AppKey, AppDefinition] = defaultApps,
-    groups: Set[Group] = defaultGroups,
+    groupsById: Map[Group.GroupKey, Group] = defaultGroups,
```
aquamatthias: Both apps and groups are available via id. Shouldn't we have the same scheme? (rename apps to appsById and have  a method to get apps)

aquamatthias:
@jmlvanre Thanks looks good. Please rebase.

src/main/scala/mesosphere/marathon/state/Group.scala
jmlvanre: I agree. I think this should be a separate follow up commit.

jasongilanfarr:
retest this please

PR #4267, opened 2016-08-29
WIP: feature/pods

jdef:
TODO:
- API Definition
  - [x] /v2/groups API w/ pods support
    - pods will be reported via the groups API (as a pod-status type) but will be read-only
    - groups API is written in older RAML, needs updating for compliance with RAML v1
    - pods will not play a role in dependencies; if they're placed into groups for which dependencies are declared from/to the pods will not be considered when calculating dependencies among groups/apps
    - ?taskKillGracePeriodSeconds is part of app.def, make it part of pods? #4341 
  - [x] RAML cleanup
    - we use `additionalProperties: false` in some places, not others; pick a consistent usage and stick to it
  - [x] Client support
    - introduce support for `HEAD` to `/v2/pods/` so that clients can query marathon to determine if it supports the pods API #4276
- Internals
  - [x] API endpoint impl
  - [x] add pods to groups
  - [x] pods persistence
  - [ ] handle task status updates for pods
  - [x] GC for pods
  - [ ] pod scheduling
  - networking
    - ~~support NONE network mode (`networks = []`)~~ (not supported by Mesos MVP)
- Plugins?
  - [ ] pod validation: probably need/want custom validation for things like secrets


docs/docs/rest-api/public/api/v2/schema/Group.json
```diff
@@ -23,6 +23,13 @@
         "$ref": "./AppDefinition.json"
       }
     },
+    "pods" : {
+      "type": "array",
+      "description": "The list of PodDefinitions in this group. See PodDefinition.json for the schema.",
+      "items": {
+        "$ref": "./PodDefinition.json"
```
jdef: we'll probably migrate group.json to RAML format and here we'll reference the pod-status RAML type instead

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
```diff
@@ -0,0 +1,66 @@
...
+properties:
+  id:
+    type: !include name.raml
+    description: |
+      Unique ID of this pod instance in the cluster.
+      TODO(jdef) Probably represents the Mesos executor ID.
+  status:
+    type: string
+    (scalaType): PodInstanceState
+    enum: [ PENDING, STAGING, STABLE, DEGRADED, TERMINAL ]
```
jdef: unclear how/why this was promoted to an enum again. need to sync w/ @jasongilanfarr

docs/docs/rest-api/public/api/v2/types/mesosContainer.raml
```diff
@@ -0,0 +1,64 @@
...
+(scalaType): MesosContainer
+description: Mesos Container
+properties:
+  name:
+    type: !include name.raml
+    description: The name of this container
+  exec?:
+    type: object
+    properties:
+      command:
```
jdef: shouldn't this be optional?

docs/docs/rest-api/public/api/v2/types/pod-instance-status.raml
jdef: resolved. we're keeping this enum. I updated the desc to match

jdef:
@jasongilanfarr looks like the RAML generator needs to escape `$` in comments (velocity builds are complaining)
```
[error] /marathon/target/scala-2.11/src_managed/main/mesosphere/marathon/raml/AppHealthCheck.scala:23: Variable PORT1 undefined in comment for class AppHealthCheck in class AppHealthCheck
[error]  *   port environment variables like $PORT1.
[error]                                       ^
[error] /marathon/target/scala-2.11/src_managed/main/mesosphere/marathon/raml/App.scala:44: Variable app.cmd undefined in comment for class App in class App
[error]  *   `/bin/sh -c ${app.cmd}`.
[error]                   ^
```

src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
```diff
@@ -47,7 +51,60 @@ class PodsResourceTest extends MarathonSpec with Matchers with Mockito {
...
+    response.getMetadata().containsKey(PodsResource.DeploymentHeader) should be(true)
+  }
+
+  test("update a simple single-container pod from docker image w/ shell command") {
+    implicit val podSystem = mock[PodManager]
+    val f = Fixture.create()
+
+    podSystem.update(any, eq(false)).returns(Future.successful(DeploymentPlan.empty))
+
+    val postJson = """
```
jasongilanfarr: You could use ```Pod()``` and have a helper: ```implicit def podToString(pod: Pod): String = Json.stringify(Json.toJson(pod))``` then it'd be a bit easier to read (you could also have it convert to bytes or whatever too if you needed, pretty sure ```Json``` has that too

jdef:
retest this please

docs/docs/rest-api/public/api/v2/types/volumes.raml
```diff
@@ -0,0 +1,70 @@
...
+  strings: stringTypes.raml
+types:
+  Volume:
+    type: object
+    properties:
+      name:
+        type: strings.Name
+        description: The name of the volume to reference.
+      host?:
+        type: string
```
jdef: using a simple type here makes it harder to add host-volume properties later. consider making this an object type with `path` property

jasongilanfarr:
merged!

PR #4268, opened 2016-08-29
Convert App and Group JsonSchemas to RAML and do a big cleanup of all the POD RAML too.

jasongilanfarr:
This actually matches and PODs is closer to app (env vars/labels in particular)

- [ ] Update the RAML Generator.

---
This change is [Reviewable](https://reviewable.io/reviews/mesosphere/marathon/4268)


docs/docs/rest-api/public/api/v2/types/app.raml
```diff
@@ -0,0 +1,318 @@
...
+    type: number
+    description: |
+      How much disk space is needed for this application.
+      This number does not have to be an integer, but can be a fraction.
+    format: double
+  env?:
+    type: object
+    properties:
+      # TODO: How do we get secretRef.raml in here?
+      /.*/: string
```
jdef: probably want the type of the properties to be `[ string | secretRef ]` or something like that

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
```diff
@@ -1,712 +0,0 @@
-{
```
aquamatthias: The json schema should be superseded by the RAML doc. We currently make the json schema available: https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/v2/SchemaResource.scala#L22

Perhaps we can create the schema from raml? (e.g.https://www.npmjs.com/package/raml-jsonschema-expander)

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -1,375 +1,355 @@
...
-                  {
-                    "path": "/upgradeStrategy/minimumHealthCapacity",
-                    "errors": [
-                      "is greater than 1"
-                    ]
-                  }
-                ]
-              }
+            #example: |
+            #    {
```
aquamatthias: Is it necessary to use a hash on every line?

aquamatthias:
We currently bundle the api-console, which is available via `/help` which does not understand raml 1.0. We need to update the console.

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
aquamatthias: It is also referenced in api.raml

aquamatthias:
`api.raml` is still 0.8

docs/docs/rest-api/public/api/v2/types/group.raml
```diff
@@ -0,0 +1,47 @@
...
+  apps?:
+    type: array
+    items: !include app.raml
+  pods?:
+    type: array
+    items: !include pod-status.raml
+  groups?:
+    type: array
+    items:
+      # TODO: unclear how to define a recursive type
```
aquamatthias: types.raml
```
#%RAML 1.0 Library
types:
  Group: !include ./group.raml
```

group.raml
```
groups?:
   type: Group
```

aquamatthias:
Made a couple of comments, but otherwise looks good.

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
jdef: Why would we want to support both at the same time? As part of some "deprecation" period for the schema?

jasongilanfarr:
Review status: 0 of 15 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 122 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4g8-r1-122:-KQQw8r4ji6dkzivc2qw:b-e9uvuu) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/apps.raml#L122)):*
<details><summary><i>Previously, aquamatthias (Matthias Veit) wrote…</i></summary>
> Is it necessary to use a hash on every line?
</details>
I haven't figured out how to get the examples to validate. I probably should have added a TODO. I don't know if yaml supports block comments.

---

*[docs/docs/rest-api/public/api/v2/types/app.raml, line 113 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4gD-r1-113:-KQQwgXpWa918Z1S5x0a:b7twv89) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/types/app.raml#L113)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> probably want the type of the properties to be `[ string | secretRef ]` or something like that
</details>
Yeah, I couldn't figure out how to do that :( The code generator will need to have a special hook for this to have /.*/ become a map.

---

*[docs/docs/rest-api/public/api/v2/types/group.raml, line 23 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4gI-r1-23:-KQQwq8CU14nDilbf8CT:b-zb10bs) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/types/group.raml#L23)):*
<details><summary><i>Previously, aquamatthias (Matthias Veit) wrote…</i></summary>
> types.raml
> 
> ```
> 
> #%RAML 1.0 Library
> 
> types:
> 
>   Group: !include ./group.raml
> 
> ```
> 
> 
> 
> group.raml
> 
> ```
> 
> groups?:
> 
>    type: Group
> 
> ```
> 
> 
</details>
I wonder if that will actually work?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4268)*

jdef:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 63 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQRXUTQInmiIX_RhCN9:-KQRXUTQInmiIX_RhCNA:bqpwmos) ([raw file](https://github.com/mesosphere/marathon/blob/023053ce99a751735313329e23be68de70a6bb92/docs/docs/rest-api/public/api/v2/apps.raml#L63)):*
> ```RAML
> 
>       enum: [ apps.tasks, apps.count, apps.deployments, apps.lastTaskFailure, apps.failures, apps.taskStats ]
>       repeat: true
> ```

instead of simply deleting this line, shouldn't embed be an array type, where item item could be an app.tasks, app.count, apps.deployments, ... etc ?

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 216 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQRXkrZFfdDBMP-AXdC:-KQRXkrZFfdDBMP-AXdD:bh3869s) ([raw file](https://github.com/mesosphere/marathon/blob/023053ce99a751735313329e23be68de70a6bb92/docs/docs/rest-api/public/api/v2/apps.raml#L216)):*
> ```RAML
> 
>         enum: [ app.tasks, app.count, app.deployments, app.lastTaskFailure, app.failures, app.taskStats ]
>         repeat: true
> ```

same comment for this `embed` as for the prior, should probably be modeled as an array of items, each item can take the value of an enum

---

*[docs/docs/rest-api/public/api/v2/types/app.raml, line 113 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4gD-r1-113:-KQRYDSUzv0Yg89eCpQj:b-393ylu) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/types/app.raml#L113)):*
<details><summary><i>Previously, jasongilanfarr (Jason Gilanfarr) wrote…</i></summary>
> Yeah, I couldn't figure out how to do that :( The code generator will need to have a special hook for this to have /.*/ become a map.
</details>
It's hard when we have all the types split into separate datatype files. If we had lib-based modules then stuff like this is easier because you can share types between libs and use formal type names instead of `!include xml.raml` everywhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4268)*

aquamatthias:
Review status: 0 of 15 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 63 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQRXUTQInmiIX_RhCN9:-KQVdZbTZX1zJk6W_IVS:b-1zm7g8) ([raw file](https://github.com/mesosphere/marathon/blob/023053ce99a751735313329e23be68de70a6bb92/docs/docs/rest-api/public/api/v2/apps.raml#L63)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> instead of simply deleting this line, shouldn't embed be an array type, where item item could be an app.tasks, app.count, apps.deployments, ... etc ?
</details>
In the api you have to supply this parameter multiple times:
`embed=apps.deployments&embed=apps.lastTaskFailure`
I would argue, that `repeat: true` is correct here.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 122 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4g8-r1-122:-KQVd6Oh7gbZ8JLjOwU2:bbzl339) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/apps.raml#L122)):*
<details><summary><i>Previously, jasongilanfarr (Jason Gilanfarr) wrote…</i></summary>
> I haven't figured out how to get the examples to validate. I probably should have added a TODO. I don't know if yaml supports block comments.
</details>
I guess my question is more: why do we need to make this a comment?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4268)*

docs/docs/rest-api/public/api/v2/types/mesosContainer.raml
```diff
@@ -1,64 +1,82 @@
...
+#%RAML 1.0 Library
+uses:
+  artifacts: artifact.raml
+  env: environmentVariable.raml
+  health: healthCheck.raml
+  labels: label.raml
+  mesosCommand: mesosCommand.raml
+  network: network.raml
+  resources: resources.raml
+  secrets: secrets.raml
```
jdef: actually let's just declare these at the pod level. v2-style envvvars in the container can reference them from there

docs/docs/rest-api/public/api/v2/types/environmentVariable.raml
```diff
@@ -1,19 +1,19 @@
-#%RAML 1.0 DataType
-type: object
-(scalaType): EnvVar
-maxProperties: 2
-additionalProperties: false
-properties:
-  name:
+#%RAML 1.0 Library
+uses:
+  strings: stringTypes.raml
```
jdef: don't see it used in this lib

jasongilanfarr:
Review status: 0 of 41 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 63 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQRXUTQInmiIX_RhCN9:-KQaiBKgsK-cckaYxNkB:b-iqwhz5) ([raw file](https://github.com/mesosphere/marathon/blob/023053ce99a751735313329e23be68de70a6bb92/docs/docs/rest-api/public/api/v2/apps.raml#L63)):*
<details><summary><i>Previously, aquamatthias (Matthias Veit) wrote…</i></summary>
> In the api you have to supply this parameter multiple times:
> `embed=apps.deployments&embed=apps.lastTaskFailure`
> I would argue, that `repeat: true` is correct here.
</details>
repeat: true isn't supported in RAML 1.0.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 122 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4g8-r1-122:-KQaiFkD-HlWavJEL0QZ:b-7srhxh) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/apps.raml#L122)):*
<details><summary><i>Previously, aquamatthias (Matthias Veit) wrote…</i></summary>
> I guess my question is more: why do we need to make this a comment?
</details>
I fixed it.

---

*[docs/docs/rest-api/public/api/v2/apps.raml, line 216 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQRXkrZFfdDBMP-AXdC:-KQaiHSkXbcIMhCsOkkq:b-yey6bq) ([raw file](https://github.com/mesosphere/marathon/blob/023053ce99a751735313329e23be68de70a6bb92/docs/docs/rest-api/public/api/v2/apps.raml#L216)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> same comment for this `embed` as for the prior, should probably be modeled as an array of items, each item can take the value of an enum
</details>
yes, I haven't figured out how to do that yet. repeat: true isn't supported in raml 1.0 and i didn't spend that much time on the API endpoint RAMLs, just the datatypes.

---

*[docs/docs/rest-api/public/api/v2/types/app.raml, line 113 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4gD-r1-113:-KQaiR9lIlL1_SfEp9go:b-6c5eqr) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/types/app.raml#L113)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> It's hard when we have all the types split into separate datatype files. If we had lib-based modules then stuff like this is easier because you can share types between libs and use formal type names instead of `!include xml.raml` everywhere.
</details>
Fixed that and its now there.

---

*[docs/docs/rest-api/public/api/v2/types/environmentVariable.raml, line 3 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQWhjU32HyeJ1_1orzo-r3-3:-KQaiTUr91U_fD3hNknL:b-wq7azj) ([raw file](https://github.com/mesosphere/marathon/blob/d6d06265ca4c6139cde687d36b53504b8f23faec/docs/docs/rest-api/public/api/v2/types/environmentVariable.raml#L3)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> don't see it used in this lib
</details>
harmless, but i'll remove

---

*[docs/docs/rest-api/public/api/v2/types/group.raml, line 23 \[r1\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQQvpYUbzspmIevk4gI-r1-23:-KQaiVj7LJACbL8Hq03r:bxnrg5p) ([raw file](https://github.com/mesosphere/marathon/blob/2fa5475e8ecf3a3525d91aae8f5404f4c353ebdd/docs/docs/rest-api/public/api/v2/types/group.raml#L23)):*
<details><summary><i>Previously, jasongilanfarr (Jason Gilanfarr) wrote…</i></summary>
> I wonder if that will actually work?
</details>
turns out it does

---

*[docs/docs/rest-api/public/api/v2/types/mesosContainer.raml, line 10 \[r3\]](https://reviewable.io:443/reviews/mesosphere/marathon/4268#-KQWhjU32HyeJ1_1orzt-r3-10:-KQaiYLbCoXU1mGEIVav:b32wtd) ([raw file](https://github.com/mesosphere/marathon/blob/d6d06265ca4c6139cde687d36b53504b8f23faec/docs/docs/rest-api/public/api/v2/types/mesosContainer.raml#L10)):*
<details><summary><i>Previously, jdef (James DeFelice) wrote…</i></summary>
> actually let's just declare these at the pod level. v2-style envvvars in the container can reference them from there
</details>
ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4268)*

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -1,150 +1,145 @@
...
-      Network settings are defined on a pod level. All containers share the same network stack.
-      At this time, only one stack is supported.
-  scaling?:
+      During an upgrade all instances of an application get replaced by a new version.
+      The upgradeStrategy controls how Marathon stops old versions and launches new versions.
+    properties:
+      minimumHealthCapacity?:
+        type: number
+        description: |
+          A number between 0and 1 that is multiplied with the instance count.
```
jdef: s/0and/0 and/

PR #4270, opened 2016-08-29
WIP: refactor pods scheduling policy

jdef:
@jasongilanfarr and I talked about trying to eliminate some of the nested `Option` fields generated from the RAML and that it might be possible to use type inheritance w/ discriminators to achieve this.

This PR does the following:
- refactor scaling types into a library so that they're formally named types, usable w/ discriminators
- refactor the `Pod` and `PodStatus` types into a library (and declare a dep on the `scaling` lib)
- mods to the pods REST api to use the `pod` library that now wraps the `Pod` and `PodStatus` types


jasongilanfarr:
Seems reasonable. I wonder how this affects the generator.

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -1,150 +1,205 @@
...
+      instances?:
+        type: array
+        items:
+          (scalaType): PodInstanceStatus
+          type: !include pod-instance-status.raml
+      terminationHistory?:
+        type: array
+        description: |
+          List of most recent instance terminations.
+          TODO(jdef) determine how many items might show up here; current thinking is .. not many
```
aquamatthias: We currently store the last task failure per app.
We probably should reflect the properties from `TaskFailure`

docs/docs/rest-api/public/api/v2/types/scaling-policy.raml
```diff
@@ -0,0 +1,30 @@
...
+    type: AbstractScalingPolicy
+    discriminatorValue: fixed
+    properties:
+      instances:
+        type: integer
+        format: int32
+        description: The number of instances of this pod to start.
+        minimum: 0
+        default: 1
+      maxInstances?:
```
jdef: The goal was to stop abusing the MARATHON_SINGLE_APP_INSTANCE label and use
a proper field

docs/docs/rest-api/public/api/v2/types/pod.raml
jdef: Keeping only the most recent failure seems rather ... limiting, but OK we could have a history of 1 here.

For those following along, TaskFailure has this useful information:
```
case class TaskFailure(
  appId: PathId,
  taskId: mesos.TaskID,
  state: mesos.TaskState,
  message: String = "",
  host: String = "",
  version: Timestamp = Timestamp.now,
  timestamp: Timestamp = Timestamp.now,
  slaveId: Option[mesos.SlaveID] = None)
```

jasongilanfarr:
I adapted this into https://github.com/mesosphere/marathon/pull/4268

PR #4277, opened 2016-08-30
Initial rev of pods API endpoints

jdef:
TODO:
- [ ] unit test

Deferred:
- remaining validation
- integration with groups
- integration with deployment

FWIW I really don't like how we bundle in all sorts of crazy functionality into one method. I'd much rather see the API endpoints composed of handlers and decorators. Instead we've guice'd the heck out of it and made each endpoint impl a mini-monolith.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/RestResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/RestResource.scala
@@ -1,6 +1,8 @@
 package mesosphere.marathon.api
 
 import java.net.URI
+import java.nio.charset.StandardCharsets
+import javax.servlet.http.HttpServletRequest
 import javax.ws.rs.core.Response
 import javax.ws.rs.core.Response.{ ResponseBuilder, Status }
 
@@ -9,8 +11,7 @@ import mesosphere.marathon.api.v2.json.Formats._
 import mesosphere.marathon.state.{ PathId, Timestamp }
 import mesosphere.marathon.upgrade.DeploymentPlan
 import play.api.libs.json.Json.JsValueWrapper
-import play.api.libs.json.{ Writes, Json }
-
+import play.api.libs.json.{ Json, Writes }
 import com.wix.accord._
 import mesosphere.marathon.api.v2.Validation._
 
@@ -72,5 +73,16 @@ trait RestResource {
       //scalastyle:on
       case Success => fn(t)
     }
+  }
+
+  def decodeBytes(data: Array[Byte], req: HttpServletRequest): String = {
+    val maybeEncoding = Option(req.getCharacterEncoding())
+    val charset = maybeEncoding match {
+      case Some(charsetName) =>
+        java.nio.charset.Charset.forName(charsetName)
+      case None =>
+        StandardCharsets.UTF_8 // TODO(jdef) should this be configurable somewhere?
+    }
+    new String(data, charset)
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
@@ -1,7 +1,6 @@
 package mesosphere.marathon.api.v2
 
 import java.net.URI
-import java.nio.charset.StandardCharsets
 import javax.inject.Inject
 import javax.servlet.http.HttpServletRequest
 import javax.ws.rs._
@@ -72,7 +71,7 @@ class PodsResource @Inject() (
           .created(new URI(createdPod.id.toString))
           .entity(marshalJson(createdPod.asPodDef))
 
-        deploymentId.foreach(did => builder.header(DEPLOYMENT_ID_HEADER, did))
+        deploymentId.foreach(id => builder.header(DEPLOYMENT_ID_HEADER, id))
         builder.build()
       }
     }(createPodValidator)
@@ -123,7 +122,7 @@ class PodsResource @Inject() (
           .ok(new URI(updatedPod.id.toString))
           .entity(marshalJson(updatedPod.asPodDef))
 
-        deploymentId.foreach(did => builder.header(DEPLOYMENT_ID_HEADER, did))
+        deploymentId.foreach(id => builder.header(DEPLOYMENT_ID_HEADER, id))
         builder.build()
       }
     }(createPodValidator and updatePodValidator)
@@ -146,7 +145,7 @@ class PodsResource @Inject() (
           .ok(new URI(deletedPod.id.toString))
           .entity(marshalJson(deletedPod.asPodDef))
 
-        deploymentId.foreach(did => builder.header(DEPLOYMENT_ID_HEADER, did))
+        deploymentId.foreach(id => builder.header(DEPLOYMENT_ID_HEADER, id))
         builder.build()
       }
     }(PodsValidation.idValidator)
@@ -241,17 +240,6 @@ protected object PodsResourceInternal {
         }.getOrElse(network)
       }
     )
-
-  def decodeBytes(data: Array[Byte], req: HttpServletRequest): String = {
-    val maybeEncoding = Option(req.getCharacterEncoding())
-    val charset = maybeEncoding match {
-      case Some(charsetName) =>
-        java.nio.charset.Charset.forName(charsetName)
-      case None =>
-        StandardCharsets.UTF_8 // TODO(jdef) should this be configurable somewhere?
-    }
-    new String(data, charset)
-  }
 
   def unmarshalJson[T: JsonReader](data: String): T = {
     data.parseJson.convertTo[T]
--- a/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
@@ -110,6 +110,7 @@ class PodsResource @Inject() (
 
     withValid(unmarshalJson(decodeBytes(body, req))) { podDef =>
 
+      // podDef.id has just been validated, requiring equality for id validates it
       require(id == podDef.id)
       val pod = PodDefinition(withDefaults(podDef, podDefaults)).withCanonizedIds()
 
--- a/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
@@ -1,14 +1,22 @@
 package mesosphere.marathon.api.v2
 
+import java.net.URI
+import java.nio.charset.StandardCharsets
 import javax.inject.Inject
 import javax.servlet.http.HttpServletRequest
 import javax.ws.rs._
 import javax.ws.rs.core.{ Context, MediaType, Response }
 
+import akka.event.EventStream
 import com.codahale.metrics.annotation.Timed
 import mesosphere.marathon.MarathonConf
 import mesosphere.marathon.api.{ AuthResource, MarathonMediaType, RestResource }
-import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer }
+import mesosphere.marathon.core.base.Clock
+import mesosphere.marathon.core.event._
+import mesosphere.marathon.core.pod.PodDefinition
+import mesosphere.marathon.plugin.auth._
+import mesosphere.marathon.raml.PodDef
+import spray.json._
 
 @Path("v2/pods")
 @Consumes(Array(MediaType.APPLICATION_JSON))
@@ -16,7 +24,10 @@ import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer }
 class PodsResource @Inject() (
     val config: MarathonConf,
     val authenticator: Authenticator,
-    val authorizer: Authorizer) extends RestResource with AuthResource {
+    val authorizer: Authorizer)(
+    implicit
+    val clock: Clock,
+    val eventBus: EventStream) extends RestResource with AuthResource {
 
   /**
     * HEAD is used to determine whether some Marathon variant supports pods.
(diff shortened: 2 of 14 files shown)


src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -1,22 +1,39 @@
...
 
 @Path("v2/pods")
 @Consumes(Array(MediaType.APPLICATION_JSON))
 @Produces(Array(MarathonMediaType.PREFERRED_APPLICATION_JSON))
 class PodsResource @Inject() (
     val config: MarathonConf,
     val authenticator: Authenticator,
-    val authorizer: Authorizer) extends RestResource with AuthResource {
+    val authorizer: Authorizer)(
+    implicit
```
jdef: does guice even support this? even if it does, is it worth being inconsistent with AppsResource (what do we really gain)?

jdef:
assigned to @jasongilanfarr for review

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +51,218 @@ class PodsResource @Inject() (
...
+
+      withAuthorization(CreateRunSpec, pod) {
+        val (createdPod, deploymentId) = podSystem.create(pod, force)
+        Events.maybePost(PodEvent(req.getRemoteAddr, req.getRequestURI, PodEvent.Created))
+
+        val builder = Response
+          .created(new URI(createdPod.id.toString))
+          .entity(marshalJson(createdPod.asPodDef))
+
+        deploymentId.foreach(did => builder.header(DEPLOYMENT_ID_HEADER, did))
```
jeschkies: I'd use `d_id` or write it out. I read "did" as in "do" ;).

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +51,218 @@ class PodsResource @Inject() (
...
+  def decodeBytes(data: Array[Byte], req: HttpServletRequest): String = {
+    val maybeEncoding = Option(req.getCharacterEncoding())
+    val charset = maybeEncoding match {
+      case Some(charsetName) =>
+        java.nio.charset.Charset.forName(charsetName)
+      case None =>
+        StandardCharsets.UTF_8 // TODO(jdef) should this be configurable somewhere?
+    }
+    new String(data, charset)
+  }
```
jeschkies: `decodeBytes` is not really pods specific, is it?

jdef: Nope, not pod-specific. Should be moved somewhere more common. I'm happy to
promote this into RestResource for now. OK?

jeschkies: I'd say so. I would not be surprised if there's not already a similar method somewhere :).

jdef: AppsResource uses play for json parsing, which appears to rely on
auto-detection of UTF variants based on the contents of the byte array
(doesn't consider request headers at all)

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jasongilanfarr: Implicits are supported by Guice. In a non-guice world, implicits are pretty phenomenal for "DI" and make testing less verbose.

src/main/scala/mesosphere/marathon/api/RestResource.scala
```diff
@@ -73,4 +74,15 @@ trait RestResource {
...
     }
   }
+
+  def decodeBytes(data: Array[Byte], req: HttpServletRequest): String = {
+    val maybeEncoding = Option(req.getCharacterEncoding())
+    val charset = maybeEncoding match {
+      case Some(charsetName) =>
+        java.nio.charset.Charset.forName(charsetName)
+      case None =>
+        StandardCharsets.UTF_8 // TODO(jdef) should this be configurable somewhere?
```
jasongilanfarr: The HTTP spec says we should use ISO-8859-1 if it is not specified. Given that UTF-8 is a strict super set, I think its the right choice. And no, it shouldn't be configurable.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+
+  // TODO(jdef) force parameter? probably since this supports deployment; but what does it actually
+  // mean to "force" a create?
+  @POST @Timed
+  def create(
+    body: Array[Byte],
+    @DefaultValue("false")@QueryParam("force") force: Boolean,
+    @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
+
+    withValid(unmarshalJson[PodDef](decodeBytes(body, req))) { podDef =>
```
jasongilanfarr: Hmm. I think we should always validate the internal model, not the json model. its also where normalization should live.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+  }
+
+  @PUT @Timed @Path("""{id:.+}""")
+  def update(
+    @PathParam("id") id: String,
+    body: Array[Byte],
+    @DefaultValue("false")@QueryParam("force") force: Boolean,
+    @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
+
+    withValid(unmarshalJson[PodDef](decodeBytes(body, req))) { podDef =>
```
jasongilanfarr: ditto above. I'd say we update first then validate

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+        }.getOrElse(network)
+      }
+    )
+
+  def unmarshalJson[T: JsonReader](data: String): T = {
+    data.parseJson.convertTo[T]
+  }
+
+  def marshalJson[T: JsonWriter](t: T): String = {
+    t.toJson.prettyPrint
```
jasongilanfarr: should be compact

src/main/scala/mesosphere/marathon/core/pod/PodModule.scala
```diff
@@ -0,0 +1,11 @@
+package mesosphere.marathon.core.pod
+
+import mesosphere.marathon.api.v2.PodsResource
+
+/**
```
jasongilanfarr: remove comment

src/main/scala/mesosphere/marathon/core/pod/PodModule.scala
```diff
@@ -0,0 +1,11 @@
...
+  * Created by jdefelice on 8/31/16.
+  */
+class PodModule {
+
+  val podSystem: PodsResource.System = new PodSubsystem()
```
jasongilanfarr: Why isn't PodSubsystem the module?

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -1,22 +1,41 @@
...
 import mesosphere.marathon.MarathonConf
+import mesosphere.marathon.api.v2.validation.PodsValidation
 import mesosphere.marathon.api.{ AuthResource, MarathonMediaType, RestResource }
-import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer }
+import mesosphere.marathon.core.base.Clock
+import mesosphere.marathon.core.event._
+import mesosphere.marathon.core.pod.PodDefinition
+import mesosphere.marathon.plugin.auth._
+import mesosphere.marathon.raml.{ Network, PodDef, PodStatus }
+import spray.json._
```
aquamatthias: Why are we use spray json here? 
I see 2 issues:
- we should try to have the same error reporting format in v2 
- having 2 json libraries in the same project is harder to deal with

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
   def capability(@Context req: HttpServletRequest): Response = authenticated(req) { _ =>
     ok()
   }
+
+  // TODO(jdef) force parameter? probably since this supports deployment; but what does it actually
```
aquamatthias: In /v2/apps you can create a deployment with multiple apps. `force=true` will kill this deployment and start a new one.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+  def create(
+    body: Array[Byte],
+    @DefaultValue("false")@QueryParam("force") force: Boolean,
+    @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
+
+    withValid(unmarshalJson[PodDef](decodeBytes(body, req))) { podDef =>
+
+      val pod = PodDefinition(withDefaults(podDef, podDefaults)).withCanonizedIds()
+
+      withAuthorization(CreateRunSpec, pod) {
```
aquamatthias: Thanks for adding authn+authz

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+          .entity(marshalJson(createdPod.asPodDef))
+
+        deploymentId.foreach(id => builder.header(DEPLOYMENT_ID_HEADER, id))
+        builder.build()
+      }
+    }(createPodValidator)
+  }
+
+  @GET @Timed
+  def findAll(@Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
```
aquamatthias: Is the information you get here always the same?
In `v2/apps` we use embed parameter to customize the response.

aquamatthias: Should we add the ability to filter pods?
We have label filters etc.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+  def update(
+    @PathParam("id") id: String,
+    body: Array[Byte],
+    @DefaultValue("false")@QueryParam("force") force: Boolean,
+    @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
+
+    withValid(unmarshalJson[PodDef](decodeBytes(body, req))) { podDef =>
+
+      // podDef.id has just been validated, requiring equality for id validates it
+      require(id == podDef.id)
```
aquamatthias: This should probably result in a 409/Conflict?

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+
+    withValid(id) { _ =>
+      val pod: PodDefinition = podSystem.find(id)
+      withAuthorization(DeleteRunSpec, pod) {
+
+        val (deletedPod, deploymentId) = podSystem.delete(id, force)
+        Events.maybePost(PodEvent(req.getRemoteAddr, req.getRequestURI, PodEvent.Deleted))
+
+        val builder = Response
+          .ok(new URI(deletedPod.id.toString))
```
aquamatthias: Is it common to define a Location for deleted entities?

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+          .entity(marshalJson(deletedPod.asPodDef))
+
+        deploymentId.foreach(id => builder.header(DEPLOYMENT_ID_HEADER, id))
+        builder.build()
+      }
+    }(PodsValidation.idValidator)
+  }
+
+  @GET @Timed @Path("""{id:.+}::status""")
```
aquamatthias: So the URL is `/v2/pods/path/to/pod::status`?
Have you checked that this path is not eaten by the `find` mapping?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -14,6 +14,7 @@ import scala.reflect.ClassTag
 import scala.util.Try
 import scala.util.matching.Regex
 
+// TODO(jdef) move this into package "validation"
```
aquamatthias: +1

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +50,207 @@ class PodsResource @Inject() (
...
+/**
+  * public interfaces, specific to the pods api, that other pieces of the system implement and/or interact with
+  */
+object PodsResource {
+
+  /**
+    * Pod operations that reach this interface have already been authenticated, authorized, and otherwise validated.
+    * TODO(jdef) does it really make sense that deployment ID's are returned an Option[String] instead of String?
+    */
+  trait System {
```
aquamatthias: I guess this is the interface you envisioned for the pods service - that will be moved somewhere else?

aquamatthias:
Added a couple of comments. 
One concern regarding spray - mainly because of the error handling reported to the user.
Otherwise looks good.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: Considered this too. However, validation of the internal model will lead to generating errors that don't necessarily make sense for people trying to map the error message to a JSON structure. Most especially where the PodDefinition internal structure differs from the v2 API structure.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: ack

src/main/scala/mesosphere/marathon/core/pod/PodModule.scala
jdef: It could be. I didn't want to make that assumption yet.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: We shifted from spray to play in the raml-10 branch for those reasons. Was waiting to rebase this PR once the raml-10 changes had landed, converting from spray to play

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: I'm not a fan of embed -- at all. Pretty sure that we should be avoiding injecting arbitrary objects into object generated by the API. Which is what embed seems to do. Filters .. yes they make sense.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: More like Bad Input or something. Because the ID param didn't match the ID supplied with the json

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: Good point. I don't see where it **hurts** to provide this information ("this is the pod that was deleted"). But if it's confusing then we should remove

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: Not yet

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: I actually like the "interface" defined here. An implemented by something else, somewhere else. It's the basis of the contract between the API and the underlying system. Plan was to mock this for unit tests.

jasongilanfarr:
While I go integrate group in, you can track progress here: https://github.com/mesosphere/marathon/compare/jdef_post_pods_endpoint...jason/pods-api

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
aquamatthias: require will throw an IllegalArgumentException, which translates to 422 with a message "requirement failed" which is not very clear

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
aquamatthias: Minor - I agree.

jdef:
changes here have been incorporated into #4314

PR #4284, opened 2016-09-01
Soften validation for PortValidations to allow `tcp,udp`

### LINKED ISSUE #4283: Port definitions protocol validation error
Marathon is returning a validation error for portDefinitions udp and tcp (`udp,tcp`). This should be changed that `tcp,udp` is allowed.

App Definition
```
{
  "id": "/test",
  "cmd": "while true; do sleep 10; done",
  "cpus": 1,
  "mem": 128,
  "disk": 0,
  "instances": 1,
  "executor": null,
  "fetch": null,
  "constraints": null,
  "acceptedResourceRoles": null,
  "user": null,
  "container": null,
  "labels": null,
  "healthChecks": null,
  "env": null,
  "portDefinitions": [
    {
      "protocol": "udp,tcp",
      "port": 0
    }
  ]
}
```
Response

```
{"message":"Object is not valid","details":[{"path":"/portDefinitions(0)/protocol","errors":["is not one of (tcp,udp)"]}]}
```

unterstein:
Fixes #4283 

@aquamatthias: Can`t find the place in code where `PortDefinitions` are mapped to `ServiceInfo`. Can you give me a hint?

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/serialization/PortDefinitionSerializer.scala
+++ b/src/main/scala/mesosphere/marathon/api/serialization/PortDefinitionSerializer.scala
@@ -6,18 +6,20 @@ import org.apache.mesos
 import scala.collection.JavaConverters._
 
 object PortDefinitionSerializer {
-  def toProto(portDefinition: PortDefinition): mesos.Protos.Port = {
-    val builder = mesos.Protos.Port.newBuilder
-      .setNumber(portDefinition.port)
-      .setProtocol(portDefinition.protocol)
+  def toProto(portDefinition: PortDefinition): Seq[mesos.Protos.Port] = {
+    portDefinition.protocol.split(',').map { protocol =>
+      val builder = mesos.Protos.Port.newBuilder
+        .setNumber(portDefinition.port)
+        .setProtocol(protocol)
 
-    portDefinition.name.foreach(builder.setName)
+      portDefinition.name.foreach(builder.setName)
 
-    if (portDefinition.labels.nonEmpty) {
-      builder.setLabels(LabelsSerializer.toMesosLabelsBuilder(portDefinition.labels))
+      if (portDefinition.labels.nonEmpty) {
+        builder.setLabels(LabelsSerializer.toMesosLabelsBuilder(portDefinition.labels))
+      }
+
+      builder.build
     }
-
-    builder.build
   }
 
   def fromProto(proto: mesos.Protos.Port): PortDefinition = {
--- a/src/main/scala/mesosphere/marathon/api/serialization/PortDefinitionSerializer.scala
+++ b/src/main/scala/mesosphere/marathon/api/serialization/PortDefinitionSerializer.scala
@@ -6,8 +6,17 @@ import org.apache.mesos
 import scala.collection.JavaConverters._
 
 object PortDefinitionSerializer {
-  def toProto(portDefinition: PortDefinition): Seq[mesos.Protos.Port] = {
-    portDefinition.protocol.split(',').map { protocol =>
+  def toProto(portDefinition: PortDefinition): mesos.Protos.Port = toProto(portDefinition, split = false).head
+
+  def toMesosProto(portDefinition: PortDefinition): Seq[mesos.Protos.Port] = toProto(portDefinition, split = true)
+
+  private def toProto(portDefinition: PortDefinition, split: Boolean): Seq[mesos.Protos.Port] = {
+    val protocols: Seq[String] = if (split) {
+      portDefinition.protocol.split(',')
+    } else {
+      Seq(portDefinition.protocol)
+    }
+    protocols.map { protocol =>
       val builder = mesos.Protos.Port.newBuilder
         .setNumber(portDefinition.port)
         .setProtocol(protocol)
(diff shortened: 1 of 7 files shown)


unterstein:
Talked to Benjamin B and he said that I must do two reservation to reserve `tcp` and `udp` on the same port.

jasongilanfarr:
The validation change is definitely correct and actually matches the json schema. Do we actually handle it correctly though?

aquamatthias:
@unterstein: 
Ports are offered as a RangeResource. So you can not reserve the same port for udp and tcp separately.
But the DiscoveryInfo has to be set accordingly. See https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/TaskBuilder.scala#L198. We probably need to do the same thing as the `PortMappingSerializer.toMesosPort` (we need one toProto to store in zk and toMesosProto to set the discoveryInfo correctly)

@gkleiman: please confirm

src/main/scala/mesosphere/marathon/state/PortDefinition.scala
```diff
@@ -14,7 +14,7 @@ case class PortDefinition(
 
 object PortDefinition {
   implicit val portDefinitionValidator = validator[PortDefinition] { portDefinition =>
-    portDefinition.protocol is oneOf(DiscoveryInfo.Port.AllowedProtocols)
+    portDefinition.protocol.split(',').toIterable is every(oneOf(DiscoveryInfo.Port.AllowedProtocols))
```
jdef: @jasongilanfarr should we update endpoints to define `protocol` as an array? or do the same thing that we're apparently doing for apps; parsing a string

aquamatthias: @jdef: Is there a term which describes tcp_and_udp? So we could use this as a separate enum value.

jdef: It's not clear to me why we restrict this field at all, if it's only used
for DI

aquamatthias: The protocol is not typed/enforced on the Mesos side - you are right.

jasongilanfarr: Given that we couldn't have supported "tcp,udp", I do think we should make it an array _or_ just add "all".

jasongilanfarr: If we changed it to an array, we'd potentially break compatibility unless we make it a union of either enum or array of enum.

jdef: huh? I'm talking about the upcoming pods RAML, suggesting that protocols
could be a string[] instead of a string. we shouldn't enforce an enum on
protocols since mesos doesn't (what would be the point of marathon limiting
the use of the field when mesos does not?)

unterstein: In portMappings we accept `tcp,udp` as string, see https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/state/Container.scala#L103

jasongilanfarr: Oh sorry, missed that. I was talking about "we can't reasonably change how
portDefs work in apps". Agree that we should allow a free form string for
pods.

unterstein:
retest this please

unterstein:
@aquamatthias @jasongilanfarr @jdef Is https://github.com/mesosphere/marathon/pull/4284/commits/2317be6f708a91040c3fe96245530eb6e458b8d9 as expected?

jdef:
that commit lgtm

src/main/scala/mesosphere/marathon/api/serialization/PortDefinitionSerializer.scala
```diff
@@ -6,18 +6,20 @@ import org.apache.mesos
 import scala.collection.JavaConverters._
 
 object PortDefinitionSerializer {
-  def toProto(portDefinition: PortDefinition): mesos.Protos.Port = {
-    val builder = mesos.Protos.Port.newBuilder
-      .setNumber(portDefinition.port)
-      .setProtocol(portDefinition.protocol)
+  def toProto(portDefinition: PortDefinition): Seq[mesos.Protos.Port] = {
```
aquamatthias: This method is used for 2 purposes:
- store the port definition as part of the app in zk 
- set discovery information on task

Where the latter now works correctly, you doubled the port information for the first one.

Please introduce a a separate method (e.g. `toMesosProto`) with this implementation and leave the original unchanged. The TaskBuilder has to use `toMesosProto`, the AppDefinitionSerializer `toProto`.

unterstein:
@aquamatthias Did I get your feedback? https://github.com/mesosphere/marathon/pull/4284/commits/146ae56e01bd4542eb329027a72312c59ce02efd

unterstein:
retest this please

unterstein:
retest this please

unterstein:
retest this please

unterstein:
retest this please

aquamatthias:
Thanks! 👍

aquamatthias:
Cherry picked in 1.3 release branch

PR #4286, opened 2016-09-01
/queue should support pods

jdef:


project/RamlGeneratorPlugin.scala
```diff
@@ -13,7 +13,10 @@ object RamlGeneratorPlugin extends AutoPlugin {
   }
   import autoImport._
   override lazy val projectSettings = inConfig(Compile)(Seq(
-    ramlFiles := Seq(baseDirectory.value / "docs" / "docs" / "rest-api" / "public" / "api" / "v2" / "pods.raml"),
+    ramlFiles := Seq(
+      baseDirectory.value / "docs" / "docs" / "rest-api" / "public" / "api" / "v2" / "pods.raml",
+      baseDirectory.value / "docs" / "docs" / "rest-api" / "public" / "api" / "v2" / "queue.raml"
```
jdef: not sure this is really necessary, but I was using it for validation and to test the limits of codegen

jasongilanfarr: The aim is to eventually have this just be the api.raml.

docs/docs/rest-api/public/api/v2/types/queue.raml
```diff
@@ -0,0 +1,35 @@
...
+types:
+  QueueDelay:
+    properties:
+      timeLeftSeconds?:
+        type: number
+        format: int32
+      overdue?:
+        type: boolean
+  QueueObject:
+    properties:
```
jasongilanfarr: need a discriminator and discriminator value if you're going to explicitly list the base type (not required).

jdef: validation of the example failed when I added a discriminator. plus, the
concrete types are distinct enough: they each have a different set of
required fields

PR #4311, opened 2016-09-06
Fixes #3957 - Load previously stored health status when becoming leader

### LINKED ISSUE #3957: COMMAND HealthChecks removed on mesos-master restart/issue

gkleiman:
See #4094 for some context.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/health/HealthResult.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/HealthResult.scala
@@ -7,15 +7,18 @@ sealed trait HealthResult {
   def taskId: Task.Id
   def version: Timestamp
   def time: Timestamp
+  def publishEvent: Boolean
 }
 
 case class Healthy(
   taskId: Task.Id,
   version: Timestamp,
-  time: Timestamp = Timestamp.now()) extends HealthResult
+  time: Timestamp = Timestamp.now(),
+  publishEvent: Boolean = true) extends HealthResult
 
 case class Unhealthy(
   taskId: Task.Id,
   version: Timestamp,
   cause: String,
-  time: Timestamp = Timestamp.now()) extends HealthResult
+  time: Timestamp = Timestamp.now(),
+  publishEvent: Boolean = true) extends HealthResult
--- a/src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala
@@ -50,20 +50,34 @@ class MarathonHealthCheckManager(
       ahcs(appId)(appVersion)
     }
 
-  override def add(app: AppDefinition, healthCheck: HealthCheck): Unit =
+  override def add(app: AppDefinition, healthCheck: HealthCheck, tasks: Iterable[Task]): Unit =
     appHealthChecks.writeLock { ahcs =>
       val healthChecksForApp = listActive(app.id, app.version)
 
-      if (healthChecksForApp.exists(_.healthCheck == healthCheck))
-        log.debug(s"Not adding duplicate health check for app [$app.id] and version [${app.version}]: [$healthCheck]")
-
-      else {
+      if (healthChecksForApp.exists(_.healthCheck == healthCheck)) {
+        log.debug(s"Not adding duplicated health check for app [$app.id] and version [${app.version}]: [$healthCheck]")
+      } else {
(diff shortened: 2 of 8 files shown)


src/main/scala/mesosphere/marathon/upgrade/DeploymentActor.scala
```diff
@@ -88,7 +88,8 @@ private class DeploymentActor(
       eventBus.publish(status)
 
       val futures = step.actions.map { action =>
-        healthCheckManager.addAllFor(action.app) // ensure health check actors are in place before tasks are launched
+        // ensure health check actors are in place before tasks are launched
+        healthCheckManager.addAllFor(action.app, Seq.empty)
```
jdef: the "ensure ... are in place" semantics depend on a side-affect of addAllFor. would prefer to see a MarathonHealthCheckManager `prepareFor(...)` or something (and maybe that impl would invoke addAllFor, hiding the impl detail from DeploymentActor). Would make for a cleaner interface between deployments and health checks.

src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala
```diff
@@ -112,13 +126,26 @@ class MarathonHealthCheckManager(
...
+          }
+      }
+
+    appRepository.get(appId).flatMap {
       case None => Future(())
       case Some(app) =>
         log.info(s"reconcile [$appId] with latest version [${app.version}]")
 
         val tasks: Iterable[Task] = taskTracker.appTasksSync(app.id)
+        val tasksByVersion = groupTasksByVersion(tasks)
```
aquamatthias: Why do we need groupByVersion? Isn't that a simple filter for task.version>=app.version?

aquamatthias:
@gkleiman One question regarding groupBy. Have you tested this fix?

src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala
gkleiman: I think that we do need it, because Marathon starts a HealthCheckActor per app version with active tasks, see line #179 (https://github.com/mesosphere/marathon/pull/4311/files#diff-82c63be54fc7bbda204fdfde44104e48R179).

gkleiman:
@aquamatthias I answered to your regarding the groupBy.

I updated the tests and also tested the patch manually, by creating an app with a COMMAND health check, waiting until it's healthy and the restarting the Marathon leader.

src/main/scala/mesosphere/marathon/core/health/HealthResult.scala
```diff
@@ -7,15 +7,18 @@ sealed trait HealthResult {
...
   def time: Timestamp
+  def publishEvent: Boolean
 }
 
 case class Healthy(
   taskId: Task.Id,
   version: Timestamp,
-  time: Timestamp = Timestamp.now()) extends HealthResult
+  time: Timestamp = Timestamp.now(),
+  publishEvent: Boolean = true) extends HealthResult
```
aquamatthias: Not a fan of adding this to the healthy/unhealthy state. Given the current implementation, this is the smallest invasive change. We need to refactor the health check manager to improve.

aquamatthias:
@gkleiman Thanks. Waiting for the tests to become green.

gkleiman:
@aquamatthias they're green =) I'll try to backport this to v1.1

aquamatthias:
Thanks. backport to 1.1 would be great

jdef:
what about cherry-picking into 1.3?

aquamatthias:
@jdef right. once we have 1.1 it should be easy to do it?
@gkleiman wdyt?

ottoyiu:
is the fix cherry-picked into the 1.1 branch? I see 347d9bb commit on it, but an integration test is failing which I can't pinpoint why...

PR #4313, opened 2016-09-06
WIP: continue to extract reusable Task things into Instance, along with some renaming

jdef:


src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
```diff
@@ -25,20 +25,26 @@ private[tracker] object InstanceOpProcessorImpl {
   class TaskStateOpResolver(directTaskTracker: InstanceTracker) {
     private[this] val log = LoggerFactory.getLogger(getClass)
 
+    def resolve(op: InstanceStateOp)(implicit ec: ExecutionContext): Future[TaskStateChange] = {
+      op match {
+        case op: InstanceStateOp.ForceExpunge => expungeTask(op.instanceId)
```
jdef: not necessarily an app-task, might be a pod-instance

jdef: already a TODO in expungeTask

src/main/scala/mesosphere/marathon/core/launcher/LauncherModule.scala
```diff
@@ -13,13 +13,13 @@ import mesosphere.marathon.metrics.Metrics
...
-    marathonSchedulerDriverHolder: MarathonSchedulerDriverHolder,
-    offerMatcher: OfferMatcher,
-    pluginManager: PluginManager) {
+                      clock: Clock,
+                      metrics: Metrics,
+                      conf: MarathonConf,
+                      taskCreationHandler: InstanceCreationHandler,
+                      marathonSchedulerDriverHolder: MarathonSchedulerDriverHolder,
+                      offerMatcher: OfferMatcher,
+                      pluginManager: PluginManager) {
```
jeschkies: Is this the proper indention? I'm just curious.

src/main/scala/mesosphere/marathon/core/task/TaskStateChange.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object TaskStateChange {
+  case class Update(newState: Task, oldState: Option[Task]) extends TaskStateChange
+  case class Expunge(task: Task) extends TaskStateChange
+  case class NoChange(taskId: Instance.Id) extends TaskStateChange
+  case class Failure(cause: Throwable) extends TaskStateChange
+  object Failure {
+    def apply(message: String): Failure = Failure(TaskStateChangeException(message))
+  }
+}
```
jeschkies: Is there some convention of object over package? In this case it seems that TaskStateChange is rather a package than an object. Just a little nitpicking on my side ;)

jdef: Unsure of our team convention. I extracted this code, nearly verbatim, from
another file in the codebase. I'm happy to factor further if it doesn't
meet our team standards.

src/main/scala/mesosphere/marathon/core/launcher/LauncherModule.scala
jdef: probably not. There's probably an IntelliJ setting that I need to change. I
recently installed it and there's a bunch of things I'm still learning how
to configure.

src/main/scala/mesosphere/marathon/core/task/TaskStateChange.scala
meichstedt: Yes, there's no obvious convention so far. I prefer packages but most of the codebase encapsulates into objects. Happy to change this.

jasongilanfarr: I don't see the point of the wrapping object. How about just removing the wrapper?

jasongilanfarr: Seems like it's a pattern from actor messages, but otherwise I don't really see the point of wrapping the subtypes in anything. The code generator had to do that only because there were name conflicts without it, but in general, I think it's cleaner without any wrapping.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceCreationHandler.scala
```diff
@@ -0,0 +1,25 @@
...
+  * Notifies the [[InstanceTracker]] of instance creation and termination.
+  */
+trait InstanceCreationHandler {
+  /**
+    * Create a new instance.
+    *
+    * If the instance exists already, the existing instance will be overwritten so make sure
+    * that you generate unique IDs.
+    */
+  def created(stateOp: InstanceStateOp): Future[Unit]
```
jasongilanfarr: Let's take the opportunity to use Done instead of Unit.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceCreationHandler.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  /**
+    * Remove the instance for the given app/pod with the given ID completely.
+    *
+    * If the instance does not exist, the returned Future will not fail.
+    */
+  def terminated(stateOp: InstanceStateOp.ForceExpunge): Future[_]
```
jasongilanfarr: Why _? I'd say be explicit or parameterize.

PR #4314, opened 2016-09-06
Keep refining the pods API and get it integrated into group more

jasongilanfarr:


docs/docs/rest-api/public/api/v2/pods.raml
```diff
@@ -31,6 +31,8 @@ traits: !include ../traits.raml
     responses:
       201:
         description: Pod created successfully.
+        headers:
+          X-Marathon-Deployment-Id: string
```
jdef: dislike `X-`

pretty sure that current recommendations are against using `X-` because it becomes somewhat meaningless over time if everyone uses `X-` for custom headers; people might as well use the non-`X-` form.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -31,4 +77,113 @@ class PodsResource @Inject() (
...
+  @PUT @Timed @Path("""{id:.+}""")
+  def update(
+    @PathParam("id") id: String,
+    body: Array[Byte],
+    @DefaultValue("false")@QueryParam("force") force: Boolean,
+    @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
+
+    withValid(unmarshal(body)) { podDef =>
+      if (id != podDef.id) {
+        Response.status(Status.BAD_REQUEST).entity(
```
jdef: :+1:

docs/docs/rest-api/public/api/v2/pods.raml
jasongilanfarr: I missed that the RFC finally deprecated the X- prefix.

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  val scalingValidator: Validator[PodScalingPolicy] = new Validator[PodScalingPolicy] {
+    override def apply(v1: PodScalingPolicy): Result = v1 match {
+      case fsf: FixedPodScalingPolicy => fixedPodScalingPolicyValidator(fsf)
+      case _ => Failure(Set(RuleViolation(v1, "Not a fixed scaling policy", None)))
+    }
+  }
+
+  def podDefValidator(enabledFeatures: Set[String]): Validator[Pod] = validator[Pod] { pod =>
```
jdef: :+1:

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -0,0 +1,162 @@
...
+import mesosphere.marathon.state.{ PathId, ResourceRole }
+
+import scala.collection.immutable.Seq
+import scala.util.Try
+// scalastyle:on
+
+/**
+  * Defines implicit validation for pods
+  */
+trait PodsValidation {
```
jdef: is this a trait so that we can mock it when unit testing elsewhere?

jasongilanfarr: No, it's a trait so it can be mixed in and any implicits are lower priority. The companion object extends it...

src/main/scala/mesosphere/marathon/api/v2/DeploymentsResource.scala
```diff
@@ -28,7 +28,10 @@ class DeploymentsResource @Inject() (
   @GET
   def running(@Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
     val infos = result(service.listRunningDeployments())
-      .filter(_.plan.affectedApplications.exists(isAuthorized(ViewRunSpec, _)))
+      .filter(steps => (steps.plan.affectedApplications.isEmpty ||
```
jdef: how did the `isEmpty` checks work their way in here?

jasongilanfarr: Previously, deployment plans always had apps. Now they can have apps or pods and they definitely have at least one of them, but either could be empty, so exists is automatically false. Hence, we have to check if they're empty.

src/main/scala/mesosphere/marathon/api/RestResource.scala
```diff
@@ -73,4 +74,15 @@ trait RestResource {
       case Success => fn(t)
     }
   }
+
+  def decodeBytes(data: Array[Byte], req: HttpServletRequest): String = {
```
aquamatthias: Do we still need this method? Looks like a left over.

src/main/scala/mesosphere/marathon/api/v2/DeploymentsResource.scala
aquamatthias: I added an `affectedRunSpecs` in my branch on deployments which hold all apps and pods that are affected.

aquamatthias:
LGTM 👍

jasongilanfarr:
Merged into feature branch

Issue #4183, opened 2016-08-05
portMappings w/o a defined hostPort don't get port-level discoveryInfo


jdef:
xref https://github.com/mesosphere/marathon/blob/9b1ee24ede277b55a7747bd27015267906c44b71/src/main/scala/mesosphere/mesos/TaskBuilder.scala#L191

/cc @sargun

jdef:
it appears that we normally stuff hostPort into the DI port's `number` field. without a host port, what value should we use instead? the service port? container port? if we skip the port then how will dcos networking know about related VIPs?

PR #4316, opened 2016-09-07
Set port if host port is not defined in mapping.

jeschkies:
Addresses #4183.

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -177,6 +177,7 @@ class TaskBuilder(
   protected def computeDiscoveryInfo(
     runSpec: RunSpec,
     hostPorts: Seq[Option[Int]]): org.apache.mesos.Protos.DiscoveryInfo = {
+
     val discoveryInfoBuilder = org.apache.mesos.Protos.DiscoveryInfo.newBuilder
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -188,7 +189,13 @@ class TaskBuilder(
           case Some(portMappings) =>
             // The run spec uses bridge and user modes with portMappings, use them to create the Port messages
             portMappings.zip(hostPorts).collect {
-              case (portMapping, Some(hostPort)) => PortMappingSerializer.toMesosPort(portMapping, hostPort)
+              case (portMapping, None) =>
+                // No host port has been defined. See PortsMatcher.mappedPortRanges, use container port instead.
+                val updatedPortMapping =
+                  portMapping.copy(labels = portMapping.labels + ("network-scope" -> "container"))
+                PortMappingSerializer.toMesosPort(updatedPortMapping, portMapping.containerPort)
```
jeschkies: I doubt that this is the right port that enables discovery. @sargun what is the appropiate way of handling this?

We have this case if `portMapping` does not define a host port.

jdef: this is what @sargun asked for. The network-scope provides context for the mapping

jeschkies:
Closed by ccae032e5cd7a0dedfb5826ddcba682eec24bafa

PR #4317, opened 2016-09-07
Define env var for port mapping without host port.

### LINKED ISSUE #4005: Also create 'PORT_<NAME>' env variables for ports defined in port mappings

jeschkies:
Fixes #4005 

Is related to #4316

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -29,7 +29,7 @@ case class PortsMatch(hostPortsWithRole: Seq[Option[PortWithRole]]) {
 /**
   * Utility class for checking if the ports resource in an offer matches the requirements of an app.
   */
-class PortsMatcher(
+class PortsMatcher private[tasks] (
   runSpec: RunSpec,
   offer: MesosProtos.Offer,
   resourceSelector: ResourceSelector = ResourceSelector.any(Set(ResourceRole.Unreserved)),
@@ -169,6 +169,12 @@ class PortsMatcher(
 }
 
 object PortsMatcher {
+
+  def apply(
+    runSpec: RunSpec,
+    offer: MesosProtos.Offer,
+    resourceSelector: ResourceSelector = ResourceSelector.any(Set(ResourceRole.Unreserved)),
+    random: Random = Random): PortsMatcher = new PortsMatcher(runSpec, offer, resourceSelector, random)
 
   // Request represents some particular type of port resource request.
   // If there is no such request for a port, then use RequestNone.
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -177,6 +177,7 @@ class TaskBuilder(
   protected def computeDiscoveryInfo(
     runSpec: RunSpec,
     hostPorts: Seq[Option[Int]]): org.apache.mesos.Protos.DiscoveryInfo = {
+
     val discoveryInfoBuilder = org.apache.mesos.Protos.DiscoveryInfo.newBuilder
     discoveryInfoBuilder.setName(runSpec.id.toHostname)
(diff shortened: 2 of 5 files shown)


src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -29,7 +29,7 @@ case class PortsMatch(hostPortsWithRole: Seq[Option[PortWithRole]]) {
 /**
   * Utility class for checking if the ports resource in an offer matches the requirements of an app.
   */
-class PortsMatcher(
+class PortsMatcher private[tasks] (
```
jdef: nit: why is the visibility modifier after the type name? doesn't it usually prefix the type declaration?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -188,7 +189,13 @@ class TaskBuilder(
...
             // The run spec uses bridge and user modes with portMappings, use them to create the Port messages
             portMappings.zip(hostPorts).collect {
-              case (portMapping, Some(hostPort)) => PortMappingSerializer.toMesosPort(portMapping, hostPort)
+              case (portMapping, None) =>
+                // No host port has been defined. See PortsMatcher.mappedPortRanges, use container port instead.
+                val updatedPortMapping =
+                  portMapping.copy(labels = portMapping.labels + ("network-scope" -> "container"))
+                PortMappingSerializer.toMesosPort(updatedPortMapping, portMapping.containerPort)
+              case (portMapping, Some(hostPort)) =>
+                PortMappingSerializer.toMesosPort(portMapping, hostPort)
```
jdef: should we also add a `network-scope` label here, with a value of `host`?

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -1838,16 +1912,22 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
       p.map { hostPort =>
         val b = MesosProtos.Port.newBuilder.setNumber(hostPort)
         if (name != "") b.setName(name)
         if (protocol != "") b.setProtocol(protocol)
-        if (vip != "") b.setLabels(MesosProtos.Labels.newBuilder().addLabels(
-          MesosProtos.Label.newBuilder().setKey("VIP").setValue(vip)
-        ))
+        if (labels.nonEmpty) {
+          val labelsBuilder = MesosProtos.Labels.newBuilder()
+          labels foreach { case(key, value) =>
```
jdef: nit: `labels.foreach`

jdef:
minor comments. otherwise looks good

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
jeschkies: It makes the constructor private but the class public. If the class is private the compiler complained that it was somewhere else. It's actually not an important change.

src/main/scala/mesosphere/mesos/TaskBuilder.scala
jeschkies: I'm not sure. @sargun what's your suggestion?

sargun: Yeah, I need to know that as well.

jeschkies:
retest this please

jdef:
LGTM

jeschkies:
Okay, I'll merge if the tests pass.

jeschkies:
retest this please

jdef:
:tada:

jdef:
@aquamatthias should we be cherry-picking this for 1.3? it was tagged for that milestone...

jasongilanfarr:
We can start 1.3.1

jdef:
looks like it already landed in v1.3.0

PR #4319, opened 2016-09-07
Make deployments aware of apps and pods.

aquamatthias:
Make the DeploymentActions generic to support RunnableSpec.
Rename properties to meet runSpec names.

This is the first step to make deployments aware of pods - mostly to get the types right.
Missing functionality: listen to instance events, handling instances instead of tasks.
Will do this in a separate PR once the events are available.

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/RunSpec.scala
+++ b/src/main/scala/mesosphere/marathon/state/RunSpec.scala
@@ -9,6 +9,7 @@ import mesosphere.marathon.state.AppDefinition.VersionInfo
 
 import scala.concurrent.duration.FiniteDuration
 import scala.collection.immutable.Seq
+import scala.language.implicitConversions
 
 /**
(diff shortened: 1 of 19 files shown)


src/main/scala/mesosphere/marathon/state/RunSpec.scala
```diff
@@ -31,6 +32,14 @@ trait RunnableSpec extends plugin.RunSpec {
   def gpus: Int
 }
 
+object RunnableSpec {
+
+  //TODO(PODS): Work in progress conversion to make the code compile
+  implicit def runSpecToApp(runnableSpec: RunnableSpec): AppDefinition = {
+    runnableSpec.asInstanceOf[AppDefinition]
+  }
```
jasongilanfarr: Let's do a match...

jasongilanfarr:
:+1: LGTM - merging manually.

PR #4322, opened 2016-09-07
Fix a few blockers in persistence store

jasongilanfarr:
- Deployment plans can be stored before the groups are.
  Instead, always store the group first.
- ZKPersistenceStore was actually failing to store the StorageVersion
  due to missing a field in the wrapper.
- Migration didn't wait on setting storage version, so the error above
  was hidden.
- StoredGroup fails silently and forever if apps couldn't
  be parsed. Log that and log if any of them are missing (which should
  never happen)

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
@@ -132,8 +132,8 @@ private[impl] class GroupManagerActor(
         plan = DeploymentPlan(from, to, resolve, version, toKill)
         _ = validateOrThrow(plan)(DeploymentPlan.deploymentPlanValidator(config))
         _ = log.info(s"Computed new deployment plan:\n$plan")
-        _ <- scheduler.deploy(plan, force)
         _ <- groupRepo.storeRoot(plan.target, plan.createdOrUpdatedApps, plan.deletedApps)
+        _ <- scheduler.deploy(plan, force)
         _ = log.info(s"Updated groups/apps according to deployment plan ${plan.id}")
       } yield plan
 
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
@@ -132,8 +132,8 @@ private[impl] class GroupManagerActor(
         plan = DeploymentPlan(from, to, resolve, version, toKill)
         _ = validateOrThrow(plan)(DeploymentPlan.deploymentPlanValidator(config))
         _ = log.info(s"Computed new deployment plan:\n$plan")
-        _ <- groupRepo.storeRoot(plan.target, plan.createdOrUpdatedApps, plan.deletedApps)
         _ <- scheduler.deploy(plan, force)
+        _ <- groupRepo.storeRoot(plan.target, plan.createdOrUpdatedApps, plan.deletedApps)
         _ = log.info(s"Updated groups/apps according to deployment plan ${plan.id}")
       } yield plan
 
--- a/src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
@@ -31,15 +31,25 @@ private[storage] case class StoredGroup(
     appIds: Map[PathId, OffsetDateTime],
     storedGroups: Seq[StoredGroup],
     dependencies: Set[PathId],
-    version: OffsetDateTime) {
+    version: OffsetDateTime) extends StrictLogging {
 
   lazy val transitiveAppIds: Map[PathId, OffsetDateTime] = appIds ++ storedGroups.flatMap(_.appIds)
 
   def resolve(
     appRepository: AppRepository)(implicit ctx: ExecutionContext): Future[Group] = async {
-    val appFutures = appIds.map { case (appId, appVersion) => appRepository.getVersion(appId, appVersion) }
+    val appFutures = appIds.map {
+      case (appId, appVersion) => appRepository.getVersion(appId, appVersion).recover {
+        case NonFatal(ex) =>
+          logger.error(s"Failed to load $appId:$appVersion for group $id ($version)", ex)
+          None
+      }
+    }
     val groupFutures = storedGroups.map(_.resolve(appRepository))
 
+    val allApps = await(Future.sequence(appFutures))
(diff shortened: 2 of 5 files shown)


src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -132,8 +132,8 @@ private[impl] class GroupManagerActor(
         plan = DeploymentPlan(from, to, resolve, version, toKill)
         _ = validateOrThrow(plan)(DeploymentPlan.deploymentPlanValidator(config))
         _ = log.info(s"Computed new deployment plan:\n$plan")
-        _ <- scheduler.deploy(plan, force)
         _ <- groupRepo.storeRoot(plan.target, plan.createdOrUpdatedApps, plan.deletedApps)
+        _ <- scheduler.deploy(plan, force)
```
jdef: maybe a comment in the code here that the order is important so be careful when refactoring?

jdef:
migration unit test is failing:

https://jenkins.mesosphere.com/service/jenkins/job/public-marathon-snapshot-packages-pulls/787/testReport/mesosphere.marathon.storage.migration/MigrationTest/Migration_should_migrate_on_an_empty_database_will_set_the_storage_version/

```
Error Message

mesosphere.marathon.MigrationFailedException: Migration Failed: null
Stacktrace

sbt.ForkMain$ForkError: mesosphere.marathon.MigrationFailedException: Migration Failed: null
	at mesosphere.marathon.storage.migration.Migration$$anonfun$1.applyOrElse(Migration.scala:152)
	at mesosphere.marathon.storage.migration.Migration$$anonfun$1.applyOrElse(Migration.scala:150)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
	at scala.util.Failure$$anonfun$recover$1.apply(Try.scala:216)
	at scala.util.Try$.apply(Try.scala:192)
	at scala.util.Failure.recover(Try.scala:216)
	at scala.concurrent.Future$$anonfun$recover$1.apply(Future.scala:326)
	at scala.concurrent.Future$$anonfun$recover$1.apply(Future.scala:326)
	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:32)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.pollAndExecAll(ForkJoinPool.java:1253)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1346)
	at scala.concurrent.forkjoin.ForkJoinPool.runWorker(ForkJoinPool.java:1979)
	at scala.concurrent.forkjoin.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:107)
Caused by: sbt.ForkMain$ForkError: java.lang.NullPointerException: null
	at mesosphere.marathon.storage.migration.Migration$stateMachine$macro$346$1.apply(Migration.scala:174)
	at mesosphere.marathon.storage.migration.Migration$stateMachine$macro$346$1.apply$mcV$sp(Migration.scala:174)
	at mesosphere.marathon.storage.migration.Migration$stateMachine$macro$346$1.apply(Migration.scala:174)
	at mesosphere.marathon.storage.migration.Migration$stateMachine$macro$346$1.apply(Migration.scala:174)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.liftedTree1$1(Future.scala:24)
	at scala.concurrent.impl.Future$PromiseCompletingRunnable.run(Future.scala:24)
	at scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:121)
	at scala.concurrent.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at scala.concurrent.forkjoin.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1339)
	... 2 more
```

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
aquamatthias: The order is definitively important here.
I would say the order was correct before and is wrong now, since a deployment can be rejected. This is most usually the case, if there is a deployment that affects the same app (no force). In that case the future will fail.
We should store the group only, if a deployment is started.

What was the reasoning to change the order?

src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
```diff
@@ -31,15 +31,25 @@ private[storage] case class StoredGroup(
...
+        case NonFatal(ex) =>
+          logger.error(s"Failed to load $appId:$appVersion for group $id ($version)", ex)
+          None
+      }
+    }
     val groupFutures = storedGroups.map(_.resolve(appRepository))
 
+    val allApps = await(Future.sequence(appFutures))
+    if (allApps.exists(_.isEmpty)) {
+      logger.warn(s"Group $id $version is missing ${allApps.count(_.isEmpty)} apps")
```
aquamatthias: If we can not resolve all app references, we log this as error/warn.
I wonder if we should fail here (exception)?

aquamatthias:
@jasongilanfarr one ask regarding the order of the deployment execution. Otherwise looks good.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
jasongilanfarr: So actually, there is a more tricky case that we should solve then - store the rootVersion then ask the scheduler to deploy, then storeRoot (setting it to the current version). I ran into some scenarios - it could have been related to the fact that pod deserialization was busted where deployment plans were referring to roots that hadn't been stored. 

Perhaps I haven't had enough coffee yet, but I think this is still the case unless we do "storeRootVersion" before we do a scheduler deploy.

src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
jasongilanfarr: I thought about that too - I chose to error if there was an error reading it, but if the application isn't found, to just log - as that would indicate the app was deleted by the user or something?

jasongilanfarr:
retest this please

jasongilanfarr:
retest this please.

PR #4326, opened 2016-09-08
Document disk unit.

jeschkies:
I regenerated the docs and update their instructions.

It seems a few things have been added since. I'm not sure if this should be merged since the docs apply to the version in Master.


aquamatthias:
@jeschkies please always create PRs against master. The gh-pages branch is synced with master regular.

PR #4329, opened 2016-09-08
Make branch build again.

jeschkies:
* Don't write empty collections to Json.
* Deal with empty `acceptedResourceRoles` collection in proto write.
* Check for non empty tasks in `TaskKiller`.
* Don't treat storage as future but as option.
* Correct `RunSpec` reference in docstrings.
* Map task killer answer correctly.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -1010,13 +1010,17 @@ trait AppAndGroupFormats {
         "dependencies" -> runSpec.dependencies,
         "upgradeStrategy" -> runSpec.upgradeStrategy,
         "labels" -> runSpec.labels,
-        "acceptedResourceRoles" -> runSpec.acceptedResourceRoles,
         "ipAddress" -> runSpec.ipAddress,
         "version" -> runSpec.version,
         "residency" -> runSpec.residency,
         "secrets" -> runSpec.secrets,
         "taskKillGracePeriodSeconds" -> runSpec.taskKillGracePeriod
       )
+
+      if (runSpec.acceptedResourceRoles.nonEmpty) {
+        appJson + ("acceptedResourceRoles" -> Json.toJson(runSpec.acceptedResourceRoles))
+      }
+
       // top-level ports fields are incompatible with IP/CT
       if (runSpec.ipAddress.isEmpty) {
         appJson = appJson ++ Json.obj(
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -1017,11 +1017,11 @@ trait AppAndGroupFormats {
       )
 
       if (runSpec.args.nonEmpty) {
-        appJson + ("args" -> Json.toJson(runSpec.args))
+        appJson = appJson + ("args" -> Json.toJson(runSpec.args))
       }
 
       if (runSpec.acceptedResourceRoles.nonEmpty) {
-        appJson + ("acceptedResourceRoles" -> Json.toJson(runSpec.acceptedResourceRoles))
+        appJson = appJson + ("acceptedResourceRoles" -> Json.toJson(runSpec.acceptedResourceRoles))
       }
 
       // top-level ports fields are incompatible with IP/CT
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -988,7 +988,6 @@ trait AppAndGroupFormats {
       var appJson: JsObject = Json.obj(
         "id" -> runSpec.id.toString,
         "cmd" -> runSpec.cmd,
-        "args" -> runSpec.args,
         "user" -> runSpec.user,
         "env" -> runSpec.env,
         "instances" -> runSpec.instances,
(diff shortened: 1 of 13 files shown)


src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -173,16 +173,17 @@ class Migration(
...
+        legacyStore.map { store =>
+          store.load(StorageVersionName).flatMap {
+            case Some(entity) => store.update(entity.withNewContent(bytes))
+            case None => store.create(StorageVersionName, bytes)
+          }
         }
-      }
-    }.getOrElse(Future.successful(Done))
-    await(storeVersionFuture)
+    }
```
jeschkies: `persistenceStore` is not a future as suggested originally. This change should make it more explicit. Do we want a future instead?

jdef: /cc @jasongilanfarr

jasongilanfarr: setting the storage version is a future and we should be waiting on it.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -1010,13 +1009,21 @@ trait AppAndGroupFormats {
...
         "labels" -> runSpec.labels,
-        "acceptedResourceRoles" -> runSpec.acceptedResourceRoles,
         "ipAddress" -> runSpec.ipAddress,
         "version" -> runSpec.version,
         "residency" -> runSpec.residency,
         "secrets" -> runSpec.secrets,
         "taskKillGracePeriodSeconds" -> runSpec.taskKillGracePeriod
       )
+
+      if (runSpec.args.nonEmpty) {
```
jasongilanfarr: Don't see why this matters

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -163,9 +163,11 @@ case class AppDefinition(
     readinessChecks.foreach { r => builder.addReadinessCheckDefinition(ReadinessCheckSerializer.toProto(r)) }
     taskKillGracePeriod.foreach { t => builder.setTaskKillGracePeriod(t.toMillis) }
 
-    val roles = Protos.ResourceRoles.newBuilder()
-    roles.addAllRole(acceptedResourceRoles.asJava)
-    builder.setAcceptedResourceRoles(roles)
+    if (acceptedResourceRoles.nonEmpty) {
+      val roles = Protos.ResourceRoles.newBuilder()
```
jasongilanfarr: Why would this matter?

jeschkies: If we set an empty list proto will set an empty list and `hasAcceptedResourceRoles` will be true. It broke this assertion: https://github.com/mesosphere/marathon/blob/feature/pods/src/test/scala/mesosphere/marathon/state/AppDefinitionTest.scala#L72.

jasongilanfarr: We should fix that then... hasAcceptedResourceRoles should be == "acceptedResourceRoles.nonEmpty"

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
jeschkies: Okay. Changed it.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
jasongilanfarr: If we don't set an empty list on the protos, when we read it, it'll still be empty ;)

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
jeschkies: It'll be an empty list and break this assetion 
`(r1 \ "args").asOpt[Seq[String]] should equal (None)` in
https://github.com/mesosphere/marathon/blob/feature/pods/src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala#L53

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
jeschkies: Hehe, yeah. However, `hasAcceptedResourceRoles` looks for `null`. I don't have a strong opinion on this. I just didn't want to change the assertions.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
jasongilanfarr: Reasonable, I guess- I still think it's awkward to treat empty list and unset lists differently.

jeschkies: Me too. It should default to an empty list. Let me fix that then.

jeschkies: Hm, so it seems we do have a semantic difference between `Set.empty` and `null` for accepted resource roles: https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala#L187

At least we expect an exception.

jasongilanfarr:
(merged)

PR #4330, opened 2016-09-08
Fixes #4028: First health check is scheduled immediately after the task was launched.

### LINKED ISSUE #4028: Marathon remains "Deploying" status when container has already deployed

zen-dog:
First health check is now applied DefaultFirstHealthCheckTime (1s by default) after the task was successfully launched. scheduler.schedule(...) is used instead of scheduler.scheduleOnce(...) which simplifies code a bit.

- a few minor scalastyle complains also fixed

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
@@ -46,8 +46,8 @@ case class HealthCheck(
 
     path foreach builder.setPath
 
-    portIndex foreach { p => builder.setPortIndex(p.toInt) }
-    port foreach { p => builder.setPort(p.toInt) }
+    portIndex foreach { p => builder.setPortIndex(p) }
+    port foreach { p => builder.setPort(p) }
 
     builder.build
   }
@@ -123,6 +123,7 @@ object HealthCheck {
   val DefaultPortIndex = None
   val DefaultCommand = None
   // Dockers can take a long time to download, so default to a fairly long wait.
+  val DefaultFirstHealthCheckTime = 1.seconds
   val DefaultGracePeriod = 5.minutes
   val DefaultInterval = 1.minute
   val DefaultTimeout = 20.seconds
--- a/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
@@ -21,7 +21,7 @@ private[health] class HealthCheckActor(
   import context.dispatcher
   import HealthCheckWorker.HealthCheckJob
 
-  var nextScheduledCheck: Option[Cancellable] = None
+  var scheduledCheck: Option[Cancellable] = None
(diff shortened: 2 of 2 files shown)


unterstein:
lgtm

should this be configurable in the future?

jdef:
* after successfully launched, or reported as RUNNING?
* and why only 1s? why not 5s? 10s? what's driving this?

zen-dog:
> It seems to me that the need is a higher frequency health check poll until the task reports healthy.
I agree, however that might not be necessary at all except for the very first health check. And yes - 1s seems to be arbitrary, but we can make it configurable via the UI (or just JSON) like the other parameters.

> ... to pick some random value between 0 and intervalSeconds/2
This would be then a source of some-times-it-works-and-sometimes-it-doesn't complains. I'd rather keep it simple, provide some meaningful default, and let the user configure it in their special cases.

jdef:
:( to even more parameters. the health check api shouldn't get any more
complicated. given that we're eventually pushing all health checks to mesos
anyway ... my former suggestion is probably moot. so lets keep it simple.
if we're going to make 1s configurable, let's please avoid polluting the API

On Fri, Sep 9, 2016 at 11:29 AM, Aleksey Dukhovniy <notifications@github.com

aquamatthias:
Close in favor of #4368

PR #4332, opened 2016-09-08
Take executor resources into account for pods.

nfnt:
A pod consist of an executor and multiple tasks. As the executor
consumes resources like a task, it's resources need to be accounted for
when matching offers. Because the default executor is used, the minimal
allowed resources suffice for that.


jdef:
works for me

nfnt:
Thanks! Merged.

PR #4339, opened 2016-09-09
WIP: Log more context in integration tests.

jeschkies:
This is a WIP to trigger integration test and see wha'ts going on.


src/test/scala/mesosphere/marathon/integration/setup/MarathonCallbackTestSupport.scala
```diff
@@ -29,29 +33,56 @@ trait MarathonCallbackTestSupport extends ExternalMarathonIntegrationTest {
...
     waitForDeploymentId(change.value.deploymentId, maxWait)
   }
 
   def waitForEventMatching(description: String, maxWait: FiniteDuration = 30.seconds)(fn: CallbackEvent => Boolean): CallbackEvent = {
-    @tailrec
-    def nextEvent: Option[CallbackEvent] = if (events.isEmpty) None else {
-      val event = events.poll()
-      if (fn(event)) Some(event) else nextEvent
+
+    def findEvent: Option[CallbackEvent] = {
```
aquamatthias: I am pretty sure, there are some integration test, that rely on the fact, that not matching events are discarded.
What is the reasoning of this change?
Perhaps make a separate `waitForEventXXX` method?

src/test/scala/mesosphere/marathon/integration/setup/MarathonCallbackTestSupport.scala
```diff
@@ -29,29 +33,56 @@ trait MarathonCallbackTestSupport extends ExternalMarathonIntegrationTest {
...
+  override def handleEvent(event: CallbackEvent): Unit = {
+    log.info(s"Add event to events list: $event")
+    events.add(event)
+  }
 
   def waitForEvent(kind: String, maxWait: FiniteDuration = 60.seconds): CallbackEvent = waitForEventWith(kind, _ => true, maxWait)
 
   def waitForDeploymentId(deploymentId: String, maxWait: FiniteDuration = 30.seconds): CallbackEvent = {
-    waitForEventWith("deployment_success", _.id == deploymentId, maxWait)
+    waitForEventWith("deployment_success", { c: CallbackEvent => c.info("id").toString == deploymentId }, maxWait)
```
aquamatthias: The `id` method does exactly this, but in one place. I would argue, this is not DRY? Any reason for doing this?

aquamatthias:
@jeschkies Thanks! There are failing integration tests. Please see my comments.

src/test/scala/mesosphere/marathon/integration/setup/MarathonCallbackTestSupport.scala
jeschkies: The problem is that some messages don't come in order. E.g. in `ResidentTaskIntegrationTest`

we have the following

```
waitForStatusUpdates(StatusUpdate.TASK_RUNNING)
waitForEvent(Event.DEPLOYMENT_SUCCESS)
waitForStatusUpdates(StatusUpdate.TASK_FINISHED)
```

I've found that `Event.DEPLOYMENT_SUCCESS` would sometimes come in before `Event.TASK_RUNNING`. `waitForStatusUpdates(StatusUpdate.TASK_RUNNING)` would then eat up `Event.DEPLOYMENT_SUCCESS` and thus `waitForEvent(Event.DEPLOYMENT_SUCCESS)` times out.

I don't like my solution. For one because of the issue you mention but also because it won't allow us to run tests in parallel.

I'd do something like:

```
with (f.createAsynchronously(app)) {
  // In any order
  expect(Event.StatusUpdate....) // Registers an event handler. e.g. a promise
  expect(Event.DEPLOYMENT_SUCCESS)
  // ...
}
```

Then in `MarathonCallbackTestSupport.handleEvent` I'd not push the event on a queue but rather really handle it an check registered events. Also, sometimes we test not for the depoyment ids of event.

src/test/scala/mesosphere/marathon/integration/setup/MarathonCallbackTestSupport.scala
jeschkies: Ah, I just got confused what type was expected and where `id` is coming from. Somtimes I like to be a little more verbose in Scala to inform the reader about types etc.

PR #4346, opened 2016-09-09
upgrade to mesos protos w/ task-group; add LaunchTaskGroup for pods

jdef:
incorporates https://github.com/mesosphere/marathon/commit/805fa0e187aac3d6ad3e02491369dcbf7568fb36

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
@@ -6,6 +6,9 @@ import mesosphere.marathon.core.task.{ Task, TaskStateOp }
 import mesosphere.marathon.tasks.ResourceUtil
 import org.apache.mesos.{ Protos => MesosProtos }
 
+import scala.collection.immutable
+import scala.collection.JavaConverters._
+
 /**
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
```diff
@@ -36,6 +36,23 @@ object InstanceOp {
...
+      executorInfo: MesosProtos.ExecutorInfo,
+      groupInfo: MesosProtos.TaskGroupInfo,
+      stateOp: InstanceStateOp,
+      oldInstance: Option[Instance] = None,
+      offerOperations: Iterable[MesosProtos.Offer.Operation]) extends InstanceOp {
+
+    override def applyToOffer(offer: MesosProtos.Offer): MesosProtos.Offer = {
+      import scala.collection.immutable
+      import scala.collection.JavaConverters._
+      val taskResources: immutable.Seq[MesosProtos.Resource] =
```
jasongilanfarr: why not use JavaConversions and make it all "nice and simple"? Also, prefer top level imports - and import scala.collection.immutable.Seq explicitly.

PR #4347, opened 2016-09-09
more launch(ed)-ephemeral glue

jdef:
xref #4352 (looks like more complete pod instance impl)


src/main/scala/mesosphere/marathon/core/instance/InstanceStateOp.scala
```diff
@@ -1,6 +1,6 @@
 package mesosphere.marathon.core.instance
 
-trait InstanceStateOp {
+trait InstanceStateOp extends Product with Serializable {
   def instanceId: Instance.Id
```
jasongilanfarr: Stable things should be 'val'

src/main/scala/mesosphere/marathon/core/instance/InstanceStatus.scala
```diff
@@ -10,6 +10,7 @@ import org.apache.mesos
   * - mapping of existing (soon-to-be deprecated) mesos.Protos.TaskStatus.TASK_LOST to the new representations
   */
 sealed trait InstanceStatus {
+  // TODO(jdef) pods was this renamed too aggressively? Should it really be TaskStatus instead?
   def toMesosStateName: String = {
```
jasongilanfarr: Lazy Val?

jdef:
/cc @nfnt 

@meichstedt merge at will

src/main/scala/mesosphere/marathon/core/pod/PodInstanceStateOp.scala
```diff
@@ -0,0 +1,13 @@
+package mesosphere.marathon.core.pod
+
+import mesosphere.marathon.core.instance.{ Instance, InstanceStateOp }
+
+sealed trait PodInstanceStateOp extends InstanceStateOp
+
+object PodInstanceStateOp {
+
+  case class LaunchedEphemeral(podInstance: PodInstance) extends PodInstanceStateOp {
```
jdef: @meichstedt do we need to differentiate between pods and apps for Launch(ed)Ephemeral, or should I nix this type in favor of TaskStateOp.LaunchEphemeral?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -75,4 +77,7 @@ object Instance {
     def isDropped: Boolean = instance.state.status == InstanceStatus.Dropped
   }
 
+  case class LaunchedEphemeral(
+    instance: Instance,
+    hostPorts: Seq[Int])
```
jdef: why did I think that I needed hostPorts here?

jdef: because Task.LaunchEphemeral has them too, and they're eventually passed to Task.launched(), and from there ... used throughout the system to determine which ports were allocated for the task

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -45,10 +47,11 @@ class InstanceOpFactoryImpl(
   }
 
   private[this] def inferNormalTaskOp(request: InstanceOpFactory.Request): Option[InstanceOp] = {
-    val InstanceOpFactory.Request(runSpec, offer, tasks, _) = request
+    val InstanceOpFactory.Request(runSpec, offer, instances, _) = request
+    val tasks: Seq[Task] = instances.values.flatMap(_.tasks)(collection.breakOut)
 
     new TaskBuilder(runSpec, Task.Id.forRunSpec, config, Some(appTaskProc)).
-      buildIfMatches(offer, tasks.values.flatMap(_.tasks)).map {
+      buildIfMatches(offer, tasks).map {
```
meichstedt: In order to match constraints for pods we should rather be passing instances instead of tasks, but before that's adjusted this is OK

jdef: there's a whole new builder coming for building pods from task-groups

jasongilanfarr: Maybe the builder could actually be in PodSpec as lazy val toTaskGroup?

jdef: um .. I stated it backwards. we'll be building task-groups from pods. :p

meichstedt:
Thanks @jdef, LGTM

Issue #4159, opened 2016-07-29
Marathon event bus stops sending messages if a marathon instance restarts


sybrandy:
We've seen this several times where an app is restarted and Traefik (short version: http proxy that can listen for Marathon events to keep it's config up to date) does not update.  This results in 502 errors until Traefik is restarted.  Up until today, we had no idea why this was occurring as it happens sporadically.  Only recently did we confirm that it was because it stopped getting events.  The initial suspect was Traefik, but digging around showed that a library that is used does reconnect if the connection is lost.  Also, no errors have been reported.

Today, this happened again and I decided to check the Marathon logs to see if there was an error.  The first one I checked had it's log starting shortly after I saw the events stop for one of my nodes.  In this case, this node had rebooted earlier in the day.  The error cropped up some time later and I investigated.  Other instances of Traefik were restarted and I could see them getting events, but not the "debugging" instance I have running.  If I connect to the events endpoint with curl, I can see events.

So, it appears that something inside of Marathon is holding the event stream connection open, but fails to read/receive events until the connections are closed/reopened.  If this happens again, I'll update the bug to confirm it, but I will have to wait until it happens again as we need to keep the system as stable as possible.

xiaods:
+1

jangie:
I am seeing similar, if not identical issues, running traefik against marathon. I have tried, from the marathon side, tcpkill-ing connections that traefik has open against the event bus, which does result in a reconnect from traefik, so it does indeed seem that traefik (specifically its dependency's dependency [eventsource](https://github.com/donovanhide/eventsource)) deals with closed connections well.

jangie:
I suspect that the issue is here: https://github.com/mesosphere/marathon/blob/6df72cdc9535e4dce3e67a3329dca87a1cfe3879/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala#L112-L114 

I am very much not a scala or akka person, but my suspicion is that handle.close() should be called in there as well, to ensure that the underlying connection, and not just the actor servicing the connection, is killed.

jangie:
@sybrandy I think I'm seeing that this does not in fact seem to be caused by election events at the very least; running against the abdication endpoint does not seem to trigger the behavior that I was hoping to replicate. I am curious to see if you've found a resolution to this, as we are at a loss as to what is going on.

sybrandy:
[...]

PR #4348, opened 2016-09-09
Add explicit close to connection

jangie:
In #4159 , we seem to be experiencing cases where connections to marathon essentially become stale, potentially due to leadership changes. I have to admit that I am not particularly well versed in scala or akka, but I suspect that it is because the actor servicing the connection is stopped but the underlying connection itself is left open in these cases.


jangie:
Withdrawing. Tests indicate that handle.close is called separately, and running an abdication manually does not seem to cause behavior. Apologies for any inconvenience.

PR #4353, opened 2016-09-12
Add 'TaskGroupBuilder'.

nfnt:
'TaskGroupBuilder' creates an 'ExecutorInfo' and a 'TaskGroupInfo' from
a 'PodDefinition' and a matching offer. These can then be used to launch
a pod in Mesos using the LAUNCH_GROUP event.

### DIFF
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -49,11 +49,7 @@ class PortsMatcher(
           endpoint <- container.endpoints
         } yield endpoint.hostPort
 
-        if (podDefinition.requirePorts) {
-          mappedPortRanges(requiredPorts)
-        } else {
-          randomPorts(requiredPorts.size)
-        }
+        mappedPortRanges(requiredPorts)
       case _ =>
         val portMappings: Option[Seq[Container.Docker.PortMapping]] =
           for {
--- a/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
@@ -1,8 +1,8 @@
 package mesosphere.marathon.tasks
 
-import mesosphere.marathon.state.{ ResourceRole, RunSpec, Container }
-import mesosphere.marathon.state.Container.Docker.PortMapping
-import mesosphere.marathon.tasks.PortsMatcher.{ Request, RequestNone, PortWithRole }
+import mesosphere.marathon.core.pod.PodDefinition
+import mesosphere.marathon.state.{ Container, ResourceRole, RunSpec }
+import mesosphere.marathon.tasks.PortsMatcher.PortWithRole
 import mesosphere.mesos.ResourceMatcher.ResourceSelector
 import mesosphere.mesos.protos
 import mesosphere.mesos.protos.{ RangesResource, Resource }
@@ -40,27 +40,42 @@ class PortsMatcher(
 
   lazy val portsMatch: Option[PortsMatch] = portsWithRoles.map(PortsMatch(_))
 
+  //scalastyle:off cyclomatic.complexity
   private[this] def portsWithRoles: Option[Seq[Option[PortWithRole]]] = {
-    val portMappings: Option[Seq[Container.Docker.PortMapping]] =
-      for {
-        c <- runSpec.container
-        pms <- c.portMappings if pms.nonEmpty
-      } yield pms
+    runSpec match {
+      case podDefinition: PodDefinition =>
+        val requiredPorts = for {
+          container <- podDefinition.containers
+          endpoint <- container.endpoints
+        } yield endpoint.hostPort
 
-    (runSpec.portNumbers, portMappings) match {
-      case (Nil, None) => // optimization for empty special case
-        Some(Seq.empty)
+        if (podDefinition.requirePorts) {
+          mappedPortRanges(requiredPorts)
+        } else {
+          randomPorts(requiredPorts.size)
+        }
+      case _ =>
+        val portMappings: Option[Seq[Container.Docker.PortMapping]] =
+          for {
+            c <- runSpec.container
+            pms <- c.portMappings if pms.nonEmpty
+          } yield pms
 
-      case (ports, Some(mappings)) =>
-        // We use the mappings from the containers if they are available and ignore any other port specification.
-        // We cannot warn about this because we autofill the ports field.
-        mappedPortRanges(mappings)
+        (runSpec.portNumbers, portMappings) match {
+          case (Nil, None) => // optimization for empty special case
+            Some(Seq.empty)
 
-      case (ports, None) if runSpec.requirePorts =>
-        findPortsInOffer(ports, failLog = true)
+          case (ports, Some(mappings)) =>
+            // We use the mappings from the containers if they are available and ignore any other port specification.
+            // We cannot warn about this because we autofill the ports field.
+            mappedPortRanges(mappings.map(_.hostPort))
 
-      case (ports, None) =>
-        randomPorts(ports.size)
+          case (ports, None) if runSpec.requirePorts =>
+            findPortsInOffer(ports, failLog = true)
+
+          case (ports, None) =>
+            randomPorts(ports.size)
+        }
     }
   }
 
@@ -101,36 +116,36 @@ class PortsMatcher(
     * Try to find all non-zero host ports in offer and use random ports from the offer for dynamic host ports (=0).
     * Return `None` if not all host ports could be assigned this way.
     */
-  private[this] def mappedPortRanges(mappings: Seq[PortMapping]): Option[Seq[Option[PortWithRole]]] = {
-    takeEnoughPortsOrNone(expectedSize = mappings.size) {
-      // non-dynamic hostPorts from port mappings
-      val hostPortsFromMappings: Set[Int] = mappings.collect {
-        case PortMapping(_, Some(hostPort), _, _, _, _) if hostPort != 0 => hostPort
-      }.toSet
+  private[this] def mappedPortRanges(ports: Seq[Option[Int]]): Option[Seq[Option[PortWithRole]]] = {
+    takeEnoughPortsOrNone(expectedSize = ports.size) {
+      // non-dynamic hostPorts
+      val staticPorts = ports.collect { case Some(port) if port != 0 => port }.toSet
 
       // available ports without the ports that have been preset in the port mappings
       val availablePortsWithoutStaticHostPorts: Iterator[PortWithRole] =
-        shuffledAvailablePorts.filter(portWithRole => !hostPortsFromMappings(portWithRole.port))
+        shuffledAvailablePorts.filter(portWithRole => !staticPorts(portWithRole.port))
 
-      mappings.iterator.map {
-        case PortMapping(containerPort, Some(hostPort), servicePort, protocol, name, labels) if hostPort == 0 =>
+      ports.iterator.map {
+        case Some(port) if port == 0 =>
           if (!availablePortsWithoutStaticHostPorts.hasNext) {
-            log.info(s"Offer [${offer.getId.getValue}]. $resourceSelector. " +
-              s"Insufficient ports in offer for run spec [${runSpec.id}]")
+            log.info(
+              s"Offer [${offer.getId.getValue}]. $resourceSelector. " +
+                s"Insufficient ports in offer for run spec [${runSpec.id}]")
             None
           } else {
             Option(availablePortsWithoutStaticHostPorts.next())
           }
-        case PortMapping(_, Some(hostPort), _, _, _, _) =>
-          offeredPortRanges.find(_.contains(hostPort)) match {
+        case Some(port) =>
+          offeredPortRanges.find(_.contains(port)) match {
             case Some(PortRange(role, _, _, reservation)) =>
-              Some(PortWithRole(role, hostPort, reservation))
+              Some(PortWithRole(role, port, reservation))
             case None =>
-              log.info(s"Offer [${offer.getId.getValue}]. $resourceSelector. " +
-                s"Cannot find range with host port ${hostPort} for run spec [${runSpec.id}]")
+              log.info(
+                s"Offer [${offer.getId.getValue}]. $resourceSelector. " +
+                  s"Cannot find range with host port ${port} for run spec [${runSpec.id}]")
               None
           }
-        case PortMapping(_, None, _, _, _, _) =>
+        case None =>
           // None has special meaning in this context: it stops the allocation process. this is a problem
           // if there's an optional host port in the middle of some mappings list. so instead of None we
           // generate Some(RequestNone) to indicate that we're not requesting a host port, but there may
--- a/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
+++ b/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
@@ -1,0 +1,106 @@
+package mesosphere.mesos
+
+import mesosphere.marathon.state.AppDefinition
+
+import scala.collection.immutable.Seq
+import scala.collection.mutable
+import scala.util.Random
+
+object EnvironmentHelper {
+  val maxEnvironmentVarLength = 512
+  val labelEnvironmentKeyPrefix = "MARATHON_APP_LABEL_"
+  val maxVariableLength = maxEnvironmentVarLength - labelEnvironmentKeyPrefix.length
+
+  // portsEnv generates $PORT{x} and $PORT_{y} environment variables, wherein `x` is an index into
+  // the portDefinitions or portMappings array and `y` is a non-zero port specifically requested by
+  // the application specification.
+  //
+  // @param requestedPorts are either declared container ports (if port mappings are specified) or host ports;
+  // may be 0's
+  // @param effectivePorts resolved non-dynamic host ports allocated from Mesos resource offers
+  // @return a dictionary of variables that should be added to a tasks environment
+  //scalastyle:off cyclomatic.complexity method.length
+  def portsEnv(
+    requestedPorts: Seq[Int],
+    effectivePorts: Seq[Option[Int]],
+    portNames: Seq[Option[String]]): Map[String, String] = {
+    if (effectivePorts.isEmpty) {
+      Map.empty
+    } else {
+      val env = Map.newBuilder[String, String]
+      val generatedPortsBuilder = Map.newBuilder[Int, Int] // index -> container port
+
+      object ContainerPortGenerator {
+        // track which port numbers are already referenced by PORT_xxx envvars
+        lazy val consumedPorts = mutable.Set(requestedPorts: _*) ++= effectivePorts.flatten
+        val maxPort: Int = 65535 - 1024
+
+        // carefully pick a container port that doesn't overlap with other ports used by this
+        // container. and avoid ports in the range (0 - 1024)
+        def next: Int = {
+          val p = Random.nextInt(maxPort) + 1025
+          if (!consumedPorts.contains(p)) {
+            consumedPorts += p
+            p
+          } else next // TODO(jdef) **highly** unlikely, but still possible that the port range could be exhausted
+        }
+      }
+
+      effectivePorts.zipWithIndex.foreach {
+        // matches fixed or dynamic host port assignments
+        case (Some(effectivePort), portIndex) =>
+          env += (s"PORT$portIndex" -> effectivePort.toString)
+
+        // matches container-port-only mappings; no host port was defined for this mapping
+        case (None, portIndex) =>
+          requestedPorts.lift(portIndex) match {
+            case Some(containerPort) if containerPort == AppDefinition.RandomPortValue =>
+              val randomPort = ContainerPortGenerator.next
+              generatedPortsBuilder += portIndex -> randomPort
+              env += (s"PORT$portIndex" -> randomPort.toString)
+            case Some(containerPort) if containerPort != AppDefinition.RandomPortValue =>
+              env += (s"PORT$portIndex" -> containerPort.toString)
+          }
+      }
+
+      val generatedPorts = generatedPortsBuilder.result
+      requestedPorts.zip(effectivePorts).zipWithIndex.foreach {
+        case ((requestedPort, Some(effectivePort)), _) if (requestedPort != AppDefinition.RandomPortValue) =>
+          env += (s"PORT_$requestedPort" -> effectivePort.toString)
+        case ((requestedPort, Some(effectivePort)), _) if (requestedPort == AppDefinition.RandomPortValue) =>
+          env += (s"PORT_$effectivePort" -> effectivePort.toString)
+        case ((requestedPort, None), _) if (requestedPort != AppDefinition.RandomPortValue) =>
+          env += (s"PORT_$requestedPort" -> requestedPort.toString)
+        case ((requestedPort, None), portIndex) if (requestedPort == AppDefinition.RandomPortValue) =>
+          val generatedPort = generatedPorts(portIndex)
+          env += (s"PORT_$generatedPort" -> generatedPort.toString)
+      }
+
+      portNames.zip(effectivePorts).foreach {
+        case (Some(portName), Some(effectivePort)) =>
+          env += (s"PORT_${portName.toUpperCase}" -> effectivePort.toString)
+        // TODO(jdef) port name envvars for generated container ports
+        case _ =>
+      }
+
+      val allAssigned = effectivePorts.flatten ++ generatedPorts.values
+      allAssigned.headOption.foreach { port => env += ("PORT" -> port.toString) }
+      env += ("PORTS" -> allAssigned.mkString(","))
+      env.result()
+    }
+  }
+
+  def labelsToEnvVars(labels: Map[String, String]): Map[String, String] = {
+
+    def escape(name: String): String = name.replaceAll("[^a-zA-Z0-9_]+", "_").toUpperCase
+
+    val validLabels = labels.collect {
+      case (key, value) if key.length < maxVariableLength
+        && value.length < maxEnvironmentVarLength => escape(key) -> value
+    }
+
+    val names = Map("MARATHON_APP_LABELS" -> validLabels.keys.mkString(" "))
+    val values = validLabels.map { case (key, value) => s"$labelEnvironmentKeyPrefix$key" -> value }
+    names ++ values
+  }
+}
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -1,0 +1,426 @@
+package mesosphere.mesos
+
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.core.pod.PodDefinition
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.raml
+import mesosphere.marathon.state.{ EnvVarString, PathId, Timestamp }
+import mesosphere.marathon.tasks.PortsMatch
+import mesosphere.mesos.ResourceMatcher.ResourceSelector
+import org.apache.mesos.{ Protos => mesos }
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable.Seq
+
+// TODO(PODS): Integrate plugin support (e.g. secrets)
+object TaskGroupBuilder {
+  val log = LoggerFactory.getLogger(getClass)
+
+  // These labels are necessary for AppC images to work.
+  // Given that Docker only works under linux with 64bit,
+  // let's (for now) set these values to reflect that.
+  val LinuxAmd64 = mesos.Labels.newBuilder
+    .addAllLabels(
+      Iterable(
+      mesos.Label.newBuilder.setKey("os").setValue("linux").build,
+      mesos.Label.newBuilder.setKey("arch").setValue("amd64").build
+    ).asJava)
+    .build
+
+  case class BuilderConfig(
+    acceptedResourceRoles: Set[String],
+    envVarsPrefix: Option[String])
+
+  def build(
+    podDefinition: PodDefinition,
+    offer: mesos.Offer,
+    runningTasks: => Iterable[Task],
+    newInstanceId: PathId => Instance.Id,
+    config: BuilderConfig
+  ): Option[(mesos.ExecutorInfo, mesos.TaskGroupInfo, Seq[Option[Int]])] = {
+    val acceptedResourceRoles: Set[String] = {
+      val roles = if (podDefinition.acceptedResourceRoles.isEmpty) {
+        config.acceptedResourceRoles
+      } else {
+        podDefinition.acceptedResourceRoles
+      }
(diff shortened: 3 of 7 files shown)


nfnt:
Some details regarding networks still need to be solved. Also there's some TODOs left.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,410 @@
...
+    builder
+  }
+
+  private[this] def taskContextEnv(
+    container: raml.MesosContainer,
+    version: Timestamp,
+    taskId: Instance.Id): Map[String, String] = {
+    Map(
+      "MESOS_TASK_ID" -> Some(taskId.idString),
+      "MARATHON_APP_ID" -> Some(container.name),
```
aquamatthias: The MARATHON_APP_ID would be the runSpecId, which is the same for all containers in the pod.
Perhaps add a MARATHON_CONTAINER_ID -> container.name

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,410 @@
...
+      "MARATHON_APP_VERSION" -> Some(version.toString),
+      "MARATHON_APP_RESOURCE_CPUS" -> Some(container.resources.cpus.toString),
+      "MARATHON_APP_RESOURCE_MEM" -> Some(container.resources.mem.toString),
+      "MARATHON_APP_RESOURCE_DISK" -> Some(container.resources.disk.toString),
+      "MARATHON_APP_RESOURCE_GPUS" -> Some(container.resources.gpus.toString)
```
aquamatthias: We probably should rename `MARATHON_APP_` to `MARATHON_CONTAINER_` ?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  private[this] def computeExecutorInfo(
+    podDefinition: PodDefinition,
+    taskId: Instance.Id): mesos.ExecutorInfo.Builder = {
+    val executorID = mesos.ExecutorID.newBuilder.setValue(f"marathon-${taskId.idString}")
+
+    val executorInfo = mesos.ExecutorInfo.newBuilder
+      .setType(mesos.ExecutorInfo.Type.DEFAULT)
+      .setExecutorId(executorID)
```
aquamatthias: I would suggest to pick the `instanceId` as value here.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,410 @@
...
+
+  private[this] def build(
+    podDefinition: PodDefinition,
+    offer: mesos.Offer,
+    newTaskId: PathId => Instance.Id,
+    config: MarathonConf,
+    resourceMatch: ResourceMatcher.ResourceMatch
+  ): Some[(mesos.ExecutorInfo, mesos.TaskGroupInfo, Seq[Option[Int]])] = {
+    // TODO: probably set unique ID for each task
+    val taskId = newTaskId(podDefinition.id)
```
aquamatthias: rename to newInstanceId

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Should we propagate the `runSpecId` as well, i.e. keep `MARATHON_APP_ID`, but set it to `runSpecId`?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jdef: I think there's a dc/os EE dependency on MARATHON_APP_ID; let's keep this
in mind in case we rename

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jdef: looks like the codes *does* use instanceId. The variable is poorly named.
s/taskId/instanceId/

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jdef: Yes, MARATHON_APP_ID should be the runSpecId

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: I kept `MESOS_TASK_ID`, `MARATHON_APP_ID` and `MARATHON_APP_VERSION` as they are using values associated with the runSpec. The rest has been renamed to `MARATHON_CONTAINER_*`.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+  private[this] def computeExecutorInfo(
+    podDefinition: PodDefinition,
+    instanceId: Instance.Id): mesos.ExecutorInfo.Builder = {
+    val executorID = mesos.ExecutorID.newBuilder.setValue(f"marathon-${instanceId.idString}")
+
+    val executorInfo = mesos.ExecutorInfo.newBuilder
+      .setType(mesos.ExecutorInfo.Type.DEFAULT)
+      .setExecutorId(executorID)
+
+    // We're deliberately not setting executor resources here. Because we're using the default executor, the
```
jdef: Who sets the executor cpu, mem? Mesos? If so .. wouldn't it be better to be explicit here? We already have constants defined that are taken into consideration during resource matching. Why not reference them here?

src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
```diff
@@ -0,0 +1,105 @@
...
+    portNames: Seq[Option[String]]): Map[String, String] = {
+    if (effectivePorts.isEmpty) {
+      Map.empty
+    } else {
+      val env = Map.newBuilder[String, String]
+      val generatedPortsBuilder = Map.newBuilder[Int, Int] // index -> container port
+
+      object ContainerPortGenerator {
+        // track which port numbers are already referenced by PORT_xxx envvars
+        lazy val consumedPorts = scala.collection.mutable.Set(requestedPorts: _*) ++= effectivePorts.flatten
```
jasongilanfarr: mutable.Set (import scala.collection above)

src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
```diff
@@ -0,0 +1,105 @@
...
+        val maxPort: Int = 65535 - 1024
+
+        // carefully pick a container port that doesn't overlap with other ports used by this
+        // container. and avoid ports in the range (0 - 1024)
+        def next: Int = {
+          val p = Random.nextInt(maxPort) + 1025
+          if (!consumedPorts.contains(p)) {
+            consumedPorts += p
+            p
+          } else next // TODO(jdef) **highly** unlikely, but still possible that the port range could be exhausted
```
jasongilanfarr: should this be handled?

jasongilanfarr: Is this unique globally or is it unique per agent?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+
+    container.exec.foreach{ exec =>
+      exec.command match {
+        case raml.ShellCommand(shell) =>
+          commandInfo.setShell(true)
+          commandInfo.setValue(shell)
+        case raml.ArgvCommand(argv) =>
+          commandInfo.setShell(false)
+          if (argv.nonEmpty) {
+            commandInfo.setValue(argv.head)
```
jdef: should check exec.overrideEntrypoint. the idea was that if overrideEntypoint == false then argv then is interpreted the parameter list for some entrypoint (so setValue would not be called). if overrideEntrypoint == true then setValue(argv[0]). In any case addAllArguments probably needs the entire argv list, not just the tail.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+      .setExecutorId(executorID)
+
+    // We're deliberately not setting executor resources here. Because we're using the default executor, the
+    // default resource values can be used.
+
+    if (podDefinition.networks.nonEmpty) {
+      val containerInfo = mesos.ContainerInfo.newBuilder
+        .setType(mesos.ContainerInfo.Type.MESOS)
+
+      // TODO: What about host networks? Is this only for labels?
```
jdef: "Is this only for labels?" .. means what?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+    val builder = mesos.TaskInfo.newBuilder
+      .setName(container.name)
+      .setTaskId(instanceId.mesosTaskId)
+      .setSlaveId(offer.getSlaveId)
+
+    builder.addResources(scalarResource("cpus", container.resources.cpus))
+    builder.addResources(scalarResource("mem", container.resources.mem))
+    builder.addResources(scalarResource("disk", container.resources.disk))
+    builder.addResources(scalarResource("gpus", container.resources.gpus.toDouble))
+
```
jdef: TODO: add host port resources

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+    val user = container.user.orElse(podDefinition.user)
+    user.foreach(commandInfo.setUser)
+
+    val uris = container.artifacts.map { artifact =>
+      val uri = mesos.CommandInfo.URI.newBuilder.setValue(artifact.uri)
+
+      artifact.cache.foreach(uri.setCache)
+      artifact.extract.foreach(uri.setExtract)
+      artifact.executable.foreach(uri.setExecutable)
+
```
jdef: what about destination path?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+    container.volumeMounts.foreach { volumeMount =>
+      podVolumes.find(_.name == volumeMount.name).map { hostVolume =>
+        val volume = mesos.Volume.newBuilder
+          .setContainerPath(volumeMount.mountPath)
+
+        hostVolume.host.foreach(volume.setHostPath)
+
+        // Read-write mode will be used when the "readOnly" option isn't set.
+        val mode = volumeMount.readOnly.getOrElse(false) match {
```
jdef: nit: use if/else vs. match/case

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+          command.setShell(true)
+          command.setValue(shell)
+        case raml.ArgvCommand(argv) =>
+          command.setShell(false)
+          if (argv.nonEmpty) {
+            command.setValue(argv.head)
+          }
+
+          if (argv.size > 1) {
+            command.addAllArguments(argv.tail.asJava)
```
jdef: are we sure that argv.tail is correct? posix normally wants arguments to include argv[0].. and this was carried over to mesos for normal task command execution. not sure if this semantic was maintained for health checks

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+
+object TaskGroupBuilder {
+  val log = LoggerFactory.getLogger(getClass)
+
+  def build(
+    podDefinition: PodDefinition,
+    offer: mesos.Offer,
+    runningTasks: => Iterable[Task],
+    newInstanceId: PathId => Instance.Id,
+    config: MarathonConf
```
jdef: can we avoid this please? create a case class instead with the config state that this builder needs

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,412 @@
...
+import org.apache.mesos.{ Protos => mesos }
+import org.slf4j.LoggerFactory
+
+import scala.collection.JavaConverters._
+import scala.collection.immutable.Seq
+
+object TaskGroupBuilder {
+  val log = LoggerFactory.getLogger(getClass)
+
+  def build(
```
jdef: probably want a TODO here for integrating plugin support (needed for secrets)

src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
jdef: this is a refactoring of what exists already on master. not unique per-cluster/agent. container ports are only used in the context of a container's network namespace. more like .. unique per runnableSpec

src/test/scala/mesosphere/mesos/EnvironmentHelperTest.scala
```diff
@@ -0,0 +1,59 @@
+package mesosphere.mesos
+
+import org.scalatest.FunSuite
+
+import scala.collection.immutable.Seq
+
+class EnvironmentHelperTest extends FunSuite {
```
jasongilanfarr: please use "UnitTest"

src/test/scala/mesosphere/mesos/TaskGroupBuilderTest.scala
```diff
@@ -0,0 +1,583 @@
...
+import mesosphere.marathon.raml
+import mesosphere.marathon.state.EnvVarString
+import mesosphere.marathon.state.PathId._
+import mesosphere.marathon.{ MarathonSpec, MarathonTestHelper }
+import org.apache.mesos.{ Protos => mesos }
+import org.scalatest.Matchers
+
+import scala.collection.JavaConversions._
+
+class TaskGroupBuilderTest extends MarathonSpec with Matchers {
```
jasongilanfarr: Please use UnitTest

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: While a name should only be set for a "container" network, what should happen here when `NetworkMode == NetworkMode.Host`? It seems that a network name shouldn't be set then, leaving only the labels as something we might be able to set here.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: From `mesos.proto, CommandInfo.shell`:

> The 'value' specified will be treated as the filename of the executable. The 'arguments' will be treated as the arguments to the executable. This is similar to how POSIX exec families launch processes (i.e., execlp(value, arguments(0), arguments(1), ...)).

So I assume this is correct here. Doing the same for health checks.

nfnt: Oh, my bad! Looks like `argv[0]` is indeed the command. Will fix it!

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Will be done at pod level in `computeExecutorInfo`.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Host networks don't need an `NetworkInfo`, hence nothing will be set here in that case.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: I want to avoid the pattern used in `TaskBuilder` here, where there's a class with a single (public) function that gets called like this: `val result = new TaskBuilder(...).build(...)`. Having that function in an object is IMO the better way here. OTOH it would make sense to create a _private_ case class in this build function that makes it easier for the other functions here to access some variables without having to pass them in the function signature.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,394 @@
...
+          val networkInfo = mesos.NetworkInfo.newBuilder
+            .setName(containerNetwork.name.get)
+
+          containerNetwork.labels.foreach { labels =>
+            networkInfo.setLabels(
+              mesos.Labels.newBuilder.addAllLabels(labels.values.map {
+                case (key, value) =>
+                  mesos.Label.newBuilder.setKey(key).setValue(value).build
+              }.asJava))
+          }
```
jeschkies: A little nit picking. I'd try to avoid too nested `foreach` calls and chain instead: 

```scala
val labels = containerNetwork.labels.iterator
    .map { case (key, value) => 
      mesos.Label.newBuilder.setKey(key).setValue(value).build
     }
     .foldLeft(mesos.Labels.newBuilder) { (builder, label) =>
       builder.addLabels(label)
     }
networkInfo.setLabels(labels)
```

or even

```scala
@inline def toMesosLabel((key, value)) = mesos.Label.newBuilder.setKey(key).setValue(value).build
@inline def addLabel(builder, label) = builder.addLabels(label)

val labels = containerNetwork.labels.iterator
    .map(toMesosLabel)
    .foldLeft(mesos.Labels.newBuilder)(addLabel)
networkInfo.setLabels(labels)
```

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,394 @@
...
+
+    if (podDefinition.labels.nonEmpty) {
+      val labels = mesos.Labels.newBuilder
+        .addAllLabels(podDefinition.labels.map {
+          case (key, value) =>
+            mesos.Label.newBuilder.setKey(key).setValue(value).build
+        }.asJava)
+
+      executorInfo.setLabels(labels)
+    }
```
jeschkies: DRY You can then reuse the method here :)

```scala
val labels = podDefinition.labels.iterator
  .map(toMesosLabel)
  .foldLeft(mesos.Labels.newBuilder)(addLabel)
  .asJava
executorInfo.setLabels(labels)
```

nfnt: Think I even found something nicer. Factored out a function that does `Map[String, String] => mesos.Labels`.

nfnt:
Except for port mappings and `DiscoveryInfo` I'd consider this good to merge now. Port mappings need to be redone from scratch. Keep in mind that because of wrong port mapping handling the test case "A TaskGroupBuilder must - support networks and port mappings for pods and containers" currently fails.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
```diff
@@ -40,27 +40,42 @@ class PortsMatcher(
...
-      case (ports, None) =>
-        randomPorts(ports.size)
+    runSpec match {
+      case podDefinition: PodDefinition =>
+        val requiredPorts = for {
+          container <- podDefinition.containers
+          endpoint <- container.endpoints
+        } yield endpoint.hostPort
+
+        if (podDefinition.requirePorts) {
```
jdef: I don't think `requirePorts` is meaningful for pods. We should always be using `mappedPortRanges`

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jdef: OK, I don't think my comment was very clear. I hate MarathonConf. I don't want to see it in new code if we can avoid it. I'd rather see a BuilderConfig case class in the parameter list here vs. MarathonConf.

src/main/scala/mesosphere/marathon/tasks/PortsMatcher.scala
nfnt: Yep, stumbled upon that as well when fixing port mappings. It's been removed.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,394 @@
...
+
+  def build(
+    podDefinition: PodDefinition,
+    offer: mesos.Offer,
+    runningTasks: => Iterable[Task],
+    newInstanceId: PathId => Instance.Id,
+    config: MarathonConf
+  ): Option[(mesos.ExecutorInfo, mesos.TaskGroupInfo, Seq[Option[Int]])] = {
+    val acceptedResourceRoles: Set[String] = {
+      val roles = if (podDefinition.acceptedResourceRoles.isEmpty) {
```
jdef: nit: use getOrElse instead of if/else

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,394 @@
...
+import scala.collection.immutable.Seq
+
+// TODO(PODS): Integrate plugin support (e.g. secrets)
+object TaskGroupBuilder {
+  val log = LoggerFactory.getLogger(getClass)
+
+  def build(
+    podDefinition: PodDefinition,
+    offer: mesos.Offer,
+    runningTasks: => Iterable[Task],
```
jdef: why this instead of Instance?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+        podDefinition.acceptedResourceRoles
+      }
+      if (log.isDebugEnabled) log.debug(s"acceptedResourceRoles $roles")
+      roles
+    }
+
+    val resourceMatchOpt: Option[ResourceMatcher.ResourceMatch] =
+      ResourceMatcher.matchResources(offer, podDefinition, runningTasks, ResourceSelector.any(acceptedResourceRoles))
+
+    resourceMatchOpt match {
```
jdef: `resourceMatchOpt.map(build(podDefinition, offer, newInstanceId, config, _))` instead?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: `acceptedResourceRoles` is a sequence, `getOrElse` doesn't work there.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+
+// TODO(PODS): Integrate plugin support (e.g. secrets)
+object TaskGroupBuilder {
+  val log = LoggerFactory.getLogger(getClass)
+
+  def build(
+    podDefinition: PodDefinition,
+    offer: mesos.Offer,
+    runningTasks: => Iterable[Task],
+    newInstanceId: PathId => Instance.Id,
```
jdef: if we pass in an instance, do we need this id generator func?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: `ResourceMatcher.matchResources` still expects `Task`, hence I'm still using it here. Maybe this should be resolved by another ticket?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -75,6 +78,30 @@ object TaskGroupBuilder {
...
+        if (mapping.containerPort.isEmpty || mapping.containerPort.get == 0) {
+          mesos.NetworkInfo.PortMapping.newBuilder
+            .setHostPort(hostPort)
+            .setContainerPort(hostPort).build
+        } else {
+          val portMapping = mesos.NetworkInfo.PortMapping.newBuilder
+            .setHostPort(hostPort)
+          mapping.containerPort.foreach(portMapping.setContainerPort)
+
+          // TODO(nfnt): set the protocol
```
jdef: .. and there might in multiple protocols, which case the mapping is (nearly) duplicated for each protocol.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jdef: yes, ok

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+      .setName(container.name)
+      .setTaskId(mesos.TaskID.newBuilder.setValue(instanceId.idString))
+      .setSlaveId(offer.getSlaveId)
+
+    builder.addResources(scalarResource("cpus", container.resources.cpus))
+    builder.addResources(scalarResource("mem", container.resources.mem))
+    builder.addResources(scalarResource("disk", container.resources.disk))
+    builder.addResources(scalarResource("gpus", container.resources.gpus.toDouble))
+
+    val labels = podDefinition.labels ++ container.labels.map(_.values).getOrElse(Map.empty[String, String])
```
jdef: someone else also thought that it would work this way. I always imagined that the pod definition labels would be placed into the executorInfo and that the container labels would be placed into the taskInfo. they would not be aggregated and then placed into taskInfo.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+
+          networkInfo.addAllPortMappings(portMappings.asJava)
+
+          containerInfo.addNetworkInfos(networkInfo)
+        }
+
+      executorInfo.setContainer(containerInfo)
+    }
+
+    executorInfo.setLabels(toMesosLabels(podDefinition.labels))
```
jdef: yes, this

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+
+    container.exec.foreach{ exec =>
+      exec.command match {
+        case raml.ShellCommand(shell) =>
+          commandInfo.setShell(true)
+          commandInfo.setValue(shell)
+        case raml.ArgvCommand(argv) =>
+          commandInfo.setShell(false)
+          commandInfo.addAllArguments(argv.asJava)
+          argv.headOption.foreach(commandInfo.setValue)
```
jdef: should only do this if exec.overrideEntrypoint is true

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+      labelEnvVars ++
+      portsEnvVars)
+      .map {
+        case (name, value) =>
+          mesos.Environment.Variable.newBuilder.setName(name).setValue(value).build
+      }
+
+    commandInfo.setEnvironment(mesos.Environment.newBuilder.addAllVariables(envVars.asJava))
+
+    commandInfo
```
jdef: nit: isn't this line redundant?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Yes, at the moment. @unterstein is working on providing instance ID and separate task IDs. Though it's not clear if that'll get rid of the generator.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+          val docker = mesos.Image.Docker.newBuilder.setName(im.id)
+
+          image.setType(mesos.Image.Type.DOCKER).setDocker(docker)
+        case raml.ImageType.Appc =>
+          // These labels are necessary for AppC images to work.
+          // Given that Docker only works under linux with 64bit,
+          // let's (for now) set these values to reflect that.
+          val labels = mesos.Labels.newBuilder
+            .addAllLabels(
+              Iterable(
```
jdef: nit: can we extract this hard-coded sequence into a val? something like `val LinuxAmd64 = Seq(...)` ?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Exactly! I'm looking for a nice way to do that. I.e. add one or two `NetworkInfo.PortMapping` instances here, depending on the set protocols. But also add one when protocols is empty.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -0,0 +1,423 @@
...
+
+    containerInfo
+  }
+
+  private[this] def computeHealthCheck(
+    healthCheck: raml.HealthCheck,
+    endpoints: Seq[raml.Endpoint]): mesos.HealthCheck.Builder = {
+    val builder = mesos.HealthCheck.newBuilder
+
+    if (healthCheck.command.isDefined) {
```
jdef: would rather see healthCheck.command.foreach { ... } (and likewise for healthCheck.http and .tcp); leave it up to some other validator to enforce the "only 1 can be set rule" and get rid of the if isDefined checking here

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: I was under the impression that they would be aggregated in `TaskInfo`.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Oh, forgot about `overrideEntrypoint`. Will fix that ASAP.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: But I'll remove them here.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: But it then also needs a `None` case so it's `resourceMatchOpt.map(build(podDefinition, offer, newInstanceId, config, _)).getOrElse(None)`

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: Got it!

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: This seems to be true only for `MesosExec`, not `CommandHealthCheck`. It's been changed there -> `computeCommandInfo`.

jdef:
lgtm

jasongilanfarr:
Future: I feel like a lot of this stuff should actually move over into PodDef and AppDef respectively. Then we just ask PodDef/AppDef "give me your __xx__ things in mesos format"

PR #4361, opened 2016-09-13
Fixes #4360 by introducing a ContainerSpec which gets implemented by …

### LINKED ISSUE #4360: Secret Interface for pods

aquamatthias:
…the MesosContainer and AppDefinition.

Also change the plugin interface of RunSpecTaskProcessor.

### DIFF
--- a/plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
+++ b/plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
@@ -1,14 +1,47 @@
 package mesosphere.marathon.plugin
 
 /**
+  * A Marathon Container definition
+  */
+trait ContainerSpec {
+  /**
+    * The user to execute the container task
+    */
+  def user: Option[String]
+
+  /**
+    * The environment of this container.
+    */
+  def env: Map[String, EnvVarValue]
+
+  /**
+    * The labels in that container
+    */
+  def labels: Map[String, String]
+}
+
+/**
   * A Marathon RunSpec Definition
   */
 trait RunSpec {
+
+  /**
+    * The uniqie id of this run specification
+    */
   def id: PathId
-  def user: Option[String]
-  def env: Map[String, EnvVarValue]
-  def labels: Map[String, String]
+
+  /**
+    * The Mesos resource roles that are accepted
+    */
   def acceptedResourceRoles: Set[String]
-  /** secrets maps a secret ID to a secret specification; a secret ID is unique within the context of the [[RunSpec]] */
+
+  /**
+    * All defined secret definitions
+    */
   def secrets: Map[String, Secret]
+
+  /**
+    * All containers of this run specification
+    */
+  def containers: Seq[ContainerSpec]
 }
--- a/plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
+++ b/plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
@@ -1,26 +1,6 @@
 package mesosphere.marathon.plugin
 
 /**
-  * A Marathon Container definition
-  */
-trait ContainerSpec {
-  /**
-    * The user to execute the container task
-    */
-  def user: Option[String]
-
-  /**
-    * The environment of this container.
-    */
-  def env: Map[String, EnvVarValue]
-
-  /**
-    * The labels in that container
-    */
-  def labels: Map[String, String]
-}
-
-/**
   * A Marathon RunSpec Definition
   */
 trait RunSpec {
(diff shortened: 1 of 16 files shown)


plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
```diff
@@ -1,14 +1,47 @@
 package mesosphere.marathon.plugin
 
 /**
-  * A Marathon RunSpec Definition
+  * A Marathon Container definition
   */
-trait RunSpec {
-  def id: PathId
+trait ContainerSpec {
```
jasongilanfarr: vals not defs.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
```diff
@@ -1,14 +1,47 @@
...
+  /**
+    * The labels in that container
+    */
   def labels: Map[String, String]
+}
+
+/**
+  * A Marathon RunSpec Definition
+  */
+trait RunSpec {
```
jasongilanfarr: vals, not defs.

src/main/scala/mesosphere/marathon/core/pod/MesosContainer.scala
```diff
@@ -0,0 +1,79 @@
...
+    case state.EnvVarSecretRef(secret) => raml.EnvVarSecretRef(secret)
+  }
+
+  def apply(c: PodContainer): MesosContainer = MesosContainer(
+    name = c.name,
+    exec = c.exec,
+    resources = c.resources,
+    endpoints = c.endpoints,
+    image = c.image,
+    env = c.environment.fold(Map.empty[String, state.EnvVarValue])(_.values.map{ case (k, v) => k -> toStateEnv(v) }),
```
jasongilanfarr: not using a helper like the others?

src/main/scala/mesosphere/marathon/state/RunSpec.scala
```diff
@@ -75,4 +75,5 @@ trait RunSpec extends plugin.RunSpec {
   // TODO(
   val fetch = Seq.empty[FetchUri]
   val portNames = Seq.empty[String]
+  def user: Option[String]
```
jasongilanfarr: val

jasongilanfarr:
I don't see the value of this change. Each pod container may have different users and environment variables etc. While I think its great that we are trying to be as generic as possible, I think we're going way too far with it and actually making mistakes instead of pattern matching when they are _different_. I put the TODOs at the bottom of RunSpec trying to say "we should remove _all of these fields_" and pattern match when they are necessary.

aquamatthias:
@jasongilanfarr 
1) vals versus def: http://stackoverflow.com/questions/19642053/when-to-use-val-or-def-in-scala-traits
The recommendation is to use def in the abstract trait, which can be implemented as vals (which is what we do)
2) state.RunSpec I totally agree. We need to get rid of this. The added method here has been deleted in the plugin.RunSpec (external interface) and added to the state.RunSpec (internal interface). 

The value of this change is: have a plugin interface, that allows for TaskInfo/TaskGroup customization that is needed for the EE Plugins.

jasongilanfarr:
I've read that article before and while interesting the restriction of
stable value is indicative of the intent. Val says it will never change,
def says it can.

As for the plugin interface, the base trait in the plugin probably needs
something more so that pods and apps can be differentiated. I don't think
container is the right approach, since pods can have many and apps may not
even have one. I'm inclined to suggest having appspec and podspec traits in
the plugin interface.

aquamatthias:
Split the RunSpec into ApplicationSpec and PodSpec.

src/main/scala/mesosphere/marathon/core/pod/MesosContainer.scala
```diff
@@ -0,0 +1,85 @@
...
+  volumeMounts: scala.collection.immutable.Seq[VolumeMount] = Nil,
+  artifacts: scala.collection.immutable.Seq[Artifact] = Nil, //TODO(PODS): use FetchUri
+  labels: Map[String, String] = Map.empty,
+  lifecycle: Option[Lifecycle] = None) extends ContainerSpec
+
+object MesosContainer {
+
+  //TODO(PODS): find a better place for this converters (should probably live in the API)
+
+  def toStateEnv(envVars: EnvVars): Map[String, state.EnvVarValue] = {
```
jasongilanfarr: We should probably have the conversion in state.EnvVarValue

jasongilanfarr:
lgtm

plugin-interface/src/main/scala/mesosphere/marathon/plugin/task/RunSpecTaskProcessor.scala
```diff
@@ -1,18 +1,25 @@
...
-import mesosphere.marathon.plugin.RunSpec
+import mesosphere.marathon.plugin.{ ApplicationSpec, PodSpec }
 import mesosphere.marathon.plugin.plugin.Plugin
-import org.apache.mesos.Protos.TaskInfo
+import org.apache.mesos.Protos.{ TaskGroupInfo, TaskInfo }
 
 /**
   * RunSpecTaskProcessor mutates a Mesos task info given some app specification.
   */
 trait RunSpecTaskProcessor extends Plugin {
```
jdef: nit: probably need a better name for this

jdef:
nit, otherwise lgtm

aquamatthias:
Rebased and merged

PR #4364, opened 2016-09-13
introduced Task.Id#instanceId

unterstein:
Introduced instanceId as calculated value from Task.Id#idString and removed Instande.Id(Task.Id)

Task.Id#forRunSpec(PathId, Instance.Id) should be used by TaskGroupBuilder

---
This change is [Reviewable](https://reviewable.io/reviews/mesosphere/marathon/4364)

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -123,12 +123,12 @@ object Task {
 
   object Id {
     private val TaskIdRegex = """^(.+)[\._]([^_\.]+)$""".r
-    private val TaskIdWithInstanceIdRegex = """^(.+)\.(.+)[\._]([^_\.]+)$""".r
+    private val TaskIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)$""".r
     private val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
 
     def runSpecId(taskId: String): PathId = {
       taskId match {
-        case TaskIdWithInstanceIdRegex(runSpecId, instanceId, uuid) => PathId.fromSafePath(runSpecId)
+        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceId, uuid) => PathId.fromSafePath(runSpecId)
         case TaskIdRegex(runSpecId, uuid) => PathId.fromSafePath(runSpecId)
         case _ => throw new MatchError(s"taskId $taskId is no valid identifier")
       }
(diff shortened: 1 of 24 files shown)


jdef:
Review status: 0 of 17 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/main/scala/mesosphere/marathon/core/task/Task.scala, line 126 at r2](https://reviewable.io:443/reviews/mesosphere/marathon/4364#-KRhDzREUlRVqIL3Vg5r:-KRhDzREUlRVqIL3Vg5s:b-k1pr35) ([raw file](https://github.com/mesosphere/marathon/blob/ac8c01fbc219d70f4558752b9d664aa7db1e87df/src/main/scala/mesosphere/marathon/core/task/Task.scala#L126)):*
> ```Scala
>     private val executorIdPrefix = "instance-"
>     private val TaskIdRegex = """^(.+)[\._]([^_\.]+)$""".r
>     private val TaskIdWithInstanceIdRegex = ("""^(.+)\.""" + executorIdPrefix + """(.+)[\._]([^_\.]+)$""").r
> ```

probably a minor point, but I think it would be great to NOT have to use something as complicated as a regex to parse our IDs. Instead if we could standardize on a single delimiter to partition ID segments and just split a string, I think that would be (1) easier to grok; (2) easier to deal with when integrating external scripting/tools; (3) more performant.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mesosphere/marathon/4364)*

jeschkies:
retest this please

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -115,35 +115,51 @@ object Task {
...
   object Id {
-    private val runSpecDelimiter = "."
     private val TaskIdRegex = """^(.+)[\._]([^_\.]+)$""".r
+    private val TaskIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)$""".r
     private val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
 
     def runSpecId(taskId: String): PathId = {
       taskId match {
+        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceId, uuid) => PathId.fromSafePath(runSpecId)
         case TaskIdRegex(runSpecId, uuid) => PathId.fromSafePath(runSpecId)
```
aquamatthias: Can we rename this to .*Legacy.*

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -115,35 +115,51 @@ object Task {
...
     }
 
+    def forInstanceId(instanceId: Instance.Id): Id = Id(instanceId.createTaskId())
+
     implicit val taskIdFormat = Format(
       Reads.of[String](Reads.minLength[String](3)).map(Task.Id(_)),
       Writes[Task.Id] { id => JsString(id.idString) }
     )
+
+    def calculateLegacyExecutorId(taskId: String): String = s"marathon-$taskId"
```
aquamatthias: Please add a note why we need this, and when we can remove this.

aquamatthias:
LGTM! Thanks! Wait for the tests to get green.

aquamatthias:
THANKS!

PR #4367, opened 2016-09-13
instead of using a CSV-formatted string for multiple protocols, make protocol an array

jdef:
I noticed that the comments don't include validation-related restrictions (like pattern) in the generated code for protocol

goal: avoid the need for CSV'd protocol values, xref #4284


jasongilanfarr:
Perhaps I'm mistaken, but if this is used for app definitions, it will break compatibility.

jdef:
it's not used for apps. endpoints are for pods

jasongilanfarr:
Lgtm then

PR #4369, opened 2016-09-13
WIP: Introduce a little more BDD for task builder tests.

jeschkies:
So it turns out we do have a lot of tests on ports and their mapping. However, it's not always clear what case they test. I tried to make it a little more explicit. There are a lot of tests so I would split them up into different suites and then mix them with

```scala
TaskBuilderTest extends Suites(
  new EnvironmentSuite,
  new TaskGenerationSuite,
  ...
)
```

Also, it might be easier to read if we use `FunSpec`

```scala
class Task GenerationSuite extends FunSpec {
  describe("TaskBuilder.buildIfMatches") {
    it("should add the network-scope label if no host port defined") {
      Given("an offer")
      // etc ...
    }
  }
}
```

Which then prints as

```
TaskBuilder.buildIfMatches
  - should add the network-scope label if no host port defined
    + Given an offer
    + When ...
    + Then
  - should ...
  - ...
```

So it would be a little more behaviour driven. However, I don't like mixing different test styles across a project too much.

Overall it just seems hard to see if we cover all use cases.


jasongilanfarr:
Could use UnitTest instead...

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -13,701 +14,952 @@ import mesosphere.mesos.protos.{ Resource, TaskID, _ }
...
+  new TaskBuilderDockerContainerSuite,
+  new TaskBuilderMesosContainerSuite,
+  new TaskBuilderEnvironmentSuite
+)
 
-  import mesosphere.mesos.protos.Implicits._
-
-  val labels = Map("foo" -> "bar", "test" -> "test")
+trait TaskBuilderSuiteBase extends UnitTestLike
+    with AppendedClues {
```
jasongilanfarr: Feel free to put AppendedClues into UnitTestLike

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -13,701 +14,952 @@ import mesosphere.mesos.protos.{ Resource, TaskID, _ }
...
+    assert(envVars.exists(v => v.getName == "HOST" && v.getValue == offer.getHostname))
+    assert(envVars.exists(v => v.getName == "PORT0" && v.getValue.nonEmpty))
+    assert(envVars.exists(v => v.getName == "PORT1" && v.getValue.nonEmpty))
+    assert(envVars.exists(v => v.getName == "PORT_8080" && v.getValue.nonEmpty))
+    assert(envVars.exists(v => v.getName == "PORT_8081" && v.getValue.nonEmpty))
 
-    assert(task.isDefined)
+    val exposesFirstPort =
+      envVars.find(v => v.getName == "PORT0").get.getValue == envVars.find(v => v.getName == "PORT_8080").get.getValue
+    assert(exposesFirstPort)
```
jasongilanfarr: Why not use a should marcher?

jdef:
much easier to read the tests, thanks for taking this on

PR #4373, opened 2016-09-13
Aligning Instance/Task updates

meichstedt:
* changed Instance to have a Map of tasks instead of a Seq
* Introduced specific InstanceUpdateOperation/Effect and TaskUpdateOperation/Effect

still kinda work in progress, feedback welcome. I'll see whether I can separate concerns and prevent propagation of these classes like they are right now.

---
This change is [Reviewable](https://reviewable.io/reviews/mesosphere/marathon/4373)


src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -55,14 +55,14 @@ class TaskKiller @Inject() (
   }
 
   private[this] def expunge(tasks: Iterable[Instance])(implicit ec: ExecutionContext): Future[Unit] = {
-    // Note: We process all tasks sequentially.
+    // Note: We process all instances sequentially.
 
-    tasks.foldLeft(Future.successful(())) { (resultSoFar, nextTask) =>
+    tasks.foldLeft(Future.successful(())) { (resultSoFar, nextInstance) =>
```
jasongilanfarr: Is there a reason we need to do this in serial? If so, maybe document way? Also, perhaps a good time to switch over to Future[Done]?

src/main/scala/mesosphere/marathon/core/flow/impl/OfferMatcherLaunchTokensActor.scala
```diff
@@ -54,7 +54,7 @@ private class OfferMatcherLaunchTokensActor(
   private[this] def healthy(status: TaskStatus): Boolean = !status.hasHealthy || status.getHealthy
 
   override def receive: Receive = {
-    case TaskChanged(InstanceStateOp.MesosUpdate(task, InstanceStatus.Running, mesosStatus, timestamp), stateChange) //
+    case TaskChanged(MesosUpdate(task, InstanceStatus.Running, mesosStatus, timestamp), stateChange) //
```
jasongilanfarr: no need for // at the end of the line

jasongilanfarr:
Only minor comments on my end.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
meichstedt: I have no idea why, the note doesn't tell unfortunately. I'll see what I can do given time constraints, but I think this will be a follow up for later

src/main/scala/mesosphere/marathon/core/flow/impl/OfferMatcherLaunchTokensActor.scala
meichstedt: ack

jdef:
boom!

PR #4375, opened 2016-09-13
Fixes #4370 Create InstanceRepository and Migrate Tasks

### LINKED ISSUE #4370: Create InstanceRepository and migrate tasks
Subtasks:
- [x] Create an `InstanceRepository`
- [x] Migrate tasks -> instances

Pods/TaskGroups have several containers that need to be created/updated/deleted. To prevent multi-step reads/writes to the persistence layer when updating instances, it makes sense to introduce a

```
trait InstanceRepository extends Repository[Instance.Id, Instance] {
  def instances(runSpecId: PathId): Source[Instance.Id, NotUsed] = {
    ids().filter(_.runSpecId == runSpecId)
  }
}
```

in `Repositories`. That also solves the problem of recreating instances in-mem after a failover.

We then need a migration that loads all existing tasks, converts them to instances (incl an executorId which can be inferred from the taskId), stores them in the `InstanceRepository` and deletes them from the `TaskRepository` afterwards.

jasongilanfarr:


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -100,6 +106,145 @@ sealed trait Task {
 
 object Task {
 
+  import mesosphere.marathon.api.v2.json.Formats.{ TimestampFormat, PathIdFormat }
```
jdef: hard to review when huge blocks like this are re-ordered.  maybe you could do the re-ordering in a separate commit?

unterstein: As far as I can see, this is just a re-ordering plus the Formats sections.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo1_4_PersistenceStore.scala
```diff
@@ -87,9 +88,33 @@ class MigrationTo1_4_PersistenceStore(migration: Migration)(implicit
...
-    migrateRepo(oldRepo, taskRepository).map("tasks" -> _)
+  def migrateTasks(legacyStore: LegacyStorageConfig, taskRepository: TaskRepository,
+    instanceRepository: InstanceRepository): Future[(String, Int)] = {
+    val oldTaskRepo = TaskRepository.legacyRepository(legacyStore.entityStore[MarathonTaskState])
+    val oldInstanceRepo = InstanceRepository.legacyRepository(legacyStore.entityStore[Instance])
+    async {
+      // first, migrate from the legacy task repositories, then overwrite with legacy instance store -> new one
+      // basically, only one of the two existed.
+      val oldTaskIds = await(oldTaskRepo.ids().runWith(Sink.seq))
+      val newTaskIds = await(taskRepository.ids().runWith(Sink.seq))
```
unterstein: this val is never used

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -1,20 +1,46 @@
...
+  }
+  override def toProto: Protos.Json = {
+    Protos.Json.newBuilder().setJson(Json.stringify(Json.toJson(this))).build()
+  }
+  override def version: Timestamp = Timestamp.zero
 }
 
 object Instance {
 
+  import mesosphere.marathon.api.v2.json.Formats
```
aquamatthias: we should discourage imports from the api.
this will break once the api is a separate build unit.

src/main/scala/mesosphere/marathon/core/task/Task.scala
jasongilanfarr: I just moved stuff. I think that when we're talking about "moving stuff around in a separate commit because it's hard to review" means "our review tool is lacking". I didn't even know I needed to move stuff around until 80-90% through the work.

On that note, can we please stop using forward references? Eg if you use Task.Id, declare it first.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jasongilanfarr: We should actually just delete Formats. The correct TODO would be to move Timestamp Json into the companion.

PR #4382, opened 2016-09-14
replace constraint matching with Placed instead of Task

jdef:


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -100,6 +101,17 @@ sealed trait Task {
 
 object Task {
 
+  object Implicits {
+    import scala.language.implicitConversions
+
+    implicit def tasksToAllPlaced(tasks: Seq[Task]): Seq[Placed] = tasks.map(taskAsPlaced(_))(collection.breakOut)
```
aquamatthias: We try to avoid implicit type conversions, since this happens "magically" behind the scenes.
```
val task: Task
def foo(placed: Placed) {}
foo(task)
```

What about defining Task as Placed?

jdef: I disliked the idea of expanding the Task API. In this case the implicits are wrapped in a lib that must be explicitly imported.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -110,6 +110,19 @@ object Task {
   // TODO PODs remove api import
   import mesosphere.marathon.api.v2.json.Formats.{ TimestampFormat, PathIdFormat }
 
+  /*
```
jdef: remove unused code

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -132,5 +134,11 @@ object Instance {
   implicit val idFormat = Json.format[Instance.Id]
   implicit val instanceStatusFormat = Json.format[InstanceStatus]
   implicit val instanceStateFormat = Json.format[InstanceState]
-  implicit val instanceJsonFormat = Json.format[Instance]
+  implicit val instanceJsonFormat: Format[Instance] = Json.format[Instance]
+
+  import mesosphere.mesos.Placed
+  implicit class PlacedInstance(instance: Instance) extends Placed {
```
jdef: @aquamatthias what about this kind of implict conversion.. ok?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -132,5 +134,11 @@ object Instance {
   implicit val idFormat = Json.format[Instance.Id]
   implicit val instanceStatusFormat = Json.format[InstanceStatus]
   implicit val instanceStateFormat = Json.format[InstanceState]
-  implicit val instanceJsonFormat = Json.format[Instance]
+  implicit val instanceJsonFormat: Format[Instance] = Json.format[Instance]
+
+  import mesosphere.mesos.Placed
```
jasongilanfarr: Please put imports at the top of the file.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jasongilanfarr: I'm not personally a fan.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -102,15 +100,19 @@ class InstanceOpFactoryImpl(
       val maybeVolumeMatch = PersistentVolumeMatcher.matchVolumes(offer, runSpec, request.reserved)
 
       maybeVolumeMatch.flatMap { volumeMatch =>
+
         // we must not consider the volumeMatch's Reserved task because that would lead to a violation of constraints
         // by the Reserved task that we actually want to launch
-        val tasksToConsiderForConstraints = tasks.filter(_.taskId != volumeMatch.task.taskId)
+        val instancesToConsiderForConstraints = instances.values.filter { inst =>
+          inst.tasks.filter(_.taskId != volumeMatch.task.taskId).nonEmpty
+        }.toVector
```
jasongilanfarr: should be able to use breakOut here

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -102,15 +100,19 @@ class InstanceOpFactoryImpl(
       val maybeVolumeMatch = PersistentVolumeMatcher.matchVolumes(offer, runSpec, request.reserved)
 
       maybeVolumeMatch.flatMap { volumeMatch =>
+
         // we must not consider the volumeMatch's Reserved task because that would lead to a violation of constraints
         // by the Reserved task that we actually want to launch
-        val tasksToConsiderForConstraints = tasks.filter(_.taskId != volumeMatch.task.taskId)
+        val instancesToConsiderForConstraints = instances.values.filter { inst =>
+          inst.tasks.filter(_.taskId != volumeMatch.task.taskId).nonEmpty
```
jasongilanfarr: exists?

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -42,7 +47,7 @@ object Constraints {
       s"{$s}"
   }
 
-  private final class ConstraintsChecker(tasks: Iterable[Task], offer: Offer, constraint: Constraint) {
+  private final class ConstraintsChecker[T <% Placed](allPlaced: Seq[T], offer: Offer, constraint: Constraint) {
```
jasongilanfarr: uhh, what? Viewable as? Aren't views super deprecated? Prefer either ```T <: Placed``` or ```T : Placed```

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -146,85 +151,86 @@ object Constraints {
...
         //sort all distributions in descending order based on distribution difference
-        .toSeq.sortBy(_.distributionDifference(toKillTasks)).reverseIterator
-        //select tasks to kill (without already selected ones)
-        .flatMap(_.tasksToKillIterator(toKillTasks)) ++
-        //fallback: if the distributions did not select a task, choose one of the not chosen ones
-        runningInstances.iterator.filterNot(task => toKillTasks.contains(task.instanceId))
-
-      val matchingTask =
-        tried.find(tryTask => distributions.forall(_.isMoreEvenWithout(toKillTasks + (tryTask.instanceId -> tryTask))))
+        .toSeq.sortBy(_.distributionDifference(toKillInstances)).reverseIterator
```
jasongilanfarr: Prefer passing in an Ordering that reverses the sort.

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -146,85 +151,86 @@ object Constraints {
...
+        //fallback: if the distributions did not select a instance, choose one of the not chosen ones
+        runningInstances.iterator.filterNot(instance => toKillInstances.contains(instance.instanceId))
+
+      val matchingInstance = tried.find { tryInstance =>
+        distributions.forall(_.isMoreEvenWithout(toKillInstances + (tryInstance.instanceId -> tryInstance)))
+      }
 
-      matchingTask match {
-        case Some(task) => toKillTasks += task.instanceId -> task
+      matchingInstance match {
```
jasongilanfarr: matchingInstances.fold { flag = false ) { ... }?

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -146,85 +151,86 @@ object Constraints {
...
+        runningInstances.groupBy(groupFn).values.map(Instance.instancesById)(collection.breakOut)
+      GroupByDistribution(constraint, instanceGroups)
     }
 
     //short circuit, if there are no constraints to align with
     if (distributions.isEmpty) return Set.empty
 
-    var toKillTasks = Map.empty[Instance.Id, Instance]
+    var toKillInstances = Map.empty[Instance.Id, Instance]
     var flag = true
```
jasongilanfarr: what's flag mean?

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -237,10 +243,10 @@ object Constraints {
       diffAfterKill <= 1 || distributionDifference() > diffAfterKill
     }
 
-    def tasksToKillIterator(without: Map[Instance.Id, Instance]): Iterator[Instance] = {
+    def instancesToKillIterator(without: Map[Instance.Id, Instance]): Iterator[Instance] = {
```
jasongilanfarr: why in the word are we giving an iterator?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -35,7 +34,7 @@ object TaskGroupBuilder {
   def build(
     podDefinition: PodDefinition,
     offer: mesos.Offer,
-    runningTasks: => Iterable[Task],
+    runningInstances: => Seq[Instance],
```
jasongilanfarr: could be better in a second parameter list as it would let you do:

```
build(pod, offer, someNewInstanceId, config) { blahBlah to get instances }
```

src/test/scala/mesosphere/marathon/core/instance/InstanceTestHelper.scala
```diff
@@ -0,0 +1,12 @@
+package mesosphere.marathon.core.instance
+
+import mesosphere.marathon.core.task.Task
+
+import scala.collection.immutable.Seq
+
+trait InstanceTestHelper {
```
jasongilanfarr: seems super useless to me.

jasongilanfarr: Why don't we have "Instance.apply(Seq[Tasks])"?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jdef: I think I tried and the compiler complained. I'll try again

src/test/scala/mesosphere/marathon/core/instance/InstanceTestHelper.scala
jdef: I didn't want to pollute the non-test API with helper code needed for tests. Is there a better plan/pattern for such?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jasongilanfarr: I have a new cleanup change that's going to rip out JavaCoverters and JavaConversions in favor of minimally copied collections (eg, go directly to the target type). Furthermore, if we transform the collection, the transform will be performed in the java8 stream API and "collected" into the Scala collection instead. I'm still working through how to make this not ugly, eg, I may make a rich stream that understands Scala types and lambas better (since we don't have 2.12 yet)

src/main/scala/mesosphere/mesos/Constraints.scala
jdef: not sure. i'm trying to avoid large amounts of refactoring because the focus is on enabling pods vs. fixing all the little things that jump out .. with a few exceptions (like preferring immutable.Seq)

src/test/scala/mesosphere/marathon/core/instance/InstanceTestHelper.scala
jasongilanfarr: I think it's useful enough and may be useful in Instance anyways. If you don't agree, maybe follow the pattern common in scalatest "InstanceSupport" and force a self-type?

src/main/scala/mesosphere/mesos/Constraints.scala
jasongilanfarr: I lean toward "leave it better than it was". This is code rot imo.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jdef: Do you have a suggested alternative?

src/main/scala/mesosphere/mesos/Constraints.scala
jasongilanfarr: Don't see anything reading flag again, so this could just be a foreach.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jasongilanfarr: I don't even fully understand what you're doing. I haven't used view bounds since 2.8, so this is just funky.

jasongilanfarr: As far as I can read, this just adds the marker trait, so if I follow the logic (whether it's view/context or implicit), if I have the implicit placement in my list, it's automatically now tagged? This seems wrong to me.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jdef: yep, compiler complained again

src/main/scala/mesosphere/mesos/Constraints.scala
jdef: ok, will change to immutable.Seq

PR #4388, opened 2016-09-15
Fix pod deployment.

nfnt:
For pods Mesos no longer sets the framework ID implicitly in
'ExecutorInfo'. Also, disk resources are validated and need to be set.
Because the default executor doesn't use disk resources, they are set to
zero.

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -2,6 +2,7 @@ package mesosphere.mesos
 
 import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.pod.{ ContainerNetwork, MesosContainer, PodDefinition }
+import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.raml
 import mesosphere.marathon.state.{ EnvVarString, PathId, Timestamp }
 import mesosphere.marathon.tasks.PortsMatch
@@ -70,7 +71,12 @@ object TaskGroupBuilder {
 
     val portMappings = computePortMappings(allEndpoints, resourceMatch.hostPorts)
 
-    val executorInfo = computeExecutorInfo(podDefinition, resourceMatch.portsMatch, portMappings, instanceId)
+    val executorInfo = computeExecutorInfo(
+      podDefinition,
+      resourceMatch.portsMatch,
+      portMappings,
+      instanceId,
+      offer.getFrameworkId)
 
     val envPrefix: Option[String] = config.envVarsPrefix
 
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -157,16 +162,19 @@ object TaskGroupBuilder {
...
     val executorID = mesos.ExecutorID.newBuilder.setValue(f"marathon-${instanceId.idString}")
 
     val executorInfo = mesos.ExecutorInfo.newBuilder
       .setType(mesos.ExecutorInfo.Type.DEFAULT)
       .setExecutorId(executorID)
+      .setFrameworkId(frameworkId)
 
     executorInfo.addResources(scalarResource("cpus", PodDefinition.DefaultExecutorCpus))
     executorInfo.addResources(scalarResource("mem", PodDefinition.DefaultExecutorMem))
+    executorInfo.addResources(scalarResource("disk", 0))
```
jdef: if disk space is ever enforced, I wonder how it will be enforced for pods. for example, pod subcontainers can already specify disk resources but we don't add any "default" disk resources for the executor. but there are almost certainly log files that will accumulate in the executor's sandbox. and the executor sandbox will also hold all of the subcontainer sandboxes .. hopefully those won't count against the pod executor's disk resource allocation.

nfnt: Looking at the [review that introduced that check in Mesos](https://reviews.apache.org/r/51270/), it seems that this will indeed be enforced.

nfnt: So this resource probably shouldn't be hardcoded here, but part of the pod definition...

nfnt: For now, I've set it to 10MB.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -142,7 +148,11 @@ object TaskGroupBuilder {
 
     val containerInfo = computeContainerInfo(podDefinition.podVolumes, container)
 
-    builder.setContainer(containerInfo)
+    // Workaround a validation problem with Mesos.
+    // Only set 'ContainerInfo' if it is non-empty.
+    if (podDefinition.podVolumes.nonEmpty || container.image.isDefined) {
```
meichstedt: Can you add a link in case of an existing Mesos Jira?

meichstedt:
LGTM, see 1 comment

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
nfnt: I've changed the code a bit, but we need to make sure, that `ContainerInfo` is only set when some of it's fields are set. Validation rules are much stricter for pods.

PR #4390, opened 2016-09-15
conversion API w/ some initial converters

jdef:
* generic conversion API
* network conversion
* envvar conversion
* internal Network object
* better network API validation
* pod conversion
* container conversion


src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -33,10 +33,23 @@ trait PodsValidation {
...
+  } and isTrue[Seq[Network]]("Duplicate networks are not allowed") { nets =>
+    // unnamed CT nets pick up the default virtual net name
+    val unnamedAtMostOnce = nets.count { n => n.name.isEmpty && n.mode == NetworkMode.Container } < 2
     val realNamesAtMostOnce: Boolean = !nets.flatMap(_.name).groupBy(name => name).exists(_._2.size > 1)
     unnamedAtMostOnce && realNamesAtMostOnce
+  } and isTrue[Seq[Network]]("May specify a single host network, or 1-to-n container networks, but not both") { nets =>
+    val countsByMode = nets.groupBy { net => net.mode }.mapValues(_.size)
+    val hostNetworks = countsByMode.get(NetworkMode.Host).getOrElse(0)
+    val containerNetworks = countsByMode.get(NetworkMode.Container).getOrElse(0)
+    (hostNetworks == 1 && containerNetworks == 0) || (hostNetworks == 0 && containerNetworks > 0)
```
aquamatthias: if you specify no host network and no container network, this will fail

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -40,22 +40,7 @@ class PodsResource @Inject() (
...
-        } else {
-          network
-        }
-      }
-      pod.copy(networks = networks)
-    } else {
-      pod
-    }
-  }
+  private def normalize(pod: PodDefinition): PodDefinition = identity(pod)
```
aquamatthias: why do we need normalize, if it calls identity?

src/main/scala/mesosphere/marathon/core/pod/Network.scala
```diff
@@ -0,0 +1,37 @@
+package mesosphere.marathon.core.pod
+
+import mesosphere.marathon.raml.{ KVLabels, NetworkMode, Network => RAMLNetwork }
+
+/**
+  * Network declared by a [[PodDefinition]].
+  */
+sealed trait Network extends Product with Serializable {
+  def toAPIObject: RAMLNetwork
```
aquamatthias: We start adding api level code into the core.
Wdyt putting this logic into a conversion class/package under api?

aquamatthias:
@jdef Please rebase

jdef:
discussed with MV, going to create a conversions package and migrate the logic there. flipping this back into WIP

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jdef: that's right. I guess we should support "no networking"?

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jdef: not sure, @jasongilanfarr added these

src/main/scala/mesosphere/marathon/core/pod/Network.scala
jdef: WIP

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jasongilanfarr: It'll make the conversion to akka http easier. Trying to get us thinking in a pattern ;)

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
```diff
@@ -0,0 +1,13 @@
+package mesosphere.marathon.api.v2.conversion
+
+trait Converter[A,B] {
```
jasongilanfarr:
```
trait Reads[A] {
   def read(a: RamlType): A
}

trait Writes[A] {
   def write(a: A): RamlType
}
```

etc.

src/main/scala/mesosphere/marathon/api/v2/conversion/EnvVarConversion.scala
```diff
@@ -0,0 +1,25 @@
+package mesosphere.marathon.api.v2.conversion
+
+import mesosphere.marathon.raml.EnvVars
+import mesosphere.marathon.{ raml, state }
+
+import scala.collection.immutable.Map
+
+trait EnvVarConversion {
+
+  implicit val fromAPIObjectToEnvironment = Converter { envVars: EnvVars =>
```
jasongilanfarr: naming: ramlToEnvVar. add type annotation.

src/main/scala/mesosphere/marathon/api/v2/conversion/EnvVarConversion.scala
```diff
@@ -0,0 +1,25 @@
...
+    def toEnvValue(envVarValue: raml.EnvVarValueOrSecret): state.EnvVarValue = envVarValue match {
+      case raml.EnvVarValue(value) => state.EnvVarString(value)
+      case raml.EnvVarSecretRef(secret) => state.EnvVarSecretRef(secret)
+    }
+    if(envVars.values.isEmpty) None else Some(envVars.values.map{ case (k, v) => k -> toEnvValue(v) })
+  }
+
+  implicit val fromEnvironmentToAPIObject = Converter { envMap: Map[String, state.EnvVarValue] =>
```
jasongilanfarr: envVarToRaml, add type annotation

src/main/scala/mesosphere/marathon/api/v2/conversion/HealthCheckConversion.scala
```diff
@@ -0,0 +1,9 @@
+package mesosphere.marathon.api.v2.conversion
+
+import mesosphere.marathon.core.health.HealthCheck
+import mesosphere.marathon.raml.{ HealthCheck => RAMLHealthCheck }
+
+trait HealthCheckConversion {
```
jasongilanfarr: see above...

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
```diff
@@ -0,0 +1,13 @@
...
+  def convert(a: A): Option[B]
```
jasongilanfarr: Should just be "B"

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
```diff
@@ -0,0 +1,13 @@
...
+}
+
+object Converter {
+  def apply[A,B](a: A)(implicit c: Converter[A,B]): Option[B] = c.convert(a)
```
jasongilanfarr: No option.

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
```diff
@@ -0,0 +1,13 @@
...
+
+  def apply[A,B](f: (A) => Option[B]): Converter[A,B] = new Converter[A,B] {
+    override def convert(a: A): Option[B] = f(a)
```
jasongilanfarr: No option

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
jdef: if only we had a RamlType :)

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
jdef: done

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
jdef: done

src/main/scala/mesosphere/marathon/api/v2/conversion/Converter.scala
jdef: done

src/main/scala/mesosphere/marathon/api/v2/conversion/ContainerConversion.scala
```diff
@@ -0,0 +1,42 @@
+package mesosphere.marathon.api.v2.conversion
```
jasongilanfarr: I'm kind of thinking this should be in mesosphere/marathon/raml (or mesosphere/marathon/core/raml)

jdef:
of the two options mentioned, core/raml preferred. the package that
contains generated code (mesosphere/marathon/raml) shouldn't have much else
inside of it.

jasongilanfarr:
I'm ok with either and actually having the conversion in the same package seems correct to me.

src/main/scala/mesosphere/marathon/core/pod/PodDefinition.scala
```diff
@@ -88,142 +88,27 @@ case class PodDefinition(
...
-      constraints = constraints,
-      version = podDef.version.fold(Timestamp.now())(Timestamp(_)),
-      podVolumes = podDef.volumes,
-      networks = networks,
-      backoffStrategy = backoffStrategy,
-      upgradeStrategy = upgradeStrategy
-    )
-  }
-  //scalastyle:on
+  def apply(podDef: Pod): PodDefinition = Converter[Pod,PodDefinition](podDef)
```
jdef: shouldn't need explicit `Converter[...]` typing here

src/main/scala/mesosphere/marathon/raml/NetworkConversion.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import mesosphere.marathon.core.pod
+
+trait NetworkConversion {
+  implicit val networkRamlReader: Reads[Network, pod.Network] =
+    Reads { raml =>
+      raml.mode match {
+        case NetworkMode.Host => pod.HostNetwork
+        case NetworkMode.Container => pod.ContainerNetwork(
+          // TODO(PODS): shouldn't this be caught by validation?
```
jdef: yes, but how else would you handle this (unexpected) case?

jdef: actually, handled by normalization

scalastyle-config.xml
```diff
@@ -1,127 +0,0 @@
-<scalastyle commentFilter="enabled">
```
jdef: why is this file removed as part of this PR?

src/main/scala/mesosphere/marathon/raml/NetworkConversion.scala
jasongilanfarr: True on both counts

scalastyle-config.xml
jasongilanfarr: I was annoyed with the formatted violating the long lines by 1 character and it doesn't exist on the branch.

jdef:
since the raml types are all specific to v2 a better home for the conversion stuff is probably under api/v2 instead of core

jasongilanfarr:
Don't agree at all. We currently only have v2 defined in raml. Please leave
it where it is.

jdef:
I left the location alone. Discussion tabled.

jasongilanfarr:
Long term, I think the raml converter itself should stay in the package and
the generator should support package annotations (and imports) so we can
organize accordingly.

jdef:
That's fine. There's probably a limit to what we'll actually want to
automated in terms of converter generation: rewards we reap vs. time
invested (and RAML readability). I suspect we'll land on a solution that
generates 90 - 95% of the converter code, keeps the RAML readable, and
still requires some manual bits to fill in the gaps.

PR #4392, opened 2016-09-16
kill Marathon if Curator cannot create a ZK factory

jdef:
Attempts to resolve a fun cluster bootstrapping problem. If ZK DNS names don't resolve then `new Zookeeper` generates some `IllegalArgumentException` (because no ZK hosts were "specified") and Curator is seemingly hosed from then on. Which means that the Marathon instance never offers leadership.

This impl fails fast in the presence of such errors.

TODO
- [ ] integration (shakedown?) test


src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
```diff
@@ -125,6 +126,25 @@ class CuratorElectionService(
...
+          try {
+            delegate.newZooKeeper(connectString, sessionTimeout, watcher, canBeReadOnly)
+          } catch {
+            // We may see exceptions here, for example, when none of the ZK hosts are resolvable at Marathon startup;
+            // in this case Curator may fail with some background exception and Marathon never offers leadership.
+            case NonFatal(e) =>
+              // TODO(jdef) should we really be committing suicide here? what about retries instead?
+              log.error("ZooKeeper initialization failed - Committing suicide")
+              CurrentRuntime.asyncExit()(scala.concurrent.ExecutionContext.global)
+              null
```
jdef: @jasongilanfarr is there a better way to handle this?

jasongilanfarr: I'd like to churn on this a bit more. My inclination is to do the resolution ourselves in a Retry.blocking call, but this gets weird because of how java caches DNS. It seems wrong to me to change their factory to get around the bug instead of giving it a list of IPs

jdef: @cmaloney recommendations?

jdef: My initial comment was re: the `null`.

However, since you've brought up DNS... I'm told that on dc/os we use special hostnames that should resolve to virtual IPs. Precisely because we can't trust Java DNS caching. I suspect we're still seeing this in dc/os at bootstrapping time because the service providing that function isn't available yet.

However, we should also degrade gracefully in non-dc/os envs. Anyone running Marathon in production should be running it with a service-runner (systemd, runit, whatever) so that if Marathon dies it will respawned shortly.

The solution as implemented is, IMO, a decent compromise because it (a) makes no assumptions about DNS; (b) makes no assumption about dc/os vs mesos, and; (c) relies on (safe) assumptions re: production-ready deployment.

Alternatives? We could build in our own retry mechanism. Or provide our own host-lookup provider. I'm not convinced that such band-aids are any better. If the community clamors for additional flexibility we could create a plugin interface for providing custom ZooKeeperFactory implementations and just let people bring their own if they hate our default behavior.

cmaloney: We hard assert the zk urls are resolvable / spartan is up before we start Marathon on DC/OS:
https://github.com/dcos/dcos/blob/master/packages/marathon/build#L26

So marathon starting but the magic url not being resolvable shouldn't be possible.

jdef:
that's a ping for leader.mesos, which I *thought* was provided by
mesos-dns. the ZK url has a bunch of .zk names, which I think are provided
by spartan (
https://github.com/mesosphere/spartan/blob/03723675f3b478fc8e0204f89cacb2c564024bec/README.md#zookeeper),
not mesos-dns. so .. where's the assertion that spartan is alive?

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
jasongilanfarr: We have our own retry mechanism already. This is why i suggested using the Java DNS resolution ourselves instead of jumping through with custom curator. If we can't resolve after 5 attempts, die.

cmaloney:
Mesos-DNS runs and listens on port 60053. Locally `/etc/resolv.conf` never points directly at Mesos-DNS.  `/etc/resolv.conf` gets set to point Spartan, Spartan forwards to Mesos-DNS

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
jasongilanfarr: We also could set the system property in main to disable java's DNS caching and instead rely on the underlying system's behavior.

jdef:
OK, well, perhaps Spartan was running. But it wasn't resolving .zk names.
The logs show it. Perhaps we should ping a .zk name as a Prestep?

jasongilanfarr:
ehh. If the real bug is that curator screws up if none of the URLs resolve, then let's test the DNS ourselves and give curator that as the list of IPs (instead of hosts) - we should wrap it in retry and disable java dns caching. Curator _will_ keep trying ZK for a while if its not there, or at least i take advantage of this all the time (because i forget to start ZK)

cmaloney:
@sargun Not being able to resolve zk names shouldn't be possible, correct?

jdef:
FWIW
```
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: [2016-09-15 22:14:26,604] WARN No IP address found for server: zk-1.zk:2181 (org.apache.zookeeper.client.StaticHostProvider:ForkJoinPool-2-worker-13)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: java.net.UnknownHostException: zk-1.zk: unknown error
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at java.net.Inet6AddressImpl.lookupAllHostAddr(Native Method)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at java.net.InetAddress$2.lookupAllHostAddr(InetAddress.java:928)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at java.net.InetAddress.getAddressesFromNameService(InetAddress.java:1323)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at java.net.InetAddress.getAllByName0(InetAddress.java:1276)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at java.net.InetAddress.getAllByName(InetAddress.java:1192)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at java.net.InetAddress.getAllByName(InetAddress.java:1126)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at org.apache.zookeeper.client.StaticHostProvider.resolveAndShuffle(StaticHostProvider.java:116)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at org.apache.zookeeper.client.StaticHostProvider.<init>(StaticHostProvider.java:77)
Sep 15 22:14:26 ip-10-10-0-37 java[17395]: at org.apache.zookeeper.ZooKeeper.<init>(ZooKeeper.java:718)
```

jdef:
I believe this happened during cluster upgrade, from dc/os 1.7 to 1.8

jasongilanfarr:
As discussed in Slack, we could change "provideCuratorClient" by wrapping its body in a Retry and fail if it ultimately results in UnknownHostException. Should be a much simpler change.

jdef:
for posterity .. the final stack trace in the "series" of ZK hostname
resolution attempts appears to show an IllegalArgumentException bubbling up
from ZooKeeper<init> all the way up through Curator to the Marathon code

jasongilanfarr:
I tried my suggestion about retrying on illegal argument exception and it worked fine. I can replace the PR or something. Funny enough, ZKPersistenceStore hit it first (but it'll retry a lot then ultimately take things down). This was just on master.

jeschkies:
Just so I get this. You don't want to resolve the IP but just retry in `provideCuratorClient`?

jeschkies:
Okay, I cannot get the code to get the UnknownHostException. It is caught somewhere else and the Curator throws an IllegalArgumentException. `provideCuratorClient` does not throw anything :/.

And may I add a combination of `lazy val`s with futures, async code, `synchronized` keywords together with latches just screams that something is at odd here ^^.

jdef:
FWIW I think there's real value here in keeping things simple :)

On Thu, Sep 29, 2016 at 9:47 AM, Karsten Jeschkies <notifications@github.com

jeschkies:
My point. I don't get this code.

Anyhow. `builder.build()` and `client.start()` don't throw any exceptions. So the the `UnknownHostException` is caught somewhere in-between. And I don't see a way to react to this. So, if Curator really retries, why not catch the `IllegalArgumentException` and then exit?

jdef:
That's pretty much what this PR does. It could be more specific in catching
**exactly** the IllegalArgumentException that is thrown, but why stop
there? It's clear that curator is dropping the call on this and not
reporting this back up the chain. This is really a workaround for something
that is better handled within curator.

jeschkies:
Hm, so I may have misunderstood @jasongilanfarr's suggestion then.

jeschkies:
Ah, I missed the IP resolution part.

jdef:
I think Jason and I differ on approach to this. :)

jeschkies:
Last comment was to catch`IllegalArgumentException` and retry on that. I'll just give an example and then you guys decide.

jeschkies:
Please see my diff in Phabricator https://phabricator.mesosphere.com/D34

aquamatthias:
@jdef @jeschkies What is the status of this PR. Looks like D34 supersedes this one?

jdef:
yes, superseded by D34. closing.

PR #4406, opened 2016-09-16
added `name` field to `portMappings`

sascala:
Added this because the docs for DC/OS Tunnel mention adding a name to the `portMappings` field (https://github.com/dcos/dcos-docs/pull/596)

@ssk2 I'm assigning this to you because you wrote this page

/cc @nlsun


docs/docs/ports.md
```diff
@@ -223,9 +223,9 @@ Alternatively, if our process running in the container had fixed ports, we might
       "image": "my-image:1.0",
       "network": "BRIDGE",
       "portMappings": [
-        { "containerPort": 80, "hostPort": 0 },
-        { "containerPort": 443, "hostPort": 0 },
-        { "containerPort": 4000, "hostPort": 0 }
+        { "containerPort": 80, "hostPort": 0, "name": "http" },
+        { "containerPort": 443, "hostPort": 0, "name": "https" },
+        { "containerPort": 4000, "hostPort": 0, "name": "tcp" }
```
ssk2: `tcp` isn't a descriptive name here. (http, https, tcp are examples of protocols and so this might confuse the concept of a `name` and `protocol` property.) I'd leave the first two as `http` and `https` but see if we can pick a better name for this third one. Perhaps a fictitious example like `monitoring` or `stats`?

aquamatthias:
@sascala Please update this branch with the suggestion from @ssk2 and reopen a PR.

PR #4410, opened 2016-09-16
Fixes #4480 Race Condition in GroupManager with new Persistence Store

jasongilanfarr:
There were situations where groups would be referenced by deployment
plans, but the group was not yet stored. Ensure that roots are always
available for any plan by storing the root version first.

### DIFF
--- a/src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
@@ -247,7 +247,6 @@ class StoredGroupRepositoryImpl[K, C, S](
           await(preStore(storedGroup))
         case _ =>
       }
-      val promise = Promise[Group]()
 
       val storeAppFutures = updatedApps.map(appRepository.store)
       val storedApps = await(Future.sequence(storeAppFutures).asTry)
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
```diff
@@ -238,6 +238,37 @@ class StoredGroupRepositoryImpl[K, C, S](
...
+  @SuppressWarnings(Array("all")) // async/await
+  override def storeRootVersion(group: Group, updatedApps: Seq[AppDefinition]): Future[Done] =
+    async { // linter:ignore UnnecessaryElseBranch
+      val storedGroup = StoredGroup(group)
+      beforeStore match {
+        case Some(preStore) =>
+          await(preStore(storedGroup))
+        case _ =>
+      }
+      val promise = Promise[Group]()
```
jdef: what is this promise actually being used for?

jasongilanfarr: oops, that was for revert root which is not something we need for this one... will double check.

jasongilanfarr:
Retest this please

jasongilanfarr:
Retest this please

jasongilanfarr:
Retest this please

PR #4411, opened 2016-09-16
Benchmark project (jmh)

jasongilanfarr:
We should probably create a new velocity job that only runs the JMH benchmarks and reports them with the performance plugin.


aquamatthias:
LGTM

Is there a way to see changes over time - or how do we detect, if a change makes the execution slow?

@jmlvanre This is for micro benchmarks only. Not sure, if this is what you are looking for.

build.sbt
```diff
@@ -201,7 +201,19 @@ lazy val `mesos-simulation` = (project in file("mesos-simulation"))
     .enablePlugins(GitBranchPrompt, CopyPasteDetector)
     .settings(commonSettings: _*)
     .settings(formatSettings: _*)
-    .dependsOn(marathon % "compile->compile; test->test").configs(IntegrationTest)
+    .dependsOn(marathon % "compile->compile; test->test")
     .settings(
       name := "mesos-simulation"
     )
+
+lazy val benchmark = (project in file("benchmark"))
```
jmlvanre: Can we add a comment referencing `benchmark/README.md` for more info?

benchmark/README.md
```diff
@@ -0,0 +1,6 @@
+Benchmarks here are generally in src/main and can be run with `benchmark/jmh:run`
+within SBT.
+
+See here for more details:
```
jmlvanre: Are there any suggested options that you can pull out from these referenced readmes?

Maybe 2-3 patterns that are most common for us in marathon?

.idea/compiler.xml
```diff
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- TODO: Remove this from git when we have linting working in test -->
```
jmlvanre: Why is this line removal part of this PR?

build.sbt
jasongilanfarr: sure, I don't think its that useful in the build files, but we should reference how we benchmark somewhere else.

benchmark/README.md
jasongilanfarr: This isn't a PR to actually merge as is imo. Its really guidance for "when someone goes to make a benchmark", please use this as a base.

.idea/compiler.xml
jasongilanfarr: IntelliJ did that :(

jasongilanfarr:
Retest this please

PR #4417, opened 2016-09-19
Also take pods into considerations for Scheduling operations.

aquamatthias:
Rename everything from app* to RunSpec


meichstedt:
LGTM

PR #4418, opened 2016-09-19
Support for creating persistent volumes on path or mount disk resources (#4107)

jasongilanfarr:
Tries and fulfill largest disk resource requests, first. This could help Marathon to do a more intelligent thing and not allocate
itself into a corner. Given the following two resources:

- path resource with 128m free
- path resource with 628m free

If we had a task attempting to allocate two pesistent disks, one for
100m, and the other for 600m, it would fail if the 100mb were subtracted
from the 628m resource, first (which is the default
precedence for path disk resources, to evenly distribute free space).

In my implementation, as of now, you must specify on which type of disk resource your persistent volume should be created, root, mount, or path. Here is an example:

{
  "type": "DOCKER",
  "volumes": [
    {
      "containerPath": "data",
      "mode": "RW",
      "persistent": {
        "size": 960679,
        "type": "mount",
        "constraints": []
      }
    }
    // ...

Failing to provide the persistent type will default it to root. If you specify a persistent type of path, the root volume resource will not be considered.

    Regarding allocation priorities, I'd suggest to make this s simple as possible for now. Specifying lower and upper limits to not need to define an exact match probably makes it easier, but I think we rather need an explicit and usable volume management sooner than later

I think this is wise. Since I've implemented path constraints already, it is possible (in a round-about way) to tell Marathon to avoid using certain disks. Adding an optional maxSize parameter for mount persistent types would be trivial to add and probably makes sense.


jasongilanfarr:
@timcharper  - I merged on master and fixed the linting errors. Once it passes CI, happy to merge into master.

jasongilanfarr:
Retest this please

PR #4422, opened 2016-09-20
WIP: Security flaw in build script.

jeschkies:
If I'm not mistaken anybody can trigger a build and just echo our docker hub password.


jeschkies:
As I thought it works.

jasongilanfarr:
:( we need to get into build-pipeline to actually fix this.

kensipe:
this is only true if the push to docker is built on a public jenkins.  we publish from a private build.

jeschkies:
I could access the logs when I was not logged in.

PR #4425, opened 2016-09-20
Fix TaskKillerTest

meichstedt:
`kill with wipe will kill running and expunge all` turned out shaky e.g. [here](https://jenkins.mesosphere.com/service/jenkins/job/public-marathon-unit-integration-tests-pulls/1215/). The underlying problem is that, if `TaskKiller` is called with `wipe=true`, it will explicitly expunge all found instances, **and** call `KillService.killInstances` with all launched instances:

````
if (wipe) expunge(foundTasks)
val launchedTasks = foundTasks.filter(_.isLaunched)
if (launchedTasks.nonEmpty) await(service.killTasks(appId, launchedTasks))
```

With the given test setup, this will effectively process two Expunge operations for the running instance, and one for the reserved one. The test only verified 2 ops, and eventually both were for the same instance which let the test fail.


aquamatthias:
Thanks!

aquamatthias:
:ship: it

PR #4426, opened 2016-09-20
Fix integration tests

aquamatthias:
3 commits that fixes integration tests:
- repair task kill
- repair down scale
- mute resident task integration tests


meichstedt:
LGTM

meichstedt:
Why does that f*** _close and comment_ button even exist?! I've never clicked on that deliberately …

PR #4427, opened 2016-09-20
Generate real PodStatus

jdef:
Lots of redundancy between AppSelector and GroupSelector. Instead of copy/paste/rename as PodSelector I opted to refactor the "selection" API.

Apparently there's an optimization in the form of AppInfoBaseData where lots of data (deployments, status, etc) is cached for use by queries that walk the group tree. Plan is to integrate pod status resolution into said cache and use that from DefaultInfoService.


src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -1,35 +1,42 @@
 package mesosphere.marathon.core.appinfo.impl
 
-import mesosphere.marathon.MarathonSchedulerService
+import java.time.OffsetDateTime
```
jeschkies: The logs says

```sbt
[error] /jenkins/workspace/public-marathon-snapshot-packages-pulls/marathon-pkg/marathon/src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala:3: Unused import
[error] import java.time.OffsetDateTime
[error]                  ^
[error] /jenkins/workspace/public-marathon-snapshot-packages-pulls/marathon-pkg/marathon/src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala:5: Unused import
[error] import mesosphere.marathon.{ DeploymentService, MarathonSchedulerService }
[error]                                                 ^
[error] two errors found
```

jeschkies:
retest this please

jeschkies:
retest this please

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
jdef: yes, I'm trying to make intellij generate errors for this so I don't commit them

jeschkies:
retest this please

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -0,0 +1,116 @@
...
+import scala.collection.immutable.Seq
+
+trait PodStatusConversion {
+
+  import PodStatusConversion._
+
+  /**
+    * generate a pod instance status RAML for some instance.
+    */
+  @throws[IllegalArgumentException]("if you provide a non-pod `spec`")
```
jasongilanfarr: Annotation is unnecessary as its a RuntimeException

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -0,0 +1,116 @@
...
+      case InstanceStatus.Staging | InstanceStatus.Starting => PodInstanceState.Staging
+      case InstanceStatus.Error | InstanceStatus.Failed | InstanceStatus.Finished | InstanceStatus.Killed |
+        InstanceStatus.Gone | InstanceStatus.Dropped | InstanceStatus.Unknown | InstanceStatus.Killing |
+        InstanceStatus.Unreachable => PodInstanceState.Terminal
+      case InstanceStatus.Running =>
+        if (instance.state.healthy.getOrElse(true)) PodInstanceState.Stable else PodInstanceState.Degraded
+    }
+
+    val networkStatus: Seq[NetworkStatus] = instance.tasks.flatMap { task =>
+      task.mesosStatus.filter(_.hasContainerStatus).fold(Seq.empty[NetworkStatus]) { mesosStatus =>
```
jasongilanfarr: withFilter()

jasongilanfarr: You may also want to do a view since you are chaining

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -0,0 +1,116 @@
...
+        val networkStatus: NetworkStatus = toMerge.reduceLeft { (merged, single) =>
+          merged.copy(addresses = merged.addresses ++ single.addresses)
+        }
+        networkStatus.copy(addresses = networkStatus.addresses.distinct)
+      }
+    }(collection.breakOut)
+
+    val resources: Option[Resources] = instance.state.status match {
+      case InstanceStatus.Staging | InstanceStatus.Starting | InstanceStatus.Running =>
+        val containerResources = pod.containers.map(_.resources).fold(Resources(0, 0, 0, 0)) { (acc, res) =>
```
jasongilanfarr: why isn't the resources exposed at the pod level?

src/main/scala/mesosphere/marathon/core/appinfo/PodStatusService.scala
```diff
@@ -0,0 +1,14 @@
+package mesosphere.marathon.core.appinfo
+
+import mesosphere.marathon.raml.PodStatus
+import mesosphere.marathon.state.PathId
+
+import scala.concurrent.Future
+
+trait PodStatusService {
```
jasongilanfarr: why isn't this in the PodModule?

jdef:
It basically provides a service similar to AppInfoService and
GroupInfoService. PodInfoService seemed like a bad name. PodStatusService
is implemented in the "appinfo" package (should perhaps be renamed) because
it depends on AppInfoBaseData (should also be renamed).The idea being that
AppInfoBaseData has a cache of information that we want to reuse when
calculating pod status (e.g. when determining pod terminal status, task
termination history (which still isn't implemented but will be soon)). And
DefaultInfoService is the gateway-to AppInfoBaseData. And it's important to
have access to this cache when walking the group tree, for performance
reasons.

jdef:
It probably is. But at some point we (Mesos/Marathon) will probably allow
tasks to die off without killing the whole pod. And this status information
should reflect the live state, not the spec'd state. So it should actually
be filtering on the containers that are still alive. I don't remember
implementing that because I was waiting on the ability to associate a Task
with a Container spec. I think that capability exists now since this PR was
rebased earlier this evening.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: I had a doc comment instead. scaladoc puked. So I got rid of the scaladoc and used this instead. I think it's useful to annotate somehow. Happy to add back as a doc item if scaladoc stops whining and failing the test phase.

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -161,8 +166,51 @@ class AppInfoBaseData(
       case NonFatal(e) => throw new RuntimeException(s"while retrieving last task failure for app [${app.id}]", e)
     }
   }
+
+  def calculatePodStatus(podDef: PodDefinition): Future[PodStatus] = {
+    val now = clock.now().toOffsetDateTime
+    val instances = instanceTracker.specInstancesSync(podDef.id)
```
jdef: Sync! good catch

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -45,73 +55,118 @@ private[appinfo] class DefaultInfoService(
...
+        } else None
       }
       queryGroup(group)
     }
   }
 
-  private[this] def resolveAppInfos(apps: Iterable[AppDefinition], embed: Set[AppInfo.Embed]): Future[Seq[AppInfo]] = {
-    val baseData = newBaseData()
+  // TODO(jdef) not a big fan of this
+  type SpecStatus = Either[AppInfo, PodStatus]
```
jdef: (A) easier to read (for me, anyway) when used later on. (B) ok

src/main/scala/mesosphere/marathon/core/appinfo/package.scala
```diff
@@ -0,0 +1,11 @@
+package mesosphere.marathon.core
+
+import mesosphere.marathon.core.pod.PodDefinition
+import mesosphere.marathon.state.{ AppDefinition, Group }
+
+package object appinfo {
+
+  type AppSelector = Selector[AppDefinition]
```
jdef: happy to remove. did this initially to help with conversion to Selector[AppDefinition]

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -36,10 +36,12 @@ class AppsResource @Inject() (
     service: MarathonSchedulerService,
     appInfoService: AppInfoService,
     val config: MarathonConf,
-    val authenticator: Authenticator,
-    val authorizer: Authorizer,
     groupManager: GroupManager,
-    pluginManager: PluginManager) extends RestResource with AuthResource {
+    pluginManager: PluginManager)(implicit
+  val authenticator: Authenticator,
```
jdef: needed to compile?

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jasongilanfarr: withFilter is lazy, filter computes the collection immediately. If you're doing a chain of transforms, in theory, views are better. I have a change incoming where you can actually flip over to the java8 stream api transparently - and in there you explicitly have "terminal operations" which "complete the stream". scala views are supposed to work similarly to this.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -270,22 +272,22 @@ class AppsResource @Inject() (
...
   private[v2] def search(cmd: Option[String], id: Option[String], label: Option[String]): AppSelector = {
     def containCaseInsensitive(a: String, b: String): Boolean = b.toLowerCase contains a.toLowerCase
-    val selectors = Seq[Option[AppSelector]](
-      cmd.map(c => AppSelector(_.cmd.exists(containCaseInsensitive(c, _)))),
-      id.map(s => AppSelector(app => containCaseInsensitive(s, app.id.toString))),
+    val selectors = Seq[Option[Selector[AppDefinition]]](
+      cmd.map(c => Selector(_.cmd.exists(containCaseInsensitive(c, _)))),
+      id.map(s => Selector(app => containCaseInsensitive(s, app.id.toString))),
       label.map(new LabelSelectorParsers().parsed)
     ).flatten
```
jdef: huh? which outer one?

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -25,13 +26,14 @@ import play.api.libs.json.Json
...
 class PodsResource @Inject() (
-    val config: MarathonConf,
-    val authenticator: Authenticator,
-    val authorizer: Authorizer)(
+    val config: MarathonConf)(
     implicit
-    podSystem: PodManager,
-    eventBus: EventStream,
-    mat: Materializer) extends RestResource with AuthResource {
+    val authenticator: Authenticator,
```
jdef: compiler disagrees

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
jasongilanfarr: i see ".map{}... .flatten" -> in general you can replace that with flatMap

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jasongilanfarr: sorry thought it was in the implicits list

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -161,8 +166,51 @@ class AppInfoBaseData(
       case NonFatal(e) => throw new RuntimeException(s"while retrieving last task failure for app [${app.id}]", e)
     }
   }
+
+  def calculatePodStatus(podDef: PodDefinition): Future[PodStatus] = {
+    val now = clock.now().toOffsetDateTime
```
jdef: the type of `clock` is a Marathon `Clock` type, not a stdlib type.

jasongilanfarr: :(

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
jdef: actually it's `Seq[Option[x]] ( maybeX1, maybeX2, maybeX3 ).flatten`

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -161,8 +166,53 @@ class AppInfoBaseData(
       case NonFatal(e) => throw new RuntimeException(s"while retrieving last task failure for app [${app.id}]", e)
     }
   }
+
+  @SuppressWarnings(Array("all")) // async/await
+  def podStatus(podDef: PodDefinition)(implicit ec: ExecutionContext): Future[PodStatus] =
+    async { // linter:ignore UnnecessaryElseBranch
+      val now = clock.now().toOffsetDateTime
+      val instances = await(instanceTracker.specInstances(podDef.id))
```
aquamatthias: Rename `tasksByAppFuture` to `instancesByRunSpecFuture` and use this method here.

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -161,8 +166,53 @@ class AppInfoBaseData(
...
+    runningDeployments.map { infos =>
+      infos.exists(_.plan.deletedPods.contains(id))
+    }
 }
 
 object AppInfoBaseData {
   private val log = LoggerFactory.getLogger(getClass)
+
+  @SuppressWarnings(Array("all")) // async/await
+  protected def podState(
```
aquamatthias: Why is this method defined in the companion?

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -58,6 +58,8 @@ trait DeploymentService {
     * @return a failed future if the deployment failed.
     */
   def deploy(plan: DeploymentPlan, force: Boolean = false): Future[Unit]
+
+  def listRunningDeployments(): Future[Seq[DeploymentStepInfo]]
```
aquamatthias: We currently use `Future[Seq[T]]` in a lot of places. Please let's defer refactoring in a separate PR.

src/main/scala/mesosphere/marathon/api/v2/GroupsResource.scala
```diff
@@ -75,17 +77,17 @@ class GroupsResource @Inject() (
 
     //format:off
     def appsResponse(id: PathId) =
-      infoService.selectAppsInGroup(id, allAuthorized, appEmbed).map(info => ok(info))
+      infoService.selectAppsInGroup(id, AppsResource.authzSelector, appEmbed).map(info => ok(info))
```
aquamatthias: should be `authorizationSelectors.appSelector`

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -29,91 +43,125 @@ private[appinfo] class DefaultInfoService(
...
-      else
-        Future.successful(Seq.empty)
-    }
+    async { // linter:ignore UnnecessaryElseBranch
+      val groupEmbedApps = groupEmbed(GroupInfo.Embed.Apps)
+      val groupEmbedPods = groupEmbed(GroupInfo.Embed.Pods)
+
+      //fetch all transitive app infos and pod statuses with one request
+      val specStatuses: Seq[SpecStatus] = await {
+        resolveSpecInfos(group.transitiveRunSpecs.collect {
```
aquamatthias: This will resolveSpecInfos for all apps and pods no matter which embed option you have.
The original goal was, to only resolve this data, if the embed option is set.

Since you split the data after resolving anyways, I think separating the 2 cases would be easier.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
aquamatthias: Once a task dies inside a pod, the resources are freed.
We should have this computation logic on the PodDefinition like:
```
  def resources(filter: Container=>Boolean): Resources
```
We can use this method to filter live resources.

jdef:
Will submit a follow-up PR w/ unit tests for pod status generation

aquamatthias:
@jdef Please also update documentation for the new embed option

PR #4435, opened 2016-09-20
Wait in tests with `isReadyWithin` instead of `futureValue`.

jeschkies:
This should make the tests a little more robust against time outs.


src/test/scala/mesosphere/marathon/upgrade/TaskStartActorTest.scala
```diff
@@ -92,7 +93,7 @@ class TaskStartActorTest
     when(f.launchQueue.get(app.id)).thenReturn(None)
     val task =
       MarathonTestHelper.startingTaskForApp(app.id, appVersion = Timestamp(1024))
-    f.taskCreationHandler.created(TaskStateOp.LaunchEphemeral(task)).futureValue
+    f.taskCreationHandler.created(TaskStateOp.LaunchEphemeral(task)).isReadyWithin(Span(1, Second))
```
jasongilanfarr: You can actually do ```.futureValue(Timeout(1.second))```

jdef:
for a while k8s used a global variable for test timeouts like this. it was
convenient on slower build systems/clusters to tweak the value of the
variable to prevent test flakes due to timeouts

jasongilanfarr:
Let's try to avoid referring to other projects directly for things like this (especially a competing one). Generally speaking, I've found that having a long timeout for futures _when_ it makes sense is valuable.

src/test/scala/mesosphere/marathon/upgrade/TaskStartActorTest.scala
jeschkies: `futureValue` was confusing me when I read it. I guess it depends if this is more part of the setup as in `given f.taskCreationHandler.created(TaskStateOp.LaunchEphemeral(task)) is ready` or if it is more of a assertion as in `should be ready within(5 seconds)`.

I'm fine with both :)

jeschkies:
I'll go with implicit timeouts in a later change. See #4443

src/test/scala/mesosphere/marathon/upgrade/TaskStartActorTest.scala
jeschkies: I'll address this is #4443

meichstedt:
FWIW `futureValue` takes an `implicit config: PatienceConfig` – we've declared longer implicit timeouts in other tests in this manner; that way, the timeout will be applied to _all_ futureValue's instead. Depending on the test and the futures that are handled, it makes sense to either declare an explicitly longer timeout for one specific call, or increase it for the whole test as @jdef suggested.

Even if we don't use the value, `futureValue` reads more clearly to me than `isReadyWithin`. To me, the latter implies that a part of the test is checking whether the future completes. I would argue that in most if not all of the cases, `futureValue` is not used as _should complete within x seconds_, but rather _block until this future completes_.

jeschkies:
@meichstedt fully agree. I mentioned this in the comment to Jason. I've just found the `FutureTestSupport` trait and replaced my fix with that.

src/test/scala/mesosphere/marathon/upgrade/TaskStartActorTest.scala
jeschkies: Okay, the newest change should address your issue.

jeschkies:
restest this please

jeschkies:
restest this please

jasongilanfarr:
retest this please

PR #4443, opened 2016-09-21
Start test traits consolidation

jeschkies:
We have several test traits and helpers that are quire similar. This is the first step to consolidate them.


src/test/scala/mesosphere/marathon/test/MarathonSpec.scala
```diff
@@ -1,6 +1,7 @@
-package mesosphere.marathon
+package mesosphere.marathon.test
 
-import org.scalatest.{ BeforeAndAfter, FunSuiteLike, OptionValues }
+import mesosphere.FutureTestSupport
 import org.scalatest.mockito.MockitoSugar
+import org.scalatest.{ BeforeAndAfter, FunSuiteLike, OptionValues }
 
-trait MarathonSpec extends FunSuiteLike with BeforeAndAfter with MockitoSugar with OptionValues
+trait MarathonSpec extends FunSuiteLike with BeforeAndAfter with MockitoSugar with OptionValues with FutureTestSupport
```
jasongilanfarr: should probably add everything that UnitTestLike has.. maybe move this over there and add a comment that one should prefer UnitTest?

jeschkies: The move is the next step.

PR #4444, opened 2016-09-21
[Don't merge] Only for discussion – residency with instances

meichstedt:
## Problem
- Tasks using persistent volumes return to state `Reserved` when they terminate
- Instances become `Reserved` as well
- InstanceChanged events that get published report `Reserved`, MesosStatusUpdateEvents report `Reserved` as well – that breaks existing logic that depends on events reporting a terminal status for these

## Ugly solution
- The InstanceChange carries an optional trigger of type `TaskStatus`, which is the trigger that led to the  state change. Events use that status, or, if None, the new status of the instance. This solution is particularly ugly as it pollutes a significant amount of code and makes it hard to read and reason about.

I will setup another PR with a hopefully much more cleaner solution; this one is just for comparison _so please do not merge_ 😬


meichstedt:
Superseded by #4457

PR #4445, opened 2016-09-21
Remove workaround for Mesos validation issues.

nfnt:
The validation issues reported in MESOS-6208 have been reverted. The
workarounds are no longer necessary.


aquamatthias:
Thanks

PR #4446, opened 2016-09-21
Enable partition awareness.

jeschkies:


src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
```diff
@@ -73,6 +73,9 @@ object MarathonSchedulerDriver {
       log.debug("GPU_RESOURCES feature enabled.")
     }
 
+    // Being partition aware allows Marathon to more granular task status, see MESOS-5344.
+    frameworkInfoBuilder.addCapabilities(Capability.newBuilder().setType(Capability.Type.PARTITION_AWARE))
```
jeschkies: What happens if this is called with Mesos 1.0?

nfnt: Nothing, the enum values are backwards-compatible.

task_status_fsm.dot
```diff
@@ -0,0 +1,49 @@
...
+    RUNNING -> FINISHED;
+    RUNNING -> UNKNOWN;
+    RUNNING -> UNREACHABLE;
+    RUNNING -> KILLING;
+
+    KILLING -> KILLED;
+
+    GONE_BY_OPERATOR -> GONE;
+    GONE_BY_OPERATOR -> RUNNING;
+}
```
jeschkies: I'd like to take the chance to discuss the different stats and their transitions. So far I inferred everything from the `mesos.proto`. I'd like to add what Marathon should be doing once it's partition aware.

jeschkies: Task State Machine

task_status_fsm.dot
```diff
@@ -0,0 +1,49 @@
+digraph task_status_fsm {
```
jasongilanfarr: :+1: :100: :+1: 

but can you put in in docs or something? or even next to the MarathonTaskStatus...

jeschkies:
Task Status FSM

This is what I got mostly from @neilconway.

As far as I can tell, Marathon will have to handle cases when a task comes back from being `UNREACHABLE`.

jeschkies:
Task Status FSM

jeschkies:
Code review moved to https://phabricator.mesosphere.com/D39

jeschkies:
See https://phabricator.mesosphere.com/D50 and follow ups.

PR #4447, opened 2016-09-21
parse container name from task ID, some unit tests to go with it

jdef:
Also changed anonymous container name from `container` to `$anon` to avoid overlap with user-specified container names.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -143,8 +143,8 @@ object Task {
 
     def containerName(taskId: String): Option[String] = {
       taskId match {
-        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceUuid, uuid) =>
-          if (uuid == anonymousContainerName) None else Some(uuid)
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -132,12 +135,21 @@ object Task {
...
-        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceId, uuid) => PathId.fromSafePath(runSpecId)
+        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceId, maybeContainer) => PathId.fromSafePath(runSpecId)
         case LegacyTaskIdRegex(runSpecId, uuid) => PathId.fromSafePath(runSpecId)
         case _ => throw new MatchError(s"taskId $taskId is no valid identifier")
       }
     }
 
+    def containerName(taskId: String): Option[String] = {
+      taskId match {
+        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceUuid, uuid) =>
```
unterstein: /uuid/maybeContainer/ ?

jdef:
ack

PR #4451, opened 2016-09-22
Fixes #4449 by also validating the PathId for pods.

### LINKED ISSUE #4449: [Pods] We need to validate the id of the pod during creation
Create pod with id `/myPod` should be rejected

aquamatthias:


unterstein:
`valid` references PathId#validPathChars which allows only non capital letters?

unterstein:
btw. tests succeed, scapegoat complaints unrelated lines

PR #4452, opened 2016-09-22
Call an optional start-hook script from /bin/start

zen-dog:
An optional start hook script defined via $MARATHON_APP_START_HOOK environment variable is called from /bin/start. This provides a possibility for additional initialization steps when starting marathon.

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -40,6 +40,19 @@ require_arg() {
   fi
 }
 
+# Run the start-hook script [optional]
+if [ -f "$MARATHON_APP_START_HOOK" ]; then
+  "$MARATHON_APP_START_HOOK"
+  # if a $MARATHON_APP_START_HOOK.env file was produced it is evaluated
+  starthook_env=${MARATHON_APP_START_HOOK%.*}.env
+  starthook_env=${starthook_env##*/}
+
+  # In case the hook script needs to add/change environment variables or otherwise access the parent shell
+  [ -f "$starthook_env" ] && source "$starthook_env"
+else
+  echo "No start hook file found (\$MARATHON_APP_START_HOOK). Proceeding with the start script."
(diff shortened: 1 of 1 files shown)


zen-dog:
retest this please

bin/start
```diff
@@ -40,6 +40,19 @@ require_arg() {
   fi
 }
 
+# Run the start-hook script [optional]
+if [ -f "$MARATHON_APP_START_HOOK" ]; then
```
jdef: maybe check with `-x` instead?

jdef:
environment variables that start with `MARATHON_` are also interpreted as marathon flags I think. is it standard practice to use envvars for things that are not also command line flags for Marathon? /cc @aquamatthias

aquamatthias:
`MARATHON_*` is translated to command line flags, but ignores `MARATHON_APP*` so I don't see a conflict with that name. 

`MARATHON_APP_*` is used to set values of the app definition, the task is derived from.

So we probably should name this env var in a way it does not start with `MARATHON_`
E.g.: `HOOK_MARATHON_START`

bin/start
zen-dog: Yeh, makes sense.

zen-dog:
> MARATHON_APP_* is used to set values of the app definition, the task is derived from.
>... E.g.: HOOK_MARATHON_START
I agree, however any name I could come up with that is not starting with MARATHON_ somehow sounds wrong. And since there is no logic that would convert **all** MARATHON_APP* prefixed variables to flags it felt ok to make an exception.  But I can change it if you insist.

aquamatthias:
🚢 it

jdef:
retest this please

aquamatthias:
retest this please

PR #4453, opened 2016-09-22
Fix KillService behavior

meichstedt:
Retry forever and don't spawn progress actors for unknown tasks

Background: there is actually no reason for setting up progress actors/returning a future when killing unknown tasks: no one is interested in the result. Yet the KillServiceActor should retry until it receives a terminal status update for that unknown task.


aquamatthias:
This means: an unknown task is eventually killed (we only send one kill for every reconciliation) - right.

LGTM.
A test is failing. Is this shaky? 
Other than that: 🚢 it

unterstein:
retest this please

meichstedt:
I'm afraid the test is shaky, I've seen it fail several times recently :|

PR #4457, opened 2016-09-22
Fix Residency behavior with Instances

### LINKED ISSUE #4429: Fix resident tasks

meichstedt:
**Note** I'm still fixing some integration tests; will add more commits.

Generate events that shall get published in the InstanceOpProcessor
    
* Events will be generated where we know what exactly should be reported
* If a status update is the trigger for an instance change, the status update state will be reported
* Resident instances that return to Reserved after they terminate will be reported with the original terminal state
    
Resolves #4429

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -2,7 +2,7 @@ package mesosphere.marathon.core.instance
 
 import java.util.Base64
 
-import com.fasterxml.uuid.{ EthernetAddress, Generators }
+import com.fasterxml.uuid.{EthernetAddress, Generators}
 import mesosphere.marathon.Protos
 import mesosphere.marathon.core.instance.Instance.InstanceState
 import mesosphere.marathon.core.instance.update.InstanceUpdateOperation
@@ -49,7 +49,6 @@ case class Instance(
         InstanceUpdateEffect.Expunge(this)
 
       case InstanceUpdateOperation.MesosUpdate(instance, status, mesosStatus, now) =>
-        // TODO(PODS): calculate the overall state afterwards
         val taskId = Task.Id(mesosStatus.getTaskId)
         val effect = tasks.find(_.taskId == taskId).map { task =>
           task.update(TaskUpdateOperation.MesosUpdate(status, mesosStatus))
@@ -73,11 +72,26 @@ case class Instance(
             InstanceUpdateEffect.Failure(cause)
         }
 
-      case InstanceUpdateOperation.LaunchOnReservation(_, version, status, hostPorts) =>
+      case InstanceUpdateOperation.LaunchOnReservation(_, version, timestamp, status, hostPorts) =>
         if (this.isReserved) {
-          // TODO(PODS) BLOCKER: implement me
-          val updated: Instance = ???
-          InstanceUpdateEffect.Update(updated, Some(this))
+          require(tasksMap.size == 1, "Residency is not yet implemented for task groups")
+
+          val task = tasksMap.values.head
+          val taskEffect = task.update(TaskUpdateOperation.LaunchOnReservation(runSpecVersion, status, hostPorts))
+          taskEffect match {
+            case TaskUpdateEffect.Update(updatedTask) =>
+              val updated = this.copy(
+                state = state.copy(
+                  status = InstanceStatus.Staging,
+                  since = timestamp
+                ),
+                tasksMap = tasksMap.updated(task.taskId, updatedTask)
+              )
+              InstanceUpdateEffect.Update(updated, oldState = Some(this))
+
+            case _ =>
+              InstanceUpdateEffect.Failure(s"Unexpected taskUpdateEffect $taskEffect")
+          }
         } else {
           InstanceUpdateEffect.Failure("LaunchOnReservation can only be applied to a reserved instance")
         }
@@ -86,17 +100,17 @@ case class Instance(
         if (this.isReserved) {
           InstanceUpdateEffect.Expunge(this)
         } else {
-          InstanceUpdateEffect.Failure("LaunchOnReservation can only be applied to a reserved instance")
+          InstanceUpdateEffect.Failure("ReservationTimeout can only be applied to a reserved instance")
         }
 
       case InstanceUpdateOperation.LaunchEphemeral(instance) =>
         InstanceUpdateEffect.Failure("LaunchEphemeral cannot be passed to an existing instance")
 
       case InstanceUpdateOperation.Reserve(_) =>
-        InstanceUpdateEffect.Failure("LaunchEphemeral cannot be passed to an existing instance")
+        InstanceUpdateEffect.Failure("Reserve cannot be passed to an existing instance")
 
       case InstanceUpdateOperation.Revert(oldState) =>
-        InstanceUpdateEffect.Failure("LaunchEphemeral cannot be passed to an existing instance")
+        InstanceUpdateEffect.Failure("Revert cannot be passed to an existing instance")
     }
   }
 
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -132,7 +132,7 @@ case class Instance(
   override def toProto: Protos.Json = {
     Protos.Json.newBuilder().setJson(Json.stringify(Json.toJson(this))).build()
   }
-  override def version: Timestamp = Timestamp.zero
+  override def version: Timestamp = runSpecVersion
 
   override def hostname: String = agentInfo.host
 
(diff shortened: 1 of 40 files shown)


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -46,57 +48,78 @@ case class Instance(
...
+            case TaskUpdateEffect.Noop =>
+              InstanceUpdateEffect.Noop(instance.instanceId)
+
+            case TaskUpdateEffect.Failure(cause) =>
+              InstanceUpdateEffect.Failure(cause)
+
+            case _ =>
+              InstanceUpdateEffect.Failure("ForceExpunge should never delegated to an instance")
+          }
+        }.getOrElse(InstanceUpdateEffect.Failure(s"$taskId not found in $instanceId"))
```
jasongilanfarr: Should there be a not found that extends failed?

jasongilanfarr: Maybe for the other cases too?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceChangedEventsGenerator.scala
```diff
@@ -0,0 +1,45 @@
+package mesosphere.marathon.core.instance.update
+
+import mesosphere.marathon.core.event.{ InstanceChanged, MarathonEvent, MesosStatusUpdateEvent }
+import mesosphere.marathon.core.instance.{ Instance, InstanceStatus }
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.state.Timestamp
+
+import scala.collection.immutable.Seq
+
+class InstanceChangedEventsGenerator {
```
jasongilanfarr: Could be an object.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
```diff
@@ -98,7 +105,8 @@ private[tracker] class InstanceOpProcessorImpl(
         // Used for task termination or as a result from a UpdateStatus action.
         // The expunge is propagated to the instanceTracker which informs the sender about the success (see Ack).
         repository.delete(change.instance.instanceId).map { _ => InstanceTrackerActor.Ack(op.sender, change) }
-          .recoverWith(tryToRecover(op)(expectedState = None, oldState = Some(change.instance)))
```
jasongilanfarr: This and others here could benefit from async/await

src/main/scala/mesosphere/marathon/core/task/update/TaskUpdateOperation.scala
```diff
@@ -1,11 +1,20 @@
...
+import scala.collection.immutable.Seq
+
 /** Trait for operations attempting to change the state of a task */
 trait TaskUpdateOperation extends Product with Serializable
 
 object TaskUpdateOperation {
   case class MesosUpdate(status: InstanceStatus, taskStatus: mesos.Protos.TaskStatus) extends TaskUpdateOperation
+
+  case class LaunchOnReservation(
+    runSpecVersion: Timestamp,
```
jasongilanfarr: Doesn't have the runSpec? Seems like that'd be handy.

jasongilanfarr: Even if it's just the id.

aquamatthias:
retest this please

aquamatthias:
retest this please

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -45,58 +49,79 @@ case class Instance(
...
-      case InstanceUpdateOperation.LaunchOnReservation(_, version, status, hostPorts) =>
+        tasks.find(_.taskId == taskId).map { task =>
+          val taskEffect = task.update(TaskUpdateOperation.MesosUpdate(status, mesosStatus, now))
+          taskEffect match {
+            case TaskUpdateEffect.Update(newTaskState) =>
+              val updated: Instance = updatedInstance(newTaskState, now)
+              val events = eventsGenerator.events(status, updated, Some(task), now)
+              updated.state.status match {
+                case t: Terminal =>
+                  InstanceUpdateEffect.Expunge(updated, events)
```
aquamatthias: We probably should expunge in the case of all tasks are terminal.
The instance can go to a terminal state in those cases:
```
    Error,
    Failed,
    Gone,
    Dropped
```
But not all tasks need to be in a terminal state.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -203,14 +229,20 @@ object Instance {
...
+    def defaultVersion: Timestamp = {
+      // TODO(PODS): fix this
+      log.error("A default Timestamp.zero breaks things!")
+      Timestamp.zero
+    }
+    new Instance(task.taskId.instanceId, task.agentInfo,
+      InstanceState(
+        status = task.status.taskStatus,
+        since = task.status.startedAt.getOrElse(task.status.stagedAt),
+        version = task.version.getOrElse(defaultVersion),
```
aquamatthias: Can we see tasks without version?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceChangedEventsGenerator.scala
```diff
@@ -0,0 +1,45 @@
...
+        status.toMesosStateName,
+        message,
+        appId = runSpecId,
+        host,
+        ipAddresses,
+        ports = ports,
+        version = version.toString,
+        timestamp = now.toString
+      )
+      Seq[MarathonEvent](instanceEvent, taskEvent)
```
aquamatthias: Nit: perhaps change the order, so that subscribers see the MesosStatusUpdate first and the InstanceChange afterwards?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -311,21 +310,22 @@ object Task {
...
         ))
         TaskUpdateEffect.Update(newState = updatedTask)
 
-      case TaskUpdateOperation.MesosUpdate(newStatus, mesosStatus) =>
+      // The Terminal extractor applies specific logic e.g. when an Unreachable task becomes Gone
+      case TaskUpdateOperation.MesosUpdate(newStatus: Terminal, mesosStatus, _) =>
+        val updated = copy(status = status.copy(
+          mesosStatus = Some(mesosStatus),
+          taskStatus = newStatus))
+        TaskUpdateEffect.Update(updated)
```
aquamatthias: The effect was expunge before the change. I guess we will update always on the level on tasks and expunge on the level of instances? (do we need TaskUpdateEffect.Expunge?)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
```diff
@@ -69,7 +75,8 @@ private[tracker] object InstanceOpProcessorImpl {
     private[this] def expungeInstance(id: Instance.Id)(implicit ec: ExecutionContext): Future[InstanceUpdateEffect] = {
       directInstanceTracker.instance(id).map {
         case Some(existingInstance: Instance) =>
-          InstanceUpdateEffect.Expunge(existingInstance)
+          val events = eventsGenerator.events(InstanceStatus.Killed, existingInstance, task = None, clock.now())
```
aquamatthias: Q: why can we say Killed here?

src/main/scala/mesosphere/marathon/core/task/Task.scala
aquamatthias: Should have read to the end. I see you deleted it already. Thanks!

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
meichstedt: Yes, Task.Reserved does currently not have a version unfortunately. An instance will always have one though.

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: Exactly 😎

src/main/scala/mesosphere/marathon/core/instance/update/InstanceChangedEventsGenerator.scala
meichstedt: Makes sense to me, will do in another PR. might break some tests that verify the order though, so I'd do this with lower prio.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
meichstedt: A: Actually we can't. I didn't want to introduce a new state but obviously forgot the `TODO`. Background is that we need to publish an event here that notifies subscribers of _this instance is terminal_. When we forcefully expunge a `Running` task, `Running` is not sufficient for that. Thought about `Deleted`, but I need to further think about that. Thoughts on this?

jdef: What about Terminating (read: this is a termination in progress). Killing might be an instance of Terminating. Suicidal might be an instance of Terminating (meaning we're expecting that Mesos will kill any remaining tasks)

PR #4458, opened 2016-09-22
improvements for pod status health checks

jdef:
* additional fixes and tests for endpoint health
* report container status conditions for `healthy`
* `Unreachable` instances should be labeled as `degraded`

### DIFF
--- a/src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
+++ b/src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
@@ -1,5 +1,7 @@
 package mesosphere.marathon.raml
 
+import java.time.OffsetDateTime
+
 import mesosphere.marathon.core.instance.{ Instance, InstanceStatus }
 import mesosphere.marathon.core.pod.{ MesosContainer, PodDefinition }
 import mesosphere.marathon.core.task.Task
@@ -22,38 +24,7 @@ trait PodStatusConversion {
     }
 
     // possible that a new pod spec might not have a container with a name that was used in an old pod spec?
-    val endpointStatuses: Seq[ContainerEndpointStatus] = maybeContainerSpec.flatMap { containerSpec =>
-      task.launched.flatMap { launched =>
-        task.taskId.containerName.flatMap { containerName =>
-          pod.containers.find(_.name == containerName).flatMap { containerSpec =>
-            val endpointRequestedHostPort: Seq[String] = containerSpec.endpoints.collect {
-              case Endpoint(name, _, Some(_), _, _) => name
-            }(collection.breakOut)
-
-            val reservedHostPorts: Seq[Int] = launched.hostPorts
-
-            assume(
-              endpointRequestedHostPort.size == reservedHostPorts.size,
-              s"number of reserved host ports ${reservedHostPorts.size} should equal number of" +
-                s"requested host ports ${endpointRequestedHostPort.size}")
-
-            // we assume that order has been preserved between the allocated port list and the endpoint list
-            // TODO(jdef) pods what actually guarantees that this doesn't change? (do we check this upon pod update?)
-            val reservedEndpointStatus: Seq[ContainerEndpointStatus] =
-              endpointRequestedHostPort.zip(reservedHostPorts).map { entry =>
-                val (name, allocated) = entry
-                val healthy: Option[Boolean] = launched.status.mesosStatus.withFilter(_.hasHealthy).map(_.getHealthy)
-                ContainerEndpointStatus(name, Some(allocated), healthy)
-              }
-
-            val unreservedEndpointStatus: Seq[ContainerEndpointStatus] = containerSpec.endpoints
-              .withFilter(_.hostPort.isEmpty).map { ep => ContainerEndpointStatus(ep.name) }
-
-            Some(reservedEndpointStatus ++ unreservedEndpointStatus)
-          }
-        }
-      }
-    }.getOrElse(Seq.empty[ContainerEndpointStatus])
+    val endpointStatus = endpointStatuses(pod, maybeContainerSpec, task)
 
     // some other layer should provide termination history
 
@@ -61,13 +32,14 @@ trait PodStatusConversion {
     // the Mesos task ID itself
     val displayName = task.taskId.containerName.getOrElse(task.taskId.mesosTaskId.getValue)
 
-    // TODO(jdef) message, conditions
+    // TODO(jdef) message
     ContainerStatus(
       name = displayName,
       status = task.status.taskStatus.toMesosStateName,
       statusSince = since,
       containerId = task.launchedMesosId.map(_.getValue),
-      endpoints = endpointStatuses,
+      endpoints = endpointStatus,
+      conditions = Seq(maybeHealthCondition(task.status, maybeContainerSpec, endpointStatus, since)).flatten,
       lastUpdated = since, // TODO(jdef) pods fixme
       lastChanged = since // TODO(jdef) pods.fixme
     )
@@ -99,7 +71,11 @@ trait PodStatusConversion {
         InstanceStatus.Gone | InstanceStatus.Dropped | InstanceStatus.Unknown | InstanceStatus.Killing |
         InstanceStatus.Unreachable => PodInstanceState.Terminal
       case InstanceStatus.Running =>
-        if (instance.state.healthy.getOrElse(true)) PodInstanceState.Stable else PodInstanceState.Degraded
+        // TODO(jdef) pods extract constant "healthy"
+        if (containerStatus.exists(_.conditions.exists { cond => cond.name == "healthy" && cond.value == "false" }))
+          PodInstanceState.Degraded
+        else
+          PodInstanceState.Stable
     }
 
     val networkStatus: Seq[NetworkStatus] = networkStatuses(instance.tasks.toVector)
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -22,52 +26,22 @@ trait PodStatusConversion {
...
-    // TODO(jdef) message, conditions
+    // TODO(jdef) message
     ContainerStatus(
       name = displayName,
       status = task.status.taskStatus.toMesosStateName,
       statusSince = since,
       containerId = task.launchedMesosId.map(_.getValue),
-      endpoints = endpointStatuses,
+      endpoints = endpointStatus,
+      conditions = Seq(maybeHealthCondition(task.status, maybeContainerSpec, endpointStatus, since)).flatten,
```
aquamatthias: Instead of `Seq(Option(T)).flatten` you can say `Option[T].toSeq`

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -92,23 +66,19 @@ trait PodStatusConversion {
...
-        if (instance.state.healthy.getOrElse(true)) PodInstanceState.Stable else PodInstanceState.Degraded
-    }
+    val (derivedStatus: PodInstanceState, message: Option[String]) = podInstanceState(
+      instance.state.status, containerStatus)
 
     val networkStatus: Seq[NetworkStatus] = networkStatuses(instance.tasks.toVector)
 
     val resources: Option[Resources] = instance.state.status match {
-      case InstanceStatus.Staging | InstanceStatus.Starting | InstanceStatus.Running =>
+      case InstanceStatus.Staging | InstanceStatus.Starting | InstanceStatus.Running | InstanceStatus.Reserved |
```
aquamatthias: If you import InstanceStatus locally in this method, than this line becomes much more readable.
You should add `Killing` as well

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+    status.taskStatus match {
+      case InstanceStatus.Created | InstanceStatus.Staging | InstanceStatus.Starting | InstanceStatus.Reserved =>
+        // not useful to report health conditions for tasks that have never reached a running state
+        None
+      case _ =>
+        val healthy: Option[(Boolean, String)] = maybeContainerSpec.flatMap { containerSpec =>
+          val usingCommandHealthCheck: Boolean = containerSpec.healthCheck.exists(_.command.nonEmpty)
+          if (usingCommandHealthCheck) {
+            Some(status.mesosStatus.withFilter(_.hasHealthy).map(_.getHealthy).fold(false -> HEALTH_UNREPORTED) { h =>
+              h -> HEALTH_REPORTED
```
aquamatthias: `h` can be ignored (make  it `_`)

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+            })
+          } else {
+            val ep = healthCheckEndpoint(containerSpec)
+            ep.map { endpointName =>
+              val epHealthy: Option[Boolean] = endpointStatuses.find(_.name == endpointName).flatMap(_.healthy)
+              // health check endpoint was specified, but if we don't have a value for health yet then generate a
+              // meaningful reason code
+              epHealthy.fold(false -> HEALTH_UNREPORTED) { h => h -> HEALTH_REPORTED }
```
aquamatthias: replace `h` with `_`

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+
+    status.taskStatus match {
+      case InstanceStatus.Created | InstanceStatus.Staging | InstanceStatus.Starting | InstanceStatus.Reserved =>
+        // not useful to report health conditions for tasks that have never reached a running state
+        None
+      case _ =>
+        val healthy: Option[(Boolean, String)] = maybeContainerSpec.flatMap { containerSpec =>
+          val usingCommandHealthCheck: Boolean = containerSpec.healthCheck.exists(_.command.nonEmpty)
+          if (usingCommandHealthCheck) {
+            Some(status.mesosStatus.withFilter(_.hasHealthy).map(_.getHealthy).fold(false -> HEALTH_UNREPORTED) { h =>
```
aquamatthias: perhaps separate this functionality into a separate method:
```
def taskHealth: Option[Boolean] = ???
```

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+              // health check endpoint was specified, but if we don't have a value for health yet then generate a
+              // meaningful reason code
+              epHealthy.fold(false -> HEALTH_UNREPORTED) { h => h -> HEALTH_REPORTED }
+            }
+          }
+        }
+        healthy.map { h =>
+          StatusCondition(
+            name = STATUS_CONDITION_HEALTHY,
+            lastChanged = since, // TODO(jdef) pods we should be checking for a real change against some history
```
aquamatthias: I think we propagate only state changes, so lastUpdate is probably wrong...

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+    task: Task): Seq[ContainerEndpointStatus] =
+
+    maybeContainerSpec.flatMap { containerSpec =>
+      task.launched.flatMap { launched =>
+
+        val taskHealthy: Option[Boolean] = // only calculate this once so we do it here
+          launched.status.mesosStatus.withFilter(_.hasHealthy).map(_.getHealthy)
+
+        task.taskId.containerName.flatMap { containerName =>
+          pod.containers.find(_.name == containerName).flatMap { containerSpec =>
```
aquamatthias: Add a method to PodDefinition:
```
def container(name: String): Option[MesosContainer]
```
and use it here.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+            val endpointRequestedHostPort: Seq[String] = containerSpec.endpoints.collect {
```
aquamatthias: I would suggest something like: 
```
containerSpec.endpoints.withFilter(_.hostPort.isDefined).map(_.name)
```

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+
+            assume(
+              endpointRequestedHostPort.size == reservedHostPorts.size,
+              s"number of reserved host ports ${reservedHostPorts.size} should equal number of" +
+                s"requested host ports ${endpointRequestedHostPort.size}")
+
+            // we assume that order has been preserved between the allocated port list and the endpoint list
+            // TODO(jdef) pods what actually guarantees that this doesn't change? (do we check this upon pod update?)
+            def reservedEndpointStatus: Map[String, ContainerEndpointStatus] =
+              endpointRequestedHostPort.zip(reservedHostPorts).map { entry =>
```
aquamatthias: instead of entry, you can write `case (name, allocated) =>`

aquamatthias: If ContainerEndpointStatus already has a name, why you use a Map[String, ContainerEndpointStatus] here?
Can it be just Seq[ContainerEndpointStatus]?

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+                val (name, allocated) = entry
+                name -> ContainerEndpointStatus(name, Some(allocated))
+              }.toMap
+
+            def unreservedEndpointStatus: Map[String, ContainerEndpointStatus] = containerSpec.endpoints
+              .withFilter(_.hostPort.isEmpty).map { ep => ep.name -> ContainerEndpointStatus(ep.name) }.toMap
+
+            def withHealth: Seq[ContainerEndpointStatus] = {
+              val allEndpoints = reservedEndpointStatus ++ unreservedEndpointStatus
+              def maybeHealthCheckEndpointName = healthCheckEndpoint(containerSpec)
```
aquamatthias: Perhaps something like this:
```
healthCheckEndpoint(containerSpec)
  .flatMap(name => allEndpoints.get(name).map(_.copy(healthy = taskHealthy))
  .fold(allEndpoints) { ep => allEndpoints += ep.name -> ep }
```

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -143,6 +114,137 @@ trait PodStatusConversion {
...
+          }
+        }
+      }
+    }.getOrElse(Seq.empty[ContainerEndpointStatus])
+
+  def podInstanceState(
+    status: InstanceStatus,
+    containerStatus: Seq[ContainerStatus]): (PodInstanceState, Option[String]) =
+    status match {
+      case InstanceStatus.Created | InstanceStatus.Reserved =>
```
aquamatthias: Please import InstanceStatus locally

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: compiler says that scala.collections.Seq isn't compatible with immutable.Seq :(

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: yes, good catch

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: I need to update the `healthy` property of the endpoint that's linked to the health check. thought it would be better to find the endpoint by looking it up from a Map instead of iterating through the sequence. It probably doesn't matter.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: ack

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jdef: As soon as we have multiple conditions here the .flatten will make more sense

jdef:
.. and rebased

PR #4461, opened 2016-09-22
Create a better set of scala/java conversions.

jasongilanfarr:
- Turn on some additional compiler linter/errors and
  generate b-code which is more native to Java8 and reduces
  the number of class files dramatically
- Create implicit conversions for all java.util.functional interfaces.
- Create package object "mesosphere.marathon" that _correctly_
  defines Seq and ImmutableSeq to be immutable.
  - Going forward, scala files should use "double" package declaration
    (which imports this package object automatically)
    See also: http://www.scala-lang.org/docu/files/ScalaReference.pdf
    Chapter 9, approximately page 135 in the pdf.
- Create Java8 stream collectors that can collect directly into a Scala
  Collection.
- Create JavaConversions that are a very simple view into the immutable
  scala collection.
- Create implicit for java8 streams that make them appear as if they are
  TraversableOnce collections and it will use the underlying stream
  operation in Java8 for most operations.
- Add a new RichTraversableLike which adds "filterAs" and "filterNotAs"
  that take in a CanBuildFrom allowing a filter to use the correct
  collection type desired, instead of always using the source collection
  type (which internally is what it does).
- Create richer conversions from JavaCollections that implement the
  appropriate scala interface and use the streaming API to implement it.
  This appears to be about 40% faster than scala's own conversions and
  allows you to duplicate a lot less.
- In a nutshell, one should always use "import mesosphere.marathon.stream._"
  and not scala.collection.JavaCon*.

Future TODOS:
- Add linter to enforce double package.
- Add linter to enforce no JavaConversions or JavaConverters and prefer
  stream package.

### DIFF
--- a/src/main/scala/mesosphere/marathon/stream/RichTraversableLike.scala
+++ b/src/main/scala/mesosphere/marathon/stream/RichTraversableLike.scala
@@ -1,4 +1,5 @@
-package mesosphere.marathon.stream
+package mesosphere.marathon
+package stream
 
 import scala.collection.TraversableLike
 import scala.collection.generic.CanBuildFrom
--- a/src/main/scala/mesosphere/marathon/stream/RichTraversableLike.scala
+++ b/src/main/scala/mesosphere/marathon/stream/RichTraversableLike.scala
@@ -1,0 +1,23 @@
+package mesosphere.marathon.stream
+
+import scala.collection.TraversableLike
+import scala.collection.generic.CanBuildFrom
+
+/**
+  * Extends traversable with a filter that isn't stupid
+  */
+class RichTraversableLike[+A, +Repr](to: TraversableLike[A, Repr]) {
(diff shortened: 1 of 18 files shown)


jasongilanfarr:
Note to reviewers, I'm about to start a big stack of reviews based on this.

jasongilanfarr:
I'm first stacking the reviews (I have a good couple piled up), then I'll fix left-over issues. This should be a good start for understanding the review stack and getting started.

src/main/scala/mesosphere/marathon/stream/RichTraversableLike.scala
```diff
@@ -0,0 +1,23 @@
+package mesosphere.marathon.stream
```
jdef: Should all of our packages use the double-package notation? If not, what are the exceptions?

jasongilanfarr:
Yes, they should. It will be progressively introduced. I'm first breaking up the reviews, then going through and cleaning them up. In part, I'm doing this to show a process where I've taken a working branch and restructured it into a series of reviews where changes to the parent review propagate through the review chain.

jasongilanfarr:
Please review in phabricator

PR #4462, opened 2016-09-22
It is invalid to supply both "cmd" and "args"  in the same app.

adyatlov:
According to the documentation on the same page.


aquamatthias:
Thanks!

PR #4463, opened 2016-09-23
If deleting the root group, pods should also be deleted.

aquamatthias:


jdef:
Surprised that we don't have a unit test for this

PR #4465, opened 2016-09-23
Refactor integration tests and add Pod tests

nfnt:
This refactors the current integration tests to support tests using the Mesos containerizer and adds a (simple) test for pods. Currently this can be tested only locally because it doesn't work inside of a Docker container -- the Mesos agent needs certain things not available there. Next step would be to update the Velocity infrastructure to support running these tests.

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala
+++ b/src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala
@@ -6,7 +6,9 @@ import java.util.Date
 import akka.actor.ActorSystem
 import mesosphere.marathon.api.v2.json.{ AppUpdate, GroupUpdate }
 import mesosphere.marathon.core.event.{ EventSubscribers, Subscribe, Unsubscribe }
+import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.integration.setup.{ RestResult, SprayHttpResponse }
+import mesosphere.marathon.raml.{ Pod, PodConversion, PodStatus, Raml }
 import mesosphere.marathon.state._
 import mesosphere.marathon.util.Retry
 import org.slf4j.LoggerFactory
@@ -62,7 +64,9 @@ case class ITDeployment(id: String, affectedApps: Seq[String])
   *
   * @param url the url of the remote marathon instance
   */
-class MarathonFacade(url: String, baseGroup: PathId, waitTime: Duration = 30.seconds)(implicit val system: ActorSystem) extends PlayJsonSupport {
+class MarathonFacade(url: String, baseGroup: PathId, waitTime: Duration = 30.seconds)(implicit val system: ActorSystem)
+    extends PlayJsonSupport
+    with PodConversion {
   implicit val scheduler = system.scheduler
   import SprayHttpResponse._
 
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -61,6 +61,21 @@ lazy val formatSettings = SbtScalariform.scalariformSettings ++ Seq(
     .setPreference(SpacesWithinPatternBinders, true)
 )
 
+lazy val integrationTestArgument: Tests.Argument = {
```
jasongilanfarr: Please do this in the test, not the build. You can use scalatest itself to cancel the test with an appropriate message.

src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala
```diff
@@ -168,6 +172,14 @@ class MarathonFacade(url: String, baseGroup: PathId, waitTime: Duration = 30.sec
     result(pipeline(Get(s"$url/v2/apps$id/versions/$version")), waitTime)
   }
 
+  //pod resource ---------------------------------------------
+
+  def createPodV2(pod: PodDefinition): RestResult[Pod] = {
+    requireInBaseGroup(pod.id)
+    val pipeline = marathonSendReceive ~> read[Pod]
+    result(pipeline(Post(s"$url/v2/pods", Raml.toRaml(pod))), waitTime)
```
jasongilanfarr: Why not map it back to PodDef?

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
```diff
@@ -33,6 +33,14 @@ object ProcessKeeper {
 
   private[this] val ENV_MESOS_WORK_DIR: String = "MESOS_WORK_DIR"
 
+  case class MesosConfig(
+    port: Int = 5050,
```
jasongilanfarr: Do not hardcode ports, especially not one I'm probably already using mesos on. (Use PortAllocator)

nfnt:
retest this please

nfnt:
A Velocity task to run integration tests that use Mesos with the Mesos containerizer is now there. I'll be triggering more test runs to make sure that it works correctly.

nfnt:
restest this please

nfnt:
retest this please

nfnt:
retest this please

nfnt:
retest this please

nfnt:
retest this please

nfnt:
retest this please

nfnt:
retest this please

build.sbt
nfnt: Okay, changed it to cancel.

src/test/scala/mesosphere/marathon/integration/setup/ProcessKeeper.scala
nfnt: Using `PortAllocator` now.

tests/integration/app_mock.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
```
jasongilanfarr: Why can't we use the existing app mock?!

src/test/scala/mesosphere/marathon/integration/setup/MesosAppIntegrationTest.scala
```diff
@@ -0,0 +1,147 @@
...
+object MesosDockerIntegrationTag extends Tag("mesosphere.marathon.MesosIntegrationTest")
+
+class MesosAppIntegrationTest
+    extends IntegrationFunSuite
+    with SingleMarathonIntegrationTest
+    with Matchers
+    with BeforeAndAfter
+    with GivenWhenThen {
+
+  override def test(testName: String, testTags: Tag*)(testFun: => Any)(implicit pos: Position): Unit = {
```
jasongilanfarr: Tag the suite?

tests/integration/app_mock.py
nfnt: The App mock needs a Docker image that provides an environment to build a scala class. For our existing integration tests (using `unit-integration-tests.sh`) that's easy: Because we've already build such an image (`build-base-image.sh`) to run the integration tests, we can just re-use that one for the app mock. When testing the Mesos containerizer, this doesn't work anymore, because these tests can't run inside a Docker container. Hence we would need to build that image and make it available to the Mesos containerizer, which is a huge effort, especially for something as simple as the app mock. IMO it's easier to just use an easily available Docker image like `python` that makes it very easy to deploy a simple application.

nfnt:
Yes, the tagging no longer makes sense, because the functionality that used them has been removed with the latest change. I'll remove them.

jasongilanfarr:
Maybe my memory is wrong (and probably is) but I could swear we just have
our classpath in the docker image or something like that. We already have a
working classes directory that can execute app mock (and it's already been
built) so I really don't see a need for another mock. Plus, I'd rather see
it do something (if the above statement isn't true) stupid like use alpine
Linux + netcat or something.

nfnt:
retest this please

nfnt:
retest this please

nfnt:
retest this please

tests/integration/app_mock.py
jasongilanfarr: So... umm, thinking about this even more. you can definitely use the existing app mock instead of adding an additional one with python:

Volume Mount the classpath and JAVA_HOME into an Alpine linux image? I think we already do something quite similar.

nfnt: I could get rid of the python app mock. Now re-using the existing one, rewriting the classpath for the container environment and mounting the target directories.

nfnt:
retest this please

src/test/scala/mesosphere/marathon/integration/setup/MesosAppIntegrationTest.scala
nfnt: Test tags have been removed.

jasongilanfarr:
While I'd prefer a new tag for these special tests (with a new sbt test target), the enviornment variables will do. Can you look into the test failures?

What do we need to do to have the mesos integration tests in the phabricator jenkins jobs too?

jasongilanfarr:
Using tags would let us specifically run only those test cases in the jenkins jobs too.

PR #4467, opened 2016-09-23
Give better error in start script if JAR is not found.

jeschkies:


bin/start
```diff
@@ -71,4 +71,11 @@ fi
 
 # Start Marathon
 marathon_jar=$(find "$FRAMEWORK_HOME"/target -name 'marathon-assembly-*.jar' | sort | tail -1)
+
+if [[ -z $marathon_jar ]]; then
```
zen-dog: -f would make more sense.

PR #4469, opened 2016-09-23
Cache hostname for MarathonConf.

jeschkies:
Somehow `java.net.InetAddress.getLocalHost.getHostName` is blocking and
takes a few seconds (at least with macOS Sierra). So my tests took more
than 40 minutes. Once the hostname is cached the tests take less than a
minute to complete.

I don't think this will ahve any impact in production since the conf
should be created only once.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -20,6 +20,10 @@ import org.rogach.scallop.ScallopConf
 
 import scala.sys.SystemProperties
 
+object FancyHostNameCache {
+  val hostname = java.net.InetAddress.getLocalHost.getHostName
+}
+
 trait MarathonConf
     extends ScallopConf
     with EventConf with GroupManagerConfig with LaunchQueueConfig with LaunchTokenConfig with LeaderProxyConf
@@ -111,7 +115,7 @@ trait MarathonConf
     "hostname",
     descr = "The advertised hostname that is used for the communication with the Mesos master. " +
       "The value is also stored in the persistent store so another standby host can redirect to the elected leader.",
-    default = Some(java.net.InetAddress.getLocalHost.getHostName))
+    default = Some(FancyHostNameCache.hostname))
 
   lazy val webuiUrl = opt[String](
     "webui_url",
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -20,6 +20,10 @@ import org.rogach.scallop.ScallopConf
 
 import scala.sys.SystemProperties
 
+object FancyHostNameCache {
```
jeschkies: I'm willing to debate this choice of name ;)

jeschkies:
I'd be fine with localhost. I don't know why it is retrieving the host name. In any case I don't see a reason why it should be fetched each time.

jeschkies:
Well, I searched for the right commit 3887006a7f41f81767af34b16932e5e37a1734ba. Unfortunately Flo did not mention why he change it from `"localhost"` to `java.net.InetAddress.getLocalHost().getHostName()`.

jasongilanfarr:
It doesn't return localhost, it returns the local hostname of the machine, e.g. for me its "heart-of-gold"

src/main/scala/mesosphere/marathon/MarathonConf.scala
aquamatthias: The `hostname` will be advertised in the cluster, so "localhost" is not a good idea unless everything is local. 
Please:
- move this property into the definition of the lazy val, so it can not be reached anywhere else
- make it lazy

Reason:
1) Nobody should use something like `java.net.InetAddress.getLocalHost.getHostName` directly - everybody should use the hostname property.
2)There are systems that do not have resolvable host names. This method will throw an exception in this case.

jeschkies:
restest this please

jeschkies:
retest this please

jeschkies:
retest this please

jeschkies:
retest this please

jeschkies:
retest this please

jeschkies:
retest this please

jeschkies:
retest this please

jeschkies:
Moved code review to https://phabricator.mesosphere.com/D35

jeschkies:
Reviewed and closed in https://phabricator.mesosphere.com/D35

PR #4473, opened 2016-09-24
WIP: DO NOT MERGE: make heartbeat reactor configuration a bit more flexible

jdef:
also:
* only log after the first N skipped beats
* additional docs

TODO:
- [ ] integration testing

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
@@ -1,6 +1,7 @@
 package mesosphere.marathon.core.heartbeat
 
 import akka.actor._
+import scala.collection.immutable.Seq
 import scala.concurrent.duration._
 
 /**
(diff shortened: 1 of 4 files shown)


jdef:
CI!
```
W: Failed to fetch http://httpredir.debian.org/debian/dists/jessie-backports/main/binary-amd64/Packages  Hash Sum mismatch
```

jdef:
retest this please

jdef:
Something's not right. I don't see any log messages now at all.

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -92,28 +103,46 @@ object Heartbeat {
...
-      new Reactor {
-        def onSkip(): Unit = {
-          log.info("detected skipped heartbeat")
-          r.onSkip()
+    /** afterNMisses only propagates onSkip events after nMisses are exceeded */
+    def afterNMisses(nMisses: Int): Decorator = Decorator { r =>
+      new Adapter(r) {
+        override def onSkip(missed: Int): Unit = {
+          if (missed > nMisses) {
+            super.onSkip(missed)
```
jmlvanre: Once you've skipped the optional heartbeat (waiting on offers, etc.) you can continue to log the `detected skipped heartbeat` right?

src/main/scala/mesosphere/marathon/core/heartbeat/HeartbeatActor.scala
```diff
@@ -66,24 +68,33 @@ class HeartbeatActor(config: Heartbeat.Config) extends LoggingFSM[HeartbeatInter
 object Heartbeat {
   import org.slf4j.LoggerFactory
 
+  /** avoid log files that are overly verbose by only logging missed heartbeats after the first */
+  val LOG_AFTER_N_MISSES = 1
```
jmlvanre: Making this configurable gives the impression that we're controlling how many skipped heartbeats are OK before we consider it a problem.
I think the purpose of this patch is to distinguish between a potential incoming heartbeat, and an explicit one. The reason we don't want to log (or at least log something else) on the initial heartbeat is because we have no control over whether it will happen or not. This means it doesn't imply anything about the stability of the connection.

Our goal should be to log information that allows us to detect partitions or potential partitions.

jeschkies:
Scapegoat is complaining :/.

jdef:
superseded by https://phabricator.mesosphere.com/D83

PR #4481, opened 2016-09-26
Extract the version of Mesos to use in Dockerfiles.

nfnt:
By making the Mesos version a build property, we don't need to repeat it in
every Dockerfile and make it accessible to external build systems that don't
necessarily rely on Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,12 @@ COPY . /marathon
 WORKDIR /marathon
 
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E56151BF && \
+    echo "deb http://repos.mesosphere.com/debian jessie-unstable main" | tee /etc/apt/sources.list.d/mesosphere.list && \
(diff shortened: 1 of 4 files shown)


nfnt:
I don't know `sbt` good enough to know if putting another property in `build.properties` is a good thing. Anyways, it would make sense to put the build version somewhere where it can easily be extracted from the a `Dockerfile`.

jasongilanfarr:
The doesn't work very well. You can put it in "mesosVersion.sbt" or something though

nfnt:
retest this please

jasongilanfarr:
The correct way to do this is to have a settingKey for the mesos version and then to set that somewhere. Then you can either put it in mesos.sbt or in build.sbt and look for something like ```mesos := "version"``` - and parse that via sed. make sense? If not, reassign this to me and i'll take over.

nfnt:
Okay, got it.

nfnt:
retest this please

aquamatthias:
@nfnt Please rebase and set the latest version correctly

aquamatthias:
Thanks!

PR #4482, opened 2016-09-26
Fix an occasional NullPointer in the DeploymentActor

zen-dog:
Using log.xxx() from a future in an actor with ActorLogging trait is bad because it tries to access this.context from another thread.

Fixes: #DCOS-9936


jdef:
is this type of usage something that we should sweep the codebase for? if so, possible to add some kind of linter rule to pick up on things like this?

zen-dog:
We could but it depends. Accessing enclosing actor from the future is bad. sender() is bad but `self` is fine since it is set at the actor initialisation. log.xxx wasn't obvious to me since it implicitly accesses actor's context (because of the ActorLogging trait) but as an actor's val e.g. `val log = Logging(context.system, this)` it would've been fine.

PR #4483, opened 2016-09-26
Address (some) TODOs

aquamatthias:


src/main/scala/mesosphere/marathon/core/pod/PodDefinition.scala
```diff
@@ -55,23 +55,21 @@ case class PodDefinition(
           containers != to.containers ||
           constraints != to.constraints ||
           podVolumes != to.podVolumes ||
-          networks != to.networks
+          networks != to.networks ||
```
jdef: wouldn't the logic be simpler to test the condition "if the IDs match and something other than scaling properties has changed"? 

`id == to.id && this != copy(instances = to.instances, maxInstances = to.maxInstances, version = to.version)`

(along with an appropriate comment)

jdef:
minor comments, otherwise lgtm

Issue #4470, opened 2016-09-23
Marathon fails to upgrade 1.1.1 -> 1.3.0


bobrik:
@alberts, I blame you for this app :)

```
2016-09-23T16:44:44.095 foobar [2016-09-23 16:44:44,094] INFO Elected (LeaderLatchListener Interface) (mesosphere.marathon.core.election.impl.CuratorElectionService:ForkJoinPool-2-worker-29-EventThread)
2016-09-23T16:44:44.095 foobar [2016-09-23 16:44:44,094] INFO As new leader running the driver (mesosphere.marathon.MarathonSchedulerService$$EnhancerByGuice$$46c953c5:ForkJoinPool-2-worker-29-EventThread)
2016-09-23T16:44:44.098 foobar [2016-09-23 16:44:44,097] INFO Migration for storage: Version(1, 1, 1) to current: Version(1, 3, 0): apply change for version: Version(1, 2, 0)  (mesosphere.marathon.state.Migration:ForkJoinPool-2-worker-29)
2016-09-23T16:44:44.098 foobar [2016-09-23 16:44:44,097] INFO Start 1.2 migration (mesosphere.marathon.state.MigrationTo1_2:ForkJoinPool-2-worker-71)
2016-09-23T16:44:44.104 foobar [2016-09-23 16:44:44,103] INFO Discovered 55 tasks for status migration (mesosphere.marathon.state.MigrationTo1_2:ForkJoinPool-2-worker-29)
2016-09-23T16:44:44.109 foobar [2016-09-23 16:44:44,108] ERROR Failed to take over leadership (mesosphere.marathon.core.election.impl.CuratorElectionService:ForkJoinPool-2-worker-29-EventThread)
2016-09-23T16:44:44.109 foobar mesosphere.marathon.MigrationFailedException: while migrating storage to 1.2
2016-09-23T16:44:44.109 foobar 	at mesosphere.marathon.state.Migration$$anonfun$migrations$5$$anonfun$apply$4.applyOrElse(Migration.scala:64)
2016-09-23T16:44:44.110 foobar 	at mesosphere.marathon.state.Migration$$anonfun$migrations$5$$anonfun$apply$4.applyOrElse(Migration.scala:63)
2016-09-23T16:44:44.110 foobar 	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
[...]

PR #4489, opened 2016-09-26
Legacy Storage should log errors if there were invalid protos

jasongilanfarr:
Summary:
#4470 shows that there are errors preventing migration from completing if the
protos changed in an incompatible way. In order to allow migration, we should instead
act as if the object didn't exist and log an error.

Should be cherry-picked into releases/1.3

Test Plan: sbt test

Reviewers: aquamatthias, jdef

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D24


aquamatthias:
LGTM. Still curious why we can not read the proto.

jasongilanfarr:
I'm very curious too. I'd need the binary dump of that node to find out though.

jasongilanfarr:
Retest this please

Issue #4493, opened 2016-09-27
[Pods] Exception during ForwardTaskOp


aquamatthias:
I started a pod with 2 containers (both sleep tasks).
I restarted Marathon and got this exception:

```
marathon[ERROR] I0927 11:01:51.432931 88170496 sched.cpp:1473] Sending ACK for status update 1�M��л<ZI�y� of task mypod2.instance-fb7c1a2a-8490-11e6-862b-2a2bf4e4eca6.sleep1 on agent 9a01f2b3-0086-4386-8042-51ba3f811281-S0 to master@127.0.0.1:5050
marathon [2016-09-27 11:01:51,443] ERROR while processing task status update task_id {
marathon   value: "mypod2.instance-fb7c1a2a-8490-11e6-862b-2a2bf4e4eca6.sleep2"
marathon }
marathon state: TASK_FINISHED
marathon message: "Command exited with status 0"
marathon slave_id {
marathon   value: "9a01f2b3-0086-4386-8042-51ba3f811281-S0"
marathon }
marathon timestamp: 1.474966899411525E9
marathon executor_id {
marathon   value: "instance-mypod2.fb7c1a2a-8490-11e6-862b-2a2bf4e4eca6"
marathon }
marathon healthy: true
marathon source: SOURCE_EXECUTOR
marathon uuid: "`\226\n\255\f\313K\244\275N\365\362\253+\031\211"
marathon container_status {
marathon   network_infos {
marathon     ip_addresses {
marathon       ip_address: "192.168.178.41"
marathon     }
marathon   }
marathon   executor_pid: 66208
marathon }
marathon  (mesosphere.marathon.MarathonScheduler$$EnhancerByGuice$$1c750585:ForkJoinPool-2-worker-19)
[...]

PR #4494, opened 2016-09-27
Related to #4493 Only expunge an instance if all tasks are terminal

meichstedt:
_Note: we're missing lots of unit tests in this area, I will add some_


aquamatthias:
Thanks!

PR #4495, opened 2016-09-27
Internal model for pod volumes, fixes for proto generation

jdef:
* this PR requires the latest mesos protos for SANDBOX_PATH volume source
* @jieyu is hacking the remaining bits of mesos support together for this, will be ready soon

sample pod: https://gist.github.com/jdef/63b0747a9a6c4763fc4a6f01ccf491a4

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/pod/Volume.scala
+++ b/src/main/scala/mesosphere/marathon/core/pod/Volume.scala
@@ -1,0 +1,17 @@
+package mesosphere.marathon.core.pod
+
+sealed trait Volume extends Product with Serializable {
+  val name: String
+}
+
+/**
+  * Ephemeral volumes share the lifetime of the pod instance they're used within and serve to
+  * provide temporary "scratch" space that sub-containers may use to share files, sockets, etc.
+  */
+case class EphemeralVolume(name: String) extends Volume
+
+/**
+  * A host volume refers to some file or directory on the host that should be made available
+  * to one or more sub-containers.
+  */
+case class HostVolume(name: String, hostPath: String) extends Volume
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -165,7 +165,7 @@ object TaskGroupBuilder {
 
     builder.setCommand(commandInfo)
 
-    computeContainerInfo(podDefinition.podVolumes, container)
+    computeContainerInfo(podDefinition.volume, container)
       .foreach(builder.setContainer)
 
     container.healthCheck.foreach { healthCheck =>
@@ -310,7 +310,7 @@ object TaskGroupBuilder {
   }
 
   private[this] def computeContainerInfo(
-    podVolumes: Seq[Volume],
+    volumeForName: String => Volume,
     container: MesosContainer): Option[mesos.ContainerInfo.Builder] = {
 
     val containerInfo = mesos.ContainerInfo.newBuilder.setType(mesos.ContainerInfo.Type.MESOS)
@@ -320,7 +320,7 @@ object TaskGroupBuilder {
       // Read-write mode will be used when the "readOnly" option isn't set.
       val mode = if (volumeMount.readOnly.getOrElse(false)) mesos.Volume.Mode.RO else mesos.Volume.Mode.RW
 
-      podVolumes.find(_.name == volumeMount.name).foreach {
+      volumeForName(volumeMount.name) match {
         case hostVolume: HostVolume =>
           val volume = mesos.Volume.newBuilder()
             .setMode(mode)
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -1,7 +1,7 @@
 package mesosphere.mesos
 
 import mesosphere.marathon.core.instance.Instance
-import mesosphere.marathon.core.pod.{ ContainerNetwork, MesosContainer, PodDefinition }
+import mesosphere.marathon.core.pod._
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.plugin.task.RunSpecTaskProcessor
 import mesosphere.marathon.raml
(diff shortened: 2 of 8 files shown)


src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -290,24 +306,38 @@ object TaskGroupBuilder {
...
+        case hostVolume: HostVolume =>
+          val volume = mesos.Volume.newBuilder()
+            .setMode(mode)
+            .setContainerPath(volumeMount.mountPath)
+            .setHostPath(hostVolume.hostPath) // TODO(jdef) use source type HOST_PATH once it's available
+
+          containerInfo.addVolumes(volume)
+
+        case e: EphemeralVolume =>
+          // TODO(PODS) this depends on some sandbox vols set at the executor level as well
```
jdef: nix this todo

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -28,6 +28,8 @@ object TaskGroupBuilder {
     ).asJava)
     .build
 
+  val ephemeralVolPathPrefix = "volumes/"
```
jdef: /cc @sascala 

ephemeral volumes are hosted in the executor sandbox at `volume/{volume-name}` and are "linked" to by sub-containers. "linked" meaning that if `filesystem/posix` isolation is used by the Mesos agent then it's symlinked to the `volumeMount.path` of the sub-container; if `filesystem/linux` isolation is being used instead (default for dc/os) then bind mounting is used.

jdef: @aquamatthias should the ephemeral volume path prefix actually be configurable?

jasongilanfarr: Maybe as a config? Should it be marathon-volumes? etc?

jdef:
rebased

jdef:
@aquamatthias PTAL

jdef:
Hmm .. looks like we need an updated mesos-utils:
```
[info] Compiling 464 Scala sources and 3 Java sources to /jenkins/workspace/marathon-snapshot-packages-pulls/marathon-pkg/marathon/target/scala-2.11/classes...
[error] /jenkins/workspace/marathon-snapshot-packages-pulls/marathon-pkg/marathon/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala:228: value SANDBOX_PATH is not a member of object org.apache.mesos.Protos.Volume.Source.Type
[error]               .setType(mesos.Volume.Source.Type.SANDBOX_PATH)
[error]                                                 ^
```

src/main/scala/mesosphere/marathon/core/pod/Volume.scala
```diff
@@ -0,0 +1,17 @@
+package mesosphere.marathon.core.pod
+
+sealed trait Volume extends Product with Serializable {
```
aquamatthias: Just curious: why Product and Serializable? (case class implements that, I know)

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -293,24 +310,39 @@ object TaskGroupBuilder {
...
-        // Read-write mode will be used when the "readOnly" option isn't set.
-        val mode = if (volumeMount.readOnly.getOrElse(false)) mesos.Volume.Mode.RO else mesos.Volume.Mode.RW
-
-        volume.setMode(mode)
 
-        containerInfo.addVolumes(volume)
+      // Read-write mode will be used when the "readOnly" option isn't set.
+      val mode = if (volumeMount.readOnly.getOrElse(false)) mesos.Volume.Mode.RO else mesos.Volume.Mode.RW
+
+      podVolumes.find(_.name == volumeMount.name).foreach {
```
aquamatthias: We have places where we reference other objects like volumes, endpoints etc.
Validation ensures, that this reference is always valid.
Perhaps it makes sense to add something like:
```
PodDefinition:
def volumeFor(mount: VolumeMount): Volume
```
which throws an exception in case of an invalid reference.
The way we handle this in a lot of places of the code is:
- complicated  (find and foreach)
- wrong (in case there is no volume this code is not executed

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
aquamatthias: The volume is created "somewhere" and mounted into the containers at the defined path.
So the "somewhere" place does not really matter or?
Your example uses relative paths for the volume mounts - I guess absolute paths would be possible as well?

src/main/scala/mesosphere/marathon/core/pod/Volume.scala
jdef: force subclasses to be case classes. i think we've adopted this pattern elsewhere as well

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jdef: agreed. do you have an exception type in mind for this? `IllegalStateExecption` perhaps?

aquamatthias: `IllegalArgumentException`

jdef: `IllegalArgumentException` has special meaning from an API perspective (see MarathonExceptionMapper). Wouldn't a State exception be more clear as to the root of the problem (we've allowed the model to devolve into a bad state)?

jdef:
@aquamatthias regarding the ephemeral vol path prefix - absolute paths are not possible. the path prefix is prepended to the volume name. the goal is that is that the resulting path is relative to the executor sandbox (since that's where the ephemeral disk is for the pod)

jdef:
.. rebased to feature/pods with the latest mesos.jar

jdef:
tested (a) without docker image; (b) split (one without, one with docker image); (c) both containers use docker image. ephemeral seems to work great in all cases.

TODO: test host vole

jdef:
host vols seem to work fine

aquamatthias:
Thanks!

PR #4501, opened 2016-09-27
Dramatically improve pod validation

jasongilanfarr:
Summary: - pods validation was missing a ton of stuff

Test Plan: sbt test

Reviewers: jdef, aquamatthias, unterstein

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D29

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
@@ -7,7 +7,7 @@ import com.wix.accord.dsl._
 import com.wix.accord.{ Failure, Result, RuleViolation, Success, Validator }
 import mesosphere.marathon.Features
 import mesosphere.marathon.api.v2.Validation
-import mesosphere.marathon.raml.{ Constraint, EnvVars, FixedPodScalingPolicy, PodContainer, Network, NetworkMode, Pod, PodPlacementPolicy, PodScalingPolicy, PodSchedulingBackoffStrategy, PodSchedulingPolicy, PodUpgradeStrategy, Resources, Secrets, Volume }
+import mesosphere.marathon.raml.{ ArgvCommand, Artifact, CommandHealthCheck, Constraint, Endpoint, EnvVars, FixedPodScalingPolicy, HealthCheck, HttpHealthCheck, Image, ImageType, Lifecycle, Network, NetworkMode, Pod, PodContainer, PodPlacementPolicy, PodScalingPolicy, PodSchedulingBackoffStrategy, PodSchedulingPolicy, PodUpgradeStrategy, Resources, Secrets, ShellCommand, TcpHealthCheck, Volume, VolumeMount }
 import mesosphere.marathon.state.{ PathId, ResourceRole }
 
 import scala.collection.immutable.Seq
@@ -17,6 +17,7 @@ import scala.util.Try
 /**
   * Defines implicit validation for pods
   */
+@SuppressWarnings(Array("all")) // wix breaks stuff
 trait PodsValidation {
   import Validation._
 
@@ -49,7 +50,7 @@ trait PodsValidation {
       isTrue[Seq[Network]]("Must specify either a single host network, or else 1-to-n container networks") { nets =>
         val countsByMode = nets.groupBy { net => net.mode }.mapValues(_.size)
         val hostNetworks = countsByMode.getOrElse(NetworkMode.Host, 0)
-        val containerNetworks = countsByMode.get(NetworkMode.Container).getOrElse(0)
+        val containerNetworks = countsByMode.getOrElse(NetworkMode.Container, 0)
         (hostNetworks == 1 && containerNetworks == 0) || (hostNetworks == 0 && containerNetworks > 0)
       }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+      case ShellCommand(shell) =>
+        (shell.length should be > 0)(shell.length)
+      case ArgvCommand(argv) =>
+        (argv.size should be >0)(argv.size)
+    }
+  }
+
+  def healthCheckValidator(endpoints: Seq[Endpoint]) = validator[HealthCheck] { hc =>
+    hc.gracePeriodSeconds should be >= 0L
+    hc.intervalSeconds should be >= 0
```
aquamatthias: What does interval 0 mean?

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+    hc.maxConsecutiveFailures should be >= 0
+    hc.timeoutSeconds should be >= 0
```
aquamatthias: timeout 0 is probably wrong. Also: the timeout should be less than the interval.

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+    hc.delaySeconds should be >= 0
+    hc.http is optional(httpHealthCheckValidator(endpoints))
+    hc.tcp is optional(tcpHealthCheckValidator(endpoints))
+    hc.command is optional(commandCheckValidator)
+    hc is isTrue("Only one of http, tcp, or command may be specified") { hc =>
+      hc.http.isDefined ^ hc.tcp.isDefined ^ hc.command.isDefined
```
aquamatthias: true ^ true ^ true ==> true
If all 3 are defined this is true.

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+    }
+  }
+
+  val endpointValidator = validator[Endpoint] { endpoint =>
+    endpoint.name.length is between(1, 63)
+    endpoint.name should matchRegexFully(NamePattern)
+    endpoint.containerPort.getOrElse(1) is between(1, 65535)
+    endpoint.hostPort.getOrElse(0) is between(0, 65535)
+    endpoint.protocol.size should be > 0
```
aquamatthias: I think we need to restrict the protocol strings to `tcp` or `udp`

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+  }
+
+  val imageValidator = validator[Image] { image =>
+    image.id.length is between(1, 1024)
+  }
+
+  // TODO(PODS): don't we need to know what path we're mounting, not just where it is in the container?
+  // kind of confused by this one.
+  @SuppressWarnings(Array("UnusedParameter"))
+  def volumeMountValidator(volumes: Seq[Volume]): Validator[VolumeMount] = validator[VolumeMount] { volumeMount => // linter:ignore:UnusedParameter
```
aquamatthias:
```scala
volumeMount.name is isTrue("Referenced Volume in VolumeMount does not exist") { volumeName =>
      volumes.exists(_.name == volumeName)
}
```

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+    artifact.destPath.map(_.length).getOrElse(1) is between(1, 1024)
+  }
+
+  val lifeCycleValidator = validator[Lifecycle] { lc =>
+    lc.killGracePeriodSeconds.getOrElse(0.0) should be > 0.0
+  }
+
+  def containerValidator(volumes: Seq[Volume]): Validator[PodContainer] = validator[PodContainer] { container =>
     container.resources is valid(resourceValidator)
+    container.endpoints is empty or every(endpointValidator)
```
aquamatthias: `every` checks every object in the seq. Is there additional value in `is empty or` ?

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+    container.image.getOrElse(Image(ImageType.Docker, "abc")) is valid(imageValidator)
     container.environment is optional(envValidator)
+    container.healthCheck is optional(healthCheckValidator(container.endpoints))
+    container.volumeMounts is empty or every(volumeMountValidator(volumes))
```
aquamatthias: remove `is empty or`

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -64,9 +65,84 @@ trait PodsValidation {
...
+    container.artifacts is empty or every(artifactValidator)
```
aquamatthias: remove `is empty or`

aquamatthias:
I would like to see some tests for this new logic.

jdef:
+1 for tests

also needed:
- endpoint names should be unique within the context of the pod
- volume names should be unique within the context of the pod
- container names should be unique within the context of the pod
- lots of duplication re: name length and pattern matching; can't we consolidate this?

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jasongilanfarr: I actually only added the validation rules that were specified in the raml across the board :/

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jasongilanfarr: Timeout >= 0? Agree on interval. I should have read all of the raml comments on not just the raml restrictions.

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jasongilanfarr: Doh.

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jasongilanfarr: That I missed entirely when reading the raml comments

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jasongilanfarr: Oh wix, so confusing. What's "empty" for?

jasongilanfarr:
Can add more validation, can't realistically remove a lot of the duplication unless we restrict to a new type too, eg, raml.Port, raml.PortAllowEmphemeral, raml.PathId, etc.

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
jdef: pretty sure that mesos doesn't restrict this. why would we?

jasongilanfarr:
Because those are the only two protocols that actually do anything?

jdef:
seems like an artificial restriction. what if someone wanted to specify
http or sctp or ... whatever. why should marathon decide what's appropriate
here?

jasongilanfarr:
Because those are L7 protocols built on L4 protocols and this is as far as
I'm aware a L4 protocol which is only tcp or udp and related to the cgroup
restrictions.

jdef:
this is nothing to do with cgroups. protocol is used for discovery,
possibly port mapping, whatever. it's not used for isolation or allocation

jasongilanfarr:
Port mapping is almost definitely L4 only, are you positive we shouldn't be
restricting it to valid L4 protocols?

jdef:
Not even mesos restricts this:
https://github.com/apache/mesos/blob/94189ee48b8d36c6f395a5b9fb7d8b92873e0802/include/mesos/mesos.proto#L1986

jasongilanfarr:
I see that, but does it work as intended with arbitrary strings? Let's get confirmation.

jdef:
It's up to some mesos module to deal with port mapping / discovery info.
AFAIK arbitrary strings were used on purpose so that module impls wouldn't
be restricted.

PR #4507, opened 2016-09-28
Wait for health status updates during deployments of pod instances

### LINKED ISSUE #4505: Deployments will not wait for health state, if the pod defines health checks
Steps to reproduce:

1. Define a Pod with HealthChecks, where the health checks fail.
1. Start a deployment
1. The deployment may not finish, but it does.

meichstedt:
Wait for health status updates during deployments of pod instances with configured health checks

Resolves #4505


aquamatthias:
Great. Thanks!

PR #4502, opened 2016-09-28
Correct the host port when using http health checks

jasongilanfarr:
Summary:
- Now that everything is validated, we should be able to reliably
  fill in the host port correctly - DCOS-10021

Test Plan:
Used the following json and checked the status after
```
{
  "id": "/podz/pod",
  "labels": {
    "values": {}
  },
  "version": "2016-09-27T21:29:28.397Z",
  "user": null,
  "environment": null,
  "containers": [
    {
      "name": "container1",
      "exec": {
        "command": {
          "shell": "sleep 100"
        },
        "overrideEntrypoint": false
      },
      "resources": {
        "cpus": 1,
        "mem": 128,
        "disk": 0,
        "gpus": 0
      },
      "endpoints": [
        {
          "name": "http",
          "containerPort": 1,
          "hostPort": 0,
          "protocol": ["tcp"],
          "labels": {
            "values": {}
          }
        }
      ],
      "image": {
        "kind": "docker",
        "id": "mesosphere/mesos-slave:latest",
        "forcePull": false
      },
      "environment": null,
      "user": "root",
      "healthCheck": {
        "http": {
          "endpoint": "http",
          "path": "/health",
          "scheme": "http"
        },
        "gracePeriodSeconds": 30,
        "intervalSeconds": 2,
        "maxConsecutiveFailures": 3,
        "timeoutSeconds": 20,
        "delaySeconds": 2
      },
      "volumeMounts": [],
      "artifacts": [],
      "labels": null,
      "lifecycle": {
        "killGracePeriodSeconds": 60
      }
    }
  ],
  "secrets": null,
  "volumes": [],
  "networks": [
    {
      "mode": "host"
    }
  ],
  "scaling": {
    "kind": "fixed",
    "instances": 1,
    "maxInstances": null
  },
  "scheduling": {
    "backoff": {
      "backoff": 1,
      "backoffFactor": 1.15,
      "maxLaunchDelay": 3600
    },
    "upgrade": {
      "minimumHealthCapacity": 1,
      "maximumOverCapacity": 1
    },
    "placement": {
      "constraints": [],
      "acceptedResourceRoles": []
    }
  }
}
```

Reviewers: jdef, aquamatthias, unterstein

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D30

PR #4510, opened 2016-09-28
Correct the host port when using http health checks

aquamatthias:


aquamatthias:
Cherry picked the commit from #4502 and fixed the lookup

jdef:
Looks like a couple of TODOs were wiped out by accident. Health checks should work with container ports when the pod is running on a container network.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -342,14 +341,20 @@ object TaskGroupBuilder {
...
+    hostPorts: Seq[Option[Int]]): mesos.HealthCheck.Builder = {
+
     val builder = mesos.HealthCheck.newBuilder
     builder.setDelaySeconds(healthCheck.delaySeconds.toDouble)
     builder.setGracePeriodSeconds(healthCheck.gracePeriodSeconds.toDouble)
     builder.setIntervalSeconds(healthCheck.intervalSeconds.toDouble)
     builder.setConsecutiveFailures(healthCheck.maxConsecutiveFailures)
     builder.setTimeoutSeconds(healthCheck.timeoutSeconds.toDouble)
 
+    lazy val hostPortsByEndpoint: Map[String, Option[Int]] = {
```
jdef: worth an `assume` to check that hostPorts and the collection of endpoint names are the same size?

jdef:
minor comment. otherwise lgtm

PR #4511, opened 2016-09-28
Every Container should have its own ports environment.

aquamatthias:


jdef:
I don't understand what's driving this. Why don't we want all sub-containers to get all the ports info? For example, one sub-container could be hosting a server on a dynamically configured port and a peer sub-container could be responsible for publishing port metadata into a registry.

jasongilanfarr:
This is related to my comment- I think the ports env bars should be containerPort_hostPort inside the container and containerName_hostPort for all containers.

meichstedt:
The test case makes this quite obvious I guess:
```
assert(task1Env("PORT0") == "8080")
assert(task2Env("PORT0") == "8081")
```

It should be possible to specify `PORT0` _relative_ to a containers endpoint, not relative to the aggregated endpoints of all containers.

jasongilanfarr:
James has a point though, we need environment bars that say container named
X has port named X at Y.

Also, numbered port env vars are kind of shitty. We have names, use them.

aquamatthias:
The `PORTXX` env vars are the ports of this container.
It is hard/impossible otherwise to know which port to bind to.
It is totally ok to extend the env with all ports of the pod.
But this should be separate env vars - possibly like Jason suggested.

jdef:
please follow this up with a PR for `PORT_EP_{endpointName}` or something so that all containers of the pod know all the endpoints/ports of all other containers.

PR #4516, opened 2016-09-29
Add Endpoint Env Vars for the whole pod on every task.

aquamatthias:
Form: `ENDPOINT_<ENDPOINTNAME>=123`

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -5,7 +5,6 @@ import mesosphere.marathon.core.pod._
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.plugin.task.RunSpecTaskProcessor
 import mesosphere.marathon.raml
-import mesosphere.marathon.raml.Endpoint
 import mesosphere.marathon.state.{ EnvVarString, PathId, Timestamp }
 import mesosphere.marathon.tasks.PortsMatch
 import mesosphere.mesos.ResourceMatcher.ResourceSelector
@@ -443,32 +442,18 @@ object TaskGroupBuilder {
     builderConfig: BuilderConfig): Map[String, String] = {
     val prefix = builderConfig.envVarsPrefix.getOrElse("").toUpperCase
     def escape(name: String): String = name.replaceAll("[^A-Z0-9_]+", "_").toUpperCase
-    def endpointEnvName(endpoint: Endpoint) = s"${prefix}ENDPOINT_${escape(endpoint.name.toUpperCase)}"
 
     val hostNetwork = pod.networks.contains(HostNetwork)
-    lazy val hostPortByEndpoint = pod.containers.flatMap(_.endpoints).zip(hostPorts).toMap.withDefaultValue(None)
+    val hostPortByEndpoint = pod.containers.flatMap(_.endpoints).zip(hostPorts).toMap.withDefaultValue(None)
     pod.containers.flatMap(_.endpoints).flatMap{ endpoint =>
       val mayBePort = if (hostNetwork) hostPortByEndpoint(endpoint) else endpoint.containerPort
-      mayBePort.map(p => endpointEnvName(endpoint) -> p.toString)
+      val envName = escape(endpoint.name.toUpperCase)
+      Seq(
+        mayBePort.map(p => s"${prefix}ENDPOINT_$envName" -> p.toString),
+        hostPortByEndpoint(endpoint).map(p => s"${prefix}EP_HOST_$envName" -> p.toString),
+        endpoint.containerPort.map(p => s"${prefix}EP_CONTAINER_$envName" -> p.toString)
+      ).flatten
     }.toMap
-  }
-
-  private[this] def portEnvVars(
-    endpoints: Seq[raml.Endpoint],
-    hostPorts: Seq[Option[Int]],
-    envPrefix: Option[String]): Map[String, String] = {
-    // TODO(nfnt): Refactor this to use portMappings
-    val declaredPorts = endpoints.flatMap(_.containerPort)
-    val portNames = endpoints.map(endpoint => Some(endpoint.name))
-
-    val portEnvVars = EnvironmentHelper.portsEnv(declaredPorts, hostPorts, portNames)
-
-    envPrefix match {
-      case Some(prefix) =>
-        portEnvVars.map{ case (key, value) => (prefix + key, value) }
-      case None =>
-        portEnvVars
-    }
   }
 
   private[this] def taskContextEnv(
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -434,6 +433,41 @@ object TaskGroupBuilder {
...
+  /**
+    * Computes all endpoint env vars for the entire pod definition
+    * Form: ENDPOINT_<ENDPOINT_NAME>=123
+    */
+  private[this] def endpointEnvVars(
+    pod: PodDefinition,
+    hostPorts: Seq[Option[Int]],
+    builderConfig: BuilderConfig): Map[String, String] = {
+    val prefix = builderConfig.envVarsPrefix.getOrElse("").toUpperCase
+    def endpointEnvName(endpoint: Endpoint) = s"${prefix}ENDPOINT_${endpoint.name.toUpperCase}"
```
jdef: endpoint names may contain hyphens, probably want to change these to underscore

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -434,6 +433,46 @@ object TaskGroupBuilder {
...
+    def endpointEnvName(endpoint: Endpoint) = s"${prefix}ENDPOINT_${escape(endpoint.name.toUpperCase)}"
+
+    val allEndpointNames = pod.containers.flatMap(_.endpoints.map(e => escape(e.name.toUpperCase)))
+
+    val hostNetwork = pod.networks.contains(HostNetwork)
+    lazy val hostPortByEndpoint = pod.containers.flatMap(_.endpoints).zip(hostPorts).toMap.withDefaultValue(None)
+    pod.containers.flatMap(_.endpoints).flatMap{ endpoint =>
+      val mayBePort = if (hostNetwork) hostPortByEndpoint(endpoint) else endpoint.containerPort
+      mayBePort.map(p => endpointEnvName(endpoint) -> p.toString)
+    }.toMap + ("ENDPOINT_NAMES" -> allEndpointNames.mkString(","))
```
jdef: dislike ENDPOINT_NAMES because if someone named their endpoint `names` then there's a conflict. need a different prefix

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -434,6 +433,46 @@ object TaskGroupBuilder {
     builder
   }
 
+  /**
+    * Computes all port env vars for the given container.
+    * Form: PORT{INDEX}=123
+    */
+  private[this] def containerPortEnvVars(
```
jasongilanfarr: This should only do named endpoints, e.g.

ENDPOINT_${NAME}_CONTAINER_PORT_${PORTNUM}=${HOSTPORT}

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -434,6 +433,46 @@ object TaskGroupBuilder {
...
+    val networkPorts = if (hostNetworkMode) hostPorts else pod.containers.flatMap(_.endpoints).map(_.containerPort)
+    val allEndpoints = pod.containers.flatMap(_.endpoints.map(_.name)).zip(networkPorts).toMap.withDefaultValue(None)
+    val containerEndpoints = allEndpoints.filter{ case (name, port) => container.endpoints.exists(_.name == name) }
+    portEnvVars(container.endpoints, containerEndpoints.values.toVector, config.envVarsPrefix)
+  }
+
+  /**
+    * Computes all endpoint env vars for the entire pod definition
+    * Form:
+    * ENDPOINT_NAMES={ENDPOINT_NAME},{ENDPOINT_NAME},..,{ENDPOINT_NAME}
```
jasongilanfarr: I think its sufficient to follow a single pattern, e.g. ENDPOINT_${NAME}=${HOST_PORT}

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
aquamatthias: Removed this variable completely.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
aquamatthias: Removed the whole containerPortEnvVars

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
aquamatthias: Removed this variable

aquamatthias:
@jdef @jasongilanfarr I removed the PORT{index} handling

jasongilanfarr:
:+1:

jasongilanfarr:
Probably not scapegoat clean yet :/

jdef:
@jasongilanfarr don't understand your last comment "..sufficient". If I declare an endpoint with both a host-port and container-port .. I may want to be able to "discover" both ports, no?

aquamatthias:
retest this please

nfnt:
Inside a container (of task A) you may only care about the container port, but outside of the container (in task B), you only care about the host port (of task A).
Though, from inside a container you may want to signal some other task, with which endpoint you may reach it. In that case, having the host port available inside the container makes sense.

jdef:
I maintain that we should provide host port mapping information for pods using container networks w/ mapped host ports.

`$ENDPOINT_{name}` -- this is what we have now, picks the host port when in `host` network node; picks the container port when in `container` network mode. Gets into trouble when a pod wants to advertise services outside of the cluster because it doesn't know its host port (and things outside of the cluster probably can't connect on the container network).

Proposal:

`EP_HOST_PORT_{name}` -- always points to an allocated host port, regardless of net mode
`EP_CT_PORT_{name}` -- points to a container port, only useful in container mode

Plugins could provide additional:
`EP_VIP_{name}_{vipname}` -- dc/os support for discoverable VIPs

jasongilanfarr:
@jdef what I am saying is that we can keep it very simple:

ENDPOINT_$NAME=$HOSTPORT 

And nothing more - there is no need for containter port mapping or anything becuase this already exposes all the information you actually need.

jasongilanfarr:
I think that anything with HOST_PORT CONTAINER_PORT etc defeats the simplicity and does not add practical value, e.g. take a hypothetical pod (like phabricator)

nginx image by default listens on port 80. "http containerPort 80 hostPort 13123"
nginx image has a layer that drops in an upstream at "$ENDPOINT_APPLICATION"
mysql image listens on 3306 "mysql containerPort 3306 hostPort 13124"
custom app listens on port 81 "application containerPort 81 hostPort 13125"

Where do we ever need to know what the container port is?

PR #4517, opened 2016-09-29
fix broken envvar name validation

jdef:


jdef:
would be nice to see `environment(someNameHere)` instead of `environment(0)` because the 0 doesn't help anyone

aquamatthias:
Thanks!

Issue #4212, opened 2016-08-12
TaskLostIntegrationTest is unstable


jasongilanfarr:

unterstein:
@jasongilanfarr still flaky? increased the timeout for a while

PR #4521, opened 2016-09-29
lazy cache for versions

jdef:
Lazy cache for versions and versioned objects. Once the versioned object cache reaches a size threshold some number of entries are purged.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
+++ b/src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
@@ -11,12 +11,14 @@ import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.Protos.StorageVersion
 import mesosphere.marathon.core.storage.store.impl.BasePersistenceStore
 import mesosphere.marathon.core.storage.store.{ IdResolver, PersistenceStore }
+import mesosphere.marathon.storage.VersionCacheConfig
 import mesosphere.util.LockManager
 
 import scala.async.Async.{ async, await }
 import scala.collection.concurrent.TrieMap
 import scala.collection.immutable.Seq
 import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.Random
 
 /**
   * A Write Ahead Cache of another persistence store that lazily loads values into the cache.
@@ -27,8 +29,8 @@ import scala.concurrent.{ ExecutionContext, Future }
   * @tparam K The persistence store's primary key type
   * @tparam Serialized The serialized format for the persistence store.
   */
-class LazyCachingPersistenceStore[K, Category, Serialized](
-    val store: BasePersistenceStore[K, Category, Serialized])(implicit
+case class LazyCachingPersistenceStore[K, Category, Serialized](
+    store: BasePersistenceStore[K, Category, Serialized])(implicit
   mat: Materializer,
     ctx: ExecutionContext) extends PersistenceStore[K, Category, Serialized] with StrictLogging {
 
@@ -64,7 +66,7 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
     delete: () => Future[Done])(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
     val category = ir.category
     val storageId = ir.toStorageId(k, None)
-    lockManager.executeSequentially(ir.category.toString) {
+    lockManager.executeSequentially(category.toString) {
       lockManager.executeSequentially(storageId.toString) {
         async { // linter:ignore UnnecessaryElseBranch
           await(delete())
@@ -85,6 +87,7 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
   override def deleteCurrent[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
     deleteCurrentOrAll(k, () => store.deleteCurrent(k))
   }
+
   override def deleteAll[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
     deleteCurrentOrAll(k, () => store.deleteAll(k))
   }
@@ -158,4 +161,195 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
     store.deleteVersion(k, version)
 
   override def toString: String = s"LazyCachingPersistenceStore($store)"
+}
+
+case class LazyVersionCachingPersistentStore[K, Category, Serialized](
+    store: PersistenceStore[K, Category, Serialized],
+    config: VersionCacheConfig = VersionCacheConfig.Default)(implicit
+  mat: Materializer,
+    ctx: ExecutionContext) extends PersistenceStore[K, Category, Serialized] with StrictLogging {
+
+  private val lockManager = LockManager.create()
+  private[store] val versionCache = TrieMap.empty[(Category, K), Seq[OffsetDateTime]]
+  private[store] val versionedValueCache = TrieMap.empty[(K, OffsetDateTime), Option[Any]]
+
+  def withVersionedValueCache[Id, V, T](
+    future: => Future[T])(implicit ir: IdResolver[Id, V, Category, K]): Future[T] =
+    lockManager.executeSequentially(s"versionedValueCache::${ir.category}")(future)
+
+  def withVersionCache[Id, V, T](
+    id: Id)(future: (Category, K) => Future[T])(implicit ir: IdResolver[Id, V, Category, K]): Future[T] = {
+    val category = ir.category
+    val storageId = ir.toStorageId(id, None)
+    lockManager.executeSequentially((category, storageId).toString())(future(category, storageId))
+  }
+
+  private[cache] def maybePurgeCachedVersions(
+    maxEntries: Int = config.maxEntries,
+    purgeCount: Int = config.purgeCount,
+    pRemove: Double = config.pRemove): Unit =
+
+    while (versionedValueCache.size > maxEntries) {
+      // randomly GC the versions
+      var counter = 0
+      versionedValueCache.retain { (k, v) =>
+        val x = Random.nextDouble()
+        x > pRemove || { counter += 1; counter > purgeCount }
+      }
+    }
+
+  /**
+    * Assumes that callers have already implemented appropriate locking via [[withVersionCache]] and
+    * [[withVersionedValueCache]].
+    */
+  private[this] def updateCachedVersions[V](
+    storageId: K,
+    version: OffsetDateTime,
+    category: Category,
+    v: Option[V]): Unit = {
+
+    maybePurgeCachedVersions()
+    versionedValueCache.put((storageId, version), v)
+    val cached = versionCache.getOrElse((category, storageId), Nil) // linter:ignore UndesirableTypeInference
+    versionCache.put((category, storageId), (version +: cached).distinct)
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  private def deleteCurrentOrAll[Id, V](
+    id: Id, delete: () => Future[Done])(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
+
+    if (!ir.hasVersions) {
+      delete()
+    } else {
+      withVersionCache(id) { (category, storageId) =>
+        withVersionedValueCache {
+          async {
+            // linter:ignore UnnecessaryElseBranch
+            await(delete())
+            versionedValueCache.retain { case ((sid, version), v) => sid != storageId }
+            versionCache.remove((category, storageId))
+            Done
+          }
+        }
+      }
+    }
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def get[Id, V](id: Id)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    um: Unmarshaller[Serialized, V]): Future[Option[V]] = {
+
+    if (!ir.hasVersions) {
+      store.get(id)
+    } else {
+      withVersionCache(id) { (category, storageId) =>
+        withVersionedValueCache {
+          async { // linter:ignore UnnecessaryElseBranch
+            val value = await(store.get(id))
+            value.foreach { v =>
+              val version = ir.version(v)
+              updateCachedVersions(storageId, version, category, value)
+            }
+            value
+          }
+        }
+      }
+    }
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def get[Id, V](id: Id, version: OffsetDateTime)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    um: Unmarshaller[Serialized, V]): Future[Option[V]] = {
+
+    withVersionCache(id) { (category, storageId) =>
+      withVersionedValueCache {
+        val cached = versionedValueCache.get((storageId, version)) // linter:ignore OptionOfOption
+        cached match {
+          case Some(v: Option[V] @unchecked) =>
+            Future.successful(v)
+          case _ =>
+            async { // linter:ignore UnnecessaryElseBranch
+              val value = await(store.get(id, version))
+              updateCachedVersions(storageId, version, category, value)
+              value
+            }
+        }
+      }
+    }
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def store[Id, V](id: Id, v: V)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized]): Future[Done] = {
+
+    if (!ir.hasVersions) {
+      store.store(id, v)
+    } else {
+      withVersionCache(id) { (category, storageId) =>
+        withVersionedValueCache {
+          async { // linter:ignore UnnecessaryElseBranch
+            await(store.store(id, v))
+            val version = ir.version(v)
+            updateCachedVersions(storageId, version, category, Some(v))
+            Done
+          }
+        }
+      }
+    }
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def store[Id, V](id: Id, v: V, version: OffsetDateTime)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized]): Future[Done] = {
+
+    withVersionCache(id) { (category, storageId) =>
+      withVersionedValueCache {
+        async { // linter:ignore UnnecessaryElseBranch
+          await(store.store(id, v, version))
+          updateCachedVersions(storageId, version, category, Some(v))
+          Done
+        }
+      }
+    }
+  }
+
+  override def deleteCurrent[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] =
+    deleteCurrentOrAll(k, () => store.deleteCurrent(k))
+
+  override def deleteAll[Id, V](k: Id)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] =
+    deleteCurrentOrAll(k, () => store.deleteAll(k))
+
+  override def deleteVersion[Id, V](
+    k: Id,
+    version: OffsetDateTime)(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] =
+    deleteCurrentOrAll(k, () => store.deleteVersion(k, version))
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def versions[Id, V](id: Id)(implicit ir: IdResolver[Id, V, Category, K]): Source[OffsetDateTime, NotUsed] = {
+    val versionsFuture = withVersionCache(id) { (category, storageId) =>
+      if (versionCache.contains((category, storageId))) {
+        Future.successful(versionCache((category, storageId)))
+      } else {
+        async { // linter:ignore UnnecessaryElseBranch
+          val children = await(store.versions(id).toMat(Sink.seq)(Keep.right).run())
+          versionCache((category, storageId)) = children
+          children
+        }
+      }
+    }
+    Source.fromFuture(versionsFuture).mapConcat(identity)
+  }
+
+  override def ids[Id, V]()(implicit ir: IdResolver[Id, V, Category, K]): Source[Id, NotUsed] = store.ids()
+
+  override def storageVersion(): Future[Option[StorageVersion]] = store.storageVersion()
+
+  override def setStorageVersion(storageVersion: StorageVersion): Future[Done] =
+    store.setStorageVersion(storageVersion)
+
+  override def toString: String = s"LazyVersionCachingPersistenceStore($store)"
 }
--- a/src/main/scala/mesosphere/marathon/storage/StorageConf.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageConf.scala
@@ -32,4 +32,13 @@ trait StorageConf extends ZookeeperConf {
     hidden = true,
     descr = "Max outstanding requests to Zookeeper persistence"
   )
+
+  lazy val versionCacheEnabled = toggle(
+    "internal_version_caching",
+    default = Some(true),
+    noshort = true,
+    hidden = true,
+    descrYes = "(Default) Enable an additional layer of caching for object versions when store_cache is enabled.",
+    prefix = "disable_"
+  )
 }
--- a/src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
@@ -346,8 +346,7 @@ object InMem {
 }
 
 object StorageConfig {
-  // TODO(jdef) disabled globally until proven; eventually the default value here will be non-empty
-  val DefaultVersionCacheConfig = Option.empty[VersionCacheConfig]
+  val DefaultVersionCacheConfig = Option(VersionCacheConfig.Default)
 
   val DefaultLegacyMaxVersions = 25
   val DefaultMaxVersions = 5000
--- a/src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
@@ -10,7 +10,7 @@ import akka.stream.Materializer
 import com.typesafe.config.{ Config, ConfigMemorySize }
 import mesosphere.marathon.core.storage.store.PersistenceStore
 import mesosphere.marathon.core.storage.store.impl.BasePersistenceStore
-import mesosphere.marathon.core.storage.store.impl.cache.{ LazyCachingPersistenceStore, LoadTimeCachingPersistenceStore }
+import mesosphere.marathon.core.storage.store.impl.cache.{ LazyCachingPersistenceStore, LazyVersionCachingPersistentStore, LoadTimeCachingPersistenceStore }
 import mesosphere.marathon.core.storage.store.impl.memory.{ Identity, InMemoryPersistenceStore, RamId }
 import mesosphere.marathon.core.storage.store.impl.zk.{ NoRetryPolicy, RichCuratorFramework, ZkId, ZkPersistenceStore, ZkSerialized }
 import mesosphere.marathon.metrics.Metrics
@@ -186,17 +186,47 @@ object CacheType {
 sealed trait PersistenceStorageConfig[K, C, S] extends StorageConfig {
   val maxVersions: Int
   val cacheType: CacheType
+  val versionCacheConfig: Option[VersionCacheConfig]
+
   protected def leafStore(implicit metrics: Metrics, mat: Materializer, ctx: ExecutionContext,
     scheduler: Scheduler, actorRefFactory: ActorRefFactory): BasePersistenceStore[K, C, S]
+
+  protected def lazyStore(implicit metrics: Metrics, mat: Materializer, ctx: ExecutionContext,
+    scheduler: Scheduler, actorRefFactory: ActorRefFactory): PersistenceStore[K, C, S] = {
+    val lazyCachingStore: PersistenceStore[K, C, S] = LazyCachingPersistenceStore(leafStore)
+    versionCacheConfig.fold(lazyCachingStore){ config => LazyVersionCachingPersistentStore(lazyCachingStore, config) }
+  }
 
   def store(implicit metrics: Metrics, mat: Materializer,
     ctx: ExecutionContext, scheduler: Scheduler, actorRefFactory: ActorRefFactory): PersistenceStore[K, C, S] = {
     cacheType match {
       case NoCaching => leafStore
-      case LazyCaching => new LazyCachingPersistenceStore[K, C, S](leafStore)
+      case LazyCaching => lazyStore
       case EagerCaching => new LoadTimeCachingPersistenceStore[K, C, S](leafStore)
     }
   }
+}
+
+case class VersionCacheConfig(
+  maxEntries: Int,
+  purgeCount: Int,
+  pRemove: Double
+)
+
+object VersionCacheConfig {
+  /**
+    * max number of entries allowed in the versioned value cache before entries are purged
+    */
+  protected val MaxVersionedCacheSize = 10000
+  /**
+    * probability that, during a purge of the versioned value cache, a given entry will be removed
+    */
+  protected val ProbabilityToRemoveFromCache = 0.05
+
+  val Default = apply(MaxVersionedCacheSize, ProbabilityToRemoveFromCache)
+
+  def apply(maxEntries: Int, pRemove: Double): VersionCacheConfig =
+    new VersionCacheConfig(maxEntries, (maxEntries * pRemove).toInt, pRemove)
 }
 
 case class CuratorZk(
@@ -213,7 +243,9 @@ case class CuratorZk(
     retryConfig: RetryConfig,
     maxConcurrent: Int,
     maxOutstanding: Int,
-    maxVersions: Int) extends PersistenceStorageConfig[ZkId, String, ZkSerialized] {
+    maxVersions: Int,
+    versionCacheConfig: Option[VersionCacheConfig]
+) extends PersistenceStorageConfig[ZkId, String, ZkSerialized] {
 
   lazy val client: RichCuratorFramework = {
     val builder = CuratorFrameworkFactory.builder()
@@ -262,7 +294,8 @@ object CuratorZk {
       retryConfig = RetryConfig(),
       maxConcurrent = conf.zkMaxConcurrency(),
       maxOutstanding = 1024,
-      maxVersions = conf.maxVersions()
+      maxVersions = conf.maxVersions(),
+      versionCacheConfig = if (conf.versionCacheEnabled()) StorageConfig.DefaultVersionCacheConfig else None
     )
 
   def apply(config: Config): CuratorZk = {
@@ -286,13 +319,16 @@ object CuratorZk {
       retryConfig = RetryConfig(config),
       maxConcurrent = config.int("max-concurrent-requests", 32),
       maxOutstanding = config.int("max-concurrent-outstanding", 1024),
-      maxVersions = config.int("max-versions", StorageConfig.DefaultMaxVersions)
+      maxVersions = config.int("max-versions", StorageConfig.DefaultMaxVersions),
+      versionCacheConfig =
+        if (config.bool("version-cache-enabled", true)) StorageConfig.DefaultVersionCacheConfig else None
     )
   }
 }
 
 case class InMem(maxVersions: Int) extends PersistenceStorageConfig[RamId, String, Identity] {
   override val cacheType: CacheType = NoCaching
+  override val versionCacheConfig: Option[VersionCacheConfig] = None
 
   protected def leafStore(implicit metrics: Metrics, mat: Materializer, ctx: ExecutionContext,
     scheduler: Scheduler, actorRefFactory: ActorRefFactory): BasePersistenceStore[RamId, String, Identity] =
@@ -310,6 +346,9 @@ object InMem {
 }
 
 object StorageConfig {
+  // TODO(jdef) disabled globally until proven; eventually the default value here will be non-empty
+  val DefaultVersionCacheConfig = Option.empty[VersionCacheConfig]
+
   val DefaultLegacyMaxVersions = 25
   val DefaultMaxVersions = 5000
   def apply(conf: StorageConf)(implicit metrics: Metrics, actorRefFactory: ActorRefFactory): StorageConfig = {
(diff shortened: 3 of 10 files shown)


src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +165,39 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
       }
     }
   }
 
-  override def versions[Id, V](id: Id)(implicit ir: IdResolver[Id, V, Category, K]): Source[OffsetDateTime, NotUsed] =
-    store.versions(id)
+  override def versions[Id, V](id: Id)(implicit ir: IdResolver[Id, V, Category, K]): Source[OffsetDateTime, NotUsed] = {
+    val category = ir.category
+    val storageId = ir.toStorageId(id, None)
+    val versionsFuture = lockManager.executeSequentially(category.toString) {
```
jdef: am I missing a lock on storageId?

jdef: fixed

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -101,8 +107,13 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
           Future.successful(v)
         case _ =>
           async { // linter:ignore UnnecessaryElseBranch
-            val value = await(store.get(id))
+            val value: Option[V] = await(store.get(id))
```
jasongilanfarr: why the type annotation? just curious.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -111,8 +122,22 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
   override def get[Id, V](id: Id, version: OffsetDateTime)(implicit
     ir: IdResolver[Id, V, Category, K],
-    um: Unmarshaller[Serialized, V]): Future[Option[V]] =
-    store.get(id, version)
+    um: Unmarshaller[Serialized, V]): Future[Option[V]] = {
+    val storageId = ir.toStorageId(id, None)
+    lockManager.executeSequentially(storageId.toString) {
+      val cached = versionedValueCache.get((storageId, version)) // linter:ignore OptionOfOption
+      cached match {
+        case Some(v: Option[V] @unchecked) =>
```
jasongilanfarr: isn't "Some(v)" sufficeint?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -111,8 +122,22 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
 
   override def get[Id, V](id: Id, version: OffsetDateTime)(implicit
     ir: IdResolver[Id, V, Category, K],
-    um: Unmarshaller[Serialized, V]): Future[Option[V]] =
-    store.get(id, version)
+    um: Unmarshaller[Serialized, V]): Future[Option[V]] = {
+    val storageId = ir.toStorageId(id, None)
+    lockManager.executeSequentially(storageId.toString) {
```
jasongilanfarr: Should this be "(storageId, version).toString"?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -101,8 +107,13 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
           Future.successful(v)
         case _ =>
           async { // linter:ignore UnnecessaryElseBranch
-            val value = await(store.get(id))
+            val value: Option[V] = await(store.get(id))
             valueCache.put(storageId, value)
+            value.foreach { v =>
+              if (ir.hasVersions) {
+                versionedValueCache.put((storageId, ir.version(v)), value)
```
jasongilanfarr: need to lock around "(storageId, ir.version(v)).toString)"

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +169,61 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
     val category = ir.category
     val storageId = ir.toStorageId(id, None)
     lockManager.executeSequentially(category.toString) {
-      async { // linter:ignore UnnecessaryElseBranch
-        await(store.store(id, v, version))
-        val old = idCache.getOrElse(category, Nil) // linter:ignore UndesirableTypeInference
-        idCache.put(category, id +: old)
-        Done
+      lockManager.executeSequentially(storageId.toString) {
```
jasongilanfarr: need to lock on ((storageId, version).toString)

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +169,61 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
       }
     }
   }
 
-  override def versions[Id, V](id: Id)(implicit ir: IdResolver[Id, V, Category, K]): Source[OffsetDateTime, NotUsed] =
-    store.versions(id)
+  protected[cache] def maybePurgeCachedVersions(toRemove: Int = 500, pRemoval: Double = 0.05)(f: () => Boolean): Unit =
+    while (f()) {
+      // randomly GC the versions
+      val counter = new AtomicInteger
```
jasongilanfarr: whats this for?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +169,61 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+      versionedValueCache.retain { (k, v) =>
+        val x = scala.util.Random.nextDouble()
```
jasongilanfarr: Don't really understand your logic here.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +169,61 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+      }
+    }
+
+  protected def updateCachedVersions[V](
+    storageId: K,
+    version: OffsetDateTime,
+    category: Category,
+    v: V): Unit = {
+
+    maybePurgeCachedVersions() { () => versionedValueCache.size > 10000 }
```
jasongilanfarr: make a constant.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +169,61 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    versionedValueCache.put((storageId, version), Some(v))
+    val cached = versionCache.getOrElse((category, storageId), Nil) // linter:ignore UndesirableTypeInference
+    versionCache.put((category, storageId), (version +: cached).distinct)
+  }
+
+  override def versions[Id, V](id: Id)(implicit ir: IdResolver[Id, V, Category, K]): Source[OffsetDateTime, NotUsed] = {
+    val category = ir.category
+    val storageId = ir.toStorageId(id, None)
+    val versionsFuture = lockManager.executeSequentially(category.toString) {
+      lockManager.executeSequentially(storageId.toString) {
```
jasongilanfarr: (storageId, version).toString?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: let's not make that flexible, pass the constant in the constructor and put the default into StorageConfig and/or the constructor (but definitely storageconfig)

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: clarity

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: not sure why I did this. copy/paste?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: wasn't clear to me if any operations (like retain) were parallelized. if that's never going to be the case then I'm happy to change this to a non-thread-safe counter.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: as discussed via slack: detecting "at max capacity" and simply removing 1 element at random is not very efficient because all subsequent writes will also be at max capacity. better to remove (at random) some percentage of existing entries.

jdef:
Additionally locking on (storageId, version) seems redundant. In most
places there's already a lock on "storageId". Why introduce a 3rd level of
locking for (storageId, version) when the existing lock is sufficient
(albeit, more coarse grained than it needs to be)?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: I don't see anything that adds clarity - the annotation is unnecessary - you can use tooling to provide a visual type annotation in this case.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Hmm, this doesn't match what I we discussed at all.

jasongilanfarr:
Umm, versioning should be relatively unrelated to the cases where we need to lock on storageId (that's protecting a different cache). The intention of these locks is to protect the key id - which for this case is (storageId, version)

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -28,13 +29,16 @@ import scala.concurrent.{ ExecutionContext, Future }
...
-    val store: BasePersistenceStore[K, Category, Serialized])(implicit
+    val store: BasePersistenceStore[K, Category, Serialized],
+    val versionedValueCacheConfig: LazyCachingPersistenceStore.VersionedValueCacheConfig)(implicit
   mat: Materializer,
     ctx: ExecutionContext) extends PersistenceStore[K, Category, Serialized] with StrictLogging {
 
   private val lockManager = LockManager.create()
   private[store] val idCache = TrieMap.empty[Category, Seq[Any]]
   private[store] val valueCache = TrieMap.empty[K, Option[Any]]
+  private[store] val versionCache = TrieMap.empty[(Category, K), Seq[Any]]
```
jasongilanfarr: `idCache` is "Category -> MarathonId", `valueCache` is "MarathonId -> Option[Some Value]".

`versionCache` is "Category, MarathonId" -> what, a list of versions? If that's the case, it should be `Seq[OffsetDateTime]` as the valuetype.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Shouldn't this also put the version into the list of versions in the versionCache?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Please do it

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Still don't like this logic. Its really strange and counter intuitive - and you still didn't address my earlier comment that "f" shouldn't be passed in in the first place. (You can test this stuff by mocking or wrapping InMemoryPersistenceStore). I still think the atomic is 100% unnecessary and can be refactored out differently. If you're really considered about the "remove 1 element" then remove X% of elements as configured. I'm hoping there is a more efficient way to do so.

In addition, this should ultimately require a lock on the entire collection, thus all reads and writes for versioning should go through this lock - and not through the other locks in the non-versioned places.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -140,22 +170,94 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+  )
+
+  object VersionedValueCacheConfig {
+    val Default = VersionedValueCacheConfig()
+  }
+
+  /**
+    * max number of entries allowed in versionedValueCache before entries are purged
+    */
+  val MAX_VERSIONED_VALUE_CACHE_SIZE = 10000
```
jasongilanfarr: MaxVersionedValueCacheSize - Please use scala style constants, not java style. This is also completely consistent with StorageConfig and all of this stuff should be _there_ not here.

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -32,4 +33,25 @@ trait StorageConf extends ZookeeperConf {
     hidden = true,
     descr = "Max outstanding requests to Zookeeper persistence"
   )
+
+  lazy val cacheMaxVersionedValueCacheSize = opt[Int](
```
jasongilanfarr: all of this should be internal only.

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -32,4 +33,25 @@ trait StorageConf extends ZookeeperConf {
...
+    "cache_max_versioned_value_cache_size",
+    default = Some(LazyCachingPersistenceStore.VersionedValueCacheConfig.Default.maxEntries),
+    hidden = true,
+    descr = "Max number of versioned objects that are lazy cached per-repository"
+  )
+
+  lazy val cachePurgeCountVersionedValuesPerCycle = opt[Int](
```
jasongilanfarr: This option is confusing and super internal

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -32,4 +33,25 @@ trait StorageConf extends ZookeeperConf {
...
+    "cache_purge_count_versioned_values_per_cycle",
+    default = Some(LazyCachingPersistenceStore.VersionedValueCacheConfig.Default.purgeCount),
+    hidden = true,
+    descr = "Upon exceeding cache_max_versioned_value_cache_size, purge this number of objects from the lazy cache"
+  )
+
+  lazy val cachePRemoveFromVersionedValuesCache = opt[Double](
+    "cache_premove_versioned_value",
```
jasongilanfarr: ditto the above.

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -186,14 +186,18 @@ object CacheType {
 sealed trait PersistenceStorageConfig[K, C, S] extends StorageConfig {
```
jasongilanfarr: The caching stuff should be configured like the other stuff in here.

src/test/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStoreTest.scala
```diff
@@ -31,9 +36,139 @@ class LazyCachingPersistenceStoreTest extends AkkaUnitTest
...
+    implicit
+    ir: IdResolver[String, TestClass1, C, K],
+    m: Marshaller[TestClass1, Serialized],
+    um: Unmarshaller[Serialized, TestClass1]): Unit = {
+
+    name should {
+      "purge the cache appropriately" in {
+        implicit val clock = new SettableClock()
+        val store = newStore
+        for (i <- 1 to 100) {
```
jasongilanfarr: no for loops in storage at all, In general please try to remove them from your scala practices. Please use `1.to(100).foreach()`

jasongilanfarr:
Finally, please isolate the version caching from the non-versioned caching. Perhaps this should be a mix in that can be added to the base caching classes.

jdef:
OK, will address

aquamatthias:
@jdef Please rebase

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
 
   override def toString: String = s"LazyCachingPersistenceStore($store)"
 }
+
+object LazyCachingPersistenceStore {
```
jasongilanfarr: could just make it a case class.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+
+  def apply[K, Category, Serialized](
+    store: BasePersistenceStore[K, Category, Serialized])(implicit
+    mat: Materializer,
+    ctx: ExecutionContext) = new LazyCachingPersistenceStore(store)
+}
+
+class LazyVersionCachingPersistentStore[K, Category, Serialized](
+    val store: LazyCachingPersistenceStore[K, Category, Serialized],
```
jasongilanfarr: why not support any arbitrary persistence store?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+            async { // linter:ignore UnnecessaryElseBranch
+              versionedValueCache.retain { case ((sid, version), v) => sid != storageId }
+              versionCache.remove((category, storageId))
+              Done
+            }
+          }
+        }
+      }
+
+    cacheDeleteOp.zip(versionDeleteOp).flatMap(_ => Future.successful(Done))
```
jasongilanfarr: why not use async/await?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+
+  override def get[Id, V](id: Id, version: OffsetDateTime)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    um: Unmarshaller[Serialized, V]): Future[Option[V]] = {
+
+    withVersionCache(id) { (category, storageId) =>
+      withVersionedValueCache {
+        val cached = versionedValueCache.get((storageId, version)) // linter:ignore OptionOfOption
+        cached match {
+          case Some(v: Option[V] @unchecked) =>
```
jasongilanfarr: just Some(v)

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+      }
+    }
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def store[Id, V](id: Id, v: V)(implicit
+    ir: IdResolver[Id, V, Category, K],
+    m: Marshaller[V, Serialized]): Future[Done] = {
+
+    if (!ir.hasVersions)
```
jasongilanfarr: if not a single line if/else (inclusive), use brackets please.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    maybePurgeCachedVersions()
+    versionedValueCache.put((storageId, version), Some(v))
+    val cached = versionCache.getOrElse((category, storageId), Nil) // linter:ignore UndesirableTypeInference
+    versionCache.put((category, storageId), (version +: cached).distinct)
+  }
+}
+
+object LazyVersionCachingPersistentStore {
+
+  case class Config(
```
jasongilanfarr: This belongs in StorageConfig

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    lockManager.executeSequentially(s"versionedValueCache::${ir.category}")(future)
+
+  final def withVersionCache[Id, V, T](
+    id: Id)(
+    future: (Category, K) => Future[T])(implicit ir: IdResolver[Id, V, Category, K]): Future[T] = {
+    val category = ir.category
+    val storageId = ir.toStorageId(id, None)
+    lockManager.executeSequentially((category, storageId).toString())(future(category, storageId))
+  }
+
```
jasongilanfarr: Why isn't update and what not up here like the these helpers?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +161,224 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    store.setStorageVersion(storageVersion)
+
+  protected[cache] def maybePurgeCachedVersions(
+    maxEntries: Int = config.maxEntries,
+    purgeCount: Int = config.purgeCount,
+    pRemove: Double = config.pRemove): Unit =
+
+    while (versionedValueCache.size > maxEntries) {
+      // randomly GC the versions
+      val counter = new AtomicInteger
```
jasongilanfarr: This doesn't need to be atomic, its totally local to the function and completely unshared.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,191 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
 
   override def toString: String = s"LazyCachingPersistenceStore($store)"
 }
+
+case class LazyVersionCachingPersistentStore[K, Category, Serialized](
+    store: PersistenceStore[K, Category, Serialized],
+    config: PersistenceStorageConfig.VersionCacheConfig = PersistenceStorageConfig.VersionCacheConfig.Default)(implicit
```
jasongilanfarr: mostly a nit, but I generally prefer passing the named values in at this level (instead of passing the config) - doesn't really matter.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -59,12 +61,12 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
   }
 
   @SuppressWarnings(Array("all")) // async/await
-  private def deleteCurrentOrAll[Id, V](
+  protected[impl] def deleteCurrentOrAll[Id, V](
```
jdef: this can probably be private again

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,191 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    v: V): Unit = {
+
+    maybePurgeCachedVersions()
+    versionedValueCache.put((storageId, version), Some(v))
+    val cached = versionCache.getOrElse((category, storageId), Nil) // linter:ignore UndesirableTypeInference
+    versionCache.put((category, storageId), (version +: cached).distinct)
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  protected def deleteCurrentOrAll[Id, V](
```
jdef: this can probably be private

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,191 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+  }
+
+  protected[cache] def maybePurgeCachedVersions(
+    maxEntries: Int = config.maxEntries,
+    purgeCount: Int = config.purgeCount,
+    pRemove: Double = config.pRemove): Unit =
+
+    while (versionedValueCache.size > maxEntries) {
+      // randomly GC the versions
+      var counter: Int = 0
```
jasongilanfarr: nit: remove type annotation.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,191 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    versionedValueCache.put((storageId, version), Some(v))
+    val cached = versionCache.getOrElse((category, storageId), Nil) // linter:ignore UndesirableTypeInference
+    versionCache.put((category, storageId), (version +: cached).distinct)
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  protected def deleteCurrentOrAll[Id, V](
+    id: Id, delete: () => Future[Done])(implicit ir: IdResolver[Id, V, Category, K]): Future[Done] = {
+
+    val storeDeleteOp: Future[Done] = delete()
```
jasongilanfarr: Seems like you can just do this inline no? Also, no type annotation

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: eek

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -32,4 +32,13 @@ trait StorageConf extends ZookeeperConf {
     hidden = true,
     descr = "Max outstanding requests to Zookeeper persistence"
   )
+
+  lazy val versionCacheEnabled = toggle(
+    "version_caching",
```
jasongilanfarr: internal_

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -186,19 +186,52 @@ object CacheType {
...
-      case LazyCaching => new LazyCachingPersistenceStore[K, C, S](leafStore)
+      case LazyCaching => lazyStore
       case EagerCaching => new LoadTimeCachingPersistenceStore[K, C, S](leafStore)
     }
   }
 }
 
+object PersistenceStorageConfig {
+
+  case class VersionCacheConfig(
```
jasongilanfarr: Make this top level

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,191 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
 
   override def toString: String = s"LazyCachingPersistenceStore($store)"
 }
+
+case class LazyVersionCachingPersistentStore[K, Category, Serialized](
```
jdef: pull this into its own file?

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -186,19 +186,52 @@ object CacheType {
...
+    maxEntries: Int,
+    purgeCount: Int,
+    pRemove: Double
+  )
+
+  object VersionCacheConfig {
+    /**
+      * max number of entries allowed in the versioned value cache before entries are purged
+      */
+    protected val maxVersionedCacheSize = 10000
```
jasongilanfarr: MaxVersionedCacheSize

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -186,19 +186,52 @@ object CacheType {
...
+    /**
+      * probability that, during a purge of the versioned value cache, a given entry will be removed
+      */
+    protected val pRemoveFromVersionedCache = 0.05
```
jasongilanfarr: ProbabilityToRemoveFromCache

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: don't care ;)

jdef:
@aquamatthias PTAL

jdef:
rebased

aquamatthias:
retest this please

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,190 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    lockManager.executeSequentially(s"versionedValueCache::${ir.category}")(future)
+
+  def withVersionCache[Id, V, T](
+    id: Id)(future: (Category, K) => Future[T])(implicit ir: IdResolver[Id, V, Category, K]): Future[T] = {
+    val category = ir.category
+    val storageId = ir.toStorageId(id, None)
+    lockManager.executeSequentially((category, storageId).toString())(future(category, storageId))
+  }
+
+  protected[cache] def maybePurgeCachedVersions(
```
aquamatthias: Just curious: would be something like an LRU cache more appropriate?
(Looks like we implement a cache with random delete to meet the maxEntries constraint).

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,190 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+    while (versionedValueCache.size > maxEntries) {
+      // randomly GC the versions
+      var counter = 0
+      versionedValueCache.retain { (k, v) =>
+        val x = Random.nextDouble()
+        x > pRemove || { counter += 1; counter > purgeCount }
+      }
+    }
+
+  protected def updateCachedVersions[V](
```
aquamatthias: This method is called only from inside lock protected code paths (correct).
Can we make this `private[this]` and add a note about this requirement?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,190 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+          val children = await(store.versions(id).toMat(Sink.seq)(Keep.right).run())
+          versionCache((category, storageId)) = children
+          children
+        }
+      }
+    }
+    Source.fromFuture(versionsFuture).mapConcat(identity)
+  }
+
+  override def ids[Id, V]()(implicit ir: IdResolver[Id, V, Category, K]): Source[Id, NotUsed] = store.ids()
```
aquamatthias: Why are ids not cached?

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -32,4 +32,13 @@ trait StorageConf extends ZookeeperConf {
     hidden = true,
     descr = "Max outstanding requests to Zookeeper persistence"
   )
+
+  lazy val versionCacheEnabled = toggle(
+    "internal_version_caching",
```
aquamatthias: I would opt for making this a standard parameter (not internal).

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,190 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+
+  protected[cache] def maybePurgeCachedVersions(
+    maxEntries: Int = config.maxEntries,
+    purgeCount: Int = config.purgeCount,
+    pRemove: Double = config.pRemove): Unit =
+
+    while (versionedValueCache.size > maxEntries) {
+      // randomly GC the versions
+      var counter = 0
+      versionedValueCache.retain { (k, v) =>
```
aquamatthias: `retain` is called for every element in the cache - no matter if the purgeCount is already met.
It should be easy to randomize the elements to delete, without walking through the list.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
```diff
@@ -159,3 +162,190 @@ class LazyCachingPersistenceStore[K, Category, Serialized](
...
+  protected def updateCachedVersions[V](
+    storageId: K,
+    version: OffsetDateTime,
+    category: Category,
+    v: V): Unit = {
+
+    maybePurgeCachedVersions()
+    versionedValueCache.put((storageId, version), Some(v))
+    val cached = versionCache.getOrElse((category, storageId), Nil) // linter:ignore UndesirableTypeInference
+    versionCache.put((category, storageId), (version +: cached).distinct)
```
aquamatthias: Do we need distinct?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: sure

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: we could. @jasongilanfarr suggest a random-purge approach so that's what I implemented. happy to implement LRU if we think it's going to work out better.

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
jdef: @jasongilanfarr what was the argument for making this `internal_`?

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jasongilanfarr: Struggling to find the exact reference, but, for example, ARM uses random eviction on its CPU caches. Random beats LRU in the worst case scenarios and is generally used as a reference/baseline for testing purge strategies.

jdef:
@jasongilanfarr @aquamatthias I'd love to land this. Can you guys come to a final decision on the flag name?

jasongilanfarr:
I'd prefer internal for now- since we then can silently drop support for it
later.

src/main/scala/mesosphere/marathon/core/storage/store/impl/cache/LazyCachingPersistenceStore.scala
jdef: Honestly I'm not very concerned about the cost of iterating over a 10k item collection. The approach you suggest seems to complicate the impl and may be a case of premature optimization

jdef:
rebased

jdef:
retest this please

jdef:
retest this please

jdef:
are these integration tests stable yet?
* mesosphere.marathon.integration.setup.MesosAppIntegrationTest.deploy a simple pod
* mesosphere.marathon.integration.setup.MesosAppIntegrationTest.deploy a simple pod with health checks

nfnt:
@jdef: They should be stable. Looking at the logs, the pods are deployed and the tests fail with
```
[2016-10-17 21:25:43,919] INFO [] marathon_42349: [2016-10-17 21:25:43,919] ERROR [] Exception while processing request
[2016-10-17 21:25:43,919] INFO [] marathon_42349: java.lang.ClassCastException: scala.Some cannot be cast to mesosphere.marathon.core.pod.PodDefinition
```
when trying to get the pod's status (see https://github.com/mesosphere/marathon/blob/13e6b8769cc9ad02e44347d30f3fed46ad6c31f5/src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala#L197-L197). So it looks like these tests are failing because of some changes of this PR.

jdef:
squashed and rebased

jdef:
retest this please

jdef:
test flake #4212

PR #4522, opened 2016-09-29
Shut down Marathon when connection to ZK fails.

jeschkies:
Summary:
* On `CuratorElectionSerivce.offerLeadershipImpl` we catch NonFatals to
  shutdown Marathon.
* This changes also proposes lightweight dependency injection via
  implicits.

Test Plan: velocity

Reviewers: jdef, jasongilanfarr

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D34


jeschkies:
Already landed.

PR #4526, opened 2016-09-30
Ad/1.3.1

zen-dog:
Fixed some compile errors as well as added one missing dependency (scala-logging).

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -6,17 +6,17 @@ import com.wix.accord._
 import com.wix.accord.combinators.GeneralPurposeCombinators
 import com.wix.accord.dsl._
 import mesosphere.marathon.Protos.Constraint
-import mesosphere.marathon.core.health.MesosCommandHealthCheck
+import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
 import mesosphere.marathon.state.Container.{ Docker, MesosAppC, MesosDocker }
 // scalastyle:off
 import mesosphere.marathon.api.serialization.{ ContainerSerializer, EnvVarRefSerializer, PortDefinitionSerializer, ResidencySerializer, SecretsSerializer }
 // scalastyle:on
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.core.externalvolume.ExternalVolumes
-import mesosphere.marathon.core.health.{ HealthCheck, MarathonHealthCheck, MesosHealthCheck }
 import mesosphere.marathon.core.plugin.PluginManager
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -373,66 +373,47 @@ case class AppDefinition(
...
-              Some(hostPort)
-            }
-
           val effectivePort =
             if (ipAddress.isDefined || portMapping.hostPort.isEmpty) {
               portMapping.containerPort
             } else {
-              hostPort.get
+              val hostPort = hostPorts.head
+              hostPorts = hostPorts.drop(1)
```
jeschkies: hostPorts.tail should work fine.

src/test/scala/mesosphere/marathon/state/AppRepositoryTest.scala
```diff
@@ -78,7 +78,7 @@ class AppRepositoryTest extends MarathonSpec {
 
     val future = Future.successful(Seq("app1", "app2") ++ allApps.map(x => s"${x.id}:${x.version}"))
 
-    when(store.names()).thenReturn(future)
+    when[Future[Seq[String]]](store.names()).thenReturn(future)
```
jeschkies: Hm, I guess it would work without if `val future` would have the type declaration. So `val future: Future[Seq[String]]`. I might be wrong though.

zen-dog: No, it's not enough. You'll still get the same ambiguity error: 

```
Error:(138, 25) overloaded method value thenReturn with alternatives:
  (x$1: scala.concurrent.Future[scala.collection.immutable.Seq[String]],x$2: scala.concurrent.Future[scala.collection.immutable.Seq[String]]*)org.mockito.stubbing.OngoingStubbing[scala.concurrent.Future[scala.collection.immutable.Seq[String]]] <and>
  (x$1: scala.concurrent.Future[scala.collection.immutable.Seq[String]])org.mockito.stubbing.OngoingStubbing[scala.concurrent.Future[scala.collection.immutable.Seq[String]]]
 cannot be applied to (scala.concurrent.Future[scala.collection.Seq[String]])
    when(store.names()).thenReturn(future)
```

unterstein:
Should https://github.com/mesosphere/marathon/pull/4526/commits/ca527cf72a1914df0e052883e9455abce44e0487 be two commits? One with the actual changes and one with the compile error fixes?

Does we have the same issues on master?

zen-dog:
No, we don't have this issue on master.

aquamatthias:
retest this please

aquamatthias:
@zen-dog Thanks
Going forward we should try to omit such kind of PRs.

zen-dog:
retest this please

PR #4414, opened 2016-09-19
Adjust the KillService to handle instances instead of tasks

meichstedt:
Resolves #4394

Note the renamings:
TaskKillProgressActor -> InstanceKillProgressActor
TaskKillServiceActor -> KillServiceActor
TaskKillServiceDelegate -> KillServiceDelegate

unfortunately GH doesn't recognize them although edit/rename actions are in separate commits.

PR #4527, opened 2016-09-30
Fixes for TaskKillService

meichstedt:
**Please don't squash merge**

Cherry picking 2 commits from `feature/pods` to master so they can be cherry-picked onto 1.3 afterwards:

### Introducing UnknownTaskTerminated event
Until now, no event was dispatched on the event stream when the related task was terminal and unknown, thus making the TaskKillService unable to track the progress of these kill requests. Promises created by TaskKillProgressActors were never completed, leading to a memory leak. With this change, an UnknownTaskTerminated event will be dispatched on the event bus when a terminal status update for an unknown task is received. This event is also visible for subscribers to the event stream. The TaskKillService handles this event and correctly recognizes it as a kill success.

Cherry-picked from feature/pods (was: Adjust the KillService to handle instances instead of tasks (#4414))

### Fix KillService behavior: retry forever and don't spawn progress actors for unknown tasks

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
@@ -64,8 +64,8 @@ private[impl] class TaskKillServiceActor(
   }
 
   override def receive: Receive = {
-    case KillUnknownTaskById(taskId, promise) =>
-      killUnknownTaskById(taskId, promise)
+    case KillUnknownTaskById(taskId) =>
+      killUnknownTaskById(taskId)
 
     case KillTasks(tasks, promise) =>
       killTasks(tasks, promise)
@@ -80,9 +80,8 @@ private[impl] class TaskKillServiceActor(
       retry()
   }
 
-  def killUnknownTaskById(taskId: Task.Id, promise: Promise[Done]): Unit = {
+  def killUnknownTaskById(taskId: Task.Id): Unit = {
     log.debug("Received KillUnknownTaskById({})", taskId)
-    setupProgressActor(Seq(taskId), promise)
     tasksToKill.update(taskId, ToKill(taskId, maybeTask = None, attempts = 0))
     processKills()
   }
@@ -150,10 +149,6 @@ private[impl] class TaskKillServiceActor(
     val now = clock.now()
 
     inFlight.foreach {
-      case (taskId, toKill) if toKill.attempts >= config.killRetryMax =>
-        log.warning("Expunging {} from state: max retries reached", taskId)
-        stateOpProcessor.process(TaskStateOp.ForceExpunge(taskId))
-
       case (taskId, toKill) if (toKill.issued + config.killRetryTimeout) < now =>
         log.warning("No kill ack received for {}, retrying", taskId)
         processKill(toKill)
@@ -172,7 +167,7 @@ private[termination] object TaskKillServiceActor {
 
   sealed trait Request extends InternalRequest
   case class KillTasks(tasks: Iterable[Task], promise: Promise[Done]) extends Request
-  case class KillUnknownTaskById(taskId: Task.Id, promise: Promise[Done]) extends Request
+  case class KillUnknownTaskById(taskId: Task.Id) extends Request
 
   sealed trait InternalRequest
   case object Retry extends InternalRequest
--- a/src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
@@ -8,6 +8,7 @@ import mesosphere.marathon.core.event.MesosStatusUpdateEvent
 import mesosphere.marathon.core.task.termination.TaskKillConfig
 import mesosphere.marathon.core.task.tracker.TaskStateOpProcessor
 import mesosphere.marathon.core.task.{ Task, TaskStateOp }
+import mesosphere.marathon.core.event.UnknownTaskTerminated
 import mesosphere.marathon.state.Timestamp
 
 import scala.collection.mutable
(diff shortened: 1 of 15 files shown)


src/main/scala/mesosphere/marathon/core/task/termination/impl/TaskKillServiceActor.scala
```diff
@@ -114,21 +116,25 @@ private[impl] class TaskKillServiceActor(
...
+      task.isGone || task.isUnknown || task.isDropped || task.isUnreachable
+    }
 
     if (taskIsLost) {
       log.warning("Expunging lost {} from state because it should be killed", taskId)
       // we will eventually be notified of a taskStatusUpdate after the task has been expunged
       stateOpProcessor.process(TaskStateOp.ForceExpunge(taskId))
     } else {
-      val knownOrNot = if (maybeTask.isDefined) "known" else "unknown"
+      val knownOrNot = if (toKill.maybeTask.isDefined) "known" else "unknown"
```
jeschkies: `toKill.fold("unknown")(task => "known")` is an alternative :)

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -72,18 +82,28 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
+    MarathonTaskStatus.Gone,
+    MarathonTaskStatus.Dropped,
+    MarathonTaskStatus.Unknown
   )
   // It doesn't make sense to kill an unknown task if it is in a terminal or killing state
   // We'd only get another update for the same task
-  private def killWhenUnknown(status: MesosProtos.TaskStatus): Boolean = {
-    !ignoreWhenUnknown.contains(status.getState)
+  private def killWhenUnknown(status: MarathonTaskStatus): Boolean = {
+    !ignoreWhenUnknown.contains(status)
```
jeschkies: Call it terminal state then

```scala
val terminalStates = Set (...)

killWhenNonTerminal(status: MarathonTaskStatus): Boolean = !terminalStates.contain(status)
```

jeschkies: Ah, the unknown is an unknown task. Hm, then I'd say `killWhenUnknownTask`. However, see above.

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -43,13 +47,19 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
         val taskStateOp = TaskStateOp.MesosUpdate(task, status, now)
         stateOpProcessor.process(taskStateOp).flatMap(_ => acknowledge(status))
 
-      case None if killWhenUnknown(status) =>
+      case None if terminalUnknown(marathonTaskStatus) =>
+        log.warn("Received terminal status update for unknown {}", taskId)
+        eventStream.publish(UnknownTaskTerminated(taskId, taskId.runSpecId, marathonTaskStatus))
+        acknowledge(status)
+
+      case None if killWhenUnknown(marathonTaskStatus) =>
```
jeschkies: I'd just say `case None if nonTerminalState(marathonTaskStatus)`.

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -72,18 +82,28 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
   lazy val name = Names.named(getClass.getSimpleName)
 
+  /** Matches all states that are considered terminal for an unknown task */
+  private[impl] def terminalUnknown(status: MarathonTaskStatus): Boolean = status match {
+    case t: MarathonTaskStatus.Terminal => true
+    case MarathonTaskStatus.Unreachable => true
+    case _ => false
+  }
+
   private[this] val ignoreWhenUnknown = Set(
```
jeschkies: I'd actually move this as `terminalStates` to `mesosphere.marathon.core.task.state.MarathonTaskStatus` since this will be a common thing and should be defined in one position.

aquamatthias:
retest this please

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
unterstein: another possible naming could be `killWhenTaskUnknown` ;)

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
meichstedt: Which wouldn't be 100% correct – `Unreachable` is a non-terminal state, but killing a task that's unreachable will just lead to another status update that says it's `Unreachable`. There's a reason for being so explicit, but it's obviously not clear, so I'll document that.

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
meichstedt: I didn't address these comments here so we can proceed with cherry-picking into 1.3. I agree we should clarify naming and use-case; let's do that in a separate PR.

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
jeschkies: Hm, why would killing an `Unreachable` task trigger a status update? I thought Mesos would not do much with it.

meichstedt: I am not sure how this behavior changes with Mesos 1.1 or future versions; until now, when killing a task that was scheduled on an agent that is currently not registered, Mesos sent a statusUpdate declaring the task is LOST (which makes sense because the master cannot forward the request to the agent).

PR #4530, opened 2016-09-30
Fixes for TaskKillService

meichstedt:
**Please don't squash merge**

Cherry picking 2 commits from `master` to `releases/1.3`:

### Introducing UnknownTaskTerminated event
Until now, no event was dispatched on the event stream when the related task was terminal and unknown, thus making the TaskKillService unable to track the progress of these kill requests. Promises created by TaskKillProgressActors were never completed, leading to a memory leak. With this change, an UnknownTaskTerminated event will be dispatched on the event bus when a terminal status update for an unknown task is received. This event is also visible for subscribers to the event stream. The TaskKillService handles this event and correctly recognizes it as a kill success.

Cherry-picked from feature/pods (was: Adjust the KillService to handle instances instead of tasks (#4414))

### Fix KillService behavior: retry forever and don't spawn progress actors for unknown tasks


aquamatthias:
Thanks!

PR #4531, opened 2016-09-30
Terminal tasks should not be send in reconciliation

unterstein:
A task group in Mesos will terminate when either all tasks are terminal or at least one task terminated abnormally (exit code != 0).
E.g., a valid state of a pod in Marathon is
> task1: Running
> task2: Finished

In this a case, the overall status of the instance is `Running`. If Marathon performs an explicit reconciliation for task2 in this case, Mesos will eventually send a status update that indicates that task2 is `Gone`, as it no longer exists.

On receiving this update, Marathon will interpret the overall instance status to be `Gone`, assuming the whole pod is not existing. A workaround is to **not** explicitly reconcile terminal tasks.

An improvement in the future would be to disallow changing a terminal task status to some other status (a task that is `Finished` should never become `Gone`).

### DIFF
--- a/src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
+++ b/src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
@@ -3,18 +3,18 @@ package mesosphere.marathon
 import java.util.concurrent.TimeoutException
 
 import akka.Done
-import akka.actor.{ ActorRef, Props }
+import akka.actor.{ActorRef, Props}
 import akka.event.EventStream
 import akka.stream.scaladsl.Source
 import akka.testkit._
 import akka.util.Timeout
 import mesosphere.marathon.MarathonSchedulerActor._
 import mesosphere.marathon.core.instance.TestInstanceBuilder
-import mesosphere.marathon.core.election.{ ElectionService, LocalLeadershipEvent }
+import mesosphere.marathon.core.election.{ElectionService, LocalLeadershipEvent}
 import mesosphere.marathon.core.event._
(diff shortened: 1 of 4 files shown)


src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
```diff
@@ -84,6 +86,33 @@ class MarathonSchedulerActorTest extends MarathonActorSupport
     }
   }
 
+  test("Terminal tasks should not be submitted in reconciliation") {
+    val f = new Fixture
+    import f._
+    val app = AppDefinition(id = "/test-app".toPath, instances = 1)
+    val instance = TestInstanceBuilder.newBuilder(app.id).addTaskUnreachable(containerName = Some("unreachable")).addTaskRunning().addTaskGone(containerName = Some("gone")).getInstance()
```
meichstedt: Can you please add _all_ available `Terminal` status to the instance and verify explicitly that none of these is sent for reconciliation? I'd like to see
> Created
> Error
> Failed
> Finished
> Killed
> Killing
> Running
> Staging
> Starting
> Unreachable
> Gone
> Dropped
> Unknown

and a verify that status is only sent for

> Created
> Killing
> Running
> Staging
> Starting
> Unreachable

Btw: It would be nice to extract the underlying functionality _which_ task status are sent to mesos. Out of scope for this PR though.

meichstedt:
@unterstein Can you address my comment in a new PR? I'd really like the test to be explicit and exhaustive.

unterstein:
Sure, created a ticket for this https://github.com/mesosphere/marathon/issues/4534

PR #4535, opened 2016-09-30
support image pulling credentials for docker images in pod containers

jdef:


jdef:
mesos API used by this PR is deprecated and should have been removed already. can re-open this later when per-image creds some up again.

PR #4537, opened 2016-09-30
when generating pod instance status xref the pod-spec by version

jdef:
xref #4521 (for performance)

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
+++ b/src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
@@ -174,9 +174,13 @@ class AppInfoBaseData(
     async { // linter:ignore UnnecessaryElseBranch
       val now = clock.now().toOffsetDateTime
       val instances = await(instancesByRunSpecFuture).specInstances(podDef.id)
-      val instanceStatus = await(Future.sequence(instances.map { inst =>
-        podInstanceStatus(podDef.id, inst)
-      }.toIndexedSeq)).flatten
+      val specByVersion: Map[Timestamp, Option[PodDefinition]] = await(Future.sequence(
+        // TODO(jdef) if repositories ever support a bulk-load interface, use it here
+        instances.map(_.runSpecVersion).toSeq.distinct.map { version =>
+          podRepository.getVersion(podDef.id, version.toOffsetDateTime).map(version -> _)
+        }
+      )).toMap
+      val instanceStatus = instances.map { inst => podInstanceStatus(inst)(specByVersion.apply) }.toIndexedSeq.flatten
       val statusSince = if (instances.isEmpty) now else instanceStatus.map(_.statusSince).max
       val state = await(podState(podDef.instances, instanceStatus, isPodTerminating(podDef.id)))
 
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -172,7 +174,9 @@ class AppInfoBaseData(
     async { // linter:ignore UnnecessaryElseBranch
       val now = clock.now().toOffsetDateTime
       val instances = await(instancesByRunSpecFuture).specInstances(podDef.id)
-      val instanceStatus = instances.map(instance => Raml.toRaml(podDef -> instance)).toVector
+      val instanceStatus = await(Future.sequence(instances.map { inst =>
+        podInstanceStatus(podDef.id, inst)
+      }.toVector)).flatten
```
jasongilanfarr: prefer `toIndexedSeq` over `toVector` if you need an indexed sequence

src/test/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseDataTest.scala
```diff
@@ -29,13 +29,15 @@ class AppInfoBaseDataTest extends MarathonSpec with GivenWhenThen with Mockito w
     lazy val healthCheckManager = mock[HealthCheckManager]
     lazy val marathonSchedulerService = mock[MarathonSchedulerService]
     lazy val taskFailureRepository = mock[TaskFailureRepository]
+    lazy val podRepository = mock[ReadOnlyPodRepository]
```
jasongilanfarr: You don't need to add any mocked methods?

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
jdef: ack

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -188,6 +192,19 @@ class AppInfoBaseData(
       )
     }
 
+  @SuppressWarnings(Array("all")) // async/await
+  def podInstanceStatus(podId: PathId, instance: Instance): Future[Option[PodInstanceStatus]] =
+    async { // linter:ignore UnnecessaryElseBranch
+      val maybePodSpec: Option[PodDefinition] = await(podRepository.getVersion(
```
aquamatthias: We know all instances (+versions) that we want to fetch. 
Perhaps it makes sense to create a separate:
```scala
lazy val podSpecByVersion: Future[Map[Timestamp, PodDefinition]]
```
which loads only the necessary versions in question?

jasongilanfarr: @jdef is correct. There is no way to bulk load in ZK; however, I do think the interface would be better as a bulk load interface so that new stores could implement a more efficient API.

jdef:
i've updated to address the concerns expressed by @aquamatthias

jdef:
Looks like failures are unrelated scapegoat issues

PR #4540, opened 2016-09-30
use mesos-1.1.0-SNAPSHOTS from Apache repo

jdef:
Not sure if we really want this, or a more specific, pinned version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -85,9 +85,10 @@ lazy val commonSettings = inConfig(IntegrationTest)(Defaults.testTasks) ++ Seq(
     "-encoding", "UTF-8", "-source", "1.8", "-target", "1.8", "-Xlint:unchecked", "-Xlint:deprecation"
   ),
   resolvers ++= Seq(
-    "Mesosphere Public Repo" at "http://downloads.mesosphere.com/maven",
     "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/",
-    "Spray Maven Repository" at "http://repo.spray.io/"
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -85,9 +85,10 @@ lazy val commonSettings = inConfig(IntegrationTest)(Defaults.testTasks) ++ Seq(
     "-encoding", "UTF-8", "-source", "1.8", "-target", "1.8", "-Xlint:unchecked", "-Xlint:deprecation"
   ),
   resolvers ++= Seq(
-    "Mesosphere Public Repo" at "http://downloads.mesosphere.com/maven",
     "Typesafe Releases" at "http://repo.typesafe.com/typesafe/releases/",
-    "Spray Maven Repository" at "http://repo.spray.io/"
+    "Spray Maven Repository" at "http://repo.spray.io/",
+    "Apache shapshot releases" at "https://repository.apache.org/content/repositories/snapshots/",
```
jasongilanfarr: Apache Snapshots

project/Dependencies.scala
```diff
@@ -74,7 +74,7 @@ object Dependency {
     val Chaos = "0.8.7"
     val Guava = "19.0"
     // FIXME (gkleiman): reenable deprecation checks after Mesos 1.0.0-rc2 deprecations are handled
```
jasongilanfarr: is this todo still valid?

jdef: I think so. We were referencing some deprecated proto fields for volumes until they were officially removed from the mesos API. I don't think we got to that point yet.

PR #4543, opened 2016-10-04
adds required mesos version for pods to dockerfile

kensipe:


jdef:
Change looks good. Wondering .. should we update all Dockerfile's?

jdef:
@nfnt does this overlap with a change that you're working on?

jdef:
xref #4481

aquamatthias:
Close in favor of #4481

PR #4546, opened 2016-10-04
incorporate per-container resources and link to parent pod-spec into pod instance status

jdef:
/cc @orlandohohmeier

### DIFF
--- a/docs/docs/rest-api/public/api/v2/types/podStatus.raml
+++ b/docs/docs/rest-api/public/api/v2/types/podStatus.raml
@@ -86,6 +86,10 @@ types:
       endpoints?:
         type: array
         items: ContainerEndpointStatus
+      resources?:
+        type: resources.Resources
+        description: |
+          Resources in use by this container.
       termination?: ContainerTerminationState
       lastUpdated:
         type: datetime
@@ -159,6 +163,10 @@ types:
       containers?:
         type: ContainerStatus[]
         description: status for each running container of this instance.
+      specReference?:
+        type: strings.Uri
+        description: |
+          Location of the version of the pod specification this instance was created from.
       lastUpdated:
         type: datetime
         description: |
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -32,6 +32,16 @@ trait PodStatusConversion {
     // the Mesos task ID itself
     val displayName = task.taskId.containerName.getOrElse(task.taskId.mesosTaskId.getValue)
 
+    val resources: Option[Resources] = {
+      import InstanceStatus._
+      task.status.taskStatus match {
+        case Staging | Starting | Running | Reserved | Unreachable | Killing =>
+          maybeContainerName.flatMap(pod.container(_)).map(_.resources)
```
aquamatthias: Use `pod.container(task).map(_.resources)`

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -80,6 +93,7 @@ trait PodStatusConversion {
       networks = networkStatus,
       containers = containerStatus,
       message = message,
+      specReference = Some(s"/v2/pods${pod.id.canonicalPath()}::versions/${instance.runSpecVersion.toOffsetDateTime}"),
```
aquamatthias: We should already have the canonicalPath here (no need to recompute).

docs/docs/rest-api/public/api/v2/types/podStatus.raml
```diff
@@ -159,6 +163,10 @@ types:
       containers?:
         type: ContainerStatus[]
         description: status for each running container of this instance.
+      specReference?:
```
aquamatthias: Like the idea of pointing to the uri.
Do we have a runSpecVersion on the instance level?

jdef: we don't report the runSpecVersion as an instance-status field. we could. if we do, would you like that in this PR or a follow-up PR?

PR #4549, opened 2016-10-04
Ken/pod functional test

kensipe:
shakedown tests which test for the full pod API.
there appear to be bugs for which jiras will be created
there is an issue with killing an instance


tests/system/test_pods.py
```diff
@@ -0,0 +1,235 @@
...
+from dcos import marathon, util
+from utils import fixture_dir, get_resource, wait_for_deployment
+
+PACKAGE_NAME = 'marathon'
+DCOS_SERVICE_URL = dcos_service_url(PACKAGE_NAME)
+WAIT_TIME_IN_SECS = 300
+
+
+def _pods_json(file="simple-pods.json"):
+    return get_resource("{}/{}".format(fixture_dir(),file))
```
colin-msphere: Any reason not to use `os.path.join(fixture_dir(), file)`?

kensipe: nope... I'll resolve that in the morning!  thx!

tests/system/test_network_part.py
```diff
@@ -113,17 +116,17 @@ def test_mom_with_network_failure_bounce_master():
...
-    json = get_json(__marathon_url("apps/sleep","marathon-user"))
+    json = get_json(__marathon_url("apps/sleep", "marathon-user"))
     original_sleep_task_id = json["app"]["tasks"][0]["id"]
 
     mom_task_ips = get_service_ips("marathon-user", "sleep")
     for task_ip in mom_task_ips:
         break
 
-    print ("\nTask IP: " + task_ip)
+    print("\nTask IP: " + task_ip)
```
colin-msphere: For another pull request, but this file has a mixture of `"some string" + var` and `"some string{}".format(var)`. It would be great to pick one style and use it throughout the code base.

tests/system/test_pods.py
```diff
@@ -0,0 +1,236 @@
...
+    client = marathon.create_client()
+    pod_id = "/pod-version"
+
+    pod_json = _pods_json()
+    pod_json["id"] = pod_id
+    pod_json["scaling"]["instances"] = 1
+    client.add_pod(pod_json)
+    wait_for_deployment(client)
+
+    pod_json["scaling"]["instances"] = 10
```
colin-msphere: Should there be a `time.sleep(1)` here as well?

kensipe: best to... resolved

PR #4551, opened 2016-10-05
remove ipAddress from RunSpec, introduce LegacyIpAddressSupport

jdef:
Addresses an outstanding TODO: removes `ipAddress` from `PodDefinition` and therefore `RunSpec`


jasongilanfarr:
I don't like this or the other change with "Legacy" - please remove them from runSpec and move them to AppDefinition only.

PR #4552, opened 2016-10-05
refactor healthChecks from RunSpec into LegacyHealthCheckSupport

jdef:


src/main/scala/mesosphere/marathon/state/RunSpec.scala
```diff
@@ -78,3 +76,9 @@ trait RunSpec extends plugin.RunSpec {
   val portNames = Seq.empty[String]
   val user: Option[String]
 }
+
+trait LegacyHealthCheckSupport {
+  _: RunSpec =>
```
jasongilanfarr: huh?

PR #4556, opened 2016-10-06
Increase task_operation_notification_timeout to 10sec

zen-dog:
- Increase the timeout to make a possible race condition that leads to an app overprovisioning less likely
- Add a comment in code to explain how the race condition may occur

Related: DCOS-10245


src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -375,17 +375,30 @@ private class TaskLauncherActor(
       // yet.
       taskOp.stateOp.possibleNewState.foreach { newState =>
         tasksMap += taskId -> newState
+        // In case we don't receive an update a TaskOpRejected message with TASK_OP_REJECTED_TIMEOUT_REASON
+        // reason is scheduled within config.taskOpNotificationTimeout milliseconds. This will trigger another
+        // attempt to launch the task.
+        //
+        // NOTE: this can lead to a race condition where the task would be launched twice:
```
zen-dog: Agree. I rephrased it to make the underlying problem clearer.

zen-dog:
test this please

meichstedt:
@zen-dog Can you please rebase against master?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -376,17 +376,33 @@ private class TaskLauncherActor(
...
+        // NOTE: this can lead to a race condition where an additional task is launched: in a nutshell if a TaskOp A
+        // is rejected due to an internal timeout, TaskLauncherActor will schedule another TaskOp.Launch B, because
+        // it's under the impression that A has not succeeded/got lost. If the timeout is triggered internally, the
+        // tasksToLaunch count will be increased by 1. The TaskOp A can still be accepted though, and if it is,
+        // two tasks (A and B) might be launched instead of one (given Marathon receives sufficient offers and is
+        // able to create another TaskOp).
+        // This would lead to the app being over-provisioned (e.g. "3 of 2 tasks" launched) but eventually converge
+        // to the target task count when tasks over capacity are killed. With a sufficiently high timeout this case
+        // should be fairly rare.
+        // A better solution would involve an overhaul of the way TaskLaunchActor works and might be
```
jdef: Is this something that we should be tracking as technical debt?

PR #4562, opened 2016-10-07
Prevent multiple ScaleRunSpec during instance gets terminal

unterstein:
When receiving multiple terminal task updated inside a pod for each task update an instance update will be triggered which leads to an ScaleRunSpec for this application. Therefore this patch makes sure that only an ScaleRunSpec is triggered if the instance reaches a terminal state.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
@@ -32,7 +32,10 @@ class ScaleAppUpdateStepImpl @Inject() (
         val state = update.status
         log.info(s"initiating a scale check for runSpec [$runSpecId] due to [$instanceId] $state")
         // TODO(PODS): we should rename the Message and make the SchedulerActor generic
-        schedulerActor ! ScaleRunSpec(runSpecId)
+        // only dispatch ScaleRunSpec if last state was not terminal and current new state is terminal
+        if (update.lastState.exists(!_.status.isTerminal) && update.status.isTerminal) {
+          schedulerActor ! ScaleRunSpec(runSpecId)
+        }
 
       case _ =>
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
```diff
@@ -32,7 +32,10 @@ class ScaleAppUpdateStepImpl @Inject() (
         val state = update.status
         log.info(s"initiating a scale check for runSpec [$runSpecId] due to [$instanceId] $state")
         // TODO(PODS): we should rename the Message and make the SchedulerActor generic
-        schedulerActor ! ScaleRunSpec(runSpecId)
+        // only dispatch ScaleRunSpec if last state was not terminal and current new state is terminal
+        if (update.lastState.exists(!_.status.isTerminal) && update.status.isTerminal) {
```
aquamatthias: exists will return false, if there is no lastState.
But if there is no lastState, this should be true.
Perhaps use `forall` instead of exists?

unterstein: correct.

aquamatthias:
There are shaky integration tests - one was successful so I will merge.

unterstein:
yes, result locally:
```
Run completed in 18 minutes, 51 seconds.
Total number of tests run: 116
Suites: completed 216, aborted 0
Tests: succeeded 116, failed 0, canceled 0, ignored 0, pending 0
All tests passed.
[info] Run completed in 18 minutes, 51 seconds.
[info] Total number of tests run: 116
[info] Suites: completed 216, aborted 0
[info] Tests: succeeded 116, failed 0, canceled 0, ignored 0, pending 0
[info] All tests passed.
[success] Total time: 1270 s, completed Oct 7, 2016 1:40:20 PM
```

PR #4563, opened 2016-10-07
Revert "when generating pod instance status xref the pod-spec by version"

jdef:
Reverts mesosphere/marathon#4537


jdef:
This should have been merged to master instead, so reverting here

PR #4564, opened 2016-10-07
when generating pod instance status xref the pod-spec by version

jdef:
same as PR #4537 but against master instead of feature/pods


aquamatthias:
Thanks.
org.apache.mesos#mesos;1.1.0-SNAPSHOT!mesos.jar could not be retrieved which happens the n.th time today :(

PR #4565, opened 2016-10-07
Fix health status in pod instances

meichstedt:
The logic to infer instance health changes was broken in that it only reported instances as healthy if **all** tasks were reported healthy. This is wrong when a task has no configured healthCheck and will never have a health status.

This patch changes this to
- an instance is healthy if
  - all tasks are running or finished AND
  - no task is reported unhealthy AND
  - at least one task is reported healthy
- an instance is unhealthy if
  - at least one task is running and reported unhealthy (this ignores unhealthy tasks which are **not** running)
- an instance has no health status if
  - not all tasks are running or finished OR
  - no task has health information

This should only act as a band aid until we provide information about the actual configured health checks. (An instance should not be reported healthy if a task that has a healthCheck configured has no health status information yet.)

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -14,6 +14,8 @@ import org.apache._
 import org.apache.mesos.Protos.Attribute
 import org.slf4j.{ Logger, LoggerFactory }
 import play.api.libs.json.{ Reads, Writes }
+
+import scala.annotation.tailrec
 // TODO: Remove timestamp format
 import mesosphere.marathon.api.v2.json.Formats.TimestampFormat
 import play.api.libs.json.{ Format, JsResult, JsString, JsValue, Json }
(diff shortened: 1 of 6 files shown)


jdef:
I wrote some logic that infers STABLE vs DEGRADED status for an instance. Looks similar to what you're doing here. I think the logic is in the PodStatusConversion class. Maybe you could factor that logic into a more reusable API?

aquamatthias:
Shaky Unit Tests:
```
mesosphere.marathon.MarathonSchedulerActorTest.Forced deployment
```
Created a ticket.


Integration tests fail:
```
mesosphere.marathon.integration.ResidentTaskIntegrationTest.persistent volume will be re-attached and keep state
mesosphere.marathon.integration.ResidentTaskIntegrationTest.Restart
mesosphere.marathon.integration.ResidentTaskIntegrationTest.Config Change
```
Which seems to be a problem of the new mesos version. 
Created a ticket.

aquamatthias:
Thanks!

PR #4568, opened 2016-10-07
Me/pods/kill strategy

meichstedt:
### Fix task state transition for terminal tasks
    
Any processed TaskUpdateOperation for a terminal task will now lead to a Noop. This is to prevent bugs when Marathon tries to kill a task that has already terminated. It will likely receive a TASK_LOST as a result, but the task is actually not lost but still e.g. Finished.

### Issue kill requests for all non-terminal tasks of an instance

When killing an instance, this patch issues a kill request for each non-terminal task of an instance. It is possible that one or more tasks have terminated already when the killService eventually retries because the instance is not terminal yet.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -269,18 +269,14 @@ object Instance {
       // E.g. if container A has a healthCheck and B doesn't, b.mesosStatus.hasHealthy will always be `false`,
       // but we don't know whether this is because no healthStatus is available yet, or because no HC is configured.
       // This is therefore simplified to `if there is no healthStatus with getHealthy == false, healthy is true`
-      log.info("1: foundHealthy = {}", foundHealthy)
       foundHealthy
     } else if (isRunningUnhealthy(tasks.head)) {
       // there is a running task that is unhealthy => the instance is considered unhealthy
-      log.info("2: foundRunningUnhealthy")
       Some(false)
     } else if (isPending(tasks.head)) {
       // there is a task that is NOT Running or Finished => None
-      log.info(s"2: foundPending: ${tasks.head.status.taskStatus}, hasHealthy = {}", tasks.head.status.mesosStatus.fold("")(_.hasHealthy.toString))
       None
     } else if (isRunningHealthy(tasks.head)) {
-      log.info("2: foundRunningHealthy")
       computeHealth(tasks.tail, Some(true))
     } else {
       computeHealth(tasks.tail, foundHealthy)
--- a/src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
@@ -13,6 +13,7 @@ import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.instance.update.InstanceUpdateOperation
 
 import scala.collection.mutable
+import scala.collection.immutable.Seq
 import scala.concurrent.Promise
 
 /**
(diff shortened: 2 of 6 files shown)


src/test/scala/mesosphere/marathon/core/instance/InstanceUpdateTest.scala
```diff
@@ -52,6 +49,16 @@ class InstanceHealthTest extends WordSpecLike
       instance = TaskStatusUpdateTestHelper.runningUnhealthy(instance, Some(f.container2)).updatedInstance
       instance.state.healthy.value shouldBe false
     }
+
+    "not transition to another state if another terminal TaskStatus update for an already terminal task is processed" in {
+      instance = TaskStatusUpdateTestHelper.running(instance, Some(f.container1)).updatedInstance
+      instance = TaskStatusUpdateTestHelper.finished(instance, Some(f.container2)).updatedInstance
+
+      TaskStatusUpdateTestHelper.unknown(instance, Some(f.container2)).effect shouldBe a[InstanceUpdateEffect.Noop]
```
aquamatthias: Think I said that before: I like your tests - so clean.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -269,18 +269,14 @@ object Instance {
       // E.g. if container A has a healthCheck and B doesn't, b.mesosStatus.hasHealthy will always be `false`,
       // but we don't know whether this is because no healthStatus is available yet, or because no HC is configured.
       // This is therefore simplified to `if there is no healthStatus with getHealthy == false, healthy is true`
-      log.info("1: foundHealthy = {}", foundHealthy)
```
aquamatthias: Thanks for removing those - should have marked that in the other PR.

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
```diff
@@ -128,25 +128,32 @@ private[impl] class KillServiceActor(
...
-    if (taskIsLost) {
-      log.warning("Expunging lost {} from state because it should be killed", taskId)
-      // we will eventually be notified of a taskStatusUpdate after the task has been expunged
+    // An instance will be expunged once all tasks are terminal. Therefore, this case is
+    // highly unlikely. Should it ever occur, this will still expunge the instance to clean up.
+    val allTerminal: Boolean = taskIds.isEmpty
+
+    if (isLost || allTerminal) {
+      val msg = if (isLost) "it is lost" else "all its tasks are terminal"
+      log.warning("Expunging {} from state because {}", instanceId, msg)
```
aquamatthias: Just a note: we should get rid of log statements with `{}`.

src/test/scala/mesosphere/marathon/core/instance/InstanceUpdateTest.scala
meichstedt: Thanks!

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
jasongilanfarr: We probably should use scala logging (StrictLogging) which implements log as a macro that does "isLevelEnabled..."

src/test/scala/mesosphere/marathon/core/instance/InstanceUpdateTest.scala
jasongilanfarr: :+1:

aquamatthias:
The test 
```mesosphere.marathon.core.election.impl.CuratorElectionServiceTest.The CuratorElectionService when given an unresolvable hostname should shut Marathon down on a NonFatal
```
is shaky. I already filed a jira for that.

meichstedt:
Cool – please don't squash to keep the history of the 2 separate commits

jdef:
lgtm

PR #4577, opened 2016-10-10
internal model should use state.Resources instead of raml.Resources

jdef:


jasongilanfarr:
What's the point of this PR? I have 0 objection to using RAML types internally _especially if they're identical_

jdef:
The point is to get off of using Raml types internally. We have a pods
ticket for this, so I picked it up. I **do** have a problem w/ using
marathon types internally.

MV and I also talked about how using the raml-to-json for a storage format
will probably bite us. standing in the way of that are the internal raml
types. unless we want to write raml-to-protobuf converters :( also a waste
since we shouldn't be using the raml types internally anyway

jasongilanfarr:
This warrants a larger discussion.

Proto bit us too- I had to fix a migration bug because somewhere the Proto was changed in an incompatible way. Json isn't going to change that (but I can do some cool stuff that actually CHECKS compatibility when you screw with Raml.)

I do NOT want to write that twice.

I object to this change until a wider discussion has been had.

jdef:
closing since there's no consensus here.

PR #4588, opened 2016-10-13
Updated doc building to use github_pages jekyll and fixes

solarkennedy:
While trying to update the docs I found that the build procedure wasn't 100%. This make the following changes:

* Uses the `github_pages` gem so that the version of jekyll that github uses is the same that we use to generate it
* Adjusted the docs readme to include installing a beta version of raml2html for RAML 1.0 support

I have some other doc changes I want to make, but I didn't want to mix them with these changes to make it easier to review.


aquamatthias:
Hey @solarkennedy Thanks for updating the gh_pages docs setup! 
Was on my list for quite some time - thanks again!

PR #4589, opened 2016-10-13
Added more docs for forcing. Closes #4581

### LINKED ISSUE #4581: Tasks managed by marathon still stick around, don't show up in the UI, but are reaped after a marathon restart
This is similar in spirit to #4236, but it happens at Yelp with individual tasks too.

The symptom is that Marathon still keeps old tasks running, even when the app that managed them no longer exists.

These tasks are _still_ in the ZK state! In fact this is how I've begun to get an early warning on this condition. Here is the psuedo-code:

```
#!/bin/bash
set -eu
CLUSTER=$1

cd /tmp/

zkdump -l /marathon/state/  | sed "s/', '/\n/g" |  sed "s/\['/\n/g" | sed "s/'\]/\n/g"  > state
grep 'app:' state | cut -f 2 -d : | sort -u> apps
grep 'task:' state  | cut -f 2 -d : | sort > tasks

rogue_task_count=0

for task in `cat tasks`; do
  app_name=$(echo $task | cut -f 1-4 -d '.')
  if ! grep -q $app_name apps; then
    echo Found a rogue task with no app associated with it: $task
    let rogue_task_count=$rogue_task_count+1
  fi
done

if [[ $rogue_task_count -eq 0 ]]; then
  echo "OK: All tasks accounted for"
  exit 0
else
  echo "CRIT: $rogue_task_count tasks have no associated marathon app!"
  echo "This means that marathon has gotten out of sync with mesos."
  echo "Try restarting all the marathons to force things to reconcile!"
  exit 2
fi
```

This happens about once per week at Yelp. Restarting Marathon (causing a re-election) _does_ cause them to get cleaned up, but we don't like to restart Marathon, because of the risk of hitting #3783!

We were not monitoring for this condition before the upgrade to 1.3, so I'm not confident if this is a regression or not.

Our tooling does use a lot of `force=true` to abort or change existing deployments, do you know if this could trigger this?

solarkennedy:
Turns out I think the docs are all pretty good once I read them :)


aquamatthias:
@solarkennedy Thanks for reading the docs :)

PR #4617, opened 2016-10-25
Log an error if we can't deserialize a task instead of throwing (#4470)

jasongilanfarr:
Summary:
- While the underlying storage caught the proto error, the migration didn't actually
  handle the inconsistency.
- This only applies to 1.3

Test Plan: - sbt test

Reviewers: jdef

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D114


jdef:
lgtm once tests pass

Issue #3783, opened 2016-04-14
Marathon "recovering deployments", killing all tasks of some apps


EvanKrall:
At Yelp, we've had a number of incidents since we upgraded to 0.15.3 where Marathon has killed all instances of an app.

Something (it's not clear what; possibly the traceback in the log snippet below?) appears to be triggering the MarathonSchedulerActor to restart, which "recovers" deployments. During the recovery of those deployments, Marathon first `Start`s the app, which sets instances = 0; then later `Scale`s the app to the desired number. Since the app already existed, the `Start` action actually _kills_ the running tasks.

This appears to hit both applications that were created recently and have been deployed for some time.

Here's some snippets of logs from a particular app that had this problem. The first timestamp on the line is when syslog received the line, and is in UTC; the second timestamp is from Marathon and is PDT.

First, the app gets deployed initially.

```

2016-04-13T17:06:31+00:00 [2016-04-13 10:06:31,167] INFO [/servicename.shard--4.git378c7549.configcbe6b7a5]: new app detected (mesosphere.marathon.upgrade.GroupVersioningUtil$:ForkJoinPool-2-worker-269)
2016-04-13T17:06:31+00:00 [2016-04-13 10:06:31,193] INFO Computed new deployment plan:
2016-04-13T17:06:31+00:00 DeploymentPlan 2016-04-13T17:06:31.160Z
2016-04-13T17:06:31+00:00 step 1:
2016-04-13T17:06:31+00:00   * Start(App(/servicename.shard--4.git378c7549.configcbe6b7a5, image="docker-paasta.yelpcorp.com:443/services-servicename:paasta-378c754959731a4c8799c58a9169d6d385151adc")), instances=0)
2016-04-13T17:06:31+00:00 step 2:
2016-04-13T17:06:31+00:00   * Scale(App(/servicename.shard--4.git378c7549.configcbe6b7a5, image="docker-paasta.yelpcorp.com:443/services-servicename:paasta-378c754959731a4c8799c58a9169d6d385151adc")), instances=5)
2016-04-13T17:06:31+00:00  (mesosphere.marathon.state.GroupManager:ForkJoinPool-2-worker-269)
2016-04-13T17:06:31+00:00 [2016-04-13 10:06:31,193] INFO Deploy plan with force=false:
2016-04-13T17:06:31+00:00 DeploymentPlan 2016-04-13T17:06:31.160Z
2016-04-13T17:06:31+00:00 step 1:
[...]

PR #4623, opened 2016-10-27
Added regression test for leadership abdication.

jparkie:
Added regression test for leadership abdication as reproduced in https://github.com/EvanKrall/reproduce_marathon_issue_3783 for https://github.com/mesosphere/marathon/issues/3783.


jasongilanfarr:
Ok to test

jasongilanfarr:
Ok to test

jparkie:
What does it mean when none of the integration tests failed, but the overall Jenkins task is a failure: https://jenkins.mesosphere.com/service/jenkins/job/public-marathon-unit-integration-tests-pulls/1819/

I see my test did not fail though: https://jenkins.mesosphere.com/service/jenkins/job/public-marathon-unit-integration-tests-pulls/1819/testReport/mesosphere.marathon.integration/LeadershipAbdicationRegressionTest/

jasongilanfarr:
In your case, a test crashed so it wasn't reported as a failure. I'm adding logic to Terminate the leader on abdication/lose of leadership, will this logical change break this test?

jasongilanfarr:
retest this please

jparkie:
If the leader terminates and the tasks that are deploying get lost or killed, I think that would be a problem. I believe the pain point in the issue is that upon several leadership abdications, tasks which were deploying/deployed get killed.

jparkie:
@jasongilanfarr Hi! Can I have status update?

jasongilanfarr:
I have some test failures to work through (they weren't exactly designed for marathon to just die and I can't easily prevent it from not doing that in a test). Im debating turning them off until another set of changes come in regarding integration tests. I hoped to get to this all today but was working out another ZK connectivity blocker where we were supposed to exit but weren't actually exiting.

I'm thinking we should likely spin 1.3.6 with both changes incorporated and I'm hoping to polish the linked review tomorrow.

Given that we're going to just kill marathon on abdication, do you think this will fix your issue, assuming a watchdog is restarting it? If not, I'll prioritize a thorough review of this PR tomorrow.

Appreciate the effort by the way, just want to make sure my change doesn't make yours totally obsolete almost immediately. I wish I was checking the github issues to let you know I was tackling a related issue on the same day.

jparkie:
```
Given that we're going to just kill marathon on abdication, do you think this will fix your issue, assuming a watchdog is restarting it?
```

I think it will.

jdef:
we recently landed updates to marathon (master) re: termination upon leader abdication. PTAL

https://phabricator.mesosphere.com/D302

unterstein:
Hi @jparkie,

Your Commit was taken over and adapted within this PR, together with the fix: https://github.com/mesosphere/marathon/pull/4900

Thanks for reporting and writing the integration test 🎉

PR #4628, opened 2016-10-29
closes #4263; minor tweaks docs

### LINKED ISSUE #4263: Development container fails to start
Following the documentation from readme under the "Running the development Docker" header causes the container to be built properly, but, running 

`docker run -p 8080:8080 marathon-head --master local --zk zk://localhost/marathon`

fails with error message:

`Failed to start a local cluster while loading agent flags from the environment: Flag 'work_dir' is required, but it was not providedI0828 22:33:41.809705    72 leveldb.cpp:304] Persisting metadata (8 bytes) to leveldb took 3.432628ms`

It seems like explicitly setting the Mesos work directory by adding  `ENV MESOS_WORK_DIR /var/lib/mesos` to the Dockerfile resolves the issue.

cglewis:


sascala:
thank you, @cglewis

PR #4642, opened 2016-11-08
first draft of task handling

sascala:


docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Task Handling
+---
+
+Marathon handles tasks in the following ways. 
+
+Can a user access these messages? How? Via the API? The DC/OS CLI?
+
+# Terminal states
```
sascala: Is this category useful for a user?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+```
+case TASK_ERROR => Error
```
sascala: what happens after this?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
...
+case TASK_FINISHED => Finished
+```
+The task finished successfully.
+
+```
+case TASK_UNKNOWN => Unknown
+```
+The mster has no knowledge of the task. This is typically because either (a) the master never had knowledge of the task, or (b) the master forgot about the task because it garbaged collected its metadata about the task. The task may or may not still be running. When Marathon receives the Unknown message, it expunges the task.
+
+# Multi-part states
```
sascala: I'm not sure about this terminology

docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
...
+```
+Task is running.
+
+```
+case TASK_UNREACHABLE => Unreachable
+```
+The task was running on an agent that has lost contact with the master, typically due to a network failure or partition. The task may or may not still be running. When Marathon receives a task unreachable message with an unreachable time older than 15 minutes, Marathon marks the task as Unknown.                              
+
+```
+case TASK_LOST => inferStateForLost(taskStatus.getReason, taskStatus.getMessage)
```
sascala: Does this need to be in user-facing documentation?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
...
+case TASK_GONE => Gone
+```
+
+The task was running on an agent that has been shutdown (e.g., the agent become partitioned, rebooted, and then reconnected to the master; any tasks running before the reboot will transition from UNREACHABLE to GONE). The task is no longer running. After Marathon marks the task as gone, it expunges the task and starts a new one.
+
+```
+case TASK_GONE_BY_OPERATOR => Gone
+```
+The task was running on an agent that the master cannot contact; the operator has asserted that the agent has been shutdown, but this has not been directly confirmed by the master. If the operator is correct, the task is not running and this is a terminal state; if the operator is mistaken, the task might still be running, and might return to the RUNNING state in the future. After Marathon marks the task as failed, it expunges the task and starts a new one.    
+
```
sascala: Is this true?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
...
+```
```
sascala: Is TASK_DROPPED a case too?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,84 @@
...
+case TASK_FINISHED => Finished
+```
+The task finished successfully.
+
+```
+case TASK_UNKNOWN => Unknown
+```
+The mster has no knowledge of the task. This is typically because either (a) the master never had knowledge of the task, or (b) the master forgot about the task because it garbaged collected its metadata about the task. The task may or may not still be running. When Marathon receives the Unknown message, it expunges the task.
+
```
sascala: Does Marathon start another after it expunges the task?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Task Handling
+---
+
+Marathon handles tasks in the following ways.
```
sascala: Can a user access these messages? How? Via the API? The DC/OS CLI?

jeschkies: The logs show when Marathon reacts to a status change. It also publishes status changes on the event stream. One can also query the status of a task.

docs/docs/task-handling.md
jeschkies: `Error` is a terminal state. That we try to start a replacement task.

docs/docs/task-handling.md
jeschkies: @neilconway could you confirm this? I think the description is correct.

docs/docs/task-handling.md
jeschkies: Yupp. You mention it at the top : ).

docs/docs/task-handling.md
jeschkies: Yes. Marathon tries to keep the number of active tasks or rather instances constant. If a task is unknown and thus expunged it is not counted as active.

docs/docs/task-handling.md
jeschkies: Hm, what do you mean by "multi-part"?

docs/docs/task-handling.md
```diff
@@ -0,0 +1,82 @@
...
+
+```
+case TASK_RUNNING => Running
+```
+Task is running.
+
+```
+case TASK_UNREACHABLE => Unreachable
+```
+The task was running on an agent that has lost contact with the master, typically due to a network failure or partition. The task may or may not still be running. When Marathon receives a task unreachable message with an unreachable time older than 15 minutes, Marathon marks the task as Unknown.
```
jeschkies: Not quite. Marathon starts a replacement task first and eventually expunges the task if it does not become running in a certain time.

docs/docs/task-handling.md
jeschkies: No. We support TASK_LOST but map it to the other statuses. Marathon will or rather should never report a `TASK_LOST`.

docs/docs/task-handling.md
sascala: Is the task description is incorrect, how could Marathon ever start a successful replacement task?

docs/docs/task-handling.md
sascala: Well, "killing" isn't a terminal state, but it leads to "killed," which is

sascala:
@jeschkies Here's a diagram draft: https://www.gliffy.com/go/share/s4ejnmtlti430eyjgj8x

Questions:
- If Marathon gives a task the Error tag when the description has an error, will it ever be able to restart the task successfully?

- Does Marathon restart tasks first, or expunge the old one and then restart the new?

- Does Marathon restart a task after every error? Does it restart a task when it has finished?

docs/docs/configure-task-handling.md
```diff
@@ -0,0 +1,18 @@
+---
+Configuring Task Handling
+---
+
+You can configure Marathon's actions on unreachable tasks. The `unreachableStrategy` parameter of your app definition allows you to configure this in two ways: by defining when a new task instance should be launched, and by defining when a task instance should be expunged.
```
sascala: Can a user set this in a pod definition as well? Or does the `unreachableStrategy` of an app in a pod dictate the pod's behavior?

docs/docs/configure-task-handling.md
```diff
@@ -0,0 +1,18 @@
...
+
+```
```
sascala: Does a user need to define both of these?

docs/docs/configure-task-handling.md
jeschkies: Yep. Pod also has an `unreachableStrategy`. Also note, that we added `killSelection`: https://phabricator.mesosphere.com/D243.

docs/docs/configure-task-handling.md
jeschkies: No, we have defaults for both ( https://phabricator.mesosphere.com/D243#88458d0e): 5 and 10 minutes.

docs/docs/configure-task-handling.md
```diff
@@ -0,0 +1,18 @@
...
+"unreachableStrategy": {
+	"timeUntilInactiveSeconds": <integer>,
+	"timeUntilExpungeSeconds": <integer>
+}
+```
+
+- `timeUntilInactiveSeconds`: If a task instance is unreachable for longer than this value, it is marked inactive and a new instance will launch. At this point, the unreachable task is not yet expunged. The minimum value for this parameter is 1.
```
jeschkies: It was renamed to `unreachableInactiveAfterSeconds`.

docs/docs/configure-task-handling.md
```diff
@@ -0,0 +1,18 @@
...
+
+- `timeUntilExpungeSeconds`: If an instance is unreachable for longer than this value, it will be expunged. An expunged task will be killed if it ever comes back. Instances are usually marked as unreachable before they are expunged, but that is not required. The minimum value for this parameter is 2.
```
jeschkies: It has been renamed to `unreachableExpungeAfterSeconds`.

"The minimum value for this parameter is 2." That is not quite true. It has to be bigger than `unreachableInactiveAfterSeconds`. The rest is good : ).

docs/docs/task-handling.md
jeschkies: That is a good point. I think it cannot.

docs/docs/configure-task-handling.md
sascala: It looks like `killSelection` is in the top level of the app definition and the `PodSchedulingPolicy` level of the pod definition. I'm not sure how much more documentation is needed about this apart from the description you've written. If more _would_ be useful, where would be the best place for it? The "Blue-Green Deployment" section, perhaps?

docs/docs/task-handling.md
jeschkies: I think it tries forever but I'm not sure. @aquamatthias can probably answer this.

docs/docs/configure-task-handling.md
```diff
@@ -15,4 +15,15 @@ You can configure Marathon's actions on unreachable tasks. The `unreachableStrat
...
 
-You can use `unreachableInactiveAfter` and `unreachableExpungeAfter` in conjunction with one another. For example, if you configure `unreachableInactiveAfter = 60` and `unreachableExpungeAfter = 120`, a task instance will be expunged if it has been unreachable for more than 120 seconds and a second instance will be started if it has been unreachable for more than 60 seconds.
\ No newline at end of file
+You can use `unreachableInactiveAfter` and `unreachableExpungeAfter` in conjunction with one another. For example, if you configure `unreachableInactiveAfter = 60` and `unreachableExpungeAfter = 120`, a task instance will be expunged if it has been unreachable for more than 120 seconds and a second instance will be started if it has been unreachable for more than 60 seconds.
+
+# Kill Selection
+You call also define a kill selection to declaring whether Marathon kills the youngest or oldest tasks first when rescaling or otherwise killing multiple tasks. The default value for this parameter is `YoungestFirst`. You can also specify `OldestFirst`.
```
jeschkies: > You call also define a kill selection to declaring...

Sounds a little off. Aside from that it's great :).

sascala:
@jeschkies Thanks for working with me on this. I'm considering this doc done, but waiting to merge until I work out with @jasongilanfarr what the best doc publishing workflow should be.

docs/docs/configure-task-handling.md
```diff
@@ -18,12 +18,12 @@ You can configure Marathon's actions on unreachable tasks. The `unreachableStrat
 You can use `unreachableInactiveAfter` and `unreachableExpungeAfter` in conjunction with one another. For example, if you configure `unreachableInactiveAfter = 60` and `unreachableExpungeAfter = 120`, a task instance will be expunged if it has been unreachable for more than 120 seconds and a second instance will be started if it has been unreachable for more than 60 seconds.
 
 # Kill Selection
-You call also define a kill selection to declaring whether Marathon kills the youngest or oldest tasks first when rescaling or otherwise killing multiple tasks. The default value for this parameter is `YoungestFirst`. You can also specify `OldestFirst`.
+You call also define a kill selection to declare whether Marathon kills the youngest or oldest tasks first when rescaling or otherwise killing multiple tasks. The default value for this parameter is `YoungestFirst`. You can also specify `OldestFirst`.
```
jeschkies: "You can also define ..."

sascala:
The docs web site is now pegged to a release branch, not master, so I can merge this. Finally.

docs/docs/task-handling.md
```diff
@@ -0,0 +1,78 @@
+
+---
+title: Task Handling
+---
+
+Marathon handles tasks in the following ways. You can consult the Marathon logs for these messages or query the status of the task via the [events stream](http://mesosphere.github.io/marathon/docs/event-bus.html) of the [Marathon REST API](https://mesosphere.github.io/marathon/docs/generated/api.html) (/v2/events).
```
janisz: File link should be used here

PR #4644, opened 2016-11-09
Fixes #1428 Converts command arguments to environment variables and back again (merge to release/1.3)

### LINKED ISSUE #1428: Remove user credentials from Marathon command line
Problem:
Currently Marathon allows to add HTTP user credentials (ID & password) via --http-credentials command line parameter. When the process is running anyone can easily see these credentials in the "ps" command output.

Proposed solutions:
1. Modify --http-credentials parameter once the process is started so it does not display in the "ps" output.
2. Read credentials from a file (preferred). This way more than one user can be defined.
3. Authenticate users against LDAP (even better).

Thanks
Nikolay

brad-peters:
This change allows environment variables to be passing into marathon so they do not show up in ps.

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -34,6 +34,31 @@ function element_in {
   return 1
 }
 
+function cmdToenv() {
+  cmds=("$@")
+  tLen=${#cmds[@]}
+  # skip run_jar
+  for (( i=1; i<${tLen}; i+=2 ));
+  do
+   key=${cmds[$i]}
+   value="" # default if no value for key
+   if (( $i+1 < ${tLen} )); then
+     value=${cmds[$i+1]}
+   fi
+
+   if [[ $value == --* ]] ; then
+     exportEnv $key ""
+     i=$i-1
+   else
+     exportEnv $key $value
+   fi
+  done
+}
+
+function exportEnv() {
+  export MARATHON_$(echo "$1" | sed "s/--//" | awk '{print toupper($0)}')=$2
+}
+
 function load_options_and_log {
   # Call mesosphere-dnsconfig if present on the system to generate config files.
   [ -x /usr/bin/mesosphere-dnsconfig ] && mesosphere-dnsconfig -write -service=marathon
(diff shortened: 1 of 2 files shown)


bin/marathon-framework
```diff
@@ -75,7 +100,6 @@ function load_options_and_log {
       cmd+=( --master "$(cat /etc/mesos/zk)" )
     fi
   fi
-  echo "${cmd[@]}"
```
brad-peters: Some commands/environment variables are secrets we do not want logged

jasongilanfarr:
ok to test

iderdik:
What is the intent here. Environment variables are still visible via `docker inspect`? What could make this more helpful is better support for treating environment variables as secrets and encrypting them.

brad-peters:
The intent was the arguments for the marathon process itself not marathon apps.

The bash script inside of the marathon_runnable.jar converts environment variables to command line arguments for the marathon jar.  So they appear when you run ps.  

Perhaps this doesn't solve both problems in that issue, only the marathon process half.  Maybe there should have been 2 different issues

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -29,12 +29,30 @@ class MarathonApp extends App {
     )
   }
 
-  override val conf = new AllConf(args)
+  private val envPrefix = "MARATHON_"
```
jasongilanfarr: Constant should be "EnvPrefix" as per scala style guidelines.

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -29,12 +29,30 @@ class MarathonApp extends App {
...
+  private val envArgs: Array[String] = {
```
jasongilanfarr: Prefer `Seq[]` please. You should be able to do something like

`
sys.env.withFilter(_._1.startsWith(EnvPrefix)).flatMap { case (key, value) => Seq(s"--${key.replaceAll(EnvPrefix, "").toLowerCase.trim}", if (value.trim.length > 0) value else Nil) }(collection.breakOut)
`

brad-peters:
@jasongilanfarr I have another branch with these changes based off of master, should I update this merge request with that branch?

jasongilanfarr:
@brad-peters Yes please. I'm open to merging into the releases branch as well, but let's first have it in master.

Issue #4338, opened 2016-09-09
taskKillGracePeriodSeconds doesn't work


robfrut135:
Hi,
I'm using Mesos 1.0.1 and Marathon 1.3.0-RC6, when I put "taskKillGracePeriodSeconds", app not deploy on Marathon, not take effect. From my view, Marathon UI not recognise this field.

unterstein:
Hi @robfrut135,

I retested this a couple of seconds ago and I could configure an app with `taskKillGracePeriodSeconds` through the UI which is recognized and passed correctly. Could you please paste your app definition?

Thanks
Johannes

robfrut135:
Hi,
I followed this steps:

From UI:

![captura de pantalla de 2016-09-09 18 52 09](https://cloud.githubusercontent.com/assets/11392338/18395398/9afb736a-76be-11e6-97fd-626c507fc520.png)

![captura de pantalla de 2016-09-09 18 37 08](https://cloud.githubusercontent.com/assets/11392338/18395238/0bb0fee6-76be-11e6-8335-28212cbc862c.png)

![captura de pantalla de 2016-09-09 18 37 24](https://cloud.githubusercontent.com/assets/11392338/18395283/471197b6-76be-11e6-9d33-73dcd67832da.png)

JSON from API:

![captura de pantalla de 2016-09-09 18 42 33](https://cloud.githubusercontent.com/assets/11392338/18395317/5f480a5e-76be-11e6-8e99-a1d7f593dcde.png)

Here the full JSON (remove some internal data)

`{
  "app": {
    "id": "/genesis/dev/plaid",
    "cmd": null,
    "args": [
      "kst-genesis/plaid/dev"
    ],
    "user": null,
    "env": {
      "CONSUL": "consul.service.consul:8500"
    },
    "instances": 1,
    "cpus": 0.2,
    "mem": 512,
    "disk": 0,
    "gpus": 0,
    "executor": "",
    "constraints": [
      [
        "hostname",
        "UNIQUE"
      ]
    ],
    "uris": [      
    ],
    "fetch": [      
    ],
    "storeUrls": [      
    ],
    "backoffSeconds": 5,
    "backoffFactor": 2,
    "maxLaunchDelaySeconds": 60,
    "container": {
      "type": "DOCKER",
      "volumes": [        
      ],
      "docker": {
        "image": "genesis-plaid:0.0.2",
        "network": "BRIDGE",
        "portMappings": [
          {
            "containerPort": 9000,
[...]

PR #4649, opened 2016-11-11
full example of app using taskKillGracePeriodSeconds

tgermain:
The doc is a bit misleading about  `taskKillGracePeriodSeconds` being a label you add to the `labels` of your application or a primary key of the app definition json. 

Hope it help a bit #4338


docs/docs/health-checks.md
```diff
@@ -211,6 +211,22 @@ done
 
 To set the necessary grace period, add the `taskKillGracePeriodSeconds` label to your application definition:
```
jdef: s/label/field/ please

jdef:
ok to test

tgermain:
Done, indeed field avoid confusion around yet another label.

docs/docs/health-checks.md
```diff
@@ -209,8 +209,24 @@ while true; do
...
 
+```json
+{
+  "id": "/foo",
+  "instances": 2,
+  "cmd": "sleep 1000",
+  "cpus": 0.1,
+  "disk": 0,
+  "mem": 16,
+  "args": [
```
gkleiman: An app specifying both `cmd` and args` is not valid:

> As of version 0.7.0, Marathon supports an args field in the app JSON. It is invalid to supply both cmd and args for the same app. The behavior of cmd is as in previous releases (the value is wrapped by Mesos via /bin/sh -c '${app.cmd}).

PR #4652, opened 2016-11-14
Fixes #1428 Converts command arguments to environment variables and back again (merge to master)

### LINKED ISSUE #1428: Remove user credentials from Marathon command line

brad-peters:
This change allows environment variables to be passing into marathon so they do not show up in ps.

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -41,6 +41,31 @@ function element_in {
   return 1
 }
 
+function cmd_to_env() {
+  cmds=("$@")
+  tLen=${#cmds[@]}
+  # skip run_jar
+  for (( i=1; i<${tLen}; i+=2 ));
+  do
+   key=${cmds[$i]}
+   value="" # default if no value for key
+   if (( $i+1 < ${tLen} )); then
+     value=${cmds[$i+1]}
+   fi
+
+   if [[ $value == --* ]] ; then
+     export_env $key ""
+     i=$i-1
+   else
+     export_env $key $value
+   fi
+  done
+}
+
+function export_env() {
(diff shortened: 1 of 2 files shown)


jdef:
ok to test

jdef:
xref #4644

bin/marathon-framework
```diff
@@ -92,9 +116,11 @@ function load_options_and_log {
         clean_cmd+=( "${i}" )
       fi
     done
-    "${clean_cmd[@]}"
+    cmd_to_env $clean_cmd
```
h0tbird: I think this should be `cmd_to_env "${clean_cmd[@]}"`

bin/marathon-framework
```diff
@@ -92,9 +116,11 @@ function load_options_and_log {
...
+    run_jar
   else
-    logged marathon "${cmd[@]}" "$@"
+    cmd_to_env $cmd
```
h0tbird: I think this should be `cmd_to_env "${cmd[@]}"`

h0tbird:
Fixed in https://github.com/mesosphere/marathon/commit/85a32314d6c60d620c1fa2cc8d5660b23190cf47

PR #4667, opened 2016-11-19
Log akka configuration on start

janisz:


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4667" target="_blank">ReviewNinja

jasongilanfarr:
ok to test

jasongilanfarr:
I like this overall. I'm a bit concerned that it could expose secrets since the command line and environment variables are in the full config. Any thoughts @janisz

janisz:
You're right it's not necessary and potentially dangerous. I saw config is printed in metronome and thought that could be useful for tuning akka configuration.

jasongilanfarr:
I think the right way to do this would be to load the config ourselves
(passing it to the actor system) and log it out with secret like things
masked out.

Issue #4637, opened 2016-11-04
Add filter option for event types on GET /v2/events endpoint


tobilg:
As far as I understand, when using server-sent events via the `/v2/events` endpoint, currently all events are sent from Marathon.

I think there are many use cases where applications registering on that endpoint are actually only interested in some specific event types, not all events. Also, in large clusters with lots of tasks, I guess there will be a lot of traffic over the connection, putting load on Marathon as well as the listening application itself.

I could imagine that something like the `embed` query string parameter of the [GET /v2/apps](http://mesosphere.github.io/marathon/docs/generated/api.html#v2_apps_get) would make sense, e.g.

`GET /v2/events?event_types=deployment_info,deployment_success,deployment_failure`

tobilg:
I just saw that there's also https://github.com/mesosphere/marathon/issues/654, which is still open.

janisz:
Related to #3154

tobilg:
Thanks @janisz for linking the issues, and to @jasongilanfarr as well for labeling it as well.

I think a good example for this kind of filtering is actually the Facebook Graph API with it's [field expansion](https://developers.facebook.com/docs/graph-api/using-graph-api/#fields) functionality. With that, it's even possible to filter/define the result on nested sub-fields (objects).

For example, a basic query can look like this: https://graph.facebook.com/?id=http://www.google.com

which will result in 

```javascript
{
   "og_object": {
      "id": "381702034999",
      "description": "Search the world's information, including webpages, images, videos and more. Google has many special features to help you find exactly what you're looking for.",
      "title": "Google",
      "type": "website",
      "updated_time": "2016-11-17T09:31:49+0000"
   },
   "share": {
      "comment_count": 2,
      "share_count": 39738503
   },
   "id": "http://www.google.com"
}
```

If one uses the field expansion like https://graph.facebook.com/?fields=og_object{description},share{share_count}&id=http://www.google.com

one receives this instead:

```javascript
{
   "og_object": {
      "description": "Search the world's information, including webpages, images, videos and more. Google has many special features to help you find exactly what you're looking for.",
      "id": "381702034999"
   },
   "share": {
      "share_count": 39738503
   },
   "id": "http://www.google.com"
}
```

janisz:
I was tying to implement it few months ago. But I hit a problem with keeping it backward compatible with old implementation. Subscriptions are stored in ZK so we need to add migration step to handle it correctly. 

Now it become critical for me because some events are > 8 MB.

[...]

PR #4669, opened 2016-11-19
[WIP] Fixes #3154 Filter events

### LINKED ISSUE #3154: Events are too big
Some of marathon events (e.g.: `DeploymentStepSuccess`) are big (> 1MB), because contains data of all groups, even if they are not changed. 
Possible solutions:
- [x] Event subscription could specify which events should be sent to subscriber https://github.com/mesosphere/marathon/issues/4637
- [] Events should contains only changed data (possible not backward compatible)

janisz:
First draft of enabling filtering of event callbacks to fix #3154
My goal it to provide configurable way of filtering events before they were sent. On big clusters events becomes a problem (especially deployments events with size 10MB). This leads to errors and increase events delay. 

My idea is to allow user use jsonpath or some other way to get field and compare it with accepted values. Accepted values contains specific field event will be delivered if no then it will be skipped.

Pleas give me some clues if this is good idea. Maybe we should stick to `event_type` and do not allow more filtering but it was request in #4637 by @tobilg 

> [here might be use cases where it would make sense to filter on field values as well, so as if an application wants to only listen to a specific applications events, such as status updates of a certain appId.](https://github.com/mesosphere/marathon/issues/4637#issuecomment-261202413)

Related to: [allegro/marathon-consul#139](https://github.com/allegro/marathon-consul/issues/139)

### DIFF
--- a/docs/docs/rest-api/mesosphere/marathon/api/v2/EventSubscriptionsResource_listSubscribers.md
+++ b/docs/docs/rest-api/mesosphere/marathon/api/v2/EventSubscriptionsResource_listSubscribers.md
@@ -29,5 +29,9 @@ Transfer-Encoding: chunked
         "http://localhost:9090/callback",
         "http://localhost:9191/callback"
     ]
+    "callbacks": {
+        "http://localhost:9090/callback": {},
+        "http://localhost:9191/callback": {}
+    }
 }
 ```
--- a/docs/docs/rest-api/mesosphere/marathon/api/v2/EventSubscriptionsResource_subscribe.md
+++ b/docs/docs/rest-api/mesosphere/marathon/api/v2/EventSubscriptionsResource_subscribe.md
@@ -30,6 +30,58 @@ Transfer-Encoding: chunked
 {
     "callbackUrl": "http://localhost:9292/callback",
     "clientIp": "127.0.0.1",
+    "filters": [],
     "eventType": "subscribe_event"
+}
+```
+
+Messages could be filtered by event type. To register filters provide JSON body
+with event filters definitions.
+
+### Example
+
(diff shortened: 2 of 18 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4669" target="_blank">ReviewNinja

tobilg:
Thanks @janisz, just a quick question: Does the PR support both event callbacks and the event stream, or is it callbacks only?

janisz:
Callbacks only. I don't believe it will be merged in nearest soon. It's quite big change involving zk migration and API change. I'll prepare another patch for SEE because it there is no persistence involved so it's easier to change.

tobilg:
I see, thanks for the fast feedback! I started to build an "event proxy" in Node to be able to have this filter functionality available quickly. It solely relies on SSEs, and acts as proxy between Marathon and the client applications. 

The better solution would be to have this directly in Marathon I guess...

jasongilanfarr:
Is Ike to consider deprecating the non-sse event style. Thanks for taking this on, I'll try to review this weekend.

We've been using phabricator lately for our own reviews, feel free to give it a shot, see http://phabricator.mesosphere.com/D48 for the updated contributing docs.

janisz:
This is only a draft to take thoughts and comments on way to implement it.
+1 to to deprecate http callbacks, maybe there is no point to review this

jasongilanfarr:
I do think we should deprecate non-SSE and only fix this in SSE style. I'll file a separate issue to deprecate non-SSE

janisz:
@jasongilanfarr Closing due to https://github.com/mesosphere/marathon/issues/4675

PR #4671, opened 2016-11-20
Fixes #4637 Allow filtering SSE events by types

### LINKED ISSUE #4637: Add filter option for event types on GET /v2/events endpoint

janisz:
Allow recievieng only specific events by providing their types
in `event_type` query params. Old behavaiour remains unchanged.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
@@ -24,13 +24,21 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends 
 
   lazy val id: String = UUID.randomUUID().toString
 
+  val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
+
+  def subscribed(eventType: String): Boolean = {
+    subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
+  }
+
   override def remoteAddress: String = request.getRemoteAddr
 
   override def close(): Unit = emitter.close()
 
-  override def sendEvent(event: String, message: String): Unit = blocking(emitter.event(event, message))
+  override def sendEvent(event: String, message: String): Unit = {
+    if (subscribed(event)) blocking(emitter.event(event, message))
+  }
 
-  override def toString: String = s"HttpEventSSEHandle($id on $remoteAddress)"
+  override def toString: String = s"HttpEventSSEHandle($id on $remoteAddress on event types from $subscribedEventTypes)"
 }
 
 /**
--- a/src/test/scala/mesosphere/marathon/core/event/impl/stream/HttpEventSSEHandleTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/event/impl/stream/HttpEventSSEHandleTest.scala
@@ -1,0 +1,59 @@
+package mesosphere.marathon.core.event.impl.stream
+
+import javax.servlet.http.HttpServletRequest
+
+import mesosphere.marathon.test.{ MarathonSpec, Mockito }
+import org.eclipse.jetty.servlets.EventSource.Emitter
+import org.scalatest.{ GivenWhenThen, Matchers }
+import collection.JavaConversions._
+
+class HttpEventSSEHandleTest extends MarathonSpec with Matchers with Mockito with GivenWhenThen {
+
+  test("events should be filtered") {
+    Given("An emiter")
+    val emitter = mock[Emitter]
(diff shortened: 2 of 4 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4671" target="_blank">ReviewNinja

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -24,13 +24,21 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
 
   lazy val id: String = UUID.randomUUID().toString
 
+  val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
```
jasongilanfarr: `val subscribedEventTypes: Set[String] = request.getParameterMap.getOrDefault("event_type", Set.empty)(collection.breakOut)` might work if you have `import mesosphere.marathon.stream._`

src/test/scala/mesosphere/marathon/core/event/impl/stream/HttpEventSSEHandleTest.scala
```diff
@@ -0,0 +1,59 @@
+package mesosphere.marathon.core.event.impl.stream
+
+import javax.servlet.http.HttpServletRequest
+
+import mesosphere.marathon.test.{ MarathonSpec, Mockito }
+import org.eclipse.jetty.servlets.EventSource.Emitter
+import org.scalatest.{ GivenWhenThen, Matchers }
+import collection.JavaConversions._
+
+class HttpEventSSEHandleTest extends MarathonSpec with Matchers with Mockito with GivenWhenThen {
```
jasongilanfarr: Please use either `UnitTest` or `FunTest` as the base type. They have all of the other traits mixed in.

jasongilanfarr:
ok to test

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
janisz: I have a problem with your code. Prefer to leave it as is.

```
[error] src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala:28: polymorphic expression cannot be instantiated to expected type;
[error]  found   : [A]scala.collection.immutable.Set[A]
[error]  required: Array[String]
[error]   val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Set.empty)(collection.breakOut)
[error]                                                                                     ^
[error] src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala:12: Unused import
[error] import mesosphere.marathon.stream._
[error]                                   ^
[error] two errors found
```

src/test/scala/mesosphere/marathon/core/event/impl/stream/HttpEventSSEHandleTest.scala
janisz: Fixed

janisz:
What type of test? Unit test or integration test? I could test that `newEventSource` create new handle with every call but this will repeat implementation. I can create integration test but spray-can doesn't support SSE and I need to introduce dependencies for this test to parse response.

Are you thinking about something like this? 
```
  test("adding an event subscriber") {

    When("read event stream without filters")
    val defaultClient = marathon.events()

    When("an filtered event subscriber is added")
    val filteredClient = marathon.events("non_existing_event_type")

    //TODO(janisz) delete when SSE support appear
    When("marathon is stopped")
    marathonServer.stop()

    Then("a notification should be sent to default subscribers")
    defaultClient.entityString.contains("type")
    And("not to filtered")
    !filteredClient.entityString.contains("type")

    // Cleanup
    marathon.unsubscribe("http://localhost:1337")
  }
```

jasongilanfarr:
No worries - can you file an issue to track adding one?

janisz:
Thanks. Issue: https://github.com/mesosphere/marathon/issues/4686

PR #4691, opened 2016-11-23
An app cannot specify both cmd and args

tgermain:
Missing correction of PR https://github.com/mesosphere/marathon/pull/4649


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4691" target="_blank">ReviewNinja

PR #4713, opened 2016-11-28
re-enabled rescheduling for resident tasks

unterstein:
The current behavior of marathon when receiving an `InstanceChange` for a resident task is, that no `ScaleRunSpec` is propagated.
This change will correct this behavior and also send `ScaleRunSpec` events if `InstanceChange` is received for an instance in `Condition. UnreachableInactive`.


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4713" target="_blank">ReviewNinja

src/test/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImplTest.scala
```diff
@@ -13,6 +13,35 @@ import mesosphere.marathon.state.{ PathId, Timestamp }
...
+    Condition.Starting,
+    Condition.Unreachable,
+    Condition.UnreachableInactive,
+    Condition.Gone,
+    Condition.Dropped,
+    Condition.Unknown
+  )
+
+  val scalingWorthyConditions = allConditions.filter(scalingWorthy)
+  val notScalingWorthyConditions = allConditions.filterNot(scalingWorthy)
```
jeschkies: I'm fine with this change but would prefer scala test tables to avoid duplicate implementations:

```scala
class ScaleAppUpdateStepImplTest extends AkkaUnitTest with TableDrivenPropertyChecks {


"ScaleAppUpdateStep" when {
  // format: OFF  
  val conditions = Table (
    ("condition",             "initially", "with previous failed condition", "with previous running condition"),
    (Condition.Failed),    true       , false,                                              true                                                ),
    (Condition.Running), false     , false,                                              false                                              )
    ...
    )
  }
  //format: ON

  forAll (conditions) { (condition, scalingWorthy, scalingWorthyAfterFailed, scalingWorthyAfterRunnig) =>
    s"receiving status update $condition" should {
      s"send no rescaling request when " in { ... } 
    }
  }
}
}
```

Something like that. Here is an example https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/core/task/TaskStatusComparisonTest.scala#L16.

Github does not align so well.

src/test/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImplTest.scala
```diff
@@ -53,6 +81,40 @@ class ScaleAppUpdateStepImplTest extends AkkaUnitTest {
       }
     }
 
+    scalingWorthyConditions.foreach { newStatus =>
+      s"receiving a scaling worthy status update '$newStatus' on a previously scaling worthy condition" should {
+        val f = new Fixture
+
+        val instance = TestInstanceBuilder.newBuilder(PathId("/app"))
+          .addTaskUnreachable(containerName = Some("unreachable1"))
```
jeschkies: Shouldn't this be failed instead of unreachable?

src/test/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImplTest.scala
```diff
@@ -53,6 +81,40 @@ class ScaleAppUpdateStepImplTest extends AkkaUnitTest {
...
+        }
+      }
+    }
+
+    scalingWorthyConditions.foreach { newStatus =>
+      s"receiving a scaling worthy status update '$newStatus' on a previously non scaling worthy condition" should {
+        val f = new Fixture
+
+        val instance = TestInstanceBuilder.newBuilder(PathId("/app"))
+          .addTaskUnreachable(containerName = Some("unreachable1"))
```
jeschkies: Shouldn't this be running?

src/test/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImplTest.scala
unterstein: thanks, changed this

src/test/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImplTest.scala
unterstein: thanks, changed this

src/test/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImplTest.scala
unterstein: Thanks, definitely will try this style next time :)

PR #4770, opened 2016-12-05
Add new deployment integration tests [only for jenkins tests]

zen-dog:
Summary:
- TwoAppsDeployWithAbdicationIntegrationTest will test an app with 2 tasks one of which becomes
HEALTHY before master abdication and second one after. Deployment should succeed with 2 updated
tasks.
- AppDeployWithReadinessWithAbdicationIntegrationTest will test an app with 2 tasks one of which
becomes READY before master abdication and second one after. Deployment should succeed with 2
updated tasks. This tests fails before D222.

Test Plan: integration-tests

Reviewers: aquamatthias, unterstein

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D254


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4770" target="_blank">ReviewNinja

PR #4771, opened 2016-12-05
Parse JSON event only once before broadcast

janisz:
Current implementation parses events for every callback.
This patch make parsing happen only once for event.


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4771" target="_blank">ReviewNinja

jdef:
ok to test

jasongilanfarr:
Can you rebase so we can merge?

janisz:
Done

Issue #3031, opened 2016-01-19
Convert all synchronous TaskTracker calls to asynchronous calls


kolloch:
Historically, the `TaskTracker` used concurrent data structures to handle state access. That allowed synchronous access to the state. Since we have now an actor based `TaskTracker` implementation, there is a compatibility layer that still allows synchronous calls but really performs an `ask` and an `Await.result` call.

The problem is that all potentially blocking constructs (like Await.result) are marked with `scala.concurrent.blocking` to prevent deadlocks. `scala.concurrent.blocking` just results in spawning more threads when needed (which is good). That means that as long as we use blocking code constructs, the number of threads that Marathon uses is, in effect, unbounded.

meichstedt:
Note: This issue has been migrated to https://jira.mesosphere.com/browse/MARATHON-3219. For more information see https://groups.google.com/forum/#!topic/marathon-framework/khtvf-ifnp8.

Issue #4693, opened 2016-11-23
Excess threads spawned due to blocking code.


janisz:
Marathon threads count is *O(n)* where *n* is number of applications with health check.  This decrease app performance and prevent using marathon at scale.

### Why
Marathon combines akka actors and object oriented services and managers. Unfortunately whenever blocking request is called by the actor new thread is spawned. [Read more](http://stackoverflow.com/a/31110111/1387612)

### Where
Every `Await`, `blocking`, `lock` called inside the actor is dangerous and should be re-implemented in asynchronous manner (returns future or callback message).
Dangerous examples that creates new thread: 
- [`mesosphere.marathon.core.task.tracker.impl.InstanceTrackerDelegate#instancesBySpecSync`](https://github.com/mesosphere/marathon/blob/c5ff20237855dfcb0869f1d158eae4455285c01b/src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala#L32-L35)
- [`mesosphere.marathon.core.launchqueue.impl.LaunchQueueDelegate#askQueueActor`](https://github.com/mesosphere/marathon/blob/d97ecff5febbd9fdc45cf884512723ccd47fd9ea/src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueDelegate.scala#L49-L55)
- [`mesosphere.marathon.core.launchqueue.impl.LaunchQueueDelegate#askQueueActorFuture`](https://github.com/mesosphere/marathon/blob/d97ecff5febbd9fdc45cf884512723ccd47fd9ea/src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueDelegate.scala#L57-L68)
- [`mesosphere.util.RWLock#withLock`](https://github.com/mesosphere/marathon/blob/31426bd1f7b091b78748ea7fd83c497db3dfea9e/src/main/scala/mesosphere/util/RWLock.scala#L9-L13)

Above examples were found byt analysing stack traces of [`akka.dispatch.MonitorableThreadFactory#newThread`](https://github.com/akka/akka/blob/4a7cd147313404eb8137e56163e54f06c23640d0/akka-actor/src/main/scala/akka/dispatch/ThreadPoolBuilder.scala#L185-L190)

### Hotfix
1. Do NOT use HTTP and TCP health-checks use Mesos command health-check instead
2. Do NOT spawn multiple apps at once (use rate limiter or somehow delay concurrent requests)
[...]

PR #4773, opened 2016-12-05
Use asynchonus call to TaskTracker

janisz:
Related to #3031


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4773" target="_blank">ReviewNinja

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -180,17 +182,21 @@ private[health] class HealthCheckActor(
...
               log.error(s"Couldn't find instance $instanceId")
               health.update(result)
-          }
+          })
       }
 
-      healthByInstanceId += (instanceId -> newHealth)
+      futureHealth.onComplete {
+        case Success(newHealth) =>
+          healthByInstanceId += (instanceId -> newHealth)
```
jdef: pretty sure that this is an unsafe operation now that it's completing in the context of a future instead of an actor mailbox. consider changing the type of `healthByInstanceId` to a TrieMap

jdef:
ok to test

jdef:
@aquamatthias any concern about out-of-order completion of health status updates?

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
janisz: Agree, this is dangerous. How about sending complete function back to actor to have proper locking and state updating? Actor will have another case class (e.g., Complete) that will contain function that need to be run in order to complete job.

jdef:
either way: actor message or TrieMap

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -180,17 +191,23 @@ private[health] class HealthCheckActor(
...
-      healthByInstanceId += (instanceId -> newHealth)
-
-      if (health.alive != newHealth.alive && result.publishEvent) {
-        eventBus.publish(HealthStatusChanged(app.id, instanceId, result.version, alive = newHealth.alive))
-        // We moved to InstanceHealthChanged Events everywhere
-        // Since we perform marathon based health checks only for apps, (every task is an instance)
-        // every health result is translated to an instance health changed event
-        eventBus.publish(InstanceHealthChanged(instanceId, result.version, app.id, Some(newHealth.alive)))
+      futureHealth.onComplete { r =>
+        self ! Finish(() => () => r match {
```
jdef: `() => () => ...` why?

jdef:
xref #4693

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -64,10 +66,15 @@ private[health] class HealthCheckActor(
       app.version,
       healthCheck
     )
-    val activeInstanceIds: Set[Instance.Id] = instanceTracker.specInstancesLaunchedSync(app.id).map(_.instanceId)(collection.breakOut)
-    // The Map built with filterKeys wraps the original map and contains a reference to activeInstanceIds.
-    // Therefore we materialize it into a new map.
-    healthByInstanceId = healthByInstanceId.filterKeys(activeInstanceIds).iterator.toMap
+    instanceTracker.instancesBySpec().map(
+      _.specInstances(app.id).filter(_.isLaunched).map(_.instanceId)(collection.breakOut)
+    ).onComplete { r =>
```
jasongilanfarr: Not really a fan of this style, instead, consider using FSM and stashing or specialized "Finished" where the logic is in the actor, not in the message itself. While I'd prefer FSM for this, it could probably be done well using become and stashing only as a smaller change.

janisz: How this should work with FSM? We will get events. If health event needs data from TaskTracker it is stashed. When actor receive message with TaskTracker data it process all stashed messages. Am I correct?

jasongilanfarr:
That's what I was thinking yes. With FSM, you could track pending updates
in the state while waiting for a updated instances info.

janisz:
I'm closing this PR. I'll reopen it when I prepare FSM/stash version.

janisz:
@jasongilanfarr @jdef I take different approach. I think it uses the idea with FSM and stash but without them. IMO the code is easier to understand than with stash/FSM

jdef:
ok to test

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -57,14 +59,21 @@ private[health] class HealthCheckActor(
...
+
+  def purgeStatusOfDoneInstances(instances: Seq[Instance]): Unit = {
     log.debug(
       "Purging health status of inactive instances for app [{}] version [{}] and healthCheck [{}]",
       app.id,
       app.version,
       healthCheck
     )
-    val activeInstanceIds: Set[Instance.Id] = instanceTracker.specInstancesLaunchedSync(app.id).map(_.instanceId)(collection.breakOut)
+    val activeInstanceIds: Set[Instance.Id] = instances.filter(_.isLaunched).map(_.instanceId)(collection.breakOut)
```
jasongilanfarr: withFilter instead of filter

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -151,20 +160,23 @@ private[health] class HealthCheckActor(
...
+      purgeStatusOfDoneInstances(instances)
+      dispatchJobs(instances)
+
     case result: HealthResult if result.version == app.version =>
       log.info("Received health result for app [{}] version [{}]: [{}]", app.id, app.version, result)
       val instanceId = result.instanceId
       val health = healthByInstanceId.getOrElse(instanceId, Health(instanceId))
 
-      val newHealth = result match {
+      (result match {
```
jasongilanfarr: could be easier to assign a val to this and call onComplete on that.

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -57,14 +59,21 @@ private[health] class HealthCheckActor(
     )
   }
 
-  def purgeStatusOfDoneInstances(): Unit = {
+  def updateInstaces(): Unit = {
```
aquamatthias: s/updateInstaces/updateInstances

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
janisz: Fixed

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
janisz: Fixed. Also extracted methods to keep `receive` small and clean.

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
janisz: Fixed

janisz:
@aquamatthias I fixed issues and cleanup `receive` by extracting methods. Should I squash commits?

aquamatthias:
@janisz The CI reports an error on `mesosphere.marathon.core.health.impl.MarathonHealthCheckManagerTest.Update`

With this stack trace:
```
health2.lastFailure.isDefined was false
Stacktrace
      org.scalatest.exceptions.TestFailedException: health2.lastFailure.isDefined was false
      at org.scalatest.Assertions$class.newAssertionFailedException(Assertions.scala:528)
      at mesosphere.marathon.core.health.impl.MarathonHealthCheckManagerTest.newAssertionFailedException(MarathonHealthCheckManagerTest.scala:34)
      at org.scalatest.Assertions$AssertionsHelper.macroAssert(Assertions.scala:501)
```

I retested it on my machine and the test was green. 
Can you have a look?

janisz:
I can't reproduce the problem on my local machine too. But I think I know why it occures.
```scala
    EventFilter.info(start = "Received health result for app", occurrences = 1).intercept {
      hcManager.update(taskStatus.toBuilder.setHealthy(false).build, app.version)
    }

    val Seq(health2) = hcManager.status(appId, instanceId).futureValue
```
It works like this:

1. Manager get unhealthy task status and schedule notification to actor.
2. When actor get health result it ask for instnaces and shedule notification when Instance health is ready
3. When actor gets instance health it updates it state

Previously it worked like this:

1. Manager get unhealthy task status and schedule notification to actor.
2. When actor get health result it ask for instance (sync) and updates it state

By adding additional step in processing that is triggered asynchronously there is a chance to ask for instance status before status is updated.

I thinkl this tests should wait for `InstanceHealthChanged` event and then perform check.

janisz:
I can reproduce the problem. I run tests in a loop and I have 9/42 (21%) fail rate.

janisz:
@aquamatthias Fixed by moving log message.

jeschkies:
Merged with https://github.com/mesosphere/marathon/commit/43c4c39c64371e40ad66895ecbaed5c7edf88103

@janisz Sorry. I've squashed it locally and it lost your authorship :/.

PR #4800, opened 2016-12-07
Use async instance tracker methods in TaskKiller

janisz:
Related to: #3031 #4693

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
+++ b/src/main/scala/mesosphere/marathon/api/TaskKiller.scala
@@ -13,7 +13,8 @@ import mesosphere.marathon.upgrade.DeploymentPlan
 import org.slf4j.LoggerFactory
 
 import scala.async.Async.{ async, await }
-import scala.concurrent.{ ExecutionContext, Future }
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
 import scala.util.control.NonFatal
 
 class TaskKiller @Inject() (
@@ -36,9 +37,6 @@ class TaskKiller @Inject() (
     result(groupManager.runSpec(runSpecId)) match {
       case Some(runSpec) =>
         checkAuthorization(UpdateRunSpec, runSpec)
-
-        // TODO: We probably want to pass the execution context as an implcit.
-        import scala.concurrent.ExecutionContext.Implicits.global
         async { // linter:ignore:UnnecessaryElseBranch
           val allTasks = await(instanceTracker.specInstances(runSpecId))
           val foundTasks = findToKill(allTasks)
@@ -57,7 +55,7 @@ class TaskKiller @Inject() (
     }
   }
 
-  private[this] def expunge(tasks: Seq[Instance])(implicit ec: ExecutionContext): Future[Unit] = {
+  private[this] def expunge(tasks: Seq[Instance]): Future[Unit] = {
     // Note: We process all instances sequentially.
 
     tasks.foldLeft(Future.successful(())) { (resultSoFar, nextInstance) =>
@@ -71,13 +69,18 @@ class TaskKiller @Inject() (
     }
   }
 
+  @SuppressWarnings(Array("all")) // async/await
   def killAndScale(
     appId: PathId,
     findToKill: (Seq[Instance] => Seq[Instance]),
-    force: Boolean)(implicit identity: Identity): Future[DeploymentPlan] = {
-    killAndScale(Map(appId -> findToKill(instanceTracker.specInstancesLaunchedSync(appId))), force)
+    force: Boolean)(implicit identity: Identity): Future[DeploymentPlan] = async {
+    val instances = await(instanceTracker.specInstances(appId))
+    val launchedInstances = instances.filter(_.isLaunched)
+    val instancesToKill = findToKill(launchedInstances)
+    await(killAndScale(Map(appId -> instancesToKill), force))
   }
 
+  @SuppressWarnings(Array("all")) // async/await
   def killAndScale(
     appTasks: Map[PathId, Seq[Instance]],
     force: Boolean)(implicit identity: Identity): Future[DeploymentPlan] = {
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4800" target="_blank">ReviewNinja

jasongilanfarr:
ok to test

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -96,8 +98,11 @@ class TaskKiller @Inject() (
       toKill = appTasks
     )
 
-    appTasks.keys.find(id => !instanceTracker.hasSpecInstancesSync(id))
-      .map(id => Future.failed(PathNotFoundException(id)))
+    Future.traverse(appTasks.keys){ id =>
```
jasongilanfarr: Can you convert to async/await for all future based code? e.g. you should rarely ever use `Future.traverse` or any monad method on a future.

janisz: Fixed

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -95,8 +96,16 @@ class TaskKiller @Inject() (
...
-      .getOrElse(killTasks)
+    val futureIdToState = Future.sequence(appTasks.keys.map(id => async {
+      id -> await(instanceTracker.hasSpecInstances(id))
+    }))
+
+    async {
+      val idToState = await(futureIdToState)
+      val deployment = idToState.find(!_._2)
+        .map(id => Future.failed(PathNotFoundException(id._1)))
+        .getOrElse(killTasks)
```
jeschkies: Hm, we should actually collect all cases that do not have a spec, right? So far we just find the first.

```scala

val idHasSpec = Future.sequence(appTasks.keys.map(id => async {
  id -> await(instanceTracker.hasSpecInstances(id))
}))

async {
  val hasSpec = await(idHasSepc).groupBy(_.2)
  hasSpec(false) match {
    case Seq(head, tail @ _*) =>
      val (id, _) = head
      throw new PathNotFoundException(id)
    case Nil => await(killTasks)
  }
}
```

The exception does not take multiple ids. That could be changed in another diff.

janisz: Fixed

janisz:
What tests failed? It's passing on my machine and travis https://travis-ci.org/janisz/marathon/builds/187161861

jeschkies:
@janisz `sbt scapegoat` is failing. See [these logs](http://pastebin.com/gLuhaxiF). Would you mind fixing the reported issues?

janisz:
@jeschkies fixed

janisz:
@jasongilanfarr would you like to rereview? I think I fixed all issues.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -75,9 +73,14 @@ class TaskKiller @Inject() (
     appId: PathId,
     findToKill: (Seq[Instance] => Seq[Instance]),
     force: Boolean)(implicit identity: Identity): Future[DeploymentPlan] = {
-    killAndScale(Map(appId -> findToKill(instanceTracker.specInstancesLaunchedSync(appId))), force)
+    instanceTracker.specInstances(appId).flatMap { instances =>
```
jasongilanfarr: no async/await?

janisz: Fixed, thanks for pointing out.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -100,8 +103,15 @@ class TaskKiller @Inject() (
       toKill = appTasks
     )
 
-    appTasks.keys.find(id => !instanceTracker.hasSpecInstancesSync(id))
-      .map(id => Future.failed(PathNotFoundException(id)))
-      .getOrElse(killTasks)
+    val idHasSpec = Future.sequence(appTasks.keys.map(id => async {
```
aquamatthias: Instead of calling `instanceTracker.hasSpecInstances` for every appId, you could get all instances by spec once:
```scala
val allInstances = await(instanceTracker.instancesBySpec())
val (existing, nonExisting) = appTasks.keys.partition(allInstances.hasSpecInstance)
if (nonExisting.nonEmpty) ...
```

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -100,8 +103,15 @@ class TaskKiller @Inject() (
...
+      id -> await(instanceTracker.hasSpecInstances(id))
+    }))
+
+    async {
+      val hasSpec = await(idHasSpec).groupBy(_._2)
```
aquamatthias: I am not a big fan of tuple property syntax (`_2`).
If you use partition like suggested above, you don't need this.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
janisz: I tried to keep old logic but you suggestion makes code cleaner. Fixed it. Thanks

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -100,8 +103,12 @@ class TaskKiller @Inject() (
       toKill = appTasks
     )
 
-    appTasks.keys.find(id => !instanceTracker.hasSpecInstancesSync(id))
-      .map(id => Future.failed(PathNotFoundException(id)))
-      .getOrElse(killTasks)
+    async {
+      val allInstances = await(instanceTracker.instancesBySpec()).instancesMap
+      val (_, nonExisting) = appTasks.keys.partition(allInstances.contains)
```
aquamatthias: If you don't need the existing ones, you don't need `partition`.
Please use `filter` or `find`.
Sorry if I moved you into that direction.

janisz: Fixed

PR #4804, opened 2016-12-08
Workaround added concerning the Docker issue with external volumes.

ArmandGrillet:
The solution is to use a Mesos container with a Docker image, it gives
the ability to use a Docker image while preventing the automatic
creation of anonymous volumes.


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4804" target="_blank">ReviewNinja

lloesche:
I think I would remove the `thanks to the Mesos containerizer` part and just leave the fact that the issue doesn't arise using the Mesos containerizer.
I love the Mesos containerizer as much as the next guy but given it's our own documentation it might sound a bit pompous. 😄 

Other than that 👍  :shipit:

aquamatthias:
Thanks @ArmandGrillet. Please change the sentence according to Lukas suggestion.

aquamatthias:
@ArmandGrillet Thanks!

PR #4812, opened 2016-12-08
Properly mark health check props as optional

### LINKED ISSUE #4811: Some properties aren't properly marked as optional in RAML
Some of the health check properties as well as potentially others aren't properly marked as optional values. This leads to confusion as they aren't required and to validation issues in the DC/OS Service UI which is using RAML to validate app definitions.

**Example**
```
 maxConsecutiveFailures:
        type: integer
        format: int32
        description: |
          Number of consecutive health check failures after which the
          unhealthy task should be killed.
        minimum: 0
        default: 3
```
_https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/healthCheck.raml#L103_

`maxConsecutiveFailures` has a default values and is as thus implicitly optional but is not marked as an optional value.

orlandohohmeier:
Mark implicit optional props as optional to avoid confusion and fix related validation errors.

Closes #4811

/cc @unterstein

### DIFF
--- a/docs/docs/rest-api/public/api/v2/types/healthCheck.raml
+++ b/docs/docs/rest-api/public/api/v2/types/healthCheck.raml
@@ -82,7 +82,7 @@ types:
     type: object
     properties:
       command?: CommandCheck
-      gracePeriodSeconds:
+      gracePeriodSeconds?:
         type: integer
         format: int32
         description: |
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4812" target="_blank">ReviewNinja

jdef:
lgtm

PR #4820, opened 2016-12-09
Change host port type  to ephemeral port number

### LINKED ISSUE #4817: Wrong Endpoint.hostPort type in the network.raml
On the `Endpoint.hostPort` type, there is a `minimum: 0` definition as seen below:

```yml
      hostPort?:
        type: numbers.Port
        description: |
          Mapped port on the host.
          All host ports are allocated from resource offers.
        usage: Specify 0 to tell Marathon to dynamically allocate a host port.
        minimum: 0
```

Unfortunately, this does not validate correctly with the RAML parser we use in the UI, since the `minimum` is not considered when already using a custom numeric type. 

However, I noticed that there is already the `EphemeralPort` in the `numberTypes.raml` that provides the minimum 0, so you could use:

```yml
      hostPort?:
        type: numbers.EphemeralPort
        description: |
          Mapped port on the host.
          All host ports are allocated from resource offers.
        usage: Specify 0 to tell Marathon to dynamically allocate a host port.
```

orlandohohmeier:
Adjust the network host port type from `numbers.Port` to `numbers.EphemeralPort` to properly document the allowed value range. This change aligns `network` with `appContainer`.

Fixes #4817

### DIFF
--- a/docs/docs/rest-api/public/api/v2/types/network.raml
+++ b/docs/docs/rest-api/public/api/v2/types/network.raml
@@ -20,12 +20,11 @@ types:
           The container port that a task's process is actually listening on.
           Required if the network mode is container
       hostPort?:
-        type: numbers.Port
+        type: numbers.EphemeralPort
         description: |
           Mapped port on the host.
           All host ports are allocated from resource offers.
         usage: Specify 0 to tell Marathon to dynamically allocate a host port.
-        minimum: 0
       protocol?:
         type: array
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4820" target="_blank">ReviewNinja

jdef:
lgtm

PR #4865, opened 2016-12-16
Include TASK_FINISHED as failure state when upgrading

fengyehong:
If we upgrade an app with health check, and newly launched task
finished before passing health check, the current task replace actor
ignore this TASK_FINISHED message and run into a strange state:
  1. it waits for task status update event, but it won't get any;
  2. it won't try to launch a new task.

### DIFF
--- a/src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
+++ b/src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
@@ -69,13 +69,13 @@ class TaskReplaceActor(
 
   def replaceBehavior: Receive = {
     // New task failed to start, restart it
-    case MesosStatusUpdateEvent(slaveId, taskId, FailedToStart(_), _, `appId`, _, _, _, `versionString`, _, _) if !oldTaskIds(taskId) => // scalastyle:ignore line.size.limit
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4865" target="_blank">ReviewNinja

fengyehong:
Issue #4866

src/main/scala/mesosphere/marathon/upgrade/TaskReplaceActor.scala
```diff
@@ -138,7 +138,7 @@ object TaskReplaceActor {
   private[this] val log = LoggerFactory.getLogger(getClass)
 
   val KillComplete = "^TASK_(ERROR|FAILED|FINISHED|LOST|KILLED)$".r
```
zen-dog: Since `KillComplete` and `FailedToStart` are now equal, I would use one and call it smth. like `IsTerminal`.

fengyehong:
updated, replace FailedToStart and KillComplete with IsTerminal

aquamatthias:
ok to test

fengyehong:
test this please

fengyehong:
retest this please

PR #4912, opened 2016-12-28
Releases/1.3 async task tracker

janisz:
It looks like Marathon 1.4 needs more time to be production ready. I prepared a PR that reflects critical changes made for 1.4 `InstanceTracker` and port it to 1.3 `TaskTracker`. With this changes threads count is more stable and keeps below 500.

Below you can see comparison between 1.3.6 before and after applying this patch (11:30 to 13:15).

async_vs_blocking_task_tracker

### DIFF
--- a/bin/marathon-framework
+++ b/bin/marathon-framework
@@ -109,10 +109,10 @@ function load_options_and_log {
         clean_cmd+=( "${i}" )
       fi
     done
-    cmd_to_env $clean_cmd
+    cmd_to_env "${clean_cmd[@]}"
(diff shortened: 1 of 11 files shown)


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4912" target="_blank">ReviewNinja

aquamatthias:
ok to test

jeschkies:
Integration tests do pass but the job fails because it tries to execute `sbt scapegoat`.

PR #4927, opened 2017-01-06
Parse JSON event only once before broadcast

janisz:
Cherry-picked: #4771
Current implementation parses events for every callback.
This patch make parsing happen only once for event.


review-ninja:
<a href="https://app.review.ninja/mesosphere/marathon/pull/4927" target="_blank">ReviewNinja

PR #4936, opened 2017-01-11
Fixes #4637 Allow filtering SSE events by types

### LINKED ISSUE #4637: Add filter option for event types on GET /v2/events endpoint

janisz:
Allow recievieng only specific events by providing their types
in `event_type` query params. Old behavaiour remains unchanged.

Cherry-picked: a95701ed755603d620e635ddacb0ff95d7fcf9bd (#4671)


aquamatthias:
ok to test

aquamatthias:
@janisz Thanks!

PR #4937, opened 2017-01-11
peg doc updates to release branch

sascala:
I believe this is the only change we need to make to get the docs published from the release branch instead of master.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -78,3 +78,4 @@ following the instructions.
         $ cd /path/to/marathon-gh-pages
         $ git commit . -m "Syncing docs with release branch"
         $ git push
+
--- a/docs/README.md
+++ b/docs/README.md
@@ -67,14 +67,14 @@ following the instructions.
         $ cd /path/to/marathon-gh-pages
         $ git checkout gh-pages
 
-3. Copy the contents of the "docs" directory in master to the root of your
+3. Check out the appropriate release branch, then copy the contents of the "docs" directory in master to the root of your
    marathon-gh-pages directory.
-
         $ cd /path/to/marathon
+        $ git checkout releases/1.x
         $ cp -r docs/** ../marathon-gh-pages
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -67,14 +67,14 @@ following the instructions.
         $ cd /path/to/marathon-gh-pages
         $ git checkout gh-pages
 
-3. Copy the contents of the "docs" directory in master to the root of your
+3. Check out the appropriate release branch, then copy the contents of the "docs" directory in master to the root of your
    marathon-gh-pages directory.
-
```
janisz: This line is required to to apply code formatting for the following lines.

docs/README.md
```diff
@@ -67,14 +67,14 @@ following the instructions.
         $ cd /path/to/marathon-gh-pages
         $ git checkout gh-pages
 
-3. Copy the contents of the "docs" directory in master to the root of your
+3. Check out the appropriate release branch, then copy the contents of the "docs" directory in master to the root of your
```
janisz: Step two (2) is redundant after this change. I think this point should be splited and replace step two.

aquamatthias: @janisz why is step 2 redundant? gh-pages still lives on a separate branch, but we sync from the release branch, instead of master.

docs/README.md
aquamatthias: Indeed - formatting is wrong now.

docs/README.md
janisz: My mistake I didn't read step 1 where you make separate copy of the Marathon.

docs/README.md
sascala: did that last commit (where I added back the newline) fix the formatting?

janisz: Nope

sascala: Ah indeed. Fixed now. I shouldn't touch documents before coffee.

PR #4986, opened 2017-01-18
Fixes #4948 | Lazy event parsing

### LINKED ISSUE #4948: Lazy event parsing
Right now every event is parsed (even if nobody will reciieve it) take a look at [HttpEventStreamHandleActor](https://github.com/mesosphere/marathon/blob/6246d55cb95267664b2ffafa1e911a9f6a9e1e59/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamHandleActor.scala#L63). We can solve this by creating case class with `lazy toJson` filed that will parse event and memorize value for next subscriber.
https://github.com/mesosphere/marathon/issues/3154

janisz:
Change SSE event stream handler to use events that are
lazy parsed to JSON. This will reduce CPU time when most events
are filtered and/or when more then one subscriper is connected.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/event/Events.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/Events.scala
@@ -1,6 +1,8 @@
 package mesosphere.marathon.core.event
 
 import akka.event.EventStream
+import com.fasterxml.jackson.annotation.JsonIgnore
+import mesosphere.marathon.api.v2.json.Formats.eventToJson
 import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.health.HealthCheck
 import mesosphere.marathon.core.instance.update.InstanceChange
@@ -9,12 +11,15 @@ import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.state.{ AppDefinition, PathId, Timestamp }
 import mesosphere.marathon.upgrade.{ DeploymentPlan, DeploymentStep }
 import org.apache.mesos.{ Protos => Mesos }
+import play.api.libs.json.Json
 
 import scala.collection.immutable.Seq
 
 sealed trait MarathonEvent {
   val eventType: String
   val timestamp: String
+  @JsonIgnore
+  lazy val jsonString: String = Json.stringify(eventToJson(this))
 }
 
 // api
--- a/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
@@ -3,6 +3,7 @@ package mesosphere.marathon.core.event.impl.stream
 import akka.actor._
 import com.google.inject.Inject
 import mesosphere.marathon.core.election.{ ElectionService, LocalLeadershipEvent }
+import mesosphere.marathon.core.event.MarathonEvent
 import mesosphere.marathon.core.event.impl.stream.HttpEventStreamActor._
 import mesosphere.marathon.metrics.Metrics.AtomicIntGauge
 import mesosphere.marathon.metrics.{ MetricPrefixes, Metrics }
@@ -16,7 +17,7 @@ import scala.util.Try
 trait HttpEventStreamHandle {
   def id: String
   def remoteAddress: String
-  def sendEvent(event: String, message: String): Unit
+  def sendEvent(event: MarathonEvent): Unit
   def close(): Unit
 }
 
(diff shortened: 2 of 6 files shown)


src/main/scala/mesosphere/marathon/core/event/Events.scala
```diff
@@ -9,9 +10,20 @@ import mesosphere.marathon.core.instance.Instance
...
+}
+
+case class JsonEvent(event: MarathonEvent) extends Event {
+  val eventType: String = event.eventType
+  lazy val message: String = Json.stringify(eventToJson(event))
+}
+
 sealed trait MarathonEvent {
   val eventType: String
   val timestamp: String
```
jasongilanfarr: Could have all of these have `lazy val jsonString....`

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
```diff
@@ -16,7 +17,7 @@ import scala.util.Try
 trait HttpEventStreamHandle {
   def id: String
   def remoteAddress: String
-  def sendEvent(event: String, message: String): Unit
+  def sendEvent(event: Event): Unit
```
jasongilanfarr: And then force these to be MarathonEvent

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamHandleActor.scala
```diff
@@ -57,10 +55,10 @@ class HttpEventStreamHandleActor(
   private[this] def sendAllMessages(): Unit = {
     if (outstanding.nonEmpty) {
       val toSend = outstanding.reverse
-      outstanding = List.empty[MarathonEvent]
+      outstanding = List.empty[JsonEvent]
```
jasongilanfarr: Prefer Seq.empty over List.empty

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamHandleActor.scala
```diff
@@ -5,11 +5,9 @@ import java.io.EOFException
 import akka.actor.{ Actor, ActorLogging, Status }
```
jasongilanfarr: Can you change the package into a double declaration?

e.g.
```
package mesosphere.marathon
package core.event.impl.stream```

File: src/main/scala/mesosphere/marathon/core/event/Events.scala [hunk 5]
```diff
@@ -1,6 +1,7 @@
 package mesosphere.marathon.core.event
```
jasongilanfarr: Can you double package this too?

jasongilanfarr:
ok to test

src/main/scala/mesosphere/marathon/core/event/Events.scala
janisz: I'm sorry I don't understand. If you are referring to put `lazy val jsonString` into `MarathonEvent` Then we need to ignore this filed during serialization. I created new class to distinct data from their view. We use JSON now but we can use Protobuf or something else in the future. Should I move `JSONEvent` to SSE Actor?

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamHandleActor.scala
janisz: Done

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamHandleActor.scala
janisz: Done

janisz:
Rebased

src/main/scala/mesosphere/marathon/core/event/Events.scala
jasongilanfarr: sorry for forgetting about this... since we control the serialization/deserialization of these, we could have the jsonString as a member and not serialize the field.

janisz: Fixed

unterstein:
Thanks @janisz !

```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__          Shippin'  
         __|_o_o_o\__         it!
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^
```

Issue #4978, opened 2017-01-18
App Container is not validate properly


janisz:
Trying to create following app:
```json
{
  "id": "docker-compose-demo",
  "cmd": "echo hello world",
  "container": {}
}
```
results with:
```
HTTP/1.1 500 Internal Server Error

{"message":"docker must not be empty"}%  
```

Should be
```
 HTTP/1.1 422 

{"message":"Object is not valid","details":...
```
Reported in: https://github.com/mesosphere/marathon/issues/4945

meichstedt:
Affects 1.4 but I don't know whether that'll land in 1.4

unterstein:
PR @ https://github.com/mesosphere/marathon/pull/4989

PR #4989, opened 2017-01-19
Bugfix #4978 |  AppDefinition.Container validation

janisz:


src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -349,7 +349,8 @@ trait ContainerFormats {
           }
         case _ =>
           if (`type` == ContainerInfo.Type.DOCKER) {
-            throw SerializationFailedException("docker must not be empty")
+            //TODO(janisz): Hack!!! Think about explicit way to validate exclusive container types.
```
jasongilanfarr: Why not just throw JsResultException with the correct parameters?

jasongilanfarr:
ok to test

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
janisz: That was my first attempt (after trying to fix it with Format validation). [`JsResultException`](https://www.playframework.com/documentation/2.5.x/api/scala/index.html#play.api.libs.json.JsResultException) has constructor that will require much more code becouse it was not meant to use outside of the parsing framework. With this approach we reuse validation from `DockerContainerParameters` so it it get changed (e.g. new required field added) no more changes in code will be required.

unterstein: What about 
```
import play.api.data.validation.ValidationError
throw JsResultException(Seq((JsPath() \ "image", Seq(ValidationError("error.path.missing")))))
```

Are you sure that `/image` is the correct path for the error message? Wondering because the error occurs because the path `/container/docker` is missing when `type="DOCKER"`.

janisz: @unterstein You are right. I think we can stick with your solution

unterstein:
retest this please

unterstein:
^^ just a trigger for CI, don`t worry :)

unterstein:
retest this please

janisz:
Rebased

unterstein:
retest this please

unterstein:
retest this please

unterstein:
Thanks! Merged master and will squash merge after tests

unterstein:
Thanks @janisz !

```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__          Shippin'  
         __|_o_o_o\__         it!
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^
```

Issue #4996, opened 2017-01-19
Generated Rest API Docs Broken


drewkerrigan:
https://mesosphere.github.io/marathon/docs/rest-api.html

If you scroll down to "DELETE /v2/apps/{appId}/tasks/{taskId}” - You’ll see where the html and markdown are shown instead of correctly generated html.

branden:
Just ran into this and was about to report it. Looks like markdown -> HTML rendering stopped in the middle of the document. Screenshot of page source:

![screenshot 2017-01-19 15 27 29](https://cloud.githubusercontent.com/assets/42242/22129625/fe07ea10-de5b-11e6-9d0a-49b57e6674d0.png)

drewkerrigan:
Yeah I can actually see the culprit right in your screenshot, `<code>wipe=false...` is an unclosed `code` tag which would result in the behavior we're seeing.

aquamatthias:
Thanks for the fix. Republished the site.

PR #4997, opened 2017-01-20
Fix code tag in rest API docs

drewkerrigan:
This addresses the problem in the source markdown file, but the generated html still needs to be... generated. The `gh-pages` branch would also need it's own PR most likely, but I haven't cut a release of the marathon docs before. Suggestions are welcome.


drewkerrigan:
Also to link - here's the related issue: #4996

PR #5027, opened 2017-01-27
Fixes #5024 by using the correct validator for validating app dependencies.

### LINKED ISSUE #5024: Dependencies validation is too restrictive
Its actually modeled on group pathing like app ids, but it should be much more permissive as we can't do dependencies that cross groups (or in some cases, even child folders).

This is a major regression introduced between 0.15 to 1.1

aquamatthias:


unterstein:
retest this please

PR #5034, opened 2017-01-28
Fixes #5006 by logging if migration is still in progress

### LINKED ISSUE #5006: Better insights in the log during a migration
When Marathon is performing a Migration, it logs to the log file:
- if the Migration starts
- if the Migration ends
More insights into that process would help a lot.

Especially in big/long migrations, Marathon appears to be "stuck", but is not.

meln1k:
Implementation of logging if migration is still in progress.


meln1k:
I'll create another PR.

PR #5061, opened 2017-02-01
Pipelines/karsten/fail pipeline

jeschkies:


Jenkinsfile
```diff
@@ -68,7 +68,7 @@ node('JenkinsMarathonCI-Debian8') {
           try {
               timeout(time: 20, unit: 'MINUTES') {
                 withEnv(['RUN_DOCKER_INTEGRATION_TESTS=true', 'RUN_MESOS_INTEGRATION_TESTS=true']) {
-                   sh "sudo -E sbt -Dsbt.log.format=false coverage test coverageReport"
+                   sh "set -e -o pipefail; sudo -E sbt -Dsbt.log.format=false coverage test coverageReport | head -c 268435456"
```
timcharper: I think we need the `2>&1`, still. (if log output is going to STDERR then the head command will not apply)

Jenkinsfile
```diff
@@ -80,7 +80,7 @@ node('JenkinsMarathonCI-Debian8') {
           try {
               timeout(time: 20, unit: 'MINUTES') {
                 withEnv(['RUN_DOCKER_INTEGRATION_TESTS=true', 'RUN_MESOS_INTEGRATION_TESTS=true']) {
-                   sh "sudo -E sbt -Dsbt.log.format=false coverage integration:test mesos-simulation/integration:test coverageReport"
+                   sh "set -e -o pipefail; sudo -E sbt -Dsbt.log.format=false coverage integration:test mesos-simulation/integration:test coverageReport | head -c 268435456"
```
timcharper: ^

jeschkies:
Since the trimming did not help on the stable tests we don't need it on our master pipeline.

PR #5094, opened 2017-02-03
Initial stab at making deployment plans cheaper. Back port of https://phabricator.mesosphere.com/D476

unterstein:
When marathon gets elected as leader and a lot of deployments were running before, marathon will need to load all of them and calculate all of the dependent applications. The current calculation is not really fast (all dependencies for all apps are calculated) and therefore a lot of memory is consumed and `on_elected_prepare_timeout` is exceeded. 
These patches will increase performance and will allow marathon to load deployments faster.


aquamatthias:
DeploymentPlanTest fails - please have a look

unterstein:
retest this please

PR #5098, opened 2017-02-03
upgrade akka

### LINKED ISSUE #5097: Akka scheduler stops workings after 248.5 days in 1.1.6 and earlier
The underlying issue is this: https://github.com/akka/akka/issues/20424

Marathon 1.3.x does not contain this bug as it has a later version of Akka

timcharper:
Fixes #5097

### DIFF
--- a/project/build.scala
+++ b/project/build.scala
@@ -284,7 +284,7 @@ object Dependency {
     val Guava = "18.0"
     val JacksonCCM = "0.1.2"
     val MesosUtils = "0.28.0"
-    val Akka = "2.3.9"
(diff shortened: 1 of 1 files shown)


nevermosby:
This patch does not fix the issue #5097, because akka v2.3.15 actually does not include the fix(https://github.com/akka/akka/pull/20435), you will have that fix if you are using v2.4.5 akka(Find the details here:https://akka.io/blog/news/2016/05/17/akka-2.4.5-released)
Thus, Using Marathon v1.1.7 will still have the time issue.

PR #5112, opened 2017-02-06
Do not update apt-get.

jeschkies:
Summary:
`apt-get update` fails sometimes. Let's remove it.

Test Plan:
see pipeline

Reviewers:
timcharper, aquamatthias

Subscribers: jenkins, marathon-team


unterstein:
If `apt-get update` is not needed to install `curl` ... ok ;)

jeschkies:
I've created an issue to track the root cause which is not fixed by this change https://github.com/mesosphere/marathon/issues/5113.

Issue #4802, opened 2016-12-07
Marathon has a cap at 2664 apps.


jasongilanfarr:

jasongilanfarr:
Also of interest, we cap at 115,685 instances for a single app too.

janisz:
What happened when you schedule more? Any hypothesis? ZK limits? Actors/threads limits?

aquamatthias:
Please define your setup + steps to reproduce, as well as your initial insights.

jasongilanfarr:
This is in the shakedown suite as well. All I'm doing is creating stupidly simple apps in a loop and checking separately the count of items in /v2/tasks and /v2/apps. My current suspect is launch queue and possibly instance tracker. Once we hit this magical cap, everything timesout with an ask timeout.

@janisz it's unlikely it's zk, when I tested the new persistence layer I could get to 100k apps.

jasongilanfarr:
Did a bunch more investigation trying to find the root cause and split off #4813 as a blocker. We likely want to just revert, but moved the discussion for that over there.

test script:
```
for i in {1..5000}; do
http POST :8888/v2/apps <<EOF
{
  "id" : "/simple-app-$i",
  "cmd": "sleep 1000",
  "instances": 1,
  "cpus": 0.01,
  "mem": 1,
  "disk": 0
}
EOF
done
```


I created https://phabricator.mesosphere.com/D305 with a bunch of random stuff that I was using for investigation. This is just my notes/temporary patches but had little effective change (the variance without nested groups was roughly +/- 300)

I'm using the simulator and I've hacked it a little to resemble a 100 node cluster with basically unlimited resources and made it send offers every 5 seconds (instead of every second) as this will more closely resemble what mesos actually does. I also made it respond to LaunchRequests faster and removed all the probabilities in there.

In addition, I started messing around with some defaults - some I think are probably actually good, like making LaunchQueue ask timeout be 10s instead of 1s. This didn't have a big effect. Thinking maxTasksPerOffer may have something to do with the problem revealed very little.

I put Kamon in since we can dive in via JMX to find out about actor queueing/time processing/errors, etc and found it really handy to combine with the profile data.

A few things do stand out:

[...]

PR #5115, opened 2017-02-06
Don't cache `Group.toProto`

timcharper:
It may not have been helping much (no data to support it), but it is
significantly burdening our heap as each DeploymentPlan has a unique
instance of a Group (even if the version is identical, protobuf
deserialization ensures a deep copy, eventaully), which leads to about
double the number of instances in memory.


janisz:
We can get hard data about this change by running test done in https://github.com/mesosphere/marathon/issues/4802 and compare GC stats and heap usage.

PR #5117, opened 2017-02-06
reordered steps

sascala:


jdef:
should this be cherry-picked onto the releases/1.4 branch (since that's
where we'll likely generate the marathon GH pages from)?

PR #5124, opened 2017-02-07
undo def -> lazy val toProto change

timcharper:
Other tests showed this significantly increased the memory overhead for
instances with concurrent deployment plans.


unterstein:
Thanks :)

PR #5143, opened 2017-02-08
app:id parameter cannot be nested

sascala:
https://mesosphere.atlassian.net/browse/DOCS-1636


jdef:
nope, untrue. left a comment on the DCOS ticket

sascala:
Thanks, @jdef

Issue #5009, opened 2017-01-25
[Design Doc] Jobs in Marathon


marcomonaco:

farra:
Will Marathon support Jobs as part of Groups such that with a single Marathon application definition could include a group of both jobs and apps that have dependencies between them? 

Would love to see the design doc when it's available!

jasongilanfarr:
We are still doing a lot of internal planning right now and plan to send
something out with the general 1.5 plan relatively soon.

The TL;DR is that an application (and eventually) pod can have a schedule
associated with it, either "Continuous", "On a Cron Schedule", or "Run on
Demand". We likely won't support any form of dependencies/workflows in the
initial implementation.

farra:
Even simple dependencies would allow this feature to effectively serve the role as Kubernetes' Init Containers: https://kubernetes.io/docs/concepts/abstractions/init-containers/

I'm looking for a similar feature in which I may want one or more jobs to run prior to launching a marathon app. These jobs should run once and should not be scaled with the marathon app.

PR #5148, opened 2017-02-09
Added CronMonitor actor

pfcoperez:
In its current state it starts apps after a given period of time.

### DIFF
--- a/src/main/scala/mesosphere/marathon/cron/CronMonitorActor.scala
+++ b/src/main/scala/mesosphere/marathon/cron/CronMonitorActor.scala
@@ -1,6 +1,6 @@
 package mesosphere.marathon.cron
 
-import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable}
+import akka.actor.{Actor, ActorLogging, ActorRef, Cancellable, Props}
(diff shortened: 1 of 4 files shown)


janisz:
Is it related to https://github.com/mesosphere/marathon/issues/5009?

src/main/scala/mesosphere/marathon/cron/CronMonitorActor.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  private def receiveCommands: Receive = {
+    case PerformCheck =>
+      launchCronApps
+  }
+
+  private def receiveInternalEvents: Receive = {
+    case LaunchPeriodicTasks(taks) =>
+      taks foreach { runSpec =>
+        launchQueue.add(runSpec)
```
pfcoperez: I've just noticed this call blocks in when using `LaunchQueueDelegate` implementation and, therefore it should be performed within a Future. This could be done at `launchCronApps`, that way, the future for the result of the request to the group repository could be composed with it via `flatMap`

aquamatthias:
This is WIP. I will close this PR for now. Please reopen, if it is finished.

PR #5157, opened 2017-02-10
PUT on /v2/apps has a PATCH semantic.

aquamatthias:
[EDIT] Add a query parameter, that allows for proper PUT semantics which will overwrite all values.

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -91,7 +91,7 @@
        You can query the deployments endoint to see the status of the deployment.
     is: [ secured, deployable ]
     queryParameters:
-      updateExisting:
+      partialUpdate:
         required: false
         default: true
         description: |
@@ -262,7 +262,7 @@
          You can query the deployments endoint to see the status of the deployment.
       is: [ secured, deployable ]
       queryParameters:
-        updateExisting:
+        partialUpdate:
           required: false
           default: true
           description: |
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -90,6 +90,16 @@
       Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
        You can query the deployments endoint to see the status of the deployment.
     is: [ secured, deployable ]
+    queryParameters:
+      updateExisting:
+        required: false
+        default: true
+        description: |
+          Without specifying this parameter, this method has a patch like semantic:
+          All values that are not defined in the json, will not change existing values.
+          This was the default behaviour in previous Marathon versions.
+          For backward compatibility, we will not change this behaviour, but let users opt in for a proper PUT.
+          Note: We will change the default behaviour in the next Marathon version to support PATCH and PUT as HTTP methods.
     body:
       application/json:
         example: !include examples/apps_create.json
@@ -251,6 +261,16 @@
         Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
          You can query the deployments endoint to see the status of the deployment.
       is: [ secured, deployable ]
+      queryParameters:
+        updateExisting:
+          required: false
+          default: true
+          description: |
+            Without specifying this parameter, this method has a patch like semantic:
+            All values that are not defined in the json, will not change existing values.
+            This was the default behaviour in previous Marathon versions.
+            For backward compatibility, we will not change this behaviour, but let users opt in for a proper PUT.
+            Note: We will change the default behaviour in the next Marathon version to support PATCH and PUT as HTTP methods.
       body:
         application/json:
           example: !include examples/app.json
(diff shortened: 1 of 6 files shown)


aquamatthias:
cc @orlandohohmeier

nLight:
We're happy with the name, thank you for addressing this!

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -90,6 +90,16 @@
       Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
        You can query the deployments endoint to see the status of the deployment.
     is: [ secured, deployable ]
+    queryParameters:
+      updateExisting:
```
jdef: i'm confused by this parameter name. aren't we updating via PUT or PATCH? how does `updateExisting` imply "I'd like to partially update"?

suggestions: `partialUpdate`, `patchUpdate`, `applyAsPatch`

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -251,6 +261,16 @@
         Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
          You can query the deployments endoint to see the status of the deployment.
       is: [ secured, deployable ]
+      queryParameters:
+        updateExisting:
```
jdef: ditto

jdef:
or else, perhaps invert the meaning parameter: call it `replaceAll` and default it to `false`?

nLight:
We went ahead and created a tag in docker hub v1.4.0-RC6.5157 then I launched is as MoM and pointed my dcos-ui to it instead of the root Marathon. Then I added the param to the API request, I can see the UI sending it but Marathon still replies with the same error.

aquamatthias:
@nLight 
That is a different problem: the problem arises, if the updated appdefinition defines an ipAdress, but no portDefinition. 

- AppDefinition.empty has a default port
- AppDefinition (L92) has a require statement not allowing ipAddress/ports at the same time

@jdef can you have a look?

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -1023,13 +1022,38 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
...
+        |  "container": {
+        |    "type": "DOCKER",
+        |    "docker": {
+        |      "image": "/test:latest",
+        |      "network": "USER"
+        |    }
+        |  },
+        |  "ipAddress": { "name": "dcos" }
+        |}""".stripMargin.getBytes("UTF-8")
+    val appUpdate = appsResource.canonicalAppUpdateFromJson(app.id, body, false)
```
jdef: should we also include a test case where `partialUpdate=true` to demonstrate that it will fail validation?

jdef:
also discussed with @aquamatthias that we should add support for the `PATCH` method so that people can migrate their patch-semantic-dependent code to the `PATCH` API during the 1.4 release cycle so that when we roll out 1.5 we don't break people who are using `PUT` to get patch-like behavior.

(The plan for 1.5 should be to, at least, change the default of `partialUpdate` to `false`)

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -141,20 +143,66 @@ class AppsResource @Inject() (
     }
   }
 
+  /**
+    * Validate and normalize an array of application updates submitted via the REST API. Validation exceptions are not
```
jdef: copy/paste doc error

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -141,20 +143,66 @@ class AppsResource @Inject() (
...
+    if (partialUpdate) {
+      validateOrThrow(Json.parse(body).as[AppUpdate].copy(id = Some(appId)).withCanonizedIds())
+    } else {
+      // this is a complete replacement of the app as we know it, so parse and normalize as if we're dealing
+      // with a brand new app because the rules are different (for example, many fields are non-optional with brand-new apps).
+      // however since this is an update, the user isn't required to specify an ID as part of the definition so we do
+      // some hackery here to pass initial JSON parsing.
+      val jsObj = Json.parse(body).as[JsObject] + ("id" -> Json.toJson(appId.toString))
+      // the version is thrown away in toUpdate so just pass `zero` for now
+      normalizedApp(validateOrThrow(jsObj.as[AppDefinition].copy(id = appId).withCanonizedIds()), Timestamp.zero).toUpdate
```
jdef: we don't actually need to copy the id in this step since we just did it in the previous one

jdef:
overall i'm not super happy w/ the efficiency when `partialUpdate==false` because we basically go from AppDefinition to AppUpdate back to AppDefinition again. but this approach minimizes the changes required for this PR and probably doesn't actually add "too much" overhead for app update operations. if performance concerns crop up in testing we can tackle a larger refactor then

jdef:
`GroupDeployIntegrationTest` seems to flake out when integration testing on CI but runs just fine locally

PR #5179, opened 2017-02-15
Fixes #4901 by removing the short circuit for root group changes.

### LINKED ISSUE #4901: Groups and apps can apparently overlap and it creates weird behaviour
Hi,
In 1.3 I have apps that were deployed using `/v2/apps`and `/v2/groups` (probably using `?force=true`at some point). The issue is that now they are visible at 2 levels when querying `/v2/groups`

```
$ curl -kL -u ${MARATHON_USER}:${MARATHON_PASSWORD} https://my-marathon.mytld/v2/groups \
    | jq '[.groups[].apps[].id], [.apps[].id]'
[
  (...)
  "/group/app1",
  "/group/app2",
  (...)
],
[
  (...)
  "/group/app1",
  "/group/app2",
  (...)
]
```

The issue it produces: now I can't deploy anymore, all my new deployments are apparently "attached" to a deployment_id in which there are app1 and app2 in "Restarting" state (probably an action initiated by the owner). These 2 apps are themselves marked as "deploying".

bug

aquamatthias:
Summary:
PathId.childOf validation yielded true for all paths relative to base root.
Wrong assumption is removed + test case added.

Test Plan:
sbt test

Start Marathon and call this curl:
curl -H "Content-Type: application/json" -X PUT -d '{"id": "/", "apps": [{"id":"/test/hello", "cmd":"while true; do echo hello; sleep 1; done"}]}' http://localhost:8080/v2/groups
Before this change, this json was accepted. After this change this is not valid and gives this result:
```
{
    "details": [
        {
            "errors": [
                "Identifier is not child of /. Hint: use relative paths."
            ],
            "path": "/apps(0)/id"
        }
    ],
    "message": "Object is not valid"
}
```

Migration:
Added a migration to 1.4, that repairs the group hierarchy.

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/PathId.scala
+++ b/src/main/scala/mesosphere/marathon/state/PathId.scala
@@ -139,11 +139,11 @@ object PathId {
 
   /**
     * Make sure that the given path is a child of the defined parent path.
-    * Every root and every relative path can be ignored.
+    * Every relative path can be ignored.
     */
   private def childOf(parent: PathId): Validator[PathId] = {
     isTrue[PathId](s"Identifier is not child of $parent. Hint: use relative paths.") { child =>
-      parent == PathId.empty || !parent.absolute || (parent.absolute && child.canonicalPath(parent).parent == parent)
+      !parent.absolute || (child.canonicalPath(parent).parent == parent)
     }
   }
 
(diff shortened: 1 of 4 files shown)


aquamatthias:
Test migration with: 
- start marathon 1.3
- create app in wrong place:  
```
curl -H "Content-Type: application/json" -X PUT -d '{"id": "/", "apps": [{"id":"/test/hello", "cmd":"while true; do echo hello; sleep 1; done"}]}' http://localhost:8080/v2/groups
```
- start marathon 1.4 with this PR
- get `/v2/groups` and see the app has moved into the correct place.

I did this locally and got this result (remove unnecessary information):
```json
{
    "id": "/",
    "apps": [],
    "dependencies": [],
    "groups": [
        {
            "id": "/test",
            "apps": [
                {
                    "id": "/test/hello",
                    "cmd": "while true; do echo hello; sleep 1; done"
                }
            ]
        }
    ]
}
```

unterstein:
Tested migration manually with curl commands from #4901 -> works as expected.

aquamatthias:
retest this please

PR #5183, opened 2017-02-15
Add support for PATCH updates to apps in 1.4

### LINKED ISSUE #5161: add support for PATCH because we're deprecating partial-update semantics for PUT in 1.4
see https://github.com/mesosphere/marathon/pull/5157#issuecomment-279020931

meichstedt:
Add support for PATCH updates to apps

This introduces support for http `PATCH` requests to one or multiple apps via `/v2/apps` as a replacement to using http `PUT` with the default setting of `partialUpdate: true`. The ability to patch with a `PUT` request will be deprecated in 1.4 and is to be removed in 1.5.

`PATCH` updates will not create an app if the specified id doesn't exist.

Fixes #5161

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -370,8 +370,8 @@
     #
     patch:
       description:
-        Replaces parameters of a running application. This will fail if no application with the given id
-        exists. If there is an application with this id, all running instances get upgraded to the new definition.
+        Replaces parameters of a running application. All running instances get upgraded to the new definition.
+        Any given application ID will be ignored.
 
         Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
          You can query the deployments endoint to see the status of the deployment.
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -158,7 +158,7 @@
     queryParameters:
     body:
       application/json:
-        example: !include examples/apps_create.json
+        example: !include examples/apps_patch.json
         type: app.App[]
     responses:
       409:
@@ -379,7 +379,7 @@
       queryParameters:
       body:
         application/json:
-          example: !include examples/app.json
+          example: !include examples/app_patch.json
           type: app.App
       responses:
         200:
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -138,6 +138,62 @@
                 }
 
   ###
+  # apply a patch update to a list of applications
+  #
+  patch:
+    description:
+      Change multiple existing applications by applying a patch.
+      All instances of these applications get replaced by the new version.
+      The order of dependencies will be applied correctly.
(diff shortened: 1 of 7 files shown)


docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -138,6 +138,62 @@
...
+
+      If you have more complex scenarios with upgrades, use the groups endpoint.
+
+      Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+       You can query the deployments endoint to see the status of the deployment.
+    is: [ secured, deployable ]
+    queryParameters:
+    body:
+      application/json:
+        example: !include examples/apps_create.json
```
jdef: this contains a PATCH example?

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -138,6 +138,62 @@
...
+        type: app.App[]
```
jdef: PATCH only accepts AppUpdate

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -310,6 +366,60 @@
...
+        Replaces parameters of a running application. This will fail if no application with the given id
+        exists. If there is an application with this id, all running instances get upgraded to the new definition.
+
+        Note&#58;  This operation will create a deployment. The operation finishes, if the deployment succeeds.
+         You can query the deployments endoint to see the status of the deployment.
+      is: [ secured, deployable ]
+      queryParameters:
+      body:
+        application/json:
+          example: !include examples/app.json
```
jdef: ditto (for both the example and type)

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -116,6 +116,27 @@ class AppsResourceTest extends MarathonSpec with MarathonActorSupport with Match
     response.getMetadata.containsKey(RestResource.DeploymentHeader) should be(true)
   }
 
+  test("Do partial update with patch methods") {
```
jdef: can we also have a test for PATCH where an app is rejected because it doesn't exist?

aquamatthias:
retest this please

aquamatthias:
retest this please

docs/docs/rest-api/public/api/v2/apps.raml
meichstedt: fixed.

docs/docs/rest-api/public/api/v2/apps.raml
meichstedt: That's unfortunate – I haven't found a way to introduce an AppUpdate type in raml w/o breaking other things. A patch w/o id is allowed and works but the doc will state that an id is required.

jdef:
don't worry about AppUpdate. mistake on my part. the scala/RAML generator
builds an AppUpdate scala class. I don't think we actually want to define
AppUpdate at the RAML level

meichstedt:
retest this please

aquamatthias:
retest this please

aquamatthias:
retest this please

aquamatthias:
retest this please

PR #5199, opened 2017-02-16
specify which params are required/optional

sascala:


jdef:
Sorry but this change only adds to the confusion. Operators (like UNIQUE or LIKE) apply to either a `field` or a `field, value` tuple. The text "This parameter is required" leaves one wondering: which parameter?

Perhaps a better layout for these constraints would be something like:
```markdown
### GROUP_BY operator
**Value**: (optional) integer, for example `"3"`

... some explanation follows ...

### LIKE operator
**Value**: (required) a valid regular expression, for example `".*-foo"`

... some explanation follows ...
```

sascala:
That makes lots of sense, @jdef . Can you point me to where in the code I can find what form the values need to take (e.g., integer)?

sascala:
@jdef That makes a lot of sense -- thanks for chiming in. Can you point me to where I can find the required value type (e.g., integer) where that's not clear fro the context of the doc?

docs/docs/constraints.md
```diff
@@ -28,108 +28,108 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
 
 ## Operators
 
-### UNIQUE operator
+### CLUSTER operator
+**Value** (required): A rack ID, set of rack IDs, or a specific node.
```
aquamatthias: The value of the attribute. This does not have to be that rack id (it is just an example)

docs/docs/constraints.md
```diff
@@ -28,108 +28,108 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
     "cmd": "sleep 60",
     "instances": 3,
-    "constraints": [["rack_id", "CLUSTER", "rack-1"]]
+    "constraints": [["hostname", "CLUSTER", "a.specific.node.com"]]
   }'
 ```
 
-You can also use this attribute to tie an application to a specific node by using the hostname property:
+### LIKE operator
+**Value** (required): A rack ID or set of rack IDs.
```
aquamatthias: A regexp for the value of the attribute. This does not have to be that rack id (it is just an example)

docs/docs/constraints.md
```diff
@@ -28,108 +28,108 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
     "cmd": "sleep 60",
     "instances": 3,
-    "constraints": [["hostname", "CLUSTER", "a.specific.node.com"]]
+    "constraints": [["rack_id", "LIKE", "rack-[1-3]"]]
   }'
 ```
 
-### GROUP_BY operator
+### UNLIKE operator
+**Value** (required): A rack ID or set of rack IDs.
```
aquamatthias: See LIKE

docs/docs/constraints.md
```diff
@@ -28,108 +28,108 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
     "id": "sleep-group-by",
     "cmd": "sleep 60",
     "instances": 3,
-    "constraints": [["rack_id", "GROUP_BY", "3"]]
+    "constraints": [["rack_id", "MAX_PER", "2"]]
   }'
 ```
 
+### UNIQUE operator
+**Value** (optional): An integer, for example `"1"`.
```
aquamatthias: This is always a string not an integer. It defines the value of the attribute, which has to be unique. The value is ignored for `hostname`

docs/docs/constraints.md
jdef: "Non-empty string value."?

sascala: 👍

docs/docs/constraints.md
sascala: 👍

docs/docs/constraints.md
sascala: What would be a good example value here?

And can you clarify what you mean by "this value is ignored for hostname"?

sascala:
Thank you for your comments, @aquamatthias  -- I have 2 small outstanding questions, then I think the doc should be complete.

sascala:
@aquamatthias For the UNIQUE operator, what would be a good example value?

And can you clarify what you mean in your comment by "this value is ignored for hostname"?

aquamatthias:
@sascala
If you specify `hostname UNIQUE`, then any value specified will be ignored.

sascala:
Thank you! If this looks good to you, please merge.

sascala:
@aquamatthias Is there a reason this was closed? Are these clarifications no longer needed?

aquamatthias:
@sascala according to the log, you closed this PR. I reopen it.

sascala:
@aquamatthias I am my own worst enemy, it seems. If this looks good, please merge.

docs/docs/constraints.md
```diff
@@ -28,108 +28,108 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
     "id": "sleep-group-by",
     "cmd": "sleep 60",
     "instances": 3,
-    "constraints": [["rack_id", "GROUP_BY", "3"]]
+    "constraints": [["rack_id", "MAX_PER", "2"]]
   }'
 ```
 
+### UNIQUE operator
+**Value** (optional): A string that defines the value of the attribute. This attribute must be unique. If you specify `hostname:UNIQUE`, any value you define will be ignored.
```
mlunoe: Marathon actually throws an error when a value is provided. So the value is _not_ optional. It is required to not be applied!

sascala:
@jdef or @aquamatthias -- is @mlunoe 's comment a bug to warn a user about or a feature to be documented?

docs/docs/constraints.md
jdef: Correct, values are not allowed for `UNIQUE`

mlunoe: In that sense, I think we should change the wording. I will not be ignored, you have to _not_ provide it.

PR #5210, opened 2017-02-18
Disable shell globbing for start script.

ferringb:
If globbing is enabled, and someone sets MARATHON_MESOS_ROLE='\*', it'll
result in '\*' being expanded.  IE, it'll literally include the files
in the cwd as arguments, rather than just passing *.

Nothing in the script relies on shell globbing, thus disable it, fixing
the problem in the process.

### DIFF
--- a/bin/start
+++ b/bin/start
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -o errexit -o pipefail
+set -o errexit -o pipefail -o noglob
 
 declare -a java_args
 declare -a app_args


jdef:
ok to test

unterstein:
hey @ferringb,

thank you for taking the time to provide this PR and taking care of the start script! We removed the `/bin/start` file a while ago in favor of the sbt native packager, see https://github.com/mesosphere/marathon/commit/94f16adc09dc8ecae03bcf3e0db593f30ecebc2d
The possible previous problems with the globbing should not happen with this packaging.

If you have further questions, please free to re-open this issue.

Thanks
Johannes

Issue #4592, opened 2016-10-14
FetchUri isExtract does not support support query strings


geoffpatehutch:
Line 40 of marathon/src/main/scala/mesosphere/marathon/state/FetchUri.scala checks if a file can be extracted by seeing if it endswith a series of extensions. However, it does not handle the case where a file has a query string (eg SAS token) attached.

Maybe it could split the uri on the '?' delimiter, and then check the extension on the path without the query?

jasongilanfarr:
The proper thing to do would be to parse out as a URL and then check the end of the path.

allquantor:
Hi, I did a fix, here the [PR](https://github.com/mesosphere/marathon/pull/5214).

jdef:
I really dislike that we automatically set the "extract" bit based on
filename. For a couple of reasons:
(1) the need to keep in sync the supported archive file types that Mesos
knows how to extract
(2) there's no way for someone to opt-out of extracting an archive. said
another way, what if I want to upload an archive and NOT have it extracted?

allquantor:
@jdef 
extract could be a function `File => R` e.g if there is no extraction needs it would be `identity`.
We would still have to keep in sync the extraction functions that can be used.

jdef:
My point is that Marathon is trying to outsmart the user here. I think it's
a mistake. The fetch API has an `extract` attribute that a user can set to
`true` if they want Mesos to extract the archive at task launch time. That
Marathon applies additional "magic" (to automatically set the extract bit)
is misguided.

allquantor:
@jdef Now I got you. Sounds reasonable to me. When removing the "magic" it still is worth to catch the error that can occur during the extraction for the sake of user-friendly log messages.

meichstedt:
Note: This issue has been migrated to https://jira.mesosphere.com/browse/MARATHON-1692. For more information see https://groups.google.com/forum/#!topic/marathon-framework/khtvf-ifnp8.

PR #5214, opened 2017-02-19
URIs in FetchUri are able to validate URIs that contain query string

allquantor:
According to the comments in [4592](https://github.com/mesosphere/marathon/issues/4592) `FetchUri` is now mapped to Java `URI` and use path for validation.

Additional:

I thought it is probably a good idea to have `uri` in `FetchUri` types as `Uri`, that imply changes in 
`mesos.marathon.Formats` (implementing additional type classes that parse a `Uri` from Json. However, that would move the possible side effect to `Formats`.  I implemented it just to see how it could look like [here](https://github.com/allquantor/marathon/tree/play/4592)


aquamatthias:
ok to test

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -21,13 +24,22 @@ case class FetchUri(
...
-    outputFile.foreach{ name => builder.setOutputFile(name) }
+    outputFile.foreach { name => builder.setOutputFile(name) }
     builder.build()
   }
 }
 
 object FetchUri {
 
+  lazy val supportedFileTypes = Seq(
+    ".tgz",
```
ichernetsky: Could you please put all the file extensions onto one line? Thank you.

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -44,13 +56,19 @@ object FetchUri {
       outputFile = if (uri.hasOutputFile) Some(uri.getOutputFile) else None
     )
 
+  /**
+    * Method check if the given string is a
+    * valid URI and if the file type in the
+    * URI is matching a supported file type.
+    *
+    * @param uri String that contain an URI.
+    * @return
```
ichernetsky: Could you please document the return value, or drop this line?

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -44,13 +56,19 @@ object FetchUri {
       outputFile = if (uri.hasOutputFile) Some(uri.getOutputFile) else None
     )
 
+  /**
+    * Method check if the given string is a
+    * valid URI and if the file type in the
+    * URI is matching a supported file type.
+    *
+    * @param uri String that contain an URI.
```
ichernetsky: Maybe just `a URI string`?

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -44,13 +56,19 @@ object FetchUri {
       outputFile = if (uri.hasOutputFile) Some(uri.getOutputFile) else None
     )
 
+  /**
+    * Method check if the given string is a
```
ichernetsky: This is way too narrow. Could you please make it wider? Thank you.

unterstein:
CI Here: https://github.com/mesosphere/marathon/pull/5624

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -21,13 +24,15 @@ case class FetchUri(
...
       .setCache(cache)
-    outputFile.foreach{ name => builder.setOutputFile(name) }
+    outputFile.foreach { name => builder.setOutputFile(name) }
     builder.build()
   }
 }
 
 object FetchUri {
 
+  lazy val supportedFileTypes = Seq(".tgz", ".tar.gz", ".tbz2", ".txz", ".tar.xz", ".zip")
```
meichstedt: Is there a reason why `.tar.bz2` is no longer contained in this list?

zen-dog: Does this correspond to what mesos executor actually supports? http://mesos.apache.org/documentation/latest/fetcher/
We should update the list.

unterstein:
Thanks @zen-dog, added list and docs ❤️

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
```diff
@@ -600,12 +600,32 @@ class AppDefinitionTest extends UnitTest with ValidationTestLike {
...
 
+    "Validate URIs with query strings correctly" in {
+      val json =
+        """
+      {
+        "id": "app-with-fetch",
+        "cmd": "brew update",
+        "uris": ["http://example.com/file1.tar.gz?foo=10&bar=meh", "http://example.com/file?foo=10&bar=meh"]
+      }
+      """
```
jeschkies: Why not use strip margin here?

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
```diff
@@ -600,12 +600,32 @@ class AppDefinitionTest extends UnitTest with ValidationTestLike {
...
+
+      val app = fromJson(json)
+
+      assert(app.fetch(0).uri == "http://example.com/file1.tar.gz?foo=10&bar=meh")
```
jeschkies: Please use ScalaTest `should`.

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -44,13 +50,18 @@ object FetchUri {
...
-    uri.endsWith(".tgz") ||
-      uri.endsWith(".tar.gz") ||
-      uri.endsWith(".tbz2") ||
-      uri.endsWith(".tar.bz2") ||
-      uri.endsWith(".txz") ||
-      uri.endsWith(".tar.xz") ||
-      uri.endsWith(".zip")
+    Try(new URI(uri)).map { u =>
+      supportedFileTypes.exists { fileType =>
+        u.getPath.endsWith(fileType)
```
meln1k: What if the provided uri doesn't have a path which ends as a file but actually serves the right file type?

zen-dog: ¯\_(ツ)_/¯

src/main/scala/mesosphere/marathon/state/FetchUri.scala
```diff
@@ -21,13 +24,16 @@ case class FetchUri(
...
       .setCache(cache)
-    outputFile.foreach{ name => builder.setOutputFile(name) }
+    outputFile.foreach { name => builder.setOutputFile(name) }
     builder.build()
   }
 }
 
 object FetchUri {
 
+  // Official extractable extensions of mesos fetcher: http://mesos.apache.org/documentation/latest/fetcher/
```
meichstedt: Thanks for the link!

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
meichstedt: +1

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
meichstedt: +1

src/main/scala/mesosphere/marathon/state/FetchUri.scala
unterstein: Just asked @ArmandGrillet how mesos fetcher would handle this.

unterstein: https://github.com/apache/mesos/blob/cc29c27ee597fbe79206d890d695c0a2e64c46aa/src/launcher/fetcher.cpp#L76-L82

It will not be extracted if extension is not matching, but the task will be started, as shown here:

https://github.com/apache/mesos/blob/cc29c27ee597fbe79206d890d695c0a2e64c46aa/src/launcher/fetcher.cpp#L385-L387

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
unterstein: I would stay consistent to the other tests in this suite and not change style nor enhance scope for this ticket. WDYT @meichstedt @jeschkies ?

unterstein:
Thanks @allquantor!

```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__          Shippin'  
         __|_o_o_o\__         it!
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^
```

PR #5236, opened 2017-02-21
Lockfree ExponentialBackoff

pfcoperez:
This PR provides a lock free implementation of the exponential backlog - used in the election service - which leverages JVM's atomic variables to replace the use of monitor lock synchronization when exponentially increasing the backoff duration value.

Atomically increasing a long variable fits nicely to this kind of functionality and protects against possible deadlocks derived of the unconscious misuse of `synchronize`. An example of this kind the situation has been added to the unit tests also included in these changes.


pfcoperez:
It is somehow unimportant, but the previous implementation could provide durations twice `maximumValue`.

pfcoperez:
@jasongilanfarr Thanks for taking a look! I've to `Retry`. Whereas it provides a whole retrying policy, `Backoff` interface just offer the exponential increase functionality.

Whilst I think it is possible, I would see implementing `Backoff` interface with `Retry` strange compared to straightforwardly using `Retry` in the `ElectionService` logic and thus rid off `Backoff` and its implementation `ExponentialBackoff`.

Another option could be to encapsulate exponential duration increase with jitter implemented at `Retry` and use it for both implement `Backoff` interface and serve `Retry`.

I am perfectly fine with closing the PR if you plan to do something like what I've mention or other approach as that would be a different change. In fact, if I had more free time, I would be more than happy to do it my self :smile: 

This PR was just to raise the attention - and try to quick fix it - on the fact that the current implementation might lead to deadlocks with some uses and provide durations twice as long as the provided maximum duration.

aquamatthias:
Ok to test

pfcoperez:
`akka.pattern` provides `BackoffSupervisor` which _implements the so-called exponential backoff supervision strategy, starting a child actor again when it fails, each time with a growing time delay between restarts._  (http://doc.akka.io/docs/akka/current/scala/general/supervision.html#delayed-restarts-with-the-backoffsupervisor-pattern).

I'd adapt the retry mechanisms to make them use the semantics provided by this supervision strategy  rather than further tweaking a custom implementation. Therefore, I am closing this PR.

PR #5244, opened 2017-02-23
ForceExpunge for missing task is a no-op, rather than a failure

### LINKED ISSUE #5142: Marathon fails to release reserved resources for deleted apps
As of this writing, with the head of e62695b, I've encountered a case in which Marathon cannot release reserved resources because they are identified as spurious, which [results in a force expunge](https://github.com/mesosphere/marathon/blob/e62695bca927dbcb99c1f7b135447d3a5e095079/src/main/scala/mesosphere/marathon/core/matcher/reconcile/impl/OfferMatcherReconciler.scala#L75). Down the line, the [force expunge fails because the app does not exist].

Expunge should probably be made idempotent, such that if it doesn't exist, then the result is successful.

timcharper:
Summary:
If a resident app is deleted before Marathon has a chance to
release the associated resources, then later attempts to release them
fail because because we ForceExpunge the task-ID when processing our
UnreserveAndDestroyVolumes operation.

This change makes ForceExpunge for a non-existent task respond with
no-op, rather than failure.

Fixes #5142

Backport of f441c57

Test Plan: n/a

Reviewers: jasongilanfarr, meichstedt

Subscribers: marathon-team

Differential Revision: https://phabricator.mesosphere.com/D547

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
@@ -46,7 +46,13 @@ private[marathon] class InstanceUpdateOpResolver(
         createInstance(op.instanceId)(updater.reserve(op, clock.now()))
 
       case op: ForceExpunge =>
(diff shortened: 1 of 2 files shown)


timcharper:
Merging as was accepted in phab

PR #5246, opened 2017-02-23
Don't restart tasks on non runtime change.

### LINKED ISSUE #4987: Don't start a new deployment when configurations like "minimumHealthCapacity" are changed only
When I change an app config via web UI, and change only configurations like `updateStrategy.minimumHealthCapacity`, making a rolling update with all other configurations unchanged is pointless.

janisz:
Fixes #4987. After changing non runtime setting, application will not
be restarted. New version won't be created.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/pod/PodDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/core/pod/PodDefinition.scala
@@ -57,9 +57,7 @@ case class PodDefinition(
           containers != to.containers ||
           constraints != to.constraints ||
           podVolumes != to.podVolumes ||
-          networks != to.networks ||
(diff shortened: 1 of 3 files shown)


aquamatthias:
Ok to test

jdef:
my understanding was that spec changes MUST fall into one of two categories: (a) scale-change; (b) upgrade (requires deployment); and that we didn't have an appropriate category for changes that don't actually imply (a) or (b).

janisz:
@jdef I agree. There should be another category for spec changes that do not influence application runtime, therfore do not require restart. What's more, there could be changes that in theory modify runtime but for some tasks restart is not necessary. For example constraints change, already running tasks could meet constraints so should not be touched.

PR #5320, opened 2017-03-14
Make AppStartActorTest stable by aligning with latest master state.

meichstedt:
Fixes MARATHON-7057

Note: This change includes changes from several commits to master that affected AppStartActorTest, e.g. by removing Await.result and changing the inheritance hierarchy. Namely 24c9dea, 9592e53, 272ca0e and e324086. I did not cherry-pick these commits completely because that would have affected too many files and pulled in even more dependent commits, especially ones that affect production code.


meichstedt:
retest this please

PR #5328, opened 2017-03-17
backport AppDeployIntegrationTest fixes and mark stable

timcharper:


src/test/python/app_mock.py
```diff
@@ -0,0 +1,119 @@
...
+            except:
+                logging.exception('Could not handle GET request')
+                raise
+
+        def do_POST(self):
+            try:
+                logging.debug("Got POST request")
+                return self.check_health()
+            except:
+                logging.exception('Could not handle GET request')
```
aquamatthias: s/GET/POST

timcharper: Good point; I'll submit a patch to fix it in master

jeschkies: Please fix this for 1.4 as well :).

timcharper:
@jeschkies I don't think we should fix the log line in this pull request as it is a back port. I think the optimal procedure would be to fix on master, and then back port again. Does this sound good to you?

jeschkies:
@timcharper sure. Go ahead.

Issue #4686, opened 2016-11-23
Create integration tests for SSE /v2/events


janisz:
Currently there are no tests for SSE.

Example test could look like this
```scala
package mesosphere.marathon
package integration
import mesosphere.AkkaIntegrationFunTest
import mesosphere.marathon.integration.setup.EmbeddedMarathonTest

@IntegrationTest
class EventStreamIntegrationTest extends AkkaIntegrationFunTest with EmbeddedMarathonTest {
  before(cleanUp())
  test("adding an event subscriber") {
    When("client read event stream")
    val defaultClientEvents = marathon.events("subscribe_event")
    When("second client read event stream with bad filter")
    val filteredClientEvents = marathon.events("non_existing_event_type")
    Then("at list one event should be sent to first subscribers")
    defaultClientEvents.entityString shouldNot be(Seq.empty)
    And("nothing should be sent to second")
    filteredClientEvents.entityString should be(Seq.empty)
  }
}
```
```scala
def events(eventTypes: String*): Seq[MarathonEvent] = ???
```

Implementation could use https://github.com/hseeberger/akka-sse to parse stream.

jasongilanfarr:
Related #4678

meichstedt:
Note: This issue has been migrated to https://jira.mesosphere.com/browse/MARATHON-2356. For more information see https://groups.google.com/forum/#!topic/marathon-framework/khtvf-ifnp8.

PR #5347, opened 2017-03-31
Add integration test for filtering SSE.

janisz:
Fixes: MARATHON-2356 (#4686)

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/EventsIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/EventsIntegrationTest.scala
@@ -1,0 +1,52 @@
+package mesosphere.marathon
+package integration
+
+import java.util.UUID
+
+import akka.stream.scaladsl.Sink
+import mesosphere.AkkaIntegrationTest
+import mesosphere.marathon.integration.setup._
+import mesosphere.marathon.state.PathId
+import mesosphere.marathon.state.PathId._
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+@IntegrationTest
+class EventsIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonTest {
+
+  before(cleanUp())
+
+  def appId(): PathId = testBasePath / s"app-${UUID.randomUUID}"
+
+  "Filter events" should {
+    "receive only app deployment event" in {
+      Given("a duration limit for connect")
+      val atMost = patienceConfig.timeout.toMillis.millis
+
+      Given("a new event source without filter is connected")
+      val allEvents = marathon.events().futureValue
+        .map(_.eventType)
+        .takeWhile(_ != "deployment_success", inclusive = true)
+        .runWith(Sink.seq)
+
+      Given("a new event source with filter is connected")
+      val filteredEvent: Future[String] = marathon.events("deployment_success").futureValue
+        .map(_.eventType)
+        .runWith(Sink.head)
+
+      When("the app is created")
+      val app = appProxy(appId(), "v1", instances = 1, healthCheck = None)
+      val result = marathon.createAppV2(app)
+
+      And("we wait for deployment")
+      waitForDeployment(result)
+      waitForTasks(app.id.toPath, 1)
+
+      Then("deployment events should be filtered")
+      val allEventsResult = allEvents.futureValue
+      filteredEvent.futureValue shouldBe "deployment_success"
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/marathon/integration/EventsIntegrationTest.scala
```diff
@@ -0,0 +1,63 @@
...
+
+      When("The app is deployed")
+      val result = marathon.createAppV2(app)
+
+      Then("Wait for deployment")
+      waitForDeployment(result)
+      waitForTasks(app.id.toPath, 1)
+
+      Then("Deployment events should be filtered")
+      filteredEventsQueue.toArray should be(Array("deployment_success"))
```
timcharper: Seems like perhaps we could test this using `filteredSource.runWith(Sink.head)`, and not need the concurrent queue (or the potential race).

The non-filtered queue could be tested via `eventsQueue.take(2).runWith(Sink.seq)`; we could make sure that the first event of the non-filtered queue is not the event in the `filteredSource` queue. What do you think?

src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala
```diff
@@ -154,13 +154,18 @@ class MarathonFacade(
...
     import EventUnmarshalling.fromEventStream
     val mapper = new ObjectMapper() with ScalaObjectMapper
     mapper.registerModule(DefaultScalaModule)
 
-    AkkaHttp().singleRequest(AkkaGet(s"$url/v2/events").withHeaders(AkkaAccept(AkkaMediaType.text("event-stream"))))
+    val filterQueryParam = eventsType
+      .map(eventType => s"event_type=$eventType")
+      .mkString("&")
+
+    AkkaHttp().singleRequest(AkkaGet(s"$url/v2/events?$filterQueryParam")
```
timcharper: This seems like it would break if eventType needs to be URL encoded. What if we used Akka http's `Uri` class?

```
Uri.parse(s"$url/v2/events").withQuery(Query("event_type" -> eventType))
```

timcharper:
Thank you @janisz ! Few questions / comments. :)

timcharper:
Hey @janisz - just pinging to see if you saw my review comments? I can just make these changes and land if you're otherwise occupied; I don't want to duplicate efforts.

janisz:
Hi, I'm sorry for the delay. I hope I can fix it over the weekend.

src/test/scala/mesosphere/marathon/integration/EventsIntegrationTest.scala
```diff
@@ -0,0 +1,63 @@
...
+
+      Given("a new app")
+      val app = appProxy(appId(), "v1", instances = 1, healthCheck = None)
+
+      When("The app is deployed")
+      val result = marathon.createAppV2(app)
+
+      Then("Wait for deployment")
+      waitForDeployment(result)
+      waitForTasks(app.id.toPath, 1)
```
janisz: Any idea how to fix it? Events should be published at once so after we wait for deployment to finish and tasks messages should be delivered. There is a chance events will be queued and delivered after but IMO this will be a bug.

src/test/scala/mesosphere/marathon/integration/EventsIntegrationTest.scala
janisz: Can you extend `take(2)` example. How to `take(2)` from `Future[Seq]`

timcharper:
Hi @janisz ! Sorry that it took so long too look at this again! I'll review and try and merge today. Thanks!

timcharper:
@janisz, here's what I had in mind:

https://github.com/mesosphere/marathon/tree/tharper/sse-stream-test

wdyt? Look good to you?

src/test/scala/mesosphere/marathon/integration/EventsIntegrationTest.scala
timcharper: (Yes; see https://github.com/mesosphere/marathon/commit/72346d1c9cf7c1c00e86ed762d1aa73d86d8ee9c#diff-d0660c96bab3390f9064663392c1877aR32)

janisz:
@timcharper It looks goo. Applied your patch and rebased to latest master. Please check it once again.

timcharper:
Thank you! Merged.

PR #5390, opened 2017-05-11
Improve Ext Vols Examples

sascala:
In response to https://jira.mesosphere.com/browse/DOCS-1505


docs/docs/external-volumes.md
```diff
@@ -84,7 +84,7 @@ You can specify an external volume in your Marathon app definition. [Learn more
       "instances": 1,
       "cpus": 0.1,
       "mem": 32,
-      "cmd": "/usr/bin/tail -f /dev/null",
+      "cmd": "mkdir my-test-vol && touch my-test-vol/test.txt && ls my-test-vol/",
```
jdef: suggest `touch my-test-vol/test.txt && ls my-test-vol/; sleep 120`

docs/docs/external-volumes.md
```diff
@@ -133,7 +133,7 @@ Below is a sample app definition that uses a Docker container and specifies firs
...
```
jdef: ditto

docs/docs/external-volumes.md
sascala: @jdef I wasn't able to get the `touch` part of this command to work without first creating the directory. Assuming it was some kind of error on my part -- retesting.

sascala: ....I still get the error, but maybe it's implied that a user already has a directory called `my-test-vol`?

from STDERR:
```
touch: cannot touch 'my-test-vol/test.txt': No such file or directory
```

jdef: the `my-test-vol` directory should already exist because of the external volume declaration. the fact that you get this error is a bit confusing. how are you launching a dcos cluster? (EE or OSS? any other parameter changes?)

sascala: That last time was an EE cluster and just cutting-and-pasting the first example and your suggested `cmd`. But I didn't expect the external volumes part to work, since I didn't do any of the necessary setup (the custom installation with `genconf/rexray.yaml` and so on).

jdef:
If you spin up a cluster via CCM (without using custom templates) then RR
should be installed automatically and the `mkdir` shouldn't be needed. Or,
am I missing something obvious here?

sascala:
@jdef No, you're not missing anything obvious. Though that means my doc is out of date -- is this as of 1.9?

It's great news that they should work out of the box! It's bad news that mine didn't work, though.

docs/docs/external-volumes.md
```diff
@@ -152,7 +91,6 @@ Below is a sample app definition that uses a Docker container and specifies firs
             "mode": "RW"
           },
           {
-            "hostPath": "var-log",
```
jdef: please add back this `hostPath` line

PR #5395, opened 2017-05-17
Ss/docs versioning

sascala:
This PR adds links to docs for previous major marathon releases.


jdef:
I might reverse the order, so that more recent versions are at the top of the list

sascala:
Can do

jdef:
I like this idea -- good first step.

I wonder if it's worth scoping a set of changes to integrate the versioned docs concept into what's published via GH pages?

/cc @meichstedt

sascala:
I'm not sure I understand -- this change would appear on GitHub pages.

Yes, agreed that this is the most minimal of versioning and it would be better to have something more robust and automated.

PR #5404, opened 2017-05-24
Include USER networked docker containers in text/plain v2/tasks output

timcharper:
Use phab for code review; posted here because CI doesn't work with phab and Marathon 1.3

Summary:
Historically, we only listed HOST or BRIDGE networked containers in v2/tasks, due to the fact that assumption USER networked containers are not necessarily reachable by other tasks.

With this change, we modify this restriction and output the containerIp and containerPort for ip-per-task containers

Test Plan: manually launch against container networks

Subscribers: marathon-dev

JIRA Issues: MARATHON-7397

Differential Revision: https://phabricator.mesosphere.com/D809


timcharper:
retest this please

timcharper:
retest this please

PR #5406, opened 2017-05-24
Fixes #4168 - Add support for Mesos HTTP/TCP health checks (#4210)

### LINKED ISSUE #4168: Support Mesos HTTP/TCP health checks

janisz:
This is a cherrypick of a @gkleiman's PR #4210. 
Marathon 1.4 introduced Mesos HTTP/TCP healthchecks. This could result in better performance of a Marathon. Unfortunately Marathon 1.4 introduced a performance degradation making it unusable in bigger installations ([MARATHON-7358](https://jira.mesosphere.com/browse/MARATHON-7358)). This PR add Mesos Healthcecks support to Marathon 1.3. This change is future proof and upgrading to 1.4 is safe.

Moving all Marathon applications from Marathon healthcecks to Mesos Healthcecks can't be done at once. Every application need to be redeployed with new configuration. This PR allow users to slowly move to Mesos Healthchecks and then upgrade to Marathon 1.4

<details>
Manually fixed Conflicts:

-	[`project/build.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/project/build.scala)
-	[`src/main/java/mesosphere/marathon/Protos.java`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/main/java/mesosphere/marathon/Protos.java)
-	[`src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala)
-	[`src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala)
-	[`src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala)
-	[`src/main/scala/mesosphere/marathon/state/AppDefinition.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/main/scala/mesosphere/marathon/state/AppDefinition.scala)
-	[`src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala)
-	[`src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckActorTest.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckActorTest.scala)
-	[`src/test/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManagerTest.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/test/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManagerTest.scala)
-	[`src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala)
-	[`src/test/scala/mesosphere/marathon/state/MarathonStoreTest.scala`](https://github.com/janisz/marathon/blob/431de4e8932b048433405131a8875904895f4d73/src/test/scala/mesosphere/marathon/state/MarathonStoreTest.scala)
</details>

### DIFF
--- a/docs/docs/command-line-flags.md
+++ b/docs/docs/command-line-flags.md
@@ -47,6 +47,7 @@ The core functionality flags can be also set by environment variable `MARATHON_O
     - "vips" can be used to enable the networking VIP integration UI.
     - "task\_killing" can be used to enable the TASK\_KILLING state in Mesos (0.28 or later)
     - "external\_volumes" can be used if the cluster is configured to use external volumes.
+    - "mesos\_healthchecks" can be used to enable the Mesos HealthChecks (Mesos 1.1 or later).
     Example: `--enable_features vips,task_killing,external_volumes`
 * `--executor` (Optional. Default: "//cmd"): Executor to use when none is
     specified.
--- a/project/build.scala
+++ b/project/build.scala
@@ -149,7 +149,7 @@ object MarathonBuild extends Build {
       "-Xlog-reflective-calls",
       "-Xlint",
       "-Ywarn-unused-import",
-      "-Xfatal-warnings",
+//      "-Xfatal-warnings",
       "-Yno-adapted-args",
       "-Ywarn-numeric-widen"
     ),
@@ -329,8 +329,8 @@ object Dependency {
     val Chaos = "0.8.7"
     val Guava = "19.0"
     // FIXME (gkleiman): reenable deprecation checks after Mesos 1.0.0-rc2 deprecations are handled
-    val MesosUtils = "1.0.2"
-    val Akka = "2.4.7"
+    val MesosUtils = "1.2.0"
+    val Akka = "2.4.8"
     val AsyncAwait = "0.9.6-RC2"
     val Spray = "1.3.3"
     val TwitterCommons = "0.0.76"
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -1,11 +1,6 @@
 <scalastyle commentFilter="enabled">
  <name>Mesosphere Marathon configuration</name>
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"/>
- <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
-  <parameters>
-   <parameter name="maxFileLength"><![CDATA[800]]></parameter>
-  </parameters>
- </check>
  <check level="error" class="org.scalastyle.file.HeaderMatchesChecker" enabled="false">
   <parameters>
    <parameter name="header"><![CDATA[]]></parameter>
--- a/src/main/scala/mesosphere/marathon/Features.scala
+++ b/src/main/scala/mesosphere/marathon/Features.scala
@@ -17,12 +17,16 @@ object Features {
   //enable GPUs
   lazy val GPU_RESOURCES = "gpu_resources"
 
+  //enable Mesos HealthChecks
+  lazy val MESOS_HEALTHCHECKS = "mesos_healthchecks"
+
   lazy val availableFeatures = Map(
     VIPS -> "Enable networking VIPs UI",
     TASK_KILLING -> "Enable the optional TASK_KILLING state, available in Mesos 0.28 and later",
     EXTERNAL_VOLUMES -> "Enable external volumes support in Marathon",
     SECRETS -> "Enable support for secrets in Marathon (experimental)",
-    GPU_RESOURCES -> "Enable support for GPU in Marathon (experimental)"
+    GPU_RESOURCES -> "Enable support for GPU in Marathon (experimental)",
+    MESOS_HEALTHCHECKS -> "Enable support for Mesos Native Healthchecks in Marathon (require Mesos version >= 1.1.0)"
   )
 
   def description: String = {
--- a/src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala
@@ -4,8 +4,8 @@ import com.wix.accord.dsl._
 import mesosphere.marathon.Features
 import mesosphere.marathon.Protos.Constraint
 import mesosphere.marathon.api.v2.Validation._
-import mesosphere.marathon.core.readiness.ReadinessCheck
 import mesosphere.marathon.core.health.HealthCheck
+import mesosphere.marathon.core.readiness.ReadinessCheck
 import mesosphere.marathon.state._
 
 import scala.collection.immutable.Seq
@@ -53,7 +53,7 @@ case class AppUpdate(
 
     container: Option[Container] = None,
 
-    healthChecks: Option[Set[HealthCheck]] = None,
+    healthChecks: Option[Set[_ <: HealthCheck]] = None,
 
     readinessChecks: Option[Seq[ReadinessCheck]] = None,
 
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -6,17 +6,16 @@ import com.wix.accord._
 import com.wix.accord.combinators.GeneralPurposeCombinators
 import com.wix.accord.dsl._
 import mesosphere.marathon.Protos.Constraint
-import mesosphere.marathon.Protos.HealthCheckDefinition.Protocol
 import mesosphere.marathon.state.Container.{ Docker, MesosAppC, MesosDocker }
 // scalastyle:off
 import mesosphere.marathon.api.serialization.{ ContainerSerializer, EnvVarRefSerializer, PortDefinitionSerializer, ResidencySerializer, SecretsSerializer }
 // scalastyle:on
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.core.externalvolume.ExternalVolumes
+import mesosphere.marathon.core.health.{ HealthCheck, MarathonHealthCheck, MesosHealthCheck, MesosCommandHealthCheck }
 import mesosphere.marathon.core.plugin.PluginManager
 import mesosphere.marathon.core.readiness.ReadinessCheck
 import mesosphere.marathon.core.task.Task
-import mesosphere.marathon.core.health.HealthCheck
 import mesosphere.marathon.plugin.validation.RunSpecValidator
 import mesosphere.marathon.state.AppDefinition.VersionInfo.{ FullVersionInfo, OnlyVersion }
 import mesosphere.marathon.state.AppDefinition.{ Labels, VersionInfo }
@@ -72,7 +71,7 @@ case class AppDefinition(
 
   container: Option[Container] = AppDefinition.DefaultContainer,
 
-  healthChecks: Set[HealthCheck] = AppDefinition.DefaultHealthChecks,
+  healthChecks: Set[_ <: HealthCheck] = AppDefinition.DefaultHealthChecks,
 
   readinessChecks: Seq[ReadinessCheck] = AppDefinition.DefaultReadinessChecks,
 
@@ -121,7 +120,7 @@ case class AppDefinition(
       hostPorts = Seq.empty,
       envPrefix = None
     )
-    val cpusResource = ScalarResource(Resource.CPUS, cpus)
+    val cpusResource: Resource = ScalarResource(Resource.CPUS, cpus)
     val memResource = ScalarResource(Resource.MEM, mem)
     val diskResource = ScalarResource(Resource.DISK, disk)
     val gpusResource = ScalarResource(Resource.GPUS, gpus.toDouble)
@@ -257,7 +256,7 @@ case class AppDefinition(
       fetch = proto.getCmd.getUrisList.asScala.map(FetchUri.fromProto).to[Seq],
       storeUrls = proto.getStoreUrlsList.asScala.to[Seq],
       container = containerOption,
-      healthChecks = proto.getHealthChecksList.iterator().asScala.map(new HealthCheck().mergeFromProto).toSet,
+      healthChecks = proto.getHealthChecksList.iterator().asScala.map(HealthCheck.fromProto).toSet,
       readinessChecks =
         proto.getReadinessCheckDefinitionList.iterator().asScala.map(ReadinessCheckSerializer.fromProto).to[Seq],
       taskKillGracePeriod = if (proto.hasTaskKillGracePeriod) Some(proto.getTaskKillGracePeriod.milliseconds)
@@ -378,10 +377,9 @@ case class AppDefinition(
       case IpAddress(_, _, DiscoveryInfo(appPorts), _) =>
         for {
           launched <- task.launched
-          effectiveIpAddress <- task.effectiveIpAddress(this)
         } yield appPorts.zip(launched.hostPorts).map {
           case (appPort, hostPort) =>
-            PortAssignment(Some(appPort.name), effectiveIpAddress, hostPort)
+            PortAssignment(Some(appPort.name), task.effectiveIpAddress(this), hostPort, Some(hostPort))
         }.toList
     }
 
@@ -390,7 +388,6 @@ case class AppDefinition(
         c <- container
         pms <- c.portMappings
         launched <- task.launched
-        effectiveIpAddress <- task.effectiveIpAddress(this)
       } yield {
         var hostPorts = launched.hostPorts
         pms.map { portMapping =>
@@ -403,7 +400,19 @@ case class AppDefinition(
               hostPort
             }
 
-          PortAssignment(portMapping.name, effectiveIpAddress, effectivePort)
+          PortAssignment(
+            portMapping.name,
+            task.effectiveIpAddress(this),
+            effectivePort,
+            if (portMapping.hostPort.isDefined)
+              if (ipAddress.isDefined) {
+              val hostPort = hostPorts.head
+              hostPorts = hostPorts.drop(1)
+              Some(hostPort)
+            } else Some(effectivePort)
+            else None,
+            Some(portMapping.containerPort)
+          )
         }
       }.toList
     }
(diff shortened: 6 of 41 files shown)


janisz:
@gkleiman Can you take a look at this?

timcharper:
Thank you, @janisz ! I'll take a look at this Tuesday. (Monday is a holiday) :)

timcharper:
@janisz reviewed; looks like a straight forward (and... probably painful) back-port of the feature. Great work!

I don't see any reason to not just merge this; the enum values are the same Going to request a second opinion from @jdef just to be sure there's not any unexpected consequences of merging.

Have you already deployed this patch to your environment? Do both types of health-checks work as expected?

janisz:
I've only tested it on development cluster. It works as exepcted with no problems. We will push this version to all environments and then slowly introduce MESOS_HTTP healthchecks to newly deployed/updated applications.

janisz:
@timcharper It works at our infrastructure for 3 weeks. We are slowly moving applications to Mesos Native healthchecks. Now about 35% of our apps has MESOS_HTTP healthcheck on our biggest cluster. During this process we didn't observed any issues.

I think it's ready to merge.

janisz:
Below is a graph presenting Marathon requests time for 1.3.10/1.4.2 and 1.3.10 with this patch and enabled Mesos Healthchecks. Currently only 40% of healthchecks are done by Mesos. As you can see responses times are more stable. It's working for a while for us. Other metrics also looks more stable then with Marathon Healthcekcs.

We roll backed from 1.4.2 because it has performance issues.

marathon-consul-test2-mesos-http

janisz:
@unterstein I added commit with AppDefinition validator to explicitly enable Mesos HealthChecks before use.

docs/docs/command-line-flags.md
```diff
@@ -47,6 +47,7 @@ The core functionality flags can be also set by environment variable `MARATHON_O
     - "vips" can be used to enable the networking VIP integration UI.
     - "task\_killing" can be used to enable the TASK\_KILLING state in Mesos (0.28 or later)
     - "external\_volumes" can be used if the cluster is configured to use external volumes.
+    - "mesos_healthchecks" can be used if ti enable the Mesos HealthChecks (Mesos 1.1 or later).
```
unterstein: I think there is some black magic rule, why the others are using `\_`

src/main/scala/mesosphere/marathon/Features.scala
```diff
@@ -17,12 +17,16 @@ object Features {
...
+  lazy val MESOS_HEALTHCHECKS = "mesos_healthchecks"
+
   lazy val availableFeatures = Map(
     VIPS -> "Enable networking VIPs UI",
     TASK_KILLING -> "Enable the optional TASK_KILLING state, available in Mesos 0.28 and later",
     EXTERNAL_VOLUMES -> "Enable external volumes support in Marathon",
     SECRETS -> "Enable support for secrets in Marathon (experimental)",
-    GPU_RESOURCES -> "Enable support for GPU in Marathon (experimental)"
+    GPU_RESOURCES -> "Enable support for GPU in Marathon (experimental)",
+    MESOS_HEALTHCHECKS -> "Enable support for Mesos Native Healthchecks in Marathon (require Mesos version >= 1.1.0)"
```
unterstein: I think mesos health checks were introduced in `1.2.0` according to https://mesosphere.com/blog/2017/01/05/introducing-mesos-native-health-checks-apache-mesos-part-1/ ?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -769,6 +770,13 @@ object AppDefinition extends GeneralPurposeCombinators {
     }
   }
 
+  private val complyWithMesosHealthChecks: Validator[AppDefinition] =
+    conditional[AppDefinition](appDef =>
+      (appDef.healthChecks.count(_.isInstanceOf[MesosHealthCheck]) -
+        appDef.healthChecks.count(_.isInstanceOf[MesosCommandHealthCheck])) > 0) {
+      featureEnabled(Features.MESOS_HEALTHCHECKS)
```
unterstein: I would add something like this additionally:

```
val mesosHealthCheckMinVersion = SemanticVersion(1, 2, 0)
...
featureEnabled(Features.MESOS_HEALTHCHECKS) && LibMesos.version >= mesosHealthCheckMinVersion
```

What do you think?

docs/docs/command-line-flags.md
unterstein: I think mesos health checks were introduced in `1.2.0` according to https://mesosphere.com/blog/2017/01/05/introducing-mesos-native-health-checks-apache-mesos-part-1/ ?

janisz: I don't know why it's done that way. Probably to prevent changing font _style_. When you take a look at [rendered html it looks the same](https://github.com/janisz/marathon/blob/ed340f969186ae99d43ad10cc42aadeae2997980/docs/docs/command-line-flags.md#optional-flags).
IMO this should be wrapped in code format (\`) not double quotes (`"`)

src/main/scala/mesosphere/marathon/Features.scala
janisz: This blog post is partially correct. Mesos Health Checks were introduced as an **experimental** feature in [Mesos 1.1.0](https://github.com/apache/mesos/blob/1.1.0/CHANGELOG#L15-L30)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
janisz: Does the libmesos version has anything to do with communication. I can use libmesos from previous version and my messages will be correctly transported. Libmesos has no correlation with real Mesos version. Instead of this we should query Mesos master to get it's version.

docs/docs/command-line-flags.md
janisz: This blog post is partially correct. Mesos Health Checks were introduced as an **experimental** feature in [Mesos 1.1.0](https://github.com/apache/mesos/blob/1.1.0/CHANGELOG#L15-L30)

janisz: I `git blame` this file and it looks like it's Jekyll implementation
(escape underscore. Otherwise jekyl will do a <em> around it)[https://github.com/mesosphere/marathon/commit/6462bb992c43bea1866afcd95f44d0c59c673d34]

janisz: Fixed

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
unterstein: hmm .. feature flag should be enough IMO. What do you say @meichstedt ?

meichstedt: tl;dr: fine with me.

Long version:
We don't check master version for other feature flags, so I can't say we must do that here. Yet in general, it would be better to combine feature flags with a minimum master version to make sure that the user gets a meaningful error message when trying to enable a feature that isn't supported by the used mesos version.

Although Mesos tries to be pretty much backwards compatible, I'm sure I have seen invalid combinations of super native library with a newer Mesos. Current master code checks for both libmesos version and master version [here](https://github.com/mesosphere/marathon/blob/23cb098275f0db1a9afda0d59ca9144e99536937/src/main/scala/mesosphere/mesos/LibMesos.scala#L7)

The `MesosMasterMinimumVersion` is checked against after having registered – if the version reported by the frameworkInfo is less, Marathon will throw and commit suicide. Libmesos will be checked during startup (even before registering).

Btw the version as reported by libmesos returned bogus values until [MESOS-6502](https://issues.apache.org/jira/browse/MESOS-6502) was fixed ([here](https://github.com/apache/mesos/commit/cbbc5a46a499eb125eac83af1f21b832f136385a))

jdef:
This is a cool PR! That said, we normally don't backport large features changes like this into a stable branch. We should **proceed with caution** regarding where these changes end up landing.

project/build.scala
```diff
@@ -329,8 +329,8 @@ object Dependency {
     val Chaos = "0.8.7"
     val Guava = "19.0"
     // FIXME (gkleiman): reenable deprecation checks after Mesos 1.0.0-rc2 deprecations are handled
-    val MesosUtils = "1.0.2"
-    val Akka = "2.4.7"
+    val MesosUtils = "1.2.0"
```
jdef: this is not guarded by feature-flag; implications is that future builds of Marathon depend on semantics of newer API, yes?

scalastyle-config.xml
```diff
@@ -1,11 +1,6 @@
 <scalastyle commentFilter="enabled">
  <name>Mesosphere Marathon configuration</name>
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"/>
- <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
```
jdef: why remove? seems unrelated to the nature of this PR

src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala
```diff
@@ -53,7 +53,7 @@ case class AppUpdate(
 
     container: Option[Container] = None,
 
-    healthChecks: Option[Set[HealthCheck]] = None,
+    healthChecks: Option[Set[_ <: HealthCheck]] = None,
```
jdef: do we really need a complicated type like this?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -72,7 +71,7 @@ case class AppDefinition(
 
   container: Option[Container] = AppDefinition.DefaultContainer,
 
-  healthChecks: Set[HealthCheck] = AppDefinition.DefaultHealthChecks,
+  healthChecks: Set[_ <: HealthCheck] = AppDefinition.DefaultHealthChecks,
```
jdef: why the complex type?

scalastyle-config.xml
unterstein: Maybe @gkleiman can give more context here.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
unterstein: Maybe @gkleiman can give more context here.

src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala
unterstein: Maybe @gkleiman can give more context here.

project/build.scala
janisz: Correct. We can downgrade to 1.1 here.

src/main/scala/mesosphere/marathon/api/v2/json/AppUpdate.scala
janisz: > [Scala immutable sets are not covariant.](https://github.com/mesosphere/marathon/pull/4210#issuecomment-244425335)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
janisz: > [Scala immutable sets are not covariant.](https://github.com/mesosphere/marathon/pull/4210#issuecomment-244425335)

docs/docs/command-line-flags.md
```diff
@@ -47,6 +47,7 @@ The core functionality flags can be also set by environment variable `MARATHON_O
     - "vips" can be used to enable the networking VIP integration UI.
     - "task\_killing" can be used to enable the TASK\_KILLING state in Mesos (0.28 or later)
     - "external\_volumes" can be used if the cluster is configured to use external volumes.
+    - "mesos\_healthchecks" can be used if ti enable the Mesos HealthChecks (Mesos 1.1 or later).
```
janisz: typo: `if ti` → `to`

scalastyle-config.xml
janisz: I agree with @jdef this shouldn't be touched. Should I revert it or keep it consistent with 1.4?

janisz:
> it introduces a dependency on a newer Mesos API (with breaking changes): for example, custom executors have new proto requirements for the 1.2 API.

@jdef What breaking changes do you refer to? I haven't noticed any breaking changes? Noting is mentioned in [Mesos changelog](https://github.com/apache/mesos/blob/1.2.0/CHANGELOG) too? Can you elaborate.

jdef:
I built a toy framework in mesos-go that uses a custom executor. I didn't
need to make any proto-building changes when upgrading from v1.0.x to
v1.1.x mesos protos. Upon upgrading to mesos v1.2.x protos the logic that
built a custom executor needs to set the executor's type to CUSTOM
otherwise mesos rejects it. Maybe this is an under-documented feature of
Mesos 1.2.x?

janisz:
We are using custom executor with Marathon and we didn't need to change anything while upgrading Marathon (1.3 then 1.4.2 then 1.3 with HC). [Marathon still doesn't set executors properly](https://github.com/mesosphere/marathon/blame/48bfffdbb2920732ba726ce19ade6deb00121794/src/main/scala/mesosphere/mesos/TaskBuilder.scala#L62-L77) ([see also MARATHON-4210](https://jira.mesosphere.com/browse/MARATHON-4210)). Executor code lives untouched from ages and [nobody knows why/how it's working](https://jira.mesosphere.com/browse/MARATHON-7191?focusedCommentId=100100&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-100100).

jdef:
oh boy. maybe there's a difference between the mesos v0 and v1 API
implementation?!

timcharper:
@janisz do you still feel like you need this change?

I'd love to just merge it as is, it is really great work! However, I got pushback on the idea of supporting a Mesos 1.1.0 feature in a version of Marathon (1.3.x) that targets Mesos 1.0.0. I thought of pushing a custom build of this change, but, this seems like a less-than-optimal road, too. Perhaps you create a release and upload the artifact to your fork, so others can find it?

With regard to this comment:

> Moving all Marathon applications from Marathon healthcecks to Mesos Healthcecks can't be done at once. 

Is it not possible to move to Mesos health checks after upgrading to Marathon 1.4.x?

Hopefully we have fixed the blocked issues to upgrade to Marathon 1.4.x. We changed a lot of internals and, unfortunately, it has been a bit of a bumpy ride.

timcharper:
Closing :(

We couldn't accept this for 1.3.x because it breaks our version compatibility matrix. The version of DCOS that ships Marathon 1.3.x also ships a version of Mesos that does not support these health checks.

Sorry about all this. I would like to find a way to improve our collaboration process where such changes can be discussed first (and consensus achieved) before changes are made. I wish Marathon were a little nimble open source project that we could just run with and change as it made sense. The reality is a pretty big picture around it, so, we need to talk about these types of changes and make sure we understand the impact. I failed to do so prior to giving my accept for the PR (and, have failed to do so, in the past, on some other ones, because I really hate saying no to PRs). This applies to core Marathon committers, too.

I still hate saying no to PRs. I'd much prefer to redirect effort, earlier, so we can say yes, later.

janisz:
@timcharper No problem. I was aware that there is small chance it will be merged (I should have written it in the first comment). I was thinking about closing this PR earlier but maybe there was someone who needed it. This PR was created to show the community there is a chance to get Mesos HealthChecks without Marathon 1.4 that was "unstable" at that time. We are happily running Marathon 1.3 with this patch and it saves from performance issues described [here](https://mesosphere.com/blog/introducing-mesos-native-health-checks-apache-mesos-part-1/). I hope we soon update to 1.4.

> I still hate saying no to PRs. I'd much prefer to redirect effort, earlier, so we can say yes, later.

Agree, I'll prepare some proposal next time before submitting a patch.

PR #5411, opened 2017-05-31
Do not block on purge in MarathonSchedulerActor.

jeschkies:
Summary:
The synchronous call to purge can block up to 33 seconds. This caused a
flakiness in GroupDeployIntegration tests. The MarathonSchedulerActor
was not able to process other message and thus the Deploy command was
not received.

This does not solve the issue of the delay withing the purge. See
MARATHON-7365 for further investigation.

Test Plan: pipeline

Reviewers: timcharper, zen-dog, jenkins

Reviewed By: timcharper, zen-dog, jenkins

Subscribers: marathon-dev, marathon-team, jenkins

JIRA Issues: MARATHON-7430

Differential Revision: https://phabricator.mesosphere.com/D814

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueue.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueue.scala
@@ -68,7 +68,7 @@ trait LaunchQueue {
   def count(specId: PathId): Int
 
   /** Remove all instance launch requests for the given PathId from this queue. */
-  def purge(specId: PathId): Unit
+  def asyncPurge(specId: PathId): Future[Done]
 
   /** Add delay to the given RunnableSpec because of a failed instance */
   def addDelay(spec: RunSpec): Unit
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueue.scala
```diff
@@ -69,6 +69,7 @@ trait LaunchQueue {
 
   /** Remove all instance launch requests for the given PathId from this queue. */
   def purge(specId: PathId): Unit
```
janisz: Can `purge` be deleted since it's replaced with `asyncPurge`?

jeschkies: That's a good point. We've discussed it for the review on master and declared it as deprecated and will remove it in 1.6. What do you think?

janisz: OK

zen-dog: This internal method is not part of the API. Removing it shouldn't be a problem.

PR #5423, opened 2017-06-17
Simplify the check for changed specs.

Tordek:
By using `Option.contains` we can avoid looking up the same key multiple
times.


Tordek:
Hi @unterstein!

Funny story: someone on a reddit thread regarding a patch adjacent to this complained about this line and proposed the fix... so I nicked it (with their approval).

janisz:
reddit discussion about this: https://www.reddit.com/r/programming/comments/6hnifm/comment/dizr266

Tordek:
Thanks!

PR #5424, opened 2017-06-17
Added perf test for large group deployment.

janisz:
Refs: MARATHON-7358


janisz:
Something is not OK. When I run this test on master I get deploy that only deploys groups no app i deployed.

janisz:
Closing due to lack of interest

PR #5432, opened 2017-07-02
App and pod validation errors for missing network name MARATHON-7398

alenkacz:
Throws validation exception instead of just NormalizationException when default network name is not specified

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
@@ -5,6 +5,7 @@ import java.util.regex.Pattern
 
 import com.wix.accord._
 import com.wix.accord.dsl._
+import mesosphere.marathon.api.v2.AppNormalization
 import mesosphere.marathon.api.v2.Validation.{ featureEnabled, _ }
 import mesosphere.marathon.core.externalvolume.ExternalVolumes
 import mesosphere.marathon.raml._
@@ -304,7 +305,7 @@ trait AppValidation {
     }
   )
 
-  def validateCanonicalAppUpdateAPI(enabledFeatures: Set[String]): Validator[AppUpdate] = forAll(
+  def validateCanonicalAppUpdateAPI(enabledFeatures: Set[String], normalizationConfig: AppNormalization.Config): Validator[AppUpdate] = forAll(
     validator[AppUpdate] { update =>
       update.id.map(PathId(_)) as "id" is optional(valid)
       update.dependencies.map(_.map(PathId(_))) as "dependencies" is optional(every(valid))
@@ -323,6 +324,9 @@ trait AppValidation {
     },
     isTrue("portDefinitions are only allowed with host-networking") { update =>
       !(update.networks.exists(_.exists(_.mode != NetworkMode.Host)) && update.portDefinitions.exists(_.nonEmpty))
+    },
+    isTrue(AppValidationMessages.NetworkNameMustBeSpecified) { update =>
+      update.networks.forall(n => n.forall(c => c.mode != NetworkMode.Container || c.name.isDefined || normalizationConfig.defaultNetworkName.isDefined))
     },
     isTrue("The 'version' field may only be combined with the 'id' field.") { update =>
       def onlyVersionOrIdSet: Boolean = update.productIterator.forall {
@@ -367,7 +371,7 @@ trait AppValidation {
     }
   )
 
-  def validateCanonicalAppAPI(enabledFeatures: Set[String]): Validator[App] = forAll(
+  def validateCanonicalAppAPI(enabledFeatures: Set[String], normalizationConfig: AppNormalization.Config): Validator[App] = forAll(
     validBasicAppDefinition(enabledFeatures),
     validator[App] { app =>
       PathId(app.id) as "id" is (PathId.pathIdValidator and PathId.absolutePathValidator and PathId.nonEmptyPath)
@@ -380,6 +384,9 @@ trait AppValidation {
     },
     isTrue("portDefinitions are only allowed with host-networking") { app =>
       !(app.networks.exists(_.mode != NetworkMode.Host) && app.portDefinitions.exists(_.nonEmpty))
+    },
+    isTrue(AppValidationMessages.NetworkNameMustBeSpecified) { app =>
+      app.networks.forall(c => c.mode != NetworkMode.Container || c.name.isDefined || normalizationConfig.defaultNetworkName.isDefined)
     }
   )
 
@@ -530,4 +537,6 @@ object AppValidationMessages {
 
   val DockerEngineLimitedToSingleContainerNetwork =
     "may only specify a single container network when using the Docker container engine"
+
+  val NetworkNameMustBeSpecified = "Network name must be specified when container network is selected."
 }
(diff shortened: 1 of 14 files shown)


unterstein:
Hey @jdef,

when your back after US holidays, could you have a look, this would be awesome! 

Thanks =)

src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
```diff
@@ -304,7 +305,7 @@ trait AppValidation {
     }
   )
 
-  def validateCanonicalAppUpdateAPI(enabledFeatures: Set[String]): Validator[AppUpdate] = forAll(
+  def validateCanonicalAppUpdateAPI(enabledFeatures: Set[String], normalizationConfig: AppNormalization.Config): Validator[AppUpdate] = forAll(
```
jdef: please use a `defaultNetworkName: Option[String]` here. let's not mix normalization APIs into the validation code

src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
```diff
@@ -367,7 +371,7 @@ trait AppValidation {
     }
   )
 
-  def validateCanonicalAppAPI(enabledFeatures: Set[String]): Validator[App] = forAll(
+  def validateCanonicalAppAPI(enabledFeatures: Set[String], normalizationConfig: AppNormalization.Config): Validator[App] = forAll(
```
jdef: ditto, `Option[String]`

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -194,7 +195,7 @@ trait PodsValidation {
     hostPorts.distinct.size == hostPorts.size
   }
 
-  def podValidator(enabledFeatures: Set[String], mesosMasterVersion: SemanticVersion): Validator[Pod] = validator[Pod] { pod =>
+  def podValidator(enabledFeatures: Set[String], mesosMasterVersion: SemanticVersion, defaultNetworkName: ScallopOption[String]): Validator[Pod] = validator[Pod] { pod =>
```
jdef: please don't use scallop here, it's not needed. use a simple `Option[String]` type instead. pretty sure that you can invoke `.get` on a `ScallopOption[T]` to get an `Option[T]` (to pass in here)

src/main/scala/mesosphere/marathon/api/v2/validation/PodsValidation.scala
```diff
@@ -211,6 +212,9 @@ trait PodsValidation {
     }
     pod.secrets is empty or (valid(secretValidator) and featureEnabled(enabledFeatures, Features.SECRETS))
     pod.networks is valid(ramlNetworksValidator)
+    pod.networks is isTrue[Seq[Network]]("network name must be specified when container network is selected") { nets =>
```
jdef: this logic is the same for both apps and pods. perhaps extract into a `NetworkValidation` class and `NetworkValidationMessages` class (used by both app and pod validation), and use the static message field (instead of a literal string) for the validation error

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
```diff
@@ -3,15 +3,19 @@ package api.v2.validation
...
 
   import Normalization._
 
-  implicit val basicValidator: Validator[App] = AppValidation.validateCanonicalAppAPI(Set.empty)
-  implicit val withSecretsValidator: Validator[App] = AppValidation.validateCanonicalAppAPI(Set("secrets"))
+  val config = AppNormalization.Configuration(None, "mesos-bridge-name")
+  val configWithDefaultNetworkName = AppNormalization.Configuration(Some("defaultNetworkName"), "mesos-bridge-name")
+  implicit val basicValidator: Validator[App] = AppValidation.validateCanonicalAppAPI(Set.empty, config)
+  implicit val withSecretsValidator: Validator[App] = AppValidation.validateCanonicalAppAPI(Set("secrets"), config)
+  implicit val withDefaultNetworkNameValidator: Validator[App] = AppValidation.validateCanonicalAppAPI(Set.empty, configWithDefaultNetworkName)
```
jdef: agree, redundant implicit types are strange here. it's not the OPs job to fix, but I'd like to suggest that we at least refrain from continuing the pattern: `withDefaultNetworkNameValidator` should not be implicit.

src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
unterstein: Due to the migration code here https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala#L63 it was not able to pass the `Option[String]` directly. Any ideas here?

jdef: sure, refactor some bits in the migration code: extract the network name logic into a separate `def` that's reused by the normalization init code

unterstein: We already had an discussion regarding this issue. There is an explicit override 
```
override def defaultNetworkName: Option[String] =
  | env.vars.get(DefaultNetworkNameForMigratedApps).orElse(networkName).orElse(throw SerializationFailedException(
  | MigrationFailedMissingNetworkEnvVar))


```
 which throws an exception if this is is None. But None is a valid state in some cases afterwards. Therefore we decided to pass this not directly and go for a lazy implementation. 

What about adding a trait `ValidationConfig` which holds the `enabledFeatures` and `defaultNetworkName` ?

jdef: Trait works. Or else a func that produces an `Option[String]` (which would be even more minimal than a trait)

alenkacz:
Thanks for the review! I'll fix those comments over the weekend :-)

alenkacz:
@jdef @unterstein can you re-review please? I fixed all your comments

PS: Sorry for the delay - holidays :-)

alenkacz:
@jdef @unterstein anything I can do to get this merged? :-)

src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
```diff
@@ -284,10 +284,9 @@ class PodsResourceTest extends AkkaUnitTest with Mockito {
 
       podSystem.create(any, eq(false)).returns(Future.successful(DeploymentPlan.empty))
 
-      val ex = intercept[NormalizationException] {
-        f.podsResource.create(podSpecJsonWithContainerNetworking.getBytes(), force = false, f.auth.request)
-      }
-      ex.msg shouldBe NetworkNormalizationMessages.ContainerNetworkNameUnresolved
+      val response = f.podsResource.create(podSpecJsonWithContainerNetworking.getBytes(), force = false, f.auth.request)
+      response.getStatus shouldBe 422
+      response.getEntity.toString should include("Network name must be specified")
```
jdef: nit: reference `NetworkValidationMessages.NetworkNameMustBeSpecified` instead of hard coding the text message

alenkacz:
@jdef good point - fixed

unterstein:
Thanks @alenkacz !

```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__          Shippin'  
         __|_o_o_o\__         it!
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^
```

PR #5435, opened 2017-07-11
Support mesos primitives for availability

alanbover:
This PR makes marathon drop those resource offers from agents that are set in maintenance.

Find more information here: https://jira.mesosphere.com/browse/MARATHON-3216


src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,23 @@
...
+object Availability {
+
+  def offerAvailable(offer: Offer): Boolean = {
+    var unavailability = offer.hasUnavailability
+    var now = System.nanoTime()
+    if (offer.hasUnavailability && offer.getUnavailability.hasStart) {
+      val start = offer.getUnavailability.getStart.getNanoseconds
+      if (now.>=(start)) {
+        if (offer.getUnavailability.hasDuration) {
+          return start.+(offer.getUnavailability.getDuration.getNanoseconds).<(now)
```
janisz: It will be cool if you allow configuring a period in which offer will be declined. For example if agent will be unavailable in next 15 minutes then there is no point in starting task there.

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.mesos.Protos.Offer
+
+object Availability {
+
+  def offerAvailable(offer: Offer): Boolean = {
+    var unavailability = offer.hasUnavailability
+    var now = System.nanoTime()
+    if (offer.hasUnavailability && offer.getUnavailability.hasStart) {
+      val start = offer.getUnavailability.getStart.getNanoseconds
+      if (now.>=(start)) {
```
janisz: do we need this dots?

ivanilves:
Fantastic job @alanbover !

alanbover:
@janisz  thanks for the comments, the changes have been done!

The PR is a little bit noisy now, mostly because it required some changes to make matchResources be able to receive marathon configurations

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -307,5 +307,12 @@ trait MarathonConf
     noshort = true,
     hidden = true,
     default = Some(MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_FAILURE_THRESHOLD))
+
+  lazy val drainingTime = opt[Long](
+    "draining_time",
+    descr = "(Default: 300 seconds) the seconds when marathon will start declining offers before a maintenance " +
```
janisz: IMO default should be 0 to not change current behavior.

src/main/scala/mesosphere/mesos/Availability.scala
unterstein: Hmm...as far as I understand maintenance mode, the agent is supposed to be offline during the scheduled maintenance frame (time between `start` and `start + duration`). If the agent is stopped to do this maintenance, the agent will not send offers.
To have a smooth behavior, we should implement this feature in multiple steps:
1) Ignore offers upfront (for a configurable amount of time) to the scheduled downtime (and during as well, if the agent is not down as planned), as @janisz suggested. I think this step could be easily added to this pr.
2) Implement a shift, where marathon moves all tasks from the to-be-maintained agent away to other agents. In the best case the start of this shift also should be configurable.

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,25 @@
+package mesosphere.mesos
+
+import java.util.concurrent.TimeUnit
+
+import org.apache.mesos.Protos.Offer
+
+object Availability {
+
+  def offerAvailable(offer: Offer, drainingTime: Long): Boolean = {
+    var unavailability = offer.hasUnavailability
```
janisz: Unused variable

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,25 @@
...
+    var now = System.nanoTime()
```
janisz: Could be `val`

janisz: In other places marathon uses `Clock`

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,25 @@
...
+      val start = offer.getUnavailability.getStart.getNanoseconds
+      if (now >= (start - TimeUnit.SECONDS.toNanos(drainingTime))) {
+        if (offer.getUnavailability.hasDuration) {
+          return start + (offer.getUnavailability.getDuration.getNanoseconds) < (now)
+        } else {
+          return false
+        }
+      }
+    }
+    return true
```
janisz: Could be simplified
```
  def offerAvailable(offer: Offer, drainingTime: Long): Boolean = {
    val now = System.nanoTime()
    val drainingTimeNanoseconds = TimeUnit.SECONDS.toNanos(drainingTime)
    if (offer.hasUnavailability && offer.getUnavailability.hasStart) {
      val start = offer.getUnavailability.getStart.getNanoseconds
      if (now >= (start - drainingTimeNanoseconds) && offer.getUnavailability.hasDuration) {
        start + offer.getUnavailability.getDuration.getNanoseconds < now
      }
      else false
    } else true
  }
```

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,25 @@
+package mesosphere.mesos
+
+import java.util.concurrent.TimeUnit
+
+import org.apache.mesos.Protos.Offer
+
+object Availability {
+
+  def offerAvailable(offer: Offer, drainingTime: Long): Boolean = {
```
janisz: Better if `drainingTime` is of `Duration` time or it name suggest it's in seconds.

src/main/scala/mesosphere/mesos/RunSpecOfferMatcher.scala
```diff
@@ -26,7 +27,7 @@ object RunSpecOfferMatcher {
     }
 
     val resourceMatchResponse =
-      ResourceMatcher.matchResources(offer, runSpec, otherInstances, ResourceSelector.any(acceptedResourceRoles))
+      ResourceMatcher.matchResources(offer, runSpec, otherInstances, ResourceSelector.any(acceptedResourceRoles), drainingTime)
```
janisz: Better if `drainingTime` is of `Duratio` time or it name suggest it's in seconds.

src/main/scala/mesosphere/mesos/Availability.scala
zmalik: @unterstein what you are saying is to implement maintenance primitives of mesos. And when a mode is in maintenence in mesos, mesos will keep offering its resources to frameworks, and its up to framework to accept the offers or not. 

 If you follow the https://jira.mesosphere.com/browse/MARATHON-3216 comments. It was already supposed to divide the task into two steps. 

The first step you are saying. This PR is already implementing only that. Dropping offers in maintenance mode. 
And the second step when you say implement a shift, it means implement inverse offers. And marathon needs to shift to v1 API of mesos to be able to do that.

src/main/scala/mesosphere/mesos/Availability.scala
zmalik: Yes, IMO better changing just name. This way this would be the first variable of MarathonConf to have the time unit in its name 🎆 
Some examples are mesosHeartbeatInterval, onElectedPrepareTimeout, scaleAppsInterval, scaleAppsInitialDelay. All are time units but in Long and without time unit in the name.

janisz: That's correct but later in the coude all this values are transformed to duration
```scala
  // Time to wait before trying to reconcile app tasks after driver starts
  val reconciliationInitialDelay =
    Duration(config.reconciliationInitialDelay(), MILLISECONDS)

  // Interval between task reconciliation operations
  val reconciliationInterval =
    Duration(config.reconciliationInterval(), MILLISECONDS)

  // Time to wait before trying to scale apps after driver starts
  val scaleAppsInitialDelay =
    Duration(config.scaleAppsInitialDelay(), MILLISECONDS)

  // Interval between attempts to scale apps
  val scaleAppsInterval =
    Duration(config.scaleAppsInterval(), MILLISECONDS)
```

zmalik: ok yeah you are right! Then should be ok with any option

alanbover:
@janisz thanks for your comments! It's already changed!

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import mesosphere.marathon.core.base.Clock
+import org.apache.mesos.Protos.Offer
+
+import scala.concurrent.duration._
+
+object Availability {
+
+  def offerAvailable(offer: Offer, drainingTime: Duration): Boolean = {
+    lazy val clock: Clock = Clock()
```
janisz: this could be moved to a field

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,22 @@
...
+    val drainingTimeNanoseconds = drainingTime.toNanos
+    if (offer.hasUnavailability && offer.getUnavailability.hasStart) {
+      val start = offer.getUnavailability.getStart.getNanoseconds
+      if (clock.now().millis >= (start - drainingTimeNanoseconds)) {
```
janisz: Comparing millis with nanoseconds? I'd prefer to keep everything as Duration

src/main/scala/mesosphere/mesos/Availability.scala
janisz: Best if clock could be injected to help testing.

src/main/scala/mesosphere/mesos/Availability.scala
alanbover: wow, you are right! massive confusion here!

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -307,5 +307,12 @@ trait MarathonConf
     noshort = true,
     hidden = true,
     default = Some(MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_FAILURE_THRESHOLD))
+
+  lazy val drainingTime = opt[Long](
+    "draining_time",
```
janisz: Please add information to https://github.com/mesosphere/marathon/blob/master/docs/docs/command-line-flags.md#optional-flags-1

src/main/scala/mesosphere/mesos/RunSpecOfferMatcher.scala
```diff
@@ -9,12 +10,14 @@ import org.apache.mesos.Protos.Offer
...
 import scala.collection.immutable.Seq
+import scala.concurrent.duration._
 
 object RunSpecOfferMatcher {
 
   val log = LoggerFactory.getLogger(getClass)
 
-  def matchOffer(runSpec: RunSpec, offer: Offer, otherInstances: => Seq[Instance], givenAcceptedResourceRoles: Set[String]): ResourceMatchResponse = {
+  def matchOffer(runSpec: RunSpec, offer: Offer, otherInstances: => Seq[Instance],
+    givenAcceptedResourceRoles: Set[String], drainingTime: FiniteDuration)(implicit clock: Clock): ResourceMatchResponse = {
```
janisz: We can make this PR much smaller by adding default value for `drainingTime`. On the other hand I prefer to avoid default values and be explicit.

alanbover: IMO, given that the parameter has a default value, we shouldn't set default value for drainingTime in other places, even if the PR gets noisy.

But I would be happy to change it if you think that's better

src/main/scala/mesosphere/marathon/MarathonConf.scala
timcharper: Does the name "draining time" imply that the parameter does more than it actually does? When I think of "draining", I would have the impression that Marathon would be killing the tasks and moving them to other machines.

Also, have we considered whether this makes more sense as a per-app setting, vs globally? It seems some tasks may be much more tolerant to failure than others, and would prefer to get as many resources as available. IE some background compute worker?

It is awesome that there is an implementation that is useful. Would it be helpful to take a step back and think of this from a macro perspective? Perhaps this change is on the road to where we ultimately want to go, I'd just like to be sure.

alanbover: We called it drainingTime because in Mesos terms, at the point an instance is scheduled for maintenance, the agent enters in "draining" state (http://mesos.apache.org/documentation/latest/maintenance/). 

I agree with you that draining sound like marathon will "migrate tasks to not-in-maintenance hosts". But currently marathon doesn't use HTTP scheduler, so it cannot migrate tasks taking as a reference the inversed offers. I hope marathon will at some point migrate to the HTTP scheduler, which will make it easy to implement the migration of tasks to another hosts.

About make it configurable for task, I do agree that it could be a good idea, but I'm unsure if this should be part of the first iteration (I would like to avoid changes on the deployment object). So I guess maybe it would be a good idea to change drainingTime to defaultDrainingTime, and then apps can override it. WDYT?

About taking an step back, I would really appreciate your feedback on this. I think because we have the specific use case for us (https://github.com/alanbover/deathnode), we could be failing into looking the big picture of the whole thing.

mattmb: Firstly, thanks for the work on this @alanbover!

I'm following this PR closely because we make use of Mesos maintenance primitives with https://github.com/Yelp/paasta

Currently when we schedule a host for maintenance and it enters a draining state we proactively try and scale up our Marathon apps to ensure that we have a new task somewhere else before we kill the draining host. However, as you know this leaves a possibility that Marathon will schedule the host back on the same box. Currently, we use http://mesos.readthedocs.io/en/0.23.1/reservation/#dynamic-reservation-since-0230 to reserve the resources to minimise the likelihood of this happening.

However, this PR would be a step in the right direction because we would eliminate that possibility if Marathon just dropped offers from maint'd hosts. I think from our point of view I'd expect Marathon to drop all offers from hosts scheduled for maintenance (i.e. draining) as well as any hosts currently in maintenance.

I hope that is useful context? And thanks again for looking at this!

alanbover: Hey @mattmb, thanks for the information! We are more or less following the same approach as you (we have an application that "restarts" the services that has one or more containers in a node in maintenance, forcing a rolling deployment following upgradeStrategy). 

So I understand that the approach that the PR follows would work also for you!

mattmb: I was a little unsure about the idea of the drainingTime. To me it feels like Marathon should decline offers as soon as a maintenance window is scheduled. The more I think about it the more sense it makes though. I guess I can set drainingTime to the same amount of time we set the start of the maintenance window in the future (or slightly longer). That way it will start declining the offers as soon as we schedule the maintenance window.

alanbover: Actually I was also thinking about this. Probably what it would have sense is to set default to 0, what would mean to directly drop the offers for those hosts that have a maintenance scheduled, no matter when the maintenance starts.

mattmb: Ah, I haven't read the code closely. If that's the meaning of 0 then that would be ideal for us.

I'm not really qualified to comment on the code itself but the approach in general gets a big 👍 from me.

alanbover: no, I mean that probably this is what it should be doing :P. I will re-think it a little bit, but I would like to get some feedback before rewriting from @timcharper

unterstein:
Hey @alanbover,

thank you again for this contribution, we really appreciate it.

We discussed this PR internally a while ago and we would love to merge this with a related feature flag to opt in for this behavior, something like `--enable_maintenance_mode` or something similar. This is a little bit annoying and also this PR currently do have some merge conflicts. Would it be ok for you if I would add and resolve those and open a PR against your fork?

alanbover:
hey @unterstein! 

Yes, I'm happy if you want to end up last details so it can be merged! Thanks a lot for that :)

unterstein:
Ok, great!

Short update: I already resolved the merge conflicts and currently adding one tiny commit to hide this feature behind a feature flag. Should be ready soon :)

janisz:
@unterstein This PR get preety big. Nearly 150 commits and over 300 changed files. Is there any chance to split it into smaller patches? It's really hard to review.

unterstein:
@janisz corrected the commit.

unterstein:
Hey @jdef, @timcharper, @jeschkies,

could you have a look at this pr now? I added the wished feature flag, resolved merge conflicts and added more tests 🎉

Cheers
Johannes

unterstein:
Also I added a phabricator diff trigger ci https://phabricator.mesosphere.com/D1069

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,32 @@
...
+          now.after(start + offer.getUnavailability.getDuration)
+      } else true
+    } else true
+  }
+
+  /**
+    * Convert Mesos DurationInfo to FiniteDuration.
+    * @return FiniteDuration for DurationInfo
+    */
+  implicit def toDuration(durationInfo: DurationInfo): FiniteDuration = {
```
jdef: maybe instead:
```scala
implicit class DurationInfoHelper(val di: DurationInfo) extends AnyVal {
  def toDuration(): FiniteDuration = FiniteDuration(di.getNanoseconds, NANOSECONDS)
}
```

src/main/scala/mesosphere/mesos/MatcherConf.scala
```diff
@@ -0,0 +1,14 @@
+package mesosphere.mesos
+
+import org.rogach.scallop.{ ScallopConf, ScallopOption }
+
+import scala.concurrent.duration._
+
+trait MatcherConf extends ScallopConf {
```
jdef: is it necessary to extend `ScallopConf` here?

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
```diff
@@ -24,7 +26,13 @@ import org.apache.mesos.Protos.Attribute
 import org.apache.mesos.{ Protos => Mesos }
 import org.scalatest.Inside
 
+import scala.concurrent.duration.Duration
+
 class ResourceMatcherTest extends UnitTest with Inside {
+
+  implicit val clock = Clock.systemUTC()
```
jdef: does this really need to be declared for the whole file, or can it be moved into the couple of test cases where it's needed? we normally use a fake clock for tests, so this seems like an exceptional case that should be appropriately scoped

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -26,6 +28,7 @@ class TaskBuilderTest extends UnitTest {
 
   import mesosphere.mesos.protos.Implicits._
 
+  implicit val clock = Clock.systemUTC()
```
jdef: any particular reason for using the system clock vs. a fake one?

src/test/scala/mesosphere/mesos/TaskGroupBuilderTest.scala
```diff
@@ -22,6 +25,10 @@ import scala.collection.JavaConverters._
 import scala.collection.breakOut
 
 class TaskGroupBuilderTest extends UnitTest with Inside {
+
+  implicit val clock = Clock.systemUTC()
```
jdef: ditto

src/test/scala/mesosphere/mesos/TaskGroupBuilderTest.scala
```diff
@@ -22,6 +25,10 @@ import scala.collection.JavaConverters._
...
+  val config = AllConf.withTestConfig("--draining_seconds", "300")
```
jdef: why this? isn't the feature disabled by default?

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
```diff
@@ -24,7 +26,13 @@ import org.apache.mesos.Protos.Attribute
...
+  val config = AllConf.withTestConfig("--draining_seconds", "300")
```
jdef: why this? isn't the feature disabled by default?

unterstein: It is off by default, removed configuration

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
unterstein: I think this happened during my merge conflict resolving, good catch! Changed this to the new `SettableClock`.

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
unterstein: This test is not operating with a fixture and have a bunch of those global vals. I filed a ticket to replace with a fixture https://jira.mesosphere.com/browse/MARATHON-7797

src/test/scala/mesosphere/mesos/TaskGroupBuilderTest.scala
unterstein: done

src/test/scala/mesosphere/mesos/TaskGroupBuilderTest.scala
unterstein: done

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,32 @@
...
+import scala.concurrent.duration._
+import scala.language.implicitConversions
+
+object Availability {
+
+  def offerAvailable(offer: Offer, drainingTime: FiniteDuration)(implicit clock: Clock): Boolean = {
+    val now = clock.now()
+    if (offer.hasUnavailability && offer.getUnavailability.hasStart) {
+      val start: Timestamp = offer.getUnavailability.getStart
+      if (now.after(start - drainingTime)) {
```
meln1k: What do you thing about moving that into separate methods to improve readability?

plugin-interface/src/main/scala/mesosphere/marathon/plugin/scheduler/SchedulerPlugin.scala
```diff
@@ -9,7 +9,7 @@ import org.apache.mesos.Protos.Offer
   * Allows to use external logic to decline offers.
   */
 trait SchedulerPlugin extends Plugin {
- 
+
```
zen-dog: rm new line

src/test/scala/mesosphere/marathon/test/MarathonTestHelper.scala
```diff
@@ -125,6 +126,17 @@ object MarathonTestHelper {
     offerBuilder
   }
 
+  def makeBasicOfferWithUnavailability(startTime: Timestamp, duration: Duration): Offer.Builder = {
```
meln1k: Does it make sense to use `FiniteDuration` here?

src/main/scala/mesosphere/mesos/Availability.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  private def currentlyInDrainingState(now: Timestamp, start: Timestamp, drainingTime: FiniteDuration) = {
+    now.after(start - drainingTime)
+  }
+
+  private def offerHasUnavailability(offer: Offer) = {
+    offer.hasUnavailability && offer.getUnavailability.hasStart
+  }
+
+  private def isAgentCurrentlyUnavailable(offer: Offer, start: Timestamp, now: Timestamp) = {
```
zen-dog: The naming is confusing. Logically it's smth. like `isAgentCurrentlyNOTUnavailable` which sounds bad.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -195,8 +198,16 @@ object ResourceMatcher extends StrictLogging {
       badConstraints.isEmpty
     }
 
+    val checkAvailability: Boolean = {
+      if (conf.availableFeatures.contains(Features.MAINTENANCE_MODE))
+        Availability.offerAvailable(offer, conf.drainingTime)
```
zen-dog: We need logging why this offer was declined.

zen-dog: And we need to add new statistic for declined offers due to `MaintenanceMode` (the `v2/queue`)

unterstein:
Thanks @meln1k and @zen-dog, please see last change to address your feedback :)

plugin-interface/src/main/scala/mesosphere/marathon/plugin/scheduler/SchedulerPlugin.scala
unterstein: This was a remove of an unwanted whitespace to prevent a reformat in the next time.

unterstein:
Thanks @alanbover!

```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__          Shippin'  
         __|_o_o_o\__         it!
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^
```

PR #5437, opened 2017-07-13
Clarify mesos vs marathon health checks

sascala:
https://jira.mesosphere.com/browse/MARATHON-7557

### DIFF
--- a/docs/docs/health-checks.md
+++ b/docs/docs/health-checks.md
@@ -16,9 +16,123 @@ application's tasks.
 A health check is considered passing if (1) its HTTP response code is between
 200 and 399 inclusive, and (2) its response is received within the
 `timeoutSeconds` period. If a task fails more than `maxConsecutiveFailures`
-health checks consecutively, that task is killed.
+health checks consecutively, that [task is terminated](#task-termination).
 
-##### Example usage
+## Mesos-level health checks versus Marathon-level health checks
+
+Marathon offers the following protocols for health checks. Declaring a protocol is optional: the default is `HTTP`.
+
+- `HTTP`
+- `HTTPS`
+- `TCP`
+- `COMMAND`
+- `MESOS_HTTP`
+- `MESOS_HTTPS`
+- `MESOS_TCP`
+
+### Marathon-level health checks
+
+Marathon-level health checks (HTTP, HTTPS, and TCP) are executed by Marathon and thus test reachability from the current Marathon leader. Marathon-level health checks have several limitations:
+
+- The API for health checks varies from one framework to the next and common functionality must be reimplemented with each new framework that requires health checks.
+
+- Marathon-level health checks create extra network traffic if the task and the scheduler are run on different nodes (this is usually the case).
+
+- Network failures between the task and the scheduler, and port mapping misconfigurations can make a healthy task look unhealthy.
+
+- If Marathon is managing a large number of tasks, performing health checks for every task can cause scheduler performance issues.
+
+These limitations may be acceptable for smaller clusters with low-scale tasks, but Marathon-level health checks should not be used on large clusters with highly scaled Marathon tasks.
+
+**Important:** Marathon-based health checks are deprecated and will be removed in a future version. See the [1.4.0 release notes](https://github.com/mesosphere/marathon/releases/tag/v1.4.0) for more information.
+
+### Mesos-level health checks
+
+Mesos-level health checks (`MESOS_HTTP`, `MESOS_HTTPS`, `MESOS_TCP`, and `COMMAND`) health checks are locally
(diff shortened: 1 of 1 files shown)


docs/docs/health-checks.md
```diff
@@ -16,9 +16,123 @@ application's tasks.
...
+- `COMMAND`
+- `MESOS_HTTP`
+- `MESOS_HTTPS`
+- `MESOS_TCP`
+
+### Marathon-level health checks
+
+Marathon-level health checks (HTTP, HTTPS, and TCP) are executed by Marathon and thus test reachability from the current Marathon leader. Marathon-level health checks have several limitations:
+
+- The API for health checks varies from one framework to the next and common functionality must be reimplemented with each new framework that requires health checks.
```
timcharper: This sounds like potentially a comment for framework developers and not users of Marathon. Does that sound right to you?

docs/docs/health-checks.md
```diff
@@ -16,9 +16,123 @@ application's tasks.
...
+These limitations may be acceptable for smaller clusters with low-scale tasks, but Marathon-level health checks should not be used on large clusters with highly scaled Marathon tasks.
+
+**Important:** Marathon-based health checks are deprecated and will be removed in a future version. See the [1.4.0 release notes](https://github.com/mesosphere/marathon/releases/tag/v1.4.0) for more information.
+
+### Mesos-level health checks
+
+Mesos-level health checks (`MESOS_HTTP`, `MESOS_HTTPS`, `MESOS_TCP`, and `COMMAND`) health checks are locally
+executed by Mesos on the agent running the corresponding task and thus test reachability from the Mesos executor. Mesos-level health checks offer the following advantages over Marathon-level health checks:
+
+- All built-in executors support native health checks via the new API and library. Custom executors can now use the same library and API to implement health checks, regardless of whether their tasks are actual processes or something else.
```
timcharper: I also wonder if this bit is for a different audience

docs/docs/health-checks.md
```diff
@@ -16,9 +16,123 @@ application's tasks.
...
+
+- Mesos-level health checks are performed as close to the task as possible, so they are are not affected by networking failures.
+
+- Mesos-level health checks delegated to the agents running the tasks, so the number of tasks that can be health checked can scale horizontally with the number of agents in the cluster.
```
timcharper: health check *are* delegated

timcharper: Should there be a `-` when using health check as a verb?

> so the number of tasks that can be health-checked

I'm not a pro tech writer, maybe it is fine as is :) this just seemed strange to me.

docs/docs/health-checks.md
```diff
@@ -16,9 +16,123 @@ application's tasks.
...
+
+#### Limitations and considerations
+
+- Mesos-level health checks consume extra resources on the agents; moreover, there is some overhead for fork-execing a process and entering the tasks’ namespaces every time a task is checked.
```
timcharper: I understand there are plans to improve the efficiency here. Confirm with Mesos team? If so, perhaps:

> there is (currently) some overhead

docs/docs/health-checks.md
```diff
@@ -16,9 +16,123 @@ application's tasks.
...
+
+- The health check processes share resources with the task that they check. Your application definition must account for the extra resources consumed by the health checks.
+
+- Mesos-level health checks require tasks to listen on the loopback interface in addition to whatever interface they require. If you run a service in production, you will want to make sure that the users can reach it.
```
timcharper: The loopback interface in a container is not the same as the host, unless if the container is not using network isolation.

As a user, I might be confused as to whether it needs to listen on the hosts loopback interface. Do you think this is a worth-while clarification?

> on the *container's* loopback interface

docs/docs/health-checks.md
```diff
@@ -16,9 +16,123 @@ application's tasks.
...
+
+#### `COMMAND` health checks
+
+You must escape any double quotes in your commands. This is required because Mesos runs the healthcheck command inside via `/bin/sh -c ""`.
+  See the example below and [MESOS-4812](https://issues.apache.org/jira/browse/MESOS-4812) for details.
```
timcharper: Ugh. This is awful. Good to mention.

docs/docs/health-checks.md
sascala: Yeah... this might be more relevant to Mesos documentation

docs/docs/health-checks.md
sascala: I see your point

docs/docs/health-checks.md
sascala: I probably shouldn't be verbifying the word in the first place.

docs/docs/health-checks.md
sascala: 👍

docs/docs/health-checks.md
sascala: yes :-)

PR #5453, opened 2017-08-02
Only leader reports metrics

fengyehong:


janisz:
I prefer all instances report metrics in case of debugging problems with standby instances e.g. huge cpu/mem usage after changing the leader.

PR #5454, opened 2017-08-02
Use NonFatal instead of Throwable in WorkQueue.

jeschkies:
This fixes the scapegoat error on releases/1.4.


jeschkies:
The CI times out cloning the repository. I filed a ticket.

timcharper:
thanks, @jeschkies ; I should not have hit merge so late at night.

I think I would actually prefer to disable the scapegoat warning.

PR #5459, opened 2017-08-09
Change unreachableStrategy to be able to start instant replacement tasks

unterstein:
Test Plan:
sbt test
sbt integration:test

Reviewers: jeschkies, zen-dog, jenkins, ichernetsky, kensipe

Reviewed By: jeschkies, zen-dog, jenkins, ichernetsky, kensipe

Subscribers: kensipe, meichstedt, marathon-dev, marathon-team

JIRA Issues: MARATHON_EE-1590

Differential Revision: https://phabricator.mesosphere.com/D963

PR #5460, opened 2017-08-09
Allow migration from previous UnreachableStrategy default

unterstein:
Summary:
To change the `unreachableStrategy` of all apps and pods, you can define the environment parameter `MIGRATION_1_4_6_UNREACHABLE_STRATEGY` which leads to the following behavior during migration:

When opting in to the unreachable migration step
1) all app and pod definitions that had a config of `UnreachableStrategy(300 seconds, 600 seconds)` (previous default) are migrated to have `UnreachableStrategy(0 seconds, 0 seconds)`
2) all app and pod definitions that had a config of `UnreachableStrategy(1 second, x seconds)` are migrated to have `UnreachableStrategy(0 seconds, x seconds)`
3) all app and pod definitions that had a config of `UnreachableStrategy(1 second, 2 seconds)` are migrated to have `UnreachableStrategy(0 seconds, 0 seconds)`

Depends D963

Test Plan:
test migration manually
sbt test

Reviewers: zen-dog, jeschkies, jenkins, ichernetsky

Reviewed By: jeschkies, jenkins

Subscribers: meichstedt, marathon-dev, marathon-team

JIRA Issues: MARATHON_EE-1591

Differential Revision: https://phabricator.mesosphere.com/D970

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,16 @@ unreachableStrategy: {
 **Note**: Instantly means as soon as marathon becomes aware of the unreachable task. By default marathon is notified after 75 seconds by mesos
   that an agent is disconnected. You can change this duration in mesos by configuring `agent_ping_timeout` and `max_agent_ping_timeouts`.
 
+#### Migrating unreachableStrategy
+If you want all of your apps and pods to adopt a `UnreachableStrategy` that retains the previous behavior where instance were immediately replaced so that you does not have to update every single app definition.
+
+To change the `unreachableStrategy` of all apps and pods, you can define the environment parameter `MIGRATION_1_4_6_UNREACHABLE_STRATEGY` which leads to the following behavior during migration:
+
+When opting in to the unreachable migration step
+1) all app and pod definitions that had a config of `UnreachableStrategy(300 seconds, 600 seconds)` (previous default) are migrated to have `UnreachableStrategy(0 seconds, 0 seconds)`
+2) all app and pod definitions that had a config of `UnreachableStrategy(1 second, x seconds)` are migrated to have `UnreachableStrategy(0 seconds, x seconds)`
+3) all app and pod definitions that had a config of `UnreachableStrategy(1 second, 2 seconds)` are migrated to have `UnreachableStrategy(0 seconds, 0 seconds)`
+
 ### Fixed issues
 
 - [MARATHON-7681](https://jira.mesosphere.com/browse/MARATHON-7681) - Fixes an issue in WorkQueue that could cause Marathon to drop exceptions and become unresponsive.
--- a/src/main/scala/mesosphere/marathon/storage/StorageModule.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageModule.scala
@@ -76,7 +76,7 @@ object StorageModule {
         val eventSubscribersStore = l.entityStore[EventSubscribers] _
         val eventSubscribersRepository = EventSubscribersRepository.legacyRepository(eventSubscribersStore)
 
-        val migration = new Migration(l.availableFeatures, legacyConfig, None, appRepository, groupRepository,
+        val migration = new Migration(l.availableFeatures, legacyConfig, None, appRepository, podRepository, groupRepository,
           deploymentRepository, taskRepository, instanceRepository, taskFailureRepository,
           frameworkIdRepository, eventSubscribersRepository)
 
@@ -107,7 +107,7 @@ object StorageModule {
             Nil
         }
 
-        val migration = new Migration(zk.availableFeatures, legacyConfig, Some(store), appRepository, groupRepository,
+        val migration = new Migration(zk.availableFeatures, legacyConfig, Some(store), appRepository, podRepository, groupRepository,
           deploymentRepository, taskRepository, instanceRepository, taskFailureRepository,
           frameworkIdRepository, eventSubscribersRepository)
         StorageModuleImpl(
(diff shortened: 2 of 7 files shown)


ShaneSaww:
@unterstein  I just pulled this branch + https://github.com/mesosphere/marathon/pull/5459 into my test environment, and the migration failed to update all of my currently running applications.  It looks like it only grabbed 1 of my applications, and migrated it. I have never set the `unreachableStrategy` in the past, so all of my apps appear to have the default 300/600 values. 

This is running on DC/OS 1.9.2 (with the patched version of marathon noted above)

unterstein:
Thanks @ShaneSaww for the feedback, this is very interesting.

I just did a migration from marathon 1.4.5 (not latest on releases/1.4 branch, because version is already to `1.4.6-snap1`) to the this branch with the commit of #5459 and I changed the `version.sbt` to hold `1.4.6`. And don`t forget to set the env var (which I dit in the first place).
I got this:

`/v2/apps` before migration (cleaned it up a little bit):

```
{  
   "apps":[  
      {  
         "id":"/default",
         "cmd":"sleep 1000",
         "instances":1,
         "cpus":0.1,
         "mem":128,
         "disk":0,
         "gpus":0,
         "unreachableStrategy":{  
            "inactiveAfterSeconds":300,
            "expungeAfterSeconds":600
         }
      },
      {  
         "id":"/hack-with-expunge",
         "cmd":"sleep 1000",
         "instances":1,
         "cpus":0.1,
         "mem":128,
         "disk":0,
         "gpus":0,
         "unreachableStrategy":{  
            "inactiveAfterSeconds":1,
            "expungeAfterSeconds":100
         }
      },
      {  
         "id":"/hack",
         "cmd":"sleep 1000",
         "instances":1,
         "cpus":0.1,
         "mem":128,
         "disk":0,
         "gpus":0,
         "unreachableStrategy":{  
            "inactiveAfterSeconds":1,
            "expungeAfterSeconds":2
         }
      }
   ]
}
```

`/v2/apps` after migration:

```
{  
   "apps":[  
      {  
         "id":"/default",
         "cmd":"sleep 1000",
         "instances":1,
         "cpus":0.1,
         "mem":128,
         "disk":0,
         "gpus":0,
         "unreachableStrategy":{  
            "inactiveAfterSeconds":0,
            "expungeAfterSeconds":0
         }
      },
      {  
         "id":"/hack-with-expunge",
         "cmd":"sleep 1000",
         "instances":1,
         "cpus":0.1,
         "mem":128,
         "disk":0,
         "gpus":0,
         "unreachableStrategy":{  
            "inactiveAfterSeconds":0,
            "expungeAfterSeconds":100
         }
      },
      {  
         "id":"/hack",
         "cmd":"sleep 1000",
         "instances":1,
         "cpus":0.1,
         "mem":128,
         "disk":0,
         "gpus":0,
         "unreachableStrategy":{  
            "inactiveAfterSeconds":0,
            "expungeAfterSeconds":0
         }
      }
   ]
}
```

BUT: I got this after restarting marathon. Unfortunately we currently having a caching problem during taking leadership (migration) and therefore this is refreshed when you restart, see this ticket for more context: https://github.com/mesosphere/marathon/pull/5461

sethp-nr:
@unterstein I was working with @ShaneSaww on this issue, and we caused the migration to run multiple times by resetting the version stored in Zookeeper by way of exhibitor and restarting all of our marathon processes. Each time it ran, we saw it report successfully modifying a single application without touching any of the others (that had existed pre-1.4.2 migration).

We took the "data bytes" from exhibitor for one of those applications and saw the following (after multiple migration runs):

```
$ pbpaste | tr -d ' ' | xxd -r -p | protoc --decode_raw
1: "/our/cool/path/to-helloworld-service"
<snip>
27 {
  1: 10085
  3: "tcp"
}
32 {
  1: 300
  2: 600
}
33: 1
```

For reference, the relevant snippet of [`marathon.proto`](https://github.com/mesosphere/marathon/blob/master/src/main/proto/marathon.proto#L100-L147) seems to be:

```
message ServiceDefinition {
  required string id = 1;
  <...>
  optional UnreachableStrategy unreachableStrategy = 32;
  optional KillSelection killSelection = 33 [ default = YoungestFirst ];
}

message UnreachableStrategy {
  // UnreachableDisabled is represented by both of these fields missing
  optional uint64 inactiveAfterSeconds = 1 [default = 900 ]; // 15 minutes
  optional uint64 expungeAfterSeconds = 2 [default = 604800 ]; // 7 days
}
```

Which suggested to me that we were being filtered out by way of [`_.unreachableStrategy != UnreachableDisabled`](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo146.scala#L99), even though I can't quite parse why that would be the case.

We then modified the upgrade task to remove the filter on`unreachableStrategy`, and doing so caused all of our applications to pass through the upgrade process. That destroyed our sample from above, but did validate that for some reason what `protoc` output as `32 { 1: 300  2: 600 }` was being decoded as `UnreachableDisabled`.

meichstedt:
@sethp-nr @ShaneSaww have you created an issue in our Marathon JIRA? What you report sounds like a severe bug we should look at.

sethp-nr:
@meichstedt at your behest, I've filed https://jira.mesosphere.com/browse/MARATHON-7704 , but we've had a lot more generalized difficulty with the behavior changes in 1.4 than is easily distilled to actionable bugs.

I've written up some of that over here: https://groups.google.com/forum/#!topic/marathon-framework/XrWbd-W-TM0 . Your thoughts and feedback would be much appreciated.

PR #5461, opened 2017-08-09
Reset root group cache when elected as leader

unterstein:
Summary:
GroupRepository are holding in memory caches of the root group. The cache is loaded when it is accessed the first time.
Therefore the root group state is loaded from zk when the marathon instance is started.
When the marathon instance is elected as leader, this cache is still in the same state as the time marathon started.
Therefore we need to re-load the root group from zk again from zookeeper when becoming leader.
The same is true after doing the migration. A migration or a restore also affects the state of zookeeper, but does not update the internal hold caches. Therefore we need to refresh the internally loaded caches after the migration.
Actually we need to do the fresh twice, before the migration, to perform the migration on the current zk state and after the migration to have marathon loaded the current valid state to the internal caches.

Test Plan:
sbt integration:test

Reviewers:
jeschkies, timcharper, zen-dog, jenkins

Reviewed By:
jeschkies, zen-dog, jenkins

Subscribers:
jdef, marathon-dev, marathon-team

JIRA Issues:
MARATHON-7565, MARATHON-7458, MARATHON-7401

Differential Revision:
https://phabricator.mesosphere.com/D847
https://phabricator.mesosphere.com/D895

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
@@ -204,6 +204,9 @@ class MarathonSchedulerService @Inject() (
   override def startLeadership(): Unit = synchronized {
     log.info("As new leader running the driver")
 
+    // refresh group repository cache
+    Await.result(groupManager.refreshGroupCache(), Duration.Inf)
+
     // execute tasks, only the leader is allowed to
     migration.migrate()
 
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
@@ -204,11 +204,7 @@ class MarathonSchedulerService @Inject() (
   override def startLeadership(): Unit = synchronized {
     log.info("As new leader running the driver")
 
-    // refresh group repository cache
-    Await.result(groupManager.refreshGroupCache(), Duration.Inf)
-
-    // execute tasks, only the leader is allowed to
-    migration.migrate()
+    refreshCachesAndDoMigration()
 
     // run all pre-driver callbacks
     log.info(s"""Call preDriverStarts callbacks on ${prePostDriverCallbacks.mkString(", ")}""")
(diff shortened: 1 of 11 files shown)


zen-dog:
This is not quite a backport since `GroupManager` is still an actor in 1.4. We ended up calling `invalidateGroupCache` in the `Migration` class itself. Otherwise it wouldn't work.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -204,7 +204,18 @@ class MarathonSchedulerService @Inject() (
...
 
-    // execute tasks, only the leader is allowed to
+    // GroupRepository is holding in memory caches of the root group. The cache is loaded when it is accessed the first time.
+    // Actually this is really bad, because each marathon will log the amount of groups during startup through metrics.
+    // Therefore the root group state is loaded from zk when the marathon instance is started.
+    // When the marathon instance is elected as leader, this cache is still in the same state as the time marathon started.
+    // Therefore we need to re-load the root group from zk again from zookeeper when becoming leader.
+    // The same is true after doing the migration. A migration or a restore also affects the state of zookeeper, but does not
+    // update the internal hold caches. Therefore we need to refresh the internally loaded caches after the migration.
+    // Actually we need to do the fresh twice, before the migration, to perform the migration on the current zk state and after
```
jdef: s/fresh/refresh/

src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
```diff
@@ -3,15 +3,15 @@ package core.group
...
-import akka.NotUsed
 import akka.stream.scaladsl.Source
 import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.state.{ AppDefinition, Group, PathId, RootGroup, RunSpec, Timestamp }
-import mesosphere.marathon.upgrade.DeploymentPlan
 
 import scala.concurrent.Future
 import scala.collection.immutable.Seq
+import akka.NotUsed
```
jdef: why move around the imports here? seems unrelated

unterstein:
Thanks for the feedback! Going to land this now.

src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
```diff
@@ -8,10 +8,10 @@ import akka.stream.scaladsl.Source
 import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.state.{ AppDefinition, Group, PathId, RootGroup, RunSpec, Timestamp }
-import mesosphere.marathon.upgrade.DeploymentPlan
 
 import scala.concurrent.Future
 import scala.collection.immutable.Seq
+import mesosphere.marathon.upgrade.DeploymentPlan
```
jdef: unclear why this commit was moved. was hoping that the last comment I left related to imports would have sufficed for both moved statements. the whole point was to avoid changes to this file that were unnecessary.

unterstein: because code style re-organizes imports alphabetically and before we annoy everyone reverting these changes before commit, it makes sense to have them in alphabetic order imo

PR #5463, opened 2017-08-12
remove broken / obsolete badges

timcharper:
Our pipeline job badge for 1.4 is not working; removing, at least for
now.

We no longer use waffle. Remove it, also.


timcharper:
🙇

PR #5464, opened 2017-08-12
Workaround for Mesos in Debian Stretch

timcharper:
openjdk:8-jdk has switched to Debian Stretch. However, there is not a
repository for Debian Stretch. We work-around by installing
libssl1.0.0 from Jessie backports.

See https://github.com/mesosphere/mesos-deb-packaging/issues/109

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,21 +6,19 @@
 # It will reresolve all dependencies on every change (as opposed to Dockerfile.development)
 # but it ultimately results in a smaller docker image.
 #
-FROM openjdk:8-jdk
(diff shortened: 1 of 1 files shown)


timcharper:
I manually tested Mesos client SSL support with this docker image and it does not work; it seg-faults when openssl is used. I think we will need to stick with Debian Jessie.

timcharper:
@jdef right - it's not needed. The base image for debian jessie has systemd

timcharper:
Oh, and to be clear, we've been installing the Mesos Debian Jessie packages all along; that was the main problem.

PR #5465, opened 2017-08-14
Fix `TaskUnreachableIntegrationTest` cherry-pick by matching appId string

zen-dog:
...instead of the `PathId`


jeschkies:
> sigh; Dotty is suppose to catch these types of comparison issues.

That'd be Scala 3, right? I really wonder why Scala 2.x cannot catch these.

PR #5467, opened 2017-08-14
Add extra clarification for changelog

timcharper:
- It was unclear that a value of "true" is needed (a reading of the code
  indicates this is the case)
- I think people may assume that if they run the migration, the default
  could be changed. Add text to make clear that is not currently the
  case.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,28 +1,39 @@
 ## Changes since 1.4.5
 
 ### Overview
+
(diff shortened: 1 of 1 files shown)


zen-dog:
Why was this only done for 1.4? We have this changelog notes in master too. @timcharper

timcharper:
@zen-dog the text on master has been updated, also.

PR #5468, opened 2017-08-15
No `Network.name` in host mode

sascala:


docs/docs/networking.md
```diff
@@ -33,6 +33,8 @@ Three networking modes are supported:
 In `host` networking, an application shares the network namespace of the Mesos agent
 process, typically the host network namespace.
 
+**Note:** The `Network.name` parameter cannot be used with `host` networking.
```
jdef: the `name` parameter shouldn't be used with `host`, nor should it be used with `container/bridge`

it's only useful for `container`

sascala: got it

sascala:
@jdef Thanks for your comment. I corrected my note.

PR #5469, opened 2017-08-15
Don't re-use taskIds for resident tasks

meichstedt:
Summary:
This introduces the notion of launch attempts for resident tasks. With this change, every task that is being launched on an existing reservation will use a new taskId, which is derived from the previous one, and that has an incremented attempt counter. Matching reservations and volumes is done against the base reservationId, which is basically the taskId without separator and attempt count. There is no migration needed, but consumers of taskIds need to be aware that newly launched resident tasks will now have an attempt appended to the ID:
Resources will be labeled with `app.b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6`, while the Mesos tasks will have IDs like `app.b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6.1`, `app.b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6.2` and so on.

There are additional changes in this set that consolidate the usage of task vs instance, e.g. in Tasklabels.

backport of 26ca21ca38bfffb01bd2592acf8a46cab8990454
includes cherry-pick of d19afde665a85af445ac4d8c57fd0f571c36f86e

JIRA Issues: MARATHON_EE-1603, MARATHON_EE-1617, MARATHON_EE-1618

Conflicts:
	src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
	src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
	src/main/scala/mesosphere/mesos/TaskBuilder.scala
	src/test/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolverTest.scala
	src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
	src/test/scala/mesosphere/marathon/core/launcher/impl/TaskLabelsTest.scala
	src/test/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactoryTest.scala
	src/test/scala/mesosphere/marathon/core/task/TaskTest.scala
	src/test/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImplTest.scala
	src/test/scala/mesosphere/marathon/state/AppDefinitionPortAssignmentsTest.scala
	src/test/scala/mesosphere/marathon/tasks/TaskIdTest.scala
	src/test/scala/mesosphere/mesos/PersistentVolumeMatcherTest.scala
	src/test/scala/mesosphere/mesos/TaskBuilderTest.scala


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -121,6 +131,10 @@ object Task {
     // Regular expression for matching taskIds before instance-era
     private val LegacyTaskIdRegex = """^(.+)[\._]([^_\.]+)$""".r
 
+    // Regular expression for matching resident app taskIds
+    private val ResidentTaskIdRegex = """^(.+)([\._])([^_\.]+)(\.\#)(\d+)$""".r
+    private val ResidentTaskIdAttemptSeparator = ".#"
```
jdef: wondering if there's any impact to things that might use taskID in http query params since the `#` has special meaning in http URLs (for web browsers)

ichernetsky: Good point.

jdef: I've not used angular, but it looks like `task_id` is part of a "route"...

https://github.com/apache/mesos/blob/8447d196a8d2385d8dd742c5d6a8ac62de9a59fe/src/webui/master/static/js/app.js#L48

.. and stackoverflow says that hashtags are special?

https://stackoverflow.com/questions/27226009/is-angularjs-routes-adding-a-special-character

meichstedt: Thanks for pointing out @jdef 
Fixed in a diff on master and including the cherry-picked fix in this PR now

PR #5475, opened 2017-08-16
Import Jenkins pipeline for Marathon 1.4.

jeschkies:


jeschkies:
Since we move on to 1.6 we can drop this.

PR #5478, opened 2017-08-18
[WIP] Update AgentInfo when launching on reservation

meichstedt:
Background: When an agent (e.g. "agent-1") reboots, the `agentId` might change under certain circumstances (to, e.g. "agent-2"). If that agent's state contained dynamic reservations and persistent volumes, Marathon would not update the AgentInfo when launching a new instance based on these. As a result, the instance as reported via the Marathon API would still say "agent-1", when the actual task and the persistent volume would run on "agent-2"; the volume and task appear to be on different nodes.

This commit will update the AgentInfo for each LaunchOnReservation.

Fixes MARATHON-7707

(I'll provide a test and remove the WIP)


meichstedt:
@jeschkies I did; I'll create a diff to have the CI run …

timcharper:
I landed the phabricator diff

PR #5481, opened 2017-08-18
Consider Reserved instances for validating constraints.

meichstedt:
Summary:
When matching offers for resident tasks, Marathon would not consider instances in state Reserved when validating constraints. This potentially resulted in multiple reservations on the same agent, even if a hostname: unique constraint was applied to the run spec. The problem boils down to the fact that the ResourceMatcher filtered for `instance.isLaunched` and only considered those instances. This was accidentally introduced via the fact that previously, instances which weren't launched had no runSpecVersion and therefore could not be compared to the last config change.

Compare 3bf03f86, which needed to access `_.launched` to compare the version. Deriving the need to check for `isLaunched` later was a wrong assumption.

While at it, StrictLogging has been introduced and unnecessary `if(xyzEnabled)` statements have been removed.

Added two unit tests for offer matching with unique constraints

Backporting df5e2e7c8424 onto 1.4

Test Plan: sbt test

JIRA Issues: MARATHON-7706, COPS-598

Original Differential Revision: https://phabricator.mesosphere.com/D1003

Conflicts:
	src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
	src/main/scala/mesosphere/mesos/ResourceMatcher.scala
	src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
	src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -2,6 +2,7 @@ package mesosphere.marathon
 package core.launcher.impl
 
 import mesosphere.marathon.core.base.Clock
+import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.instance.Instance.{ AgentInfo, InstanceState }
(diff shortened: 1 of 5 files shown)


timcharper:
Manually tested against a cluster locally; confirmed it works.

PR #5484, opened 2017-08-21
Do not suppress offers.

timcharper:
Summary:
Marathon requires offers not only to launch instances but also to free
reserved resources. In some situation the `ReviveOffersActor` would not
revive offers in time after a suppress has been issued. This change
removes the suppress completely. The original reason for suppressing
offers is to throttle them and not overload Marathon. However, Marathon
already declines offers with a time period. This is an implicit
throttling.

This diff is mainly intended for 1.4 and might be superseded by D941.

Backport of bb6420c

Test Plan: pipeline.

Reviewers: unterstein, zen-dog, timcharper, jenkins, kensipe

Reviewed By: timcharper, jenkins, kensipe

Subscribers: marathon-dev, marathon-team, jenkins

JIRA Issues: MARATHON-7711, MARATHON-7338

Original Differential Revision: https://phabricator.mesosphere.com/D1005

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
@@ -21,7 +21,7 @@ private[flow] object ReviveOffersActor {
   }
 
   private[impl] case object TimedCheck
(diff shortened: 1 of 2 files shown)


timcharper:
Tested in local cluster; confirmed that the fix works.

PR #5501, opened 2017-09-08
Add 'forceIp' to force use IP from "IpAddressess" for containers (for marathon 1.3)

ikuzniatsou:
Add boolean "forceIp" parameter for containers that allows use ip (Task.effectiveIp) from "ipAddressess" without conditions.

It resolves issue when using docker container with custom network:
Marathon try to get incorrect "agentInfo.host" for ip (host of container's custom network)  instead of IpAddressess

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -85,7 +85,7 @@ sealed trait Task {
   }
 
   def effectiveIpAddress(runSpec: RunSpec): Option[String] = {
-    if (runSpec.ipAddress.isDefined)
+    if (runSpec.container.exists(_.forceIp) || runSpec.ipAddress.isDefined)
       launched.flatMap(_.ipAddresses).flatMap(_.headOption).map(_.getIpAddress)
     else
       Some(agentInfo.host)
--- a/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
+++ b/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
@@ -99,6 +99,10 @@
               "type": "boolean",
               "description": "The container will be pulled, regardless if it is already available on the local system."
             },
+            "forceIp": {
+              "type": "boolean",
+              "description": "The container will use values from ipAddressess ."
+            },
             "image": {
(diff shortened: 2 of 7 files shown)


ichernetsky:
@meichstedt @jdef could you please review it?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -85,7 +85,7 @@ sealed trait Task {
   }
 
   def effectiveIpAddress(runSpec: RunSpec): Option[String] = {
-    if (runSpec.ipAddress.isDefined)
+    if (runSpec.container.exists(_.forceIp) || runSpec.ipAddress.isDefined)
```
jdef: referring to `forceIp` isn't necessary here. For apps that use docker containers, it may be inferred that custom (USER) networking is in use by examining the network mode of the docker container.

timcharper:
Hi @ikuzniatsou ! Could you open up a ticket describing the motive and proposal for this change? Please include the app definition and any information about the custom network that you are using.

BTW - I once worked around this myself when I launched containers using MacVLAN... set `"ipAddress": {}`. It was a work-around that worked, but, I think enriching the logic so that Marathon is more aware of the Docker container network specified is the right way to go.

ikuzniatsou:
Hi @timcharper @jdef  !

Sorry, maybe I didn't  quite clearly describe situation.

For now 1.3 is tested and worked properly in our environment, migration to 1.4\1.5  is planned but not now. 
So, it was look like the best way was a quick fix for 1.3 only. 



About issue:  we have a host-machine with it's own IP.
There is a bunch of docker containers ( mesos-agents ) started on it. Each of these containers (agents) have it's own IP as required in preconfigured docker network policy.  Agents are registered in mesos with it's container-IP.
It's work perfectly except health-check in marathon 1.3. 

( Yes, Health-checks works with 1.5, but as I said above, we can't migrate now...)

Sorry, I can't remember exactly reason why is this so, but marathon 1.3 does not receive runSpec.ipAddress variable in this case. So, as provied in health-check logic, it does not use IP from ipAddresses variable (with correct IP of container (agent), but get IP of host-machine.

I hope, it's more clearly now.

PR #5504, opened 2017-09-13
Constraints doc edits

sascala:
https://jira.mesosphere.com/browse/MARATHON-7193

### DIFF
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -30,20 +30,21 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
 
 ### CLUSTER operator
 **Value** (optional): A string value.
-When a value is specified then tasks are launched on agent nodes with an attribute whose value matches exactly.
+When a value is specified, tasks are launched on agent nodes with an attribute whose value matches exactly.
 
 `CLUSTER` allows you to run all of your app's tasks on agent nodes that share a certain attribute.
-This is useful for example if you have apps with special hardware needs, or if you want to run them on the same rack for low latency.
+This is useful, for example, if you have apps with special hardware needs, or if you want to run them on the same rack for low latency.
 
-A field of `"hostname"` tells Marathon that launched tasks of the app/pod have affinity for each other and should be launched together on the same agent:
-* When value is specified then tasks are launched on the agent whose hostname matches the value.
-* When value is empty or unspecified then the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
+The <hostname> field tells Marathon that the launched tasks of the app or pod should be launched together on the same agent:
+* When a value for <hostname> is specified, tasks are launched on the agent whose hostname matches the value.
+* WWhen the <hostname> value is empty or unspecified, the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
 
 Attribute fields are handled differently:
-* When value is specified then tasks are launched on any agent with an attribute named according the field **and** with a value matching that of the constraint.
-* When value is empty or unspecified then the first instance is launched on any agent with an attribute named according to the field; the value of the attribute on that agent is used for future constraint matches.
+* When attribute fields are specified, tasks are launched on any agent with an attribute named according the field **and** with a value matching that of the constraint.
+* When attribute fields are empty or unspecified, the first instance is launched on any agent with an attribute named according to the field; the value of the attribute on that agent is used for future constraint matches.
 
-You could specify the exact rack on which to run app tasks:
+## Examples
+The example below specifies the exact rack on which to run app tasks:
 
 ``` bash
 $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
@@ -54,7 +55,7 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
-Alternatively, you could specify that all of the app tasks should run on the same rack, but without specifying which rack:
+This example leaves the attribute field empty. This tells Marathon that all of the app tasks should run on the same rack, but does not specify which.
 
 ``` bash
 $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
@@ -65,7 +66,7 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
-You can also tie an app to a specific node by using the `hostname` field:
+This example uses the `hostname` field to specify that an app must run on a specific node.
 
 ``` bash
 $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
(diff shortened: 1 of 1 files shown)


docs/docs/constraints.md
```diff
@@ -30,20 +30,21 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
-* When value is empty or unspecified then the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
+The <hostname> field tells Marathon that the launched tasks of the app or pod should be launched together on the same agent:
+* When a value for <hostname> is specified, tasks are launched on the agent whose hostname matches the value.
+* WWhen the <hostname> value is empty or unspecified, the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
 
 Attribute fields are handled differently:
-* When value is specified then tasks are launched on any agent with an attribute named according the field **and** with a value matching that of the constraint.
-* When value is empty or unspecified then the first instance is launched on any agent with an attribute named according to the field; the value of the attribute on that agent is used for future constraint matches.
+* When attribute fields are specified, tasks are launched on any agent with an attribute named according the field **and** with a value matching that of the constraint.
```
jdef: I really did mean `value` and not `attribute`. "When a value is specified, tasks are launched ..."

docs/docs/constraints.md
```diff
@@ -30,20 +30,21 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
+* When attribute fields are empty or unspecified, the first instance is launched on any agent with an attribute named according to the field; the value of the attribute on that agent is used for future constraint matches.
```
jdef: "When the value is empty or unspecified ..."

docs/docs/constraints.md
```diff
@@ -54,7 +55,7 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
-Alternatively, you could specify that all of the app tasks should run on the same rack, but without specifying which rack:
+This example leaves the attribute field empty. This tells Marathon that all of the app tasks should run on the same rack, but does not specify which.
```
jdef: "This example leaves the value field empty."

Because the attribute is "rack_id"

docs/docs/constraints.md
jdef: Maybe this is confusing because "Hostname field" and "Attribute field" are bad section names? Maybe a better way to explain it would be:
```
FIELD := "hostname" | ATTRIBUTE      // FIELD is the literal "hostname" or else some ATTRIBUTE
ATTRIBUTE := {some mesos agent attribute name}
```

docs/docs/constraints.md
```diff
@@ -76,7 +77,7 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
-Or in a similar way, you can run all app tasks together on the same node but without specifying which agent to use:
+Below, the attribute field is empty. This tells Marathon that all app tasks must run on the same node, but does not specify which.
```
jdef: "Below, the value field is empty."

sascala:
Thanks, @jdef

docs/docs/constraints.md
```diff
@@ -30,20 +30,21 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
+When a value is specified, tasks are launched on agent nodes with an attribute whose value matches exactly.
 
 `CLUSTER` allows you to run all of your app's tasks on agent nodes that share a certain attribute.
-This is useful for example if you have apps with special hardware needs, or if you want to run them on the same rack for low latency.
+This is useful, for example, if you have apps with special hardware needs, or if you want to run them on the same rack for low latency.
 
-A field of `"hostname"` tells Marathon that launched tasks of the app/pod have affinity for each other and should be launched together on the same agent:
-* When value is specified then tasks are launched on the agent whose hostname matches the value.
-* When value is empty or unspecified then the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
+The <hostname> field tells Marathon that the launched tasks of the app or pod should be launched together on the same agent:
```
jdef: I'm still confused by this syntax:
* `The <hostname> field...`
* `The <hostname> value...`

This reads like there's a `<hostname>` element has a field and value. Which is not the case. A constraint is composed of the following:
* `field`: is the subject being tested. `"hostname"` (literally), or else is a string that reflects an attribute name of a mesos agent
  - "a field of `hostname`" means "... when the field is specified as `'hostname'`"
  - I don't think I was consistent here when I rewrote parts of this doc .. and I should have been
* `operator`: the test being performed
* `value`: what the field is being tested with/against

I am open to suggestions w/ respect to better ways to write about this stuff. I just want to make sure we're on the same page first.

docs/docs/constraints.md
```diff
@@ -30,20 +30,21 @@ Regex is allowed for LIKE and UNLIKE operators; to match ANY value, use the stri
...
-* When value is specified then tasks are launched on the agent whose hostname matches the value.
-* When value is empty or unspecified then the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
+Naming the field name `hostname` tells Marathon that the launched tasks of the app or pod should be launched together on the same agent:
+* When a value for `hostname` is specified, tasks are launched on the agent whose hostname matches the value.
+* When the `hostname` value is empty or unspecified, the first instance is launched on **any** agent node and the remaining tasks are launched alongside it on the same agent.
 
-Attribute fields are handled differently:
-* When value is specified then tasks are launched on any agent with an attribute named according the field **and** with a value matching that of the constraint.
-* When value is empty or unspecified then the first instance is launched on any agent with an attribute named according to the field; the value of the attribute on that agent is used for future constraint matches.
+When the field name is an attribute, Mararthon handles tasks differently:
```
ichernetsky: `Mararthon` -> `Marathon`.

docs/docs/constraints.md
```diff
@@ -65,7 +66,7 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
-You can also tie an app to a specific node by using the `hostname` field:
+This example uses the specifies a `hostname`, which means an app must run on a specific node.
```
ichernetsky: I think it should be `The example uses the specified hostname`.

ichernetsky:
@jdef could you review it again? Thanks in advance!

docs/docs/constraints.md
```diff
@@ -6,21 +6,21 @@ title: Constraints
...
-The `hostname` field matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
+Entering `hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
 
-`hostname` field supports all operators of Marathon.
+All Marathon operators are supported when the field name is `hostname`. 
 
-### Attribute field
+### Attribute as field name
 
 If the field name is not the agent node hostname, it will be treated as a Mesos agent node attribute. A Mesos agent node attribute allows you to tag an agent node. See `mesos-slave --help` to learn how to set the attributes.
```
jdef: "If the field name is not the agent node hostname..."

should perhaps be

"If `hostname` is not specified as the field name then the field name is interpreted as a Mesos agent node attribute."

sascala:
Changes made. Thank you, @jdef and @ichernetsky

PR #5505, opened 2017-09-13
mesos_role clarification

sascala:
https://jira.mesosphere.com/browse/MARATHON-7359


docs/docs/command-line-flags.md
```diff
@@ -78,7 +77,7 @@ The core functionality flags can be also set by environment variable `MARATHON_O
     service ports to apps. If you assign your service port statically in your app definition, it does
     not have to be in this range.
 * `--mesos_role` (Optional. Default: None): Mesos role for this framework. If set, Marathon receives resource offers
-    for the specified role in addition to resources with the role designation '*'.
+    for the specified role in addition to resources with the role designation '*'. This parameter is only applied the first time the framework registers with Mesos.
```
meichstedt: I think this could even be more clear:
> If you change this parameter to another value, or leave it out, this will not have an effect on the role for which Marathon is registered.

PR #5508, opened 2017-09-15
Set a fixed ephemeral port range to avoid bing errors.

jeschkies:
Summary:
We would sometimes see binding errors for the forwarder test services or the
health check endpoint. Mesos launches executor with port 0 thus an executor
could block the port for the mentioned services.

This change will set the port range for port 0 outside of the ports assigned to
our test services.

Note that `set_port_range.sh` as to be added to any loop.

Test Plan: loop

Reviewers: unterstein, zen-dog, ichernetsky, timcharper, kensipe, meln1k, jenkins

Reviewed By: zen-dog, timcharper, meln1k, jenkins

Subscribers: marathon-team, marathon-dev, jenkins

Differential Revision: https://phabricator.mesosphere.com/D1058

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,6 +48,7 @@ node('JenkinsMarathonCI-Debian8') {
         }
         stage("Provision Jenkins Node") {
             sh "sudo -E ci/provision.sh"
+            sh "sudo -E ci/set_port_range.sh"
         }
         stageWithCommitStatus("1. Compile") {
           try {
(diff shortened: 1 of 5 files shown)


jeschkies:
This PR depends on #5509 and should only be merged or rather rebased with

```
git rebase karsten/1.4/backport/D1058 karsten/1.4/fix-provisioning master
```

after `karsten/1.4/fix-provisioning master` has been merged into `master`.

PR #5516, opened 2017-09-20
Add issue template pointing people towards JIRA

meichstedt:
Github issues are still enabled for reference and to keep links alive. Instead of having people write a nicely formatted issue in markdown syntax and then letting them see it being closed, locked and migrated to JIRA, we should directly hint that they might consider creating an issue directly in JIRA.

We should consider adding a pull request template as well, and move github related files to a `.github` folder later. See https://github.com/blog/2111-issue-and-pull-request-templates for reference.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,5 +1,5 @@
 ## Head over to JIRA!
 
-Please note that we use JIRA for issue tracking and only keep GitHub issues around to not break existing links. Please prefer filing issues [directly in JIRA](https://jira.mesosphere.com/secure/CreateIssue!default.jspa?pid=10401).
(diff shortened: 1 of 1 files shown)


janisz:
Github issues created by a bot are automatically assigned and has some labels. Is it important for the issue to have this labels and to be assigned?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+## Head over to JIRA!
+
+Please note that we use JIRA for issue tracking and only keep GitHub issues around to not break existing links. Please prefer filing issues [directly in JIRA](https://jira.mesosphere.com/secure/CreateIssue!default.jspa?pid=10401).
+
+You can sign in to JIRA using github or google as identity providers.
```
ichernetsky: `Github` and `Google`?

meichstedt: Will do ;)

meichstedt:
@janisz Actually, no. The labels were used to find issues created in Github (a.k.a by the community), but we try to process issues regardless of their origin now … which seems politically more correct, and also more helpful for us. We can differentiate in JIRA wrt the permissions of the user (Mesosphere employee vs the rest of the world) anyways.

I recently changed the labels to be `github` only, but that hasn't been updated yet.

unterstein:
@meichstedt I would add that logs are always highly appreciated :)

LGTM

jeschkies:
@meichstedt, can we merge this?

PR #5519, opened 2017-09-21
Ken/docker test

kensipe:
moved from phabricator


Jenkinsfile.docker
```diff
@@ -0,0 +1,64 @@
...
+    properties([
+        parameters([
+          string(name: 'dcos_url',
+                defaultValue: '',
+                description: 'URL to DC/OS Enterprise Cluster'
+                ),
+
+          string(name: 'marathon',
+                defaultValue: '1.4.6',
+                description: '''Version of Marathon (MoM).  This is the version tag of the docker image.
```
kensipe: comment originally by @zen-dog "I would not exclude v here. Instead of passing the version, I would pass the tag here which might include v but also might look completely different in the future e.g. 1.5.0-SNAPSHOT-470-g660a7bd tags."

tests/system/test_docker_marathon_on_marathon.py
```diff
@@ -0,0 +1,66 @@
...
+pytestmark = [pytest.mark.usefixtures('wait_for_marathon_user_and_cleanup')]
+
+
+@pytest.fixture(scope="function")
+def marathon_service_name():
+    return "marathon-user"
+
+
+def setup_module(module):
+    mom_version = "v1.4.5"
```
jeschkies: Should this be really hardcoded? How about passing it to the test?

tests/system/test_docker_marathon_on_marathon.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        mom_version = os.environ['DOCKER_MOM_VERSION']
```
jeschkies: Ok, I see. I think we should fail if the version if not provided.

tests/system/test_docker_marathon_on_marathon.py
```diff
@@ -0,0 +1,66 @@
...
+import time
+
+from datetime import timedelta
+from dcos import mesos
+from shakedown import marathon
+
+# the following lines essentially do:
+#     from marathon_common_tests import test_*
+import marathon_common_tests
+for attribute in dir(marathon_common_tests):
```
jeschkies: Hm, I'm not sure this is the right approach. Are there other files that should not be imported? Why not import them explicitly by name?

jeschkies:
@kensipe what should we do with this PR?

jeschkies:
Ok, let's close and reopen once there is feedback.

PR #5523, opened 2017-09-22
Changed default of UnreachableEnabled to (0,0)

unterstein:
Summary:
Most users would expect the same behavior in terms of unreachable task as marathon had before introducing the unreachableStrategy.
Therefore this patch re-introduces the old behavior as default configuration.

Test Plan:
sbt test
sbt integration:test

JIRA Issues: MARATHON-7576

Differential Revision (abandoned): https://phabricator.mesosphere.com/D1072

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
@@ -22,7 +22,7 @@ import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.bus.TaskStatusUpdateTestHelper
 import mesosphere.marathon.core.task.state.TaskConditionMapping
 import mesosphere.marathon.core.task.tracker.InstanceTracker
-import mesosphere.marathon.state.{ AppDefinition, PathId, Timestamp }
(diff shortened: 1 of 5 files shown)


src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -206,10 +206,11 @@ class TaskLauncherActorTest extends AkkaUnitTest {
         .setOperator(Operator.UNIQUE)
         .setValue("")
         .build
-      val constraintApp: AppDefinition = f.app.copy(constraints = Set(uniqueConstraint))
+      val unreachableStrategy = UnreachableEnabled(5.minutes, 10.minutes)
```
meln1k: Do we have a test for (0,0) defaults?

unterstein: Kind of. We do test this in an integration test: https://github.com/mesosphere/marathon/blob/0f471239370d702ce2b042a4c8cb7b8ff837f952/src/test/scala/mesosphere/marathon/integration/TaskUnreachableIntegrationTest.scala#L90

To be sure, I`m going to a unit test to cover this!

meln1k: Awesome! Thanks @unterstein!

PR #5524, opened 2017-09-22
Added option to disable plugin

tsydd:
Option to disable plugin is useful, when plugins config is generated by template engine (for example, Ansible) and plugins are optional


unterstein:
Maybe @timcharper or @jdef want to provide some feedback <3

jdef:
Would be great to see a unit test for this change

PR #5527, opened 2017-09-25
Create pull request template.

jeschkies:
Summary:
After moving back to GitHub pull requests we should try to keep the form
of our commit messages.

JIRA issues: MARATHON-7605

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -3,5 +3,3 @@ MARATHON-<<n>> <<Replace this line with your Revision Title>>
 Summary:
 
 Test Plan:
-
-JIRA issues:
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,4 +1,4 @@
-<<Replace this line with your Revision Title>>
+MARATHON-<<n>> <<Replace this line with your Revision Title>>
 
 Summary:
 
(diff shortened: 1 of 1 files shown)


jeschkies:
> Amazing!! ❤️🎉😂

I know, right?

meln1k:
Do you think it makes sense to put Jira issue number in the first line?

jeschkies:
> Do you think it makes sense to put Jira issue number in the first line?

Generally, yes. However, I'm following the pattern we used to have. Then again this won't break the pattern. I'll change it.

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,7 @@
+MARATHON-<<n>> <<Replace this line with your Revision Title>>
+
+Summary:
+
+Test Plan:
+
+JIRA issues:
```
janisz: It looks like JIRA issue number will be doubled at the beginning and at the bottom.

jeschkies: True. I was unsure because we sometimes have multiple issue numbers and this would follow the old pattern a little bit.

kensipe: has anyone ever filled out `Test Plan:` with anything meaningful?

unterstein: @kensipe I use this field usually when I want to highlight a special test case, for example I do:
`sbt 'test-only *PortsMatcherTest'` or I when something should be tested manually. But I also see maybe 60% only `sbt test && sbt integration:test` there.

jeschkies: I thought we agreed in the meeting that we would remove `Test Plan:`.

PR #5528, opened 2017-09-25
leader proxy refactor

kensipe:
this is moved from phabricator... it had a down vote by @jeschkies for the follow very good reasons:

It seems a little strange that the input stream is not used explicitly in the block. How about 
```
scala
val in = request.getInputStream
val out = response.getOutputStream
try {
 ...
} finally {
  IOUtils.closeQuietly(in, out) // IO.using is already doing this
}
```
--
279 | copyRequestToConnection is using `IO.using` which already closes the connection.
306 | How about using Apache commons IO
```
try { body }
finally { IOUtils.closeQuietly(closables:_*) }
```
308 | Why not a try-catch?
310 | Should this rather be a warn?
311 | I think we should use `Done` instead since it's used in other cases as well.

https://phabricator.mesosphere.com/D1023


src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -270,42 +270,49 @@ class JavaUrlConnectionRequestForwarder @Inject() (
...
 
+  /**
+    * Closes the closeables when the provided body returns
+    * If close fails, log and proceed to close next closeable.
+    */
+  def withTryClosing(closeables: => Seq[Closeable])(body: => Unit): Unit = {
+    try { body }
+    finally {
+      closeables.foreach { closeable =>
+        Try(if (closeable != null) closeable.close()) match {
```
jeschkies: You could use Apache commons `IOutils.closeQuietly(closesables)` here.

jeschkies:
@kensipe, what is the state of this patch?

jeschkies:
Superseded by #5829.

PR #5530, opened 2017-09-25
Do sync before reading from or writing to ZooKeeper

ichernetsky:
In addition to that,
- do not load anything from the storage until becoming a leader
- improve the handling of connecting to ZK in RichCuratorFramework and CuratorElectionService

The relevant tickets:
https://jira.mesosphere.com/browse/MARATHON-7763
https://jira.mesosphere.com/browse/DCOS_OSS-1652
https://jira.mesosphere.com/browse/MARATHON-7785

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
+++ b/src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
@@ -252,26 +252,9 @@ class CuratorElectionService(
 
   private def startCuratorClientAndConnect(): Unit = {
     client.start()
-
-    val zkTimeout: Int = client.getZookeeperClient.getConnectionTimeoutMs
-    val timeoutAt: Long = System.currentTimeMillis() + zkTimeout
-    var connected = false
-
-    while (!connected && System.currentTimeMillis <= timeoutAt) {
-      if (!lifecycleState.isRunning) {
-        throw new InterruptedException("Not waiting for connection to zookeeper; Marathon is shutting down")
-      }
-
-      try {
-        connected = client.blockUntilConnected(zkTimeout, java.util.concurrent.TimeUnit.MILLISECONDS)
-      } catch {
-        case _: InterruptedException => // ignore
-      }
-    }
-
-    if (!connected) {
-      throw new InterruptedException("timed out while waiting for zookeeper connection")
-    }
+    client.blockUntilConnected(
+      client.getZookeeperClient.getConnectionTimeoutMs,
+      java.util.concurrent.TimeUnit.MILLISECONDS)
   }
 
   /**
--- a/src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
+++ b/src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
@@ -4,7 +4,7 @@ package core.election.impl
 import java.util
 import java.util.Collections
 import java.util.concurrent.atomic.{ AtomicBoolean, AtomicReference }
-import java.util.concurrent.{ ExecutorService, Executors, TimeUnit }
+import java.util.concurrent.{ ExecutorService, Executors }
 
 import akka.actor.ActorSystem
 import akka.event.EventStream
@@ -111,7 +111,7 @@ class CuratorElectionService(
     if (acquiringLeadership.compareAndSet(false, true)) {
       require(leaderLatch.get.isEmpty, "leaderLatch is not empty")
 
-      startCuratorClient()
+      startCuratorClientAndConnect()
       val latch = new LeaderLatch(
         client, config.zooKeeperLeaderPath + "-curator", hostPort)
       latch.addListener(LeaderChangeListener, threadExecutor)
@@ -250,9 +250,28 @@ class CuratorElectionService(
     client
   }
 
-  private def startCuratorClient(): Unit = {
+  private def startCuratorClientAndConnect(): Unit = {
     client.start()
-    client.blockUntilConnected(config.zkTimeoutDuration.toMillis.toInt, TimeUnit.MILLISECONDS)
+
+    val zkTimeout: Int = client.getZookeeperClient.getConnectionTimeoutMs
+    val timeoutAt: Long = System.currentTimeMillis() + zkTimeout
+    var connected = false
+
+    while (!connected && System.currentTimeMillis <= timeoutAt) {
+      if (!lifecycleState.isRunning) {
+        throw new InterruptedException("Not waiting for connection to zookeeper; Marathon is shutting down")
+      }
+
+      try {
+        connected = client.blockUntilConnected(zkTimeout, java.util.concurrent.TimeUnit.MILLISECONDS)
+      } catch {
+        case _: InterruptedException => // ignore
+      }
+    }
+
+    if (!connected) {
+      throw new InterruptedException("timed out while waiting for zookeeper connection")
+    }
   }
 
   /**
--- a/src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
@@ -9,7 +9,7 @@ import kamon.metric.instrument.Time
 import mesosphere.marathon.core.group.impl.GroupManagerImpl
 import mesosphere.marathon.storage.repository.GroupRepository
 
-import scala.concurrent.{ Await, ExecutionContext }
+import scala.concurrent.ExecutionContext
 
 /**
   * Provides a [[GroupManager]] implementation.
@@ -20,16 +20,7 @@ class GroupManagerModule(
     groupRepo: GroupRepository)(implicit ctx: ExecutionContext, eventStream: EventStream) {
 
   val groupManager: GroupManager = {
-    val groupManager = new GroupManagerImpl(config, Await.result(groupRepo.root(), config.zkTimeoutDuration), groupRepo, scheduler)
-
-    // We've already released metrics using these names, so we can't use the Metrics.* methods
-    Kamon.metrics.gauge("service.mesosphere.marathon.app.count")(
-      groupManager.rootGroup().transitiveApps.size.toLong
-    )
-
-    Kamon.metrics.gauge("service.mesosphere.marathon.group.count")(
-      groupManager.rootGroup().transitiveGroupsById.size.toLong
-    )
+    val groupManager = new GroupManagerImpl(config, None, groupRepo, scheduler)
 
     val startedAt = System.currentTimeMillis()
     Kamon.metrics.gauge("service.mesosphere.marathon.uptime", Time.Milliseconds)(
--- a/src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
@@ -22,19 +22,6 @@ class GroupManagerModule(
   val groupManager: GroupManager = {
     val groupManager = new GroupManagerImpl(config, None, groupRepo, scheduler)
 
-    // We've already released metrics using these names, so we can't use the Metrics.* methods
-    Kamon.metrics.gauge("service.mesosphere.marathon.app.count")(
-      groupManager.rootGroupOption().foldLeft(0L) { (_, group) =>
-        group.transitiveApps.size.toLong
-      }
-    )
-
-    Kamon.metrics.gauge("service.mesosphere.marathon.group.count")(
-      groupManager.rootGroupOption().foldLeft(0L) { (_, group) =>
-        group.transitiveGroupsById.size.toLong
-      }
-    )
-
     val startedAt = System.currentTimeMillis()
     Kamon.metrics.gauge("service.mesosphere.marathon.uptime", Time.Milliseconds)(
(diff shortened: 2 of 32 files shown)


src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -43,9 +44,25 @@ class GroupManagerImpl(
     * Even though updates go through the workqueue, we want to make sure multiple readers always read
     * the latest version of the root. This could be solved by a @volatile too, but this is more explicit.
     */
-  private[this] val root = LockedVar(initialRoot)
-
-  override def rootGroup(): RootGroup = root.get()
+  private[this] var root = LockedVar(initialRoot)
```
meln1k: Mutable reference to the mutable state is a little bit too much :)
What do think about using `val` with `LockedVar` or to remove `LockedVar` and use `@volatile`?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
```diff
@@ -159,22 +158,25 @@ class RichCuratorFramework(val client: CuratorFramework) {
...
-    val timeoutAt: Long = System.currentTimeMillis() + client.getZookeeperClient.getConnectionTimeoutMs
+    val zkTimeout: Int = client.getZookeeperClient.getConnectionTimeoutMs
+    val timeoutAt: Long = System.currentTimeMillis() + zkTimeout
+    var connected = false
 
-    @tailrec def poll(): Unit = {
-      if (System.currentTimeMillis > timeoutAt)
-        throw new InterruptedException("timed out while waiting for zookeeper connection")
-      else if (!lifecycleState.isRunning)
+    while (!connected && System.currentTimeMillis <= timeoutAt) {
```
meln1k: Do you think we should use some kind of exponential backoff here to limit amount of requests we send?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -43,9 +44,25 @@ class GroupManagerImpl(
...
+
+  override def rootGroup(): RootGroup =
+    root.get() match { // linter:ignore:UseGetOrElseNotPatMatch
+      case None =>
```
meln1k: In my view if would be great to have a test for this scenario too.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
ichernetsky: Yeah, it slipped through the cracks, while I was experimenting (with `@volatile` included). It was `val` before. Thanks for catching it.

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
ichernetsky: I don't think it will add any real value here, because it is supposed to happen only once during start-up, and it is supposed to fail after the predefined timeout.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
ichernetsky: I will add one.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -45,7 +45,16 @@ class GroupManagerImpl(
     */
   private[this] val root = LockedVar(initialRoot)
```
meln1k: Do you think it's possible to use an `AtomicReference` here?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
```diff
@@ -159,22 +158,25 @@ class RichCuratorFramework(val client: CuratorFramework) {
     */
   @SuppressWarnings(Array("CatchFatal"))
   def blockUntilConnected(lifecycleState: LifecycleState): Unit = {
-    val timeoutAt: Long = System.currentTimeMillis() + client.getZookeeperClient.getConnectionTimeoutMs
+    val zkTimeout: Int = client.getZookeeperClient.getConnectionTimeoutMs
```
zen-dog: Again - why is the timeout coming from the client and not from `zk.config`?

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -100,6 +100,13 @@ class Migration(
 
   @SuppressWarnings(Array("all")) // async/await
   def migrateAsync(): Future[Seq[StorageVersion]] = async {
+    // Before reading to and writing from the storage, let's ensure that
+    // no stale values are read from the persistence store.
+    // Although in case of ZK it is done at the time of creation of CuratorZK,
+    // it is better to be safe than sorry.
+    logger.info(s"$persistenceStore")
+    await(persistenceStore.sync())
```
zen-dog: So what happens if `sync` fails here (e.g. `StoreCommandFailedException`)? Did you test this case?

src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala
```diff
@@ -250,9 +250,11 @@ class CuratorElectionService(
     client
   }
 
-  private def startCuratorClient(): Unit = {
+  private def startCuratorClientAndConnect(): Unit = {
     client.start()
-    client.blockUntilConnected(config.zkTimeoutDuration.toMillis.toInt, TimeUnit.MILLISECONDS)
+    client.blockUntilConnected(
+      client.getZookeeperClient.getConnectionTimeoutMs,
```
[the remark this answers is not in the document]
ichernetsky: It is already set when the client is created, before it is started. In in line with how it is done in `RichCuratorFramework`. Purely for consistency.

src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
```diff
@@ -20,15 +20,19 @@ class GroupManagerModule(
     groupRepo: GroupRepository)(implicit ctx: ExecutionContext, eventStream: EventStream) {
 
   val groupManager: GroupManager = {
-    val groupManager = new GroupManagerImpl(config, Await.result(groupRepo.root(), config.zkTimeoutDuration), groupRepo, scheduler)
+    val groupManager = new GroupManagerImpl(config, None, groupRepo, scheduler)
 
     // We've already released metrics using these names, so we can't use the Metrics.* methods
     Kamon.metrics.gauge("service.mesosphere.marathon.app.count")(
-      groupManager.rootGroup().transitiveApps.size.toLong
+      groupManager.rootGroupOption().foldLeft(0L) { (_, group) =>
```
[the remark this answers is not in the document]
ichernetsky: Unfortunately, yes, but it has to get broken this time, because reading anything from ZK before becoming a leader is unacceptable.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -45,7 +45,16 @@ class GroupManagerImpl(
...
 
-  override def rootGroup(): RootGroup = root.get()
+  override def rootGroup(): RootGroup =
+    root.get() match { // linter:ignore:UseGetOrElseNotPatMatch
+      case None =>
+        val group = Await.result(groupRepository.root(), config.zkTimeoutDuration)
```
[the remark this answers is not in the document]
ichernetsky: There is no race condition because of some subtleties. It is called first by a leader when performing migration, which is sequential, and there is a synchronization point which is `serializeUpdates: WorkQueue`. But if considering this code not in the overall context, I do agree that a race condition is possible (assuming there is no sync point), though `val group = root := Some(Await.result(groupRepository.root(), config.zkTimeoutDuration))` has the same problem. I will update the code to handle this properly no matter how and when this method is invoked. Thanks a lot for pointing this out.

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
ichernetsky: Because `zk.config` is used at the time of client creation.

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
ichernetsky: Yes. I will add a test.

src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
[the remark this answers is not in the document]
timcharper: Can we just not even register this metric until after leadership is acquired? Better to report nothing than a lie?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -45,7 +45,20 @@ class GroupManagerImpl(
     */
   private[this] val root = LockedVar(initialRoot)
 
-  override def rootGroup(): RootGroup = root.get()
+  @SuppressWarnings(Array("OptionGet"))
+  override def rootGroup(): RootGroup =
```
timcharper: What would you think if we had GroupManagerImpl queried to see if the post sync leadership has been completed, and then threw an exception if not? Ideas include potentially having the sync method complete some promise _before_ notifying anything else that the sync is complete

src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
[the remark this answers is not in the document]
zen-dog: No data would also trigger alerts. But I'm fine with both approaches

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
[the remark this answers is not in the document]
zen-dog: You're right - one line suffers from the same problem. And `root.update()` seems to solve the problem. Thx.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -208,11 +212,33 @@ class GroupManagerImpl(
...
     val currentRoot = await(groupRepository.root())
     root := Option(currentRoot)
+
+    registerMetrics()
     Done
   }
+
+  private[this] val metricsRegistered: AtomicBoolean = new AtomicBoolean(false)
+  private[this] def registerMetrics(): Unit = {
+    if (metricsRegistered.compareAndSet(false, true)) {
```
timcharper: 👏 👏 👏

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -208,11 +212,33 @@ class GroupManagerImpl(
...
   override def invalidateGroupCache(): Future[Done] = async {
+    root := None
+
     // propagation of reset group caches on repository is needed,
     // because manager and repository are holding own caches
     await(groupRepository.invalidateGroupCache())
     val currentRoot = await(groupRepository.root())
     root := Option(currentRoot)
+
+    registerMetrics()
```
timcharper: Why do we have two calls to registerMetrics? Can we have invalidateGroupCache just clear the rootGroup, and then call rootGroup() to repopulate?

ichernetsky: I did it this way In order to not worry about the order (which one gets invoked first). For instance, right now when Marathon becomes a leader it calls `invalidateGroupCache` twice. Also, with the current implementation `root` is set to `Some(...)` here first. But it may change in the future. So, whoever gets called first, registers metrics.

timcharper: This won't work?

```diff
diff --git a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
index 82a74f95d..a7dbd5864 100644
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
@@ -217,10 +217,7 @@ class GroupManagerImpl(
     // propagation of reset group caches on repository is needed,
     // because manager and repository are holding own caches
     await(groupRepository.invalidateGroupCache())
-    val currentRoot = await(groupRepository.root())
-    root := Option(currentRoot)
-
-    registerMetrics()
+    rootGroup()
     Done
   }
 
```

ichernetsky: I think not, because `groupRepository.root()` has to be invoked.

PR #5531, opened 2017-09-26
MARATHON_EE-1673: Introduce lightweight SSE flag

wavesoft:
⚠️  Depends on #5625

---

This PR adds support for the URL argument `/v2/events?plan-format=light` that subscribes to a "lighter" version of the SSE events, where some of the deployment plan fields are excluded.

~The flag event_stream_lightweight enables "lightweight" events. Such events have the origin and target dicts in the deployment plan removed and thus a major part of the SEE size reduced.~

For more information regarding this issue you can refer to this document: https://docs.google.com/document/d/1BsfSoglJ2Z2RM1CQIqrQl_zDkiSLbd68fQoEmCeOgqA/edit#heading=h.7t0ffuvu6r3h


src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -27,6 +27,8 @@ import scala.concurrent.duration._
 // https://mesosphere.atlassian.net/browse/MARATHON-1292
 object Formats extends Formats {
 
+  type LightweightDeploymentPlanFlag = Boolean
```
wavesoft: _According to https://phabricator.mesosphere.com/D1056 discussion:_
* This type should be removed and the `Boolean` now flag should be inlined

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -359,7 +369,7 @@ trait EventFormats {
     )
   }
 
-  def eventToJson(event: MarathonEvent): JsValue = event match {
+  def eventToJson(event: MarathonEvent)(implicit lightweightEvents: LightweightDeploymentPlanFlag): JsValue = event match {
```
wavesoft: In line with my previous comment: The implicit should be conversed to an explicit argument

src/main/scala/mesosphere/marathon/core/event/EventConf.scala
```diff
@@ -14,5 +14,13 @@ trait EventConf extends ScallopConf {
     default = Some(1024)
   )
 
+  lazy val eventStreamLightweight: ScallopOption[Boolean] = opt[Boolean](
+    "event_stream_lightweight",
```
wavesoft: Bringing in @meichstedt comment:
> Let's talk about naming :)
>
> For (at least some[1]) existing command line args, we use the [disable_]fancy_mode notation. That way, you  either specify --fancy_mode to enable it, or --disable_fancy_mode to, well, disable it.
> 
> [1] For references, see metrics, store_cache, internal_version_caching, zk_compression, ha, checkpoint, mesos_authentication, revive_offers_for_new_apps.
19

wavesoft:
To bring in the discussion from Phabricator:

* I have to clarify that I consider this to be a **bug**. If you have about 1000 deployed apps the event bus becomes practically unusable. Therefore I am opting for a deprecation plan and thus a command-line flag, or api versioning + deprecation plan.
* This flag can help towards a later deprecation of the current state of the SSE message format.

jdef:
Can we please avoid the use of custom JSON formats, and instead use the RAML types? Custom JSON formats are a mess and have led to code that's nearly impossible to maintain w/o introducing bugs.

wavesoft:
After a discussion with @jeschkies I would agree with @jdef ... think that the best would be to move to RAML types.

jeschkies:
@wavesoft, is this the diff about the get parameter for lightweight events?

wavesoft:
@jeschkies yes, this is, I forgot to push :smile:

meln1k:
So, what was decided regarding usage of RAML here?

wavesoft:
This implementation is the "fastest possible solution on the current state". It's not currently taking RAML into account.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -52,15 +52,19 @@ class EventsController(
...
-                .fromGraph[MarathonEvent, NotUsed](
-                  new EventStreamSourceGraph(eventBus, conf.eventStreamMaxOutstandingMessages(), clientIp)
-                )
-                .filter(isAllowed(events.toSet))
-                .map(event => ServerSentEvent(`type` = event.eventType, data = Json.stringify(Formats.eventToJson(event))))
-                .keepAlive(5.second, () => ServerSentEvent.heartbeat)
+          parameter("plan-format".?) { light =>
+            extractClientIP { clientIp =>
+              complete {
+                Source
```
meln1k: Do you think it make sense to move this to a different method?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -52,15 +52,19 @@ class EventsController(
...
-          extractClientIP { clientIp =>
-            complete {
-              Source
-                .fromGraph[MarathonEvent, NotUsed](
-                  new EventStreamSourceGraph(eventBus, conf.eventStreamMaxOutstandingMessages(), clientIp)
-                )
-                .filter(isAllowed(events.toSet))
-                .map(event => ServerSentEvent(`type` = event.eventType, data = Json.stringify(Formats.eventToJson(event))))
-                .keepAlive(5.second, () => ServerSentEvent.heartbeat)
+          parameter("plan-format".?) { light =>
```
meln1k: Will something like this work here? 
```scala
(parameters("plan-format".?, 'event_type.*) & extractClientIP) { (light, events, clientIp) =>
```

wavesoft: True, I forgot that `parameters` can accept more than one argument :smile:

wavesoft: But I think the `extractClientIP` type is incompatible and cannot be conjugated.

meln1k: in AppsContoller we have something like `(entity(as[AppDefinition]) & forceParameter & extractClientIP & extractUri) { (rawApp, force, remoteAddr, reqUri) =>`, different types shouldn't be a problem :)

wavesoft: Ah, true :smile: (forgot the parenthesis)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
wavesoft: I am indifferent on this. Yes, the logic is a bit mixed, but on the other hand with the trick you suggested above the code becomes quite small and easy to understand :man_shrugging:

meln1k:
Looks great!

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -49,19 +49,20 @@ class EventsController(
...
-            }
+        (parameters('event_type.*, "plan-format".?) & extractClientIP) { (events, light, clientIp) =>
+          complete {
+            Source
+              .fromGraph[MarathonEvent, NotUsed](
+                new EventStreamSourceGraph(eventBus, conf.eventStreamMaxOutstandingMessages(), clientIp)
+              )
+              .filter(isAllowed(events.toSet))
+              .map(event => ServerSentEvent(`type` = event.eventType, data = Json.stringify(
+                Formats.eventToJson(event, light.exists(_.contentEquals("light")))
```
meichstedt: I've only skimmed this PR but wouldn't it be lighter to explicitly pass a `Writes` in the controller instead of passing down the config into the formats definition? Something like (watch out, imperative pseudo code 😱 )
```
val writes = if (light) lightWrites else heavyWrites
toJson(event)(writes)
```
.. where `lightWrites` contains writes for the given events using light raml generated formats, and `heavyWrites` uses the heavy formats. Passing down the flag seems kind of 🐟 'y.

So instead of importing `mesosphere.marathon.api.v2.json.Formats`, you'd explicitly reference the format you want.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -239,6 +238,14 @@ trait EventFormats {
     )
   }
 
+  lazy val LightDeploymentPlanWrites: Writes[DeploymentPlan] = Writes { plan =>
+    Json.obj(
```
meichstedt: As @jdef noted in a comment: can you please use raml and have the formats for this and the other messages generated?

wavesoft:
@meichstedt @jdef @jeschkies I am also onboard with RAML type definitions, so let's do this. 

Note that this is going to change a lot of things since we currently don't have any RAML types for most of the deployment-related responses. For this purpose I started working on this on a new PR, so I redirect your attention there and I am putting this PR on halt: https://github.com/mesosphere/marathon/pull/5625

wavesoft:
Just a comment on this PR:

I replaced all the hand-written types with RAML types but since that addition is quite big I decided to split it in a separate PR. However, since this has been around for too long and is currently functional I am opting for merging it now and replacing the types in a separate PR. 

What do you think @meichstedt , @jeschkies ?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
jeschkies: +1 Please introduce RAML types.

PR #5538, opened 2017-09-27
Set a constant time for task status updates in InstanceTest.

jeschkies:
Summary:
The UnreachableStrategy defaults were change to zero. since we used the
current time as status update times the was a small race condition
since we could hit the inactive threshold.

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/instance/InstanceTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/instance/InstanceTest.scala
@@ -107,15 +107,15 @@ class InstanceTest extends UnitTest with TableDrivenPropertyChecks {
     def tasks(statuses: Seq[Condition]): Map[Task.Id, Task] =
       statuses.map { status =>
         val taskId = Task.Id.forRunSpec(id)
-        val mesosStatus = MesosTaskStatusTestHelper.mesosStatus(status, taskId, Timestamp.now())
-        val task = TestTaskBuilder.Helper.minimalTask(taskId, Timestamp.now(), mesosStatus, status)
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/marathon/core/instance/InstanceTest.scala
```diff
@@ -107,15 +107,15 @@ class InstanceTest extends UnitTest with TableDrivenPropertyChecks {
...
         task.taskId -> task
       }(collection.breakOut)
 
     def instanceWith(condition: Condition, conditions: Seq[Condition]): (Instance, Map[Task.Id, Task]) = {
       val currentTasks = tasks(conditions.map(_ => condition))
       val newTasks = tasks(conditions)
-      val state = Instance.InstanceState(None, currentTasks, Timestamp.now(), UnreachableStrategy.default())
+      val state = Instance.InstanceState(None, currentTasks, clock.now(), UnreachableStrategy.default())
       val instance = Instance(Instance.Id.forRunSpec(id), agentInfo, state, currentTasks,
         runSpecVersion = Timestamp.now(), UnreachableStrategy.default())
```
zen-dog: Doesn't play a role but I would use `runSpecVersion = clock.now()` too. 🚢

jeschkies:
Please note. We do not expunge immediately if the update reaches us at the same time. See https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/task/Task.scala#L95. If `since` and `now` are the same time `since.expired(now, by = timeout)` is false even if `timeout` is `0`.

PR #5545, opened 2017-09-28
Analyze logs with ELK stack

jeschkies:
Summary:
This is the result of our Dreamathon. So far it assumes that the logs have been stripped from any ANSI color marks. Also, the logs for production are different than from integration tests (go figure). The app id extraction is buggy.


logstash/config/logstash.conf
```diff
@@ -0,0 +1,120 @@
...
+    }
+    overwrite => [ "message" ]
+#    remove_field => [ "@version", "port" ]
+    add_tag => ["unclassified"]
+  }
+
+  # Match content of messages
+
+  # Deployment related events.
+  grok {
```
jeschkies: We should extract these into separate files.

logstash/config/logstash.conf
```diff
@@ -0,0 +1,120 @@
...
+    remove_field => [ "logdate" ]
+  }
+}
+
+input {
+  file {
+    path => '/usr/share/logstash/bundle'
+    start_position => 'beginning'
+    type => 'log'
+    codec => multiline {
```
jeschkies: It would be nice to strip ANSI here.

timcharper:
Seems like a good start. Some of the features I think would be useful:

[ ] Script to target loading from some DCOS bundle, and generate logstash config to load the files, setting the `host` field, appropriately (host can not be derived from the log file in this case and must be derived from the path of the log file, IE `master_{ip}/dcos-marathon.service.gz`)
[ ] "configure" mechanism to load configuration in "debug" mode; that is to say that logstash would be launched listening for log input on a port, with config auto-reload enabled, so that a series of lines could be fed via netcat.
[ ] "configure" mechanism to detect log format. The prefix changes between logging systems. It'd be nice if we could have our logstash config generator detect this, and then generate the appropriate multiline regex and date parser section.

timcharper:
It could be _really_ nice to be able to load logs for integration tests, or for a diagnostic bundle.

jeschkies:
Superseded by https://github.com/mesosphere/marathon/commit/537f45a2b3bcc1947255288e64aad16f704c5f30#diff-9c3b4d40e281410f3401501922c49343

PR #5403, opened 2017-05-22
EventsControllerTest - tests for EventStreamSourceGraph stage [MARATHON-7317]

gosubpl:
This is work-in-progress. And my first PR to the `marathon` repo.
I am trying to understand what sort of tests could be useful for you. I have drafted two tests for the `EventStreamSorceGraph` stage. There should probably be some more, for the edge cases resulting from commands like `Standby` and `ElectedAsLeader`.
Please let me know if this is useful, and if yes - what additional tests you would you like me to add.
https://jira.mesosphere.com/browse/MARATHON-7317

PR #5548, opened 2017-09-29
Simply Akka EventsController Stream Logic

timcharper:
* Replace the custom stage implementation with stream composed via the DSL.
* Add unit tests

JIRA Issues: MARATHON-7317

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
@@ -6,7 +6,7 @@ import akka.event.EventStream
 import akka.http.scaladsl.model.RemoteAddress
 import akka.http.scaladsl.server.Route
 import akka.stream.OverflowStrategy
-import akka.stream.scaladsl.Source
+import akka.stream.scaladsl.{ Keep, Source }
 import com.typesafe.scalalogging.StrictLogging
 import de.heikoseeberger.akkasse.ServerSentEvent
 import mesosphere.marathon.api.akkahttp.Controller
@@ -90,15 +90,17 @@ object EventsController extends StrictLogging {
 
     EnrichedSource.eventBusSource(classOf[MarathonEvent], eventStream, bufferSize, OverflowStrategy.fail)
       .via(EnrichedFlow.stopOnFirst(leaderLossKillSwitch))
-      .watchTermination() { (m, completed) =>
-        eventStream.publish(EventStreamAttached(remoteAddress = remoteAddress.toString()))
-        logger.info(s"EventStream attached: $remoteAddress")
+      .watchTermination()(Keep.both)
+      .mapMaterializedValue {
+        case (cancellable, completed) =>
+          eventStream.publish(EventStreamAttached(remoteAddress = remoteAddress.toString()))
+          logger.info(s"EventStream attached: $remoteAddress")
 
-        completed.onComplete { _ =>
-          eventStream.publish(EventStreamDetached(remoteAddress = remoteAddress.toString()))
-          logger.info(s"EventStream detached: $remoteAddress")
-        }(ExecutionContexts.callerThread)
-        m
+          completed.onComplete { _ =>
+            eventStream.publish(EventStreamDetached(remoteAddress = remoteAddress.toString()))
+            logger.info(s"EventStream detached: $remoteAddress")
+          }(ExecutionContexts.callerThread)
+          cancellable
       }
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
@@ -1,27 +1,23 @@
 package mesosphere.marathon
 package api.akkahttp.v2
 
-import akka.NotUsed
-import akka.actor.Actor.Receive
-import akka.actor.{ ActorSystem, Cancellable }
+import akka.actor.ActorSystem
 import akka.event.EventStream
 import akka.http.scaladsl.model.RemoteAddress
 import akka.http.scaladsl.server.Route
-import akka.stream._
-import akka.stream.scaladsl.{ Sink, Source }
-import akka.stream.stage.GraphStageLogic.StageActor
-import akka.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
+import akka.stream.OverflowStrategy
+import akka.stream.scaladsl.Source
 import com.typesafe.scalalogging.StrictLogging
 import de.heikoseeberger.akkasse.ServerSentEvent
 import mesosphere.marathon.api.akkahttp.Controller
-import mesosphere.marathon.api.akkahttp.v2.EventsController.EventStreamSourceGraph
 import mesosphere.marathon.api.v2.json.Formats
+import mesosphere.marathon.core.async.ExecutionContexts
 import mesosphere.marathon.core.election.{ ElectionService, LeadershipState }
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -1,42 +1,37 @@
...
-    val eventBus: EventStream)(
+    eventStreamMaxOutstandingMessages: Int,
+    eventBus: EventStream)(
     implicit
-    val actorSystem: ActorSystem,
     val executionContext: ExecutionContext,
     val authenticator: Authenticator,
     val authorizer: Authorizer,
-    val electionService: ElectionService
+    val electionService: ElectionService // TODO - it's inconsistent to make this an implicit
```
meln1k: Do we need it to be an implicit parameter?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -70,76 +61,44 @@ class EventsController(
...
-        val actor = getStageActor(receive)
-        val doCompleteStage = getAsyncCallback[Unit] { _ => completeStage() }
-        electionSubscription = leaderStateEvents.to(Sink.foreach({
-          case _: LeadershipState.Standby => doCompleteStage.invoke(())
-          case _ => ()
-        })).run()(this.materializer)
+    EnrichedSource.eventBusSource(classOf[MarathonEvent], eventStream, bufferSize, OverflowStrategy.fail)
+      .via(EnrichedFlow.stopOnFirst(leaderLossKillSwitch))
+      .watchTermination() { (m, completed) =>
         eventStream.publish(EventStreamAttached(remoteAddress = remoteAddress.toString()))
```
meln1k: Do you think it worth moving this to `.mapMaterializedValue` part?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
timcharper: No :) It is just already that way for this and other controllers. I suggest we change in a separate commit.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
timcharper: After thinking about this, I think you're right. It is a bit confusing to hijack the materialization reducer function. Better to do `.watchTermination()(Keep.both).mapMaterializedValue { m => ... }`

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
```diff
@@ -0,0 +1,126 @@
...
+
+    val output = response.entity.dataBytes
+      .via(EventStreamParser(Int.MaxValue, Int.MaxValue))
+      .drop(1)
+      .take(1)
+      .runWith(Sink.head)
+
+    f.eventBus.publish(testTerminatedEvent)
+
+    val sse = output.futureValue
```
jeschkies: What type is `sse`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -70,76 +61,46 @@ class EventsController(
...
+    val leaderLossKillSwitch =
+      leaderEvents.collect { case evt: LeadershipState.Standby => evt }
 
-      setHandler(out, new OutHandler {
-        override def onPull(): Unit = if (messages.nonEmpty) push(out, messages.dequeue())
-      })
-
-      override def preStart(): Unit = {
-        super.preStart()
-        val actor = getStageActor(receive)
```
jeschkies: Why did we have this actor? What was it doing?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    sse.`type` shouldBe Some("app_terminated_event")
+    sse.data.get should include("/my-app")
+  }
+
+  "Respects filtering conditions" in withFixture() { f =>
+    val response = f.eventsRequest(Get(Uri("/?event_type=app_terminated_event")))
+
+    val output = response.entity.dataBytes
```
jeschkies: What type is `output` here?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
```diff
@@ -0,0 +1,126 @@
...
+      .futureValue
+
+    attachedEvent.`type` shouldBe Some("event_stream_attached")
+    attachedEvent.data.get should include("1.2.3.4")
+  }
+
+  "Relays events that are published after it is connected" in withFixture() { f =>
+    val response = f.eventsRequest(Get())
+
+    val output = response.entity.dataBytes
```
jeschkies: What type is `output` here.

timcharper:
Closing patch for now; need to sort out the upstream Curator leader election simplification patch, first

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
timcharper: Server sent event. IE the second event sent over HTTP, and parsed via the `EventStreamParser`. (we drop the first event, which is the event saying "we're connected").

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
timcharper: output is the materialized type of the stream. Down below, on line 103, you see `.runWith(Sink.queue())`.

Looking at the docs for [.runWith](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Source.html#runWith[Mat2](sink:akka.stream.Graph[akka.stream.SinkShape[Out],Mat2])(implicitmaterializer:akka.stream.Materializer):Mat2), we see that it returns the materialized type for the provided Sink:

```
runWith[Mat2](sink: Graph[SinkShape[Out], Mat2])(implicit materializer: Materializer): Mat2
```

The sink we pass it is of type [Sink.queue](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/Sink$.html#queue[T]():akka.stream.scaladsl.Sink[T,akka.stream.scaladsl.SinkQueueWithCancel[T]]), which has this type signature:

```
Sink[T, SinkQueueWithCancel[T]]
```

Therefore, the type of `output` is [SinkQueueWithCancel](https://doc.akka.io/api/akka/current/akka/stream/scaladsl/SinkQueueWithCancel.html)

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
timcharper: Future[SSE] (server sent event)

There is a reason for this order.

When we get the Future[SSE], we know that we have established the HTTP connection, received a successful response, but haven't necessarily received any events yet. This is important because before the akka http controller sends the http OK response, it subscribes to the event stream, providing us the guarantee that the event published by `f.eventBus.publish` on line 91 will be received and published through the http SSE stream.

gosubpl:
@timcharper if you ever need the test from https://github.com/mesosphere/marathon/pull/5403 ported to new setup, it is here: https://github.com/gosubpl/marathon/commit/67b2a16931dab81f79a69938358b3cc392e8f62c

PR #5550, opened 2017-09-29
Entity marshallers for apps controller

meln1k:
Summary: entity marshallers for apps controller

Test Plan: sbt test

JIRA issues: MARATHON-7299


src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -95,10 +96,49 @@ object EntityMarshallers {
...
+  implicit def appDefinitionUnmarshaller(
     implicit
-    normalization: Normalization[raml.App], validator: Validator[AppDefinition]) = {
-    validEntityRaml(playJsonUnmarshaller[raml.App])
+    normalization: Normalization[raml.App],
+    validator: Validator[AppDefinition]): FromEntityUnmarshaller[AppDefinition] = {
+    validEntityRaml(playJsonUnmarshaller[raml.App]).handleValidationErrors
+  }
+
+  implicit val appDefinitionMarshaller: ToEntityMarshaller[AppDefinition] =
```
jeschkies: If I'm not mistaken this could just be

```scala
implicit val appDefinitionMarshaller = playJsonMarshaller[appDefinitionMarshaller]
```
or what did I miss?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -95,10 +96,46 @@ object EntityMarshallers {
...
+  implicit def appDefinitionUnmarshaller(
     implicit
-    normalization: Normalization[raml.App], validator: Validator[AppDefinition]) = {
-    validEntityRaml(playJsonUnmarshaller[raml.App])
+    normalization: Normalization[raml.App],
+    validator: Validator[AppDefinition]): FromEntityUnmarshaller[AppDefinition] = {
+    validEntityRaml(playJsonUnmarshaller[raml.App]).handleValidationErrors
+  }
+
+  implicit val appDefinitionMarshaller: ToEntityMarshaller[AppDefinition] = playJsonMarshaller[AppDefinition]
```
timcharper: Does this implicitly convert to raml, and then to json? Might be clearer to use a different method name?

Maybe something like:

```
ramlJsonMarshaller[raml.App, AppDefinition]
```

?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -136,4 +184,8 @@ object EntityMarshallers {
     case ValidationFailed(failure) =>
       complete(StatusCodes.UnprocessableEntity -> failure)
   }
+
+  import scala.language.implicitConversions
+  implicit def entityMarshallerToMessageUnmarshaller[T](um: FromEntityUnmarshaller[T]): FromMessageUnmarshaller[T] =
```
timcharper: seems like the name here should be `entityUnmarshallerToMessageUnmarshaller`

timcharper: also, how bad would it be if this conversion were explicit? My default is to avoid implicit conversions.

jeschkies: +1 on that. Let's try to avoid the implicit conversion unless we can find a great benefit.

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
jeschkies: I've used the same for the info endpoint. However, it should be `implicit val appDefinitionMarshaller: ToEntityMarshaller[raml.App] = playJsonMarshaller[raml.App]` if I'm not mistaken.

meln1k: `internalToRamlJsonMarshaller[raml.App, AppDefinition]` doesn't compile for some reason. I'd just wait to use it in AppsController and then refactor it properly

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -119,22 +156,33 @@ object EntityMarshallers {
   implicit val loggerChangeUnmarshaller = playJsonUnmarshaller[LoggerChange]
   implicit val stringMapMarshaller = playJsonMarshaller[Map[String, String]]
   implicit val pluginDefinitionsMarshaller = playJsonMarshaller[PluginDefinitions]
+  implicit val deploymentResultMarshaller = playJsonMarshaller[Messages.DeploymentResult]
+  implicit val enrichedTaskMarshaller = playJsonMarshaller[EnrichedTask]
```
jeschkies: Ideally this is `raml.EnrichedTask`. I'm currently working on this in https://github.com/mesosphere/marathon/pull/5556. So I think it is fine here.

src/main/scala/mesosphere/marathon/api/akkahttp/Messages.scala
```diff
@@ -0,0 +1,26 @@
...
+package api.akkahttp
+
+import mesosphere.marathon.core.deployment.DeploymentPlan
+import mesosphere.marathon.state.Timestamp
+import play.api.libs.json.Json
+
+/**
+  * Classes for generating json responses
+  */
+object Messages {
```
jeschkies: These should be modelled in the RAML.

docs/docs/rest-api/public/api/v2/types/deploymentResult.raml
```diff
@@ -0,0 +1,11 @@
...
+uses:
+  strings: stringTypes.raml
+
+types:
+  DeploymentResult:
+    type: object
+    description: Result of the deployment
+    properties:
+      deploymentId: string
+      version: datetime
```
jeschkies: Nit: I think there has to be a newline.

PR #5553, opened 2017-10-01
Upgrade wix-accord to 0.7.1

timcharper:
This is a pre-requisite to moving to Scala 2.12

A common source of frustration when testing validations has been to
figure out the right matcher format. To help with that aim, we
introduce the shouldViolate and shouldSucceed test helpers which have
much easier-to-grok output.

Started with @aquamatthias's branch https://github.com/mesosphere/marathon/tree/mv/wix_06

JIRA Issues: MARATHON-1779

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -63,7 +63,6 @@ object Dependencies {
     Test.akkaHttpTestKit % "test",
     Test.junit % "test",
     Test.scalacheck % "test",
-    Test.wixAccordScalatest % "test",
     Test.curatorTest % "test"
   ) ++ Kamon.all).map(
     _.excludeAll(excludeSlf4jLog4j12)
@@ -191,7 +190,6 @@ object Dependency {
     val diffson = "org.gnieh" %% "diffson" % V.Diffson
     val junit = "junit" % "junit" % V.JUnit
     val scalacheck = "org.scalacheck" %% "scalacheck" % V.ScalaCheck
-    val wixAccordScalatest = "com.wix" %% "accord-scalatest" % V.WixAccord
     val curatorTest = "org.apache.curator" % "curator-test" % V.Curator
   }
 }
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -108,7 +108,7 @@ object Dependency {
     val MarathonApiConsole = "3.0.8-accept"
     val Logback = "1.1.3"
     val Logstash = "4.9"
-    val WixAccord = "0.5"
+    val WixAccord = "0.7.1"
     val Curator = "2.11.1"
     val Java8Compat = "0.8.0"
     val ScalaLogging = "3.5.0"
--- a/src/main/scala/mesosphere/marathon/api/RestResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/RestResource.scala
@@ -64,16 +64,15 @@ trait RestResource {
     * See [[assumeValid]], which is preferred to this.
     *
     * @param t object to validate
-    * @param description optional description which might be injected into the failure message
     * @param fn function to execute after successful validation
     * @param validator validator to use
     * @tparam T type of object
     * @return returns a 422 response if there is a failure due to validation. Executes fn function if successful.
     */
-  protected def withValid[T](t: T, description: Option[String] = None)(fn: T => Response)(implicit validator: Validator[T]): Response = {
+  protected def withValid[T](t: T)(fn: T => Response)(implicit validator: Validator[T]): Response = {
     validator(t) match {
       case f: Failure =>
-        val entity = Json.toJson(description.map(f.withDescription).getOrElse(f)).toString
+        val entity = Json.toJson(f).toString
         Response.status(StatusCodes.UnprocessableEntity.intValue).entity(entity).build()
       case Success => fn(t)
     }
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -80,7 +80,8 @@ trait Validation {
               case Success => accum
               case Failure(violations) =>
                 val scopedViolations = violations.map { violation =>
-                  violation.withPath(Generic(key) +: violation.path) }
+                  violation.withPath(Generic(key) +: violation.path)
+                }
                 accum.and(Failure(scopedViolations))
             }
         }
@@ -116,7 +117,8 @@ trait Validation {
               case Success => accum
               case Failure(violations) =>
                 val scopedViolations = violations.map { violation =>
-                  violation.withPath(Indexed(index.toLong) +: violation.path) }
+                  violation.withPath(Indexed(index.toLong) +: violation.path)
+                }
                 accum.and(Failure(scopedViolations))
             }
         }
--- a/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/Validation.scala
@@ -3,9 +3,11 @@ package api.v2
 
 import java.net._
 
+import com.wix.accord.Descriptions._
 import com.wix.accord._
 import com.wix.accord.dsl._
 import com.wix.accord.ViolationBuilder._
+import mesosphere.marathon.api.v2.Validation.ConstraintViolation
 import mesosphere.marathon.state.FetchUri
 import mesosphere.marathon.stream.Implicits._
 import org.slf4j.LoggerFactory
@@ -13,7 +15,6 @@ import play.api.libs.json._
 
 import scala.collection.GenTraversableOnce
 import scala.util.matching.Regex
-
 import scala.language.implicitConversions
 
 // TODO(jdef) move this into package "validation"
@@ -35,7 +36,7 @@ trait Validation {
   def definedAnd[T](implicit validator: Validator[T]): Validator[Option[T]] = {
     new Validator[Option[T]] {
       override def apply(option: Option[T]): Result = option.map(validator).getOrElse(
-        Failure(Set(RuleViolation(None, "not defined", None)))
+        Failure(Set(RuleViolation(None, "not defined")))
       )
     }
   }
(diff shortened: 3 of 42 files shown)


project/Dependencies.scala
```diff
@@ -108,7 +108,7 @@ object Dependency {
     val MarathonApiConsole = "3.0.8-accept"
     val Logback = "1.1.3"
     val Logstash = "4.9"
-    val WixAccord = "0.5"
+    val WixAccord = "0.7.1"
```
jeschkies: 🎉

src/main/scala/mesosphere/marathon/api/RestResource.scala
```diff
@@ -64,16 +64,15 @@ trait RestResource {
...
     *
     * @param t object to validate
-    * @param description optional description which might be injected into the failure message
     * @param fn function to execute after successful validation
     * @param validator validator to use
     * @tparam T type of object
     * @return returns a 422 response if there is a failure due to validation. Executes fn function if successful.
     */
-  protected def withValid[T](t: T, description: Option[String] = None)(fn: T => Response)(implicit validator: Validator[T]): Response = {
+  protected def withValid[T](t: T)(fn: T => Response)(implicit validator: Validator[T]): Response = {
```
jeschkies: Where is the description coming from now?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -52,35 +53,45 @@ trait Validation {
...
-          case m: Map[_, _] =>
-            m.map(item => (item, validator(item))).collect {
-              case ((k, v), f: Failure) => GroupViolation(k, "not valid", Some(s"($k)"), f.violations)
-            }(collection.breakOut)
-          case _ =>
-            seq.map(item => (item, validator(item))).zipWithIndex.collect {
-              case ((item, f: Failure), pos: Int) => GroupViolation(item, "not valid", Some(s"($pos)"), f.violations)
-            }(collection.breakOut)
+  implicit def everyKeyValue[T](implicit validator: Validator[T]): Validator[Iterable[(String, T)]] = {
+    def mapViolation(violation: Violation, desc: Description): Violation = {
```
jeschkies: Is this where the description comes from now?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -52,35 +53,45 @@ trait Validation {
...
-        val violations: Set[Violation] = seq match {
-          case m: Map[_, _] =>
-            m.map(item => (item, validator(item))).collect {
-              case ((k, v), f: Failure) => GroupViolation(k, "not valid", Some(s"($k)"), f.violations)
-            }(collection.breakOut)
-          case _ =>
-            seq.map(item => (item, validator(item))).zipWithIndex.collect {
-              case ((item, f: Failure), pos: Int) => GroupViolation(item, "not valid", Some(s"($pos)"), f.violations)
-            }(collection.breakOut)
+  implicit def everyKeyValue[T](implicit validator: Validator[T]): Validator[Iterable[(String, T)]] = {
```
jeschkies: A doc string would help understand the use case. Also, it's too many implicits for my taste :/. Why not convert it explicitly?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -52,35 +53,45 @@ trait Validation {
...
+    def mapViolation(violation: Violation, desc: Description): Violation = {
+      violation match {
+        case RuleViolation(value, constraint, path) =>
+          RuleViolation(value, constraint, desc +: path)
+        case GroupViolation(value, constraint, children, path) =>
+          GroupViolation(value, constraint, children, desc +: path)
+      }
+    }
+    new Validator[Iterable[(String, T)]] {
```
jeschkies: I thought wix accord has this already, no?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -299,10 +263,42 @@ trait Validation {
     )
 
   def validateAll[T](x: T, all: Validator[T]*): Result = all.map(v => validate(x)(v)).fold(Success)(_ and _)
+
+  def allViolations(result: Result): Seq[ConstraintViolation] = {
```
jeschkies: Could you add a doc string?

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
```diff
@@ -299,10 +263,42 @@ trait Validation {
...
+    }
+  }
 }
 
 object Validation extends Validation {
 
+  case class ConstraintViolation(path: String, constraint: String)
+
+  implicit class AsDescription(val optional: Option[String]) extends AnyVal {
+    def asDescription: Description = ??? //optional.fold[Description](Empty)(Explicit)
```
jeschkies: Is `???` the right choice here?

src/test/scala/mesosphere/UnitTest.scala
```diff
@@ -79,26 +79,51 @@ trait ValidationTestLike extends Validation {
...
+  def shouldViolate[T](entity: T, expectedViolations: (String, String)*)(implicit validator: Validator[T]): Unit = {
+    validator(entity) match {
+      case Success => fail("Validation succeeded, but expected failure")
+      case f: Failure =>
+        val violations = Validation.allViolations(f)
+        expectedViolations.foreach {
+          case (path, constraint) =>
+            val haveMatch = violations.exists { v => v.path == path && v.constraint == constraint }
+
+            if (!haveMatch)
```
jeschkies: Thanks a ton! We needed this.

src/test/scala/mesosphere/UnitTest.scala
```diff
@@ -79,26 +79,51 @@ trait ValidationTestLike extends Validation {
...
+    s"""- "${c.path}" -> "${c.constraint}""""
 
-  def containViolation(path: String, message: String): Matcher[Result] = {
-    Matcher {
-      case Success =>
-        MatchResult(
-          false,
-          s"result had no violations; expected ${path} -> ${message}",
-          s"result was success")
+  def shouldViolate[T](entity: T, expectedViolations: (String, String)*)(implicit validator: Validator[T]): Unit = {
```
jeschkies: Why make the validator implicit?

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala
```diff
@@ -411,15 +408,7 @@ class AppDefinitionFormatsTest extends UnitTest
...
-      ) should failWith(
-        GroupViolationMatcher(
-          description = "container",
-          constraint = "is invalid",
-          violations = Set(
-            group("docker", "is invalid", "credential" -> "must be empty")
-          )
-        )
-      )
+      shouldViolate(app, "/container/docker/credential" -> "must be empty")(AppValidation.validateOldAppAPI)
```
jeschkies: Nit: As a follow up we could make it a ScalaTest assertion:
```scala
AppValidation.validateOldAppAPI shouldViolate("/container/docker/credential" -> "must be empty") on(app)
```

or something similar. I would not do so in this diff though.

src/test/scala/mesosphere/UnitTest.scala
timcharper: Great question; I actually sure about this change, but went with it since it it was what MV had proposed in his branch. I really disliked the wix accord matchers. I find the error messages to be cryptic and the matchers to be particularly verbose to write. Things are better with wix 0.7, but I still prefer the simple "haveViolations" matcher.

I'll rewrite this to exclude the `shouldFail` and `shouldSucceed` with our opinionated matchers (path and violation message only matter). should `haveViolations` and `haveNoViolations`

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: Lol, crazy! The tests all pass with this so looks like we don't use this method. I overlooked this when I was doing the effort of getting the original wix 0.6 changes to build.

I'll just remove it.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: Yeah, copied from original path; great point. Implicits don't seem to be necessary or less helpful, here

src/main/scala/mesosphere/marathon/api/RestResource.scala
timcharper: It isn't needed

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala
timcharper: I agree; I'd like to do it in this diff.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: ✅

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: I simplified this code; description is... a path component. Anyways next patch is clearer.

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: It has the "each" operator, but it is insufficient for many cases.

It is fine for his case:

```
    // before
    portMappings is every(portMappingNetworkNameValidator(networks.flatMap(_.name)))
    // after
    portMappings.each is portMappingNetworkNameValidator(networks.flatMap(_.name))
```

It does not seem to work well for these cases:

```
      isTrue("Volume names must be unique") { (vols: Seq[AppVolume]) =>
        val names: Seq[String] = vols.collect{ case v: AppExternalVolume => v.external.name }.flatten
        names.distinct.size == names.size
      } and every(validVolume(container, enabledFeatures, secrets))

      update.uris is optional(every(api.v2.Validation.uriIsValid) and isTrue(
        "may not be set in conjunction with fetch") { (uris: Seq[String]) =>
          !(uris.nonEmpty && update.fetch.fold(false)(_.nonEmpty))
        })

```

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: ✅

src/main/scala/mesosphere/marathon/api/v2/Validation.scala
timcharper: ✅

src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionTest.scala
```diff
@@ -248,62 +187,57 @@ class AppDefinitionTest extends UnitTest with ValidationTestLike {
...
       }
 
       app = correct.copy(
         resources = Resources(gpus = 1),
-        container = Some(Container.Docker())
-      )
+        container = Some(Container.Docker()))
 
-      shouldViolate(app, "/", "GPU resources only work with the Mesos containerizer")
+      validator(app) should haveViolations("/" -> "GPU resources only work with the Mesos containerizer")
```
jeschkies: Splendid!

PR #5555, opened 2017-10-02
Fixes bug in which some Condition values were improperly read

timcharper:
The Condition "Finished" was read as "Unknown" from the storage layer

JIRA Issues: MARATHON-7806


ichernetsky:
Should it be backported?

timcharper:
@ichernetsky yes - we should definitely back-port this

PR #5556, opened 2017-10-02
[WIP] Remove formats import from marshalling. 

jeschkies:
Summary:
This change forces us to define all API models in RAML and not convert from internal models straight to HTML entities but always to RAML models first. 

JIRA issues: MARATHON-7803

### DIFF
--- a/docs/docs/rest-api/public/api/v2/types/app.raml
+++ b/docs/docs/rest-api/public/api/v2/types/app.raml
@@ -20,30 +20,6 @@ uses:
   killSelection: killSelection.raml
 
 types:
-  AppInfo:
-    type: object
-    properties:
-      app:
-        type: App
-        properties:
-        # Task counts
-          tasksStaged?:
-            type: number
-            format: int32
-          tasksRunning?:
-            type: number
-            format: int32
-          tasksHealthy?:
-            type: number
-            format: int32
-          tasksUnhealthy?:
-            type: number
-            format: int32
-       #   deployments?: Deployment[]
-       #   readinessCheckResults?: ReadinessCheckResult[]
-       #   tasks?: EnrichedTask[]
-       #   lastTaskFailures?: TaskFailure
-       #   taskStats?: TaskStatsByVersion
   AppResidency:
     type: object
     description: |
--- a/docs/docs/rest-api/public/api/v2/types/app.raml
+++ b/docs/docs/rest-api/public/api/v2/types/app.raml
@@ -25,25 +25,25 @@ types:
     properties:
       app:
         type: App
-        # properties:
+        properties:
         # Task counts
-      #  tasksStaged?:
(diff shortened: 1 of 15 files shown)


docs/docs/rest-api/public/api/v2/types/app.raml
```diff
@@ -20,6 +20,28 @@ uses:
...
+      tasksUnhealthy:
+        type: number
+        format: int32
+  AppInfo:
+    type: object
+    properties:
+      app:
+        type: App
+      # TODO: It is actually flattened
+      #task_counts?: TaskCounts
```
jeschkies: This is a blocker.

docs/docs/rest-api/public/api/v2/types/pluginDefinition.raml
```diff
@@ -0,0 +1,14 @@
+#%RAML 1.0 Library
+types:
+  PluginDefinition:
+    type: object
+    properties:
+      id: string
+      plugin: string
+      implementation: string
+      tags?: string[]
```
jeschkies: This is interpreted as `Option[Seq[String]]`. What do you think about treating it as an empty string instead?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PluginsController.scala
```diff
@@ -39,7 +40,18 @@ class PluginsController(
     * GET /v2/plugins
     * @return plugin definitions for all available plugins
     */
-  def plugins: Route = complete(definitions)
+  def plugins: Route = {
+    val plugins: Seq[raml.PluginDefinition] = definitions.plugins.map { internal =>
```
jeschkies: Should this be moved to `PluginDefinitionConversion`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -105,7 +106,8 @@ class AppsController(
           reject(Rejections.EntityNotFound.app(appId))
         case Some(info) =>
           authorized(ViewResource, info.app).apply {
-            complete(Json.obj("app" -> info))
+            val ramlAppInfo = raml.AppInfo(app = raml.Raml.toRaml(info.app))
```
jeschkies: `raml.Raml.toRaml` is obviously not nice. However, I am really in favour of being explicit here. I don't see a great benefit in `info.app.toRaml`.

jeschkies:
I'll pull out the change for plugins so this will become mergable.

jeschkies:
This grew out of proportations. See https://github.com/mesosphere/marathon/pull/5571 for `EnrichedTasks` models.

jeschkies:
Let's close this but keep the branch.

PR #5558, opened 2017-10-02
Fixes bug in which some Condition values were improperly read (#5555)

timcharper:
Summary:
The Condition "Finished" was read as "Unknown" from the storage layer.

Backport-Of: 1e9ac43

JIRA Issues: MARATHON-7806


timcharper:
Failures appear to be flakes

PR #5513, opened 2017-09-18
Replace Phabricator client with GitHub client.

jeschkies:
Summary:
We want every pull request build to make the same reports as Phabricator
builds do.

Test Plan: manually

Subscribers: jenkins, marathon-team

JIRA issues: MARATHON-7602

PR #5560, opened 2017-10-03
Replace phabricator client with github client also in documentation

alenkacz:
Summary:
Since phabricatorClient was removed in #5513 I think these docs should have been updated too. Feel free to alter the final text - I did my best since I did not develop that feature :-)

JIRA issues: MARATHON-7602

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -47,33 +47,20 @@ The `createPackages` target assembles Marathon binary packages and generates the
 sha1 checksums for the zip and tarball packages. See `createPackageSha1s` in the
 code base for details.
 
-There are several targets in `phabricatorClient`:
+There are several targets in `githubClient`:
 
-  * `accept`
   * `reject`
   * `comment`
-  * `reportTestResults`
+  * `collectTestResults`
   * `reportSuccess`
   * `reportFailure`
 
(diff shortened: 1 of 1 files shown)


ci/README.md
```diff
@@ -47,33 +47,20 @@ The `createPackages` target assembles Marathon binary packages and generates the
...
-
-Methods `accept` and `reject` accept or reject a Phabriactor Differential review.
-The `comment` method posts a comment to Phabriactor Differential reviews.
-`reportTestResults` uploads unit and integration test results to a build
-triggered by Phabriactor Harbormaster. It checks `target/phabricator-test-reports/`
-for JSON files and joins them. The methods `reportSuccess` and `reportFailure`
-call the previous methods to report back.
+Method `reject` and `reportFailure` rejects a GitHub pull request review while `reportSuccess` approves it.
+The `comment` method posts a comment to the pull request.
+`collectTestResults` gathers and returns unit and integration test results. It checks `target/phabricator-test-reports/`
```
unterstein: It would be super nice, if we would also could rename the `phabricator-test-reports` folder

alenkacz: there is already ticket for that https://jira.mesosphere.com/browse/MARATHON-7780

PR #5525, opened 2017-09-25
Further simplify leader election by using an event based approach

timcharper:
We encapsulate Curator election logic using a generic Akka Stream source
of events: ElectedLeader, Standby(currentLeader). When not leader, the
curator stream polls for current leader changes and submits events as it
changes (before, we queried every HTTP request)

By using a generic stream interface, PsuedoElectionService becomes an
incredibly simple stream that simply emits a leader elected event when
it starts, and ends with a standby event when closed.

Testing also becomes easier with event production and event reacting
concerns better separated.

Further, we remove all uses of locks in ElectionService. These were
previously a source of deadlock, and a source of unseen complexity.

We add a generic Repeater BroadcastHub type stream component which
allows for late subscribers to get the last published event (or
exception).

Further, we previously had leader ready state concerns meddled with the
election service. Leader ready is the responsibility of
MarathonSchedulerService, and the concern has been moved there
appropriately. This helped improve the legibility and substantially
reduced the state dependencies of election service on
MarathonSchedulerService. ElectionService simply handles "am I elected"
and no longer needs to know anything about migration state.

PR #5562, opened 2017-10-04
Further simplify leader election by using an event based approach [WIP]

timcharper:
Continuation of https://github.com/mesosphere/marathon/pull/5525

We encapsulate Curator election logic using a generic Akka Stream source
of events: ElectedLeader, Standby(currentLeader). When not leader, the
curator stream polls for current leader changes and submits events as it
changes (before, we queried every HTTP request)

By using a generic stream interface, PsuedoElectionService becomes an
incredibly simple stream that simply emits a leader elected event when
it starts, and ends with a standby event when closed.

Testing also becomes easier with event production and event reacting
concerns better separated.

Further, we remove all uses of locks in ElectionService. These were
previously a source of deadlock, and a source of unseen complexity.

We add a generic Repeater BroadcastHub type stream component which
allows for late subscribers to get the last published event (or
exception).

Further, we previously had leader ready state concerns meddled with the
election service. Leader ready is the responsibility of
MarathonSchedulerService, and the concern has been moved there
appropriately. This helped improve the legibility and substantially
reduced the state dependencies of election service on
MarathonSchedulerService. ElectionService simply handles "am I elected"
and no longer needs to know anything about migration state.


timcharper:
Scaling back this change; too ambitious. Improving our bootstrap and initialization logic difficult to improve without making wide changes.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+     * 2) We query the current leader
+     * 3) We receive an event (indicating child removal / addition)
+     * 4) Repeat step 1
+     *
+     * By re-registering the watch before querying the state, we will not miss out on the latest leader change.
+     */
+    def longPollLeaderChange(retries: Int = 0): Unit = Future {
+      try {
+        if (latch.getState == LeaderLatch.State.STARTED)
+          client.getChildren
```
meln1k: Do you think it's possible to refactor this using Promise into something like `def leaderEvent: Future[WatchedEvent]` ?

timcharper: We use `Future` as a helper to run in it in the serial execution thread devoted to curator concerns.

I'm not sure I see how a Promise would improve the readability. Can you provide an example? This is an asynchronous, trampoline-style recursive loop. I think that's a pretty good pattern.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -107,23 +114,25 @@ object EventsController extends StrictLogging {
       override def preStart(): Unit = {
         super.preStart()
         val actor = getStageActor(receive)
-        eventStream.subscribe(actor.ref, classOf[MarathonEvent])
-        eventStream.subscribe(actor.ref, classOf[LocalLeadershipEvent])
+        val doCompleteStage = getAsyncCallback[Unit] { _ => completeStage() }
+        electionSubscription = leaderStateEvents.to(Sink.foreach({
```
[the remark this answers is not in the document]
timcharper: We can't. We don't want the materialized result of the sink, we want the materialized value of leaderStateEvents.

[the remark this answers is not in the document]
timcharper: (`runForeach` is a shortcut for `events.toMat(Sink.foreach { evt => ... })(Keep.right)`)

timcharper:
This commit became too large. Reduced scope version is here: https://github.com/mesosphere/marathon/pull/5592

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+     * The timeline of events looks like this
+     *
+     * 1) We register a watch
+     * 2) We query the current leader
+     * 3) We receive an event (indicating child removal / addition)
+     * 4) Repeat step 1
+     *
+     * By re-registering the watch before querying the state, we will not miss out on the latest leader change.
+     */
+    def longPollLeaderChange(retries: Int = 0): Unit = Future {
```
zen-dog: Return type should be `Future[Done?]`

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+      try {
+        if (latch.getState == LeaderLatch.State.STARTED)
```
zen-dog: I'm not sure why this method has to poll (and implement backoff). Curator's `LeaderLatch` has `await()` and `await(long timeout, TimeUnit unit)` methods which would block current thread until leadership acquired. Wouldn't that suffice?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+          client.getChildren
+            .usingWatcher(new CuratorWatcher {
+              override def process(event: WatchedEvent): Unit =
+                if (!_isCancelled) longPollLeaderChange()
+            })
+            .forPath(curatorLeaderLatchPath)
+        emitLeader()
```
zen-dog: So `emitLeader` is called every call? Or is supposed to be part of the `if` statement?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+          longPollLeaderChange(retries + 1)
+        case ex: Throwable =>
+          sq.fail(ex)
+      }
+    }(singleThreadEC)
+
+    /**
+      * Emit current leader. Does not fail on connection error
+      */
+    private def emitLeader(): Unit = {
```
zen-dog: return type ` Future[QueueOfferResult]`?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+        case Some(`hostPort`) =>
+          sq.offer(LeadershipState.ElectedAsLeader)
+        case otherwise =>
+          sq.offer(LeadershipState.Standby(otherwise))
+      }
+    }
+
+    private val closeHook: () => Unit = cancel
+
+    def start(): Unit = synchronized {
```
zen-dog: return type...

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+          logger.info("Leader latch closed")
+        } catch {
+          case ex: Throwable =>
+            logger.error("Error closing CuratorElectionStream latch", ex)
+        }
+        Try(sq.complete()) // if not already completed
+      }
+      _isCancelled
+    }
+  }
```
zen-dog: Overall I don't think I understand how we use curator's `LeaderLatch` here. The canonical recipe seems to be very simple: ```leaderLatch.start(), await() // the leadership and close() //when not needed```
but I fail to recognize it here. Maybe we can chat and you could walk me through your thought process?

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
```diff
@@ -0,0 +1,39 @@
...
+  def beforeClose(fn: () => Unit): Unit
+  def removeBeforeClose(fn: () => Unit): Unit
+}
+
+class LifeCycledCloseable[T <: Closeable](val closeable: T) extends LifeCycledCloseableLike[T] {
+  @volatile private[this] var closed: Boolean = false
+  @volatile private[this] var beforeCloseHooks: List[() => Unit] = Nil
+
+  override def close(): Unit = {
+    val hooks = synchronized {
```
zen-dog: Shouldn't we check if _closeable_ wasn't already closed?

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
```diff
@@ -0,0 +1,39 @@
...
+      closed = true
+      val oldHooks = beforeCloseHooks
+      beforeCloseHooks = Nil
+      oldHooks
+    }
+    hooks.foreach { hook =>
+      Try(hook())
+    }
+    closeable.close()
+    beforeCloseHooks = Nil
```
zen-dog: second Nil-ing outside of `synchronized` block?

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
```diff
@@ -0,0 +1,39 @@
...
+import scala.util.Try
+
+trait LifeCycledCloseableLike[T <: Closeable] extends Closeable {
+  def closeable: T
+  def beforeClose(fn: () => Unit): Unit
+  def removeBeforeClose(fn: () => Unit): Unit
+}
+
+class LifeCycledCloseable[T <: Closeable](val closeable: T) extends LifeCycledCloseableLike[T] {
+  @volatile private[this] var closed: Boolean = false
```
zen-dog: I'm pretty sure both variables doesn't need to be `volatile` since they are accessed within `synchronized` blocks.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
meln1k: My idea was to move `new CuratorWatcher` part somewhere else to have something like this:
```scala
def watchLeaderEvent: Future[WatchedEvent] = {
  val p = Promise[WatchedEvent]()
  client.getChildren
  .usingWatcher(new CuratorWatcher {
    override def process(event: WatchedEvent): Unit = p.success(event)
  }
  .forPath(curatorLeaderLatchPath)
  p.future
}
```
And then we can keep the loop: 
```scala
def longPollLeaderChange(retries: Int = 0): Unit = Future {
      try {
        if (latch.getState == LeaderLatch.State.STARTED)
          watchLeaderEvent.foreach { _ => 
            if (!_isCancelled) longPollLeaderChange()
          }
        emitLeader()
      } catch {
        case ex: KeeperException.NoNodeException if retries < 100 =>
          // Wait for node to be created
          logger.info("retrying")
          Thread.sleep(retries * 10L)
          longPollLeaderChange(retries + 1)
        case ex: Throwable =>
          sq.fail(ex)
      }
    }(singleThreadEC)
```

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
meln1k: maybe `AtomicBoolean` will be a better option here?

kensipe:
It is hard to tell what is happening with this PR... twice there is the commitment to abandon but it is unclear if that is happening...

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: We discussed this in Slack; the reason is that given Marathon A, B, and C. I'll add a comment

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -32,6 +32,7 @@ object ModuleNames {
   final val HISTORY_ACTOR_PROPS = "HISTORY_ACTOR_PROPS"
 
   final val MESOS_HEARTBEAT_ACTOR = "MesosHeartbeatActor"
+  final val ELECTION_EXECUTOR = "ELECTION_EXECUTOR"
```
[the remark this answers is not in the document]
timcharper: ¯\_(ツ)_/¯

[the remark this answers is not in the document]
timcharper: Patch went through a few iterations. No reason to use Guice for this in the end :) removed.

src/main/scala/mesosphere/marathon/MarathonSchedulerService.scala
```diff
@@ -109,6 +111,8 @@ class MarathonSchedulerService @Inject() (
   // be reused (i.e. once stopped they can't be started again. Thus,
   // we have to allocate a new driver before each run or after each stop.
   var driver: Option[SchedulerDriver] = None
+  @volatile private[this] var _leaderReady = false
+  def leaderReady = _leaderReady
```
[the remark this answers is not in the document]
timcharper: This was removed in a later revision of the patch ✅

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+      zooKeeperLeaderPath: String,
+      hostPort: String,
+      sq: SourceQueueWithComplete[LeadershipState]) extends Cancellable {
+
+    val client = clientCloseable.closeable
+    private lazy val singleThreadEC: ExecutionContext = ExecutionContext.fromExecutor(singleThreadExecutor)
+    private lazy val leaderHostPortMetric: Timer = Metrics.timer(ServiceMetric, getClass, "current-leader-host-port")
+    private val curatorLeaderLatchPath = zooKeeperLeaderPath + "-curator"
+    private lazy val latch = new LeaderLatch(client, curatorLeaderLatchPath, hostPort)
+    @volatile private var isStarted = false
```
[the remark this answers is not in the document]
timcharper: yeah, you're right. I'm always afraid that some future change will break that convention :)

✅

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+    @volatile private var _isCancelled = false
+
+    /* It is important that we re-register our watch _BEFORE_ we get the current the current leader. In Zookeeper,
```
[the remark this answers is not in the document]
timcharper: ✅

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: No. :) Switched to `singleThreadEC.execute { () => ` for clarity.

With Scala 2.11 we had to instantiate the Runnable manually but with 2.12 we can just pass a Function0 🎉 🎉 🎉

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Added docs ✅

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: If latch is closed emitLeader will emit a `Standby(None)`

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Not sure if this is helpful. `emitLeader()` emits the event on the event stream. Seems straightforward to me?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Can you expound? What's wrong with the return type?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Discussed. Added explanation to longPollLeaderChange() method. :)

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
timcharper: Maybe? What problem will it avoid? Closing is an idempotent operation. Calling `close` twice on a java.io.InputStream does not throw an exception.

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
timcharper: Yeah, good catch; this is not necessary.

src/main/scala/mesosphere/marathon/util/LifeCycledCloseable.scala
timcharper: ✅ 

Put the whole state in a single `AtomicReference`, instead :). This completely avoids the potential deadlock scenario (callings hooks inside of synchronized) and reduces any confusion about what is volatile.

PR #5565, opened 2017-10-04
Region-aware capability in marathon

meln1k:
Summary: added region-aware capability and store the default region in a MarathonScheduler

JIRA issues: MARATHON_EE-1663

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -30,7 +30,7 @@ class MarathonScheduler @Inject() (
   private[this] val log = LoggerFactory.getLogger(getClass.getName)
 
   private var lastMesosMasterVersion: Option[SemanticVersion] = Option.empty
-  @volatile private[this] var defaultRegion: Option[String] = Option.empty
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -144,9 +148,19 @@ class MarathonScheduler @Inject() (
     }
   }
 
+  protected def updateDefaultRegion(masterInfo: MasterInfo): Unit = {
```
alenkacz: It is called home region in the design doc so maybe we should stick with that? https://github.com/mesosphere/marathon-design/commit/813c6a1f303b12edb65a0ca1c8eae41c8f6e9d7f#diff-af3ac823b06139f90539055737e0a188

timcharper: Yeah, let's stick with home region nomenclature; I think this is what they are calling it in other DCOS components

timcharper:
Looks like you also solved https://jira.mesosphere.com/browse/MARATHON_EE-1664 at the same time, @meln1k ?

meln1k:
@timcharper it looks like it :)

PR #5566, opened 2017-10-04
Do sync before reading from or writing to ZooKeeper

ichernetsky:
In addition to that,

- Prevent any access to the persistence storage while
  not being a leader, which means that Marathon reads from
  and writes to ZK only after it becomes a leader.
- Register service.mesosphere.marathon.app.count and
  service.mesosphere.marathon.app.count metrics once
  the root group gets loaded for the first time

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-7763
https://jira.mesosphere.com/browse/MARATHON-7785

It is a cherry-pick of https://github.com/mesosphere/marathon/commit/5137ed9949c98f0c759bd2ba2b6ce6f8cd4d60db

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
@@ -21,7 +21,19 @@ import scala.collection.immutable.Seq
   */
 trait GroupManager {
 
+  /**
+    * Get a root group, fetching it from a persistence store if necessary.
+    *
+    * @return a root group
+    */
   def rootGroup(): RootGroup
+
+  /**
+    * Get a root group.
+    *
+    * @return a root group if it has been fetched already, otherwise [[None]]
+    */
+  def rootGroupOption(): Option[RootGroup]
 
   /**
     * Get all available versions for given group identifier.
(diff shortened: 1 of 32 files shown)


src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
```diff
@@ -21,9 +21,21 @@ import scala.collection.immutable.Seq
...
+    * @return a root group
+    */
   def rootGroup(): RootGroup
 
   /**
+    * Get a root group.
+    *
+    * @return a root group if it has been fetched already, otherwise [[None]]
+    */
+  def rootGroupOption(): Option[RootGroup]
```
meln1k: how about `findRootGroup`?

ichernetsky: I named it in line with the Scala standard library. Like `head` and `headOption`. What do you think?

zen-dog: It's like this in the original patch so let's not change it in the cherry-pick (unless it's a bug)

PR #5571, opened 2017-10-05
Model EnrichedTask in RAML.

jeschkies:
Summary:
Model the `EnrichedTask` response in RAML. This is the first step of removing the infamous handwritten `Formats.scala`. I used the raw JSON in `EnrickedTaskWritesTest` as examples.

JIRA issues:

### DIFF
--- a/docs/docs/rest-api/public/api/api.raml
+++ b/docs/docs/rest-api/public/api/api.raml
@@ -7,6 +7,7 @@ baseUri: ../
 
 uses:
   app: v2/types/app.raml
+  enrichedTask: v2/types/enrichedTask.raml
   error: v2/types/error.raml
   info: v2/types/info.raml
   queue: v2/types/queue.raml
--- a/docs/docs/rest-api/public/api/v2/types/healthCheck.raml
+++ b/docs/docs/rest-api/public/api/v2/types/healthCheck.raml
@@ -148,3 +148,15 @@ types:
         description: Amount of time to wait until starting the health checks.
         minimum: 0
         default: 15
+  Health:
+    type: object
+    properties:
+      alive: boolean
+      consecutiveFailures:
+        type: number
+        format: int32
+      firstSuccess?: string
+      instanceId: strings.InstanceId
+      lastSuccess?: string
+      lastFailure?: string
+      lastFailureCause?: string
--- a/docs/docs/rest-api/public/api/v2/types/stringTypes.raml
+++ b/docs/docs/rest-api/public/api/v2/types/stringTypes.raml
@@ -33,6 +33,12 @@ types:
     type: string
     enum: [ HTTP, HTTPS ]
     description: The http scheme to use
+  IpProtocol:
+    type: string
+    enum: [ IPv4 ]
+  PersistenceId:
+    type: string
+    pattern: ^([^#]+)[#]([^#]+)[#]([^#]+)$
   ReadMode:
     type: string
     enum: [RO, RW]
--- a/docs/docs/rest-api/public/api/v2/types/stringTypes.raml
+++ b/docs/docs/rest-api/public/api/v2/types/stringTypes.raml
@@ -1,5 +1,8 @@
 #%RAML 1.0 Library
 types:
+  InstanceId:
+    type: string
+    pattern: ^(.+)\.(instance-|marathon-)([^\.]+)$
   Name:
     type: string
     pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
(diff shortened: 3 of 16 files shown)


jeschkies:
Should I also replace the tests in `EnrichedTaskWritesTest` with conversions? This way we can be sure that the everything still works.

src/main/scala/mesosphere/marathon/raml/DefaultConversions.scala
```diff
@@ -37,4 +42,22 @@ trait DefaultConversions {
...
+
+  implicit val timestampWrites: Writes[Timestamp, String] = Writes { _.toString }
+
+  implicit val pathIdWrites: Writes[PathId, String] = Writes { _.toString }
+
+  implicit val instanceIdWrites: Writes[Instance.Id, String] = Writes { _.toString }
+
+  implicit val taskStateWrites: Writes[mesos.TaskState, MesosTaskState] = Writes { taskState =>
+    val name = taskState.name()
+    MesosTaskState.fromString(name).getOrElse(throw new IllegalArgumentException(s"$name is an unknown Mesos task state"))
```
jeschkies: The exception should not really happen since it's a write. What shall we do here?

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
```diff
@@ -92,7 +94,7 @@ class TasksResource @Inject() (
 
     val enrichedTasks: Iterable[EnrichedTask] = result(futureEnrichedTasks)
     ok(jsonObjString(
-      "tasks" -> enrichedTasks
+      "tasks" -> enrichedTasks.toIndexedSeq.toRaml
```
meln1k: why not `.toVector`?

src/main/scala/mesosphere/marathon/raml/DefaultConversions.scala
meln1k: I think if it's something that shouldn't really happen then it's ok to use an exception.

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
jeschkies: No specific reason. Mostly because it is more general than a vector.

jeschkies:
> maybe just use the new types in docs somehow, otherwise LGTM 👍

Good point. I'll follow up.

docs/docs/rest-api/public/api/v2/types/stringTypes.raml
```diff
@@ -49,3 +58,18 @@ types:
...
+      - TASK_ERROR
+      - TASK_FAILED
+      - TASK_FINISHED
+      - TASK_KILLED
+      - TASK_KILLING
+      - TASK_RUNNING
+      - TASK_STAGING
+      - TASK_STARTING
+      - TASK_UNREACHABLE
+      - TASK_UNREACHABLE
```
meichstedt: I think the second `TASK_UNREACHABLE` can be removed :)

docs/docs/rest-api/public/api/api.raml
```diff
@@ -7,6 +7,7 @@ baseUri: ../
 
 uses:
   app: v2/types/app.raml
+  enrichedTask: v2/types/enrichedTask.raml
```
meichstedt: IIRC the `EnrichedTask` format was named as such because it serialized more information that a task alone would have. I think it's OK to keep this name, but from an API perspective, it _could_ also just be named `task.raml`.

docs/docs/rest-api/public/api/v2/types/healthCheck.raml
```diff
@@ -148,3 +148,15 @@ types:
...
         minimum: 0
         default: 15
+  Health:
+    type: object
+    properties:
+      alive: boolean
+      consecutiveFailures:
+        type: number
+        format: int32
+      firstSuccess?: string
```
meichstedt: isn't this a `datetime`? Or is the format different? Compare e.g. to `app.version`.

docs/docs/rest-api/public/api/v2/types/healthCheck.raml
```diff
@@ -148,3 +148,15 @@ types:
...
+      instanceId: strings.InstanceId
+      lastSuccess?: string
+      lastFailure?: string
```
meichstedt: same for these two: `datetime`?

src/test/scala/mesosphere/marathon/api/v2/json/EnrichedTaskWritesTest.scala
```diff
@@ -77,17 +78,20 @@ class EnrichedTaskWritesTest extends UnitTest {
...
         |{
         |  "appId": "${f.runSpecId}",
+        |  "healthCheckResults" : [],
         |  "id": "${f.taskWithoutIp.task.taskId.idString}",
+        |  "ipAddresses" : [],
         |  "host": "agent1.mesos",
         |  "state": "TASK_STAGING",
         |  "ports": [],
-        |  "startedAt": null,
+        |  "servicePorts" : [],
```
meichstedt: If I interpret this change correctly, it means that previously, there was always a field `"startedAt"`, even if the value was null. That field does not seem to be the case after this change. Given that tooling might expect this field to be present, wouldn't this be a breaking change?

PR #5572, opened 2017-10-05
Ignore offers from non-home regions

alenkacz:
Summary:
Until the remote region support is finished we want to be able to pre-filter all offers from remote regions and ignore them.
This logic will be adjusted we move further with remote regions.

JIRA issues: MARATHON_EE-1662

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
@@ -17,7 +17,7 @@ import scala.concurrent.{ Future, Promise }
   * @param actorRef Reference to actor that matches offers.
   * @param precedenceFor Defines which matcher receives offers first. See [[mesosphere.marathon.core.matcher.base.OfferMatcher.precedenceFor]].
   */
-class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId])(implicit scheduler: akka.actor.Scheduler)
+class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], marathonScheduler: MarathonScheduler)(implicit scheduler: akka.actor.Scheduler)
     extends OfferMatcher with StrictLogging {
 
   def matchOffer(offer: Offer): Future[MatchedInstanceOps] = {
@@ -27,6 +27,14 @@ class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[P
   }
 
   override def toString: String = s"ActorOfferMatcher($actorRef)"
+
+  override def isInterestedIn(offer: Offer) = {
+    def isFromHomeRegion(offer: Offer): Boolean = {
+      !offer.hasDomain || !offer.getDomain.hasFaultDomain || marathonScheduler.getHomeRegion.forall(_ == offer.getDomain.getFaultDomain.getRegion.getName)
+    }
+
+    isFromHomeRegion(offer)
+  }
 }
 
 object ActorOfferMatcher {
--- a/src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
+++ b/src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
@@ -2,6 +2,7 @@ package mesosphere.marathon
 package core.matcher.base.util
 
 import akka.actor.ActorRef
+import com.google.inject.Provider
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.matcher.base.OfferMatcher
 import mesosphere.marathon.core.matcher.base.OfferMatcher.MatchedInstanceOps
@@ -17,7 +18,7 @@ import scala.concurrent.{ Future, Promise }
   * @param actorRef Reference to actor that matches offers.
   * @param precedenceFor Defines which matcher receives offers first. See [[mesosphere.marathon.core.matcher.base.OfferMatcher.precedenceFor]].
   */
-class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], marathonScheduler: MarathonScheduler)(implicit scheduler: akka.actor.Scheduler)
+class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], marathonScheduler: Provider[MarathonScheduler])(implicit scheduler: akka.actor.Scheduler)
     extends OfferMatcher with StrictLogging {
 
   def matchOffer(offer: Offer): Future[MatchedInstanceOps] = {
@@ -30,7 +31,7 @@ class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[P
 
   override def isInterestedIn(offer: Offer) = {
     def isFromHomeRegion(offer: Offer): Boolean = {
-      !offer.hasDomain || !offer.getDomain.hasFaultDomain || marathonScheduler.getHomeRegion.forall(_ == offer.getDomain.getFaultDomain.getRegion.getName)
+      !offer.hasDomain || !offer.getDomain.hasFaultDomain || marathonScheduler.get().getHomeRegion.forall(_ == offer.getDomain.getFaultDomain.getRegion.getName)
     }
 
     isFromHomeRegion(offer)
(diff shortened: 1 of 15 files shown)


src/test/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActorTest.scala
```diff
@@ -195,6 +195,21 @@ class OfferMatcherManagerActorTest extends AkkaUnitTest with Eventually {
       Then("offer-1 is declined, since the actor did not respond in time")
       offerMatch1.future.futureValue.opsWithSource should be('empty)
     }
+
+    "not send offers to matchers that are not interested" in new Fixture {
```
meln1k: I think we also need a test which checks the other case when matchers are interested in the offer.

alenkacz:
I hopefully fixed that issue and added more tests so feel free to review :)

src/test/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActorTest.scala
alenkacz: agreed, added another one

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
```diff
@@ -27,6 +28,14 @@ class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[P
   }
 
   override def toString: String = s"ActorOfferMatcher($actorRef)"
+
+  override def isInterestedIn(offer: Offer) = {
+    def isFromHomeRegion(offer: Offer): Boolean = {
```
jeschkies: Out of curiosity, why is this inside the method and not outside?

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
```diff
@@ -174,7 +174,7 @@ private[impl] class OfferMatcherManagerActor private (
       .map(_.getDisk.getPersistence.getId)
       .collect { case LocalVolumeId(volumeId) => volumeId.runSpecId }
       .toSet
-    val (reserved, normal) = matchers.toSeq.partition(_.precedenceFor.exists(appReservations))
+    val (reserved, normal) = matchers.toSeq.filter(_.isInterestedIn(offer)).partition(_.precedenceFor.exists(appReservations))
```
jeschkies: As I remember the proposal was to register with an interest in offers or not. What is the motivation to ask the offer matcher instead of being told by it? I like this approach more but am just curious.

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
```diff
@@ -17,7 +18,7 @@ import scala.concurrent.{ Future, Promise }
   * @param actorRef Reference to actor that matches offers.
   * @param precedenceFor Defines which matcher receives offers first. See [[mesosphere.marathon.core.matcher.base.OfferMatcher.precedenceFor]].
   */
-class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId])(implicit scheduler: akka.actor.Scheduler)
+class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], marathonScheduler: Provider[MarathonScheduler])(implicit scheduler: akka.actor.Scheduler)
```
jeschkies: Do we really want to inject the scheduler or an abstraction? I have an uneasy feeling about the `MarathonScheduler` leaking into this code.

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
```diff
@@ -28,12 +68,36 @@ class ActorOfferMatcherTest extends AkkaUnitTest {
...
+        })(scheduler)
         val offerMatch: MatchedInstanceOps = offerMatcher.matchOffer(offer).futureValue
 
         offerMatch.offerId should not be (offer.getId)
         offerMatch.offerId.getValue should be("other")
       }
     }
+
+    def offerWithFaultRegion(faultRegion: String) = {
+      Offer.newBuilder()
```
jeschkies: I really wish we would move to a Scala native protobuf package with case classes to get rig of these builders. Anyhow, thanks for the factory method :).

In the passed we would follow the pattern of a `class Fixture` which would hold this method and then create the fixture in each test. What do you think?

src/test/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActorTest.scala
```diff
@@ -195,6 +195,36 @@ class OfferMatcherManagerActorTest extends AkkaUnitTest with Eventually {
...
+      val offerMatch1 = Promise[OfferMatcher.MatchedInstanceOps]
+      offerMatcherManager.underlyingActor.launchTokens = 100
+      val matcherMock = matcher(isInterestedIn = false)
+      offerMatcherManager.underlyingActor.matchers += matcherMock
+
+      When("Offer is sent to MatcherManager")
+      offerMatcherManager ! ActorOfferMatcher.MatchOffer(offer1, offerMatch1)
+
+      Then("OfferMatcher not interested in offer should not receive any offer")
+      verify(matcherMock, times(0)).matchOffer(any)
```
jeschkies: Isn't this flaky? There is no guarantee that the message arrived when we do the check.

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
alenkacz: by an abstraction you mean introducing some trait (something like HomeRegionProvider) and injecting that instead of the whole scheduler (although it will be implemented by the scheduler class)?

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
alenkacz: I introduced this function just to improve readability so I like to nest it within the function that's using it (if it is actually not that meaningful for ActorOfferMatcher itself)

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
alenkacz: In the end I moved toward this approach as I found it fitting what we wanted to do the most... But I am up for a discussion if you or someone else would approach it differently...

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
alenkacz: by scala protobuf you mean this one? https://github.com/scalapb/ScalaPB we actually did not like about it that it requires python installed on your computer...

alenkacz: regarding the fixture - will do :)

src/test/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActorTest.scala
alenkacz: good point, I changed those...

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
jeschkies: Hm, @meichstedt proposed the filter on registration approach. My vote would go to yours.

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
jeschkies: Oh, yeah. That is a minus. Is there no alternative around? @timcharper, you basically know every Scala package right? Do you know an alternative?

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
jeschkies: I should have been more precise. It could be a trait or just a case class.

PR #5573, opened 2017-10-05
Refactoring of AppsController routes

meln1k:
Summary: Moved routes structure in a different commit, so it will be easier to review separate controller method implementation

JIRA issues: MARATHON-7299

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -50,68 +50,35 @@ class AppsController(
 
   import mesosphere.marathon.api.v2.json.Formats._
 
-  private def listApps(implicit identity: Identity): Route = {
-    parameters('cmd.?, 'id.?, 'label.?, 'embed.*) { (cmd, id, label, embed) =>
-      def index: Future[Seq[AppInfo]] = {
-        def containCaseInsensitive(a: String, b: String): Boolean = b.toLowerCase contains a.toLowerCase
+  private val forceParameter = parameter('force.as[Boolean].?(false))
 
-        val selectors = Seq[Option[Selector[AppDefinition]]](
-          cmd.map(c => Selector(_.cmd.exists(containCaseInsensitive(c, _)))),
-          id.map(s => Selector(app => containCaseInsensitive(s, app.id.toString))),
-          label.map(new LabelSelectorParsers().parsed),
-          Some(authzSelector)
-        ).flatten
-        val resolvedEmbed = InfoEmbedResolver.resolveApp(embed.toSet) + AppInfo.Embed.Counts + AppInfo.Embed.Deployments
-        appInfoService.selectAppsBy(Selector.forall(selectors), resolvedEmbed)
-      }
-      onSuccess(index)(apps => complete(Json.obj("apps" -> apps)))
-    }
-  }
+  private def listApps(implicit identity: Identity): Route = ???
 
-  private def createApp(app: AppDefinition, force: Boolean)(implicit identity: Identity): Route = {
-    def create: Future[(DeploymentPlan, AppInfo)] = {
+  private def replaceMultipleApps(implicit identity: Identity): Route = ???
 
-      def createOrThrow(opt: Option[AppDefinition]) = opt
-        .map(_ => throw ConflictingChangeException(s"An app with id [${app.id}] already exists."))
-        .getOrElse(app)
+  private def patchMultipleApps(implicit identity: Identity): Route = ???
 
-      groupManager.updateApp(app.id, createOrThrow, app.version, force).map { plan =>
-        val appWithDeployments = AppInfo(
-          app,
-          maybeCounts = Some(TaskCounts.zero),
-          maybeTasks = Some(Seq.empty),
-          maybeDeployments = Some(Seq(Identifiable(plan.id)))
-        )
-        plan -> appWithDeployments
-      }
-    }
-    authorized(CreateRunSpec, app).apply {
-      onSuccess(create) { (plan, app) =>
-        //TODO: post ApiPostEvent
-        complete((StatusCodes.Created, Seq(Headers.`Marathon-Deployment-Id`(plan.id)), app))
-      }
-    }
-  }
+  private def createApp(implicit identity: Identity): Route = ???
 
-  private def showApp(appId: PathId)(implicit identity: Identity): Route = {
-    parameters('embed.*) { embed =>
-      val resolvedEmbed = InfoEmbedResolver.resolveApp(embed.toSet) ++ Set(
-        // deprecated. For compatibility.
-        AppInfo.Embed.Counts, AppInfo.Embed.Tasks, AppInfo.Embed.LastTaskFailure, AppInfo.Embed.Deployments
-      )
+  private def showApp(appId: PathId)(implicit identity: Identity): Route = ???
 
-      onSuccess(appInfoService.selectApp(appId, authzSelector, resolvedEmbed)) {
-        case None =>
-          reject(Rejections.EntityNotFound.app(appId))
-        case Some(info) =>
-          authorized(ViewResource, info.app).apply {
-            complete(Json.obj("app" -> info))
-          }
-      }
-    }
-  }
+  private def patchSingle(appId: PathId)(implicit identity: Identity): Route = ???
 
-  val RemainingPathId = RemainingPath.map(_.toString.toRootPath)
+  private def putSingle(appId: PathId)(implicit identity: Identity): Route = ???
+
+  private def deleteSingle(appId: PathId)(implicit identity: Identity): Route = ???
+
+  private def restartApp(appId: PathId)(implicit identity: Identity): Route = ???
+
+  private def listRunningTasks(appId: PathId)(implicit identity: Identity): Route = ???
+
+  private def killTasks(appId: PathId)(implicit identity: Identity): Route = ???
+
+  private def killTask(appId: PathId, taskId: TaskId)(implicit identity: Identity): Route = ???
+
+  private def listVersions(appId: PathId)(implicit identity: Identity): Route = ???
+
+  private def getVersion(appId: PathId, version: Timestamp)(implicit identity: Identity): Route = ???
 
   // format: OFF
   val route: Route = {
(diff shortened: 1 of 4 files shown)


timcharper:
I have mixed feelings :) On the one hand, I see the value in being able to focus on reviewing and negotiating routes, on the other hand, it seems like this approach compromises the atomicity of our commits?

Just a thought. I don't have an extremely strong objection :) Interested to hear what others on the team think.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -50,93 +53,101 @@ class AppsController(
...
-    authorized(CreateRunSpec, app).apply {
-      onSuccess(create) { (plan, app) =>
-        //TODO: post ApiPostEvent
-        complete((StatusCodes.Created, Seq(Headers.`Marathon-Deployment-Id`(plan.id)), app))
-      }
-    }
-  }
+  private def listApps(implicit identity: Identity): Route = ???
 
-  private def showApp(appId: PathId)(implicit identity: Identity): Route = {
```
jeschkies: Why is this removed?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+      val entity = HttpEntity(body).withContentType(ContentTypes.`application/json`)
+
+      Post(Uri./, entity) ~> route ~> check {
+        Then("It is successful")
+        status shouldEqual StatusCodes.Created
+        header[Headers.`Marathon-Deployment-Id`] should not be 'empty
+
+        And("the JSON is as expected, including a newly generated version")
+        import mesosphere.marathon.api.v2.json.Formats._
+        val expected = AppInfo(
```
jeschkies: How about using a JSON string here? What do you think? This avoids serialization

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+          normalizeAndConvert(app).copy(versionInfo = VersionInfo.OnlyVersion(clock.now())),
+          maybeTasks = Some(immutable.Seq.empty),
+          maybeCounts = Some(TaskCounts.zero),
+          maybeDeployments = Some(immutable.Seq(Identifiable(plan.id)))
+        )
+
+        Try(JsonTestHelper.assertThatJsonString(responseAs[String]).correspondsToJsonOf(expected))
```
jeschkies: Why wrapping it in a try?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+
+    "Create a new app with IP/CT, no default network name, Alice does not specify a network" in new Fixture {
+      Given("An app and group")
+      val app = App(
+        id = "/app",
+        cmd = Some("cmd"),
+        networks = Seq(Network(mode = NetworkMode.Container))
+      )
+      the[NormalizationException] thrownBy {
+        prepareApp(app, groupManager)
```
jeschkies: Is this the right spot for testing this? Isn't this independent of the controller?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+      clock += 5.seconds
+      val entity = HttpEntity(body).withContentType(ContentTypes.`application/json`)
+      Post(Uri./, entity) ~> route ~> check {
+        Then("It is successful")
+        status shouldEqual StatusCodes.Created withClue s"body = ${ByteString(body).utf8String}, response = ${responseAs[String]}"
+        header[Headers.`Marathon-Deployment-Id`] should not be 'empty
+
+        And("the JSON is as expected, including a newly generated version")
+        import mesosphere.marathon.api.v2.json.Formats._
+        val expected = AppInfo(
```
jeschkies: How about avoiding the serialization.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+import scala.language.postfixOps
+import scala.util.Try
+import scala.util.control.NonFatal
+
+class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRouteTest {
+
+  case class Fixture(
+      clock: SettableClock = new SettableClock(),
+      auth: TestAuthFixture = new TestAuthFixture {
+        override val UnauthorizedStatus = 403
```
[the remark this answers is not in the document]
meln1k: Yep, it is! So it is temporary workaround until we fix all the other tests depending on it

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+      healthCheckManager = healthCheckManager,
+      instanceTracker = instanceTracker,
+      taskKiller = taskKiller,
+      config = config,
+      groupManager = groupManager,
+      pluginManager = PluginManager.None
+    )
+
+    implicit val rejectionHandler: RejectionHandler = AkkaHttpMarathonService.rejectionHandler
+    val route: Route = Route.seal(appsController.route)
```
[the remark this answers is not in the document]
meln1k: We want to make sure that our own rejection handler is used instead of some default.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+      val migrated = AppNormalization.forDeprecated(normalizationConfig).normalized(app)
+      AppNormalization(normalizationConfig).normalized(migrated)
+    }
+
+    def normalizeAndConvert(app: App): AppDefinition = {
+      val normalized = normalize(app)
+      Raml.fromRaml(normalized)
+    }
+
+    def prepareApp(app: App, groupManager: GroupManager): (Array[Byte], DeploymentPlan) = {
```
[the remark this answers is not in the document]
meln1k: If I understand it correctly, it will update mocked group manager with appropriate values for the given app.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1433 @@
...
+           |  }
+           |}
+      """.stripMargin
+
+      When("The request is processed")
+      HttpEntity(body).withContentType(ContentTypes.`application/json`)
+    }
+  }
+
+  case class FixtureWithRealGroupManager(
```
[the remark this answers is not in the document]
meln1k: Because we need a real implementation here to simulate a non-existing app.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: I don't know what's wrong with serialization, but yes, I can do it.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: I'm not sure it's possible because deployment id is generated dynamically.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: It seems it's not possible because of the dynamic id generation.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
jeschkies: The can be easily set with string interpolation.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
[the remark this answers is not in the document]
jeschkies: Aren't we breaking the API then?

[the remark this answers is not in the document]
jeschkies: Also, I've just checked. This value is never used. I'd then remove these changes,

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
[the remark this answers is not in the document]
jeschkies: Hm, this seems too much for a test to me but I understand that you've just copied it. Since I cannot think of a better solution for now I'm fine with it.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
jeschkies: Just used string interpolation :).

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -0,0 +1,1856 @@
...
+          |        "provider": "dvdi",
+          |        "name": "namedfoo",
+          |        "options": {"dvdi/driver": "bar"}
+          |      },
+          |      "mode": "RW"
+          |    }]
+        """.stripMargin, groupManager, auth
+      )
+
+      Then("The return code indicates create success")
```
jeschkies: I think we should check the rejections instead. Let's decide tomorrow to follow up. I don't see this as a blocker for this diff.

jeschkies: I think these should check the rejections.

PR #5574, opened 2017-10-05
Add constraint support for Zone and Region

timcharper:
* Zone / Region info added to agentInfo
* Consolidate matcher code

JIRA Issues: MARATHON_EE-1665

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -8,6 +8,7 @@ import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.instance.Instance.{ AgentInfo, InstanceState }
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.state.{ MarathonState, PathId, Timestamp, UnreachableStrategy, UnreachableDisabled, UnreachableEnabled }
+import mesosphere.marathon.tasks.OfferUtil
 import mesosphere.marathon.stream.Implicits._
 import mesosphere.mesos.Placed
 import mesosphere.marathon.raml.Raml
@@ -69,6 +70,10 @@ case class Instance(
   override def hostname: String = agentInfo.host
 
   override def attributes: Seq[Attribute] = agentInfo.attributes
+
+  override def zone: Option[String] = agentInfo.zone
+
+  override def region: Option[String] = agentInfo.region
 }
 
 @SuppressWarnings(Array("DuplicateImport"))
@@ -267,12 +272,16 @@ object Instance {
   case class AgentInfo(
     host: String,
     agentId: Option[String],
-    attributes: Seq[mesos.Protos.Attribute])
+    region: Option[String],
+    zone: Option[String],
+    attributes: Seq[Attribute])
 
   object AgentInfo {
     def apply(offer: org.apache.mesos.Protos.Offer): AgentInfo = AgentInfo(
       host = offer.getHostname,
       agentId = Some(offer.getSlaveId.getValue),
+      region = OfferUtil.region(offer),
+      zone = OfferUtil.zone(offer),
       attributes = offer.getAttributesList.toIndexedSeq
     )
   }
@@ -290,7 +299,7 @@ object Instance {
       throw new IllegalStateException(s"No task in ${instance.instanceId}"))
   }
 
-  implicit object AttributeFormat extends Format[mesos.Protos.Attribute] {
+  implicit object AttributeFormat extends Format[Attribute] {
     override def reads(json: JsValue): JsResult[Attribute] = {
       json.validate[String].map { base64 =>
         mesos.Protos.Attribute.parseFrom(Base64.getDecoder.decode(base64))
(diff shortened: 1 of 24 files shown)


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -267,12 +272,16 @@ object Instance {
...
-    attributes: Seq[mesos.Protos.Attribute])
+    region: String,
+    zone: String,
+    attributes: Seq[Attribute])
 
   object AgentInfo {
     def apply(offer: org.apache.mesos.Protos.Offer): AgentInfo = AgentInfo(
       host = offer.getHostname,
       agentId = Some(offer.getSlaveId.getValue),
+      region = OfferUtil.region(offer).getOrElse(""),
```
meln1k: Does it make sense to make `region` and `zone` fields optional?

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -86,77 +90,51 @@ object Constraints {
...
-        case Operator.GROUP_BY => checkGroupBy(offer.getHostname, (p: Placed) => Some(p.hostname))
-        case Operator.MAX_PER => checkMaxPer(offer.getHostname, value.toInt, (p: Placed) => Some(p.hostname))
-        case Operator.CLUSTER =>
-          // Hostname must match or be empty
-          (value.isEmpty || value == offer.getHostname) &&
-            // All running tasks must have the same hostname as the one in the offer
-            allPlaced.forall(_.hostname == offer.getHostname)
-        case _ => false
-      }
+    private def checkCluster(ov: String, placedValue: Placed => Option[String]) = {
```
meln1k: Do you think we can use another name instead of `ov`?

src/main/scala/mesosphere/mesos/Constraints.scala
```diff
@@ -53,14 +56,15 @@ object Constraints {
...
-      if (field == "hostname") {
-        checkHostName
-      } else if (attr.nonEmpty) {
-        checkAttribute
-      } else {
-        // This will be reached in case we want to schedule for an attribute
-        // that's not supplied.
-        checkMissingAttribute
+      field match {
+        case "hostname" | "*hostname" =>
```
alenkacz: did we actually decided to go with the * prefix in the end?

timcharper: I hadn't heard that! Did we?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
timcharper: I thought about this; can you think of a reason?

One thing I was considering is if region and zone are unspecified, IE, in the case of the home region, we still need to match them. Having them be None would make our matching logic more complicated (there is no way to make a constraint `region is None`, only `region is ""`)

timcharper: Thinking about it again, this may have been more valid with the implicit constraints approach. Maybe it would just be better as None. If the user wants to avoid the home region, they can specify that tasks should launch in a region where region is specified.

ichernetsky: What have you decided to use?

src/main/scala/mesosphere/mesos/Constraints.scala
ichernetsky: Sorry for bringing it up this late, but `*hostname:LIKE:.*foo.*` looks a bit confusing. The LHS looks a bit confusing as if it is a glob-style pattern. What other characters to designate special fields have been considered?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
timcharper: Went with None

✅

src/main/scala/mesosphere/mesos/Constraints.scala
timcharper: ✅

src/main/scala/mesosphere/mesos/Constraints.scala
timcharper: We had considered `:hostname:LIKE:.*foo.*`.

`<hostname>:LIKE:.*foo.*`
`{hostname}:LIKE:.*foo.*`
`&hostname:LIKE:.*foo.*`
`@hostname:LIKE:.*foo.*`
`$hostname:LIKE:.*foo.*`
`(hostname):LIKE:.*foo.*`

¯\_(ツ)_/¯ :)

Any favorites?

I'm kind of partial towards `@hostname:LIKE:.*foo.*`, now that I think about it more. But, not really opinionated.

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
```diff
@@ -88,311 +114,153 @@ class ConstraintsTest extends UnitTest {
...
+      val instance3_rack3 = makeSampleInstanceWithScalarAttrs(appId, Map(rackIdField -> 3.0))
+      val instance4_rack1 = makeSampleInstanceWithScalarAttrs(appId, Map(rackIdField -> 1.0))
+      val instance5_rack2 = makeSampleInstanceWithScalarAttrs(appId, Map(rackIdField -> 2.0))
 
       var groupRack = Seq.empty[Instance]
 
-      val groupByRack = makeConstraint("rackid", Constraint.Operator.GROUP_BY, "3")
+      val groupByRack = makeConstraint(rackIdField, GROUP_BY, "3")
 
       val clusterFreshRackMet = Constraints.meetsConstraint(
```
timcharper: I got tired of rewriting these and left half of them this way.

PR #5577, opened 2017-10-06
Add akka http queue controller

zen-dog:
Summary:
Added queue controller implementation with authentication and authorisation. Respective tests were ported from the `QueueResourceTest` class.

JIRA issues: MARATHON-7307, MARATHON-7088

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
@@ -1,6 +1,8 @@
 package mesosphere.marathon
 package api.akkahttp
 
+import java.time.Clock
+
 import akka.http.scaladsl.marshalling.{ Marshaller, ToEntityMarshaller }
 import akka.http.scaladsl.model.MediaTypes.`application/json`
 import akka.http.scaladsl.model.StatusCodes
@@ -13,6 +15,7 @@ import kamon.metric.SubscriptionsDispatcher.TickMetricSnapshot
 import mesosphere.marathon.api.v2.Validation
 import mesosphere.marathon.core.appinfo.{ AppInfo, EnrichedTask }
 import mesosphere.marathon.plugin.PathId
+import mesosphere.marathon.core.launchqueue.LaunchQueue.QueuedInstanceInfoWithStatistics
 import mesosphere.marathon.raml.{ LoggerChange, MarathonInfo, Metrics }
 import mesosphere.marathon.core.plugin.PluginDefinitions
 import mesosphere.marathon.state.AppDefinition
@@ -150,12 +153,12 @@ object EntityMarshallers {
   implicit val messageMarshaller = playJsonMarshaller[Rejections.Message]
   implicit val appInfoMarshaller = playJsonMarshaller[AppInfo]
   implicit val infoMarshaller = playJsonMarshaller[MarathonInfo]
-  implicit val infoUnmarshaller = playJsonUnmarshaller[MarathonInfo]
   implicit val metricsMarshaller = internalToRamlJsonMarshaller[TickMetricSnapshot, Metrics]
   implicit val loggerChangeMarshaller = playJsonMarshaller[LoggerChange]
   implicit val loggerChangeUnmarshaller = playJsonUnmarshaller[LoggerChange]
   implicit val stringMapMarshaller = playJsonMarshaller[Map[String, String]]
   implicit val pluginDefinitionsMarshaller = playJsonMarshaller[PluginDefinitions]
+  implicit val queueMarashaller = internalToRamlJsonMarshaller[(Seq[QueuedInstanceInfoWithStatistics], Boolean, Clock), raml.Queue]
   implicit val deploymentResultMarshaller = playJsonMarshaller[raml.DeploymentResult]
   implicit val enrichedTaskMarshaller = playJsonMarshaller[EnrichedTask]
 
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
@@ -12,10 +12,15 @@ object Rejections {
   }
 
   val defaultEntityNotFoundMessage = Message("Entity was not found")
+
   case class EntityNotFound(message: Message = defaultEntityNotFoundMessage) extends Rejection
   object EntityNotFound {
     def app(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
+    }
+
+    def queueApp(appId: PathId): EntityNotFound = {
+      EntityNotFound(Message(s"Application $appId not found in tasks queue."))
     }
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/V2Controller.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/V2Controller.scala
@@ -2,7 +2,7 @@ package mesosphere.marathon
 package api.akkahttp
 
 import akka.http.scaladsl.server.Route
-import v2.{ AppsController, EventsController, InfoController, PluginsController }
+import v2._
 
 /**
   * Dispatches to various v2 controller routes
@@ -11,7 +11,8 @@ class V2Controller(
     appsController: AppsController,
     eventsController: EventsController,
     pluginsController: PluginsController,
-    infoController: InfoController
+    infoController: InfoController,
+    queueController: QueueController
 ) extends Controller {
   import Directives._
   override val route: Route = {
@@ -26,6 +27,9 @@ class V2Controller(
       } ~
       pathPrefix("info") {
         infoController.route
+      } ~
+      pathPrefix("queue") {
+        queueController.route
       }
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
@@ -4,17 +4,22 @@ package api.akkahttp.v2
 import java.time.Clock
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.{ StatusCodes }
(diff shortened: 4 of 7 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -150,12 +153,12 @@ object EntityMarshallers {
   implicit val messageMarshaller = playJsonMarshaller[Rejections.Message]
   implicit val appInfoMarshaller = playJsonMarshaller[AppInfo]
   implicit val infoMarshaller = playJsonMarshaller[MarathonInfo]
-  implicit val infoUnmarshaller = playJsonUnmarshaller[MarathonInfo]
   implicit val metricsMarshaller = internalToRamlJsonMarshaller[TickMetricSnapshot, Metrics]
   implicit val loggerChangeMarshaller = playJsonMarshaller[LoggerChange]
   implicit val loggerChangeUnmarshaller = playJsonUnmarshaller[LoggerChange]
   implicit val stringMapMarshaller = playJsonMarshaller[Map[String, String]]
   implicit val pluginDefinitionsMarshaller = playJsonMarshaller[PluginDefinitions]
+  implicit val queueMarashaller = internalToRamlJsonMarshaller[(Seq[QueuedInstanceInfoWithStatistics], Boolean, Clock), raml.Queue]
```
meln1k: Have you considered to use a case class instead of tuple3 here?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  @SuppressWarnings(Array("all")) // async/await
+  private def list(implicit identity: Identity): Route = {
+    def listWithStatisticsAsync(): Future[Seq[LaunchQueue.QueuedInstanceInfoWithStatistics]] = async {
+      val info = await(launchQueue.listWithStatisticsAsync)
+      info.filter(t => t.inProgress && authorizer.isAuthorized(identity, ViewRunSpec, t.runSpec))
+    }
+
+    parameters('embed.*) { embed =>
+      val embedLastUnusedOffers = embed.exists(_ == QueueController.EmbedLastUnusedOffers)
```
meln1k: Can we use `embed.contains(EmbedLastUnusedOffers)` here?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
zen-dog: I did and I couldn't come up with any good name. So I left it a tuple

src/main/scala/mesosphere/marathon/api/akkahttp/V2Controller.scala
```diff
@@ -2,7 +2,7 @@ package mesosphere.marathon
 package api.akkahttp
 
 import akka.http.scaladsl.server.Route
-import v2.{ AppsController, EventsController, InfoController, PluginsController }
+import v2._
```
jeschkies: What's the benefit of this? Does it import some implicits? Otherwise I'd keep the explicit imports. This makes reading the code much easier.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
```diff
@@ -0,0 +1,84 @@
...
+
+class QueueController(
+    clock: Clock,
+    launchQueue: LaunchQueue)(
+    implicit
+    val actorSystem: ActorSystem,
+    val executionContext: ExecutionContext,
+    val authenticator: Authenticator,
+    val authorizer: Authorizer,
+    val electionService: ElectionService
```
jeschkies: I'm not sure we should pass the election service as an implicit. I've done it for the info endpoint and am not happy. What do you think?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
```diff
@@ -0,0 +1,84 @@
...
+          }
+        case None => reject(Rejections.EntityNotFound.queueApp(appId))
+      }
+    }
+  }
+}
+
+object QueueController {
+  val EmbedLastUnusedOffers = "lastUnusedOffers"
+}
```
jeschkies: I think a new line is missing here.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
```diff
@@ -0,0 +1,84 @@
...
+            complete(StatusCodes.NoContent)
+          }
+        case None => reject(Rejections.EntityNotFound.queueApp(appId))
+      }
+    }
+  }
+}
+
+object QueueController {
+  val EmbedLastUnusedOffers = "lastUnusedOffers"
```
jeschkies: What is this? Is it a parameter or a message?

EDIT: Just saw that it is a query parameter name. Maybe this could be expressed in the name.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
```diff
@@ -0,0 +1,84 @@
...
+  private def list(implicit identity: Identity): Route = {
+    def listWithStatisticsAsync(): Future[Seq[LaunchQueue.QueuedInstanceInfoWithStatistics]] = async {
+      val info = await(launchQueue.listWithStatisticsAsync)
+      info.filter(t => t.inProgress && authorizer.isAuthorized(identity, ViewRunSpec, t.runSpec))
+    }
+
+    parameters('embed.*) { embed =>
+      val embedLastUnusedOffers = embed.toSeq.contains(QueueController.EmbedLastUnusedOffers)
+      onSuccess(listWithStatisticsAsync) { info =>
+        complete((info, embedLastUnusedOffers, clock))
```
jeschkies: Hm, why not create a response model for this?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+  }
+
+  private def reset(appId: PathId)(implicit identity: Identity): Route = {
```
jeschkies: Why is it private? What is the benefit here?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
jeschkies: Is `Clock` the correct type? Why is this not a simple `playJsonMarshaller[raml.QueueResponse]`?

zen-dog: because `Seq[QueuedInstanceInfoWithStatistics]` is filtered with  `Boolean, Clock` to form the `Queue.raml` response: https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/raml/QueueInfoConversion.scala#L21 
Didn't want to change current implementation - we can do it later.

src/main/scala/mesosphere/marathon/api/akkahttp/V2Controller.scala
zen-dog: Intellij is set to to automatically convert >= 5 explicit imports to `_`. I don't see any particular benefit in changing this - long import lists aren't exactly handy. And look what I found in the leader controller PR:
https://github.com/mesosphere/marathon/pull/5579/files#diff-9dab674675063e0e5738657574bd846eR5
;)

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
```diff
@@ -12,10 +12,15 @@ object Rejections {
...
 
   val defaultEntityNotFoundMessage = Message("Entity was not found")
+
   case class EntityNotFound(message: Message = defaultEntityNotFoundMessage) extends Rejection
   object EntityNotFound {
     def app(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
     }
+
+    def queueApp(appId: PathId): EntityNotFound = {
```
alenkacz: did you mean queuedApp?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
zen-dog: I agree. Moved

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
zen-dog: I don't expect anybody else calling these. Why?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
zen-dog: See above

src/main/scala/mesosphere/marathon/api/akkahttp/V2Controller.scala
jeschkies: Damn it. Ok. If it's part of the style...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
jeschkies: Not a big thing but I like to keep methods public unless there is a good reason not to.

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
jeschkies: Oh man. You are right.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/QueueController.scala
meln1k: More public methods => longer compilation times :) private is a good default.

PR #5579, opened 2017-10-09
Create leader controller for Akka HTTP refactoring.

jeschkies:
Summary:
We reimplement the `LeaderResource` as an Akka HTTP `LeaderController`.
The response is modelled in RAML. New tests for leader abdication have
been added.

JIRA issues: MARATHON-7303

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
@@ -14,10 +14,10 @@ object Rejections {
   val defaultEntityNotFoundMessage = Message("Entity was not found")
   case class EntityNotFound(message: Message = defaultEntityNotFoundMessage) extends Rejection
   object EntityNotFound {
-    def app(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
+    def noApp(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
     }
-    def leader(): EntityNotFound = {
+    def noLeader(): EntityNotFound = {
       EntityNotFound(Message("There is no leader"))
     }
   }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
@@ -17,5 +17,8 @@ object Rejections {
     def app(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
     }
+    def leader(): EntityNotFound = {
+      EntityNotFound(Message("There is no leader"))
+    }
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
@@ -34,12 +34,10 @@ case class LeaderController(
     }
 
   def deleteLeader(): Route =
-    authenticated.apply { implicit identity =>
-      authorized(UpdateResource, AuthorizedResource.SystemConfig).apply {
-        parameters('backup.?, 'restore.?) { (backup, restore) =>
-          // This follows the LeaderResource implementation. Seems like a bug to me. Why should we not proxy the request?
-          if (electionService.isLeader) {
-
+    asLeader(electionService) {
+      authenticated.apply { implicit identity =>
+        authorized(UpdateResource, AuthorizedResource.SystemConfig).apply {
+          parameters('backup.?, 'restore.?) { (backup, restore) =>
             //TODO: validate backup and restore parameters
             complete {
               async {
@@ -48,8 +46,6 @@ case class LeaderController(
                 raml.Message("Leadership abdicated")
               }
             }
-          } else {
-            reject(Rejections.EntityNotFound.leader())
           }
         }
       }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
@@ -4,14 +4,21 @@ package api.akkahttp.v2
 import akka.http.scaladsl.server.Route
 import mesosphere.marathon.api.akkahttp.{ Controller, Rejections }
 import mesosphere.marathon.core.election.ElectionService
-import mesosphere.marathon.plugin.auth.{ Authenticator, AuthorizedResource, Authorizer, ViewResource }
+import mesosphere.marathon.plugin.auth._
 import com.typesafe.scalalogging.StrictLogging
+import mesosphere.marathon.raml.RuntimeConfiguration
+import mesosphere.marathon.storage.repository.RuntimeConfigurationRepository
+
+import scala.async.Async._
 import scala.concurrent.ExecutionContext
 
-case class LeaderController(val electionService: ElectionService)(
+case class LeaderController(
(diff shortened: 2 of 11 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
```diff
@@ -0,0 +1,70 @@
...
+          case Some(leader) => complete(raml.LeaderInfo(leader))
+        }
+      }
+    }
+
+  def deleteLeader(): Route =
+    authenticated.apply { implicit identity =>
+      authorized(UpdateResource, AuthorizedResource.SystemConfig).apply {
+        parameters('backup.?, 'restore.?) { (backup, restore) =>
+          // This follows the LeaderResource implementation. Seems like a bug to me. Why should we not proxy the request?
```
jeschkies: I'd say this should be an `asLeader`. What do you say @meln1k and @timcharper?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
```diff
@@ -0,0 +1,70 @@
...
+          if (electionService.isLeader) {
+
+            //TODO: validate backup and restore parameters
```
jeschkies: @meln1k, how should we do this validation?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
meln1k: I think there should be some kind of directive for that. But I'm not sure if it's merged yet.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
meln1k: There are 2 options: https://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/directives/misc-directives/validate.html and https://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/case-class-extraction.html#case-class-validation

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
meln1k: `asLeader` should work fine here. It will also proxy the request.

zen-dog: Yep. Proxying most of the requests should probably happen elsewhere. Otherwise the code below is only executed `asLeader`

meln1k: Can we place `asLeader` in the most upper level of the routes structure?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
zen-dog: I like the first one. Whatever you chose - by default akka http handles validation errors by throwing `ValidationRejection` and converting them to `HTTP 400 Bad Request` in default rejection handler. Marathon (at least for `v2/leader?backup=foo`) returns a `HTTP 422 Unprocessable Entity` with:
```
{
    "message": "Object is not valid",
    "details": [
        {
            "path": "/",
            "errors": [
                "Invalid URI or unsupported scheme: foo"
            ]
        }
    ]
}
```

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
jeschkies: Ok. It seems the original resource has a bug then.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
jeschkies: Hm, I thought we have validation with wix? Are we using the Akka HTTP validations for other models as well?

Personally I really do not like that we use that RAML model deep down in the storage layer.

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
```diff
@@ -17,5 +17,8 @@ object Rejections {
     def app(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
     }
+    def leader(): EntityNotFound = {
```
ichernetsky: Should this be renamed to `noLeader`? Also, maybe `noApp` too?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+
+  @SuppressWarnings(Array("all")) // async/await
+  def deleteLeader(): Route =
+    asLeader(electionService) {
+      authenticated.apply { implicit identity =>
+        authorized(UpdateResource, AuthorizedResource.SystemConfig).apply {
+          parameters('backup.?, 'restore.?) { (backup: Option[String], restore: Option[String]) =>
+            //TODO: validate backup and restore parameters
```
ichernetsky: Were they validated before? Is there a ticket to track this afterwards?

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
```diff
@@ -14,8 +14,11 @@ object Rejections {
   val defaultEntityNotFoundMessage = Message("Entity was not found")
   case class EntityNotFound(message: Message = defaultEntityNotFoundMessage) extends Rejection
   object EntityNotFound {
-    def app(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
+    def noApp(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
     }
+    def noLeader(): EntityNotFound = {
```
meln1k: Why not `def noLeader: EntityNotFound`? It seems there are no side effects :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
zen-dog: I am not sure if that's actually a bug. May be `DELETE /leader` was not supposed to be proxied for some reason?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/LeaderController.scala
```diff
@@ -0,0 +1,75 @@
...
+    implicit
+    val authenticator: Authenticator,
+    val authorizer: Authorizer,
+    val executionContext: ExecutionContext) extends Controller with Validation with StrictLogging {
+
+  import mesosphere.marathon.api.akkahttp.Directives._
+  import mesosphere.marathon.api.akkahttp.EntityMarshallers._
+
+  def leaderInfo(): Route =
+    authenticated.apply { implicit identity =>
```
meln1k: Can we move `authenticated.apply { implicit identity =>` between the lines 63 and 64?

jeschkies: Hm, `authorized` requires the identity but `authorized` cannot be moved since the resources diff (`UpdateResource` vs `ViewResource`).

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
jeschkies: I have to admit that I do not like method definitions without the braces much. I'd rather use val.

PR #5581, opened 2017-10-09
Enable successful SI Test runs with cluster of agents with different core counts

kensipe:
Enable successful SI Test runs with cluster of agents with different core counts

Summary:
The test was originally written assuming an agent core count of 4 which has been true for years.   Based on cost cutting measures the latest clusters are made of 2 core agents.   However CCM still creates the 4 core count agents.   This change allows this test to pass for changes in the agent core count.

JIRA issues:
??

### DIFF
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -643,11 +643,12 @@ def test_pinned_task_does_not_scale_to_unpinned_host():
     app_def = apps.sleep_app()
     app_id = app_def['id']
 
-    # the size of cpus is designed to be greater than 1/2 of a node
-    # such that only 1 task can land on the node. 
-    app_def['cpus'] = 1.5
     host = common.ip_other_than_mom()
     print('Constraint set to host: {}'.format(host))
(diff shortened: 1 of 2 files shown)


tests/system/marathon_common_tests.py
```diff
@@ -642,9 +642,13 @@ def test_pinned_task_does_not_scale_to_unpinned_host():
 
     app_def = apps.sleep_app()
     app_id = app_def['id']
-    app_def['cpus'] = 3.5
+
     host = common.ip_other_than_mom()
     print('Constraint set to host: {}'.format(host))
+    # the size of cpus is designed to be greater than 1/2 of a node
```
ichernetsky: What if a node has only 1 cpu? Should it be something like `max(0.6, cores - 0.5)` instead?

kensipe: it's a good catch... however if we start using 1 core vms I quit :)

ichernetsky:
Oops, I thought it's been in review for > 24h. Noticed that it was not the case only after I merged it. I beg your pardon, dear colleagues.

PR #5583, opened 2017-10-09
Refactor Openable trait

ichernetsky:
- Rename its open() and close() methods to markOpen() and markClosed()
  respectively, because the former are too generic.
- Rename Opeable to OpenableOnce because it better reflects what
  it does.


kensipe:
it isn't clear to me how `markOpen` is more informative than `open`.   It isn't a big deal either way to me. 
I agree that openOnce makes sense.   
I'm good with the change.

ichernetsky:
Thanks, @kensipe I believe `markOpen` is more informative because it is really about marking an object as open, and not doing any resource allocation. Another reason I did this is because while I was working on similar changes to the legacy persistent stores in `1.4` there was name collision because of different inheritance hierarchy.

PR #5584, opened 2017-10-09
Wire marathon scheduler manually, not via guice

alenkacz:
Summary:
Follow-up for my previous review - this time just removing MarathonScheduler being wired as a guice dependency - now it is instantiated manually. 

JIRA issues:

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -104,7 +104,7 @@ class CoreModuleImpl @Inject() (
     // infrastructure
     clock, random, marathonConf, actorSystem.scheduler,
     leadershipModule,
-    () => marathonScheduler
+    () => marathonScheduler.getHomeRegion
   )
 
   private[this] lazy val offerMatcherReconcilerModule =
@@ -150,7 +150,7 @@ class CoreModuleImpl @Inject() (
     // external guice dependencies
     taskTrackerModule.instanceTracker,
     launcherModule.taskOpFactory,
-    () => marathonScheduler
+    () => marathonScheduler.getHomeRegion
   )
 
   // PLUGINS
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -4,7 +4,7 @@ package core
 import java.time.Clock
 import javax.inject.Named
 
-import akka.actor.ActorSystem
+import akka.actor.{ ActorRef, ActorSystem }
 import akka.event.EventStream
 import com.google.inject.{ Inject, Provider }
 import mesosphere.marathon.core.async.ExecutionContexts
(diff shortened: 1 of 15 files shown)


src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
```diff
@@ -18,7 +18,7 @@ import scala.concurrent.{ Future, Promise }
   * @param actorRef Reference to actor that matches offers.
   * @param precedenceFor Defines which matcher receives offers first. See [[mesosphere.marathon.core.matcher.base.OfferMatcher.precedenceFor]].
   */
-class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], homeRegionProvider: Provider[HomeRegionProvider])
+class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], homeRegionProvider: () => HomeRegionProvider)
```
meln1k: Do you think it's possible to pass a function here? Something like `homeRegion: () => string`

alenkacz: right ... good point. I don't know how I would do it via guice but now it should be pretty straightforward

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -53,7 +55,8 @@ class CoreModuleImpl @Inject() (
   scheduler: Provider[DeploymentService],
   instanceUpdateSteps: Seq[InstanceChangeHandler],
   taskStatusUpdateProcessor: TaskStatusUpdateProcessor,
-  homeRegionProvider: Provider[HomeRegionProvider]
+  mesosLeaderInfo: MesosLeaderInfo,
+  @Named(ModuleNames.MESOS_HEARTBEAT_ACTOR) heartbeatActor: ActorRef
```
jeschkies: Why was `@Named(ModuleNames.MESOS_HEARTBEAT_ACTOR) heartbeatActor: ActorRef` added?

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -101,7 +104,7 @@ class CoreModuleImpl @Inject() (
     // infrastructure
     clock, random, marathonConf, actorSystem.scheduler,
     leadershipModule,
-    homeRegionProvider
+    () => marathonScheduler.getHomeRegion
```
jeschkies: What motivated this change?

alenkacz: @meln1k suggested it and I liked it... I would actually do it like this even from the beginning but then it was wired by guice and it was easier for me to introduce the interface.

You don't like it?

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
alenkacz: that is a dependency of MesosHeartbeatMonitor which I instantiated here instead of by guice. Also MarathonSchedulerService depends on it so I kept it managed by guice for now...

jeschkies: Ah, too much Guice magic for me ^^.

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
jeschkies: I was just a little confused and too tired yesterday.

PR #5587, opened 2017-10-09
Add IS constraint operator

timcharper:
JIRA Issues: MARATHON_EE-1667

### DIFF
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -102,6 +102,8 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
+Note, if the attribute in question is a scalar, it is rounded to the nearest thousandth using the half-even rounding strategy; zeroes after the decimal are dropped.
+
 ### UNLIKE operator
 **Value** (required): A regular expression for the value of the attribute.
 
@@ -185,7 +187,7 @@ text : [a-zA-Z0-9_/.-]
 
 #### Comparing text values
 
-When an `IS` constraint is specified, a task is only launched on nodes that have the specified attribute value. Note that `IS` behaves similar to `CLUSTER` with a value, and has been added mostly for semantic value (except its behavior differs when comparing scalars).
+When an `IS` constraint is specified, a task is only launched on nodes that have the specified value.
 
 ``` bash
 $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -10,9 +10,9 @@ Constraints control where apps run to allow optimizing for either fault toleranc
 
 ### Hostname as field name
 
-Entering `hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
+Entering `@hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
 
-All Marathon operators are supported when the field name is `hostname`. 
+All Marathon operators are supported when the field name is `@hostname`.
 
 ### Attribute as field name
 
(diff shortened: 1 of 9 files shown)


src/test/scala/mesosphere/mesos/ConstraintsTest.scala
```diff
@@ -645,6 +645,30 @@ class ConstraintsTest extends UnitTest {
...
+  "IS operator" should {
+    "require that a value match exactly" in {
+      makeOffer("righthost.com") should meetConstraint(hostnameField, IS, "righthost.com")
+      makeOffer("wronghost.com") should meetConstraint(hostnameField, IS, "wronghost.com")
+    }
+  }
+
+  "IN operator" should {
+    "require that the offer value be in the comma delimited list of values" in {
+      makeOffer("host1") should meetConstraint(hostnameField, IN, "host1,host2")
```
ichernetsky: I am worried that values might have commas in them. In this case, we would need to support escaping and/or strings inside value strings of constraint definitions. Although I would vote for supports of arrays of strings as constraint values. In this case the constraint in an app definition could look like `["some-field", "IN", ["value1", "value2"]`. What do you think?

alenkacz: I like what Ivan is proposing! :)

timcharper: This would break API compatibility

timcharper:
Great question :) On the one hand, it'd probably be a good idea to update the docs with this PR. Suzanne usually does our docs.

On the other hand, no sense in writing docs until we fully settle on the behavior of these constraints.

alenkacz:
@timcharper ok, thanks :)

then LGTM

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
ichernetsky: Please forgive me my ignorance, but I do not see how it would break the user-facing API compatibility. Isn't it about addition/extension only? Are you referring to the case when Java/Python/etc clients would fail if their receive an array value, and not a string one?

src/main/proto/marathon.proto
```diff
@@ -26,6 +26,10 @@ message Constraint {
       UNLIKE = 4;
       // Field will be grouped by field. Value specifies the maximum size of each group.
       MAX_PER = 5;
+      // Field must be the specified value
+      IS = 6;
```
jeschkies: Our API is also defined in Protos?!

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
```diff
@@ -645,6 +645,30 @@ class ConstraintsTest extends UnitTest {
...
+      makeOffer("host3") shouldNot meetConstraint(hostnameField, IN, "host1,host2")
+    }
+
+    "trims whitespace after the commas (but not before)" in {
+      makeOffer("host1") should meetConstraint(hostnameField, IN, "host1, host2")
+      makeOffer("host1") should meetConstraint(hostnameField, IN, "host1 , host2")
```
jeschkies: Isn't here a whitespace before the comma? `host1` is not equal to `host1[ ]`. What am I missing?

ichernetsky: The separator is a comma optionally surrounded by whitespaces.

src/main/proto/marathon.proto
timcharper: 🎉

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
timcharper: err lol the test description is a lie 😢

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
timcharper: It would change the type, specified here:

https://github.com/mesosphere/marathon/blob/v1.6.0-pre/docs/docs/rest-api/public/api/v2/types/constraint.raml#L20-L29

The specification is that constraints are a 2-3 item array of strings. Putting a non-string in there breaks the API.

ichernetsky: I was thinking of supporting strings _and_ arrays: arrays for `IN` and strings for everything else.

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
timcharper: I don't believe Mesos constraints what values can be specified on an agent attribute value. A user could potentially put trailing or leading spaces in the value. That would be an awful thing to do, but, none-the-less, possible.

I think it is more likely that people accidentally put a space in comma-delimited list.

ichernetsky: Yes, in agreement here.

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
timcharper: I understood this :)

timcharper:
Waiting for this proposal to be accepted by DCOS SDK / Storage Services team:

https://docs.google.com/document/d/e/2PACX-1vSFvPol0pcHC2Web7EaNU0oSDS5wrOWSgFcmuslYBtISV2NB2JZ_D-B4wpWy_Vutaf08m2LX6WZVy6s/pub

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
timcharper: @ichernetsky did you mean to write this earlier?

```
["some-field", "IN", "value1", "value2"]
```

(sans the additional `[`)?

@jdef mentioned this as a possibility and it made me wonder if perhaps this was what you were trying to suggest :)

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
timcharper: @vinodkone corrected me; spaces _are explicitly_ not allowed in attribute values, and [they are removed](https://github.com/apache/mesos/blob/master/src/common/values.cpp#L561) by Mesos.

src/test/scala/mesosphere/mesos/ConstraintsTest.scala
ichernetsky: No, I was suggesting `["field", "IN", ["value1", "value2"]]`.

timcharper:
Proposal is accepted and change has been updated to match proposal

docs/docs/constraints.md
```diff
@@ -102,6 +102,8 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
   }'
 ```
 
+Note, if the attribute in question is a scalar, it is rounded to the nearest thousandth using the half-even rounding strategy; zeroes after the decimal are dropped.
```
sascala: `**Note:** If...`

docs/docs/constraints.md
```diff
@@ -168,3 +170,44 @@ $ curl -X POST -H "Content-type: application/json" localhost:8080/v2/apps -d '{
...
+    "id": "sleep-cluster",
+    "cmd": "sleep 60",
+    "instances": 3,
+    "constraints": [["", "IS", "rack-1"]]
+  }'
+```
+
+#### Comparing scalar values
+
+When comparing scalars, the value is compared to the nearest thousandth (using half-even rounding strategy)
```
sascala: ^Period at the end of the sentence.

docs/docs/constraints.md
timcharper: ✅

docs/docs/constraints.md
timcharper: ✅

timcharper:
Thanks! Merged

PR #5589, opened 2017-10-10
[WIP] Run unit tests on dedicated host in Travis CI.

jeschkies:
Summary:
This is just a quick way to run unit tests and linting on Travis CI for
fast feedback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,15 @@
-sudo: false
+sudo: required
+dist: trusty
+
 language: scala
 jdk:
   - oraclejdk8
 scala:
   - 2.12.3
-cache:
-  directories:
-    - $HOME/.sbt
-    - $HOME/.ivy2
-script:
-  - sbt clean scapegoat doc coverage test
+
 before_script: # the automated download fails sometimes
     - mkdir -p $HOME/.sbt/launchers/0.13.11/
     - test -r $HOME/.sbt/launchers/0.13.11/sbt-launch.jar || curl -L -o $HOME/.sbt/launchers/0.13.11/sbt-launch.jar http://dl.bintray.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.11/sbt-launch.jar
-after_success:
-  - sbt coverageReport coveralls
-notifications:
-  slack:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,20 +1,15 @@
-sudo: false
+sudo: required
+dist: trusty
+
 language: scala
 jdk:
   - oraclejdk8
 scala:
   - 2.12.3
-cache:
```
alenkacz: why did we remove the cache?

.travis.yml
```diff
@@ -1,20 +1,15 @@
-sudo: false
+sudo: required
```
alenkacz: and why do we require sudo now?

.travis.yml
```diff
@@ -1,20 +1,15 @@
...
   - 2.12.3
-cache:
-  directories:
-    - $HOME/.sbt
-    - $HOME/.ivy2
-script:
-  - sbt clean scapegoat doc coverage test
+
 before_script: # the automated download fails sometimes
     - mkdir -p $HOME/.sbt/launchers/0.13.11/
```
ichernetsky: We use `0.13.13` in the master branch.

.travis.yml
jeschkies: This enables the VM instead of a container: https://docs.travis-ci.com/user/reference/overview/.

.travis.yml
jeschkies: We don't benefit from caching since the cache is downloaded from S3 (https://docs.travis-ci.com/user/caching#How-does-the-caching-work%3F). For sbt it might makes sense since it's rebuilt each time. I wanted to simplify things.

.travis.yml
jeschkies: Thanks for catching that.

.travis.yml
```diff
@@ -1,20 +1,16 @@
 sudo: false
```
jeschkies: Let's simplify things and go from there :).

jeschkies:
Travis cannot run the unit tests withing 60 minutes. See my branch: https://github.com/jeschkies/marathon/pull/2.

PR #5592, opened 2017-10-10
Simplify LeaderElection by using an event based approach

timcharper:
We encapsulate Curator election logic using a generic Akka Stream source of
events: ElectedLeader, Standby(currentLeader). Previously, we used to query
ZooKeeper each time an http endpoint was hit on the non leader. Now, we use
Zookeeper watches to be notified of leadership changes.

By using a generic stream interface, PsuedoElectionService becomes an
incredibly simple stream that simply emits a leader elected event when
it starts, and ends with a standby event when closed.

Testing also becomes easier with event production and event reacting
concerns better separated.

Further, we remove all uses of locks in ElectionService. These were
previously a source of deadlock, and a source of unseen complexity.


src/main/scala/mesosphere/marathon/core/CoreGuiceModule.scala
```diff
@@ -212,6 +213,9 @@ class CoreGuiceModule(config: Config) extends AbstractModule {
   @Singleton
   def provideExecutionContext: ExecutionContext = ExecutionContexts.global
 
+  @Provides @Singleton @Named(ModuleNames.ELECTION_EXECUTOR)
+  def electionExecutor(): ExecutorService = Executors.newSingleThreadExecutor()
```
jeschkies: Is there a way we could avoid Guive here?

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -53,7 +54,8 @@ class CoreModuleImpl @Inject() (
   scheduler: Provider[DeploymentService],
   instanceUpdateSteps: Seq[InstanceChangeHandler],
   taskStatusUpdateProcessor: TaskStatusUpdateProcessor,
-  homeRegionProvider: Provider[HomeRegionProvider]
+  homeRegionProvider: Provider[HomeRegionProvider],
+  @Named(ModuleNames.ELECTION_EXECUTOR) electionExecutor: ExecutorService,
```
jeschkies: Why not pass it?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+    * Materialized cancellable is used to abdicate leadership; which will do so followed by a closing of the stream.
+    */
+  def apply(
+    clientCloseable: LifeCycledCloseableLike[CuratorFramework],
+    zooKeeperLeaderPath: String,
+    zooKeeperConnectionTimeout: FiniteDuration,
+    hostPort: String,
+    singleThreadExecutor: ExecutorService): Source[LeadershipState, Cancellable] = {
+    val singleThreadEC = ExecutionContext.fromExecutor(singleThreadExecutor)
+    Source.queue[LeadershipState](16, OverflowStrategy.dropHead)
```
jeschkies: Could we guarantee that only one event at a time can be processed?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+  @SuppressWarnings(Array("CatchThrowable"))
+  private class CuratorEventEmitter(
+      singleThreadExecutor: ExecutorService,
+      clientCloseable: LifeCycledCloseableLike[CuratorFramework],
+      zooKeeperLeaderPath: String,
+      hostPort: String,
+      sq: SourceQueueWithComplete[LeadershipState]) extends Cancellable {
+
+    val client = clientCloseable.closeable
+    private lazy val singleThreadEC: ExecutionContext = ExecutionContext.fromExecutor(singleThreadExecutor)
```
jeschkies: Why is a new executor required? Because of the polling?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+    private lazy val leaderHostPortMetric: Timer = Metrics.timer(ServiceMetric, getClass, "current-leader-host-port")
+    private val curatorLeaderLatchPath = zooKeeperLeaderPath + "-curator"
+    private lazy val latch = new LeaderLatch(client, curatorLeaderLatchPath, hostPort)
+    @volatile private var isStarted = false
+    @volatile private var _isCancelled = false
+
+    /* It is important that we re-register our watch _BEFORE_ we get the current the current leader. In Zookeeper,
```
jeschkies: Why do we have to re-register at all?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * By re-registering the watch before querying the state, we will not miss out on the latest leader change.
+     */
+    def longPollLeaderChange(retries: Int = 0): Unit = Future {
+      try {
+        if (latch.getState == LeaderLatch.State.STARTED)
+          client.getChildren
+            .usingWatcher(new CuratorWatcher {
+              override def process(event: WatchedEvent): Unit =
+                if (!_isCancelled) longPollLeaderChange()
```
jeschkies: Why not publish the event directly?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+            })
+            .forPath(curatorLeaderLatchPath)
+        emitLeader()
+      } catch {
+        case ex: KeeperException.NoNodeException if retries < 100 =>
+          // Wait for node to be created
+          logger.info("retrying")
+          Thread.sleep(retries * 10L)
```
jeschkies: Hm, 🤔. I wonder if we should use Akka scheduling for polling instead. What do you think?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+          } catch {
+            case ex: Throwable =>
+              logger.error("Error while getting current leader", ex)
+              None
+          }
+        }
+      }
+
+      currentLeader match {
+        case Some(`hostPort`) =>
```
jeschkies: I don't quite get this. What is the type of `currentLeader`? What is `` `hostPort` ``? Just a name?

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,231 @@
...
+          sq.offer(LeadershipState.ElectedAsLeader)
+        case otherwise =>
+          sq.offer(LeadershipState.Standby(otherwise))
+      }
+    }
+
+    private val closeHook: () => Unit = cancel
+
+    def start(): Unit = synchronized {
+      assert(!isStarted, "already started")
```
jeschkies: Is the assertion the right approach? How about ignoring a second start?

src/main/scala/mesosphere/marathon/core/election/ElectionModule.scala
```diff
@@ -1,40 +1,42 @@
...
+) {
 
-  lazy val service: ElectionService = if (config.highlyAvailable()) {
+  lazy private val electionBackend: Source[LeadershipState, Cancellable] = if (config.highlyAvailable()) {
     config.leaderElectionBackend.get match {
       case Some("curator") =>
-        new CuratorElectionService(
-          config,
+        val client = new LifeCycledCloseable(CuratorElectionStream.newCuratorConnection(config))
+        sys.addShutdownHook { client.close() }
```
jeschkies: Don't we have to close the stream as well?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+class ElectionServiceImpl(
+    eventStream: EventStream,
+    hostPort: String,
+    leaderEventsSource: Source[LeadershipState, Cancellable],
+    crashStrategy: CrashStrategy,
+    electionEC: ExecutionContext
+)(implicit system: ActorSystem) extends ElectionService with StrictLogging {
+
+  import ElectionService._
+  @volatile private[this] var lastState: LeadershipState = LeadershipState.Standby(None)
```
jeschkies: What's the advantage of `@volatile` over and atomic reference?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+  /**
+    * Releases leadership
+    *
+    * Has no effect if called before offerLeadership
+    */
+  override def abdicateLeadership(): Unit = {
+    leaderSubscription.foreach(_.cancel())
+  }
+
+  private def initializeStream(leadershipTransitionsFlow: Flow[LeadershipState, LeadershipTransition, NotUsed]) = {
```
jeschkies: I think this method need some docs :).

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+    val localEventListenerSink = Sink.foreach[LeadershipState] { event =>
+      lastState = event
+    }
+    val localTransitionSink = Sink.foreach[LeadershipTransition] { e =>
+      eventStream.publish(e)
+      if (e == LeadershipTransition.Standby) {
```
jeschkies: Is this true? I thought it should be `if(lastState == LeadershipTransition.ElectedAsLeader && e == LeadershipTransition.Standby)`.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+        logger.error("Lost leadership; crashing")
+        crashStrategy.crash()
+      }
+    }
+
+    val (leaderStream, leaderStreamDone) =
+      RunnableGraph.fromGraph(GraphDSL.create(
+        leaderEventsSource, localEventListenerSink)(
+        (_, _)) { implicit b =>
```
jeschkies: What is `b`?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+          { (leaderEventsSource, localEventListenerSink) =>
+            import GraphDSL.Implicits._
+            // We defensively specify eagerCancel as true; if any of the components in the stream close or fail, then
+            // we'll help to make it obvious by closing the entire graph (and, by consequence, crashing).
+            // Akka will log all stream failures, by default.
+            val stateBroadcast = b.add(Broadcast[LeadershipState](2, eagerCancel = true))
```
jeschkies: What is a broadcast?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+    val localTransitionSink = Sink.foreach[LeadershipTransition] { e =>
+      eventStream.publish(e)
+      if (e == LeadershipTransition.Standby) {
+        logger.error("Lost leadership; crashing")
+        crashStrategy.crash()
+      }
+    }
+
+    val (leaderStream, leaderStreamDone) =
+      RunnableGraph.fromGraph(GraphDSL.create(
```
jeschkies: How about moving this graph into another builder method so we can test it?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+  def offerLeadership(candidate: ElectionCandidate): Unit = {
+
+    /**
+      * Deduped event stream with current leader removed. Specified this way to maintain compatibility with the rest of
+      * the code base
+      */
+    val leadershipTransitionsFlow =
+      Flow[LeadershipState]
+        .map {
+          case LeadershipState.ElectedAsLeader => LeadershipTransition.ElectedAsLeader
```
jeschkies: What is the difference between `LeadershipState` and `LeadershipTransition`. I think we should verify that it is an actual transition. What guarantees that we won't receive two `LeadershipState.ElectedAsLeader`?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+          case _: LeadershipState.Standby => LeadershipTransition.Standby
+        }
+        .via(EnrichedFlow.dedup(initialFilterElement = LeadershipTransition.Standby)) // if the first elements are standby, emit nothing
```
jeschkies: Why `dedup`?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,155 @@ trait ElectionCandidate {
...
+        .mapAsync(1) {
```
jeschkies: This ensures no parallel execution, right?

zen-dog:
I left a bunch of comments in the previous version of this PR here: https://github.com/mesosphere/marathon/pull/5562 They still apply, however we can have the discussion in either of the PRs

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
meln1k: it's a builder, akka-streams graphDSL thingie :)

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
meln1k: We can define executor in application.conf and then get it from actor system

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
meln1k: Another idea is to wrap all this in actor and use `BackoffSupervisor` to handle restarts.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: It is how we managed the curator connection logic in the old election logic.

It doesn't seem necessary anymore, although, putting the IO logic in its own thread isn't the worst idea. Election logic is important enough that I think it is okay to have its own thread.

We definitely don't need to pass in the java ExecutorService anymore, since we no longer pass the executor to curator.

✅

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: This is the idea with streams

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: It's actually very important that query we query the state _after_ we register the watch, and not after the watch fires.

If we query the state, and then register the watch, there is a small window in which we could miss a leader change event.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: This was my earlier implementation.

Akka scheduling means we would have to query periodically. Seems suboptimal if no leader change happened. The idea of using watches is that we just have to query when something was detected to have changed.

I'm personally not enthusiastic about introducing actors here. They can lead to an explosion of complexity and can be more difficult to follow. Not saying that this is the easiest code to follow, but, it doesn't seem awful, either. I wrote it, though, so I'm biased.

I think using BackoffSupervisor would be too heavy handed. We have 19 lines of code. Once we introduce an actor, and backoff supervisor, etc., it'll become 50-100. I don't think that helps legibility.

I empathize with the hesitation. Excessive confidence in concurrency code is the mark of an inexperienced programmer. I did stress test this pretty heavily to verify it has consistent behavior. While this was only against a single Zookeeper node, I do feel the test was sufficient to bring out some bugs (it is actually why I introduced the backoff).

Might be a good idea to make it easy to run the stress test again. With an environment variable? I ran the detect hand-off test 10,000 times (the test `"Monitors leadership changes"`). It behaved completely deterministically.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: `currentLeader` is the value we queried from ZooKeeper. `hostPort` is passed to the constructor above.

I'll remove the back ticks and replace it with an if statement to be more clear. Kind of a fringe Scala feature.

timcharper: (this code is equivalent)

```
      currentLeader match {
        case Some(leader) if leader == hostPort =>
          sq.offer(LeadershipState.ElectedAsLeader)
        case otherwise =>
          sq.offer(LeadershipState.Standby(otherwise))
      }
```

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: This is a private class. Calling start twice could be an indication of some upstream inconsistency / error that I would prefer be surfaced.

src/main/scala/mesosphere/marathon/core/election/ElectionModule.scala
timcharper: stream is closed implicitly by the latch closing

timcharper: (watch events fire on connection loss)

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: AtomicReference would work.

AtomicReference creates one more box, since it is generic.

I generally use AtomicReference when I need to apply atomic operations, such as compare and swap. That isn't needed here. `@volatile` var suffices.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: But it's private! :)

Okay.

✅ 

This work?

```

  /**
    * Construct the stream topology and run it.
    *
    * Topology looks like this:
    *
    * +----------------------------+    +-----------+    +----------------------------+
    * |    leaderEventsSource      | -> | broadcast | -> |   localEventListenerSink   |
    * | (IE CuratorElectionStream) |    +-----------+    | (keep track of last state) |
    * |                            |          |          +----------------------------+
    * |     [LeadershipState]      |          |
    * +----------------------------+          |
    *                                         V
    *                             +--------------------------------------+
    *                             |      leadershipTransitionFlow        |
    *                             | (Emits LeadershipTransition when we  |
    *                             | lost or gain leadership, but calling |
    *                             | marathon initialization logic before |
    *                             | emitting ObtainedLeadership          |
    *                             |                                      |
    *                             |      [LeadershipTransition]          |
    *                             +--------------------------------------+
    *                                              |
    *                                              V
    *          +------------------------+    +-----------+    +------------------------+
    *          |   localTransitionSink  | <- | broadcast | -> |    metricsSink         |
    *          | Monitors for lost      |    +-----------+    | Watches transitions,   |
    *          | leadership; crashes if |                     | records how long we've |
    *          | that happens.          |                     | had leadership         |
    *          +------------------------+                     +------------------------+
    *
    * If any component throws an exception, or finishes, then the _whole stream_ shuts down.
    *
    * When the stream ends (exception or not), we report accordingly and crash.
    */
```

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: `LeadershipTransition.Standby` is only published, by definition, if leadership was previously had.

Enriched documentation for these events to make more clear.

There is already a test for this.

✅

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: Graph is already covered by tests. I don't feel like more is necessary.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: https://doc.akka.io/docs/akka/current/scala/stream/stream-graphs.html#constructing-graphs :)

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: https://doc.akka.io/docs/akka/current/scala/stream/stream-graphs.html#constructing-graphs

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: If the first events are Standby, then they are removed from the stream (`initialFilterElement`).

If leadership hands off from one instance to another (both external to this instance), then while we get multiple upstream events, we only want to emit a single transition event, because our own leadership status did not change.

(in below table, empty cell indicates no event emitted; each row indicates an event emitted in time)

| Sequence | LeadershipState  | LeadershipTransition (pre-dedup) | LeadershipTransition (after dedup) |
| -------- | ---------------- | -------------------------------- | ---------------------------------- |
| 1        | Standby(None)    | Standby                          |                                    |
| 2        | Standby(host1)   | Standby                          |                                    |
| 3        | ElectedAsLeader  | ElectedAsLeader                  | ElectedAsLeader                    |
| 4        | Standby(None)    | Standby                          | Standby                            |
| 5        | Standby(host3)   | Standby                          |                                    |

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: Yes, it does.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Stated differently, currently the order of events are as follows:

1) We register a watch
2) We query the current leader
3) We receive an event (indicating child removal / addition)
4) Repeat step 1

If we published the event directly, then the order of events would look like this:

1) We register a watch
2) We receive an event (indicating child removal / addition)
3) We query the current leader
4) Repeat step 1

Between step 3 and 1 in the re-arranged sequence is the window of opportunity to miss an event.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
timcharper: > What guarantees that we won't receive two LeadershipState.ElectedAsLeader

dedup

I've augmented the docs for LeadershipState and LeadershipTransition to make this more clear.

✅

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,243 @@
...
+          sq.offer(LeadershipState.ElectedAsLeader)
+        case otherwise =>
+          sq.offer(LeadershipState.Standby(otherwise))
+      }
+    }
+
+    private val closeHook: () => Unit = { () => cancel() }
+
+    def start(): Unit = synchronized {
+      assert(!isStarted, "already started")
```
meln1k: Should we use `require` here?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
```diff
@@ -91,10 +105,237 @@ trait ElectionCandidate {
...
+    *          |   localTransitionSink  | <- | broadcast | -> |    metricsSink         |
+    *          | Monitors for lost      |    +-----------+    | Watches transitions,   |
+    *          | leadership; crashes if |                     | records how long we've |
+    *          | that happens.          |                     | had leadership         |
+    *          +------------------------+                     +------------------------+
+    *
+    * If any component throws an exception, or finishes, then the _whole stream_ shuts down.
+    *
+    * When the stream ends (exception or not), we report accordingly and crash.
+    */
```
meln1k: This picture is so cool!

timcharper: Thanks!

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Any specific reason?

`require` throws an `IllegalArgumentException`, and `assert` throws an `AssertionError`

meln1k: you can disable `assert` in runtime but not `require`. Just wanted to clarify if we need this.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,243 @@
...
+    private lazy val latch = new LeaderLatch(client, curatorLeaderLatchPath, hostPort)
+    private var isStarted = false
+    @volatile private var _isCancelled = false
+
+    /* Long-poll trampoline-style recursive method which calls emitLeader() each time it detects that the leadership
+     * state has changed.
+     *
+     * Given instance A, B, C, Curator's Leader latch recipe only provides A the ability to be notified if it gains or
+     * loses leadership, but not if leadership transitions between B and C. This method allows us to monitor any change
+     * in leadership state.
```
zen-dog: 👏  thx for the comment!

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+      * Emit current leader. Does not fail on connection error, but throws if multiple election candidates have the same
+      * ID.
+      */
+    private def emitLeader(): Unit = {
+      val currentLeader = leaderHostPortMetric.blocking {
+        val participants = try {
+          if (client.getState == CuratorFrameworkState.STOPPED)
+            Nil
```
zen-dog: Because `Nil` list will give an empty iterator? 🤔  Interesting

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -0,0 +1,243 @@
...
+          else
+            latch.getParticipants.asScala.toList
+        } catch {
+          case ex: Throwable =>
+            logger.error("Error while getting current leader", ex)
+            Nil
+        }
+        if (participants.iterator.filter(_.getId == hostPort).size > 1)
+          throw new IllegalStateException(s"Multiple election participants have the same ID: ${hostPort}. This is not allowed.")
```
zen-dog: I guess this should normally never happen?

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,40 @@
...
-case class ProcessOutputToLogStream(process: String) extends ProcessLogger {
-  val log = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
-  override def out(msg: => String): Unit = log.debug(msg)
-  override def err(msg: => String): Unit = log.warn(msg)
+final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, periodMillis: Long = 1000L) extends ProcessLogger {
+  private[this] var nextFlush = System.currentTimeMillis() + periodMillis
+  private[this] var callsSinceLastFlush = 0
+  val log: Logger = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
+
+  private def rateLimited(msg: String): Boolean = synchronized {
```
zen-dog: I guess this in now in it's own PR?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
zen-dog: 👏 👏

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
timcharper: Yeah, I'll remove this piece.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: If you start two marathon instances with the same hostname, and the same port, then it would happen.

This would also be illegal.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Yeah, we presume no known participants :) This is translated to `Standby(None)` (we're not leader, and we don't know who is)

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Okay, that's a good reason to use `require`. I have never, _not once_, wanted to exclude assertions. I didn't even know that was possible :)

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
timcharper: Curator's leader latch actually allows for this to happen. The trouble is, when querying the participant list, you can't tell which is your participant record, and which is some other participant record.

Just forbidding this state altogether kept it simple and allowed us to use a query mechanism to determine if we're the leader, or somebody else is.

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
jeschkies: I'm on your side. We do not need to introduce more akka dependencies here.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
jeschkies: `@volatile` is fine 🙂 .

src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
jeschkies: Jeez. That is indeed some fringe feature 😉. Thanks for making the code more expressive.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
jeschkies: 😮 
Karsten: I think we need some docs.
Tim: How many pages.

Overall, I like to document code also for myself and other engineers. Ideally methods are so expressive that they need no docs. However, for complex systems it's super hard to achieve this.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
jeschkies: Thanks for clearing this up. I did not know that the graph would ensure this.

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
jeschkies: `leaderStream` is unit tested?

src/main/scala/mesosphere/marathon/core/election/ElectionService.scala
jeschkies: Thanks for clearing this up.

jeschkies:
@timcharper, I took the liberty and resolved the merge conflicts. The loop is running [here](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-tim-election/).

src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
```diff
@@ -125,8 +125,11 @@ class EventsControllerTest extends AkkaUnitTest with Inside {
       .runWith(Sink.queue())
 
     f.leaderStateEventsInput.offer(LeadershipTransition.Standby)
+    inside(events.pull().futureValue) {
+      case Some(Some(tpe)) => tpe shouldBe "event_stream_attached"
```
jeschkies: The tests fails here. This was added with https://github.com/mesosphere/marathon/commit/fc863df666242fcd95d69c2d0af62d59ff2db092. @timcharper, is this still relevant?

jeschkies:
The loop shows a few time outs. E.g. [marathon-loop-tim-election/137](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-tim-election/137/) did not finish within 30 minutes.

zen-dog:
It has also a few error in `CuratorElectionStreamTest` e.g.:
https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-tim-election/193/

jeschkies:
The loop had a first green build again.

jeschkies:
failures

Failures time line of loop.

PR #5593, opened 2017-10-11
Handle validation results in directive.

jeschkies:
Summary:
In some cases we have to validate queyr parameters and handled the
result. This change introduces a simple directive that rejects requests
with a failed validation.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
@@ -3,7 +3,8 @@ package api.akkahttp
 
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.model.{ DateTime, HttpHeader, HttpMethods, HttpProtocols }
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -62,4 +63,17 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
       }
     }
   }
+
+  /**
+    * Rejects the request if the validation result is a failure. Proceeds otherwise.
+    * @param result The result of a Wix validation.
+    * @return The passed inner route.
+    */
+  def assumeValid(result: ValidationResult): Directive0 = Directive { f: (Unit => Route) =>
```
meln1k: `f: () => Route`?

meln1k: Why do we need `f: (Unit => Route) =>` ? Can't we write
```scala
result match {
       case failure: Failure => reject(ValidationFailed(failure))
       case Success => pass
     }
```

PR #5522, opened 2017-09-22
Implementation of AppsController

meln1k:
Summary: AppsController is implemented with akka-http

Test Plan: N/A

Reviewers: unterstein, zen-dog, ichernetsky, timcharper, kensipe

Subscribers: jenkins, marathon-team

JIRA: MARATHON-7299

PR #5594, opened 2017-10-11
AppsController split: replace multiple apps method

meln1k:
Summary: Migration of apps resource to akka http. This PR only implements replaceMultipleApps method, so it should be easier to review it. Full implementation of AppsController is available at #5522.

JIRA issues: MARATHON-7299

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
@@ -197,6 +197,6 @@ object EntityMarshallers {
       complete(StatusCodes.UnprocessableEntity -> failure)
   }
 
-  def entityUnmarshallerToMessageUnmarshaller[T](um: FromEntityUnmarshaller[T]): FromMessageUnmarshaller[T] =
+  implicit def entityUnmarshallerToMessageUnmarshaller[T](um: FromEntityUnmarshaller[T]): FromMessageUnmarshaller[T] =
     Unmarshaller.withMaterializer { implicit ec ⇒ implicit mat ⇒ request ⇒ um(request.entity) }
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -158,6 +158,7 @@ class AppsController(
   }
 
   private def patchSingle(appId: PathId)(implicit identity: Identity): Route = ???
+
   private[this] def updateMultiple(partialUpdate: Boolean, allowCreation: Boolean)(implicit identity: Identity): Route = {
     val version = clock.now()
     (forceParameter & entity(as(appUpdatesUnmarshaller(partialUpdate)))) { (force, appUpdates) =>
@@ -204,7 +205,6 @@ class AppsController(
     case Failure(ex) =>
       throw ex
   }
-
 
   private def putSingle(appId: PathId)(implicit identity: Identity): Route = ???
 
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -197,7 +197,7 @@ class AppsController(
     case Failure(_: AppNotFoundException) =>
       reject(
         maybeAppId.map { appId =>
-          Rejections.EntityNotFound.app(appId)
+          Rejections.EntityNotFound.noApp(appId)
         } getOrElse Rejections.EntityNotFound()
       )
     case Failure(RejectionError(rejection)) =>
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -165,9 +165,9 @@ class AppsController(
       def updateGroup(rootGroup: RootGroup): RootGroup = appUpdates.foldLeft(rootGroup) { (group, update) =>
         update.id.map(PathId(_)) match {
           case Some(id) =>
-            val res = Try(group.updateApp(id, AppHelpers.updateOrCreate(id, _, update, partialUpdate, allowCreation, clock.now(), marathonSchedulerService), version))
-            res.get
-          case None => group
+            group.updateApp(id, AppHelpers.updateOrCreate(id, _, update, partialUpdate, allowCreation, clock.now(), marathonSchedulerService), version)
+          case None =>
+            group
         }
       }
 
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -8,6 +8,7 @@ import akka.event.EventStream
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.server.{ Directive1, Rejection, RejectionError, Route }
 import mesosphere.marathon.api.v2.{ AppHelpers, AppNormalization, InfoEmbedResolver, LabelSelectorParsers }
 import mesosphere.marathon.core.appinfo._
 import mesosphere.marathon.core.deployment.DeploymentPlan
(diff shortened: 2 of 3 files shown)


src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1852,5 +1852,62 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+          |   "id": "$app1Id",
+          |   "cmd": "$newApp1Cmd"
+          | },
+          | {
+          |   "id": "$app2Id",
+          |   "cmd": "$newApp2Cmd"
+          | }
+          ]""".stripMargin.getBytes("UTF-8")
+
+      When("The application is updated")
```
alenkacz: I see this line with the same text on line 1891 (maaan, that is a long file :)

So maybe you did not meant to have it two times there?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -136,6 +159,53 @@ class AppsController(
...
+        complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+      }
+    }
+  }
+
+  /**
+    * It'd be neat if we didn't need this. Would take some heavy-ish refactoring to get all of the update functions to
+    * take an either.
+    */
+  private def onSuccessLegacy[T](maybeAppId: Option[PathId])(f: => Future[T])(implicit identity: Identity): Directive1[T] = onComplete({
```
alenkacz: I would really like to refactor this somehow, but right now I did not come to any good idea to do this - maybe someone else will jump in :)

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1852,5 +1852,62 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+      val entity = HttpEntity(body).withContentType(ContentTypes.`application/json`)
+      Put(Uri./, entity) ~> route ~> check {
+        Then("The application is updated")
+        status shouldEqual StatusCodes.OK
+        header[Headers.`Marathon-Deployment-Id`] should not be 'empty
+      }
+    }
```
zen-dog: I'd like to see another test case covering `AppNotFoundException`.

meln1k: Another test won't make much sense because this endpoint will always allow creation

kensipe:
also please add the PR to the jiras so when the jira closes we can see all that was completed.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: Yep, you're right.

meln1k:
@kensipe thanks for the feedback. Regarding JIRA, it should automatically link all PRs to the ticket.

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -199,6 +199,6 @@ object EntityMarshallers {
       complete(StatusCodes.UnprocessableEntity -> failure)
   }
 
-  def entityUnmarshallerToMessageUnmarshaller[T](um: FromEntityUnmarshaller[T]): FromMessageUnmarshaller[T] =
+  implicit def entityUnmarshallerToMessageUnmarshaller[T](um: FromEntityUnmarshaller[T]): FromMessageUnmarshaller[T] =
```
jeschkies: What is this method doing? `implicit ec ⇒ implicit mat ⇒ request ⇒ um(request.entity)` seems non trivial to me. Adding the extra `implicit` does not help 😉.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -82,7 +101,11 @@ class AppsController(
     Selector.forall(Seq(authzSelector, fn))
   }
 
-  private def replaceMultipleApps(implicit identity: Identity): Route = ???
+  private def replaceMultipleApps(implicit identity: Identity): Route = {
```
jeschkies: I thinks it's time to introduce scala docs.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -136,6 +159,53 @@ class AppsController(
 
   private def patchSingle(appId: PathId)(implicit identity: Identity): Route = ???
 
+  private[this] def updateMultiple(partialUpdate: Boolean, allowCreation: Boolean)(implicit identity: Identity): Route = {
```
jeschkies: As we've discussed the identity probably should not be passed implicitly but rather with the request object. Aside from that this seems to be a helper method so why not pass the identity explicitly?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -136,6 +159,53 @@ class AppsController(
...
+    val version = clock.now()
+    (forceParameter & entity(as(appUpdatesUnmarshaller(partialUpdate)))) { (force, appUpdates) =>
```
jeschkies: What is `(forceParameter & entity(as(appUpdatesUnmarshaller(partialUpdate))))`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -136,6 +159,53 @@ class AppsController(
...
+      def updateGroup(rootGroup: RootGroup): RootGroup = appUpdates.foldLeft(rootGroup) { (group, update) =>
```
jeschkies: How about moving this method out and test it separately? I actually thought that we mix a lost of controller and service/model logic. Ideally the controllers call one method or so. I haven't been following myself, though.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
jeschkies: Why is this method required at all?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -136,6 +159,53 @@ class AppsController(
...
+  private def onSuccessLegacy[T](maybeAppId: Option[PathId])(f: => Future[T])(implicit identity: Identity): Directive1[T] = onComplete({
+    try { f }
+    catch {
+      case NonFatal(ex) =>
+        Future.failed(ex)
+    }
+  }).flatMap {
+    case Success(t) =>
+      provide(t)
+    case Failure(ValidationFailedException(_, failure)) =>
```
jeschkies: I think the failures should be handled by the rejection handlers. Why not just throw them?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -136,6 +159,53 @@ class AppsController(
...
+      reject(EntityMarshallers.ValidationFailed(failure))
+    case Failure(AccessDeniedException(msg)) =>
```
jeschkies: Where is this coming from?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1852,5 +1852,61 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+
+      val entity = HttpEntity(body).withContentType(ContentTypes.`application/json`)
+      Put(Uri./, entity) ~> route ~> check {
+        Then("The application is updated")
+        status shouldEqual StatusCodes.OK
+        header[Headers.`Marathon-Deployment-Id`] should not be 'empty
+      }
+    }
   }
 }
```
jeschkies: Is a new line missing? 🙂

kensipe:
@meln1k it doesn't link afaik... so please add them and if this is a feature we want let create a jira to make sure it gets done.

PR #5597, opened 2017-10-11
[RFC] Use explicit RAML conversion instead of implicits.

jeschkies:
Summary:
This leverages method overloading instead of implicits. This way it is
much easier to find the implementation of a conversion and their usages.
It should also speed up compilation since searching for implicits takes
the longest time.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
@@ -17,12 +17,13 @@ import mesosphere.marathon.core.appinfo.AppInfo
 import mesosphere.marathon.plugin.PathId
 import mesosphere.marathon.core.launchqueue.LaunchQueue.QueuedInstanceInfoWithStatistics
 import mesosphere.marathon.core.plugin.PluginDefinitions
+import mesosphere.marathon.raml.AppConversion
 import mesosphere.marathon.state.AppDefinition
 import play.api.libs.json._
 
 import scala.collection.breakOut
 
-object EntityMarshallers {
+object EntityMarshallers extends AppConversion {
   import Directives.complete
   import mesosphere.marathon.api.v2.json.Formats._
   import mesosphere.marathon.raml.MetricsConversion._
@@ -130,7 +131,7 @@ object EntityMarshallers {
         // however since this is an update, the user isn't required to specify an ID as part of the definition so we do
         // some hackery here to pass initial JSON parsing.
         val app = (jsObj + ("id" -> JsString(appId.toString))).as[raml.App]
-        appNormalization.normalized(app).toRaml[raml.AppUpdate]
+        asRaml(appNormalization.normalized(app))
       }
   }.handleValidationErrors
 
@@ -144,7 +145,7 @@ object EntityMarshallers {
       // this is a complete replacement of the app as we know it, so parse and normalize as if we're dealing
       // with a brand new app because the rules are different (for example, many fields are non-optional with brand-new apps).
       // the version is thrown away in toUpdate so just pass `zero` for now.
-      playJsonUnmarshaller[Seq[raml.App]].map(_.map(app => appNormalization.normalized(app).toRaml[raml.AppUpdate]))
+      playJsonUnmarshaller[Seq[raml.App]].map(_.map(app => asRaml(appNormalization.normalized(app))))
   }
 
   implicit val jsValueMarshaller = playJsonMarshaller[JsValue]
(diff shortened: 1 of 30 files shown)


jeschkies:
It is not complete yet.

src/main/scala/mesosphere/marathon/raml/VolumeConversion.scala
```diff
@@ -45,7 +45,7 @@ trait VolumeConversion extends ConstraintConversion with DefaultConversions {
         case DiskType.Path => PersistentVolumeType.Path
         case DiskType.Root => PersistentVolumeType.Root
       })
-      PersistentVolume(pvType, pv.size, pv.maxSize, pv.constraints.toRaml[Set[Seq[String]]])
+      PersistentVolume(pvType, pv.size, pv.maxSize, pv.constraints.map(constraintProtosToStringSeq))
```
timcharper: This change should definitely stay

timcharper:
I'm conflicted about how I feel about this. If we were to implement RAML conversion in Rust, it would fromRaml and toRAML would be implemented in a RAMLConversion trait, and then each implementor would define those methods. The benefit is that we get uniformity in definition, and ad-hoc polymorphism. If a method accepts any T for which a RAMLConversion is implemented, then we can simply provide it said T.

So, effectively, I would stand in support for RAMLConversion in the account that it is implemented and treated as a type-class, and the majority of definitions are defined in the companion object for the respective type (this makes for a simple lookup rule for finding implicit definitions).

If a toRaml / fromRAML type needs additional parameters, because normalization is involved, then I'd be inclined to extract that normalization out of the RAML conversion layer (it probably does not belong there). This would make our use of implicits more straightforward, if it is the case that they are currently used this way.

I don't think the implicit class methods are necessary. `.toRaml` etc.

WRT to compile speed, have we run a clean compile benchmark to see if this actually improves things?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -17,12 +17,13 @@ import mesosphere.marathon.core.appinfo.AppInfo
...
 import mesosphere.marathon.core.launchqueue.LaunchQueue.QueuedInstanceInfoWithStatistics
 import mesosphere.marathon.core.plugin.PluginDefinitions
+import mesosphere.marathon.raml.AppConversion
 import mesosphere.marathon.state.AppDefinition
 import play.api.libs.json._
 
 import scala.collection.breakOut
 
-object EntityMarshallers {
+object EntityMarshallers extends AppConversion {
```
meln1k: Why extend when we can just import a method?

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -52,13 +50,13 @@ class AppTasksResource @Inject() (
           val groupPath = gid.toRootPath
           val maybeGroup = groupManager.group(groupPath)
           withAuthorization(ViewGroup, maybeGroup, unknownGroup(groupPath)) { group =>
-            ok(jsonObjString("tasks" -> runningTasks(group.transitiveAppIds, instancesBySpec).toRaml))
+            ok(jsonObjString("tasks" -> runningTasks(group.transitiveAppIds, instancesBySpec).map(asRaml)))
```
meln1k: Is it the same as before?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -445,7 +445,7 @@ trait AppAndGroupFormats {
         info.maybeCounts.map(TaskCountsWrites.writes(_).as[JsObject]),
         info.maybeDeployments.map(deployments => Json.obj("deployments" -> deployments)),
         info.maybeReadinessCheckResults.map(readiness => Json.obj("readinessCheckResults" -> readiness)),
-        info.maybeTasks.map(tasks => Json.obj("tasks" -> Raml.toRaml(tasks))),
+        info.maybeTasks.map(tasks => Json.obj("tasks" -> tasks.map(asRaml))),
```
meln1k: I'm not sure if it's equivalent

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -40,7 +37,7 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       backoffFactor = app.backoffStrategy.factor,
       backoffSeconds = app.backoffStrategy.backoff.toSeconds.toInt,
       cmd = app.cmd,
-      constraints = app.constraints.toRaml[Set[Seq[String]]],
+      constraints = app.constraints.map(constraintProtosToStringSeq),
```
meln1k: Is it the same as it was before?

jeschkies:
> If a toRaml / fromRAML type needs additional parameters, because normalization is involved, then I'd be inclined to extract that normalization out of the RAML conversion layer (it probably does not belong there). This would make our use of implicits more straightforward, if it is the case that they are currently used this way.

+1 on that. Let's revisit this later.

PR #5598, opened 2017-10-11
Move method to GroupManager

alenkacz:
Summary:
Before actually working on the TasksController I moved part of the code to GroupManager as it was already suggested in the comment.

JIRA issues: MARATHON-7308

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/v2/TasksResourceTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/v2/TasksResourceTest.scala
@@ -77,7 +77,7 @@ class TasksResourceTest extends UnitTest with GroupCreation {
       Given("no apps")
       config.zkTimeoutDuration returns 5.seconds
       taskTracker.instancesBySpec returns Future.successful(InstanceTracker.InstancesBySpec.empty)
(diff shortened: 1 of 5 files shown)


src/test/scala/mesosphere/marathon/api/v2/TasksResourceTest.scala
```diff
@@ -77,7 +77,7 @@ class TasksResourceTest extends UnitTest with GroupCreation {
       Given("no apps")
       config.zkTimeoutDuration returns 5.seconds
       taskTracker.instancesBySpec returns Future.successful(InstanceTracker.InstancesBySpec.empty)
-      groupManager.rootGroup() returns createRootGroup()
```
ichernetsky: Why is this line deleted? Shouldn't the following one be just added?

meln1k: I have the same question.

alenkacz: indexJson method does not call rootGroup(), just indexTxt does so I think it was a dead line of code even before I started touching this so I just removed it...

PR #5600, opened 2017-10-11
Update Mesos version to 1.4.0 which is used for building packages and Docker images

ichernetsky:


ichernetsky:
@timcharper it is a nice idea, but unfortunately, it won't always work for us, say, when we use RC builds. For instance, if `Mesos` is `1.4.0-rc5`, `MesosDebian` could be `1.4.0-1.0.1.rc5`.

PR #5602, opened 2017-10-12
AppsController split: `replace app` method

meln1k:
This is another part of migration to akka-http

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -6,6 +6,8 @@ import java.time.Clock
 
 import akka.event.EventStream
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.{ StatusCodes, Uri }
+import akka.http.scaladsl.model.headers.Location
 import akka.http.scaladsl.model.StatusCodes
 import akka.http.scaladsl.server.Route
 import akka.http.scaladsl.server.{ Directive1, Rejection, RejectionError, Route }
@@ -14,6 +16,7 @@ import mesosphere.marathon.api.akkahttp.{ Controller, EntityMarshallers }
 import mesosphere.marathon.api.v2.AppHelpers.{ appNormalization, appUpdateNormalization, authzSelector }
 import mesosphere.marathon.api.v2.Validation.validateOrThrow
 import mesosphere.marathon.api.v2.AppHelpers.{ appNormalization, appUpdateNormalization, authzSelector }
+
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.appinfo._
 import mesosphere.marathon.core.deployment.DeploymentPlan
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -222,6 +259,14 @@ class AppsController(
 
   private def getVersion(appId: PathId, version: Timestamp)(implicit identity: Identity): Route = ???
 
+  private def completeWithDeploymentForApp(appId: PathId, plan: DeploymentPlan) =
```
jeschkies: Could you add a comment so refactoring would be easier? Is this used somewhere else?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -206,7 +209,41 @@ class AppsController(
...
+    * @return http servlet response
+    */
+  private[this] def update(appId: PathId, partialUpdate: Boolean, allowCreation: Boolean)(implicit identity: Identity): Route = {
+    val version = clock.now()
+
+    (forceParameter &
+      extractClientIP &
+      extractUri &
+      entity(as(appUpdateUnmarshaller(appId, partialUpdate)))) { (force, remoteAddr, requestUri, appUpdate) =>
+        //         Note - this function throws exceptions and handles authorization synchronously. We need to catch and map these
```
jeschkies: Nit: What is the reason for the space in the comment?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -206,7 +209,41 @@ class AppsController(
...
+        //         exceptions to the appropriate rejections
+        val fn = updateOrCreate(
+          appId, _: Option[AppDefinition], appUpdate, partialUpdate, allowCreation, clock.now(), marathonSchedulerService)
```
jeschkies: Hm, this is an interesting pattern. Why not create a proper method with `def`? What is the motivation for using the underscore?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -206,7 +209,41 @@ class AppsController(
...
+
+        onSuccessLegacy(Some(appId))(groupManager.updateApp(appId, fn, version, force)).apply { plan =>
```
jeschkies: It's really hard to follow all the nested methods :/. It feels it would make more sense to factor out `groupManager.updateApp(appId, fn, version, force)).apply ...` and not `completeWithDeploymentForApp`. ...ok, I throught `apply` belong to the update but it's part of the `onComplete` directive returned by  `onSuccessLegacy`. I have a hard time accepting this. We mix directives and actual logic too much.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: IntelliJ misbehaved.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: I agree, but refactoring will require quite a lot of changes, let's do it in another PR.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: 1. sure
2. yes, it is, but not in this PR

PR #5603, opened 2017-10-12
 AppsController split: `patch multiple apps` method

meln1k:
Summary: splitting of the big apps controller pr, this time `patch multiple apps` method

JIRA issues: MARATHON-7299

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -38,9 +38,8 @@ import mesosphere.marathon.core.task.tracker.InstanceTracker
 import mesosphere.marathon.core.task.tracker.InstanceTracker.InstancesBySpec
 import mesosphere.marathon.core.task.Task.{ Id => TaskId }
 import PathMatchers._
-import mesosphere.marathon.raml.DeploymentResult
+import mesosphere.marathon.raml.{ AnyToRaml, AppUpdate, DeploymentResult }
 import mesosphere.marathon.raml.EnrichedTaskConversion._
-import mesosphere.marathon.raml.AnyToRaml
 
 import scala.concurrent.{ ExecutionContext, Future }
 import scala.util.{ Failure, Success, Try }
@@ -164,17 +163,33 @@ class AppsController(
   private[this] def updateMultiple(partialUpdate: Boolean, allowCreation: Boolean)(implicit identity: Identity): Route = {
     val version = clock.now()
     (forceParameter & entity(as(appUpdatesUnmarshaller(partialUpdate)))) { (force, appUpdates) =>
-      def updateGroup(rootGroup: RootGroup): RootGroup = appUpdates.foldLeft(rootGroup) { (group, update) =>
-        update.id.map(PathId(_)) match {
-          case Some(id) =>
-            group.updateApp(id, AppHelpers.updateOrCreate(id, _, update, partialUpdate, allowCreation, clock.now(), marathonSchedulerService), version)
-          case None =>
-            group
-        }
-      }
-
-      onSuccessLegacy(None)(groupManager.updateRoot(PathId.empty, updateGroup, version, force)).apply { plan =>
+      val updateGroupFn = updateAppsRootGroupModifier(appUpdates, partialUpdate, allowCreation, version)
+      onSuccessLegacy(None)(groupManager.updateRoot(PathId.empty, updateGroupFn, version, force)).apply { plan =>
         complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+      }
+    }
+  }
+
+  /**
+    * Helper function to update apps insude rootgroup
+    *
+    * @param appUpdates application updates
+    * @param partialUpdate do we allow partial updates or not
+    * @param allowCreation can we create a new app if appId is missing
+    * @param version version
+    * @return function for updating rootgroup
+    */
+  private[v2] def updateAppsRootGroupModifier(
+    appUpdates: Seq[AppUpdate],
+    partialUpdate: Boolean,
+    allowCreation: Boolean,
+    version: Timestamp)(implicit identity: Identity): RootGroup => RootGroup = { rootGroup: RootGroup =>
+    appUpdates.foldLeft(rootGroup) { (group, update) =>
+      update.id.map(PathId(_)) match {
+        case Some(id) =>
+          group.updateApp(id, AppHelpers.updateOrCreate(id, _, update, partialUpdate, allowCreation, clock.now(), marathonSchedulerService), version)
+        case None =>
+          group
       }
     }
   }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -171,7 +171,7 @@ class AppsController(
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -106,6 +106,18 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
       (body, plan)
     }
 
+    def addAppsToRootGroupRootGroup(apps: Seq[App]): (Seq[AppDefinition], DeploymentPlan, RootGroup) = {
```
zen-dog: `addAppsToRootGroupRootGroup` is this like "FactoryManagerManager`? ^..^

zen-dog: It seems that `preapreApp` and this one has a lot in common. Shouldn't one call the other (plus extras)? And shouldn't that be clear from their names?

jeschkies: I don't know. Such helper methods in tests often point to issue in the actual code. Why is it so hard to add something to the root group?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -106,6 +106,18 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+      val normed = apps.map(normalize)
+      val appDefs = normed.map(Raml.fromRaml(_))
```
jeschkies: What's happening here? Why do we have to normalize apps and the transform them back from the RAML?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1908,5 +1904,105 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+        appDefs.map { appDef =>
+        appDef.id -> appDef
+      }.toMap
+      )
+      val plan = DeploymentPlan(rootGroup, rootGroup)
+
+      groupManager.updateRoot(any, any, any, any, any) answers { args =>
+        val fn = args(1).asInstanceOf[RootGroup => RootGroup]
+        val updated = fn(rootGroup)
+        updated.app(app1Id).get.cmd.get shouldEqual newApp1Cmd
```
jeschkies: Just be explicit. `newApp1Cmd` is used only here so why not state the string explicitly

`updated.app(app1Id).get.cmd.get shouldEqual "bla1"`

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1908,5 +1904,105 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+        updated.app(app2Id).get.cmd.get shouldEqual newApp2Cmd
```
jeschkies: 🤔  hm, I'm not sure I like the assertions here. This diff is super close but I'm wondering if we should rather match the arguments. What do you think? What are we testing here?

meln1k: We're testing that original commands in the new root group was replaced by provided ones. I think without this the test won't make much sense at all :) But we can refactor this once we introduce a 'service' layer.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: Actually there is a json below where we use it too :)

jeschkies: True. Still. I don't think we save much here :).

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
jeschkies: I agree that these assertion should be there. However, I'd like them to be not in a mock. Is there not a way to match the arguments? Cause that's what you do. right? What is the argument anyways?

```scala
val fn = args(1).asInstanceOf[RootGroup => RootGroup]
val updated = fn(rootGroup)
```

seems a little odd. So the first arguments if cast to a function?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: If I inline it, someone will have to change code in two places instead of one. In my view it will make this test less developer-friendly.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -162,17 +163,33 @@ class AppsController(
...
+      onSuccessLegacy(None)(groupManager.updateRoot(PathId.empty, updateGroupFn, version, force)).apply { plan =>
+        complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
       }
+    }
+  }
 
-      onSuccessLegacy(None)(groupManager.updateRoot(PathId.empty, updateGroup, version, force)).apply { plan =>
-        complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+  /**
+    * Helper function to update apps insude rootgroup
```
jeschkies: Typo `insude` vs `inside`.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1908,5 +1899,146 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+          ]""".stripMargin.getBytes("UTF-8")
+
+      val entity = HttpEntity(body).withContentType(ContentTypes.`application/json`)
+      Patch(Uri./, entity) ~> route ~> check {
+        Then("404: Entity not found")
+        status shouldEqual StatusCodes.NotFound
+      }
+    }
+
+    "Correctly update the RootGroup when updating multiple apps and app creation is allowed" in new Fixture {
```
jeschkies: Nice. I like this.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -1908,5 +1899,146 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+        List(
+          AppUpdate(id = Some("/app1"), cmd = Some("newCmd")),
+          AppUpdate(id = Some("/newapp"), cmd = Some("newCmd2"))
+        ),
+        partialUpdate = true,
+        allowCreation = false,
+        Timestamp.now()
+      )(new Identity {})
+      Then("rootgroup is not updated")
+      Try(rootGroupUpdatefn(rootGroup)).isFailure shouldEqual true
```
jeschkies: Nit: ScalaTest is nice matchers for that.

```scala
an [Exception] should be thrownBy rootGroupUpdatefn(rootGroup)
```

PR #5605, opened 2017-10-12
Configure probot/stale to mark pull requests as stale.

jeschkies:
Summary:
This uses [probot](https://probot.github.io/apps/stale/) to manage our pull requests.
When a pull request has seen no activity in the last seven days it is marked as staled.
A stale PR is closed after sixty days without any activity.

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/akkahttp/SystemControllerTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/akkahttp/SystemControllerTest.scala
@@ -1,0 +1,5 @@
+package mesosphere.marathon.api.akkahttp
+
+class SystemControllerTest {
+
+}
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/marathon/api/akkahttp/SystemControllerTest.scala
```diff
@@ -0,0 +1,5 @@
+package mesosphere.marathon.api.akkahttp
```
ichernetsky: Why is it a part of this PR?

jeschkies: Ah, that slipped in. Sorry.

alenkacz:
I agree with @unterstein - to some level.

I thought this is intended for PRs that we attempt to help to be merged, but e.g. the other side is not interested in working on it (and I think this happens in OS projects often - someone submits PR, received feedback, realizes there is much more work that would have to be put into that and just abandon the PR)

But I agree that we probably have to change the internal process too, not just merge this bot...

jeschkies:
Let's drop it.

PR #5610, opened 2017-10-12
Store a flag node in ZK to indicate that Marathon performs data migration

ichernetsky:
JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-7788

### DIFF
--- a/docs/docs/data-migration.md
+++ b/docs/docs/data-migration.md
@@ -14,4 +14,4 @@ The following tips will help you avoid data loss during a Marathon upgrade.
 
 - Do not interrupt the leading Marathon instance while it is performing data migration.
 
-If data migration gets interrupted for any reason, restore the data and start another Marathon instance of the newer version to repeat the process. While performing data migration, Marathon creates a `migration-in-progress` node. If a Marathon is interrupted during migration, it will fail when it becomes leader again and detects the existence of the `migration-in-progress` node.
+If data migration gets interrupted for any reason, restore the data and start another Marathon instance of the newer version to repeat the process. While performing data migration, Marathon creates a `state/migration-in-progress` node. If a Marathon is interrupted during migration, it will fail when it becomes leader again and detects the existence of the `state/migration-in-progress` node.
--- a/docs/docs/data-migration.md
+++ b/docs/docs/data-migration.md
@@ -1,0 +1,16 @@
+---
+title: Data migration
+---
+
+# Data migration
+
+Marathon can store its data either in ZooKeeper or in-memory. In this document only Zookeeper is going to be considered, because there is no data migration possible in case of the in-memory only persistence store.
+
+Marathon stores data in ZooKeeper according to a certain layout, and different version of Marathon tend to have different data layouts in ZooKeeper. When a Marathon instance of a newer version becomes a leader for the first time, the first thing it does, is data migration. Therefore it is important
+
+1. to create a data backup before upgrading a Marathon just in case a connection to ZooKeeper drops or there is a bug in Marathon.
+1. to not interrupt the leading Marathon instance while it is performing data migration.
+
+If data migration gets interrupted for any reason, it is recommended to restore the data, and start another Marathon instance of a newer version to repeat the process. While performing migration, Marathon stores a `migration-in-progress` node. If it gets interrupted while working on it, it is going fail once it becomes a leader again and detects that there is the node lingering.
+
+Data migration might depend on environment variables or command-line flags or both. In the following sections all such variables and flags are explained.
--- a/docs/docs/data-migration.md
+++ b/docs/docs/data-migration.md
@@ -4,13 +4,14 @@ title: Data migration
 
 # Data migration
 
-Marathon can store its data either in ZooKeeper or in-memory. In this document only Zookeeper is going to be considered, because there is no data migration possible in case of the in-memory only persistence store.
+Marathon can store its data either in ZooKeeper or in-memory. If data is stored in ZooKeeper, your data will be migrated when you upgrade Marathon.
 
-Marathon stores data in ZooKeeper according to a certain layout, and different version of Marathon tend to have different data layouts in ZooKeeper. When a Marathon instance of a newer version becomes a leader for the first time, the first thing it does, is data migration. Therefore it is important
+Different versions of Marathon can have different data layouts in ZooKeeper. When a newer version of Marathon becomes leader for the first time, it moves the data in ZooKeeper to the new layout.
 
-1. to create a data backup before upgrading a Marathon just in case a connection to ZooKeeper drops or there is a bug in Marathon.
-1. to not interrupt the leading Marathon instance while it is performing data migration.
+The following tips will help you avoid data loss during a Marathon upgrade.
 
-If data migration gets interrupted for any reason, it is recommended to restore the data, and start another Marathon instance of a newer version to repeat the process. While performing migration, Marathon stores a `migration-in-progress` node. If it gets interrupted while working on it, it is going fail once it becomes a leader again and detects that there is the node lingering.
+- Back up your data before upgrade in case the connection to ZooKeeper drops or there is a bug in Marathon.
 
-Data migration might depend on environment variables or command-line flags or both. In the following sections all such variables and flags are explained.
+- Do not interrupt the leading Marathon instance while it is performing data migration.
+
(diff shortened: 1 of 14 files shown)


src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -100,15 +100,22 @@ class Migration(
...
     await(config.backup.map(uri => backup.backup(new URI(uri))).getOrElse(Future.successful(Done)))
     // step 2: restore state from given backup
     await(config.restore.map(uri => backup.restore(new URI(uri))).getOrElse(Future.successful(Done)))
+
+    // mark migration as started
+    await(persistenceStore.startMigration())
     // last step run the migration, to ensure we can operate on the zk state
     val result = await(migrateStorage(backupCreated = config.backup.isDefined || config.restore.isDefined))
+    // mark migration as completed
+    await(persistenceStore.endMigration())
```
timcharper: There are some migrations that legitimately abort due to a configuration issue. IE https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala#L64

Perhaps we should introduce a MigrationCancelled exception or something? And catch and endMigration ?

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
```diff
@@ -148,6 +148,26 @@ class InMemoryPersistenceStore(implicit
 
     Future.successful(Done)
   }
+
+  @volatile private[this] var migrationInProgress: Boolean = false
```
meln1k: What do you think about using AtomicBoolean here?

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
```diff
@@ -148,6 +148,26 @@ class InMemoryPersistenceStore(implicit
...
+
+  override def startMigration(): Future[Done] = {
+    require(isOpen, "the store must be opened before it can be used")
+    if (migrationInProgress) {
+      throw new IllegalStateException("Migration is already in progress")
```
meln1k: Does it make sense to return`Future.failed` here?

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
```diff
@@ -148,6 +148,26 @@ class InMemoryPersistenceStore(implicit
...
+    }
+    migrationInProgress = true
```
meln1k: I'm afraid it's not very thread-safe. We can fix this using atomic boolean.

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
ichernetsky: A very valid point. Thanks, Tim. I will do it.

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
ichernetsky: Initially, this method was doing nothing, because it doesn't store anything in an external storage, so there is not distribution. It is purely for completeness and uniformity just for unit-tests. Though I agree that it makes sense to do it right, since it is just one-two lines more.

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
ichernetsky: Will do.

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
ichernetsky: Done.

ichernetsky:
@timcharper yes, I was just waiting till the team is happy with the code, before writing any guide.

timcharper:
Changes look great so far!

We discussed in chat that before we merge we will have some documentation and point the user to it via the logged exception message when a migration refuses to start due to a prior failure.

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
```diff
@@ -148,6 +149,24 @@ class InMemoryPersistenceStore(implicit
...
   }
+
+  private[this] val migrationInProgress: AtomicBoolean = new AtomicBoolean(false)
+
+  override def startMigration(): Future[Done] = {
+    require(isOpen, "the store must be opened before it can be used")
+    if (!migrationInProgress.compareAndSet(false, true)) {
+      Future.failed(new IllegalStateException("Migration is already in progress"))
+    }
+    Future.successful(Done)
```
meln1k: I think we need to wrap this Future in `else` block otherwise we will always return success :)

src/main/scala/mesosphere/marathon/core/storage/store/impl/memory/InMemoryPersistenceStore.scala
```diff
@@ -148,6 +149,24 @@ class InMemoryPersistenceStore(implicit
...
+  }
+
+  override def endMigration(): Future[Done] = {
+    require(isOpen, "the store must be opened before it can be used")
+    if (!migrationInProgress.compareAndSet(true, false)) {
+      Future.failed(new IllegalStateException("Migration has not been started"))
+    }
+    Future.successful(Done)
```
meln1k: Same here :)

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
```diff
@@ -351,4 +351,42 @@ class ZkPersistenceStore(
...
+  override def startMigration(): Future[Done] = {
+    require(isOpen, "the store must be opened before it can be used")
+
+    async {
+      await(client.create("/migration-in-progress").asTry) match {
+        case Success(_) =>
+          Done
+        case Failure(e: KeeperException.NodeExistsException) =>
+          throw new StoreCommandFailedException(
+            "Migration is already in progress; /migration-in-progress node already exists", e)
```
jeschkies: Hm, is the exception really called `NodeExistsException`? This is odd if the node already exists.

ichernetsky: Why is it odd?

ichernetsky:
@sascala could you please review `data-migration.md`? Thanks in advance.

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
timcharper: This is sensible to me. If you try to create a node, but it already exists, then this is an exception.

This keeps it atomic. If, by chance, two marathon instances were trying to run migrations simultaneously, only one of them would succeed, and the other would fail.

docs/docs/data-migration.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Data migration
+---
+
+# Data migration
+
+Marathon can store its data either in ZooKeeper or in-memory. If data is stored in ZooKeeper, your data will be migrated when you upgrade Marathon.
```
timcharper: I wouldn't even talk about in-memory store. The backend is only for testing purposes and I would never encourage a user to actually use it.

ichernetsky: Sure. I will remove it.

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
jeschkies: Oh jeez. You know, I read `NoNodeExistsException`. Shame on me.

PR #5612, opened 2017-10-13
Extend the set of command-line flags returned by /v2/info

ichernetsky:


ichernetsky:
@alenkacz why do you think it would be nice to do? `/v2/info` supports only `GET` method.

jeschkies:
I agree with @alenkacz. For master we have to add the fields to the RAML. However, they do not exist for 1.4 and 1.5.

ichernetsky:
@jeschkies I don't get what the benefit is, if have a RAML type for JSON objects returned by `/v2/info`? Am I missing something?

PR #5613, opened 2017-10-13
Extend the set of command-line flags returned by /v2/info

ichernetsky:


ichernetsky:
@alenkacz why do you think it would be nice to do? `/v2/info` supports only `GET` method.

alenkacz:
@ichernetsky not sure I understand your comment, so maybe I misunderstood the whole thing... ? I thought that what you added here is actually what is returned by /v2/info endpoint described here https://mesosphere.github.io/marathon/api-console/index.html and all those additional fields should be present on this page too...

ichernetsky:
@alenkacz I created PRs with the updated `info.json` file to have that example in the docs updated too.

PR #5614, opened 2017-10-13
Replace last usage of InstanceCreationHandler with TaskStateOpProcessor

meichstedt:
Replace last usage of InstanceCreationHandler with TaskStateOpProcessor

Summary: TaskStateOpProcessor was introduced to replace InstanceCreationHandler. There was one remaining usage of the legacy trait, which has been removed.

Renamed `TaskStateOpProcessor` and `InstanceCreationHandlerAndUpdaterDelegate` to better reflect their purpose. Both traits had misleading names: `TaskStateOpProcessor` is actually about processing instance operations, and `InstanceCreationHandlerAndUpdaterDelegate` ... simply didn't have a very readable name. Since the latter now only implements `InstanceStateOpProcessor`, it was renamed accordingly. Arguably, `InstanceStateOpProcessor` could be named `InstanceUpdateOpProcessor`, or `InstanceUpdateOperation` could be renamed to `InstanceStateOperation`. The latter would imply too many file changes in my opinion.

JIRA issues: MARATHON-7830


meichstedt:
Even though the change is hopefully considered small, explaining a little:

`InstanceCreationHandlerAndUpdaterDelegate` has 3 methods: `created, terminated, process`. All three had the same argument list, while `process` returns a `Future[InstanceUpdateEffect]`. `created` and `terminated` used `process` internally and mapped the result to a `Future[Done]`. These methods have been removed and the call sites are adjusted to use `process` instead.

This change introduces only a few real additions where, in tests, an `InstanceUpdateEffect` is stubbed instead of a `Done`. On top of that, I adjusted the names of the trait and its implementation.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceStateOpProcessorDelegate.scala
```diff
@@ -3,47 +3,34 @@ package core.task.tracker.impl
...
 
   import mesosphere.marathon.core.async.ExecutionContexts.global
 
   private[impl] implicit val timeout: Timeout = conf.internalTaskUpdateRequestTimeout().milliseconds
 
   override def process(stateOp: InstanceUpdateOperation): Future[InstanceUpdateEffect] = {
-    taskUpdate(stateOp.instanceId, stateOp)
-  }
-
-  override def created(taskStateOp: InstanceUpdateOperation): Future[Done] = {
```
jeschkies: So this is also implemented somewhere else? Why is this not required anymore?

meichstedt: `created` and `terminated` forwarded the call to `process` until now; I've adjusted all call sites to use `process(taskStateOp)` instead, to get rid of the redundant interface

meichstedt: Look at the removed impl: it returned the result of `process(taskStateOp).map(_ => Done)`. Now the call sites directly call `process`.

zen-dog: Yep, here: https://github.com/mesosphere/marathon/pull/5614/files/d164bb1e4998d6ed044126d7f1e8d8d7b4b966b3#diff-61f68bd4b1b588ea69af252b8377f3efR126
is basically the actual change.

Issue #476, opened 2014-08-24
Task API should distinguish between tasks that are running and staging


guenter:
The json version of the tasks API should include a `status` key that is `(running|staging)`
The text/plain tasks API should only return tasks that are running.

PR #5618, opened 2017-10-15
Implement list tasks API method using Akka.HTTP

alenkacz:
Summary:
It implements first part of TasksController - method that lists the tasks. (for now, it handles output only in json, text/plain output will be in one of the following PRs)

I had to add few tests because some part of code were not covered in previous implementation.

JIRA issues: MARATHON-7308


src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+      pathEndOrSingleSlash {
+        listTasks()
+      }
+    }
+  }
+
+  private def listTasks(): Route = {
+    authenticated.apply { implicit identity =>
+      authorized(ViewResource, AuthorizedResource.SystemConfig).apply {
+        parameters("status".?, "status[]".as[String].*) { (statusParameter, statusesParameter) =>
```
jeschkies: Hm, should `statusesParameter` rather be `statusParameters`? I'm actually surprised that we need `"status".?` at all. Is this not covered by `"status[]".as[String].*`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+          val statuses = statusParameter.map(s => Seq(s) ++ statusesParameter).getOrElse(Seq[String](statusesParameter.toSeq: _*))
```
jeschkies: Hm, how about

```scala
val statuses = statusParameter.fold(Seq.empty[String])(s => Seq(s)) ++ statusesParameter
```

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+    val appToPorts: Map[PathId, Seq[Int]] = groupManager.apps(appIds).map {
+      case (appId, app) => appId -> app.map(_.servicePorts).getOrElse(Nil)
+    }
+
+    val instancesHealth = await(
+      Future.sequence(appIds.map { appId =>
+        healthCheckManager.statuses(appId)
+      })).foldLeft(Map[Id, Seq[Health]]())(_ ++ _)
+
+    val enrichedTasks: Iterable[Iterable[EnrichedTask]] = for {
```
jeschkies: We try to avoid for comprehensions for performance reasons.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+          instance.agentInfo,
+          instancesHealth.getOrElse(instance.instanceId, Nil),
+          appToPorts.getOrElse(appId, Nil)
+        )
+      }
+    }
+    Seq(enrichedTasks.flatten.toSeq: _*)
+  }
+
+  private def toTaskState(state: String): Option[Condition] = state.toLowerCase match {
```
jeschkies: Don't we have a mapping already? Also, I think it should be `toCondition`. @unterstein introduced `Condition`. He probably can explain the difference better. We should also document it then.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
```diff
@@ -0,0 +1,176 @@
...
+      val app = AppDefinition(id = "/app".toPath, instances = 1)
+      val (runningInstance, runningTask) = runningInstanceAndItsTask(app, clock)
+      val (stagingInstance, stagingTask) = stagingInstanceAndItsTask(app, clock)
+
+      instanceTracker.instancesBySpec() returns Future.successful(InstanceTracker.InstancesBySpec.of(InstanceTracker.SpecInstances.forInstances("app".toPath, Seq(stagingInstance, runningInstance))))
+      groupManager.apps(any) returns Map(app.id -> Some(app))
+      healthCheckManager.statuses(any) returns Future.successful(Map.empty)
+
+      When("Getting the tasks list for running as well as staging")
+      Get(Uri./.withQuery(Query("status[]" -> "staging", "status[]" -> "running"))) ~> controller.route ~> check {
```
jeschkies: Ah, the query parameter is called `status[]`? I don't think this is defined in the RAML.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: hmm, never heard of that (not saying it is not true). Can you maybe explain more or provide some article... ?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
alenkacz: yeah, in the old tasks resource, there is status (single) but it also accepts array of status[] (the old php style, probably). I thought I should actually keep it backward compatible...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: for some reason, this endpoint works only with two conditions (running and staging), rest is ignored. So if we want to keep that, we probably need this... I'll look into it more

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: I think this was already answered in one of your other comments - to keep backward compatibility, I need to accept "status" and "status[]"

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: better, thanks

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: Btw. I found that status[] was introduced here https://github.com/mesosphere/marathon/issues/476

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+        EnrichedTask(
+          appId,
+          task,
+          instance.agentInfo,
+          instancesHealth.getOrElse(instance.instanceId, Nil),
+          appToPorts.getOrElse(appId, Nil)
+        )
+      }
+    }
+    Seq(enrichedTasks.flatten.toSeq: _*)
```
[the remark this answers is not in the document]
alenkacz: that's not possible because of the scala.collection vs scala.collection.immutable but I found a nicer way to do this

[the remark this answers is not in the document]
meln1k: Will `.toList`/`.toVector` work?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+    val instances: Iterable[(PathId, Instance)] = instancesBySpec.instancesMap.values.flatMap { appTasks =>
+      appTasks.instances.map(i => appTasks.specId -> i)
+    }
+    val appIds: Set[PathId] = instancesBySpec.allSpecIdsWithInstances
+
+    val appToPorts: Map[PathId, Seq[Int]] = groupManager.apps(appIds).map {
+      case (appId, app) => appId -> app.map(_.servicePorts).getOrElse(Nil)
+    }
+
+    val instancesHealth = await(
```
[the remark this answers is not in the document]
alenkacz: did not know that, thanks

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,105 @@
...
+    val authenticator: Authenticator,
+    val authorizer: Authorizer
+) extends Controller with StrictLogging {
+
+  import mesosphere.marathon.api.akkahttp.Directives._
+  import mesosphere.marathon.api.akkahttp.EntityMarshallers._
+  import mesosphere.marathon.raml.EnrichedTaskConversion._
+
+  override val route = {
+    get {
```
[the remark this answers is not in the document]
alenkacz: we apparently do, thanks

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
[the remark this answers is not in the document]
alenkacz: I think the correct way is to use .to[Seq] that I now found it used only in other parts of the codebase

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
jeschkies: I could not find the source anymore. I guess overall the downside is that it's hiding performance bottlenecks.

jeschkies: Sorry, I could not find the source. I guess in general it's hiding bottlenecks. E.g. this could be:

```scala
instances
  .filter( (appId, instance) => conditionSet.isEmpty || conditionSet(instance.state.condition))
  .flatMap( (appId, instance)  =>
    tasks = instance.tasksMap.values
    tasks.map { task => EnrichedTask(...)
  }(breakOut)
```

As you can see I already have issues placing the if-clause. The for comprehension is not really making this clear. A collect should work as well.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
alenkacz: I documented the status[] in RAML

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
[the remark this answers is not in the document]
alenkacz: added to the listTasks method if you are looking for it :)

alenkacz:
@jeschkies @meln1k and others - this PR is ready for your re-review. Thanks :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,115 @@
...
+            listTasks()
+          }
+        }
+      }
+    }
+  }
+
+  private def listTasks()(implicit identity: Identity): Route = {
+    parameters("status".?, "status[]".as[String].*) { (statusParameter, statusParameters) =>
+      val statuses = statusParameter.fold(Seq.empty[String])(s => Seq(s)) ++ statusParameters
```
meln1k: Will `statusParameters ++ statusParameter` work?

docs/docs/rest-api/public/api/v2/types/enrichedTask.raml
```diff
@@ -36,3 +36,12 @@ types:
...
       localVolumes?: LocalVolumeId[]
+
+  EnrichedTasksList:
+    type: object
+    properties:
+      tasks: EnrichedTask[]
+
+  TaskStatusCondition:
+    type: string
+    enum: [ running, staging ]
```
jeschkies: Nit: I think a newline is missing.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -0,0 +1,115 @@
...
+      }
+      .flatMap {
+        case (appId, instance) => instance.tasksMap.values.map(t => EnrichedTask(
+          appId,
+          t,
+          instance.agentInfo,
+          instancesHealth.getOrElse(instance.instanceId, Nil),
+          appToPorts.getOrElse(appId, Nil)
+        ))
+      }.to[Seq]
```
jeschkies: Nice!

PR #5619, opened 2017-10-15
Update also condition when creating running task via test builder

alenkacz:
Summary:
From my opinion, when creating task by test builder, condition should be updated too, otherwise we create task that have status Running and condition Starting (which seems strange to me - but I don't fully understand the condition vs status in the codebase).

So from my opinion, this is how the code should look like... feel free to correct me :-)

JIRA issues:


alenkacz:
@jeschkies yep, I think I heard about it from @meichstedt too...

I just fixed it because it was actually working really weird for me. I created instance with running task, but the state of the instance was starting, not running, because the condition was not adjusted accordingly and instance reads just the condition.

BTW. it would be nice if the class did not even allow to instantiate it in such invalid state (condition and mesos status with different values)

PR #5620, opened 2017-10-16
Rate limit ProcessOutputToLogStream

timcharper:
Sometimes, our sub-processes (such as Mesos), can get in a really tight loop and produce an overwhelming amount of output that overwhelms Jenkins and inhibits the test tear-down code from running.

Looking at a few of such cases, it seems obvious that we can effectively limit it via our ProcessOutputToLogStream logger class.

JIRA Issues: MARATHON-7809

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
+++ b/src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
@@ -4,25 +4,28 @@ package integration.setup
 import org.slf4j.{ LoggerFactory, Logger }
 
 import scala.sys.process.ProcessLogger
+import scala.concurrent.duration._
 
-final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, periodMillis: Long = 1000L) extends ProcessLogger {
-  private[this] var nextFlush = System.currentTimeMillis() + periodMillis
-  private[this] var callsSinceLastFlush = 0
+final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, period: FiniteDuration = 1.second) extends ProcessLogger {
+  private[this] var nextPeriodStart = System.currentTimeMillis() + periodMillis
+  private[this] var callsThisPeriod = 0
+  private val periodMillis = period.toMillis
+
   val log: Logger = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
 
   private def rateLimited(msg: String): Boolean = synchronized {
     val now = System.currentTimeMillis
-    if (now > nextFlush) {
-      val suppressedLines = callsSinceLastFlush - maxLinesPerPeriod
+    if (now > nextPeriodStart) {
+      val suppressedLines = callsThisPeriod - maxLinesPerPeriod
       if (suppressedLines > 0)
         log.error(s"ProcessLogger ${process} was rate limited; ${suppressedLines} lines were supressed")
 
-      callsSinceLastFlush = 1
-      nextFlush = now + periodMillis
+      callsThisPeriod = 1
+      nextPeriodStart = now + periodMillis
       false
     } else {
-      callsSinceLastFlush += 1
-      callsSinceLastFlush > maxLinesPerPeriod
+      callsThisPeriod += 1
+      callsThisPeriod > maxLinesPerPeriod
     }
   }
 
--- a/src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
+++ b/src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
@@ -1,13 +1,40 @@
 package mesosphere.marathon
 package integration.setup
 
-import org.slf4j.LoggerFactory
+import org.slf4j.{ LoggerFactory, Logger }
 
 import scala.sys.process.ProcessLogger
 
-case class ProcessOutputToLogStream(process: String) extends ProcessLogger {
-  val log = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
-  override def out(msg: => String): Unit = log.debug(msg)
-  override def err(msg: => String): Unit = log.warn(msg)
+final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, periodMillis: Long = 1000L) extends ProcessLogger {
+  private[this] var nextFlush = System.currentTimeMillis() + periodMillis
+  private[this] var callsSinceLastFlush = 0
+  val log: Logger = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
+
+  private def rateLimited(msg: String): Boolean = synchronized {
+    val now = System.currentTimeMillis
+    if (now > nextFlush) {
+      val suppressedLines = callsSinceLastFlush - maxLinesPerPeriod
+      if (suppressedLines > 0)
+        log.error(s"ProcessLogger ${process} was rate limited; ${suppressedLines} lines were supressed")
+
+      callsSinceLastFlush = 1
+      nextFlush = now + periodMillis
+      false
+    } else {
+      callsSinceLastFlush += 1
+      callsSinceLastFlush > maxLinesPerPeriod
+    }
+  }
+
+  override def out(getMsg: => String): Unit = if (log.isDebugEnabled()) {
+    val msg = getMsg
+    if (!rateLimited(msg))
+      log.debug(msg)
+  }
+  override def err(getMsg: => String): Unit = if (log.isWarnEnabled()) {
+    val msg = getMsg
+    if (!rateLimited(msg))
+      log.warn(msg)
+  }
   override def buffer[T](f: => T): T = f
 }
--- a/src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
+++ b/src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
@@ -15,7 +15,7 @@ final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: In
 
   private def rateLimited(msg: String): Boolean = synchronized {
     val now = System.currentTimeMillis
-    if (now > nextPeriodStart) {
+    if (now >= nextPeriodStart) {
       val suppressedLines = callsThisPeriod - maxLinesPerPeriod
       if (suppressedLines > 0)
         log.error(s"ProcessLogger ${process} was rate limited; ${suppressedLines} lines were supressed")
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,40 @@
...
-import org.slf4j.LoggerFactory
+import org.slf4j.{ LoggerFactory, Logger }
 
 import scala.sys.process.ProcessLogger
 
-case class ProcessOutputToLogStream(process: String) extends ProcessLogger {
-  val log = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
-  override def out(msg: => String): Unit = log.debug(msg)
-  override def err(msg: => String): Unit = log.warn(msg)
+final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, periodMillis: Long = 1000L) extends ProcessLogger {
```
jeschkies: Why not use `FiniteDuration` for `period`?

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,40 @@
...
+  private[this] var nextFlush = System.currentTimeMillis() + periodMillis
+  private[this] var callsSinceLastFlush = 0
+  val log: Logger = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
+
+  private def rateLimited(msg: String): Boolean = synchronized {
```
jeschkies: Unfortunately I have a hard time understanding the logic.
 * When do we flush? I do not see any call to flush.
 * In both if-branches we might rate limit, right? So why do we log only once that we rate limit?

Also, shouldn't this method be threadsafe? By changing `callsSinceLastFlush` it does not seem to be.

timcharper: On second thought, flush was a bad choice in naming; probably better to use reset.

How is the method not threadsafe? It is wrapped in a synchronized block.

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
timcharper: It could; I was wanting to avoid converting a duration to milliseconds on every call to `rateLimited`; it is logging code and performance here is more critical.

timcharper: But ¯\\\_(ツ)\_/¯ not a big deal to just pull out the millis in the constructor. :) ✅

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,43 @@
...
+import org.slf4j.{ LoggerFactory, Logger }
 
 import scala.sys.process.ProcessLogger
+import scala.concurrent.duration._
 
-case class ProcessOutputToLogStream(process: String) extends ProcessLogger {
-  val log = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
-  override def out(msg: => String): Unit = log.debug(msg)
-  override def err(msg: => String): Unit = log.warn(msg)
+final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, period: FiniteDuration = 1.second) extends ProcessLogger {
```
zen-dog: Is `100 lines/s` a good number? I never checked how many lines we produce during "high-peak times" e.g. bootstrap phase of the IT.

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,43 @@
...
+  private[this] var nextPeriodStart = System.currentTimeMillis() + periodMillis
+  private[this] var callsThisPeriod = 0
+  private val periodMillis = period.toMillis
+
+  val log: Logger = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
+
+  private def rateLimited(msg: String): Boolean = synchronized {
```
zen-dog: Rate limiting is one option. But I'm thinking may be it's even better to limit the overall size? E.g. Our test in a normal case are ~50-60Mb which roughly corresponds to `N` lines. So we place an total limit of lines to e.g. `2*N` and that's fine?

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
zen-dog: I just checked the log from https://jenkins.mesosphere.com/service/jenkins/job/marathon-pipelines/job/PR-5620/10/console and it seems that we're limiting a lot of "normal IT execution logging" which will of course make debugging harder.

timcharper:
@jeschkies My hypothesis was that when the Jenkins build jobs get stuck, it is not a deadlock, but it is too much logging overwhelming the Jenkins agent. This is corroborated by the fact that when the "logsplosion" happens, seconds worth of logs can take up to 1GB. I suspect, that at some point, the logsplosion starts to eat in to the JVMs ability to schedule threads effectively, cancel, and tear down.

At any rate, it seems that my hypothesis has been confirmed. https://jenkins.mesosphere.com/service/jenkins/job/marathon-pipelines/job/PR-5620/10/console was a stuck build. `SystemResourceIntegrationTest-MesosMaster-32839` went in to the frenzy. After we suppressed `7,492,039` log lines over 5 minutes, the Mesos Master process was finally killed, as designed, when the test timed out.

Of course, it's always possible that there are other causes for jenkins builds getting stuck. All of the issues I've seen, recently, are from when the Mesos Master gets into a tight logging loop. Evidence confirms that that case is solved by this patch.

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
timcharper: Limiting the overall size was my first hunch, too (and approach, as seen in Slack). Going down that path is what led me to discover that either logback or Jenkins starts to duplicate log lines once some buffer overflow event happens. I saw it happen as early as 1.4MB produced by a single logger. We run integration tests in parallel, so, it is possible that we could have multiple processes concurrently outputting spurious amounts of logs. If their combined rate causes us to hit this overflow threshold, then game over. It doesn't matter how much we rate limit, because whatever we log will be repeated (I believe forever).

I prefer rate limiting.

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
timcharper: Seems like we could just use a value of 500 and be safe. However, on the flip side, it would be nice to detect that a process is being noisy and curb its output.

Perhaps we start with a limit of 500, and once that is reached, we set the limit to 100?

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
jeschkies: Oh, I've missed the `synchronized`.

jeschkies:
> After we suppressed 7,492,039 log lines over 5 minutes, the Mesos Master process was finally killed, as designed, when the test timed out.

but this should not happen right? Why is the master not stopping? To me that's a bug.

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,58 @@
...
+  final class RateLimiter(log: Logger, maxLinesPerPeriod: Int, period: FiniteDuration, clock: Clock) {
+    private val periodMillis = period.toMillis
+    private[this] var nextPeriodStart = System.currentTimeMillis() + periodMillis
+    private[this] var linesThisPeriod = 0
+    def incrementAndCheck(): Boolean = synchronized {
+      val now = clock.millis()
+      if (now >= nextPeriodStart) {
+        val suppressedLines = linesThisPeriod - maxLinesPerPeriod
+        if (suppressedLines > 0)
+          log.error(s"Logger was rate-limited; ${suppressedLines} lines were supressed")
```
jeschkies: Nit:
`log.error(s"Logger was rate-limited; ${suppressedLines} lines were suppressed during last period.")`

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
```diff
@@ -1,13 +1,58 @@
...
+      }
+    }
+  }
+}
+
+final case class ProcessOutputToLogStream(process: String, maxLinesPerPeriod: Int = 100, period: FiniteDuration = 1.second) extends ProcessLogger {
+  private val log: Logger = LoggerFactory.getLogger(s"mesosphere.marathon.integration.process.$process")
+  private val rateLimiter = new RateLimiter(log, maxLinesPerPeriod, period, Clock.systemUTC())
+
+  override def out(getMsg: => String): Unit = if (log.isDebugEnabled()) {
```
jeschkies: I might be wrong but shouldn't this line be thread safe as well? I do not see an apparent issue but just checking.

meln1k: Looks pretty thread-safe to me.

timcharper:
Yes, it is not the core issue. The test times out and fails. This at least keeps Jenkins from crapping out.

src/test/scala/mesosphere/marathon/integration/setup/ProcessOutputToLogStream.scala
timcharper: Yeah, it's fine; rateLimiter is thread safe. `out` and `err` are independently called serially, but, as I understand it, could each be called concurrently.

zen-dog:
I hope that this PR will also solve the issue (with some additional perks): https://github.com/mesosphere/marathon/pull/5633
 by simply pumping `stdout` in the file - `stderr` alone is very few lines and would (in theory) never be able to break jenkins agent. 
However I can see both approaches coexist - we should make sure however that `maxLinesPerPeriod` is high enough (1k is enough?) and is never seen during normal test runtime.

zen-dog:
And since we already have `Guava` in dependencies - shouldn't we just take it?
```
import com.google.common.util.concurrent.RateLimiter

val rateLimiter: RateLimiter = RateLimiter.create(1000.0); 
if (rateLimiter.tryAcquire()){
  ...
}
```

PR #5622, opened 2017-10-16
Status parameter is not required on the tasks endpoint

alenkacz:
Status parameter is not required on the tasks endpoint

Summary:
Small fix for the tasks endpoint RAML documentation. If status parameter is not provided, task of all statuses are returned. If you don't believe me, feel free to test it out here :) https://soak.mesosphere.com/service/marathon/v2/tasks

JIRA issues: MARATHON-7308

### DIFF
--- a/docs/docs/rest-api/public/api/v2/tasks.raml
+++ b/docs/docs/rest-api/public/api/v2/tasks.raml
@@ -5,7 +5,7 @@ get:
   queryParameters:
     status:
       required: false
-      description: Filter the list of tasks by status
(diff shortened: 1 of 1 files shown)


docs/docs/rest-api/public/api/v2/tasks.raml
```diff
@@ -1,9 +1,10 @@
 get:
   description:
-    List all running tasks.
+    List all tasks.
   is: [ secured ]
   queryParameters:
     status:
+      required: false
```
unterstein: Please add description about the behavior when status parameter is not provided.

PR #5627, opened 2017-10-16
Test system controller.

jeschkies:
Summary:
The controller was already implemented. This diff adds some tests. Please see my comment on the metrics endpoint test.

JIRA issues: MARATHON-7311

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpModule.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpModule.scala
@@ -66,7 +66,7 @@ class AkkaHttpModule(conf: MarathonConf with HttpConf) extends AbstractModule {
       pluginManager = pluginManager)
 
     val resourceController = new ResourceController
-    val systemController = new SystemController(config)
+    val systemController = new SystemController(conf, config, electionService)
     val eventsController = new EventsController(conf, eventBus)
     val infoController = InfoController(mesosLeaderInfo, storageModule.frameworkIdRepository, conf)
     val pluginsController = new PluginsController(pluginManager.plugins[HttpRequestHandler], pluginManager.definitions)
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
@@ -89,7 +89,7 @@ class SystemController(
   def changeLoggers: Route = {
     authenticated.apply { implicit identity =>
       authorized(UpdateResource, SystemConfig).apply {
-        entity(as[LoggerChange]) { change =>
+        entity(as[raml.LoggerChange]) { change: raml.LoggerChange =>
           LoggerFactory.getILoggerFactory.getLogger(change.logger) match {
             case log: Logger =>
               val level = Level.valueOf(change.level.value.toUpperCase)
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
@@ -5,13 +5,13 @@ import akka.actor.ActorSystem
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.Accept
 import akka.http.scaladsl.server.Route
-import ch.qos.logback.classic.{Level, Logger, LoggerContext}
-import com.typesafe.config.{Config, ConfigRenderOptions}
+import ch.qos.logback.classic.{ Level, Logger, LoggerContext }
+import com.typesafe.config.{ Config, ConfigRenderOptions }
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.election.ElectionService
 import mesosphere.marathon.metrics.Metrics
-import mesosphere.marathon.plugin.auth.AuthorizedResource.{SystemConfig, SystemMetrics}
-import mesosphere.marathon.plugin.auth.{Authenticator, Authorizer, UpdateResource, ViewResource}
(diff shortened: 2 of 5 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
```diff
@@ -45,6 +67,7 @@ class SystemController(cfg: Config)(
   def metrics: Route = {
     authenticated.apply { implicit identity =>
       authorized(ViewResource, SystemMetrics).apply {
+        //        complete(Metrics.snapshot().toRaml)
```
meln1k: Let's do something with this :)

src/test/scala/mesosphere/marathon/api/akkahttp/SystemControllerTest.scala
```diff
@@ -0,0 +1,149 @@
...
+          response.status should be(statusCode)
+          response.entity.contentType should be(responseContentType)
+        }
+      }
+    }
+
+    "return a snapshot of the metrics" in {
+      val controller = Fixture().controller()
+      Get("/metrics") ~> controller.route ~> check {
+        response.status should be(StatusCodes.OK)
```
jeschkies: I cannot inject the time so it's harder to test the response body.

src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
jeschkies: Yep. I've forgot it.

src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpModule.scala
```diff
@@ -66,7 +66,7 @@ class AkkaHttpModule(conf: MarathonConf with HttpConf) extends AbstractModule {
       pluginManager = pluginManager)
 
     val resourceController = new ResourceController
-    val systemController = new SystemController(config)
+    val systemController = new SystemController(conf, config, electionService)
```
alenkacz: conf and config as two different parameters are very unclear. Can we find different (more specific name) for at least one of those parameters?

ichernetsky: +1

jeschkies: I've renamed one to `marathonConfig`. It is a weird setup. Ideally we would just use the typesafe config.

src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
```diff
@@ -20,23 +23,40 @@ import scala.concurrent.ExecutionContext
...
-    val authorizer: Authorizer,
-    val electionService: ElectionService
-) extends Controller with StrictLogging {
+    val authorizer: Authorizer
+) extends Controller with MetricsConversion with StrictLogging {
 
   import Directives._
   import EntityMarshallers._
 
+  def acceptsTextPlain(ranges: Seq[MediaRange]): Boolean = {
```
meln1k: Do you think we can build a `Unmarshaller which will handle all those types instead of matching them by our own?

src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
```diff
@@ -20,23 +23,40 @@ import scala.concurrent.ExecutionContext
...
+    ranges.exists(range => range.matches(MediaTypes.`application/json`))
+  }
   /**
     * GET /ping
     * @return a simple pong as text/plain
     */
-  def ping: Route = complete("pong")
+  def ping: Route = extract(_.request.header[Accept]) { maybeAccept: Option[Accept] =>
+    // Match by priority. Always prefer text/plain.
+    maybeAccept match {
```
meln1k: Akka http supports content negotiation :) for example: 
```scala
val respFuture = Marshal(response).toResponseFor(request) // with content negotiation!
```

meln1k: Wait, it actually should work implicitly, you just need some marshallers

jeschkies: Hm. I wonder if we really to go the implicit route here. I don't a benefit in hiding it just for this request. If we'd support this behaviour for more than one route, I'd agree with you. I'll look into to the content negotiation.

src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
```diff
@@ -20,23 +23,40 @@ import scala.concurrent.ExecutionContext
...
+      case None =>
+        complete("pong")
+      case Some(accept) if (acceptsTextPlain(accept.mediaRanges)) =>
+        complete("pong")
+      case Some(accept) if (acceptsApplicationJson(accept.mediaRanges)) =>
+        complete(JsString("ping"))
```
jdef: shouldn't this read `pong` instead of `ping`?

src/test/scala/mesosphere/marathon/api/akkahttp/SystemControllerTest.scala
```diff
@@ -0,0 +1,125 @@
...
+import mesosphere.marathon.core.election.ElectionService
+import org.scalatest.prop.TableDrivenPropertyChecks
+import org.slf4j.LoggerFactory
+import play.api.libs.json.{ JsDefined, JsObject, JsString, Json }
+
+class SystemControllerTest extends UnitTest with ScalatestRouteTest with RouteBehaviours with TableDrivenPropertyChecks {
+
+  "SystemController" should {
+    // format: OFF
+    val pingCases = Table[Option[MediaRange], ContentType, StatusCode](
```
jdef: maybe also test the returned entity matches expectations?

PR #5628, opened 2017-10-16
AppsController split: `delete single app` method

meln1k:
AppsController split: `delete single app` method

Summary: splitting of the big apps controller pr, this time `delete single app` method

JIRA issues: MARATHON-7299


src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
```diff
@@ -124,12 +125,22 @@ trait GroupManager {
...
     */
-  def updateRoot(
+  final def updateRoot(
     id: PathId,
     fn: RootGroup => RootGroup,
     version: Timestamp = Group.defaultVersion,
     force: Boolean = false,
-    toKill: Map[PathId, Seq[Instance]] = Map.empty): Future[DeploymentPlan]
+    toKill: Map[PathId, Seq[Instance]] = Map.empty): Future[DeploymentPlan] = {
+    updateRootEither[Nothing](id, fn.andThen(Right(_)), version, force, toKill)
```
alenkacz: If we use Either with Nothing, maybe it makes more sense to model this using Option?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -103,43 +103,54 @@ class GroupManagerImpl(
...
-        await(deploymentService.get().deploy(plan, force))
-        await(groupRepository.storeRoot(plan.target, plan.createdOrUpdatedApps, plan.deletedApps, plan.createdOrUpdatedPods, plan.deletedPods))
-        logger.info(s"Updated groups/apps/pods according to plan ${plan.id}")
-        // finally update the root under the write lock.
-        root := Option(plan.target)
-        plan
+    val maybeDeploymentPlan: Future[Either[T, DeploymentPlan]] = serializeUpdates {
+      logger.info(s"Upgrade root group version:$version with force:$force")
+
+      val from = rootGroup()
```
alenkacz: I read it few times but I still don't completely understand what motivated the change here - can you maybe explain more, since I don't know this part of code that well. Before your change, it always returned a deployment plan and now it does not have to? What changed?

src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
meln1k: It's done for the sake of backwards-compatibility with the current code. Once all required methods will accept either, it will be removed.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
meln1k: ### Before:
This method returned deployment plan and threw an exception if something would go wrong.
### Now:
This method returns deployment plan and an error object (in the most cases - rejection) explicitly using `Either`.
### Motivation:
In akka-http we explicitly handle **errors** (via rejections) and throw exceptions only when unexpected **failure** happened. 

Difference between these two is that **failure is an unexpected event** that prevents us from continuing to function normally. In contrast an **error is an expected event**, which we know how to handle, (for validation error or other hand-written thing). Errors will be communicated to the client as part of the normal processing of the message. 

You can find more here: https://www.reactivemanifesto.org/glossary#Failure

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -245,7 +248,30 @@ class AppsController(
       }
   }
 
-  private def deleteSingle(appId: PathId)(implicit identity: Identity): Route = ???
+  private def deleteSingle(appId: PathId)(implicit identity: Identity): Route =
+    forceParameter { force =>
+      lazy val notFound: Either[Rejection, RootGroup] =
```
jeschkies: What is the motivation for this helper? Why not return `Left(Rejections.EntityNotFound.noApp(appId))` directly? ...ah, since it makes `deleteApp` more readable, how about moving it out of `deleteSingle`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -245,7 +248,30 @@ class AppsController(
...
+          case None =>
+            notFound
+          case Some(app) =>
+            if (authorizer.isAuthorized(identity, DeleteRunSpec, app))
+              Right(rootGroup.removeApp(appId))
+            else
+              notAuthorized
+        }
+      }
+      onSuccess(groupManager.updateRootEither(appId.parent, deleteApp, force = force)) {
```
jeschkies: I'm sorry for my ignorance but is `updateRootEither` the new pattern?

src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
```diff
@@ -124,12 +125,22 @@ trait GroupManager {
...
+      .map(_.right.getOrElse(throw new RuntimeException("Either must be Right here")))(ExecutionContexts.callerThread)
+  }
+
+  def updateRootEither[T](
+    id: PathId,
+    fn: RootGroup => Either[T, RootGroup],
```
jeschkies: What's the benefit over exceptions? I seems we inject `Rejections` into the group manager with this.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
jeschkies: I do like the approach. However, `Rejections` should not leak into this area. Ideally we'd avoid not found and not authorized errors before we even call the logic layer. I used disjunctions in previous projects but I don't think we should introduce Scalaz or Cats. What kind of expected error can we not handle early on the controller layer?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -262,7 +265,33 @@ class AppsController(
...
+      onSuccess(groupManager.updateRootEither(appId.parent, deleteAppRootGroupModifier(appId), force = force)) {
+        case Right(plan) =>
+          completeWithDeploymentForApp(appId, plan)
+        case Left(rej) =>
+          reject(rej)
+      }
+    }
+
+  /**
+    * We have to pass this function to the groupManager to make sure updates are serialized
```
jeschkies: Hm, I think we need even more details.

```
We have to pass this function to the groupManager to make sure updates are serialized
	
Another request might remove the app before we call the update. That is why we cannot check if the app exists before the update call. We have to do so during the call. Since we do some app handling anyways we can also check if the caller is authroized.
```

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
alenkacz: @meln1k thanks for the explanation 👍

PR #5630, opened 2017-10-16
AppsController split: `restart app` method

meln1k:
Summary: splitting of the big apps controller pr, this time `restart app` method

JIRA issues: MARATHON-7299


src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -299,7 +299,28 @@ class AppsController(
     }
   }
 
-  private def restartApp(appId: PathId)(implicit identity: Identity): Route = ???
+  private def restartApp(appId: PathId)(implicit identity: Identity): Route = {
+    forceParameter { force =>
+      val newVersion = clock.now()
+      onSuccessLegacy(Some(appId))(
+        groupManager.updateApp(
```
jeschkies: Why no `updateAppEither`? 😉

meln1k: It will requite quite a lot of changes :) I think we can have a separate task for such refactoring.

PR #5632, opened 2017-10-17
Fix 1.4.6 UnreachableStrategy migration

fengyehong:
As inactiveAfter and expungeAfter defaults to 0 seconds now, we need
to match the old default values to make the migration work.


ichernetsky:
It should be backported/cherry-picked too, I guess.

unterstein:
Thanks @fengyehong!

```
    .  o ..                  
     o . o o.o                
          ...oo               
            __[]__          Shippin'  
         __|_o_o_o\__         it!
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^
```

PR #5633, opened 2017-10-17
CI `stdout` is printed to file instead of console

zen-dog:
Summary:
We're printing CI logs (sbt test integration:test...) directly to a log file. This should:
- improve CI run time since we don't print to console and upload zipped artifacts (1/10 of log size)
- theoretically solve our "jenkins log loop" problem by printing only very few lines (`stderr`) to console
- logs are archived by Jenkins
Additionally I removed an old remnant (`review.py`) of phabricator times


Jenkinsfile
```diff
@@ -16,6 +16,8 @@ ansiColor('gnome-terminal') {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
         junit allowEmptyResults: true, testResults: 'target/test-reports/*integration/*.xml'
         archive includes: 'sandboxes.tar.gz'
+        archive includes: 'ci.tar.gz'
+        archive includes: 'ci.log'  // Only in case the build was  aborted and the logs weren't zipped
```
jeschkies: Hm, this is done each time now. I'd just do it if the `ci.tar.gz` does not exist. It seems we do not have the `timeout` directive anymore.

zen-dog: Done. The `ci.log` is removed if the run was successful.

PR #5635, opened 2017-10-17
Wait for specific app in SI tests.

jeschkies:
Summary:
Some tests time out waiting for a deployment. We should be more specific
while waiting. This should also enable us to run tests in parallel in
the future.


tests/system/README.md
```diff
@@ -2,4 +2,21 @@
...
+First authenticate with your cluster with
+
+```
+dcos cluster setup --no-check --username=$DCOS_USER --password=$DCOS_PWD https://...
+```
+
+Then run a test with
+
+```
+shakedown test_marathon...py
```
zen-dog: Add: To run a specific test within a test file: ```shakedown test_marathon_root.py::test_private_repository_mesos_app```

tests/system/marathon_common_tests.py
```diff
@@ -39,7 +39,7 @@ def test_launch_docker_container():
 
     client = marathon.create_client()
     client.add_app(app_def)
-    shakedown.deployment_wait()
+    shakedown.deployment_wait(app_id=app_def["id"])
```
zen-dog: for almost all this tests `app_def["id"]` is used at least 2x times in the test. It probably makes sense to extract it to a var e.g. `app_id = app_def["id"]`

tests/system/system_integration.txt
```diff
@@ -0,0 +1,503 @@
+test_marathon_basics.py:def test_launch_mesos_container():
```
zen-dog: what is this? :)

jeschkies: This is some awesome stuff which should not be checked in.

tests/system/marathon_common_tests.py
```diff
@@ -449,7 +451,7 @@ def test_app_with_no_health_check_not_healthy():
     client = marathon.create_client()
     client.add_app(app_def)
 
-    shakedown.deployment_wait()
+    shakedown.deployment_wait(app_id=app_def["id"])
```
zen-dog: `app_id = ...`

tests/system/marathon_common_tests.py
```diff
@@ -242,7 +244,7 @@ def test_run_app_with_specified_user():
 
     client = marathon.create_client()
     client.add_app(app_def)
-    shakedown.deployment_wait()
+    shakedown.deployment_wait(app_id=app_def["id"])
```
zen-dog: `app_id = ...`

jeschkies:
All tests pass locally.

jeschkies:
> However, was this really the culprit for the timed out deployments? Since we only had one deployment and no parallel tests running? 🤔

I have my doubts as well. However, this change can rule out that option.

PR #5641, opened 2017-10-18
Akka EventsController Stream Logic and Tests

timcharper:
* Replace the custom stage implementation with stream composed via the DSL.
* Add unit tests

Jira Issues: MARATHON-7317

### DIFF
--- a/src/main/scala/mesosphere/marathon/stream/EnrichedFlow.scala
+++ b/src/main/scala/mesosphere/marathon/stream/EnrichedFlow.scala
@@ -1,0 +1,28 @@
+package mesosphere.marathon
+package stream
+
+import akka.NotUsed
+import akka.stream.scaladsl.{ Flow, Source }
+
+@SuppressWarnings(Array("AsInstanceOf"))
+object EnrichedFlow {
+  /**
+    * Drops all elements and has an output type of Nothing.
+    */
+  val ignore: Flow[Any, Nothing, NotUsed] =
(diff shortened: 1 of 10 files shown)


src/main/scala/mesosphere/marathon/stream/EnrichedFlow.scala
```diff
@@ -0,0 +1,28 @@
...
+    * Stops the current stream the moment an element is produced by specified source, without merging any elements from
+    * the specified source into the current stream
+    *
+    * In the following example, 4 TimeToPoll elements would be emitted:
+    *
+    *    Source.tick(0.mills, 1000.millis, TimeToPoll).via(stopOnFirst(Source(3500.millis, 3500.millis, 'stop)))
+    *
+    * @param source The Source which, upon producing its first element, causes the current stream to stop
+    */
+  def stopOnFirst[T](source: Source[Any, Any]): Flow[T, T, NotUsed] = {
```
meln1k: What do you thing about this?
```scala
def stopOnFirst[T](source: Source[_, _]): Flow[T, T, NotUsed] = {
```

timcharper: ¯\\\_(ツ)\_/¯ 

Source is covariant in both parameters.

timcharper:
```scala
final class Source[+Out, +Mat] 
```

PR #5642, opened 2017-10-18
Fix 1.4.6 UnreachableStrategy migration (#5632)

unterstein:
As inactiveAfter and expungeAfter defaults to 0 seconds now, we need
to match the old default values to make the migration work.

Back port of #5632 to releases/1.4


unterstein:
Not necessary needed in `releases/1.4`, because default is still 5 minutes / 10 minutes. But if we want to change the default in the future for 1.4, we should have this patch in place.

unterstein:
related to #5649, where we introduced the back port of the new defaults (0,0).

ichernetsky:
@unterstein then it means that we should not merge this PR at the moment?

ichernetsky:
Please ignore me. I found it #5649

PR #5644, opened 2017-10-18
CI `stdout` is printed to file instead of console (#5633)

jeschkies:
Summary:

We're printing CI logs (sbt test integration:test...) directly to a log file. This should:
- improve CI run time since we don't print to console and upload zipped artifacts (1/10 of log size)
- theoretically solve our "jenkins log loop" problem by printing only very few lines (`stderr`) to console
- logs are archived by Jenkins
Additionally an old remnant (`review.py`) of phabricator times was removed.


review.py
```diff
@@ -1,133 +0,0 @@
-#!/usr/bin/env python
```
meln1k: Why this was removed?

jeschkies: The script would create statics for Phabricator reviews E.g. how long are reviews open or how long would it take to get a first respond.

PR #5647, opened 2017-10-18
Mark SI test runs as unstable on cluster launch failure.

jeschkies:
Summary:
We want the system integration tests to be marked as unstable and not as
failed if the cluster launch failed. This gives us a more fine grained
overview of the daily runs.

### DIFF
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -60,6 +60,9 @@ function create-junit-xml {
 	EOF
 }
 
+create-junit-xml "dcos-launch" "cluster.create" "Cluster launch failed."
+exit 0
+
 if ! ./dcos-launch create; then
   create-junit-xml "dcos-launch" "cluster.create" "Cluster launch failed."
   exit 0
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -62,11 +62,11 @@ function create-junit-xml {
 
 if ! ./dcos-launch create; then
   create-junit-xml "dcos-launch" "cluster.create" "Cluster launch failed."
-  exit 1
+  exit 0
 fi
 if ! ./dcos-launch wait; then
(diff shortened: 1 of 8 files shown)


ci/launch_cluster.sh
```diff
@@ -62,11 +62,11 @@ function create-junit-xml {
 
 if ! ./dcos-launch create; then
   create-junit-xml "dcos-launch" "cluster.create" "Cluster launch failed."
-  exit 1
+  exit 0
```
jeschkies: If a build returns 0 but JUnit finds and error it will report the build as unstable.

zen-dog: Oh yes, this should be a comment in code. Otherwise the next guy will wonder.

ichernetsky: 👍

jeschkies:
Ok, it works now https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-open/job/karsten%252Fcluster-launch-unstable//

ci/launch_cluster.sh
```diff
@@ -44,29 +44,11 @@ template_parameters:
...
-	      </testcase>
-	  </testsuite>
-	</testsuites>
-	EOF
-}
-
 if ! ./dcos-launch create; then
-  create-junit-xml "dcos-launch" "cluster.create" "Cluster launch failed."
-  exit 1
+  exit 2
```
zen-dog: should we not comment the possible exit codes at the top?

jeschkies:
The build is green https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-open/job/karsten%252Fcluster-launch-unstable/14/

ci/launch_cluster.sh
ichernetsky: 👍 , since they are used in a case statement in the other script.

PR #5650, opened 2017-10-18
Refactored `SchedulerActions` object

zen-dog:
Summary:
Removed `stopRunnable` method completely and moved missing parts to the `DeploymentActor`. Partially that code (killing instances) was called twice between those two.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -379,35 +379,6 @@ class SchedulerActions(
     scale(runSpec)
   }
 
-  @SuppressWarnings(Array("all")) // async/await
-  def stopRunSpec(runSpec: RunSpec): Future[Done] = {
-    logger.info(s"Stopping runSpec ${runSpec.id}")
-
-    healthCheckManager.removeAllFor(runSpec.id)
-
-    async {
-      val tasks = await(instanceTracker.specInstances(runSpec.id))
-
-      tasks.foreach { instance =>
-        if (instance.isLaunched) {
-          logger.info("Killing {}", instance.instanceId)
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
```diff
@@ -186,20 +186,29 @@ private class DeploymentActor(
...
 
-    // Note: This is an asynchronous call. We do NOT wait for the run spec to stop. If we do, the DeploymentActorTest
-    // fails.
-    scheduler.stopRunSpec(runnableSpec)
+    // The tasks will be removed from the InstanceTracker when their termination
+    // was confirmed by Mesos via a task update.
+    eventBus.publish(AppTerminatedEvent(runSpec.id))
 
     Done
+  }.recover {
```
meln1k: Do you think it's a good idea to suppress the exception? What if a caller would like to know if something goes wrong?

zen-dog: I agree but at the very least it the same logic as previously.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -379,35 +379,6 @@ class SchedulerActions(
...
-  def stopRunSpec(runSpec: RunSpec): Future[Done] = {
-    logger.info(s"Stopping runSpec ${runSpec.id}")
-
-    healthCheckManager.removeAllFor(runSpec.id)
-
-    async {
-      val tasks = await(instanceTracker.specInstances(runSpec.id))
-
-      tasks.foreach { instance =>
-        if (instance.isLaunched) {
```
unterstein: Code in DeploymentActor is way clearer btw.

PR #5653, opened 2017-10-18
Edited logs to print `runSpecId` in all relevant deployment log lines

zen-dog:
Summary:
This should allow us to grep logs for some `runSpecId` and get all relevant deployment lines. Also complete deployment plan is  only printed once - this should further improve log readability.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/DeploymentPlan.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/DeploymentPlan.scala
@@ -186,6 +186,8 @@ case class DeploymentPlan(
       } else " NO STEPS"
     s"DeploymentPlan id=$id,$version$stepString\n"
   }
+
+  def targetIdsString = affectedRunSpecIds.mkString(", ")
 }
 
 object DeploymentPlan {
--- a/src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
+++ b/src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
@@ -28,17 +28,17 @@ object GroupVersioningUtil {
     def updateAppVersionInfo(maybeOldApp: Option[AppDefinition], newApp: AppDefinition): AppDefinition = {
       val newVersionInfo = maybeOldApp match {
         case None =>
-          log.info(s"[${newApp.id}]: new app detected")
+          log.info(s"${newApp.id}: new app detected")
           VersionInfo.forNewConfig(newVersion = version)
         case Some(oldApp) =>
           if (oldApp.isUpgrade(newApp)) {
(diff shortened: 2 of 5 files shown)


src/main/scala/mesosphere/marathon/core/deployment/DeploymentPlan.scala
```diff
@@ -186,6 +186,8 @@ case class DeploymentPlan(
       } else " NO STEPS"
     s"DeploymentPlan id=$id,$version$stepString\n"
   }
+
+  def targetIdsString = affectedRunSpecIds.mkString(", ")
```
unterstein: In this PR some occurrences of `plan.id` was replaced with `plan.targetIdsString` and some occurrences of `plan.target.id` was replaced with `plan.targetIdsString`.

`plan.target.id` is always `/` for RootGroup and we are only doing deployments on RootGroup afaik. On the other hand `plan.id` holds the UUID of the deployment and should be included in the log statements as well to make it searchable.

src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
```diff
@@ -28,17 +28,17 @@ object GroupVersioningUtil {
     def updateAppVersionInfo(maybeOldApp: Option[AppDefinition], newApp: AppDefinition): AppDefinition = {
       val newVersionInfo = maybeOldApp match {
         case None =>
-          log.info(s"[${newApp.id}]: new app detected")
+          log.info(s"${newApp.id}: new app detected")
```
unterstein: I would keep logging statements stable as long as there is no additional value in changing them. Probably we do have users/customers running regex on those logs and maybe they need `[` and `]`. But I understand the wish to align the messages with others...I don`t have strong feelings about one or the other solution, maybe the other reviewers? :)

zen-dog: This is the only place we use brackets `[]`. Anyway, "keeping logs stable" is not part of our contract with customers - otherwise this argument can be applicable to any other change in this PR ;)

src/main/scala/mesosphere/marathon/core/deployment/DeploymentPlan.scala
zen-dog: What do you mean? I kept `plan.id` everywhere.

unterstein: ... well, you are right. I`m sorry 🐼😪

src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
unterstein: Any other opinions? 📦 

// I`m ok with a merge 😘

PR #5655, opened 2017-10-19
Delete tasks in TasksController

alenkacz:
Delete tasks in TasksController

Summary:
Re-implemented delete tasks endpoint in TasksController using Akka.HTTP.

JIRA issues: MARATHON-7308


src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
     }
   }
 
+  private def deleteTasks()(implicit identity: Identity): Route = {
+    def doKill(force: Boolean, scale: Boolean, wipe: Boolean, tasksIdToAppId: Map[Id, PathId]) = {
+      def isAuthorized(tasksIdToAppId: Map[Id, PathId]): Boolean = tasksIdToAppId.values.exists(id => !authorizer.isAuthorized(identity, UpdateRunSpec, id))
+
+      // TODO authorization check is performed twice - in the taskKiller and also here, this one must remain
+      if (isAuthorized(tasksIdToAppId)) {
```
alenkacz: this is a change I introduced in comparison to the old implementation. For all kill methods, authorization is checked in TaskKiller. For kill (not kill and scale) some time ago, duplicit authorization was introduced because we need to check authorization for all tasks before we start killing anything.

The change I introduced (which I hope should be fine) is that the authorization is checked in advance also for kill and scale (not just for kill).

src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpMarathonService.scala
```diff
@@ -97,6 +96,9 @@ object AkkaHttpMarathonService {
       .handle(EntityMarshallers.handleNonValid)
       .handle(AuthDirectives.handleAuthRejections)
       .handle {
```
meln1k: What do you thing about this? 
```scala
.handle {
  case Rejections.BadRequest(message) => complete(BadRequest -> message)
  case Rejections.EntityNotFound(message) => complete(NotFound -> message)
}
```

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
...
+        }
+      }
+    }
+
+    (entity(as[TasksToDelete])
+      & parameter('force.as[Boolean].?(false))
+      & parameter('scale.as[Boolean].?(false))
+      & parameter('wipe.as[Boolean].?(false))) { (taskIds, force, scale, wipe) =>
+        if (scale && wipe) {
+          reject(Rejections.BadRequest(Message("You cannot use scale and wipe at the same time.")))
```
meln1k: There is an built-in akka-http rejection for this. If you want to use it, the code would look like this:
```scala
reject(MalformedQueryParamRejection("scale, wipe", "You cannot use scale and wipe at the same time."))
```

src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpMarathonService.scala
alenkacz: better, thanks

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: used the built in one, thanks

docs/docs/rest-api/public/api/v2/types/enrichedTask.raml
```diff
@@ -45,4 +45,3 @@ types:
   TaskStatusCondition:
     type: string
     enum: [ running, staging ]
-
```
ichernetsky: Please add a newline here.

src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpMarathonService.scala
kensipe: nice!

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -48,7 +53,10 @@ class TasksController(
           pathEndOrSingleSlash {
             listTasks()
           }
-        }
+        } ~
+          (path("delete") & post) {
```
jeschkies: Out of curiosity, why did you opt for `(path("delete") & post)` instead of `post { path("delete") }` as done for `get`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
jeschkies: Is this still a todo then? What needs to be done?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
...
+  private def deleteTasks()(implicit identity: Identity): Route = {
+    def doKill(force: Boolean, scale: Boolean, wipe: Boolean, tasksIdToAppId: Map[Id, PathId]) = {
+      def isAuthorized(tasksIdToAppId: Map[Id, PathId]): Boolean = tasksIdToAppId.values.exists(id => !authorizer.isAuthorized(identity, UpdateRunSpec, id))
+
+      // TODO authorization check is performed twice - in the taskKiller and also here, this one must remain
+      if (isAuthorized(tasksIdToAppId)) {
+        reject(AuthDirectives.NotAuthorized(HttpPluginFacade.response(authorizer.handleNotAuthorized(identity, _))))
+      } else {
+        val maybeInstances: Future[Iterable[Option[Instance]]] = Future.sequence(tasksIdToAppId.view
+          .map { case (taskId, _) => instanceTracker.instancesBySpec.map(_.instance(taskId)) })
```
jeschkies: How about `collect` which filters out `None`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
...
+        val tasksByAppId: Future[Map[PathId, Seq[Instance]]] = maybeInstances.map(i => i.flatten
+          .groupBy(instance => instance.instanceId.runSpecId)
+          .map { case (appId, instances) => appId -> instances.to[Seq] }(collection.breakOut))
+
+        if (scale) {
+          onSuccess(killAndScale(tasksByAppId, force)) { deploymentResult =>
```
jeschkies: What type is `deploymentResult`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
     }
   }
 
+  private def deleteTasks()(implicit identity: Identity): Route = {
+    def doKill(force: Boolean, scale: Boolean, wipe: Boolean, tasksIdToAppId: Map[Id, PathId]) = {
```
jeschkies: What do you think about moving this out? It seems to be easier to oversee `deleteTasks` then.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
...
+    }
+
+    (entity(as[TasksToDelete])
+      & parameter('force.as[Boolean].?(false))
+      & parameter('scale.as[Boolean].?(false))
+      & parameter('wipe.as[Boolean].?(false))) { (taskIds, force, scale, wipe) =>
+        if (scale && wipe) {
+          reject(MalformedQueryParamRejection("scale, wipe", "You cannot use scale and wipe at the same time."))
+        } else {
+          val maybeTasksIdToAppId: Try[Map[Id, PathId]] = Try(taskIds.ids.map { id =>
```
jeschkies: Out of curiosity, what is the benefit of `Try` here?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -62,6 +70,53 @@ class TasksController(
...
+            try { Task.Id(id).instanceId -> Task.Id.runSpecId(id) }
```
jeschkies: Isn't `Try` supposed to catch this?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -108,8 +163,43 @@ class TasksController(
     case _ => None
   }
 
+  @SuppressWarnings(Array("all")) // async/await
+  private def doKillTasks(toKillFuture: Future[Map[PathId, Seq[Instance]]], tasksIdToAppId: Map[Id, PathId], wipe: Boolean)(implicit identity: Identity): Future[Seq[EnrichedTask]] = async {
```
jeschkies: Could you add docs? It is not quite clear from the method names how this differs to `doKill` and `deleteTasks`.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -108,8 +163,43 @@ class TasksController(
...
+    val toKill = await(toKillFuture)
+    val affectedApps = tasksIdToAppId.values.flatMap(appId => groupManager.app(appId))(collection.breakOut)
+
+    val killedTasks = await(Future.sequence(toKill
+      .filter { case (appId, _) => affectedApps.exists(app => app.id == appId) }
```
jeschkies: Why do we have to filter here again? It's a little hard to follow how we come up with the right tasks to kill. So `toKill` gives a map of app ids (`PathId`) to a sequence of instances. Why do we require `tasksIdToAppId` then?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: I talked with the author of the original todo and it was meant to point out a potential performance issue with the fact that the authorization is checked twice (once in controller and then somewhere deep in task killer). So I kept the comment in place if we ever want to refactor it (but it was totally out of scope of this PR)

The change that I made on top what was there before is that now the check is done twice for kill as well as killAndScale (previously it was for some reason just kill)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: Good question - It was the pattern that I have seen in few other controllers so I just stick with it. I also don't like deep nesting...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: the none comes from this very last call _.instance(taskId) and the rest of the map is actually mapping a future so I think it is not possible to use collect here... feel free to correct me :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: DeploymentResult :) that is derived from DeploymentPlan in killAndScale method

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: since I am not mapping/transforming the Try in any way, the benefit is probably not that big. I just generally use Try object instead of oldschool try/catch everywhere, I prefer this "more functional style". If it does not conform to the coding style of the rest of the codebase, just tell me and I'll change it

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: I struggled with this part of the code so much :( I tried to rewrite it many times, this is the best I could get. Maybe it would be nice to pair over this or if you have any idea from the top of your head just suggest it...

the problem is, that this inner try is there just for the purpose of persisting the "id" of the task where the exception happened. If I catched it on the outer level, then I wouldn't know which input task id it was...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: Added docs and did some renaming. Let me know if it did not help.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: I tried to refactor it but it was still not enough (I agree). I will try to dive into this deeper once more

alenkacz: Sooo ... I found out this filter was introduced by @unterstein so I will go over the code with him next week. But I think it is not blocker of merging this patch because this code is there even in the taskResource so it should be working...

alenkacz:
@jeschkies @meln1k ready for re-review. I did a little bit of refactoring, now the flow of the whole method should be a little bit better (I think).

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
jeschkies: Ok, let's not touch it in this diff then :).

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -108,8 +167,51 @@ class TasksController(
...
+      .map { case (appId, instances) => taskKiller.kill(appId, _ => instances, wipe) }))
+      .flatten
+
+    killedTasks.flatMap { instance =>
+      instance.tasksMap.valuesIterator.map { task =>
+        EnrichedTask(task.runSpecId, task, instance.agentInfo, Seq.empty)
```
meln1k: Haven't we introduce a breaking change in the API here?

alenkacz: Just if someone finds this comment - it looks like apps/delete returns a different json than /delete ... so this change is fine, but the API sucks :)

PR #5656, opened 2017-10-19
Fix flaky `EventsControllerTest`

timcharper:
Stream was getting closed due to leader loss before first element could be published.

Resolved by waiting for first element, _then_ publish leader loss event.

Passed 1000 times in a row now. Seems deterministic.

JIRA Issues: MARATHON-7317


src/test/scala/mesosphere/marathon/api/akkahttp/v2/EventsControllerTest.scala
```diff
@@ -123,10 +121,14 @@ class EventsControllerTest extends AkkaUnitTest with Inside {
 
     val events = response.entity.dataBytes
       .via(EventStreamParser(Int.MaxValue, Int.MaxValue))
-      .runWith(Sink.seq)
+      .map(_.`type`)
+      .runWith(Sink.queue())
 
-    f.leaderStateEventsInput.offer(LocalLeadershipEvent.Standby)
+    inside(events.pull().futureValue) {
+      case Some(Some(tpe)) => tpe shouldBe "event_stream_attached"
```
zen-dog: s/tpe/type ?

PR #5658, opened 2017-10-19
 Update MoM-EE image to `v1.5.1_1.10.0` in SI tests

zen-dog:
Summary:
- Update MoM-EE image to `v1.5.1_1.10.0` in SI tests. Replaced `max_tasks_per_offer` parameter with `max_instances_per_offer` for 1.5 MoM-EE definitions

### DIFF
--- a/tests/system/test_marathon_on_marathon_ee.py
+++ b/tests/system/test_marathon_on_marathon_ee.py
@@ -26,7 +26,7 @@ PUBLIC_KEY_FILE = 'public-key.pem'
 DEFAULT_MOM_IMAGES = {
     'MOM_EE_1.4': 'v1.4.7_1.9.8',
     'MOM_EE_1.3': '1.3.13_1.1.5',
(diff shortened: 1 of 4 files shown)


zen-dog:
I tested it manually on a strict cluster and it works.

tests/system/test_marathon_on_marathon_ee.py
```diff
@@ -25,7 +25,8 @@
 
 DEFAULT_MOM_IMAGES = {
     'MOM_EE_1.4': 'v1.4.7_1.9.8',
-    'MOM_EE_1.3': '1.3.13_1.1.5'
+    'MOM_EE_1.3': '1.3.13_1.1.5',
+    'MOM_EE_1.5': 'v1.5.1_1.10.0'
```
ichernetsky: Why not keeping the descending order?

kensipe: what ivan said!

PR #5659, opened 2017-10-19
Migrate UnreachableStrategy to migrate saved and running Instances as well

unterstein:
Summary:

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-7790

### DIFF
--- a/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
+++ b/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
@@ -13,14 +13,14 @@ import mesosphere.marathon.storage.repository.InstanceRepository
 import scala.concurrent.{ ExecutionContext, Future }
 
 @SuppressWarnings(Array("ClassNames"))
-class MigrationTo149(instanceRepository: InstanceRepository)(implicit ctx: ExecutionContext, mat: Materializer) extends StrictLogging {
+class MigrationTo152(instanceRepository: InstanceRepository)(implicit ctx: ExecutionContext, mat: Materializer) extends StrictLogging {
 
   def migrate(): Future[Done] = {
-    MigrationTo149.migrateUnreachableInstances(instanceRepository)(Environment(sys.env), ctx, mat)
+    MigrationTo152.migrateUnreachableInstances(instanceRepository)(Environment(sys.env), ctx, mat)
   }
 }
 
-object MigrationTo149 extends StrictLogging {
+object MigrationTo152 extends StrictLogging {
 
   private def changeUnreachableStrategyForInstances(instance: Instance): Instance = {
     instance.copy(unreachableStrategy = MigrationTo146.changeUnreachableStrategy(instance.unreachableStrategy))
--- a/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
+++ b/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
@@ -27,7 +27,7 @@ object MigrationTo152 extends StrictLogging {
   }
 
   def migrateUnreachableInstances(instanceRepository: InstanceRepository)(implicit env: Environment, ctx: ExecutionContext, mat: Materializer): Future[Done] = {
-    logger.info("Starting unreachable strategy migration to 1.4.9")
+    logger.info("Starting unreachable strategy migration to 1.5.2")
 
     val instanceSink =
       Flow[Instance]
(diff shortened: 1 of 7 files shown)


unterstein:
This PR should also be backported to `releases/1.4`

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
```diff
@@ -0,0 +1,61 @@
...
+  private def changeUnreachableStrategyForInstances(instance: Instance): Instance = {
+    instance.copy(unreachableStrategy = MigrationTo146.changeUnreachableStrategy(instance.unreachableStrategy))
+  }
+
+  def migrateUnreachableInstances(instanceRepository: InstanceRepository)(implicit env: Environment, ctx: ExecutionContext, mat: Materializer): Future[Done] = {
+    logger.info("Starting unreachable strategy migration to 1.4.9")
+
+    val instanceSink =
+      Flow[Instance]
+        .mapAsync(Int.MaxValue)(instanceRepository.store)
```
meln1k: Wow, that's huge!

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+object MigrationTo152 extends StrictLogging {
+
+  private def changeUnreachableStrategyForInstances(instance: Instance): Instance = {
+    instance.copy(unreachableStrategy = MigrationTo146.changeUnreachableStrategy(instance.unreachableStrategy))
+  }
+
+  def migrateUnreachableInstances(instanceRepository: InstanceRepository)(implicit env: Environment, ctx: ExecutionContext, mat: Materializer): Future[Done] = {
+    logger.info("Starting unreachable strategy migration to 1.4.9")
```
meln1k: Why it says 1.4.9 when migration is called 1.5.2?

src/test/scala/mesosphere/marathon/storage/migration/MigrationTo152Test.scala
```diff
@@ -0,0 +1,53 @@
...
+import mesosphere.marathon.storage.migration.MigrationTo146.Environment
+import mesosphere.marathon.storage.repository.InstanceRepository
+import mesosphere.marathon.test.GroupCreation
+
+import scala.concurrent.duration._
+import scala.concurrent.{ ExecutionContextExecutor, Future }
+
+class MigrationTo152Test extends AkkaUnitTest with GroupCreation with StrictLogging {
+
+  "Migration to 1.4.9" should {
```
meln1k: 1.4.9 or 1.5.2? :)

unterstein:
thanks @meln1k, my head was already at the backport :)

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
```diff
@@ -0,0 +1,61 @@
...
+  private def changeUnreachableStrategyForInstances(instance: Instance): Instance = {
+    instance.copy(unreachableStrategy = MigrationTo146.changeUnreachableStrategy(instance.unreachableStrategy))
+  }
+
+  def migrateUnreachableInstances(instanceRepository: InstanceRepository)(implicit env: Environment, ctx: ExecutionContext, mat: Materializer): Future[Done] = {
+    logger.info("Starting unreachable strategy migration to 1.5.2")
+
+    val instanceSink =
+      Flow[Instance]
+        .mapAsync(Int.MaxValue)(instanceRepository.store)
```
meln1k: Do you think it makes sense to restrict amount of simultaneous calls to some reasonable default? Let's say 8 for example (must common number of CPUs * 2).

jeschkies: +1 `Int.MaxValue` seem to big.

unterstein:
@jeschkies @meln1k:
The usage of `Int.MaxValue` is aligned with implementation for 1.4.2 migration and 1.4.6 migration and I don't think it does really matter, because we are operating in memory anyways.
If you go down the path to the storage layer, you will find a `WorkQueue` in `BasePersistenceStore` to queue all update requests for a single appId and if you go further down the path, you will find a `WorkQueue` in the `ZkPersistenceStore` to queue up all zookeeper requests. In `ZkPersistenceStore` a maxConcurrency of 8 is configured.

I don't have a strong opinion on changing this to 8 or whatever number, but I would question if another queue would help and bring add value at this point.
If we decide to go with a change, we should change the other migrations as well.

meln1k:
@unterstein Personally I'd prefer to be extra safe, but if you're sure that `Int.MaxValue` won't cause anything bad - go for it.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo152.scala
ichernetsky: 👍

unterstein:
@meln1k I`m going to merge this and open a follow up to change all migrations using streams to have max concurrency to 8. Who knows when we change the storage layer again and how this will be implemented.

PR #5662, opened 2017-10-19
Store a flag node in ZK to indicate that Marathon performs data migration

ichernetsky:
Summary:
 - Store a flag node in ZK to indicate that Marathon performs data migration
 - Delete the flag if migration gets cancelled
 - An integration test to ensure that Marathon fails to start in presence of
   the migration flag

It is a backport of #5610

JIRA issues: MARATHON-7788

### DIFF
--- a/src/main/scala/mesosphere/marathon/Exception.scala
+++ b/src/main/scala/mesosphere/marathon/Exception.scala
@@ -89,3 +89,11 @@ class ResolveArtifactsCanceledException(msg: String) extends DeploymentFailedExc
 class StoreCommandFailedException(msg: String, cause: Throwable = null) extends Exception(msg, cause)
 @SuppressWarnings(Array("NullAssignment"))
 class MigrationFailedException(msg: String, cause: Throwable = null) extends Exception(msg, cause)
+
+/**
+  * Instances of this exception class are expected to be used to cancel
+  * an on-going migration. It is imperative to throw such an exception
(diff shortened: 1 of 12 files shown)


ichernetsky:
Yes, you are correct, @kensipe It is stored under `state`, therefore it gets automatically deleted if a user does a restore after a failed migration.

kopax:
Hi guys, this is super frustrating, I have created this: https://jira.mesosphere.com/browse/MARATHON-7972

I am running marathon `1.4.9-1.0.668.ubuntu1604`, can you please tell me if this will save my time and life?

jeschkies:
Could you update to a newer Marathon 1.4 before going to 1.5? Maybe that works.

kopax:
I am usine 1.4 already

ichernetsky:
@kopax, 1.4.13 is the latest 1.4 release. @jeschkies advises to upgrade to it first, before going to 1.5 and so on.

kopax:
Will the `/migration-in-progress` disappear in 1.5 ? Because that's a good motivation to upgrade.

ichernetsky:
The concept of it is kept in 1.5 too. The purpose of this flag is to be able to detect if a previous migration attempt failed for some reason (interruption, exception and so forth). If it exists, it means that something went wrong during the previous attempt. In such a case, it is recommended that Marathon state should be restored from a backup, and upgrade is retried. Does it make sense, @kopax?

PR #5665, opened 2017-10-19
List tasks as txt

alenkacz:
List tasks as txt

Summary:
Last part of TasksController rewrite using Akka.HTTP. Prints out tasks in plaintext.

JIRA issues: MARATHON-7308

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
@@ -2,7 +2,8 @@ package mesosphere.marathon
 package api.akkahttp
 
 import akka.http.scaladsl.model.headers._
-import akka.http.scaladsl.model.{ DateTime, HttpHeader, HttpMethods, HttpProtocols }
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.server.Directives.extractRequest
 import akka.http.scaladsl.server.{ Directive, Directive0, Route, Directives => AkkaDirectives }
 import com.wix.accord.{ Failure, Success, Result => ValidationResult }
 
@@ -40,6 +41,24 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
       headers += `Access-Control-Max-Age`(1.day.toSeconds)
 
       respondWithHeaders (headers.result())
+    }
+  }
+
+  def accepts(mediaType: MediaType): Directive0 = {
+    extractRequest.flatMap { request =>
+      request.header[Accept] match {
+        case Some(accept) if accept.mediaRanges.exists(range => range.matches(mediaType)) => pass
+        case _ => reject
+      }
+    }
+  }
+
+  def acceptsAnything: Directive0 = {
+    extractRequest.flatMap { request =>
+      request.header[Accept] match {
+        case None => pass
+        case _ => reject
+      }
     }
   }
 
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
@@ -2,10 +2,16 @@ package mesosphere.marathon
 package api.akkahttp.v2
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.marshalling.{ Marshaller, ToEntityMarshaller }
+import akka.http.scaladsl.model.headers.Accept
+import akka.http.scaladsl.model.{ HttpResponse, MediaRange, MediaTypes, StatusCodes }
 import akka.http.scaladsl.server.Route
 import akka.stream.scaladsl.Source
 import akka.stream.Materializer
+import akka.http.scaladsl.model.MediaTypes.`text/plain`
 import com.typesafe.scalalogging.StrictLogging
+import mesosphere.marathon.api.EndpointsHelper
+import mesosphere.marathon.api.EndpointsHelper.ListTasks
 import mesosphere.marathon.api.akkahttp.Controller
 import mesosphere.marathon.api.akkahttp.Directives.pathEndOrSingleSlash
 import mesosphere.marathon.core.appinfo.EnrichedTask
@@ -17,8 +23,9 @@ import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.instance.Instance.Id
 import mesosphere.marathon.core.task.tracker.InstanceTracker
 import mesosphere.marathon.plugin.auth.{ Authenticator, _ }
-import play.api.libs.json.Json
+import play.api.libs.json.{ JsString, Json }
 import mesosphere.marathon.raml.{ AnyToRaml, EnrichedTasksList, Writes }
+import mesosphere.marathon.stream.Implicits._
 import mesosphere.marathon.state.PathId
 
 import scala.async.Async._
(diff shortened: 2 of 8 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -43,6 +44,24 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
     }
   }
 
+  def accepts(mediaType: MediaType): Directive0 = {
```
jeschkies: Could you be so nice and use `accept` here https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala#L37 as well? Thanks a ton.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -45,15 +52,32 @@ class TasksController(
     asLeader(electionService) {
       authenticated.apply { implicit identity =>
         get {
-          pathEndOrSingleSlash {
-            listTasks()
+          (accepts(MediaTypes.`text/plain`) & pathEndOrSingleSlash) {
```
meln1k: Looks good, but do you think we can do the same thing using marshallers? It should be possible to define one marshaller based on several others using `Marshaller.oneOf`, and akka will do the rest of content negotiation.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -45,15 +52,32 @@ class TasksController(
...
+            listTasksJson() // when no accept header present, json is the default choice
           }
         }
       }
     }
   }
 
-  private def listTasks()(implicit identity: Identity): Route = {
+  @SuppressWarnings(Array("all")) // async/await
+  private def listTasksTxt()(implicit identity: Identity): Route = {
```
jeschkies: How come this method differs so much from `listTasksJson`?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
```diff
@@ -161,6 +162,63 @@ class TasksControllerTest extends UnitTest with ScalatestRouteTest with Inside w
...
+      val app = AppDefinition("/foo".toRootPath, portDefinitions = Seq(PortDefinition(0), PortDefinition(0)))
+      val instance = TestInstanceBuilder.newBuilder(app.id).addTaskRunning().getInstance()
+
+      val tasksByApp = InstanceTracker.InstancesBySpec.forInstances(instance)
+      instanceTracker.instancesBySpec returns Future.successful(tasksByApp)
+
+      val rootGroup = createRootGroup(apps = Map(app.id -> app))
+      groupManager.rootGroup() returns rootGroup
+
+      assert(app.servicePorts.size > instance.appTask.status.networkInfo.hostPorts.size)
```
jeschkies: Please use the `should` matcher.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
```diff
@@ -161,6 +162,63 @@ class TasksControllerTest extends UnitTest with ScalatestRouteTest with Inside w
...
+
+      When("Getting the txt tasks index")
+      Get(Uri./).addHeader(Accept(MediaTypes.`text/plain`)) ~> controller.route ~> check {
+        Then("The status should be 200")
+        status should be(StatusCodes.OK)
+
+        responseAs[String] should include (s"${app.id.safePath}\t${app.servicePorts.head}")
```
jeschkies: Could we be explicit here? I think that's easier to understand.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
```diff
@@ -161,6 +162,63 @@ class TasksControllerTest extends UnitTest with ScalatestRouteTest with Inside w
...
+        status should be(StatusCodes.OK)
+
+        responseAs[String] should include (s"${authorizedApp.id.safePath}")
+        responseAs[String] should not include (s"${notAuthorizedApp.id.safePath}")
+      }
+    }
+
+    {
+      val controller = Fixture(authenticated = false).controller
+      behave like unauthenticatedRoute(forRoute = controller.route, withRequest = Get(Uri./).addHeader(Accept(MediaTypes.`text/plain`)))
```
jeschkies: How about

```scala
behave like unauthenticatedRoute(forRoute = controller.route, withRequest = Get(Uri./).addHeader(Accept(MediaTypes.`application/json`)))
```

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
```diff
@@ -161,6 +162,63 @@ class TasksControllerTest extends UnitTest with ScalatestRouteTest with Inside w
...
+
+      val rootGroup = createRootGroup(apps = Map(authorizedApp.id -> authorizedApp, notAuthorizedApp.id -> notAuthorizedApp))
+      f.groupManager.rootGroup() returns rootGroup
+
+      When("Getting the txt tasks index")
+      Get(Uri./).addHeader(Accept(MediaTypes.`text/plain`)) ~> f.controller.route ~> check {
+        Then("The status should be 200")
+        status should be(StatusCodes.OK)
+
+        responseAs[String] should include (s"${authorizedApp.id.safePath}")
```
jeschkies: Please use explicit assertions here. That is easier to understand.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: well I thought about it but the thing is that the data from which one or the other response is created from is not the same - for json endpoint there is much more data fetched so the underlying model it different. That's why I did it this way, not the marshaller way. Does it make sense... ?

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
alenkacz: yes, I was actually planning to do that but in a separate PR once this one is approved (if you liked this approach - as it looks now you do :))

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: I would say that much less data is actually serialized to the text output than to json ... so even the processing and preparation of the data is a little bit different...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
alenkacz: what do you mean by being explicit?

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
meln1k: Well, in general we don't need to explicitly filter accept headers as long as we have a proper marshaller. Akka http can do it transparently

alenkacz: I can try to rewrite ping using marshallers. I think there it should work...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
alenkacz: added

alenkacz:
@meln1k @jeschkies should be ready for re-review. Thanks!

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
jeschkies: I'd stay explicit here. I know that Akka should be able to do so. However, our API unfortunately does not seem to follow those rules exactly.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
jeschkies: Ok, thanks for clearing this up.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/TasksControllerTest.scala
```diff
@@ -135,12 +137,16 @@ class TasksControllerTest extends UnitTest with ScalatestRouteTest with Inside w
       behave like unauthenticatedRoute(forRoute = controller.route, withRequest = Get(Uri./))
     }
 
+    {
+      val controller = Fixture(authenticated = false).controller
+      behave like unauthenticatedRoute(forRoute = controller.route, withRequest = Get(Uri./).addHeader(Accept(MediaTypes.`application/json`)), " using text/json")
```
jeschkies: Why not move it into the same block as 
```scala
behave like unauthenticatedRoute(forRoute = controller.route, withRequest = Get(Uri./).addHeader(Accept(MediaTypes.`text/plain`)))
```

PR #5666, opened 2017-10-19
Switch to use Debian Slim base image

timcharper:
this shaves off 40MB from our base image, and reduces the number of
dependencies we pull in (limiting the number of CVE vulnerabilities that
show up in image scans)

JIRA Issues: MARATHON-7852


timcharper:
Needs backport

I built Marathon and pushed docker tag v1.5.1-1 based on this patch so we could get a scan and see if there is any improvement for CVE vulnerabilities.

PR #5670, opened 2017-10-20
Migrate UnreachableStrategy to migrate saved and running Instances as well (#5659)

unterstein:
Summary:

JIRA issues: MARATHON-7790

backport of #5659

### DIFF
--- a/src/main/scala/mesosphere/marathon/storage/migration/legacy/MigrationTo_1_4_9.scala
+++ b/src/main/scala/mesosphere/marathon/storage/migration/legacy/MigrationTo_1_4_9.scala
@@ -1,5 +1,4 @@
-package mesosphere.marathon
-package storage.migration
+package mesosphere.marathon.storage.migration.legacy
 
 import akka.Done
 import akka.stream.Materializer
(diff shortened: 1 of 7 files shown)


src/main/scala/mesosphere/marathon/storage/migration/legacy/MigrationTo_1_4_9.scala
```diff
@@ -0,0 +1,60 @@
...
+  private def changeUnreachableStrategyForInstances(instance: Instance): Instance = {
+    instance.copy(unreachableStrategy = MigrationTo_1_4_6.changeUnreachableStrategy(instance.unreachableStrategy))
+  }
+
+  def migrateUnreachableInstances(instanceRepository: InstanceRepository)(implicit env: Environment, ctx: ExecutionContext, mat: Materializer): Future[Done] = {
+    logger.info("Starting unreachable strategy migration to 1.4.9")
+
+    val instanceSink =
+      Flow[Instance]
+        .mapAsync(Int.MaxValue)(instanceRepository.store)
```
meln1k: In my view `.mapAsync(Int.MaxValue)` looks like we don't want to have any backpressure :) Do you think it makes sense to change `Int.MaxValue` to a lower value?

unterstein: Migrating discussion to https://github.com/mesosphere/marathon/pull/5659#issuecomment-338158484

PR #5673, opened 2017-10-20
Default behaviour for workload bursting

meln1k:
Summary: when there are no region/zone related constraints, we should accept offers only from the local region

JIRA issues: MARATHON_EE-1695

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -7,6 +7,7 @@ import mesosphere.marathon.core.base._
 import mesosphere.marathon.core.event.{ SchedulerRegisteredEvent, _ }
 import mesosphere.marathon.core.launcher.OfferProcessor
 import mesosphere.marathon.core.task.update.TaskStatusUpdateProcessor
+import mesosphere.marathon.state.FaultDomain
 import mesosphere.marathon.storage.repository.FrameworkIdRepository
 import mesosphere.marathon.stream.Implicits._
 import mesosphere.marathon.util.SemanticVersion
@@ -30,7 +31,7 @@ class MarathonScheduler(
   private[this] val log = LoggerFactory.getLogger(getClass.getName)
 
   private var lastMesosMasterVersion: Option[SemanticVersion] = Option.empty
-  @volatile private[this] var homeRegion: Option[String] = Option.empty
+  @volatile private[this] var homeFaultDomain: Option[FaultDomain] = Option.empty
 
   import mesosphere.marathon.core.async.ExecutionContexts.global
 
@@ -42,7 +43,7 @@ class MarathonScheduler(
     master: MasterInfo): Unit = {
     log.info(s"Registered as ${frameworkId.getValue} to master '${master.getId}'")
     masterVersionCheck(master)
-    updateHomeRegion(master)
+    updateHomeFaultDomain(master)
     Await.result(frameworkIdRepository.store(FrameworkId.fromProto(frameworkId)), zkTimeout)
     mesosLeaderInfo.onNewMasterInfo(master)
     eventBus.publish(SchedulerRegisteredEvent(frameworkId.getValue, master.getHostname))
@@ -51,7 +52,7 @@ class MarathonScheduler(
   override def reregistered(driver: SchedulerDriver, master: MasterInfo): Unit = {
     log.info("Re-registered to %s".format(master))
     masterVersionCheck(master)
-    updateHomeRegion(master)
+    updateHomeFaultDomain(master)
     mesosLeaderInfo.onNewMasterInfo(master)
     eventBus.publish(SchedulerReregisteredEvent(master.getHostname))
   }
@@ -148,11 +149,14 @@ class MarathonScheduler(
     }
   }
 
-  protected def updateHomeRegion(masterInfo: MasterInfo): Unit = {
+  protected def updateHomeFaultDomain(masterInfo: MasterInfo): Unit = {
     if (masterInfo.hasDomain && masterInfo.getDomain.hasFaultDomain) {
-      homeRegion = Some(masterInfo.getDomain.getFaultDomain.getRegion.getName)
+      homeFaultDomain = Some(FaultDomain(
+        masterInfo.getDomain.getFaultDomain.getRegion.getName,
+        masterInfo.getDomain.getFaultDomain.getZone.getName
+      ))
     } else {
-      homeRegion = None
+      homeFaultDomain = None
     }
   }
 
@@ -160,10 +164,10 @@ class MarathonScheduler(
   def mesosMasterVersion(): Option[SemanticVersion] = lastMesosMasterVersion
 
   /**
-    * Current home region of the mesos master
-    * @return name of the region if it's available, None otherwise
+    * Current home fault domain of the mesos master
+    * @return fault domain if it's available, None otherwise
     */
-  def getHomeRegion: Option[String] = homeRegion
+  def getHomeFaultDomain: Option[FaultDomain] = homeFaultDomain
 
   /**
     * Exits the JVM process, optionally deleting Marathon's FrameworkID
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -7,7 +7,7 @@ import mesosphere.marathon.core.base._
 import mesosphere.marathon.core.event.{ SchedulerRegisteredEvent, _ }
 import mesosphere.marathon.core.launcher.OfferProcessor
 import mesosphere.marathon.core.task.update.TaskStatusUpdateProcessor
-import mesosphere.marathon.state.FaultDomain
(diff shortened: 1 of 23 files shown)


src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
```diff
@@ -67,16 +72,16 @@ class ActorOfferMatcherTest extends AkkaUnitTest {
...
       Offer.newBuilder()
         .setId(MesosOfferIdProto.newBuilder().setValue(UUID.randomUUID().toString))
         .setFrameworkId(FrameworkID.newBuilder().setValue("notanidframework"))
         .setSlaveId(SlaveID.newBuilder().setValue(s"slave1"))
         .setHostname("hostname")
         .setDomain(DomainInfo.newBuilder()
-          .setFaultDomain(FaultDomain.newBuilder()
-            .setRegion(RegionInfo.newBuilder().setName(faultRegion))
-            .setZone(ZoneInfo.newBuilder().setName("zone")))
+          .setFaultDomain(FaultDomainPB.newBuilder()
```
zen-dog: nit: dislike the `PB` in `FaultDomainPB`. May be `Protos.FaultDomain` or similar?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -133,7 +133,7 @@ object ResourceMatcher extends StrictLogging {
     * the reservation.
     */
   def matchResources(offer: Offer, runSpec: RunSpec, knownInstances: => Seq[Instance],
-    selector: ResourceSelector, conf: MatcherConf, schedulerPlugins: Seq[SchedulerPlugin] = Seq.empty)(implicit clock: Clock): ResourceMatchResponse = {
+    selector: ResourceSelector, conf: MatcherConf, schedulerPlugins: Seq[SchedulerPlugin] = Seq.empty, homeFaultDomain: Option[FaultDomain])(implicit clock: Clock): ResourceMatchResponse = {
```
zen-dog: May be give it a default of `None` and avoid passing it altogether in tests later? Makes for a lot less changed lines.

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
```diff
@@ -29,13 +29,6 @@ class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[P
 
   override def toString: String = s"ActorOfferMatcher($actorRef)"
 
-  override def isInterestedIn(offer: Offer) = {
```
zen-dog: Where did this method go?

meln1k: It's true by default, and we need exactly this

zen-dog: If it's `true` in the trait and we never override it - do we really need it?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
alenkacz: 👍

src/main/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcher.scala
```diff
@@ -18,7 +18,7 @@ import scala.concurrent.{ Future, Promise }
   * @param actorRef Reference to actor that matches offers.
   * @param precedenceFor Defines which matcher receives offers first. See [[mesosphere.marathon.core.matcher.base.OfferMatcher.precedenceFor]].
   */
-class ActorOfferMatcher(actorRef: ActorRef, override val precedenceFor: Option[PathId], homeRegion: () => Option[String])
```
alenkacz: I think the whole homeFaultDomain is not needed here anymore ... ?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -176,6 +176,12 @@ object ResourceMatcher extends StrictLogging {
 
     def portsMatchOpt: Option[PortsMatch] = PortsMatcher(runSpec, offer, selector).portsMatch
 
+    val meetsFaultDomainRequirements: Boolean = {
+      val faultDomainFields = Set(Constraints.regionField, Constraints.zoneField)
+      val hasFaultDomainConstraints = runSpec.constraints.exists(c => faultDomainFields.contains(c.getField))
+      hasFaultDomainConstraints || homeFaultDomain.map(_.region) == OfferUtil.region(offer)
```
jeschkies: Hm, I'm not such a big fan of comparing options. How about

```scala
homeFaultDomain.exists( domain => OfferUtil.region(offer).contains(domain.region) )
```

I don't have a strong opinion though.

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
jeschkies: +1

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
jeschkies: +1 this would make things a little simpler :).

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
meln1k: Well, it will be a bit wordy: `Protos.DomainInfo.FaultDomain`

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
meln1k: I agree that it would make it simpler, but in this case I'd make it explicit to not miss a single important part. If it looks too scary, can change it for sure :)

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -148,22 +149,24 @@ class MarathonScheduler(
     }
   }
 
-  protected def updateHomeRegion(masterInfo: MasterInfo): Unit = {
```
zen-dog: I would save `MasterInfo` completely providing helper methods like `getHomeRegion` to access parts of it.

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
```diff
@@ -1,46 +1,18 @@
...
 import mesosphere.AkkaUnitTest
 import mesosphere.marathon.core.matcher.base.OfferMatcher.MatchedInstanceOps
 import mesosphere.marathon.test.MarathonTestHelper
 import mesosphere.mesos.protos.Implicits._
 import mesosphere.mesos.protos.OfferID
-import org.apache.mesos.Protos.DomainInfo.FaultDomain
-import org.apache.mesos.Protos.DomainInfo.FaultDomain.{ RegionInfo, ZoneInfo }
-import org.apache.mesos.Protos.{ DomainInfo, FrameworkID, Offer, SlaveID, OfferID => MesosOfferIdProto }
 
 class ActorOfferMatcherTest extends AkkaUnitTest {
```
zen-dog: I understand that we removed `isInterestedIn` method but we should still unit test the use cases below.

src/main/scala/mesosphere/marathon/core/matcher/manager/OfferMatcherManagerModule.scala
```diff
@@ -41,6 +41,6 @@ class OfferMatcherManagerModule(
     * offers.
     */
   val globalOfferMatcherWantsOffers: Observable[Boolean] = offersWanted
-  val globalOfferMatcher: OfferMatcher = new ActorOfferMatcher(offerMatcherMultiplexer, None, homeRegion)
+  val globalOfferMatcher: OfferMatcher = new ActorOfferMatcher(offerMatcherMultiplexer, None)
```
jeschkies: Why won't we pass the region anymore? Is this a bug?

jeschkies: I'm probably missing the obvious but it seems to me we are not matching the region anymore. Where is that done?

src/test/scala/mesosphere/marathon/core/matcher/base/util/ActorOfferMatcherTest.scala
meln1k: Done.

src/main/scala/mesosphere/marathon/core/matcher/manager/OfferMatcherManagerModule.scala
meln1k: In the resource matcher

meln1k:
@jeschkies yep, we have

PR #5674, opened 2017-10-20
 Define Jenkinsjob for on demand SI tests. 

jeschkies:
Summary:
This enables on demand system integration runs in Jenkins that report back to specific pull requests. We can then test system integration test changes.

### DIFF
--- a/Jenkinsfile.pr.si
+++ b/Jenkinsfile.pr.si
@@ -15,20 +15,19 @@ node('shakedown') {
             string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d', variable: 'DOCKER_HUB_USERNAME'),
             string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD')
           ]) {
-            echo """$params.Pull_Request_Id"""
+            // Checkout commit of pull request.
             sshagent(credentials: ['4ff09dce-407b-41d3-847a-9e6609dd91b8']) {
                 sh """git fetch origin pull/$params.Pull_Request_Id/head:$params.Pull_Request_Id"""
                 sh """git checkout $params.Pull_Request_Id"""
             }
             sh """env"""
+
             sh """./ci/si_pipeline.sh $params.Channel $params.Variant"""
           }
         } finally {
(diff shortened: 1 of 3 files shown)


ci/github_status.py
```diff
@@ -0,0 +1,34 @@
...
+jenkins_github = {
+        'PENDING': 'pending',
+        'SUCCESS': 'success',
+        'null': 'success',
+        'UNSTABLE': 'failure',
+        'FAILURE': 'failure'}
+
+
+def main(gh_user, gh_password, commit, status, target_url, context):
+    uri = 'https://api.github.com/repos/mesosphere/marathon/statuses/{}'.format(commit)
```
alenkacz: I actually wanted to ask you today during our 1:1 so we can still do it then but why don't we use github statuses also for regular PR builds? I find it the more "native" way in GitHub.

It is just a general question, not something against this PR...

jeschkies: We do use Github statuses. The should show up for each build. However, I've added the bot notifications to transfer more information.

jeschkies:
> There will be a job that I will run from jenkins gui or how does this work?

Good point. I wanted to add a link to the comment by the CI bot. I will follow up in this PR. Thanks for reminding me.

Jenkinsfile.pr.si
```diff
@@ -0,0 +1,39 @@
...
+            string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD')
+          ]) {
+            // Checkout commit of pull request.
+            sshagent(credentials: ['4ff09dce-407b-41d3-847a-9e6609dd91b8']) {
+                sh """git fetch origin pull/$params.Pull_Request_Id/head:$params.Pull_Request_Id"""
+                sh """git checkout $params.Pull_Request_Id"""
+            }
+            sh """python3 ./ci/github_status.py "$JOB_NAME" "$JOB_URL" \$(git rev-parse HEAD) PENDING"""
+
+            sh """./ci/si_pipeline.sh $params.Channel $params.Variant"""
```
zen-dog: too many `sh` calls here. Shouldn't we have a script for that?

jeschkies: I thought so at first but there is hardly any added value since the github status is only used here. At some point we might want to unify it or have the Github plugin report the status.

PR #5675, opened 2017-10-20
Define minimum executor resources.

janisz:
When running a custom executor Mesos will log the following statement

> W1012 14:34:57.062808 20260 validation.cpp:932] Executor 'marathon-app.c15de967-af49-11e7-88c1-020b7e07fc76' for task 'app.c15de967-af49-11e7-88c1-020b7e07fc76' uses less memory (None) than the minimum required (32MB). Please update your executor, as this will be mandatory in future releases.

this PR sets CPU/MEM resources so this statement is not logged.

Refs:
* https://issues.apache.org/jira/browse/MESOS-1807
* https://github.com/apache/mesos/blob/1.4.0/src/master/validation.cpp#L1292-L1314

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskBuilder.scala
@@ -11,6 +11,7 @@ import mesosphere.marathon.state._
 import mesosphere.marathon.stream.Implicits._
 import mesosphere.mesos.ResourceMatcher.ResourceMatch
 import mesosphere.mesos.protos.Implicits._
+import mesosphere.mesos.protos.ScalarResource
 import org.apache.mesos.Protos.Environment._
 import org.apache.mesos.Protos._
 
@@ -65,7 +66,10 @@ class TaskBuilder(
 
         val info = ExecutorInfo.newBuilder()
           .setExecutorId(ExecutorID.newBuilder().setValue(executorId))
-
+          .addAllResources(Seq(
+            ScalarResource.cpus(0.1),
+            ScalarResource.memory(32.0)
+          ).map(resourceToProto).asJava)
         containerProto.foreach(info.setContainer)
 
         val command =
--- a/src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
+++ b/src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
@@ -3,23 +3,24 @@ package mesosphere.mesos
 import com.google.protobuf.TextFormat
 import mesosphere.UnitTest
 import mesosphere.marathon.api.serialization.PortDefinitionSerializer
-import mesosphere.marathon.core.instance.{ Instance, TestInstanceBuilder }
-import mesosphere.marathon.core.pod.{ BridgeNetwork, ContainerNetwork }
+import mesosphere.marathon.core.instance.{Instance, TestInstanceBuilder}
+import mesosphere.marathon.core.pod.{BridgeNetwork, ContainerNetwork}
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.state.NetworkInfo
-import mesosphere.marathon.raml.{ App, Resources }
-import mesosphere.marathon.state.Container.{ Docker, PortMapping }
+import mesosphere.marathon.raml.{App, Resources}
+import mesosphere.marathon.state.Container.{Docker, PortMapping}
 import mesosphere.marathon.state.PathId._
 import mesosphere.marathon.state.VersionInfo.OnlyVersion
-import mesosphere.marathon.state.{ AppDefinition, Container, PathId, Timestamp, _ }
+import mesosphere.marathon.state.{AppDefinition, Container, PathId, Timestamp, _}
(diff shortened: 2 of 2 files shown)


src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -3,23 +3,24 @@ package mesosphere.mesos
...
 import mesosphere.marathon.api.serialization.PortDefinitionSerializer
-import mesosphere.marathon.core.instance.{ Instance, TestInstanceBuilder }
-import mesosphere.marathon.core.pod.{ BridgeNetwork, ContainerNetwork }
+import mesosphere.marathon.core.instance.{Instance, TestInstanceBuilder}
+import mesosphere.marathon.core.pod.{BridgeNetwork, ContainerNetwork}
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.state.NetworkInfo
-import mesosphere.marathon.raml.{ App, Resources }
-import mesosphere.marathon.state.Container.{ Docker, PortMapping }
+import mesosphere.marathon.raml.{App, Resources}
```
ichernetsky: Could you please run `sbt compile` to have these imports properly formatted and updated this PR?

src/main/scala/mesosphere/mesos/TaskBuilder.scala
```diff
@@ -65,7 +66,10 @@ class TaskBuilder(
 
         val info = ExecutorInfo.newBuilder()
           .setExecutorId(ExecutorID.newBuilder().setValue(executorId))
-
+          .addAllResources(Seq(
+            ScalarResource.cpus(0.1),
+            ScalarResource.memory(32.0)
```
timcharper: Nit: extract these as constants (and let's add a comment here)

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
timcharper: 👍

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -990,6 +991,10 @@ class TaskBuilderTest extends UnitTest {
 
       assert(!taskInfo.hasCommand)
       assert(cmd.getValue == "chmod ug+rx '/custom/executor' && exec '/custom/executor' a b c")
+      assert(taskInfo.getExecutor.getResourcesList == Seq(
+        ScalarResource.cpus(0.1),
+        ScalarResource.memory(32.0)
+      ).map(resourceToProto).asJava)
```
timcharper: Test will be easier to read with the constant; `TaskBuilder.defaultExecutorResources`

src/main/scala/mesosphere/mesos/TaskBuilder.scala
timcharper: I suggest

```
object TaskBuilder {
  val executorResources = Seq(...)
}

In a future commit we'll probably want to parameterize this.

timcharper:
@meln1k Great point about breaking compatibility. New Marathon versions don't support old Mesos versions (but old Marathon versions generally work with new Mesos versions).

timcharper:
@kensipe brings up some great points. After thinking about this, I don't think setting a default is the right way to go. Mesos effectively has a default, and is disabling it in order to force one to think about custom executor resources. We should follow suit.

janisz:
@kensipe @timcharper I totally agree that setting minimal resources does not solve any problem. 

@kensipe Custom executor fixes following issues:
- registering task in Service Discovery — like in [Aurora Executor](http://aurora.apache.org/documentation/latest/getting-started/overview/)
- Handling logs — like [ContainerLogger](http://mesos.apache.org/documentation/latest/logging/#containerlogger-module) but without dependency on Mesos Module and C++
- Graceful startup and shutdown — like in [Singularity](https://github.com/HubSpot/Singularity/blob/master/Docs/about/how-it-works.md#optional-components)
- Handling configuration for tasks — like [Mesos Secrets](http://mesos.apache.org/documentation/latest/secrets/#secrets) but easier to implement

Some other implementation: https://github.com/Devatoria/go-mesos-executor

---

> I think the direction we should go, if we are to continue to support custom executors, is to add a way to specify executorResources to the app definition, and not provide a default. 

I'll go in that direction. This might take some time so I'll close this PR now.

timcharper:
@janisz thanks for being so awesome, understanding, (and patient!!!) <3

We've discussed this as a team, and it dawned on us that our contributors should follow a similar process that we use.

I took some time to update our contributing document: https://github.com/mesosphere/marathon/pull/5719 . As a team, we're converging on this being our contributing process (which is the same process that applies to us).

Part of this proposal is that we submit a JIRA ticket for an improvement so that we can have a place to discuss some change proposal and get consensus on the direction, prior to implementation. Similar to what I just mentioned on the your other PR, this ideally will enable us to collaborate with the contributors in a design role before too much effort is spent polishing patches or coding.

How does this sound to you?

PR #5676, opened 2017-10-23
Change `Int.MaxValue` to 8 for maximum concurrency during migration.

unterstein:
Summary:
As discussed in #5659, we should introduce a restriction on the  amount of simultaneous calls. Therefore we change `Int.MaxValue` to 8 for maximum concurrency during migration.


jeschkies:
Why does this diff have a `backport` label?

unterstein:
Because I will backport this once it`s merged.

PR #5678, opened 2017-10-23
AppsController split: `list running tasks` method

meln1k:
AppsController split: `list running tasks` method

Summary: splitting of the big apps controller pr, this time `list running tasks` method

JIRA issues: MARATHON-7299

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -72,6 +72,7 @@ class AppsController(
     val pluginManager: PluginManager)(
     implicit
     val actorSystem: ActorSystem,
+    val materializer: Materializer,
     val executionContext: ExecutionContext,
     val authenticator: MarathonAuthenticator,
     val authorizer: Authorizer,
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -353,7 +353,7 @@ class AppsController(
     Source(appIds).mapAsync(1) { appId =>
       async {
         val instances = instancesBySpec.specInstances(appId)
-        lazy val healthStatuses = await(healthCheckManager.statuses(appId))
+        val healthStatuses = await(healthCheckManager.statuses(appId))
         instances.flatMap { instance =>
           val health = healthStatuses.getOrElse(instance.instanceId, Nil)
           instance.tasksMap.values.map { task => EnrichedTask(appId, task, instance.agentInfo, health) }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -8,9 +8,14 @@ import akka.event.EventStream
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.{ StatusCodes, Uri }
 import akka.http.scaladsl.model.headers.Location
-import akka.http.scaladsl.model.StatusCodes
-import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.model.{ StatusCodes, Uri }
+import akka.http.scaladsl.model.headers.Location
 import akka.http.scaladsl.server.{ Directive1, Rejection, RejectionError, Route }
+import mesosphere.marathon.api.akkahttp.AuthDirectives.NotAuthorized
+import mesosphere.marathon.api.akkahttp.PathMatchers.ExistingAppPathId
+import akka.stream.Materializer
+import akka.stream.scaladsl.Source
+import mesosphere.marathon.api.TaskKiller
 import mesosphere.marathon.api.akkahttp.AuthDirectives.NotAuthorized
 import mesosphere.marathon.api.akkahttp.PathMatchers.ExistingAppPathId
 import mesosphere.marathon.api.v2.{ AppHelpers, AppNormalization, InfoEmbedResolver, LabelSelectorParsers }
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -322,7 +331,36 @@ class AppsController(
     }
   }
 
-  private def listRunningTasks(appId: PathId)(implicit identity: Identity): Route = ???
+  private def listRunningTasks(appId: PathId)(implicit identity: Identity): Route = {
+    val maybeApp = groupManager.app(appId)
```
jeschkies: What is the benefit of this temporary variable? Why not match `groupManager.app(appId)` directly?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -322,7 +331,36 @@ class AppsController(
...
+    maybeApp.map { app =>
+      authorized(ViewRunSpec, app).apply {
+        val tasksF = instanceTracker.instancesBySpec flatMap { instancesBySpec =>
+          runningTasks(Set(appId), instancesBySpec)
+        }
+        onSuccess(tasksF) { tasks =>
+          complete(Json.obj("tasks" -> tasks.toRaml))
```
jeschkies: I thinks we have a Raml object that has `tasks` as a field. Could you create one if we don't have it?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -322,7 +331,36 @@ class AppsController(
...
+        }
+      }
+    } getOrElse {
+      reject(Rejections.EntityNotFound.noApp(appId))
```
jeschkies: Why not match the app and use `case None` here? I think that is easier to follow.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -322,7 +331,36 @@ class AppsController(
...
+    }
+  }
+
+  private def runningTasks(appIds: Set[PathId], instancesBySpec: InstancesBySpec): Future[Set[EnrichedTask]] = {
+    Source(appIds)
+      .filter(instancesBySpec.hasSpecInstances)
+      .mapAsync(1)(id => healthCheckManager.statuses(id).map(_ -> id))
+      .mapConcat {
+        case (health, id) =>
+          instancesBySpec.specInstances(id).flatMap { instance =>
```
jeschkies: We already have the instances, don't we?

jeschkies: Ok, so after the quick discussion I think it should be more like this.

First we should use async/await instead of `flatMap`:

```scala
async {
  val instances = await(instancesBySpec.specInstances(id))
  instance.tasksMap.values.map { task =>
    EnrichedTask(id, task, instance.agentInfo, health.getOrElse(instance.instanceId, Nil))
  }
}
```
I'm surprised that `mapConcat` has no troubles with the future here.

Then it seems we would not need the filtering since `instancesBySpec.specInstances` is already doing that. So, I'd first fetch the instances and then attach the health for each:

```scala
Future.sequence(appIds.map { appId =>
  async {
    val instances = await(instancesBySpec.specInstances(appId))
    val healthStatuses = await(healthCheckManager.statuses(appId))
    instances.flatMap { instance =>
      val healths: Seq[???] = healthStatuses.getOrElse(instance.id, Nil)
      instance.tasksMap.values.map { task => EnrichedTask(appId, task, instance.agentInfo, healths)
    }
  }
})
```

that seems to be easier to comprehend. I guess a stream could parallelize the flat mapping over the instances.

jeschkies: Let's just introduce some state:

```scala
Future.sequence(appIds.map { appId =>
  async {
    val instances = await(instancesBySpec.specInstances(appId))
    val healthStatuses = await(healthCheckManager.statuses(appId))
    val tasksSetBuilder = Set.newBuilder[EnrichedTask]
    instances.foreach { instance =>
        val healths: Seq[???] = healthStatuses.getOrElse(instance.id, Nil)
        instance.tasksMap.valuesIterator { task =>
           tasksSetBuilder +=  EnrichedTask(appId, task, instance.agentInfo, healths)
        }
    }
    tasksSetBuilder.build()
  }
})
```

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -322,7 +332,36 @@ class AppsController(
...
+            complete(raml.EnrichedTasksList(tasks))
+          }
+        }
+      case None =>
+        reject(Rejections.EntityNotFound.noApp(appId))
+    }
+  }
+
+  private def runningTasks(appIds: Set[PathId], instancesBySpec: InstancesBySpec): Source[EnrichedTask, NotUsed] = {
+    Source(appIds).mapAsync(1) { appId =>
```
alenkacz: this code is very similar to the tasks controller ... it would really make so much sense to extract this to some service. I am not saying it should be done in this PR...

ichernetsky: I am not sure about it. Code duplication is quite often a lesser evil.

meln1k: Actually we can extract such logic to a service level, e.g. `TaskService`, but I think this goes beyond the scope of the current PR.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -322,7 +331,38 @@ class AppsController(
...
+          }
+        }
+      case None =>
+        reject(Rejections.EntityNotFound.noApp(appId))
+    }
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  private def runningTasks(appIds: Set[PathId], instancesBySpec: InstancesBySpec): Source[EnrichedTask, NotUsed] = {
+    Source(appIds).mapAsync(1) { appId =>
```
jeschkies: Hm, I don't see the benefit of creating a stream here. Why not simply use `Future.sequence`?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -2206,5 +2214,41 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+        InstanceTracker.SpecInstances.forInstances(app.id, Seq(instance))
+      ))
+      healthCheckManager.statuses(app.id) returns Future.successful(Map(instance.instanceId -> Seq(Health(instance.instanceId))))
+
+      val entity = HttpEntity.Empty
+
+      val uri = Uri./.withPath(Path(app.id.toString) / "tasks")
+
+      Get(uri, entity) ~> route ~> check {
+        status shouldEqual StatusCodes.OK
```
jeschkies: Shouldn't this test also verify the tasks?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: The main benefit of having a stream instead of `Future.sequence` is that `sequence` will fire everything at once, and `mapAsync` will respect backpressure.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
meln1k: Done!

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
jeschkies: Ah, ok cool!

PR #5679, opened 2017-10-23
AppsController split: `kill tasks` method

meln1k:
AppsController split: `kill tasks` method

Summary: splitting of the big apps controller pr, this time `kill tasks` method

JIRA issues: MARATHON-7299

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
@@ -3,7 +3,7 @@ package api.akkahttp
 
 import akka.http.scaladsl.model.headers._
 import akka.http.scaladsl.model.{ DateTime, HttpHeader, HttpMethods, HttpProtocols }
-import akka.http.scaladsl.server.{ Directive, Directive0, Route, Directives => AkkaDirectives }
+import akka.http.scaladsl.server.{ Directive, Directive0, Directive1, MalformedQueryParamRejection, Route, Directives => AkkaDirectives }
 import com.wix.accord.{ Failure, Success, Result => ValidationResult }
 
 import scala.concurrent.duration._
@@ -76,4 +76,30 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
       case Success => f(Unit)
     }
   }
+
+  def extractTaskKillingMode: Directive1[TaskKillingMode] =
+    (parameter("scale".as[Boolean].?(false)) & parameter("wipe".as[Boolean].?(false))).tflatMap {
+      case (scale, wipe) =>
+        if (scale && wipe) {
+          reject(MalformedQueryParamRejection("scale, wipe", "You cannot use scale and wipe at the same time."))
+        } else if (scale) {
+          provide(TaskKillingMode.Scale)
+        } else if (wipe) {
+          provide(TaskKillingMode.Wipe)
+        } else {
+          provide(TaskKillingMode.KillWithoutWipe)
+        }
+    }
+
+  sealed trait TaskKillingMode
+  object TaskKillingMode {
+
+    case object Scale extends TaskKillingMode
+
+    case object KillWithoutWipe extends TaskKillingMode
+
+    case object Wipe extends TaskKillingMode
+
+  }
+
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -324,7 +324,30 @@ class AppsController(
 
   private def listRunningTasks(appId: PathId)(implicit identity: Identity): Route = ???
 
-  private def killTasks(appId: PathId)(implicit identity: Identity): Route = ???
+  private def killTasks(appId: PathId)(implicit identity: Identity): Route = {
+    // the line below doesn't look nice but it doesn't compile if we use parameters directive
+    (forceParameter & parameter("host") & extractTaskKillingMode) {
+      (force, host, mode) =>
+        def findToKill(appTasks: Seq[Instance]): Seq[Instance] = {
+          appTasks.filter(_.agentInfo.host == host || host == "*")
+        }
+        mode match {
+          case TaskKillingMode.Scale =>
+            val deploymentPlanF = taskKiller.killAndScale(appId, findToKill, force)
+            onSuccess(deploymentPlanF) { plan =>
+              complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+            }
+          case TaskKillingMode.Wipe =>
+            onSuccess(taskKiller.kill(appId, findToKill, wipe = true)) { instances =>
+              complete(Json.obj("tasks" -> instances))
+            }
+          case TaskKillingMode.KillWithoutWipe =>
+            onSuccess(taskKiller.kill(appId, findToKill, wipe = false)) { instances =>
+              complete(Json.obj("tasks" -> instances))
+            }
+        }
+    }
+  }
 
   private def killTask(appId: PathId, taskId: TaskId)(implicit identity: Identity): Route = ???
 
--- a/src/test/scala/mesosphere/marathon/api/akkahttp/DirectivesTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/akkahttp/DirectivesTest.scala
@@ -1,0 +1,44 @@
+package mesosphere.marathon
+package api.akkahttp
+
+import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.model.Uri.{ Path, Query }
+import akka.http.scaladsl.server.MalformedQueryParamRejection
+import akka.http.scaladsl.server.PathMatcher.{ Matched, Unmatched }
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import mesosphere.UnitTest
+import mesosphere.marathon.state.{ AppDefinition, PathId }
+import mesosphere.marathon.test.GroupCreation
+
+class DirectivesTest extends UnitTest with ScalatestRouteTest {
+  import Directives._
+
+  def objectName(obj: AnyRef) = obj.getClass.getName
+
+  val route = extractTaskKillingMode { mode =>
+    complete(objectName(mode))
+  }
+
+  "Directives" should {
+    "extract scale killing mode from request" in {
+      Get().withUri(Uri./.withQuery(Query("scale" -> "true"))) ~> route ~> check {
+        responseAs[String] shouldEqual objectName(TaskKillingMode.Scale)
+      }
+    }
+    "extract wipe killing mode from request" in {
+      Get().withUri(Uri./.withQuery(Query("wipe" -> "true"))) ~> route ~> check {
+        responseAs[String] shouldEqual objectName(TaskKillingMode.Wipe)
+      }
+    }
+    "extract killWithoutWipe killing mode from request" in {
+      Get().withUri(Uri./) ~> route ~> check {
+        responseAs[String] shouldEqual objectName(TaskKillingMode.KillWithoutWipe)
+      }
+    }
+    "reject when scale and wipe are both true" in {
+      Get().withUri(Uri./.withQuery(Query("wipe" -> "true", "scale" -> "true"))) ~> route ~> check {
+        rejection shouldBe a[MalformedQueryParamRejection]
+      }
+    }
+  }
+}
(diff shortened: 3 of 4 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -324,7 +324,30 @@ class AppsController(
...
+  private def killTasks(appId: PathId)(implicit identity: Identity): Route = {
+    // the line below doesn't look nice but it doesn't compile if we use parameters directive
+    (forceParameter & parameter("host") & extractTaskKillingMode) {
+      (force, host, mode) =>
+        def findToKill(appTasks: Seq[Instance]): Seq[Instance] = {
+          appTasks.filter(_.agentInfo.host == host || host == "*")
+        }
+        mode match {
+          case TaskKillingMode.Scale =>
+            val deploymentPlanF = taskKiller.killAndScale(appId, findToKill, force)
```
jeschkies: What is the F pattern? Du you mean `deploymentPlan` is a future? Why not use the type `deploymentPlan: Future[???]`?

src/test/scala/mesosphere/marathon/api/akkahttp/DirectivesTest.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def objectName(obj: AnyRef) = obj.getClass.getName
+
+  val route = extractTaskKillingMode { mode =>
+    complete(objectName(mode))
+  }
+
+  "Directives" should {
+    "extract scale killing mode from request" in {
+      Get().withUri(Uri./.withQuery(Query("scale" -> "true"))) ~> route ~> check {
```
jeschkies: Create tests! Is there a way to test the directive without `route` and `check`?

meln1k: I think there is, but since directives are highly coupled to routes (for example `apply` of the directive returns a `Route`), it will be less concise compared to current way of testing them.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: Yep, I do. Usually it's useful if we use maps/flatmaps so we can write like this:
```scala
val valueF = fetchValue
valueF.map { value =>
  ...
}
```

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -98,6 +101,31 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
     }
   }
 
+  def extractTaskKillingMode: Directive1[TaskKillingMode] =
```
zen-dog: put it into `AppDirectives` may be? Let's try not to create a `UtilsDirective` class - those always end up being a big pile of "garbage"

src/main/scala/mesosphere/marathon/api/akkahttp/AppsDirectives.scala
```diff
@@ -0,0 +1,33 @@
+package mesosphere.marathon
+package api.akkahttp
```
jeschkies: Hm, I'm not sure this class should go into this package. Isn't it v2 specific? Not a blocker though 🙂.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -46,6 +46,7 @@ import mesosphere.marathon.core.task.Task.{ Id => TaskId }
 import PathMatchers._
 import mesosphere.marathon.raml.{ AnyToRaml, AppUpdate, DeploymentResult }
 import mesosphere.marathon.raml.EnrichedTaskConversion._
+import AppsDirectives._
```
jeschkies: Why not `import AppsDirectives.{extractTaskKillingMode, TaskKillingMode}`? What is the benefit of the underscore import?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -324,7 +325,30 @@ class AppsController(
...
+        }
+        mode match {
+          case TaskKillingMode.Scale =>
+            val deploymentPlan = taskKiller.killAndScale(appId, findToKill, force)
+            onSuccess(deploymentPlan) { plan =>
+              complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+            }
+          case TaskKillingMode.Wipe =>
+            onSuccess(taskKiller.kill(appId, findToKill, wipe = true)) { instances =>
+              complete(Json.obj("tasks" -> instances))
```
jeschkies: As I mentioned in the comment before. I think the is a RAML type. If not, could you please create one?

src/test/scala/mesosphere/marathon/api/akkahttp/AppsDirectivesTest.scala
```diff
@@ -0,0 +1,45 @@
...
+class AppsDirectivesTest extends UnitTest with ScalatestRouteTest {
+  import AppsDirectives._
+
+  def objectName(obj: AnyRef) = obj.getClass.getName
+
+  val route = extractTaskKillingMode { mode =>
+    complete(objectName(mode))
+  }
+
+  "Directives" should {
```
jeschkies: Nit: `AppDirectives`

src/test/scala/mesosphere/marathon/api/akkahttp/AppsDirectivesTest.scala
```diff
@@ -0,0 +1,45 @@
...
+import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.testkit.ScalatestRouteTest
+import mesosphere.UnitTest
+import mesosphere.marathon.state.{ AppDefinition, PathId }
+import mesosphere.marathon.test.GroupCreation
+
+class AppsDirectivesTest extends UnitTest with ScalatestRouteTest {
+  import AppsDirectives._
+
+  def objectName(obj: AnyRef) = obj.getClass.getName
```
jeschkies: I'm not a big fan of this way but since I cannot come up with a better one this'll do.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -2206,5 +2206,63 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
+
+    "Kill tasks and wipe" in new Fixture {
+      val app = AppDefinition(id = PathId("/app"))
+      val rootGroup = createRootGroup(Map(app.id -> app))
+      val plan = DeploymentPlan(rootGroup, rootGroup)
+      groupManager.app(PathId("/app")) returns Some(app)
+
+      groupManager.rootGroup() returns rootGroup
+      taskKiller.killAndScale(any, any, any)(any) returns Future.successful(plan)
+      taskKiller.kill(any, any, any)(any) returns Future.successful(Seq.empty)
```
jeschkies: Why `any` as parameters and not the values we expect?

src/main/scala/mesosphere/marathon/api/akkahttp/AppsDirectives.scala
zen-dog: I would be ok with it here for now. We'll probably have to rework package structure anyway.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: No specific reason.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -2206,5 +2206,63 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
       }
     }
+    "Kill tasks and scale" in new Fixture {
+      val app = AppDefinition(id = PathId("/app"))
+      val rootGroup = createRootGroup(Map(app.id -> app))
+      val plan = DeploymentPlan(rootGroup, rootGroup)
+      groupManager.app(PathId("/app")) returns Some(app)
+
+      groupManager.rootGroup() returns rootGroup
+      taskKiller.killAndScale(any, any, any)(any) returns Future.successful(plan)
```
zen-dog: Here and below I guess it could be at least `taskKiller.killAndScale(app.id, any...)` to make sure we're not trying to kill anything else. Even if that's a bit paranoid.

src/test/scala/mesosphere/marathon/api/akkahttp/DirectivesTest.scala
jeschkies: Btw I meant `Great tests`. Shame on me for this Freudian slip.

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -500,7 +500,8 @@
               If scale=true, than a deployment is triggered and the deployment is returned.
             body:
               application/json:
-                example: !include examples/app_tasks.json
+                type: instance.InstanceList | deploymentResult.DeploymentResult
+#                example: !include examples/app_tasks.json //todo fixit
```
jeschkies: 🤔 shouldn't the changes to `app_taks.json` fix the RAML generation errors. Why is this commented?

docs/docs/rest-api/public/api/v2/examples/app_tasks.json
```diff
@@ -1,16 +1,44 @@
...
+      "agentInfo": {
+        "host": "srv7.hw.ca1.mesosphere.com",
+        "attributes": []
+      },
+      "taskMap": {
+        "taskName": {
+          "taskId": "minecraft_survival-world.564bd685-4c30-11e5-98c1-be5b2935a987",
+          "reservation" : {
+            "volumeIds": {
+              "runSpecId": "/minecraft/survival-world",
```
jeschkies: Oh man. You are starting something here!

docs/docs/rest-api/public/api/v2/types/Instance.raml
```diff
@@ -0,0 +1,46 @@
...
+
+  Instance:
+    type: object
+    properties:
+      instanceId: strings.InstanceId
+      agentInfo: AgentInfo
+      taskMap:
+        type: object
+        properties:
+          /.*/: task.Task
```
jeschkies: Should this rather be the regex for valid task ids? What do you think?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -325,7 +324,30 @@ class AppsController(
...
+            onSuccess(deploymentPlan) { plan =>
+              complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+            }
+          case TaskKillingMode.Wipe =>
+            onSuccess(taskKiller.kill(appId, findToKill, wipe = true)) { instances =>
+              complete(raml.InstanceList(instances.map(_.toRaml)))
+            }
+          case TaskKillingMode.KillWithoutWipe =>
+            onSuccess(taskKiller.kill(appId, findToKill, wipe = false)) { instances =>
+              complete(Json.obj("tasks" -> instances))
```
jeschkies: I thought this could be a RAML now 🙂.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -3,18 +3,18 @@ package raml
 
 import java.time.OffsetDateTime
 
-import mesosphere.marathon.core.condition.Condition
+import mesosphere.marathon.core.condition
```
jeschkies: What is the motivation for this change>

docs/docs/rest-api/public/api/v2/types/Instance.raml
jeschkies: Ok, as you've told me we should be more flexible for the output type.

docs/docs/rest-api/public/api/v2/apps.raml
meln1k: forgot to uncomment this, thanks!

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -501,7 +501,7 @@
             body:
               application/json:
                 type: instance.InstanceList | deploymentResult.DeploymentResult
-#                example: !include examples/app_tasks.json //todo fixit
+        #        example: !include examples/app_tasks.json
```
jeschkies: Should this stay this way?

PR #5680, opened 2017-10-23
AppsController split: `kill task` method

meln1k:
Summary: splitting a big appsController pr, this time `kill task` method

JIRA issues: MARATHON-7299


src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -76,4 +76,30 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
       case Success => f(Unit)
     }
   }
+
+  def extractTaskKillingMode: Directive1[TaskKillingMode] =
+    (parameter("scale".as[Boolean].?(false)) & parameter("wipe".as[Boolean].?(false))).tflatMap {
```
jeschkies: What is `tflatMap`?

meln1k:
@jeschkies no, it doesn't, although they both contain new directive.

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
meln1k: A something like `flatMap` but for directives :)

alenkacz: I did not know that one either. Thanks for asking and answering :)

alenkacz:
@meln1k now that I look at it the other day with clear head ... why don't we need any kind of authorization for killing tasks? For listing tasks it is present...

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
jeschkies: I actually stumbled upon this as well.

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -98,6 +101,31 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
     }
   }
 
+  def extractTaskKillingMode: Directive1[TaskKillingMode] =
```
jeschkies: How is this different from `AppsDirectives.extractTaskKillingMode`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -326,7 +326,36 @@ class AppsController(
 
   private def killTasks(appId: PathId)(implicit identity: Identity): Route = ???
 
-  private def killTask(appId: PathId, taskId: TaskId)(implicit identity: Identity): Route = ???
+  private def killTask(appId: PathId, taskId: TaskId)(implicit identity: Identity): Route = {
```
jeschkies: This looks incredibly similar to `killTasks` from #5679. The only difference seems to be `findToKill`. What do you think about consolidating these two diffs? You could make one branch the base of the other.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/AppsControllerTest.scala
```diff
@@ -2343,5 +2352,60 @@ class AppsControllerTest extends UnitTest with GroupCreation with ScalatestRoute
...
     }
 
+    "Kill task and scale" in new Fixture {
+      val app = AppDefinition(id = PathId("/app"))
+      val rootGroup = createRootGroup(Map(app.id -> app))
+      val plan = DeploymentPlan(rootGroup, rootGroup)
+      groupManager.app(PathId("/app")) returns Some(app)
+
+      groupManager.rootGroup() returns rootGroup
+      taskKiller.killAndScale(any, any, any)(any) returns Future.successful(plan)
```
jeschkies: Hm, shouldn't we match the arguments?

meln1k: Good catch!

PR #5682, opened 2017-10-24
 Draft pods controller with stubs.

jeschkies:
Summary:
This creates the routes as stubs for the pods controller. We also define a directive to extract Instance ids or reject the request if the id is invalid.


src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -76,4 +78,20 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
       case Success => f(Unit)
     }
   }
+
+  /**
+    * Matches the remaining path and transforms it into an instance id or rejects if it is not a valid id.
+    */
+    class ExtractInstanceId extends Directive[Instance.Id] {
```
zen-dog: We already have `PathMatchers` to match for `pathId`. Shouldn't we put this method there?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,87 @@
...
+    head {
+      capability()
+    } ~
+      get {
+        findAll()
+      } ~
+      post {
+        create()
+      } ~
+      pathPrefix(ExistingAppPathId(groupManager.rootGroup)) { podId =>
```
zen-dog: Should we rename the matcher to `ExistingPathId` or `ExistingRunSpecId`?

jeschkies: We actually should use another one since `ExistingAppPathId` is not returning 400 if the app id is invalid.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,89 @@
...
+  def version(version: String): Route = ???
+
+  def allStatus(): Route = ???
+
+  def killInstance(instanceId: Instance.Id): Route = ???
+
+  def killInstances(podId: PathId): Route = ???
+
+  override val route: Route =
+    head {
```
alenkacz: no asLeader and authenticated directives here?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,113 @@
...
+      capability()
+    } ~
+    get {
+      pathEnd {
+        findAll()
+      } ~
+      path("::status" ~ PathEnd) {
+        allStatus()
+      } ~
+      path(PodsPathIdLike ~ PathEnd) { runSpecId: String =>
```
alenkacz: @jeschkies I like this version more. From my point of view much more readable!

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,113 @@
...
+        assumeValid(validatePathId(runSpecId)) {
+          find(PathId(runSpecId))
+        }
+      } ~
+      path(PodsPathIdLike ~ "::status" ~ PathEnd) { runSpecId: String =>
+        assumeValid(validatePathId(runSpecId)) {
```
alenkacz: to make it even more readable here I would probably go towards direction of having a tailor-made directive for this type called assumeValidPathId but it's up to a discussion. Your approach is better if we want to validate many different entities... But I am not sure whether that is the case.

zen-dog:
Looks good. What about `authenticated` directive?

src/main/scala/mesosphere/marathon/api/akkahttp/PathMatchers.scala
```diff
@@ -19,6 +19,10 @@ object PathMatchers {
   final val RemainingTaskId = Remaining.map(s => Task.Id(s))
 
   /**
+    * Matches
```
meln1k: Matches what? :)

src/main/scala/mesosphere/marathon/api/akkahttp/PathMatchers.scala
```diff
@@ -49,6 +53,10 @@ object PathMatchers {
     override def apply(path: Path) = iter(Nil, path)
   }
 
+  case object PodsPathIdLike extends PathMatcher1[String] {
```
meln1k: Why do we need a matcher which doesn't match anything?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,113 @@
...
+import mesosphere.marathon.core.group.GroupManager
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.plugin.auth.Authenticator
+import mesosphere.marathon.state.PathId
+import akka.http.scaladsl.server.PathMatchers
+
+class PodsController(
+    val groupManager: GroupManager)(
+    implicit
+    val authenticator: Authenticator) extends Controller {
```
meln1k: Can we pass it explicitly?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  import mesosphere.marathon.api.akkahttp.Directives._
+
+  def capability(): Route =
+    authenticated.apply { implicit identity =>
+      complete("")
```
meln1k: You can also use `complete(StatusCodes.OK)`

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -0,0 +1,113 @@
...
+      }
+    } ~
+    delete {
+      path(PodsPathIdLike ~ PathEnd) { runSpecId: String =>
+        assumeValid(validatePathId(runSpecId)) {
+          remove(PathId(runSpecId))
+        }
+      } ~
+      path(PodsPathIdLike ~ "::instances" ~ PathEnd) { runSpecId: String =>
+        assumeValid(validatePathId(runSpecId)) {
```
meln1k: What do you think about moving `assumeValid(validatePathId(runSpecId))` to a separate directive? Something like `forceParameter` in apps controller

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -88,11 +90,21 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
     * @param result The result of a Wix validation.
     * @return The passed inner route.
     */
-  def assumeValid(result: ValidationResult): Directive0 = Directive { f: (Unit => Route) =>
+  def assumeValid(result: ValidationResult): Directive0 = {
```
meln1k: I like this change :)

src/main/scala/mesosphere/marathon/api/akkahttp/PathMatchers.scala
jeschkies: Take a guess 😉

src/main/scala/mesosphere/marathon/api/akkahttp/PathMatchers.scala
jeschkies: I've updated it.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
jeschkies: Sure. So I've finally convinced you?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
meln1k: @jeschkies I think so

src/test/scala/mesosphere/marathon/api/akkahttp/PathMatchersTest.scala
```diff
@@ -67,4 +67,31 @@ class PathMatchersTest extends UnitTest with GroupCreation with ScalatestRouteTe
       AppPathIdLike(Path("/restart")) shouldBe Unmatched
     }
   }
+
+  "PodsPathIdLike matcher" should {
```
meln1k: Property based tests! Yay!

jeschkies: Hehe, yeah. We could actually use random strings. That's be fun :).

PR #5683, opened 2017-10-24
Do not kill pods even when calling kill and scale

alenkacz:
Pods cannot be killed using /tasks endpoint

Summary:
In this PR https://github.com/mesosphere/marathon/commit/89740d036d2107116cb18dba56644fc35def077c we introduced filtering of tasks to be killed to avoid killing pods. I think the same thing should be done for killAndScale not just kill. I changed this in the old as well as new implementation.

JIRA issues:


src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -100,14 +101,23 @@ class TasksController(
         }
       }
   }
-  private def getTasksByAppId(instanceIdsToAppId: Map[Id, PathId])(implicit identity: Identity): Future[Map[PathId, Seq[Instance]]] = {
+
+  private def getTasksToKill(instanceIdsToAppId: Map[Id, PathId], affectedApps: IndexedSeq[AppDefinition])(implicit identity: Identity): Future[Map[PathId, Seq[Instance]]] = async {
+    /**
+      * Pods can't be killed using /tasks endpoint, only apps
+      */
+    def isNotPod(appId: PathId): Boolean = affectedApps.exists(app => app.id == appId)
```
jeschkies: How is `affectedApps.exists(app => app.id == appId)` verifying that the app is not a pod?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -100,14 +101,23 @@ class TasksController(
         }
       }
   }
-  private def getTasksByAppId(instanceIdsToAppId: Map[Id, PathId])(implicit identity: Identity): Future[Map[PathId, Seq[Instance]]] = {
+
+  private def getTasksToKill(instanceIdsToAppId: Map[Id, PathId], affectedApps: IndexedSeq[AppDefinition])(implicit identity: Identity): Future[Map[PathId, Seq[Instance]]] = async {
```
jeschkies: Could you add docs?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -193,10 +201,14 @@ class TasksController(
     * @return new deployment created as the result of scale operation
     */
   @SuppressWarnings(Array("all")) // async/await
-  def killAndScale(tasksByAppIdFuture: Future[Map[PathId, Seq[Instance]]], force: Boolean)(implicit identity: Identity): Future[DeploymentResult] = async {
-    val tasksByAppId = await(tasksByAppIdFuture)
-    val deploymentPlan = await(taskKiller.killAndScale(tasksByAppId, force))
-    DeploymentResult(deploymentPlan.id, deploymentPlan.version.toOffsetDateTime)
+  def killAndScale(toKillFuture: Future[Map[PathId, Seq[Instance]]], force: Boolean)(implicit identity: Identity): Future[Either[Rejection, DeploymentResult]] = async {
```
jeschkies: I missed this on the first review. Why do we pass `toKillFuture` as a future? What is the motivation?

jeschkies: Also, why are the passed? I rather inverse the call. This should call the method that defines what should be killed.

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
```diff
@@ -159,9 +157,15 @@ class TasksResource @Inject() (
       val tasksByAppId: Map[PathId, Seq[Instance]] = maybeInstances.flatten
         .groupBy(instance => instance.instanceId.runSpecId)
         .map { case (appId, instances) => appId -> instances.to[Seq] }(collection.breakOut)
+      val affectedApps = tasksIdToAppId.values.flatMap(appId => groupManager.app(appId))(collection.breakOut)
+      // FIXME (gkleiman): taskKiller.kill a few lines below also checks authorization, but we need to check ALL before
+      // starting to kill tasks
+      affectedApps.foreach(checkAuthorization(UpdateRunSpec, _))
+      val toKill = tasksByAppId
+        .filter { case (appId, _) => affectedApps.exists(app => app.id == appId) }
```
jeschkies: This is strange to me. Why do we have to filter at all? If we replace `affectedApps` with its assignment we get

```scala
tasksByAppId.foreach { case (appId, _) =>
  tasksIdToAppId.values
    .flatMap(appId => groupManager.app(appId))
    .exists(app => app.id == appId) }
}
``` 

How do `tasksByAppId` and `tasksIdToAppId` differ? It seems we are just trying to find the apps that are in both but I cannot find how they differ at all.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
alenkacz: from what I understood from @unterstein affected apps is collection of apps that was retrieved via groupManager.app which should return None if the PathId actually belong to Pod...

Any idea how to have this more obvious ... ?

alenkacz:
@meln1k your comment fixed
@jeschkies I rewrote this code so many times and I agree it's still not perfect. I will try to dig into this once again and will ping you when it is ready

PR #5684, opened 2017-10-24
Small refactoring of System controller

alenkacz:
Summary:
Use newly introduced directives instead of custom accept header parsing code.

JIRA issues:

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
@@ -116,9 +116,9 @@ class SystemController(val marathonConfig: MarathonConf, val cfg: Config, val el
     asLeader(electionService) {
       path("ping") {
         (get & acceptsAnything) { pingPlainText } ~
-          (get & accepts(MediaTypes.`text/plain`)) { pingPlainText } ~
-          (get & accepts(MediaTypes.`application/json`)) { pingJson } ~
-          get { complete(HttpResponse(StatusCodes.NoContent)) }
+        (get & accepts(MediaTypes.`text/plain`)) { pingPlainText } ~
+        (get & accepts(MediaTypes.`application/json`)) { pingJson } ~
+         get { complete(HttpResponse(StatusCodes.NoContent)) }
       } ~
         path("metrics") {
           get { metrics }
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
```diff
@@ -137,7 +115,10 @@ class SystemController(val marathonConfig: MarathonConf, val cfg: Config, val el
     // It would make sense to allow the functionality of this controller on every instance.
     asLeader(electionService) {
```
jeschkies: Nit: Could you add `// format: OFF` here and `// format: On` after the routes so the alignment will not be screwed up by scala formatting?

src/main/scala/mesosphere/marathon/api/akkahttp/SystemController.scala
```diff
@@ -34,29 +33,8 @@ class SystemController(val marathonConfig: MarathonConf, val cfg: Config, val el
...
-      case Some(accept) if (acceptsTextPlain(accept.mediaRanges)) =>
-        complete("pong")
-      case Some(accept) if (acceptsApplicationJson(accept.mediaRanges)) =>
-        complete(JsString("ping"))
-      case Some(_) =>
-        complete(HttpResponse(StatusCodes.NoContent))
-    }
-  }
+  def pingPlainText: Route = complete("pong")
+  def pingJson: Route = complete(JsString("ping"))
```
wavesoft: What is the default content-type responded here? Do you think we should explicitly stick on some particular type (ex. "text/plain") in case the default changes in the future?

alenkacz: the default one is actually specified in route like this. 

(get & acceptsAnything) { pingPlainText } ~

meln1k: Do you think it's possible to do something like this?
```scala
  sealed trait Pong
  case object Pong extends Pong

  implicit val pongMarshaller: ToResponseMarshaller[Pong]  = Marshaller.oneOf(
    Marshaller.combined{case Pong => JsString("pong")},
    Marshaller.StringMarshaller.compose { case Pong => "pong" },
    Marshaller.opaque{case Pong => "pong"},
    Marshaller.withFixedContentType(ContentTypes.NoContentType){ case Pong => HttpResponse(StatusCodes.NoContent)}
  )
```
and then in the route it will be just 
```scala
path("ping") {
  complete(Pong)
}
```

alenkacz: it is possible, but I discussed this with @jeschkies before implementing this and he wanted this to be explicitly done via routes...

I don't mind if it is implemented in either way...

I'll merge this one but I am happy to reimplement it using marshallers - or feel free to submit PR and we can discuss it there

PR #5686, opened 2017-10-24
allow underscore for network names, not endpoint name

timcharper:
JIRA Issue: MARATHON-7848

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -573,7 +573,6 @@ object AppDefinition extends GeneralPurposeCombinators {
     // constraints are only validated in RAML layer
     appDef.unreachableStrategy is valid
     appDef.networks is valid(NetworkValidation.modelNetworksValidator)
-    appDef.networks is every(NetworkValidation.modelNetworkValidator)
   } and ExternalVolumes.validApp and EnvVarValue.validApp
 
   @SuppressWarnings(Array("TraversableHead"))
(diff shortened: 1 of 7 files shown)


src/test/scala/mesosphere/ValidationTestLike.scala
```diff
@@ -17,6 +19,20 @@ import org.scalatest.matchers.{ BePropertyMatchResult, BePropertyMatcher, MatchR
 trait ValidationTestLike extends Validation {
   this: Assertions =>
 
+  /**
+    * Validator which takes an object, serializes it to JSON, and then parses it back, allowing up to test validations
```
kensipe: "allowing up to test" doesn't make any sense

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -573,7 +573,6 @@ object AppDefinition extends GeneralPurposeCombinators {
     // constraints are only validated in RAML layer
     appDef.unreachableStrategy is valid
     appDef.networks is valid(NetworkValidation.modelNetworksValidator)
-    appDef.networks is every(NetworkValidation.modelNetworkValidator)
```
timcharper: all of the network-level validation is handled in RAML now

timcharper:
@kensipe it is called "round trip" because we serialize it to JSON, and then parse it back, and check for violations.

I am open for naming suggestions. I used roundTrip because it was there before.

kensipe:
I read the javadocs and understand what it does... It is a crappy name and I did give several naming suggestions.

PR #5687, opened 2017-10-24
Allow underscore for network names, not endpoint name

timcharper:
backport of 7a926370 (#5686)

* allow underscore for network names, not endpoint name

JIRA Issue: MARATHON-7848


timcharper:
I did not backport the unit test because we don't have the fancy wix validation matchers. I did manually test this change on 1.5, however.

PR #5690, opened 2017-10-25
Define build loop target and unique ci log name.

jeschkies:
Summary:
The ci logs included logs of another run because the ci.log file would
not always be removed. This introduces a unique name per builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,8 +21,8 @@ ansiColor('gnome-terminal') {
             reportName: 'Scapegoat Report', reportTitles: ''
         ])
         archive includes: 'sandboxes.tar.gz'
-        archive includes: '*.tar.gz'
-        archive includes: '*.log'  // Only in case the build was  aborted and the logs weren't zipped
+        archive includes: 'ci-*.tar.gz'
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -21,8 +21,8 @@ ansiColor('gnome-terminal') {
             reportName: 'Scapegoat Report', reportTitles: ''
         ])
         archive includes: 'sandboxes.tar.gz'
-        archive includes: 'ci.tar.gz'
-        archive includes: 'ci.log'  // Only in case the build was  aborted and the logs weren't zipped
+        archive includes: '*.tar.gz'
+        archive includes: '*.log'  // Only in case the build was  aborted and the logs weren't zipped
```
zen-dog: `ci-*.log`

ci/pipeline
```diff
@@ -22,11 +22,13 @@ val PACKAGE_DOCS_DIR: Path = pwd / 'target / "universal-docs"
 
 /**
  * Compile Marathon and run unit and integration tests followed by scapegoat.
+ *
+ * @param logFileName Name of file which collects logs.
```
zen-dog: please add: `Default value exits to be able to run this target locally`

jeschkies: I've removed the default because it led to bugs.

PR #5691, opened 2017-10-25
Fix flaky `GracefulTaskKillIntegrationTest`

zen-dog:
Summary:
Fix flaky `GracefulTaskKillIntegrationTest` by scaling down when killing tasks. Otherwise we have a race condition where a new task is started before the test is finished which leads to a failure during clean up. The underlying race condition will be fixed in another PR.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
@@ -207,7 +207,8 @@ private class DeploymentActor(
 
     Done
   }.recover {
-    case NonFatal(error) => logger.warn(s"Error in stopping runSpec ${runSpec.id}", error);
+    case NonFatal(error) =>
+      logger.warn(s"Error in stopping runSpec ${runSpec.id}", error);
       Done
   }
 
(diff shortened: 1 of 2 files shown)


unterstein:
Where does the test fail when it is flaky?
I think the 'TASK_KILLED` will be fetched even if there is a newly started task, because `waitForEventWith` is waiting until the condition is fulfilled.

Do you have a failed test run to get more information?

I would prefer to fix the issue and leave the test as is, but just my personal preference.

zen-dog:
The test itself works, it's the cleanup afterwards that fails because there is task still running (the replacement task for the one killed). Here is a failed run:
https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-pipelines/job/master/438/console

unterstein:
So the race is between starting a new a replacement task and cleaning up the cluster?

zen-dog:
> So the race is between starting a new a replacement task and cleaning up the cluster?
Exactly. We send a `LaunchEphemeral` but there is not tasks running, so nothing to kill either. Deployment that cleans everything completes successfully and then the `TASK_RUNNING` comes. This is an inherent race against mesos and normally task reconciliation will take care of that eventually. But in ITs we don't wait long enough.

PR #5692, opened 2017-10-25
Support pragma.serializeOnly in RAML generator

wavesoft:
Summary:
This commit introduces the "(pragma.serializeOnly):" RAML attribute
that can be used in RAML types that have no discriminator, forcing
the generator to materialize only the `writes` part of the Format.

_JIRA issues: [MARATHON-7841](https://jira.mesosphere.com/browse/MARATHON-7841)_

Example input:

```raml
  DeploymentPlan:
    type: object
    (pragma.serializeOnly): ; <-- Used on base class
    properties:
      id:
        type: string
        description: The UUID of the deployment plan

  DeploymentStepInfo:
    type: DeploymentPlan
    properties:
      affectedApps:
        type: array
        items: strings.PathId
        description: The IDs of the apps affected by this deployment

```

Example Output:

```scala
trait DeploymentPlan extends RamlGenerated with Product with Serializable {
  val id: String
}

case class DeploymentStepInfo(id: String, affectedApps: scala.collection.immutable.Seq[String]) extends DeploymentPlan

object DeploymentStepInfo {
  implicit object playJsonFormat extends play.api.libs.json.Format[DeploymentStepInfo] {
    def reads(json: play.api.libs.json.JsValue): play.api.libs.json.JsResult[DeploymentStepInfo] = {
      val id = json.\("id").validate[String](play.api.libs.json.JsPath.read[String])
      val affectedApps = json.\("affectedApps").validate[scala.collection.immutable.Seq[String]](play.api.libs.json.JsPath.read[scala.collection.immutable.Seq[String]])
      val _errors = Seq(("id", id), ("affectedApps", affectedApps)).collect({
        case (field, e:play.api.libs.json.JsError) => e.repath(play.api.libs.json.JsPath.\(field)).asInstanceOf[play.api.libs.json.JsError]
      })
      if (_errors.nonEmpty) _errors.reduceOption[play.api.libs.json.JsError](_.++(_)).getOrElse(_errors.head)
      else play.api.libs.json.JsSuccess(DeploymentStepInfo(id = id.get, original = original.get, target = target.get, steps = steps.get, version = version.get, affectedApps = affectedApps.get, affectedPods = affectedPods.get, currentActions = currentActions.get, currentStep = currentStep.get, totalSteps = totalSteps.get))
    }
    def writes(o: DeploymentStepInfo): play.api.libs.json.JsValue = {
      val id = play.api.libs.json.Json.toJson(o.id)
      val affectedApps = play.api.libs.json.Json.toJson(o.affectedApps)
  }
  val DefaultOriginal: Option[Group] = None
  val DefaultTarget: Option[Group] = None
}

object DeploymentPlan {
  implicit object playJsonFormat extends play.api.libs.json.Writes[DeploymentPlan] {
    def writes(o: DeploymentPlan): play.api.libs.json.JsValue = {
      val id = play.api.libs.json.Json.toJson(o.id)
    }
  }
  val DefaultOriginal: Option[Group] = None
  val DefaultTarget: Option[Group] = None
}
```

Previously, `DeploymentPlan` was becoming:

```scala
object DeploymentPlan {
  val DefaultOriginal: Option[Group] = None
  val DefaultTarget: Option[Group] = None
}
```

### DIFF
--- a/docs/docs/rest-api/public/api/v2/types/pragma.raml
+++ b/docs/docs/rest-api/public/api/v2/types/pragma.raml
@@ -42,5 +42,5 @@ annotationTypes:
   serializeOnly:
     type: nil
     description:  |
-      Mark a type as "serialize-only" for which no de-serialization code will be generated.
+      Mark an object type as "serialize-only" for which no de-serialization code will be generated.
 
--- a/docs/docs/rest-api/public/api/v2/types/pragma.raml
+++ b/docs/docs/rest-api/public/api/v2/types/pragma.raml
@@ -40,7 +40,7 @@ annotationTypes:
       No functional change - only to indicate the use of one type.
       Use only on type level.
   serializeOnly:
-    type: string
+    type: nil
     description:  |
-      Mark a type as "serialize-only" for which no de-serialization code will be genrated.
+      Mark a type as "serialize-only" for which no de-serialization code will be generated.
 
--- a/docs/docs/rest-api/public/api/v2/types/pragma.raml
+++ b/docs/docs/rest-api/public/api/v2/types/pragma.raml
@@ -39,5 +39,8 @@ annotationTypes:
       Mark a type as internal to indicate, that it is not used in the external API.
       No functional change - only to indicate the use of one type.
       Use only on type level.
-
+  serializeOnly:
+    type: string
+    description:  |
+      Mark a type as "serialize-only" for which no de-serialization code will be genrated.
 
--- a/project/src/main/scala/com/mesosphere/RamlTypeGenerator.scala
+++ b/project/src/main/scala/com/mesosphere/RamlTypeGenerator.scala
@@ -185,7 +185,6 @@ object RamlTypeGenerator {
     override def toString: String = s"Enum($name, $values)"
 
     override def toTree(): Seq[Tree] = {
-      System.err.println(s"Enum: ${name}")
       val baseTrait = TRAITDEF(name) withParents("Product", "Serializable", "RamlGenerated") withFlags Flags.SEALED := BLOCK(
         VAL("value", StringClass),
         DEF("toString", StringClass) withFlags Flags.OVERRIDE := REF("value")
@@ -421,7 +420,6 @@ object RamlTypeGenerator {
     override def toString: String = parentType.fold(s"$name(${fields.mkString(", ")})")(parent => s"$name(${fields.mkString(" , ")}) extends $parent")
 
     override def toTree(): Seq[Tree] = {
-      System.err.println(s"` ${name}")
       val actualFields = fields.filter(_.rawName != discriminator.getOrElse(""))
       val params = actualFields.map(_.param)
       val klass = if (childTypes.nonEmpty) {
@@ -490,13 +488,7 @@ object RamlTypeGenerator {
           OBJECTDEF("playJsonFormat") withParents PLAY_JSON_FORMAT(name) withFlags Flags.IMPLICIT := BLOCK(
             DEF("reads", PLAY_JSON_RESULT(name)) withParams PARAM("json", PlayJsValue) := {
               if (serializeOnly) BLOCK(
-                THROW(
-                  NEW(
-                    REF(s"NotImplementedError") APPLY (
-                      LIT("This type can only be de-serialized")
-                      )
-                  )
-                )
+                THROW( NEW( REF(s"NotImplementedError") APPLY LIT("This type can only be de-serialized") ) )
               )
               else BLOCK(
                 actualFields.map { field =>
@@ -589,7 +581,7 @@ object RamlTypeGenerator {
             DEF("writes", PlayJsValue) withParams PARAM("o", name) := BLOCK(
               REF("o") MATCH
                 childDiscriminators.map { case (k, v) =>
-                  CASE(REF(s"f:${v.name}")) ==> (REF(PlayJson) DOT "toJson" APPLY REF("f") APPLY (REF(v.name) DOT "playJsonFormat"))
+                  CASE(REF(s"f:${v.name}")) ==> (REF(PlayJson) DOT "toJson" APPLY REF("f") APPLY(REF(v.name) DOT "playJsonFormat"))
                 }
             )
           )) ++ defaultFields ++ defaultInstance
--- a/project/src/main/scala/com/mesosphere/RamlTypeGenerator.scala
+++ b/project/src/main/scala/com/mesosphere/RamlTypeGenerator.scala
@@ -485,59 +485,59 @@ object RamlTypeGenerator {
       } else if (actualFields.size > 22 || actualFields.exists(f => f.repeated || f.omitEmpty || f.constraints.nonEmpty) ||
         actualFields.map(_.toString).exists(t => t.toString.startsWith(name) || t.toString.contains(s"[$name]"))) {
         actualFields.map(_.constraints).requiredImports ++ Seq(
-          OBJECTDEF("playJsonFormat") withParents PLAY_JSON_FORMAT(name) withFlags Flags.IMPLICIT := BLOCK(
-            DEF("reads", PLAY_JSON_RESULT(name)) withParams PARAM("json", PlayJsValue) := {
-              if (serializeOnly) BLOCK(
-                THROW( NEW( REF(s"NotImplementedError") APPLY LIT("This type can only be de-serialized") ) )
-              )
-              else BLOCK(
-                actualFields.map { field =>
-                  VAL(field.name) := field.playValidator
-                } ++ Seq(
-                  VAL("_errors") := SEQ(actualFields.map(f => TUPLE(LIT(f.rawName), REF(f.name)))) DOT "collect" APPLY BLOCK(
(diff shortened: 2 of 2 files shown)


project/RamlTypeGenerator.scala
```diff
@@ -411,7 +414,7 @@ object RamlTypeGenerator {
     }
   }
 
-  case class ObjectT(name: String, fields: Seq[FieldT], parentType: Option[String], comments: Seq[String], childTypes: Seq[ObjectT] = Nil, discriminator: Option[String] = None, discriminatorValue: Option[String] = None) extends GeneratedClass {
+  case class ObjectT(name: String, fields: Seq[FieldT], parentType: Option[String], comments: Seq[String], childTypes: Seq[ObjectT] = Nil, discriminator: Option[String] = None, discriminatorValue: Option[String] = None, serializeOnly: Boolean = false) extends GeneratedClass {
```
jeschkies: You've look at this method. Could you add some docs?

wavesoft:
> Also, what do you think about making the generator another sbt project in this repository as a follow up?

I already have a plan for introducing a replacement to the RAML generator as a separate project. Possibly also replacing treehugger. When I am done with patching the performance tests I will clean-up what I already have and open it up for discussion :)

project/RamlTypeGenerator.scala
wavesoft: Why not. Let me add some :smile:

docs/docs/rest-api/public/api/v2/types/pragma.raml
```diff
@@ -39,5 +39,8 @@ annotationTypes:
       Mark a type as internal to indicate, that it is not used in the external API.
       No functional change - only to indicate the use of one type.
       Use only on type level.
-
+  serializeOnly:
+    type: string
```
jdef: why `string` instead of `nil` type? what's the value of making it a string?

jdef: better yet, can we come up with a more generic name for what this attribute is actually doing (declaring support for serialization modes)?

for example:
```
serializationModes:
  enum: [read, write, all]
  default: all
```

wavesoft: Updated description with an example.

> why `string` instead of `nil` type? what's the value of making it a string?

Just copied the previous pragmas, for sake of consistence.

> better yet, can we come up with a more generic name for what this attribute is actually doing (declaring support for serialization modes)?

Actually, that's not a bad idea :smile:

docs/docs/rest-api/public/api/v2/types/pragma.raml
```diff
@@ -39,5 +39,8 @@ annotationTypes:
...
+    description:  |
+      Mark a type as "serialize-only" for which no de-serialization code will be genrated.
```
ichernetsky: A type: `genrated`.

jeschkies:
> When making changes to the RAML generator, it's also useful to include an example of the newly/changed generated code in the PR summary.

I'm usually a fan of checking in the generated code. Unit tests for the generator might help as well :).

jdef:
Are you volunteering to write unit tests for the RAML generator? :)

On Thu, Oct 26, 2017 at 8:58 AM, Karsten Jeschkies <notifications@github.com

wavesoft:
I am volunteering re-writing the RAML generator :trollface:

zen-dog:
Awesome! I was kinda hoping you would get to this point :D

On 26 October 2017 at 17:03, Ioannis Charalampidis <notifications@github.com

jeschkies:
> I am volunteering re-writing the RAML generator :trollface: , with unit tests

Please do not rewrite it 😉. Let's just make it another sbt project. That should be half a day's work.

jeschkies:
> Possibly also replacing treehugger.

Let's have a chat. I had to work on a Swagger generator that used Markdown. So far it seems that [StringTemplate](http://www.stringtemplate.org/) would be a great fit. Did you great a ticket for tracking and discussing the efforts?

wavesoft:
Since `serializationModes: enum` is taking longer than expected (satisfying all code branches proven to be trickier than expected) I am opting for merging this with `pragma.serializeOnly` support and then workign on a proper solution (with unit tests, as a separate project).

And I am saying this because this PR is blocking a chain of other PRs...

docs/docs/rest-api/public/api/v2/types/pragma.raml
```diff
@@ -39,5 +39,8 @@ annotationTypes:
       Mark a type as internal to indicate, that it is not used in the external API.
       No functional change - only to indicate the use of one type.
       Use only on type level.
-
+  serializeOnly:
+    type: nil
```
meln1k: why `nil`?

meichstedt: b/c the flag will be added w/o any value. That's how other similar flags are defined and used.

project/RamlTypeGenerator.scala
```diff
@@ -481,21 +484,26 @@ object RamlTypeGenerator {
...
-                IF(REF("_errors") DOT "nonEmpty") THEN (
-                  REF("_errors") DOT "reduceOption" APPLYTYPE PlayJsError APPLY (REF("_") DOT "++" APPLY REF("_")) DOT "getOrElse" APPLY (REF("_errors") DOT "head")
-                  ) ELSE (
-                  REF(PlayJsSuccess) APPLY (REF(name) APPLY
-                    actualFields.map { field =>
-                      REF(field.name) := (REF(field.name) DOT "get")
-                    }))
+            DEF("reads", PLAY_JSON_RESULT(name)) withParams PARAM("json", PlayJsValue) := {
+              if (serializeOnly) BLOCK(
+                THROW( NEW( REF(s"NotImplementedError") APPLY LIT("This type can only be de-serialized") ) )
```
meichstedt: Actually it would be nicer if no reads function would be generated in this case – instead of generating one that throws. Could be changed later on if we feel like it.

wavesoft: It's a requirement from the `play.api.libs.json.Format` interface, where the `reads` and `writes` are exposed. It cannot be blank :disappointed:

meichstedt: Yes and no. The generator is written to create a
```
implicit object playJsonFormat extends play.api.libs.json.Format[Whatever]
```

A `Format` requires both `Reads` and `Writes`, but the generator could as well only create a
```
implicit object playJsonReads extends play.api.libs.json.Reads[Whatever]
```

meaning no `Writes` and no `Format`. As I said,
> Could be changed later on if we feel like it.

I don't see a strong reason now to mess with the generator if we're eventually replacing it.

wavesoft: True, I was considering performing the minimum possible interventions in the code since I consider every change there quite fragile and untested :smile: 

But I don't think changing `Format` to `Writes` would cause much harm..

jdef:
can you refresh the summary w/ an updated sample of generated code?

wavesoft:
So the changes are the following:

* Instead of using `play.api.libs.json.Formatter` interface it uses `play.api.libs.json.Writes`
* Instead of throwing an error it does not generate the `reads` function at all

This way the compiler will throw a warning if the type is used for de-serialization.

project/RamlTypeGenerator.scala
```diff
@@ -480,54 +483,59 @@ object RamlTypeGenerator {
...
-                  REF(PlayJsObject) APPLY (SEQ(
-                    actualFields.withFilter(_.name != AdditionalProperties).map { field =>
-                      TUPLE(LIT(field.rawName), REF(field.name))
-                    }) DOT "filter" APPLY (REF("_._2") INFIX("!=") APPLY PlayJsNull) DOT("++") APPLY(
-                      actualFields.find(_.name == AdditionalProperties).fold(REF("Seq") DOT "empty") { extraPropertiesField =>
-                      REF("o.additionalProperties") DOT "fields"
-                    })
+          OBJECTDEF("playJsonFormat") withParents (if (serializeOnly) PLAY_JSON_WRITES(name) else PLAY_JSON_FORMAT(name)) withFlags Flags.IMPLICIT := BLOCK(
+            if (serializeOnly) {
+              Seq()
```
jdef: nit: Seq.empty?

docs/docs/rest-api/public/api/v2/types/pragma.raml
```diff
@@ -39,5 +39,8 @@ annotationTypes:
...
+    description:  |
+      Mark a type as "serialize-only" for which no de-serialization code will be generated.
```
jdef: it looks like this is only support for object types (vs enum, string, array, map types). please indicate that in the docs

project/RamlTypeGenerator.scala
```diff
@@ -129,6 +129,9 @@ object RamlTypeGenerator {
   def pragmaForceOptional(o: TypeDeclaration): Boolean =
     o.annotations().exists(_.name() == "(pragma.forceOptional)")
 
+  def pragmaSerializeOnly(o: TypeDeclaration): Boolean =
+    o.annotations().exists(_.name() == "(pragma.serializeOnly)")
```
jdef: maybe validate that this is only being applied to supported types?

PR #5693, opened 2017-10-25
Migrate UnreachableStrategy to migrate saved and running Instances as…

unterstein:
… well (#5659)

Summary:

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-7790

backport of #5659


unterstein:
After this is merged, I would do the backport of https://github.com/mesosphere/marathon/pull/5676 which changes the `Max.IntValue` to `8` usage.

PR #5697, opened 2017-10-25
Stabilized ITs by removing excessive usage of `cleanUp` method

zen-dog:
Summary:
- we used to call `cleanUp` between individual tests in some test suites. It's removed now (with one explicit exception
  in `ResidentTaskIntegrationTest`)
- every test should be a snow flake in a sense that it is independent of what else happens on the cluster which e.g. includes
  unique app names
- added a few helper methods in the `MarathonFacade` class to e.g. fetch deployment for a specific app
- fixed a few unused imports

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/GroupDeployIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/GroupDeployIntegrationTest.scala
@@ -15,9 +15,6 @@ class GroupDeployIntegrationTest extends AkkaIntegrationTest with EmbeddedMarath
 
   import PathId._
 
-  //clean up state before running the test case
-  before(cleanUp())
-
   val appIdCount = new AtomicInteger()
   val groupIdCount = new AtomicInteger()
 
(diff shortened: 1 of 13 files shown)


zen-dog:
Do not merge - currently pointing to `ad/fix-gracefull-taskkill-it` as base to avoid flaky `GracefullTaskkillIntegrationTest`

meln1k:
So, now we don't clean any state?

zen-dog:
> So, now we don't clean any state?

The state is always cleaned at the end of the test suite but not anymore after each test.

src/test/scala/mesosphere/marathon/integration/GroupDeployIntegrationTest.scala
```diff
@@ -74,7 +71,7 @@ class GroupDeployIntegrationTest extends AkkaIntegrationTest with EmbeddedMarath
       Then("The group is deleted")
       result should be(OK)
       // only expect the test base group itself
-      marathon.listGroupsInBaseGroup.value.filter { group => group.id != testBasePath } should be('empty)
+      marathon.listGroupsInBaseGroup.value.exists { g => g.id == group.id } shouldBe false
```
jeschkies: How about 

```scala
 marathon.listGroupsInBaseGroup.value.map(_.id) should contain(group.id)
```

this should make debugging easier.

PR #5698, opened 2017-10-26
Define build loop target and unique ci log name. (#5690)

jeschkies:
Summary:
The ci logs included logs of another run because the ci.log file would
not always be removed. This introduces a unique name per builds.


jeschkies:
Original commit diverted too much.

PR #5700, opened 2017-10-26
Define pr, master and loop targets.

jeschkies:
Summary:
The master loop currently fails because it tries to upload snapshots. To ease the definitions of different targets we separate build and upload logic.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,9 +20,11 @@ ansiColor('gnome-terminal') {
             reportDir: 'target/scala-2.12/scapegoat-report', reportFiles: 'scapegoat.html',
             reportName: 'Scapegoat Report', reportTitles: ''
         ])
-        archive includes: 'sandboxes.tar.gz'
-        archive includes: 'ci-$env.BUILD_TAG.log.tar.gz'
-        archive includes: 'ci-$env.BUILD_TAG.log'  // Only in case the build was  aborted and the logs weren't zipped
+        sh "ls *.tar.gz"
+        echo "tag: $env.BUILD.TAG"
+        archive includes: "sandboxes.tar.gz"
+        archive includes: "ci-$env.BUILD_TAG.log.tar.gz"
+        archive includes: "ci-$env.BUILD_TAG.log"  // Only in case the build was  aborted and the logs weren't zipped
       }
     }
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,10 +21,10 @@ ansiColor('gnome-terminal') {
             reportName: 'Scapegoat Report', reportTitles: ''
         ])
         sh "ls *.tar.gz"
-        echo "tag: $env.BUILD_TAG"
+        echo "tag: ${env.BUILD_TAG}"
         archive includes: "sandboxes.tar.gz"
-        archive includes: "ci-$env.BUILD_TAG.log.tar.gz"
-        archive includes: "ci-$env.BUILD_TAG.log"  // Only in case the build was  aborted and the logs weren't zipped
+        archive includes: "ci-${env.BUILD_TAG}.log.tar.gz"
+        archive includes: "ci-${env.BUILD_TAG}.log"  // Only in case the build was  aborted and the logs weren't zipped
       }
     }
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,7 @@ ansiColor('gnome-terminal') {
             reportName: 'Scapegoat Report', reportTitles: ''
         ])
         sh "ls *.tar.gz"
-        echo "tag: $env.BUILD.TAG"
+        echo "tag: $env.BUILD_TAG"
         archive includes: "sandboxes.tar.gz"
         archive includes: "ci-$env.BUILD_TAG.log.tar.gz"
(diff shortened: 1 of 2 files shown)


ci/pipeline
```diff
@@ -75,13 +75,14 @@ def uploadTarballToS3(): Option[awsClient.Artifact] = utils.stage("Upload Packag
...
+ * @param snapshot Indicates whether we want to upload the snapshot.
  * @return Version and artifact description of Marathon build.
  */
-def createAndUploadPackages(): (String, Option[awsClient.Artifact]) = {
+def createAndUploadPackages(snapshot: Boolean = true): (String, Option[awsClient.Artifact]) = {
   val version = createPackages()
   val docsPath = PACKAGE_DOCS_DIR / s"marathon-docs-$version.tgz"
-  val artifact = uploadTarballToS3()
-  awsClient.upload(docsPath)
+  val artifact = if(snapshot) uploadTarballToS3() else None
```
alenkacz: shouldn't this condition be inverted? The title says we do NOT want to upload when snapshot build but this says if snapshot, then upload... or am I missing something?

jeschkies:
Do not merge yet! The archive does not work yet.

ci/pipeline
jeschkies: Good point. The logic is as I want it but the name `snapshot` is not fitting 🤔.

Jenkinsfile
```diff
@@ -13,16 +13,18 @@ ansiColor('gnome-terminal') {
           sh """sudo -E ci/pipeline jenkins"""
         }
       } finally {
-        junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
-        junit allowEmptyResults: true, testResults: 'target/test-reports/*integration/*.xml'
+        //junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
```
zen-dog: is this temporary?

jeschkies: Of course :). This is for faster iterations on Jenkins.

ci/pipeline
```diff
@@ -30,12 +30,13 @@ def compileAndTest(logFileName: String): Unit = utils.stage("Compile and Test")
 
   def run(cmd: String *) = utils.runWithTimeout(30.minutes, logFileName)(cmd)
 
-  run("sbt", "clean", "test", "integration:test", "scapegoat")
+  //run("sbt", "clean", "test", "integration:test", "scapegoat")
```
zen-dog: don't forget to un-comment

jeschkies: Thanks, Mr. Obvious ^^.

jeschkies: You are right though. I've added the `wip` tag.

ci/pipeline
```diff
@@ -75,13 +76,14 @@ def uploadTarballToS3(): Option[awsClient.Artifact] = utils.stage("Upload Packag
 /**
  * Packages Marathon and uploads its artifacts alongside sha1 checksum to S3.
  *
+ * @param upload Indicates whether we want to upload the tarball package.
  * @return Version and artifact description of Marathon build.
  */
-def createAndUploadPackages(): (String, Option[awsClient.Artifact]) = {
+def createAndUploadPackages(upload: Boolean = true): (String, Option[awsClient.Artifact]) = {
```
zen-dog: This grew and got out of hand. We should refactor here a little. Lets do 2 methods:
- `createTarballPackage()`
- `uploadArtifatct(artifact: Artifact)`

ci/pipeline
```diff
@@ -75,13 +76,14 @@ def uploadTarballToS3(): Option[awsClient.Artifact] = utils.stage("Upload Packag
...
   val version = createPackages()
   val docsPath = PACKAGE_DOCS_DIR / s"marathon-docs-$version.tgz"
-  val artifact = uploadTarballToS3()
-  awsClient.upload(docsPath)
+  val artifact = if(upload) uploadTarballToS3() else None
+  if(upload) awsClient.upload(docsPath)
 
   createDockerAndLinux()
```
zen-dog: May be we should split that in `createDockerImage` and `createLinuxPackages`

ci/pipeline
zen-dog: And we should put all creation of packages (Tarball, Docker, Linux) into one stage.

ci/pipeline
```diff
@@ -199,14 +202,14 @@ def run(publish: Boolean = false): (String, Option[awsClient.Artifact]) = {
 @main
 def jenkins(): Unit = {
   if(utils.isPullRequest) {
```
zen-dog: What do you think about calling 2 distinct `@main` methods from here where each one of those should then call it's own targets? E.g. PR's don't publish artifacts and master builds do etc.

kensipe: we are likely to add another distinct `@main` as well... I'm expecting release to be it's own path.

zen-dog:
We just discussed this with Karsten and I'm ready to lower the acceptance bar 😄 But there will be a few minor improvements that Karsten will push soon

zen-dog:
Karsten's proposing that we should leverage `sbt` more extensively, effectively making most of this script into sbt's own targets.

ci/pipeline
jeschkies: Done.

PR #5706, opened 2017-10-27
Removed old workarounds in `ResidentTaskIntegrationTest`

zen-dog:
Summary:
We implemented a few workarounds in `ResidentTaskIntegrationTest` when we reused the taskIds. This is not necessary anymore.


unterstein:
Ok, no proof 😜

PR #5702, opened 2017-10-26
Rescue Billy

wavesoft:
It seems `user_billy` fixture was accidentally removed in a previous commit : https://github.com/mesosphere/marathon/compare/e3425ac9376975207799871f75b604550ba1dd5e...99a596785b77cac74173742d58ce7182f0ced328#diff-984c5562f7187dee9e6d3e161c7b52f8L51

This commit re-imports `user_billy` that would hopefully solve the failing SI tests.

PR #5707, opened 2017-10-30
Report error and warning count from flake8.

jeschkies:
Summary:
Some builds such as in #5702 have been confsing because the fail without
any apparent issue in the stderr logs. This change reports the flake8
error count to stderr. We should think about a better style reporting
for shakedown and flake8.


alenkacz:
@jeschkies are you sure it is printed to stderr? I know the documentation says so, but I think it was removed in 3.0.0 release https://github.com/PyCQA/flake8/blob/52180995046583cacf9d78a4423b82448a47fef1/docs/source/release-notes/3.0.0.rst

I tested it locally and it prints out the count of errors to stdout for me...

jeschkies:
@alenkacz thanks for pointing this out. We probably want to use something like [flake8-html](https://github.com/lordmauve/flake8-html). What do you think?

PR #5709, opened 2017-10-30
Small refactoring of matching for AppPathIdLike

alenkacz:
Summary:
Small refactoring of matching for AppPathIdLike.

JIRA issues:


alenkacz:
@meln1k I find this better readable than the previous one with the slash. But let me know what do you think...

alenkacz:
Will close this one because I found a case where it is not working correctly. See #5715

PR #5710, opened 2017-10-30
Import `user_billy` fixture in order to run SI tests.

kensipe:
Summary:
import is needed in order to have the fixture injected.  Added notes to the common file.

JIRA issues:  MARATHON-7892

### DIFF
--- a/tests/system/marathon_auth_common_tests.py
+++ b/tests/system/marathon_auth_common_tests.py
@@ -27,7 +27,8 @@ def test_non_authorized_user():
             error = exec_info.value
             assert str(error) == "You are not authorized to perform this operation"
 
-
+# NOTE:  this is a common test file.   All test suites which import this common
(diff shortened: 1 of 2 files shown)


unterstein:
I might be wrong, but it is possible that #5702 is doing the same thing?

kensipe:
Thanks @unterstein !!   That PR from @wavesoft doesn't work (at least with my testing).  this approach works.

tests/system/marathon_auth_common_tests.py
```diff
@@ -28,6 +28,8 @@ def test_non_authorized_user():
             assert str(error) == "You are not authorized to perform this operation"
 
 
+# NOTE:  this is a common test file. All test suites which import this common
```
jeschkies: Is there another way to fix this?

PR #5711, opened 2017-10-31
Detail of requested group

alenkacz:
Summary:
Returns detail of requested group.

JIRA issues: MARATHON-7301

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
@@ -5,7 +5,6 @@ package v2
 import akka.actor.ActorSystem
 import akka.http.scaladsl.server.{ Directive1, Route }
 import mesosphere.marathon.api.akkahttp.PathMatchers.GroupPathIdLike
-import mesosphere.marathon.api.akkahttp.v2.GroupsController._
 import mesosphere.marathon.api.v2.{ AppHelpers, PodsResource }
 import mesosphere.marathon.core.appinfo.{ AppInfo, GroupInfo, GroupInfoService, Selector }
 import mesosphere.marathon.core.election.ElectionService
@@ -97,11 +96,17 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
   }
   // format: On
 
+  /**
+    * Initializes rules for selecting groups to take authorization into account
+    */
   def authorizationSelectors(implicit identity: Identity): GroupInfoService.Selectors = {
     GroupInfoService.Selectors(
       AppHelpers.authzSelector,
       PodsResource.authzSelector,
       authzSelector)
+  }
+  private def authzSelector(implicit authz: Authorizer, identity: Identity) = Selector[Group] { g =>
+    authz.isAuthorized(identity, ViewGroup, g)
   }
 
   import mesosphere.marathon.api.v2.InfoEmbedResolver._
@@ -113,12 +118,5 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
     parameter('embed.*).tflatMap {
       case Tuple1(embeds) => provide(resolveAppGroup(if (embeds.isEmpty) defaultEmbeds else embeds.toSet))
     }
-  }
-}
-
-object GroupsController {
-
-  private def authzSelector(implicit authz: Authorizer, identity: Identity) = Selector[Group] { g =>
-    authz.isAuthorized(identity, ViewGroup, g)
   }
 }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
@@ -43,34 +43,34 @@ class GroupsController(electionService: ElectionService)(
             get {
               groupDetail(groupId)
             } ~
-              post {
-                createGroup(groupId)
-              } ~
-              put {
-                updateGroup(groupId)
-              } ~
-              delete {
-                deleteGroup(groupId)
-              }
+            post {
+              createGroup(groupId)
+            } ~
+            put {
+              updateGroup(groupId)
+            } ~
+            delete {
+              deleteGroup(groupId)
+            }
           }
         } ~
-          path(GroupPathIdLike ~ "apps" ~ Slash.? ~ PathEnd) { groupId =>
+        path(GroupPathIdLike ~ "apps" ~ Slash.? ~ PathEnd) { groupId =>
+          get {
+            appsList(groupId)
+          }
+        } ~
+        path(GroupPathIdLike / "versions") { groupId =>
+          pathEndOrSingleSlash {
             get {
-              appsList(groupId)
+              listVersions(groupId)
             }
           } ~
-          path(GroupPathIdLike / "versions") { groupId =>
-            pathEndOrSingleSlash {
-              get {
-                listVersions(groupId)
-              }
-            } ~
-              path(Remaining ~ Slash.? ~ PathEnd) { version =>
-                get {
-                  versionDetail(groupId, version)
-                }
-              }
+          path(Remaining ~ Slash.? ~ PathEnd) { version =>
+            get {
+              versionDetail(groupId, version)
+            }
           }
+        }
       }
     }
   }
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
@@ -1,0 +1,76 @@
+package mesosphere.marathon
+package api.akkahttp
+package v2
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.server.Route
+import mesosphere.marathon.api.akkahttp.PathMatchers.{ AppPathIdLike, GroupPathIdLike }
+import mesosphere.marathon.core.election.ElectionService
+import mesosphere.marathon.plugin.auth.{ Authorizer, Authenticator => MarathonAuthenticator }
+import mesosphere.marathon.state.PathId
+
+import scala.concurrent.ExecutionContext
+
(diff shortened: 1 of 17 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -20,8 +29,19 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
   import Directives._
+  import mesosphere.marathon.api.akkahttp.EntityMarshallers._
+  import mesosphere.marathon.api.v2.json.Formats._
 
-  def groupDetail(groupId: PathId): Route = ???
+  def groupDetail(groupId: PathId)(implicit identity: Identity): Route = {
+    extractEmbeds {
+      case (appEmbed, groupEmbed) =>
+        onSuccess(infoService.selectGroup(groupId, authorizationSelectors, appEmbed, groupEmbed)) {
+          case Some(info) => complete(Json.toJson(info))
```
jeschkies: Hm, shouldn't there be a RAML for `info`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -20,8 +29,19 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+          case None if groupId.isRoot => complete(Json.toJson(GroupInfo.empty))
```
jeschkies: Same here. I thought it would be `raml.GroupInfo.empty` or something similar.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -76,4 +96,29 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
     }
   }
   // format: On
+
+  def authorizationSelectors(implicit identity: Identity): GroupInfoService.Selectors = {
```
jeschkies: What is this method doing? Could you add docs?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -76,4 +96,29 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+      PodsResource.authzSelector,
+      authzSelector)
+  }
+
+  import mesosphere.marathon.api.v2.InfoEmbedResolver._
+  /**
+    * For backward compatibility, we embed always apps, pods, and groups if nothing is specified.
+    */
+  val defaultEmbeds = Set(EmbedApps, EmbedPods, EmbedGroups)
+  def extractEmbeds: Directive1[(Set[AppInfo.Embed], Set[GroupInfo.Embed])] = {
```
jeschkies: Should this be part of the `Directives.scala`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -76,4 +96,29 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+    parameter('embed.*).tflatMap {
+      case Tuple1(embeds) => provide(resolveAppGroup(if (embeds.isEmpty) defaultEmbeds else embeds.toSet))
+    }
+  }
+}
+
+object GroupsController {
```
jeschkies: What is the motivation for this object?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -1,26 +1,97 @@
...
+    }
+
+    "returns group detail" in {
+      val infoService = mock[GroupInfoService]
+      infoService.selectGroup(eq(PathId("/test/group")), any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId("/test/group")), None, None, None)))
+      val f = new Fixture(infoService = infoService)
+
+      Get(Uri./.withPath(Path("/test/group"))) ~> f.groupsController.route ~> check {
+        status should be(StatusCodes.OK)
+        responseAs[String] should include (""""id" : "/test/group"""")
```
jeschkies: I'm torn here. I've used the whole JSON string but find it too much. This string include is maybe not enough. How about parsing the raw json and query the id field? What do you think? I think I've seen this in other older tests.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -1,26 +1,97 @@
...
+      }
+    }
+
+    "returns empty group if root and no group found" in {
+      val infoService = mock[GroupInfoService]
+      infoService.selectGroup(any, any, any, any) returns Future.successful(None)
```
jeschkies: Should we match a specific parameter here?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: well there is not... :( I was hoping it was one of those types @meln1k was preparing because he is not using RAML type for this in AppsController too.

So what should we do? Keep it as it is? One of us should prepare the type? (which is potentially huge...)

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
jeschkies: I've found it in @meln1k's code ` (Json.parse(responseAs[String]) \ "tasks" \ 0 \ "id").get shouldEqual JsString(taskId)`. We could create some nice helper matchers.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: I would prefer not have this here since it is very specific to groups - it has specific defaults for groups and calls different embed resolver than e.g. apps controller

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: there is some weird concept of selectors in group-related methods... I will add comment but it won't make it that much clearer...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: I took this whole thing from Resource class, I guess it does not have to be an object

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
alenkacz: I matched at least the pathid, the rest would be too much implementation details in my opinion...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
alenkacz: thanks

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: for GroupInfo RAML, AppInfo RAML is needed - we don't have that and I think it should not be part of this diff

alenkacz:
@jeschkies ready for re-review

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
jeschkies: Ah, this is the AppInfo, right? Ok, let's just add a comment and a JIRA.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
jeschkies: You are right. Could you create a JIRA as I mentioned above?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: no this is groupInfo - but AppInfo is part of GroupInfo - ticket here https://jira.mesosphere.com/browse/MARATHON-7900

Once AppInfo is done, GroupInfo will be very easy...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: https://jira.mesosphere.com/browse/MARATHON-7900

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -1,26 +1,100 @@
...
+    }
+
+    "returns group detail" in {
+      val infoService = mock[GroupInfoService]
+      infoService.selectGroup(eq(PathId("/test/group")), any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId("/test/group")), None, None, None)))
+      val f = new Fixture(infoService = infoService)
+
+      Get(Uri./.withPath(Path("/test/group"))) ~> f.groupsController.route ~> check {
+        status should be(StatusCodes.OK)
+        (Json.parse(responseAs[String]) \ "id").get shouldEqual JsString("/test/group")
```
jeschkies: I like these. Thanks! The error message would be a little of but let's clean this up a different time. I'd have to revisit my older tests as well.

PR #5714, opened 2017-10-31
Automated Community Release Process

kensipe:
Automated Community Release Process

Summary:
* Full build and testing
* Docker and Linux Binary Builds with Testing
* Landing in the Release Folder on S3

This has been tested against building 1.5 and 1.6. 
It changes our releases defined here:  https://wiki.mesosphere.com/display/MARATHON/Release+a+1.5.x+Version+of+Marathon
and automates steps 1, 2, 3

We still need to add in native linux releases and mom-ee releases.
The rest will remain manual.

JIRA issues:  MARATHON-7902

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -33,7 +33,6 @@ ansiColor('xterm') {
             sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD}"""
             sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
           }
-        }
       } finally {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
         junit allowEmptyResults: true, testResults: 'target/test-reports/*integration/*.xml'
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -32,7 +32,9 @@ ansiColor('xterm') {
             string(credentialsId: '4551c307-10ae-40f9-a0ac-f1bb44206b5b',variable: 'DOCKER_HUB_EMAIL'),
         ]) {
           sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD} -e ${DOCKER_HUB_EMAIL}"""
-          sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
+          sshagent (credentials: ['0f7ec9c9-99b2-4797-9ed5-625572d5931d']) {
+            sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
+          }
         }
       } finally {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -1,0 +1,51 @@
+#!/usr/bin/env groovy
+
+ansiColor('xterm') {
+  node('JenkinsMarathonCI-Debian8-2017-10-23') {
+
+    properties([
+      parameters([
+        string(name: 'gitsha',
+          defaultValue: '',
+          description: 'Git Commit SHA to Build and Release (ex. 167d7fb86 or 167)'
+        ),
+        string(name: 'version',
+          defaultValue: '',
+          description: 'Community Release Version (ex. v1.5.2)'
+        ),
+        string(name: 'latest',
+          defaultValue: 'False',
+          description: 'True if docker mesosphere/marathon:latest should be pushed'
+        )]
+      )
+    ])
+
+    stage("Run Pipeline") {
+      try {
+        checkout scm
+        withCredentials([
+            usernamePassword(credentialsId: 'a7ac7f84-64ea-4483-8e66-bb204484e58f', passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USER'),
+            string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
+            string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY'),
+            string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d',variable: 'DOCKER_HUB_USERNAME'),
+            string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825',variable: 'DOCKER_HUB_PASSWORD'),
+            string(credentialsId: '4551c307-10ae-40f9-a0ac-f1bb44206b5b',variable: 'DOCKER_HUB_EMAIL'),
+        ]) {
+          sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD} -e ${DOCKER_HUB_EMAIL}"""
+          sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
+        }
+      } finally {
+        junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
+        junit allowEmptyResults: true, testResults: 'target/test-reports/*integration/*.xml'
+        publishHTML([
+            allowMissing: true, alwaysLinkToLastBuild: false, keepAll: true,
+            reportDir: 'target/scala-2.12/scapegoat-report', reportFiles: 'scapegoat.html',
+            reportName: 'Scapegoat Report', reportTitles: ''
+        ])
+        archive includes: "sandboxes.tar.gz"
+        archive includes: "ci-${env.BUILD_TAG}.log.tar.gz"
+        archive includes: "ci-${env.BUILD_TAG}.log"  // Only in case the build was  aborted and the logs weren't zipped
+      }
+    }
+  }
+}
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -27,14 +27,11 @@ ansiColor('xterm') {
             usernamePassword(credentialsId: 'a7ac7f84-64ea-4483-8e66-bb204484e58f', passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USER'),
             string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
             string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY'),
-            string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d',variable: 'DOCKER_HUB_USERNAME'),
-            string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825',variable: 'DOCKER_HUB_PASSWORD'),
-            string(credentialsId: '4551c307-10ae-40f9-a0ac-f1bb44206b5b',variable: 'DOCKER_HUB_EMAIL'),
         ]) {
-          sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD} -e ${DOCKER_HUB_EMAIL}"""
           sshagent (credentials: ['0f7ec9c9-99b2-4797-9ed5-625572d5931d']) {
+          withDockerRegistry([credentialsId: 'docker-hub-credentials']) {
             sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
-          }
+          }}
         }
       } finally {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -13,7 +13,7 @@ ansiColor('xterm') {
           defaultValue: '',
           description: 'Community Release Version (ex. v1.5.2)'
         ),
-        boolean(name: 'latest',
+        booleanParam(name: 'latest',
           defaultValue: false,
           description: 'True if docker mesosphere/marathon:latest should be pushed'
         )]
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -29,10 +29,11 @@ ansiColor('xterm') {
             string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY'),
             string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d',variable: 'DOCKER_HUB_USERNAME'),
             string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825',variable: 'DOCKER_HUB_PASSWORD'),
-          ]) {
-            sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD}"""
-            sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
-          }
+            string(credentialsId: '4551c307-10ae-40f9-a0ac-f1bb44206b5b',variable: 'DOCKER_HUB_EMAIL'),
+        ]) {
+          sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD} -e ${DOCKER_HUB_EMAIL}"""
+          sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
+        }
       } finally {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
         junit allowEmptyResults: true, testResults: 'target/test-reports/*integration/*.xml'
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -27,8 +27,10 @@ ansiColor('xterm') {
             usernamePassword(credentialsId: 'a7ac7f84-64ea-4483-8e66-bb204484e58f', passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USER'),
             string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
             string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY'),
-        ]) {
-          withDockerRegistry([credentialsId: 'docker-hub-credentials', url: 'https://index.docker.io/v1/']) {
+            string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d',variable: 'DOCKER_HUB_USERNAME'),
+            string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825',variable: 'DOCKER_HUB_PASSWORD'),
+          ]) {
+            sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD}"""
             sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
           }
         }
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -13,8 +13,8 @@ ansiColor('xterm') {
           defaultValue: '',
           description: 'Community Release Version (ex. v1.5.2)'
         ),
-        string(name: 'latest',
-          defaultValue: 'False',
+        boolean(name: 'latest',
+          defaultValue: false,
           description: 'True if docker mesosphere/marathon:latest should be pushed'
         )]
       )
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -27,12 +27,10 @@ ansiColor('xterm') {
             usernamePassword(credentialsId: 'a7ac7f84-64ea-4483-8e66-bb204484e58f', passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USER'),
             string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
             string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY'),
-            string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d',variable: 'DOCKER_HUB_USERNAME'),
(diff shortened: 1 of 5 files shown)


kensipe:
This works all the way thru uploading to s3 in the correct folders.  It has an issue with the Docker push currently.    I believe the docker credentials are the remaining component.

I also want to give the option to update the latest Docker.

After that... I want to push the tags back to the origin.

unterstein:
I`d love to review this pr and provide feedback but I think I am missing some information. Do we have a JIRA issue with more context on this one? :)

Jenkinsfile.release
```diff
@@ -0,0 +1,47 @@
...
+          defaultValue: '',
+          description: 'Git Commit SHA to Build and Release (ex. 167d7fb86 or 167)'
+        ),
+        string(name: 'version',
+          defaultValue: '',
+          description: 'Community Release Version (ex. v1.5.2)'
+        ),
+        string(name: 'latest',
+          defaultValue: 'False',
+          description: 'True if docker mesosphere/marathon:latest should be pushed'
```
jeschkies: 🤔 in which cases would this be false?

ci/awsClient.sc
```diff
@@ -39,9 +39,14 @@ case class Artifact(path: S3Path, sha1: String) {
   def downloadUrl: String = s"$base/${path.bucket}/${path.key}"
 }
 
-val S3_PREFIX = S3Path(
-  sys.env.getOrElse("S3_BUCKET", "downloads.mesosphere.io"),
-  Path(sys.env.getOrElse("S3_PATH" , "/marathon/snapshots")))
+/**
+ * builds = snapshots / builds, milestones, releases
```
jeschkies: Why `/ builds`?

Why not
```
/**
  * @return S3 path to buildLocation in S3 bucket
  */
```

ci/awsClient.sc
```diff
@@ -67,32 +72,32 @@ def doesS3FileExist(path: S3Path): Boolean = {
  *
  *  @return Artifact description if it was uploaded. None otherwise.
  */
-def archiveArtifact(uploadFile: Path): Option[Artifact] = {
+def archiveArtifact(uploadFile: Path, s3path: S3Path): Option[Artifact] = {
   // is already uploaded.
-  if(doesS3FileExist(S3_PREFIX / uploadFile.last)) {
-    println(s"Skipping File: ${uploadFile.last} already exists on S3 at ${S3_PREFIX / uploadFile.last}")
+  if(doesS3FileExist(s3path / uploadFile.last)) {
```
jeschkies: Great catch. Would you mind adding the parameter descriptions to the docstring?

ci/awsClient.sc
```diff
@@ -67,32 +72,32 @@ def doesS3FileExist(path: S3Path): Boolean = {
...
-    Some(uploadFileAndSha(uploadFile))
+    Some(uploadFileAndSha(uploadFile, s3path))
 }
 
 /**
  *  Uploads marathon artifacts to the default bucket, using the env var credentials.
  *  Upload process creates the sha1 file
  */
-def uploadFileAndSha(uploadFile: Path): Artifact = {
+def uploadFileAndSha(uploadFile: Path, s3path: S3Path): Artifact = {
```
jeschkies: Thanks. Would you mind adding the parameters to the docstring?

ci/pipeline
```diff
@@ -38,11 +39,12 @@ def compileAndTest(logFileName: String): Unit = utils.stage("Compile and Test")
   run("sbt", "plugin-interface/compile")
 
   // Check system integraion tests.
-  checkSystemIntegrationTests(logFileName)
+  // checkSystemIntegrationTests(logFileName)
```
jeschkies: Why is the commented?

ci/pipeline
```diff
@@ -38,11 +39,12 @@ def compileAndTest(logFileName: String): Unit = utils.stage("Compile and Test")
...
 }
 
 @main
 def checkSystemIntegrationTests(logFileName: String): Unit = {
+  // TODO: (kgs)  There is NO need to kill processes for this!
```
jeschkies: I don't understand the TODO. Is the todo to remove the timeout?

So, I've left it because the `runWithTimeout` redirect stdout into a file. The name is probably not accurate anymore. We could introduce a infinite timeout. What do you think?

ci/pipeline
```diff
@@ -60,21 +62,21 @@ def zipLogs(logFileName: String = "ci.log"): Unit = {
...
 @main
 /**
- * Upload Marathon tgz tarballs, its cha1 checksum and docs to S3.
+ * Upload Marathon tgz tarballs, its sha1 checksum and docs to S3.
  *
  * @param version The version to upload.
  * @return Artifact description if it was uploaded.
  */
-def uploadTarballPackagesToS3(version: String): Option[awsClient.Artifact] = utils.stage("Upload Tarball Packages") {
+def uploadTarballPackagesToS3(version: String, buildLocation: String): Option[awsClient.Artifact] = utils.stage("Upload Tarball Packages") {
```
jeschkies: Please document the parameter.

ci/pipeline
```diff
@@ -92,14 +94,18 @@ def createTarballPackages(): String = utils.stage("Package Tarballs") {
   // ending with random characters
   // we need to regex this string because we do have colored output in the `sbt version` command
   val VersionLineRegex = "^.*(\\d+\\.\\d+\\.\\d+[-A-Za-z\\d]+).*$".r
+  // release version example v1.5.2
+  val releaseVersionLineRegex = "^.*(\\d+\\.\\d+\\.\\d+).*$".r
```
jeschkies: Nit: Why is this lower case when `VersionLineRegex` is upper case?

ci/pipeline
```diff
@@ -168,11 +174,19 @@ def build(): String = {
...
+def buildDockerAndLinuxPackages(): Unit = {
   utils.stage("Package Docker Image, Debian and RedHat Packages") {
     %('sbt, "docker:publishLocal", "packageLinux")
-  }
 
-  version
+    // create test docker images and tests packages
+    val testPath = pwd/'tests/'package
+    %('make, "all")(testPath)
+    %('amm, "test.sc", "all")(testPath)
```
jeschkies: Hm, I don't think `test.sc` would be found. Should it be something like `%('amm, testPath / "test.sc", "all")(testPath)`.

Also, it seems weird to call ammonite from ammonite. I'd actually expect `make all test`. That seems cleaner to me. What do you think?

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+@main
+def release(releaseVersion: String, gitSha: String, latest: String = "False"): Unit = {
+  println(s"Releasing version: $releaseVersion ")
+  %('git, "checkout", gitSha)
+  %('git, "tag", "-a", releaseVersion, "-m", releaseVersion)
+  val version = build()
+
+  // Uploads
+  uploadTarballPackagesToS3(version, s"releases/$releaseVersion")
+  // push docker
```
jeschkies: Nit: I don't think this comment is required since the command actually says `docker push` 🙂.

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+  %('docker, "push", s"mesosphere/marathon:$releaseVersion")
```
jeschkies: Isn't sbt offering a build target?

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+ *  3. git tag of version (pushed to server)
+ *
+ * @return Version and artifact description of Marathon build.
+ */
+@main
+def release(releaseVersion: String, gitSha: String, latest: String = "False"): Unit = {
+  println(s"Releasing version: $releaseVersion ")
+  %('git, "checkout", gitSha)
+  %('git, "tag", "-a", releaseVersion, "-m", releaseVersion)
+  val version = build()
```
jeschkies: 🤔 hm, it feels odd that we pass a release version and then have something returned by `build()`. I know the background but am wondering if we could avoid this. Also `version` is not used in this method.

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+
+  // Uploads
+  uploadTarballPackagesToS3(version, s"releases/$releaseVersion")
```
jeschkies: Shouldn't this be `uploadTarballPackagesToS3(version, s"releases/$version")`?

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+  // push docker
+  %('docker, "push", s"mesosphere/marathon:$releaseVersion")
+
+  // TODO: if latest == True
+  // docker tag :latest
+  // %('docker, "push", "mesosphere/marathon:latest")
```
jeschkies: Why is the commented?

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+
+  // push tags
+  // %('git, "push", "--tags")
```
jeschkies: Why not activate it?

ci/pipeline
```diff
@@ -221,3 +235,34 @@ def jenkins(): Unit = {
...
+
+  // TODO: Publish swagger files.
+  // TODO: Publish native packages to unstable.
```
jeschkies: Hm, I guess this should be
> // TODO: Publish native packages to ~un~stable.

🙂

ci/utils.sc
```diff
@@ -94,6 +93,20 @@ def runWithTimeout(timeout: FiniteDuration, logFileName: String)(commands: Seq[S
...
+
+/**
+ * Run a process with given commands and time out it runs too long.
+ *
+ * @param timeout The maximum time to wait.
+ * @param logFileName Name of file which collects all logs.
+ * @param commands The commands that are executed in a process. E.g. "sbt",
+ *  "compile".
+ */
+def runWithTimeoutAndCleanup(timeout: FiniteDuration, logFileName: String)(commands: Seq[String]): Unit = {
```
jeschkies: What is the motivation for this?

Jenkinsfile.release
kensipe: Today when we release 1.5.2 it will be True.   Tomorrow when we release 1.6.0 it will be True.   in 3 days when release 1.5.3, it will be False.

kensipe:
Example Release:  https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/ken/job/community-release/15/console

ci/awsClient.sc
kensipe: It is not clear what is being asked.     Our release process: https://docs.google.com/document/d/1BRFbCjgu0N4nwr4f9daxJLIr_j9nqI19u5CAS-0jQio/edit#
indicates that we will have regular master builds with have been called snaps, snapshots and not pre-X, which currently land in a snapshots directory.   We are going to move those to `builds` directory on S3.   Planning to get rid of snapshots completely.   the word carries meaning which is not true for us.

ci/pipeline
kensipe: 1st... it works as written.   the `(testPath)` sets the working directory... which took some time for me to figure out.   As this is the working dir of the invocation the `test.sc` script all is good.

2nd.   I was codifying what our manual release process has been.    We could add a test target to the Makefile.   I discussed the need to call ammonite from ammonite with @timcharper seemed like the thing to do.    Changing it to make adds a layer of indirection but doesn't change the fact we are calling amm from amm.
I'm good the way it is... but totally willing to change it if you feel strongly about it.

ci/utils.sc
kensipe: because the call for flake8 is a python thing that can't possibly need Java process cleanup.  It was likely called with it out of convenience to reuse what was there and gain the benefits of a timeout... 
this is better.

kensipe:
keep getting CI failures with
```

File: Jenkinsfile.release (hunk 1 above)
Comment (jeschkies) [reply Jenkinsfile.release]: Ah, that makes a lot of sense. Thanks.

File: ci/awsClient.sc (hunk 2 above)
Comment (jeschkies) [reply ci/awsClient.sc]: That was the missing information. I'd make it `builds` then. No more `snapshots`.

File: Jenkinsfile.release [hunk 18]
```diff
@@ -0,0 +1,51 @@
...
+        checkout scm
+        withCredentials([
+            usernamePassword(credentialsId: 'a7ac7f84-64ea-4483-8e66-bb204484e58f', passwordVariable: 'GIT_PASSWORD', usernameVariable: 'GIT_USER'),
+            string(credentialsId: '3f0dbb48-de33-431f-b91c-2366d2f0e1cf',variable: 'AWS_ACCESS_KEY_ID'),
+            string(credentialsId: 'f585ec9a-3c38-4f67-8bdb-79e5d4761937',variable: 'AWS_SECRET_ACCESS_KEY'),
+            string(credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d',variable: 'DOCKER_HUB_USERNAME'),
+            string(credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825',variable: 'DOCKER_HUB_PASSWORD'),
+            string(credentialsId: '4551c307-10ae-40f9-a0ac-f1bb44206b5b',variable: 'DOCKER_HUB_EMAIL'),
+        ]) {
+          sh """docker login -u ${DOCKER_HUB_USERNAME} -p ${DOCKER_HUB_PASSWORD} -e ${DOCKER_HUB_EMAIL}"""
```
jeschkies: Checkout the [Jenkinsfile from marathon-dcos-plugins](https://github.com/mesosphere/marathon-dcos-plugins/blob/master/Jenkinsfile#L19). We used

```
withDockerRegistry([credentialsId: 'docker-hub-credentials', url: 'https://index.docker.io/v1/']) {
  sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
}
```

that seems a little cleaner to me.

ci/pipeline
```diff
@@ -43,6 +44,7 @@ def compileAndTest(logFileName: String): Unit = utils.stage("Compile and Test")
 
 @main
 def checkSystemIntegrationTests(logFileName: String): Unit = {
+  // TODO: (kgs)  There is NO need to kill processes for this!
```
jeschkies: [Nit]: I guess the todo is done, right?

ci/pipeline
```diff
@@ -60,21 +62,22 @@ def zipLogs(logFileName: String = "ci.log"): Unit = {
 
 @main
 /**
- * Upload Marathon tgz tarballs, its cha1 checksum and docs to S3.
+ * Upload Marathon tgz tarballs, its sha1 checksum and docs to S3.
  *
  * @param version The version to upload.
+ * @param buildLocation subfolder location to upload tarball to.  Example: "builds"
```
jeschkies: [Super duper nit]: Please use French  instead of English spacing.

ci/pipeline
jeschkies: Thanks for enlighten me on the work path. That makes sense now.

As to the ammonite call I'm on the fence. To me this file is like a meta build tool that triggers other build tools such as sbt and flake8. In that sense the build tool for testing the Docker image would be `make`. However, I don't think it's a blocker.

ci/pipeline
```diff
@@ -183,7 +198,7 @@ def master(): Unit = {
   val version = build()
 
   // Uploads
-  val maybeArtifact = uploadTarballPackagesToS3(version)
+  val maybeArtifact = uploadTarballPackagesToS3(version, "snapshots")
```
jeschkies: Should we change it to `builds` then?

ci/pipeline
```diff
@@ -199,7 +214,7 @@ def pr(): Unit = asPullRequest {
   val version = build()
 
   // Uploads
-  val artifact = uploadTarballPackagesToS3(version)
+  val artifact = uploadTarballPackagesToS3(version, "snapshots")
```
jeschkies: I have to read up on the doc but should this be `builds` as well? I guess not.

ci/pipeline
```diff
@@ -221,3 +236,45 @@ def jenkins(): Unit = {
...
+/**
+ * Executes the Community Release which includes:
+ *  1. tarball with version details
+ *  2. release of docker with that version
+ *  3. git tag of version (pushed to server)
+ *
+ * @return Version and artifact description of Marathon build.
+ */
+@main
+def release(requestVersion: String, gitSha: String, latest: String = "False"): Unit = {
```
jeschkies: The docs are already good but could you add docs for the parameters?

ci/pipeline
```diff
@@ -221,3 +236,45 @@ def jenkins(): Unit = {
...
+
+  val releaseVersion = cleanVersion(requestVersion)
+  val tagVersion = s"v$releaseVersion"
+  println(s"Releasing version: $releaseVersion ")
+  %('git, "checkout", gitSha)
+  %('git, "tag", "-a", tagVersion, "-m", tagVersion)
+  val version = build()
+
+  uploadTarballPackagesToS3(version, s"releases/$releaseVersion")
```
jeschkies: As I mentioned before, should this be `s"releases/$version"`?

ci/pipeline
```diff
@@ -221,3 +236,45 @@ def jenkins(): Unit = {
...
+
+  %('docker, "push", s"mesosphere/marathon:$tagVersion")
+
+  if (latest.toUpperCase == "TRUE") {
```
jeschkies: Hm, how about making `latest` a boolean?

ci/pipeline
```diff
@@ -221,3 +236,45 @@ def jenkins(): Unit = {
...
+  val releaseVersion = cleanVersion(requestVersion)
+  val tagVersion = s"v$releaseVersion"
+  println(s"Releasing version: $releaseVersion ")
+  %('git, "checkout", gitSha)
+  %('git, "tag", "-a", tagVersion, "-m", tagVersion)
+  val version = build()
+
+  uploadTarballPackagesToS3(version, s"releases/$releaseVersion")
+
+  %('docker, "push", s"mesosphere/marathon:$tagVersion")
```
jeschkies: What are the differences between `tagVersion`, `version` and `releaseVersion`? Shouldn't they all be the same?

ci/pipeline
```diff
@@ -221,3 +236,45 @@ def jenkins(): Unit = {
...
+
+  if (latest.toUpperCase == "TRUE") {
+    %('docker, "tag", s"mesosphere/marathon:$tagVersion", "mesosphere/marathon:latest")
+    %('docker, "push", "mesosphere/marathon:latest")
+  } else {
+    println("Docker image mesosphere/marathon:latest NOT updated")
+  }
+
+  %('git, "push", "--tags")
```
jeschkies: We both talked about this before, what happens if the tag already exists? Is there a way to skip the uploads? How can we make it idempotent?

ci/pipeline
```diff
@@ -221,3 +236,45 @@ def jenkins(): Unit = {
...
+
+  // TODO: Publish swagger files.
+  // TODO: Publish native packages to unstable.
+}
+
+def cleanVersion(releaseVersion: String): String = {
```
jeschkies: I'd call it `withoutVersionPrefix` 🙂.

ci/utils.sc
jeschkies: Ok, makes sense. How about this

```scala
def runWithTimeout(timeout: FiniteDuration, logFileName: String)(commands: Seq[String]): Unit = ???

def withCleanUp[T](block :=> T): T {
  try { block }
  finally { provision.killStaleTestProcesses() }
}

def run(cmd: String *) = withCleanUp { utils.runWithTimeoutAndCleanup(30.minutes, logFileName)(cmd) }
```

Jenkinsfile.release
kensipe: I agree!!  on it!

ci/pipeline
kensipe: yeah... this was a PR for releases...   I don't know the complete consequence of changing the name of snapshots and thought it would be follow up work (like native packages).   This is likely really really simple and I'll change it.

ci/pipeline
kensipe: same as above... and yeah.

ci/pipeline
kensipe: there is a lot of "versions" :) .   if using $version the tarball would land in `/releases/v1.6.0/tarball` and we want `releases/1.6.0/tarball`

ci/pipeline
kensipe: the tag early in the the process fails if the tag exists and will not reach this point.

kensipe:
JIRA added:  https://jira.mesosphere.com/browse/MARATHON-7902

ci/pipeline
kensipe: `requestVersion` is the request attempt and will be cleaned to be the `releaseVersion`
`releaseVersion`  is the version with out the `v` prefix
`tagVersion` is the version with the `v` prefix
`version` could go away... it is the version determined by the underlying build which should be the same as tagVersion.   It is there being consistent with the other builds (master and pr) which require it because the version isn't passed in but is calculated.

ci/pipeline
kensipe: much better!

kensipe:
@jeschkies  All feedback addressed

ci/pipeline
jeschkies: Ok, we can follow up if we want to.

ci/pipeline
jeschkies: We should document this 🙂.

ci/pipeline
```diff
@@ -221,3 +235,47 @@ def jenkins(): Unit = {
...
+ * @param requestVersion The version attempting to be released v1.4.6
+ * @param gitSha The git commit sha.  This can be shorthand (ex. 0e1)
+ * @param latest Boolean to indicat if the Docker latest should be updated to this version.
+ * @return Version and artifact description of Marathon build.
+ */
+@main
+def release(requestVersion: String, gitSha: String, latest: Boolean = false): Unit = {
+
+  val releaseVersion = withoutVersionPrefix(requestVersion)
+  val tagVersion = s"v$releaseVersion"
```
jeschkies: Hehe, ok, this reads funny now:

```scala
val releaseVersion = withoutVersionPrefix(requestVersion)
val tagVersion = s"v$releaseVersion"
```

one would wonder if `requestVersion == tagVersion`. I guess `withoutVersionPrefix` jus removes a prefix if there is one.

ci/utils.sc
```diff
@@ -94,10 +93,11 @@ def runWithTimeout(timeout: FiniteDuration, logFileName: String)(commands: Seq[S
       val cmd = commands.mkString(" ")
       throw new java.util.concurrent.TimeoutException(s"'$cmd' timed out after $timeout.")
     }
-  } finally {
-    // This also cleans forked SBT processes.
-    provision.killStaleTestProcesses()
-  }
+}
+
+def withCleanUp[T](block: => Unit): Unit = {
```
jeschkies: [Nit] Please add a docstring.

jeschkies: Either

```scala
def withCleanUp(block: => Unit): Unit
```

or

```scala
def withCleanUp[T](block: => T): T
```

ci/pipeline
timcharper: @jeschkies that's not a bad idea! We could create a "make test" target which builds all necessary docker images and then runs the tests.

timcharper: Looks like that was already done :)

PR #5715, opened 2017-11-01
Another test verifying app matcher

alenkacz:
Summary:
In the end I found a case where #5709 was not behaving correctly. Just documented that in this unit test and will discard the #5709

JIRA issues:

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/akkahttp/PathMatchersTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/akkahttp/PathMatchersTest.scala
@@ -66,6 +66,10 @@ class PathMatchersTest extends UnitTest with GroupCreation with ScalatestRouteTe
     "considers it an unmatch if path starts with keyword" in {
       AppPathIdLike(Path("/restart")) shouldBe Unmatched
     }
+
+    "match path with multiple slashes" in {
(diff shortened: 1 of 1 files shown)


src/test/scala/mesosphere/marathon/api/akkahttp/PathMatchersTest.scala
```diff
@@ -66,6 +66,10 @@ class PathMatchersTest extends UnitTest with GroupCreation with ScalatestRouteTe
     "considers it an unmatch if path starts with keyword" in {
       AppPathIdLike(Path("/restart")) shouldBe Unmatched
     }
+
+    "match path with multiple slashes" in {
+      AppPathIdLike(Path("//////app")) shouldBe Matched(Path.Empty, Tuple1("//////app".toPath))
```
meln1k: I'm not sure if this is the behaviour we want to have. Let's ask @jeschkies

alenkacz: well ... I tested it on the current API and it behaves like this... So I guess we must be backward compatible ... ?

Also what PathId does is that there is sanitization that actually removes all those //// and normalizes it to a correct PathId with just one /

alenkacz:
@meln1k I changed the assert - now it is more obvious what is it actually testing - because there is sanitization happening on the background

PR #5722, opened 2017-11-03
DeploymentSteps RAML type added, dryRun parameter documented MARATHON-3317

alenkacz:
Summary:
While implementing groups controller, I found a parameter that was not documented.

JIRA issues: MARATHON-3317


alenkacz:
@meln1k well I can't test RAML and this is adding just RAML...

I'll add more tests once I submit the implementation. Sounds ok?

alenkacz:
@meln1k by the way it is not new parameter. It is there since 0.8.0 just wasn't documented

alenkacz:
@jeschkies I asked both to review and they did, can you change your rejection now, pls? :)

PR #5724, opened 2017-11-04
Workaround MESOS-8171 by setting failover_timeout to non-zero in tests

timcharper:
JIRA Issues: MARATHON-7793


timcharper:
This will make our test cleanup cycle slightly slower. Overall, though, considering how frequently we hit the infinite loop issue, we are not that much slower.

We cannot use a value of 0 until MESOS-8171 is resolved.

PR #5726, opened 2017-11-06
Remove duplicate imports and unused prameters.

jeschkies:
Summary:
Following `sbt scapegoat` this removes quite a lot of unused imports and
parameters. This makes the linter output a little easier to comprehend.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -8,45 +8,27 @@ import akka.event.EventStream
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.{ StatusCodes, Uri }
 import akka.http.scaladsl.model.headers.Location
-import akka.http.scaladsl.model.{ StatusCodes, Uri }
-import akka.http.scaladsl.model.headers.Location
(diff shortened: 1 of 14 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -81,7 +63,7 @@ class AppsController(
   import Directives._
 
   private implicit lazy val validateApp = AppDefinition.validAppDefinition(config.availableFeatures)(pluginManager)
-  private implicit lazy val updateValidator = AppValidation.validateCanonicalAppUpdateAPI(config.availableFeatures, () => normalizationConfig.defaultNetworkName)
+  //  private implicit lazy val updateValidator = AppValidation.validateCanonicalAppUpdateAPI(config.availableFeatures, () => normalizationConfig.defaultNetworkName)
```
alenkacz: why is this just commented out and not removed?

zen-dog: I guess we'll need it soon?

jeschkies: No, it should go. I just tested whether it would work without. I'll follow up.

PR #5728, opened 2017-11-07
Implement list of versions, version detail and create group MARATHON-7301

alenkacz:
Summary:
Implement three more methods - list of versions, detail of version and create group.
As a part of this I did some refactoring in the Resource class to be able to re-use some logic there.

JIRA issues: MARATHON-7301


src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -0,0 +1,71 @@
...
+class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorizer, executionContext: ExecutionContext) {
+
+  /**
+    * Encapsulates the group update logic that is following:
+    * - if version is set, group is reverted to that version
+    * - if scaleBy is set, group is scaled up
+    * - if none version nor scaleBy is set then the group is updated
+    */
+  @SuppressWarnings(Array("all")) // async/await
+  def getUpdatedGroup(
```
alenkacz: this is a piece of code I did not alter much just took it from the resource class and move it to service class so that I can reuse it between resource and controller.

I also added tests for this logic because it feels to me that this is quite complex business logic but it was not covered by tests at all.

In the future, we can talk about using Either or actually do this in more functional way...

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -0,0 +1,71 @@
...
+    /**
+      * roll back to a previous group version
+      */
+    def revertToOlderVersion: Future[Option[RootGroup]] = groupUpdate.version match {
+      case Some(version) =>
+        val targetVersion = Timestamp(version)
+        checkAuthorizationOrThrow(UpdateGroup, group)
+        groupManager.group(group.id, targetVersion)
+          .filter(checkAuthorizationOrThrow(ViewGroup, _))
+          .map(g => g.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion")))
```
meln1k: What do you think about having only one `.map` here?

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
meln1k: The name of this method is a bit confusing. It says get something, and performs side effects in the same time.

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -0,0 +1,71 @@
...
+    * Encapsulates the group update logic that is following:
+    * - if version is set, group is reverted to that version
+    * - if scaleBy is set, group is scaled up
+    * - if none version nor scaleBy is set then the group is updated
+    */
+  @SuppressWarnings(Array("all")) // async/await
+  def getUpdatedGroup(
+    rootGroup: RootGroup,
+    groupId: PathId,
+    groupUpdate: raml.GroupUpdate,
```
meln1k: What do you thing about separating raml and domain objects?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -163,6 +177,7 @@ object EntityMarshallers {
   implicit val enrichedTaskMarshaller = playJsonMarshaller[raml.EnrichedTask]
   implicit val enrichedTasksListMarshaller = playJsonMarshaller[raml.EnrichedTasksList]
   implicit val deleteTasksUnmarshaller = playJsonUnmarshaller[raml.DeleteTasks]
+  implicit val seqDateTimeMarshalled = playJsonMarshaller[Seq[Timestamp]]
```
meln1k: Interesting. So it can't figure out how to serialize `Seq[T]` given a marshaller for type `T`?

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
alenkacz: that would be nice, but it would require bigger refactoring because some code inside those methods actually expect these raml types. Btw. this is not really a domain service, I intentionally placed it to the api package, so that it is very close to API.

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
alenkacz: that would be nice, but it would require bigger refactoring because some code inside those methods actually expect these raml types. Btw. this is not really a domain service, I intentionally placed it to the api package, so that it is very close to API.

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
alenkacz: apparently no :( any idea where is the problem? If I remove it, this is the error

found   : scala.collection.immutable.Seq[mesosphere.marathon.state.Timestamp]
[error]  required: akka.http.scaladsl.marshalling.ToResponseMarshallable
[error]             complete(versions)

alenkacz:
@meln1k all your comments addressed... you can re-review :)

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
zen-dog: I'm not sure if we should already introduce service classes. So far we've been copy-pasting existing resource logic into controllers (as bad as it is) with an option of extracting this later.

alenkacz: well this code is quite complex and contains a lot of logic that should be tested... and you just cannot test it that easy without extracting it to a separate class...

But I can remove the class and delete the tests if you all think it is not a good idea

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.async.Async._
+import scala.concurrent.{ ExecutionContext, Future }
+
+class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorizer, executionContext: ExecutionContext) {
+
+  /**
+    * Encapsulates the group update logic that is following:
+    * - if version is set, group is reverted to that version
+    * - if scaleBy is set, group is scaled up
+    * - if none version nor scaleBy is set then the group is updated
```
jeschkies: Nit: `if neither a version nor scaleBy is set`

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -0,0 +1,72 @@
...
+    */
+  @SuppressWarnings(Array("all")) // async/await
+  def updateGroup(
+    rootGroup: RootGroup,
+    groupId: PathId,
+    groupUpdate: raml.GroupUpdate,
+    newVersion: Timestamp)(implicit identity: Identity): Future[RootGroup] = async {
+    val group = rootGroup.group(groupId).getOrElse(Group.empty(groupId))
+
```
jeschkies: How about calling `checkAuthorizationOrThrow(UpdateGroup, group)` here? What do you think?

src/main/scala/mesosphere/marathon/api/akkahttp/EntityMarshallers.scala
```diff
@@ -132,6 +134,18 @@ object EntityMarshallers {
       }
   }.handleValidationErrors
 
+  def groupUpdateUnmarshaller(groupId: PathId)(implicit normalization: Normalization[mesosphere.marathon.raml.App]): FromEntityUnmarshaller[GroupUpdate] = {
```
jeschkies: Hm, I avoided such a marshaller in the pods controller and called the validation and normalization explicitly. What is the benefit of such a marshaller?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -3,29 +3,53 @@ package api.akkahttp
...
+    val materializer: Materializer
 ) extends Controller {
   import Directives._
   import mesosphere.marathon.api.akkahttp.EntityMarshallers._
   import mesosphere.marathon.api.v2.json.Formats._
 
+  private val forceParameter = parameter('force.as[Boolean].?(false))
+
+  /** convert app to canonical form */
+  private implicit val appNormalization: Normalization[raml.App] = {
```
jeschkies: I'd go for an explicit normalization as in the pods controller. These implict normalizations are hard to track.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -39,50 +63,101 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
 
   def appsList(groupId: PathId): Route = ???
 
-  def createGroup(groupId: PathId): Route = ???
+  def createGroup(groupId: PathId)(implicit identity: Identity): Route = {
+    (forceParameter & entity(as(groupUpdateUnmarshaller(groupId)))) { (force, groupUpdate) =>
```
jeschkies: Could you add a type to `groupUpdate`. Is it `raml.GroupUpdate`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -39,50 +63,101 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+          case Success(deploymentResult) => complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(deploymentResult.deploymentId)), deploymentResult))
+          case Failure(ex: IllegalArgumentException) => complete(StatusCodes.UnprocessableEntity -> ex.getMessage)
+          case Failure(ex) => throw ex
+        }
+      }
+    }
+  }
+
+  /**
+    * Until there is async version of updateRoot we have to block here
```
jeschkies: I don't quite understand. Why are we blocking?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -39,50 +63,101 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+    */
+  protected def result[T](fn: Awaitable[T]): T = Await.result(fn, config.zkTimeoutDuration)
+
+  @SuppressWarnings(Array("all")) // async/await
+  private def updateOrCreate(id: PathId, update: raml.GroupUpdate, force: Boolean)(implicit identity: Identity): Future[DeploymentResult] = async {
+    val version = Timestamp.now()
+
+    val effectivePath = update.id.map(PathId(_).canonicalPath(id)).getOrElse(id)
+    val deploymentPlan = await(groupManager.updateRoot(
+      id.parent, group => result(groupApiService.updateGroup(group, effectivePath, update, version)), version, force))
```
jeschkies: Are we blocking because the lambda cannot handle asynchronous calls?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -39,50 +63,101 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+            complete(versions)
+          }
+        }
+      case None => reject(EntityNotFound.noGroup(groupId))
+    }
+  }
 
-  def versionDetail(groupId: PathId, version: String): Route = ???
+  def versionDetail(groupId: PathId, version: Timestamp)(implicit identity: Identity): Route = extractEmbeds {
+    case (appEmbed, groupEmbed) =>
```
jeschkies: Why `case` here?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -39,50 +63,101 @@ class GroupsController(electionService: ElectionService, infoService: GroupInfoS
...
+                updateGroup(groupId)
+              } ~
+              delete {
+                deleteGroup(groupId)
+              }
             }
           }
         } ~
-        path(GroupPathIdLike ~ "apps" ~ Slash.? ~ PathEnd) { groupId =>
+        path(GroupPathIdLike ~ Slash.? ~ "apps" ~ Slash.? ~ PathEnd) { groupId =>
```
jeschkies: I think there is a `PathEndOrSlash`, isn't there?

src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
```diff
@@ -17,7 +19,7 @@ import scala.concurrent.ExecutionContext
 class GroupManagerModule(
     config: GroupManagerConfig,
     scheduler: Provider[DeploymentService],
-    groupRepo: GroupRepository)(implicit ctx: ExecutionContext, eventStream: EventStream) {
+    groupRepo: GroupRepository)(implicit ctx: ExecutionContext, eventStream: EventStream, authorizer: Authorizer) {
```
jeschkies: Where is the authorizer used?

src/test/scala/mesosphere/marathon/api/GroupApiServiceTest.scala
```diff
@@ -0,0 +1,97 @@
...
+    Given("Group manager with no group of required version")
+    val groupManager = mock[GroupManager]
+    val groupId = PathId.empty
+    val version = Timestamp.now
+    groupManager.group(groupId, version).returns(Future.successful(None))
+    val f = Fixture(groupManager = groupManager)
+
+    When("Calling update with version provided")
+    Then("Exception will be thrown")
+    val ex = f.groupApiService.updateGroup(createRootGroup(), PathId.empty, GroupUpdate(version = Some(version.toOffsetDateTime)), version).failed.futureValue
```
jeschkies: You could also do

```scala
an[IllegalArgumentException] should be thrownBy {
  f.groupApiService
   .updateGroup(createRootGroup(), 
                PathId.empty, 
                GroupUpdate(version = Some(version.toOffsetDateTime)),
                version)
   .futureValue
}
```

src/test/scala/mesosphere/marathon/api/GroupApiServiceTest.scala
```diff
@@ -0,0 +1,97 @@
...
+  "scale when scaleBy provided" in {
+    Given("Initialized service with root group and one app")
+    val f = Fixture()
+    val app = AppDefinition("/app".toRootPath, cmd = Some("cmd"), networks = Seq(ContainerNetwork("foo")))
+    val originalInstancesCount = app.instances
+    val rootGroup = createRootGroup(apps = Map(
+      "/app".toRootPath -> app
+    ))
+    When("Calling update with scaleBy")
+    val updatedGroup = f.groupApiService.updateGroup(rootGroup, PathId.empty, GroupUpdate(scaleBy = Some(2)), Timestamp.now()).futureValue
```
jeschkies: Nit: The line is a little to long.

src/test/scala/mesosphere/marathon/api/GroupApiServiceTest.scala
```diff
@@ -0,0 +1,97 @@
...
+    val groupManager = mock[GroupManager]
+    val groupId = PathId.empty
+    val version = Timestamp.now
+    groupManager.group(groupId, version).returns(Future.successful(None))
+    val f = Fixture(groupManager = groupManager)
+
+    When("Calling update with version provided")
+    Then("Exception will be thrown")
+    val ex = f.groupApiService.updateGroup(createRootGroup(), PathId.empty, GroupUpdate(version = Some(version.toOffsetDateTime)), version).failed.futureValue
+    ex shouldBe a[IllegalArgumentException]
```
jeschkies: Nit: `an[IllegalArgumentException]`

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -62,6 +80,116 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+    "list all versions of given group" in {
+      val clock = new SettableClock()
+      val groupManager = mock[GroupManager]
+      val groupVersions = Seq(clock.now(), clock.now())
+      groupManager.versions("groupname".toRootPath) returns Source(groupVersions)
+      groupManager.group("groupname".toRootPath) returns Some(createGroup(PathId.empty))
+      val f = new Fixture(groupManager = groupManager)
+
+      Get(Uri./.withPath(Path("/groupname/versions"))) ~> f.groupsController.route ~> check {
+        responseAs[String] should be ("[ \"2015-04-09T12:30:00.000Z\", \"2015-04-09T12:30:00.000Z\" ]")
```
jeschkies: If you use triple quotes """ there is not need to escape the quites.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -62,6 +80,116 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+    "list all versions of root group" in {
+      val clock = new SettableClock()
+      val groupManager = mock[GroupManager]
+      val groupVersions = Seq(clock.now())
+      groupManager.versions(PathId.empty) returns Source(groupVersions)
+      groupManager.group(PathId.empty) returns Some(createGroup(PathId.empty))
+      val f = new Fixture(groupManager = groupManager)
+
+      Get(Uri./.withPath(Path("/versions"))) ~> f.groupsController.route ~> check {
+        responseAs[String] should be ("[ \"2015-04-09T12:30:00.000Z\" ]")
```
jeschkies: See above.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -62,6 +80,116 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+
+  "Version detail" should {
+
+    "show app in a given version" in {
+      val infoService = mock[GroupInfoService]
+      infoService.selectGroupVersion(any, any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId("/test/group")), None, None, None)))
+      val f = new Fixture(infoService = infoService)
+
+      Get(Uri./.withPath(Path("/test/group/versions/2017-10-30T16:08:53.852Z"))) ~> f.groupsController.route ~> check {
+        responseAs[String] should include (""""id" : "/test/group"""")
```
jeschkies: I'd say we parse the json and check as in the other controller tests.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -62,6 +80,116 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+
+  "Create a group" should {
+    "fail with rejection for group that already exists" in {
+      val groupManager = mock[GroupManager]
+      groupManager.rootGroup() returns createRootGroup()
+      val f = new Fixture(groupManager = groupManager)
+      val entity = HttpEntity("{}").withContentType(ContentTypes.`application/json`)
+
+      Post(Uri./.withPath(Path("/")), entity) ~> f.groupsController.route ~> check {
+        rejection should be (Rejections.ConflictingChange(Message(s"Group / is already created. Use PUT to change this group.")))
```
jeschkies: Check the `LeaderControllerTest`. One can do

```scala
rejection shouldBe a[Rejections.ConflictingChange]
inside(rejection) {
  case Rejections.ConflictingChang(error) =>
    error.message should be("Group / is already created. Use PUT to change this group.")
}
```

This makes the error messages nicer.

Please not that scapegoat adds a warning for `s"..."` if there is not string interpolation.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -62,6 +80,116 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+      val groupManager = mock[GroupManager]
+      val rootGroup = createRootGroup(apps = Map(
+        "/appname".toRootPath -> AppDefinition("/appname".toRootPath, cmd = Some("cmd"), networks = Seq(ContainerNetwork("foo")))
+      ))
+      groupManager.rootGroup() returns rootGroup
+      val f = new Fixture(groupManager = groupManager)
+      val entity = HttpEntity("{}").withContentType(ContentTypes.`application/json`)
+
+      Post(Uri./.withPath(Path("/appname")), entity) ~> f.groupsController.route ~> check {
+        rejection should be (Rejections.ConflictingChange(Message(s"An app with the path /appname already exists.")))
```
jeschkies: See above.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -62,6 +80,116 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+      val groupApiService = mock[GroupApiService]
+      groupApiService.updateGroup(any, any, any, any)(any) returns Future.successful(createRootGroup())
+      val groupManager = mock[GroupManager]
+      groupManager.rootGroup() returns createRootGroup()
+      groupManager.updateRoot(eq(PathId.empty), any, any, eq(false), any) returns Future.successful(DeploymentPlan.empty.copy(id = "plan", version = Timestamp.zero))
+      val f = new Fixture(groupManager = groupManager, groupApiService = groupApiService)
+      val entity = HttpEntity("{}").withContentType(ContentTypes.`application/json`)
+
+      Post(Uri./.withPath(Path("/newgroup")), entity) ~> f.groupsController.route ~> check {
+        responseAs[String] should be ("""{
```
jeschkies: I know I've been doing it this way. However, I'm wondering if we should not all use the json parsing with some nice ScalaTest matchers. What do you think?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: but this is not path end, right... ?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
meln1k: It's a good point

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
meln1k: I'd change `updateRoot` method to accept `Future`

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
meln1k: What do you think about moving this into a separate `PathMatcher`?

alenkacz: @meln1k not sure I understand... Slash.? is already a path matcher :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
alenkacz: because the input is tuple and I want to unwrap that...

src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
alenkacz: GroupApiService

meln1k: Does it make sense to make `authorizer` a dependency of the `GroupApiService` class then?

src/test/scala/mesosphere/marathon/api/GroupApiServiceTest.scala
alenkacz: unfortunately I can't because of https://github.com/scalatest/scalatest/issues/1172

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
alenkacz: Sounds good. Have you anything in mind or are you working on those or should I try to come up with sth?

Btw. I actually find it OK to compare response in full as string *once* for every type of response...

alenkacz:
@meln1k @jeschkies ready for your re-review :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
jeschkies: I meant the last part `~ Slash.? ~ PathEnd`.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
jeschkies: Ok, it was just a little confusing.

src/test/scala/mesosphere/marathon/api/GroupApiServiceTest.scala
jeschkies: Ah, bummer.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
jeschkies: I've finally put the proposal into review https://github.com/mesosphere/marathon/pull/5741.

src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
```diff
@@ -130,14 +130,21 @@ trait GroupManager {
...
+    toKill: Map[PathId, Seq[Instance]] = Map.empty): Future[DeploymentPlan] = updateRootAsync(id, (g: RootGroup) => Future.successful(fn(g)), version, force, toKill)
+
+  final def updateRootAsync(
+    id: PathId,
+    fn: RootGroup => Future[RootGroup],
+    version: Timestamp = Group.defaultVersion,
+    force: Boolean = false,
     toKill: Map[PathId, Seq[Instance]] = Map.empty): Future[DeploymentPlan] = {
-    updateRootEither[Nothing](id, fn.andThen(Right(_)), version, force, toKill)
+    updateRootEither[Nothing](id, fn(_).map(Right(_))(ExecutionContexts.callerThread), version, force, toKill)
```
jeschkies: Do we want this? Was this a bug before?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
jeschkies: I'd do

```scala
val jsonResponse = Json.parse(responseAs[String])
jsonResponse shoould have(appId("/test/group"))
```

src/main/scala/mesosphere/marathon/core/group/GroupManager.scala
alenkacz: I needed to do that because it now operates on future ... it is not just a method composition as it was before

src/main/scala/mesosphere/marathon/core/group/GroupManagerModule.scala
alenkacz: it is dependency of GroupApiService, but in order to inject that dependency, I need it to be available in the GroupManagerModule

PR #5730, opened 2017-11-07
Update UCR limitations documentation

unterstein:
The UCR don`t have this restrictions anymore, so we should remove them to not confuse readers.


unterstein:
Please don't bother about the red highlights in the rest of the markdown..I just retested this with another markdown file and I am seeing the same issues :(

jdef:
lgtm

PR #5731, opened 2017-11-07
Create pods in new Akka http controller.

jeschkies:
Summary:
This implements the `/create` route in the new pods controller. Note that not all tests have been ported. They will be introduced in another PR.

JIRA issues: MARATHON-7873


src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -0,0 +1,115 @@
...
+      val entity = HttpEntity(podSpecJson).withContentType(ContentTypes.`application/json`)
+      val request = Post(Uri./.withQuery(Query("force" -> "false")))
+        .withEntity(entity)
+        .withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12"))))
+      request ~> controller.route ~> check {
+        response.status should be(StatusCodes.Created)
+        response.header[Headers.`Marathon-Deployment-Id`].value.value() should be(deploymentPlan.id)
+        response.header[Location].value.value() should be("/mypod")
+
+        (Json.parse(responseAs[String]) \ "networks" \ 0 \ "mode") shouldBe JsDefined(JsString(raml.NetworkMode.Host.value))
```
zen-dog: Nit: you could probably parse the response only once here.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -1,31 +1,87 @@
...
+        extractRequest { req =>
+          entity(as[raml.Pod]) { podDef =>
+            val normalizedPodDef = podNormalizer.normalized(podDef)
+            val pod = Raml.fromRaml(normalizedPodDef).copy(version = clock.now())
+            assumeValid(PodsValidation.pluginValidators(pluginManager).apply(pod)) {
+              authorized(CreateRunSpec, pod).apply {
+                val p = async {
+                  val deployment = await(podManager.create(pod, force))
+
+                  // TODO: How should we get the ip?
```
alenkacz: @jeschkies is this WIP or not? Just asking because of these TODO comments...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
jeschkies: There is a follow up which makes this much nicer.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
jeschkies: It's an open question. I don't think this is the proper way to get the IP since not every client will set the header. What do you think?

src/main/scala/mesosphere/marathon/api/akkahttp/PathMatchers.scala
```diff
@@ -45,6 +45,10 @@ object PathMatchers {
   }
 
   /**
+    * Matches and extracts the force parameter.
```
meln1k: Nice!

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -1,31 +1,88 @@
...
+                  val ip = clientIp.getAddress().get.toString
+                  eventBus.publish(PodEvent(ip, req.uri.toString(), PodEvent.Created))
```
meln1k: What do you think about `clientIp.toString`?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -1,31 +1,88 @@
...
+
+                  deployment
+                }
+                onSuccess(p) { plan =>
+                  // TODO: Set pod id uri
+                  val ramlPod = PodConversion.podRamlWriter.write(pod)
```
meln1k: No more `pod.toRaml` implicits? 😉

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
jeschkies: No, they need to go ;).

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -1,31 +1,87 @@
...
     }
 
-  def create(): Route = ???
+  @SuppressWarnings(Array("all")) // async/await
+  def create(): Route =
+    authenticated.apply { implicit identity =>
+      (extractClientIP & forceParameter) { (clientIp, force) =>
+        extractRequest { req =>
+          entity(as[raml.Pod]) { podDef =>
+            val normalizedPodDef = podNormalizer.normalized(podDef)
```
alenkacz: maybe I am missing something but how do we deal with NormalizationException here? Thanks

jeschkies: Good point. I don't think they should be handled here but outside. What kind of normalization exceptions do we have? How can this be?

meln1k: How about having some kind of normalization directive? It will normalize and reject in case of errors.

alenkacz: Yeah, I am not saying it should be handled here, I was just looking for a place where that is handled both here and in AppsController and did not find one... I am not against directive since that is how we handle validation errors too...

jeschkies: Do you know cases where the `NormalizationException` is thrown? I could not find it for pods.

jeschkies: Oh, the `PodNormalizationTest` have a case. However, it does not make much sense to me to have a normalization that can fail.

alenkacz: well I found this and it is used in PodNormalization https://github.com/mesosphere/marathon/blob/77ccca15b91d86e71736961f6ba00155eb27e45f/src/main/scala/mesosphere/marathon/api/v2/NetworkNormalization.scala#L37

jeschkies: Ok, I'll add a handler.

jeschkies: It is so strange. #5741 introduces a test which should result in this kind of error. However, it is a validation error. I'm really not sure is there should be normalization issues. In any case, that should be part of a separated diff.

jeschkies: Hm, it seems we do not handle `NormalizationException` at all.

PR #5732, opened 2017-11-08
Added taskId filtering in eventbus

ror6ax:
Summary:
Added a filter by taskId.

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-7882

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
@@ -25,21 +25,19 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends 
 
   lazy val id: String = UUID.randomUUID().toString
 
-
   private val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
   private val subscribedTaskIds = request.getParameterMap.getOrDefault("task_id", Array.empty).toSet
 
   def subscribedForEventType(eventType: String): Boolean = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
-  def subscribedForTaskId(taskId: String): Boolean       = subscribedEventTypes.isEmpty || subscribedTaskIds.contains(taskId)
+  def subscribedForTaskId(taskId: String): Boolean = subscribedEventTypes.isEmpty || subscribedTaskIds.contains(taskId)
 
   override def remoteAddress: String = request.getRemoteAddr
 
   override def close(): Unit = emitter.close()
 
   override def sendEvent(event: MarathonEvent): Unit = {
-  	if subscribedForEventType(event.eventType)) && subscribedForTaskId(event.taskId)) blocking(emitter.event(event.eventType, event.jsonString))
-  	if subscribedForEventType(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
-	if subscribedForTaskId(event.taskId)) blocking(emitter.event(event.taskId, event.jsonString))
+    if (subscribedForEventType(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+    /*if (subscribedForTaskId(event.taskId)) blocking(emitter.event(event.taskId, event.jsonString))*/
   }
 
   override def toString: String = s"HttpEventSSEHandle($id on $remoteAddress on event types from $subscribedEventTypes)"
--- a/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
@@ -27,16 +27,17 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends 
 
   private val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
 
-  def subscribed(eventType: String): Boolean = {
-    subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
-  }
+  def subscribedForEventType(eventType: String): Boolean = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
+  def subscribedForTaskId(taskId: String): Boolean       = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(task_id)
 
   override def remoteAddress: String = request.getRemoteAddr
 
   override def close(): Unit = emitter.close()
 
   override def sendEvent(event: MarathonEvent): Unit = {
-    if (subscribed(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+  	if subscribedForEventType(event.eventType)) && subscribedForTaskId(event.taskId)) blocking(emitter.event(event.eventType, event.jsonString))
+  	if subscribedForEventType(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+	if subscribedForTaskId(event.taskId)) blocking(emitter.event(event.taskId, event.jsonString))
   }
 
   override def toString: String = s"HttpEventSSEHandle($id on $remoteAddress on event types from $subscribedEventTypes)"
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -63,7 +63,6 @@ class AppsController(
   import Directives._
 
   private implicit lazy val validateApp = AppDefinition.validAppDefinition(config.availableFeatures)(pluginManager)
-  //  private implicit lazy val updateValidator = AppValidation.validateCanonicalAppUpdateAPI(config.availableFeatures, () => normalizationConfig.defaultNetworkName)
 
   import AppHelpers._
   import EntityMarshallers._


alenkacz:
@ror6ax your code does not compile. Can you adjust it so that it passes our CI, please?

ror6ax:
@alenkacz - need your help here. I've branched off latest master. sbt clean test never worked for me with "error: eof expected but ';' found" error. I fixed that by adding newlines in between addSbtPlugin. Not sure how that even works in CI without them.  Still, error in Jenkins is not about me code, as far as I understand, it seems like a job misconfiguration...

Locally, my build also fails, but with different error "sbt.InvalidComponent: Could not find required component 'compiler-interface-src".
Please advise.

ichernetsky:
@ror6ax could you please rebase your branch on top of `master` branch, and update this PR by doing force push?

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -27,16 +27,17 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
 
   private val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
```
ichernetsky: Shouldn't `subscribedTaskIds` be introduced similarly to `subscribedEventTypes`?

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -27,16 +27,17 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
...
 
-  def subscribed(eventType: String): Boolean = {
-    subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
-  }
+  def subscribedForEventType(eventType: String): Boolean = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
+  def subscribedForTaskId(taskId: String): Boolean       = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(task_id)
```
ichernetsky: I believe a newly introduced `subscribedForTaskIds` variable should be used here instead. Also, this code fails to compile, because there is no `task_id` variable.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -27,16 +27,17 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
...
 
   override def remoteAddress: String = request.getRemoteAddr
 
   override def close(): Unit = emitter.close()
 
   override def sendEvent(event: MarathonEvent): Unit = {
-    if (subscribed(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+  	if subscribedForEventType(event.eventType)) && subscribedForTaskId(event.taskId)) blocking(emitter.event(event.eventType, event.jsonString))
```
ichernetsky: This line is redundant. Actually, as is, this function will be duplicating events.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -27,16 +27,17 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
...
+  	if subscribedForEventType(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+	if subscribedForTaskId(event.taskId)) blocking(emitter.event(event.taskId, event.jsonString))
```
ichernetsky: It seems that `event_type` and `task_id` are `OR`ed, not `AND`ed. Is it as you want it to be? Could you please explain why? As far as I understand, the purpose of filtering is to reduce the number of events received, therefore `AND` logic would be more appropriate here.

ror6ax:
@ichernetsky , thanks, I'll address those.

ichernetsky:
Thanks, @ror6ax 
I think the error doesn't originate from where you have pointed to. Instead, if you address some of the things I referred to earlier, it will compile just fine, I believe.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
ror6ax: In fact, the problem is, there's no mention of Task in any way in events except for UnhealthyInstanceKillEvent. At this stage, we know event type and it's timestamp. Is there a way to look up TaskId based on those inputs?

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
ror6ax: The idea was to make a case for filtering by taskId, filtering by eventType and case of filtering by both. I can't figure out how to return `blocking...` for both.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
ror6ax: Done.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -26,17 +26,18 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
   lazy val id: String = UUID.randomUUID().toString
 
   private val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
+  private val subscribedTaskIds = request.getParameterMap.getOrDefault("task_id", Array.empty).toSet
 
-  def subscribed(eventType: String): Boolean = {
-    subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
-  }
+  def subscribedForEventType(eventType: String): Boolean = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
+  def subscribedForTaskId(taskId: String): Boolean = subscribedEventTypes.isEmpty || subscribedTaskIds.contains(taskId)
```
ichernetsky: `subscribedEventTypes` -> `subscribedTaskIds`, please.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -26,17 +26,18 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
...
 
   override def remoteAddress: String = request.getRemoteAddr
 
   override def close(): Unit = emitter.close()
 
   override def sendEvent(event: MarathonEvent): Unit = {
-    if (subscribed(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+    if (subscribedForEventType(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+    /*if (subscribedForTaskId(event.taskId)) blocking(emitter.event(event.taskId, event.jsonString))*/
```
ichernetsky: Please delete this line.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -26,17 +26,18 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends
...
+  def subscribedForEventType(eventType: String): Boolean = subscribedEventTypes.isEmpty || subscribedEventTypes.contains(eventType)
+  def subscribedForTaskId(taskId: String): Boolean = subscribedEventTypes.isEmpty || subscribedTaskIds.contains(taskId)
 
   override def remoteAddress: String = request.getRemoteAddr
 
   override def close(): Unit = emitter.close()
 
   override def sendEvent(event: MarathonEvent): Unit = {
-    if (subscribed(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
+    if (subscribedForEventType(event.eventType)) blocking(emitter.event(event.eventType, event.jsonString))
```
ichernetsky: I believe that

```
if (subscribedForEventType(event.eventType) && subscribedForTaskId(event.taskId)) blocking(emitter.event(event.eventType, event.jsonString))
```

is what you want, but since `event.taskId` is not available, you need to adjust `MarathonEvent` type and propagate task IDs throughout the entire pipeline.

ror6ax: I browsed the code and tried a couple of things, but I lack the detailed knowledge of this codebase to figure out where to inject task into event flow. Any suggestions? 

Also, some of the methods that get messed up by adding taskId to MarathonEvent look like they are not designed to handle tasks, they do stuff before tasks are even created e.g ApiPostEvent. How do I deal with these? I'd suggest to use some special case of marathon event in sendEvent method. This way there's less changes to be made.

ror6ax: Hi @ichernetsky , can you suggest something about the stuff I described?

jeschkies: Hi @ror6ax,

I've worked on the code recently. You could do a pattern match for certain events. Checkout [Events.scala](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/event/Events.scala#L31) for different types.

jeschkies: Btw, do you really mean task id and not instance id?

timcharper:
Closing; idle :)

PR #5734, opened 2017-11-08
Upload native packages

timcharper:
Also:

- assert valid tag
- assert prior release is direct ancestor
- assert target tag has new commits since prior release

JIRA Issues: MARATHON-7829

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -21,17 +21,26 @@ To execute a particular function just invoke the script with function appended a
 The `ci/pipeline` script defines two primary targets
 
 1. `jenkins`
-2. `phabricator`
+2. `pr`
 
-The `jenkins` target is excuted on every branch build by jenkins. It runs
+The `jenkins` target is executed on every release branch, on master, and on each Github pull request. The `jenkins`
+target will check if a build is for a pull-request, and if so, run forward to the `pr` target.
 
-  * `provision.killStaleTestProcesses()`
-  * `provision.installMesos()`
-  * `compileAndTest()`
-  * `build()`
+The `pr` target runs the test pipeline followed by Github PR review reporting. It is triggered with each diff
+update.
 
-The `phabricator` target runs the `jenkins` pipeline followed by Phabricator
-review reporting. It is triggered with each diff update.
+The test pipeline involves the following steps:
+
+* Compile
+* Run unit tests
+* Run integration tests
+* Run scapegoat
+* Build and upload snapshot artifacts
+
+Additionally, for master builds, we:
+
+* Update the DCOS packages for Marathon by pushing an update to the snapshot Marathon DCOS branch (for OSS and
+  Enterprise).
 
 ## Sub Targets
 
--- a/ci/README.md
+++ b/ci/README.md
@@ -28,7 +28,7 @@ The `jenkins` target is excuted on every branch build by jenkins. It runs
   * `provision.killStaleTestProcesses()`
   * `provision.installMesos()`
   * `compileAndTest()`
-  * `createPackages()`
+  * `build()`
 
 The `phabricator` target runs the `jenkins` pipeline followed by Phabricator
 review reporting. It is triggered with each diff update.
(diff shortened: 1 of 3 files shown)


ci/README.md
```diff
@@ -28,7 +28,7 @@ The `jenkins` target is excuted on every branch build by jenkins. It runs
   * `provision.killStaleTestProcesses()`
   * `provision.installMesos()`
   * `compileAndTest()`
-  * `createPackages()`
+  * `build()`
 
 The `phabricator` target runs the `jenkins` pipeline followed by Phabricator
```
alenkacz: I know it is not connected to your changes but this part of readme is probably outdated because it mentions phabricator... ?

timcharper:
Oh, yes :)

ci/pipeline
```diff
@@ -273,9 +285,79 @@ def release(requestVersion: String, gitSha: String, latest: Boolean = false): Un
...
+/**
+  * Given tagName v1.5.1, and rev DEADBEEF, we would assert that v1.5.0 is a direct ancestor of rev DEADBEEF
+  *
+  * Also, we assert that rev contains new commits since the last release. If the proposal is to tag v1.5.1 with same
+  * revision as v1.5.0, then we fail, saying that there v1.5.0 points to the commit porposed for v1.5.1
+  *
+  * @param tagName The tag name / version we are going to tag. eg v1.5.1
+  * @param gitSha The commit sha / rev that will be tagged.
+  */
+def assertPriorVersionIsAncestor(tagName: String, gitSha: String): Unit = {
```
jeschkies: Why is this required?

ci/pipeline
```diff
@@ -273,9 +285,79 @@ def release(requestVersion: String, gitSha: String, latest: Boolean = false): Un
...
+  })
+  val pkgserverHost = sys.env.getOrElse("PKG_SSH_HOST", {
+    throw new IllegalStateException("PKG_SSH_HOST environment variable must be set")
+  })
+
+  // Note - the key is expected to be provided via an SSH agent
+  utils.printStageTitle(s"Uploading native packages")
+  %("rsync", "-avz",
+    (pwd / 'target / 'packages) + "/",
+    s"${pkgserverUser}@${pkgserverHost}:repo/incoming/marathon-${tagName}/")
```
jeschkies: So we first rsync the packages and then move them on the server? Where is the unstable channel defined?

ci/pipeline
```diff
@@ -273,9 +285,79 @@ def release(requestVersion: String, gitSha: String, latest: Boolean = false): Un
...
+}
+
 def withoutVersionPrefix(releaseVersion: String): String = {
     if(releaseVersion.toLowerCase.startsWith("v"))
       releaseVersion.substring(1)
     else
       releaseVersion
 }
+
+def uploadLinuxPackagesToRepos(tagName: String): Unit = {
```
jeschkies: This does not seem to be called anywhere. Should we upload each master build? As far as I understood we agreed to have the master builds on the unstable release channel.

ci/pipeline
kensipe: this ensures that the sha that is being released is on the correct release branch.   It is designed to prevent human error... because I understand that humans like errors so much they create them or something

ci/pipeline
kensipe: that may be true... I understand that if we do release from master that should be on the unstable release channel.  I was sure or clear if we agreed we should do that.   At this point I would really like to publish our releases.

ci/pipeline
kensipe: nothing yet created for the unstable channel.     Perhaps we will have agreement out of our release meeting.

ci/pipeline
jeschkies: Ah, so I cannot release a 1.6 commit as 1.5? Could we express this in the comments?

ci/pipeline
jeschkies: Ok, can we follow up then? Also, shouldn't this method be called in the release target?

ci/pipeline
kensipe: it seems like it is there but I made a clarification statement.

ci/pipeline
kensipe: it is called in the release method... line 269

PR #5739, opened 2017-11-09
Use the right value for zone field

timcharper:
JIRA Issues: MARATHON_EE-1665


kensipe:
ditto on zen-dog comment... shouldn't there be tests that would have capture this?

meln1k:
@kensipe @zen-dog tests used the value from the `Constraints` object, that explains why they didn't fail.

timcharper:
😂 

Yeah, at some point we need to write a test to make sure the test is testing the right thing.

PR #5742, opened 2017-11-10
First batch of changes related to persistent volumes for pods

ichernetsky:
## Things that are done so far

- Volume and Residency RAML types
- Conversion to and from RAML, JSON, and Protocol Buffers
- App and pod volume refactoring
- Some changes to task and task group builders
- Reserving and creating volumes
- Launching pods on reservation
- Supporting pod restarts
- Persistent volume profiles
- System integration testing

## Things to do yet

- Move reservation info from the task level to the instance level
- Once reservation info is moved, probably there won't be any need in task polymorphism (Reserved, Ephemeral etc)
- Documentation
- Unit- and integration testing
- More system integration testing
- Instance update care only about last task status update, which doesn't really work well with pods with persistent volumes, because if one of pod tasks exits normally, this is considered worthy of checking and scaling the pod
- Migration


src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -80,11 +80,15 @@ object ContainerSerializer {
...
-      case dv: DockerVolume => builder.addVolumes(VolumeSerializer.toMesos(dv))
-      case _: SecretVolume => // SecretVolumes are handled differently
+    container.volumes.foreach { volumeWithMount =>
+      val volume = volumeWithMount.volume
+      val mount = volumeWithMount.mount
+      volume match {
+        case _: PersistentVolume => // PersistentVolumes are handled differently
+        case ev: ExternalVolume => ExternalVolumes.build(builder, ev, mount) // this also adds the volume
+        case dv: HostVolume => builder.addVolumes(VolumeSerializer.toMesos(dv, mount))
+        case _: SecretVolume => // SecretVolumes are handled differently
```
meln1k: Do you think it's a good idea to return something different than Unit?

ichernetsky: Could you please tell me what is better to return here and why?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -369,24 +417,47 @@ object ResourceMatcher extends StrictLogging {
           Right(DiskResourceMatch(DiskType.Mount, resourcesConsumed, scope))
         case nextAllocation :: restAllocations =>
           resources.find { resource =>
+            val matchesProfileName = nextAllocation.persistent.profileName match {
```
meln1k: Could we somehow refactor those conditional statements?

ichernetsky:
Rebased.

build.sbt
```diff
@@ -17,6 +17,9 @@ def formattingTestArg(target: File) = Tests.Argument("-u", target.getAbsolutePat
 credentials ++= loadM2Credentials(streams.value.log)
 resolvers ++= loadM2Resolvers(sLog.value)
 
+lazy val myRepo = "My Repo" at "https://s3.us-east-2.amazonaws.com/ichernetsky/maven/"
```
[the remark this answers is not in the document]
ichernetsky: It is a temporary until the Mesos team publishes an RC build.

ichernetsky:
Moved to #5813

PR #5743, opened 2017-11-10
Apps persistent volume profiles

ichernetsky:


src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -213,20 +215,25 @@ object PersistentVolumeInfo {
...
+      profileName is notEmpty
+    }
+
     validator[PersistentVolumeInfo] { info =>
       info.size should be > 0L
       info.constraints.each must complyWithVolumeConstraintRules
       info should meetMaxSizeConstraint
       info should notHaveConstraintsOnRoot
       info should haveProperlyOrderedMaxSize
+      info.profileName is optional(haveProperProfileName)
```
jeschkies: Nit: Is there a `should optionally(...)`?

Also, does this check make sense? `haveProperProfileName` validates that `profileName` is set but then the rule is optional. So it passes every time, doesn't it?

docs/docs/rest-api/public/api/v2/types/volumes.raml
```diff
@@ -58,6 +58,11 @@ types:
         minimum: 0
         description: |
           For `mount` mesos disk resources, the optional maximum size of an exclusive mount volume to be considered.
+      profileName?:
+        type: string
```
jeschkies: Should this be an enum?

ichernetsky: Why? I do not get what you mean here. A profile name is an optional arbitrary non-empty string.

src/main/scala/mesosphere/marathon/state/Volume.scala
ichernetsky: AFAIU, `should` is a synonym of `is`. I am going to change it to `should` for consistency.

`optional` here means that if `profileName` is `Some(...)` then that validator gets applied to `...`, otherwise it results in `Success`. Does it clear it out?

ichernetsky:
I am closing this PR, because I am about to manually merge this functionality into #5742

PR #5744, opened 2017-11-12
Resolve deadlock scenario in test code

timcharper:
StubLifeCycledCloseable called hooks inside of a synchronized block,
and this could deadlock.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
+++ b/src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
@@ -50,7 +50,7 @@ object CuratorElectionStream extends StrictLogging {
       .map { e =>
         e match {
           case LeadershipState.ElectedAsLeader =>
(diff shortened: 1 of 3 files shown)


timcharper:
Deadlock scenario only affects test code (stubbed life cycle closable). The actual implementation of `LifeCycledClosable` uses atomics instead of locks.

Deadlock scenario happened as follows:

* CuratorElectionStream stream crashes
* CuratorElectionStream cancel method is called, CuratorElectionStream lock acquired
* Test suite starts to tear down
* We start calling close hooks. Lock acquired in StubLifeCycledCloseable
* StubLifeCycledCloseable calls hook which is registered to call CuratorElectionStream cancel method, but is blocked, because other thread is still in CuratorElectionStream cancel method
* CuratorElectionStream cancel method tries to clear the hook (on line 187), but can't, because StubLifeCycledCloseable lock is acquired.
* Deadlock. The two wait forever.

PR #5745, opened 2017-11-13
Fixed broken integration with scheduler plugin

tsydd:
Integration with SchedulerPlugin is broken. I've made `shedulerPlugins` parameter obligatory and fixed tests

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -70,7 +70,7 @@ class InstanceOpFactoryImpl(
 
     val matchedOffer =
       RunSpecOfferMatcher.matchOffer(pod, request.offer, request.instances,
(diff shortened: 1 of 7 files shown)


jeschkies:
@tsydd, thanks for you efforts.

> Integration with SchedulerPlugin is broken. I've made shedulerPlugins parameter obligatory and fixed tests

Which integration tests do you mean? If we don't run them in our pipeline we have to change that.

alenkacz:
@jeschkies I think he meant tests in general, not integration tests... and those are fixed as part of this PR. There is probably no integration test at all ... ? (or is there? @kensipe )

zen-dog:
> Which integration tests do you mean? 
Some tests has now one parameter more (e.g. `schedulerPlugins: Set[SchedulerPlugin] = Seq.empty`). However we don't currently test plugin integration at all in marathon ITs.

jeschkies:
@alenkacz our CI should catch these issues. So we don't run everything?

jeschkies:
Ok, I've misread the comment. Apparently we do not have tests. How could we have prevented this from breaking in the first place?

PR #5747, opened 2017-11-14
Migration from joda-time to java.time

meln1k:
Migration from joda-time to java.time

Summary: joda-time removed from the project and underlying implementation of Timestamp is changed from joda.DateTime to java.time.Instant

JIRA issues: MARATHON-7914

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/Timestamp.scala
+++ b/src/main/scala/mesosphere/marathon/state/Timestamp.scala
@@ -2,7 +2,7 @@ package mesosphere.marathon
 package state
 
 import java.time.format.DateTimeFormatter
-import java.time._
+import java.time.{ OffsetDateTime, Instant, ZoneOffset, Duration }
 import java.util.concurrent.TimeUnit
 
 import org.apache.mesos
--- a/src/main/scala/mesosphere/marathon/state/Timestamp.scala
+++ b/src/main/scala/mesosphere/marathon/state/Timestamp.scala
@@ -1,11 +1,11 @@
 package mesosphere.marathon
 package state
 
-import java.time.{ Instant, OffsetDateTime }
+import java.time.format.DateTimeFormatter
+import java.time._
 import java.util.concurrent.TimeUnit
 
 import org.apache.mesos
-import org.joda.time.{ DateTime, DateTimeZone }
 
 import scala.concurrent.duration.FiniteDuration
 import scala.language.implicitConversions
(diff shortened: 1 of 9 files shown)


src/main/scala/mesosphere/marathon/state/Timestamp.scala
```diff
@@ -14,72 +14,71 @@ import scala.math.Ordered
...
   }
 
   /**
     * @return true if this timestamp is more than "by" duration older than other timestamp.
     */
   def expired(other: Timestamp, by: FiniteDuration): Boolean = this.until(other) > by
 
-  def +(duration: FiniteDuration): Timestamp = Timestamp(utcDateTime.getMillis + duration.toMillis)
-  def -(duration: FiniteDuration): Timestamp = Timestamp(utcDateTime.getMillis - duration.toMillis)
+  def +(duration: FiniteDuration): Timestamp = Timestamp(instant.plusMillis(duration.toMillis))
```
alenkacz: you can do just instant.plus(duration) right?

src/main/scala/mesosphere/marathon/state/Timestamp.scala
```diff
@@ -14,72 +14,71 @@ import scala.math.Ordered
...
 
   /**
     * Returns a new Timestamp representing the current instant.
     */
-  def now(): Timestamp = Timestamp(System.currentTimeMillis)
+  def now(): Timestamp = Timestamp(Instant.now())
 
   def now(clock: java.time.Clock): Timestamp =
-    Timestamp(clock.millis())
+    Timestamp(Instant.now(clock))
```
alenkacz: clock.instant() ?

src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
```diff
@@ -619,7 +619,7 @@ class PodsResourceTest extends AkkaUnitTest with Mockito {
           implicit val podManager = PodManagerImpl(groupManager)
           val f = Fixture()
 
-          val response = f.podsResource.version("/id", "2008", f.auth.request)
+          val response = f.podsResource.version("/id", "2008-01-01T12:00:00Z", f.auth.request)
```
alenkacz: is it OK that we must have changed this value in a test? Wasn't this a behavior that we were relying on also somewhere else?

src/test/scala/mesosphere/marathon/core/instance/InstanceFormatTest.scala
```diff
@@ -15,9 +15,9 @@ class InstanceFormatTest extends UnitTest {
       |{
       |  "instanceId": { "idString": "app.instance-1337" },
       |  "tasksMap": {},
-      |  "runSpecVersion": "2015-01-01",
+      |  "runSpecVersion": "2015-01-01T12:00:00Z",
```
alenkacz: I am a little bit woried that if we had to change so manye values in tests we will break the behavior for our clients... or what am I missing?

src/main/scala/mesosphere/marathon/state/Timestamp.scala
```diff
@@ -1,11 +1,11 @@
 package mesosphere.marathon
 package state
 
-import java.time.{ Instant, OffsetDateTime }
+import java.time.format.DateTimeFormatter
+import java.time._
```
jeschkies: Why underscore import? Do we have so many Java time classes below?

src/main/scala/mesosphere/marathon/state/Timestamp.scala
```diff
@@ -14,72 +14,71 @@ import scala.math.Ordered
...
   * An ordered wrapper for UTC timestamps.
   */
-abstract case class Timestamp private (private val utcDateTime: DateTime) extends Ordered[Timestamp] {
+abstract case class Timestamp private (private val instant: Instant) extends Ordered[Timestamp] {
   def toOffsetDateTime: OffsetDateTime =
     OffsetDateTime.ofInstant(
-      Instant.ofEpochMilli(utcDateTime.toInstant.getMillis),
-      utcDateTime.getZone.toTimeZone.toZoneId)
+      instant,
+      ZoneOffset.UTC)
```
jeschkies: Hm, did we always assume UTC before?

src/main/scala/mesosphere/marathon/state/Timestamp.scala
jeschkies: +1

src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
jeschkies: I don't think we can change the test here.

src/test/scala/mesosphere/marathon/core/instance/InstanceFormatTest.scala
jeschkies: I agree. This should not change.

meln1k: In our api we have datetime type for this field, so it's impossible to have such data passed our validation layer.

src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
meln1k: see the comment below.

src/main/scala/mesosphere/marathon/state/Timestamp.scala
meln1k: what's the difference? :)

src/main/scala/mesosphere/marathon/state/Timestamp.scala
meln1k: Unfortunately there is no such method.

src/main/scala/mesosphere/marathon/state/Timestamp.scala
meln1k: it seems that we stored `utcDateTime` before, so I think so

meln1k: you can check the deleted line 60:
```scala
def apply(dateTime: DateTime): Timestamp = new Timestamp(dateTime.toDateTime(DateTimeZone.UTC))
```

src/main/scala/mesosphere/marathon/state/Timestamp.scala
alenkacz: ah, we are not using java.time.Duration... ok then...

src/main/scala/mesosphere/marathon/state/Timestamp.scala
alenkacz: I just like it more :)

src/test/scala/mesosphere/marathon/core/instance/InstanceFormatTest.scala
alenkacz: valid, point, thanks 👍

PR #5749, opened 2017-11-16
Use task killing mode in TasksController

alenkacz:
Summary:
At some point, TaskKillingMode directive was introduced. Now it is used even in TasksController.

JIRA issues:

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
@@ -92,32 +92,29 @@ class TasksController(
   }
 
   private def deleteTasks()(implicit identity: Identity): Route = {
+    import AppsDirectives._
+
     (entity(as[TasksToDelete])
       & parameter('force.as[Boolean].?(false))
-      & parameter('scale.as[Boolean].?(false))
-      & parameter('wipe.as[Boolean].?(false))) { (taskIds, force, scale, wipe) =>
-        if (scale && wipe) {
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -92,32 +92,29 @@ class TasksController(
...
-
-                if (scale) {
-                  onSuccess(killAndScale(tasksByAppId, force)) { deploymentResult =>
-                    complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(deploymentResult.deploymentId)), deploymentResult))
-                  }
-                } else {
-                  onSuccess(kill(tasksByAppId, instanceIdsToAppId, wipe)) { tasks =>
-                    complete(TasksList(tasks).toRaml)
-                  }
+                onSuccess(kill(tasksByAppId, instanceIdsToAppId, taskKillingMode == TaskKillingMode.Wipe)) { tasks =>
```
meln1k: What if we want to just kill the task?

meln1k: And it seems we don't have a test for this case (kill without wipe)

alenkacz: well that is not a problem, right? TaskKillingMode will not be Wipe, false will be passed to the kill method and the task will be killed without wipe...

meln1k: well, yep, but I'd write it a little bit more explicit :)

PR #5751, opened 2017-11-16
Transfer all bytes with Guava instead of parts with IOUtils.

jeschkies:
Summary:
A user report that the event stream would stutter. He found that the
change to Apache IOUtils is not reading the complete input stream and
also buffers the input before it's copied. We switch to Guava
ByteStreams.copy which copies all bytes.

JIRA issues: MARATHON-7919

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
+++ b/src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
@@ -10,12 +10,12 @@ import javax.servlet.http.{ HttpServletRequest, HttpServletResponse }
 
 import akka.Done
 import akka.http.scaladsl.model.StatusCodes._
+import com.google.common.io.{ ByteStreams, Closeables }
 import com.google.inject.Inject
 import mesosphere.chaos.http.HttpConf
 import mesosphere.marathon.core.election.ElectionService
 import mesosphere.marathon.io.IO
 import mesosphere.marathon.stream.Implicits._
-import org.apache.commons.io.IOUtils
 import org.slf4j.LoggerFactory
 
 import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -294,13 +294,13 @@ class JavaUrlConnectionRequestForwarder @Inject() (
...
-          Try(leaderConnection.getErrorStream.close())
+          Closeables.closeQuietly(leaderConnection.getInputStream())
+          Closeables.closeQuietly(leaderConnection.getErrorStream())
         }
       }
     } finally {
-      Try(request.getInputStream.close())
-      Try(response.getOutputStream.close())
+      Closeables.closeQuietly(request.getInputStream())
+      Closeables.close(response.getOutputStream(), true)
```
zen-dog: Why is the output stream not closed quietly?

jeschkies: It is. The parameter is called `swallowIOException`. There is not `closeQuietly` for output streams.

alenkacz:
I reviewed but since there is still ongoing discussion about whether it helps or not I will wait with my 👍 :) https://jira.mesosphere.com/browse/MARATHON-7919

jeschkies:
> I reviewed but since there is still ongoing discussion about whether it helps or not I will wait with my 👍 

Yep, could you ask for changes though? I'm kind of surprised that this patch would not work.

jeschkies:
> Based on https://jira.mesosphere.com/browse/MARATHON-7919 I am not sure if we should merge this. Maybe we need to test this or verify in a way that it helped... ?

I agree. Let's drop it for now.

PR #5741, opened 2017-11-10
Cover all pod creation tests.

jeschkies:
Summary:
Follow up to #5731. Cover all other cases from `PodsResourceTest` and introduce have property matcher for JSON responses.

JIRA issues: MARATHON-7873

PR #5752, opened 2017-11-16
Pods contoller update endpoint implementation

meln1k:
Summary: 
implementation and tests for the update pods endpoint

JIRA issues: MARATHON-7874


src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -85,7 +85,30 @@ class PodsController(
       }
     }
 
-  def update(podId: PathId): Route = ???
+  @SuppressWarnings(Array("all")) // async/await
+  def update(podId: PathId): Route = {
+    authenticated.apply { implicit identity =>
+      (entity(as[raml.Pod]) & forceParameter & extractClientIP & extractUri) {
+        case (ramlPod, force, host, uri) =>
+          normalized(ramlPod, podNormalizer) { normalizedPodDef =>
```
jeschkies: I think you miss the normal validation before. Or is this done in `entity[as[raml.Pod]]`?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -290,6 +290,65 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
...
+      val entity = HttpEntity(postJson).withContentType(ContentTypes.`application/json`)
+      val request = Put("/mypod")
+        .withEntity(entity)
+        .withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12"))))
+
+      request ~> controller.route ~> check {
+        response.status should be(StatusCodes.OK)
+        response.header[Headers.`Marathon-Deployment-Id`].value.value() should be(deploymentPlan.id)
+
+        val pod = Json.parse(responseAs[String]).as[raml.Pod]
```
jeschkies: Check out the tests in https://github.com/mesosphere/marathon/pull/5741. I've added have matchers for json responses that make testing a little cleaner.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -290,6 +290,65 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
     //        response.getMetadata.containsKey(RestResource.DeploymentHeader) should be(true)
     //      }
     //    }
+
+    "update a simple single-container pod from docker image w/ shell command" in {
```
jeschkies: What about failing cases?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
jeschkies: Ok, apparently we used the implicit validation in `create` already. So this should be fine.

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
```diff
@@ -19,6 +19,9 @@ object Rejections {
     def noApp(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"App '$id' does not exist" + version.fold("")(v => s" in version $v")))
     }
+    def noPod(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
+      EntityNotFound(Message(s"Pod '$id' does not exist" + version.fold("")(v => s" in version $v")))
```
jeschkies: Checkout https://github.com/mesosphere/marathon/pull/5754/files#diff-ac498c575af88454a7f4eabdb3498b23. We probably should agree on one pattern. Which do you like more?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -180,4 +209,31 @@ class PodsController(
     }
   // format: ON
 
+  private def onSuccessLegacy[T](maybePodId: Option[PathId])(f: => Future[T])(implicit identity: Identity): Directive1[T] = onComplete({
```
jeschkies: Don't we have this for apps already?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -388,6 +391,92 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
...
+                       |     "image": { "kind": "DOCKER", "id": "busybox" },
+                       |     "exec": { "command": { "shell": "sleep 1" } } } ] }
+                     """.stripMargin
+      val entity = HttpEntity(postJson).withContentType(ContentTypes.`application/json`)
+      val request = Put(podId.toString)
+        .withEntity(entity)
+        .withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12"))))
+
+      request ~> controller.route ~> check {
+        rejection shouldBe a[EntityNotFound]
```
jeschkies: Please verify the message.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -388,6 +391,92 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
...
+      }
+    }
+
+    "save pod with more than one instance" in {
+      implicit val podSystem = mock[PodManager]
```
jeschkies: No need. The fixture already comes with `podManager`.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/ResponseMatchers.scala
```diff
@@ -86,4 +86,17 @@ trait ResponseMatchers { this: UnitTest =>
       HavePropertyMatchResult(matches, "podContainerSecret", Some(secret), maybeSecret)
     }
   }
+
+  /**
+    * Have property matcher for scaling policy instances.
+    * @param instances expected number of instances
+    * @return Match result
+    */
+  def scalingPolicyInstances(instances: Int) = new HavePropertyMatcher[JsValue, Option[String]] {
```
jeschkies: Cool 🙂! Why is it `Option[String]` instead of `Option[Int]`, though?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -388,6 +391,92 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
...
+                       |     "image": { "kind": "DOCKER", "id": "busybox" },
+                       |     "exec": { "command": { "shell": "sleep 1" } } } ] }
+                     """.stripMargin
+      val entity = HttpEntity(postJson).withContentType(ContentTypes.`application/json`)
+      val request = Put("/mypod")
+        .withEntity(entity)
+        .withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12"))))
+
+      request ~> controller.route ~> check {
+        response.status should be(StatusCodes.OK)
```
jeschkies: Is there a response body to test?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
alenkacz: +1 - if it is the same (and I guess it is) I wouldn't copy-paste it...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
alenkacz: yeah - maybe it can use Karsten's Response matchers

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -388,6 +391,92 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
...
+      implicit val podSystem = mock[PodManager]
+      val f = Fixture()
+      val controller = f.controller()
+
+      val podId = PathId("/unknownpod")
+
+      val deploymentPlan = DeploymentPlan.empty
+      f.podManager.update(any, eq(false)).returns(Future.failed(new PodNotFoundException(podId)))
+
+      val postJson = """
```
alenkacz: btw. if we evere remove a field or sth. from the input json, it would be super-hard to track down all these string things and alter those, because it won't fail in compile time... just saying... 

Maybe we can at least share those through some method since some of those only differ in pod name ... ?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/ResponseMatchers.scala
alenkacz: +1

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
meln1k: I'll use yours :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
meln1k: Yep, but I'd prefer not to share the code here.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
meln1k: Yes, but it's the same as what we send there. Do you think it's worth testing?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
meln1k: @jeschkies what do you think?

jeschkies: Hm, if I understood you correctly the issue is when we drop a field from the API these JSON strings would still include it. I think that is fine for now. This is the v2 API and it really should not change that much anymore.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -92,7 +99,32 @@ class PodsController(
       }
     }
 
-  def update(podId: PathId): Route = ???
+  @SuppressWarnings(Array("all")) // async/await
+  def update(podId: PathId): Route = {
+    authenticated.apply { implicit identity =>
+      (entity(as[raml.Pod]) & forceParameter & extractClientIP & extractUri) {
+        case (ramlPod, force, host, uri) =>
```
jeschkies: Why `case`? It should work without. See line 69.

src/main/scala/mesosphere/marathon/api/akkahttp/Directives.scala
```diff
@@ -121,4 +127,46 @@ object Directives extends AuthDirectives with LeaderDirectives with AkkaDirectiv
     }
   }
 
+  /**
+    * Directive for handling "legacy" method calls (ones which don't return an exception)
```
jeschkies: Which *do* or do *not* throw exceptions?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -92,7 +99,32 @@ class PodsController(
...
+            normalized(ramlPod, podNormalizer) { normalizedPodDef =>
+              val pod = Raml.fromRaml(normalizedPodDef).copy(version = clock.now())
+              assumeValid(PodsValidation.pluginValidators(pluginManager).apply(pod)) {
+                authorized(UpdateRunSpec, pod).apply {
+                  val deploymentPlan = async {
+                    val plan = await(podManager.update(pod, force))
+                    eventBus.publish(PodEvent(host.toString(), uri.toString(), PodEvent.Updated))
+                    plan
+                  }
+                  onSuccessLegacy(deploymentPlan).apply { plan =>
```
jeschkies: I must be missing something. Which exception is thrown by `update` that has to be handled? `onSuccessLegacy` lists `ValidationFailedException`, `AccessDeniedException`, `AuthDirectives.NotAuthorized` which are already handled. This leaves `ConflictingChangeException`. This could be handled in the async block, no? So we would not requires the legacy implementation.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -466,6 +470,101 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
...
+                     """.stripMargin
+      val entity = HttpEntity(postJson).withContentType(ContentTypes.`application/json`)
+      val request = Put(podId.toString)
+        .withEntity(entity)
+        .withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12"))))
+
+      request ~> controller.route ~> check {
+        rejection shouldBe a[EntityNotFound]
+        inside(rejection) {
+          case r: EntityNotFound => r.message.message shouldEqual s"Pod '/unknownpod' does not exist"
```
jeschkies: Checkout the changes in https://github.com/mesosphere/marathon/pull/5767. One of us will have to adopt this 🙂.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/ResponseMatchers.scala
```diff
@@ -107,4 +107,17 @@ trait ResponseMatchers { this: UnitTest =>
       HavePropertyMatchResult(matches, "podId", Some(id), maybeId)
     }
   }
+
+  /**
+    * Have property matcher for scaling policy instances.
+    * @param instances expected number of instances
+    * @return Match result
+    */
+  def scalingPolicyInstances(instances: Int) = new HavePropertyMatcher[JsValue, Option[Int]] {
```
jeschkies: Nice. Thanks.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -97,7 +104,34 @@ class PodsController(
...
+                val deploymentPlan = async {
+                  val plan = await(podManager.update(pod, force))
+                  eventBus.publish(PodEvent(host.toString(), uri.toString(), PodEvent.Updated))
+                  plan
+                }
+                onComplete(deploymentPlan) {
+                  case Success(plan) =>
+                    val ramlPod = PodConversion.podRamlWriter.write(pod)
+                    complete((StatusCodes.OK, Seq(Headers.`Marathon-Deployment-Id`(plan.id)), ramlPod))
+                  case Failure(e: ConflictingChangeException) =>
```
jeschkies: Nice!

PR #5753, opened 2017-11-20
Delete group in GroupsController MARATHON-7301

alenkacz:
Summary:
Re-implement delete group in akka.http GroupsController.

JIRA issues: MARATHON-7301

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
@@ -4,9 +4,10 @@ package v2
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.StatusCodes
-import akka.http.scaladsl.server.{ Directive1, Route }
+import akka.http.scaladsl.server.{ Directive1, Rejection, Route }
 import akka.stream.Materializer
 import mesosphere.marathon.api.GroupApiService
+import mesosphere.marathon.api.akkahttp.AuthDirectives.NotAuthorized
 import mesosphere.marathon.api.akkahttp.PathMatchers.GroupPathIdLike
 import mesosphere.marathon.api.akkahttp.Rejections.{ EntityNotFound, Message }
 import mesosphere.marathon.api.v2.GroupsResource.normalizeApps
@@ -14,9 +15,9 @@ import mesosphere.marathon.api.v2.{ AppHelpers, AppNormalization, PodsResource }
 import mesosphere.marathon.core.appinfo.{ AppInfo, GroupInfo, GroupInfoService, Selector }
 import mesosphere.marathon.core.election.ElectionService
 import mesosphere.marathon.core.group.GroupManager
-import mesosphere.marathon.plugin.auth.{ Authorizer, Identity, ViewGroup, Authenticator => MarathonAuthenticator }
+import mesosphere.marathon.plugin.auth.{ Authorizer, DeleteGroup, DeleteRunSpec, Identity, ViewGroup, Authenticator => MarathonAuthenticator }
 import mesosphere.marathon.raml.DeploymentResult
-import mesosphere.marathon.state.{ Group, PathId, Timestamp }
+import mesosphere.marathon.state.{ Group, PathId, RootGroup, Timestamp }
 import mesosphere.marathon.stream.Sink
 import play.api.libs.json.Json
 
@@ -102,7 +103,26 @@ class GroupsController(
 
   def updateGroup(groupId: PathId): Route = ???
 
-  def deleteGroup(groupId: PathId): Route = ???
+  def deleteGroup(groupId: PathId)(implicit identity: Identity): Route = {
+    forceParameter { force =>
+      val version = Timestamp.now()
+
+      def deleteGroupEither(rootGroup: RootGroup) = {
+        rootGroup.group(groupId) match {
+          case Some(group) if !authorizer.isAuthorized(identity, DeleteGroup, group) =>
+            Future.successful(Left(NotAuthorized(HttpPluginFacade.response(authorizer.handleNotAuthorized(identity, _)))))
+          case Some(_) => Future.successful(Right(rootGroup.removeGroup(groupId, version)))
+          case None => Future.successful(Left(Rejections.EntityNotFound.noGroup(groupId)))
+        }
+      }
+
+      onSuccess(groupManager.updateRootEither(groupId.parent, deleteGroupEither, version, force)) {
+        case Left(rejection) => reject(rejection)
+        case Right(plan) =>
+          complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
+      }
+    }
+  }
 
   def listVersions(groupId: PathId)(implicit identity: Identity): Route = {
     groupManager.group(groupId) match {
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/GroupsController.scala
```diff
@@ -102,7 +103,26 @@ class GroupsController(
 
   def updateGroup(groupId: PathId): Route = ???
 
-  def deleteGroup(groupId: PathId): Route = ???
+  def deleteGroup(groupId: PathId)(implicit identity: Identity): Route = {
+    forceParameter { force =>
+      val version = Timestamp.now()
+
+      def deleteGroupEither(rootGroup: RootGroup) = {
```
jeschkies: Hm, I know we had this pattern before but this seems odd in comparison to

```scala
def deleteGroup(rootGroup: RootGroup): Either[Rejection, DeploymentPlan]
```

why limit the compiler here?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -125,35 +116,30 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
       }
     }
 
-    "show app in a given version for root group" in {
-      val infoService = mock[GroupInfoService]
-      infoService.selectGroupVersion(eq(PathId.empty), any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId.empty), None, None, None)))
+    "show app in a given version for root group" in new Fixture {
```
jeschkies: Hm, I don't think `new Fixture` is required here since a fixture is created below.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -125,35 +116,30 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+      infoService.selectGroupVersion(org.mockito.Matchers.eq(PathId.empty), any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId.empty), None, None, None)))
       val f = new Fixture(infoService = infoService)
 
-      Get(Uri./.withPath(Path("/versions/2017-10-30T16:08:53.852Z"))) ~> f.groupsController.route ~> check {
+      Get(Uri./.withPath(Path("/versions/2017-10-30T16:08:53.852Z"))) ~> groupsController.route ~> check {
         (Json.parse(responseAs[String]) \ "id").get shouldEqual JsString("/")
```
jeschkies: How about the `should have` matcher here?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -125,35 +116,30 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
     }
 
-    "show app in a given version for root group" in {
-      val infoService = mock[GroupInfoService]
-      infoService.selectGroupVersion(eq(PathId.empty), any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId.empty), None, None, None)))
+    "show app in a given version for root group" in new Fixture {
+      infoService.selectGroupVersion(org.mockito.Matchers.eq(PathId.empty), any, any, any) returns Future.successful(Some(GroupInfo(createGroup(PathId.empty), None, None, None)))
       val f = new Fixture(infoService = infoService)
 
-      Get(Uri./.withPath(Path("/versions/2017-10-30T16:08:53.852Z"))) ~> f.groupsController.route ~> check {
```
jeschkies: Hm, if `f` is removed `val f = new Fixture(infoService = infoService)` should be removed as well. However, I fail to see the benefit as it violates the code legibility a tiny bit.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -239,4 +245,26 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
 
     val groupsController: GroupsController = new GroupsController(electionService, infoService, groupManager, groupApiService, config)
   }
+
+  case class FixtureWithRealGroupManager(
```
jeschkies: Why is the mock not working for us? Should we just go with the actual group manager in get rid off the mocked one?

alenkacz: hmm maybe... ? groupsresource actually had also two kinds of tests - with real and mocked one. 

The reason why I need a real one here is that part of the behavior is injected to the groupManager via function as parameter and I need to test that behavior so I need to actually execute the method...

alenkacz:
@jeschkies all your comments should be addressed now

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
alenkacz: btw. I would probably stick with this two fixture setup - the problem is, that some methods  calling group repository on the background, and those are just more easily mocked than done on real group manager... But this concrete thing (with the injected function) I cannot easily test on mock. It is a mess (kind of)... :(

jeschkies: I'd opt for not mocking at all then 🙂.

jeschkies: A [Hangout](https://www.youtube.com/watch?v=z9quxZsLcfo&list=PLJb2p0qX8R_qSRhs14CiwKuDuzERXSU8m) with DHH, Kent Beck and Martin Fowler convinced me that we should not mock too much. Why not call the real thing?

alenkacz: (disclaimer - I did not watch the video yet, will do that later today)

jeschkies: Hm, the name is confusing then. What is bad about mocking the repository instead of the manager?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/ResponseMatchers.scala
```diff
@@ -86,4 +87,11 @@ trait ResponseMatchers { this: UnitTest =>
       HavePropertyMatchResult(matches, "podContainerSecret", Some(secret), maybeSecret)
     }
   }
+
+  def id(pathId: PathId) = new HavePropertyMatcher[JsValue, Option[PathId]] {
```
jeschkies: Could you add a quick doc?

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
alenkacz: let's continue this discussion on slack - I am still interested in your input even though I am merging this...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/ResponseMatchers.scala
```diff
@@ -87,6 +88,18 @@ trait ResponseMatchers { this: UnitTest =>
     }
   }
 
+  /**
+    * Have property matcher for id (e.g. of an app)
+    * @param pathId id of an app
+    * @return Match result
+    */
+  def id(pathId: PathId) = new HavePropertyMatcher[JsValue, Option[PathId]] {
```
jeschkies: I've just noticed that this is better than `podId`. Would you mind replacing `have(podId)` with `have(id)`?

PR #5756, opened 2017-11-21
Remove link to deprecated artifact store

meichstedt:
The artifact store was deprecated with Marathon 1.4 and has been removed in version 1.5.0.

Summary:
The artifact store was deprecated with Marathon 1.4 and has been removed in 1.5.0. This commit removes a link that survived somehow.


meichstedt:
Note that the gh pages need to be updated for this to take effect.

PR #5754, opened 2017-11-20
 Define and test pod find endpoint.

jeschkies:
Summary:
Implement the pod controllers find ability.

JIRA issues: MARATHON-7876

PR #5758, opened 2017-11-21
Delete pod from pod controllers route.

jeschkies:
Summary:
This implements the removal of pod through the pods controller.

JIRA issues: MARATHON-7877

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
@@ -7,11 +7,11 @@ import akka.event.EventStream
 import akka.http.scaladsl.model.{ StatusCodes, Uri }
 import akka.http.scaladsl.model.headers.Location
 import akka.http.scaladsl.server.Route
-import mesosphere.marathon.api.akkahttp.{ Controller, Headers }
+import mesosphere.marathon.api.akkahttp.{ Controller, Headers, Rejections }
 import mesosphere.marathon.api.akkahttp.PathMatchers.{ PodsPathIdLike, forceParameter }
 import mesosphere.marathon.core.group.GroupManager
 import mesosphere.marathon.core.instance.Instance
-import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer, CreateRunSpec }
+import mesosphere.marathon.plugin.auth.{ Authenticator, Authorizer, CreateRunSpec, DeleteRunSpec }
 import mesosphere.marathon.state.PathId
 import akka.http.scaladsl.server.PathMatchers
 import com.wix.accord.Validator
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -100,7 +100,36 @@ class PodsController(
 
   def find(podId: PathId): Route = ???
 
-  def remove(podId: PathId): Route = ???
+  def remove(podId: PathId): Route =
+    authenticated.apply { implicit identity =>
+      (extractClientIP & forceParameter) { (clientIp, force) =>
+        extractRequest { req =>
```
meln1k: Can we extract URI instead of request here?

jeschkies: Sure. Much better. Thanks.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -100,7 +100,37 @@ class PodsController(
 
   def find(podId: PathId): Route = ???
 
-  def remove(podId: PathId): Route = ???
+  @SuppressWarnings(Array("all")) // async/await
+  def remove(podId: PathId): Route =
+    authenticated.apply { implicit identity =>
+      (extractClientIP & forceParameter) { (clientIp, force) =>
```
meln1k: What do you think about this?
```scala
(extractClientIP & forceParameter & extractUri) { (clientIp, force, uri) =>
```

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
```diff
@@ -55,7 +59,12 @@ class PodsControllerTest extends UnitTest with ScalatestRouteTest with RouteBeha
       val request = Post(Uri./.withQuery(Query("force" -> "false")))
         .withEntity(entity)
         .withHeaders(`Remote-Address`(RemoteAddress(InetAddress.getByName("192.168.3.12"))))
+
+      val podDefinition = PodDefinition(id = PathId("mypod"))
+      f.podManager.find(any).returns(Some(podDefinition))
```
meln1k: How about this? `f.podManager.find(any)` -> `f.podManager.find(eq(PathId("mypod")))`

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
alenkacz: +1

jeschkies: ✅

src/test/scala/mesosphere/marathon/api/akkahttp/v2/PodsControllerTest.scala
jeschkies: Same as in #5754. We do not want to test whether `find` was invoked but if the request is rejection due to unauthorized requests.

PR #5759, opened 2017-11-21
Introduce dcos-type-generator 

wavesoft:
This PR brings in the `dcos-type-generator` project that can read RAML files and generate code based on [StringTemplate4](https://github.com/antlr/stringtemplate4/blob/master/doc/cheatsheet.md) templates.

You can consider this "fluffy" I guess 👼  , as long as you have something like this in mind:

Fluffy

Or you could [browse this on it's own repository](https://github.com/mesosphere/dcos-type-generator)


jeschkies:
Hey, you did not break the record of 10 thousand lines of change 😉.

type-generator/.gitignore
```diff
@@ -0,0 +1,107 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
```
jeschkies: Why does this include Python related ignores?

type-generator/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+            <artifactId>snakeyaml</artifactId>
+            <version>1.19</version>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.12</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
jeschkies: This dependency does not seem to be used.

type-generator/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+            <artifactId>compiler</artifactId>
+            <version>0.9.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>ST4</artifactId>
+            <version>4.0.8</version>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
```
jeschkies: Where is this used?

type-generator/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>1.8</maven.compiler.target>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.raml</groupId>
```
jeschkies: Where is this used?

type-generator/.gitignore
wavesoft: You are actually right. I started this as a python project. I am going to change this 👍

type-generator/pom.xml
wavesoft: Good point. Forgotten. 👍

type-generator/pom.xml
wavesoft: Haha, yeah, this was from the first implementation that I was using moustache

type-generator/pom.xml
wavesoft: Not sure, this dependency was added from IntelliJ when I started the project, but it might have done some optimistic import

wavesoft: This is for parsing `ini` files. It is possible to define some configuration parameters for the compiler in an ini file.

Check https://github.com/mesosphere/marathon/pull/5759/files#diff-3300810e4d40efdaf2a598765dddea02

jdef:
Some questions:
(a) why added as a sub-tree? this is inconsistent w/ everything else in marathon
(b) what does the "fluffy" label mean? this reads like an inside joke
(c) is there a related JIRA that provides context for the motivation of this work? if so, can you link to it in the description?

wavesoft:
Hey @jdef 
(a) This project is part of a bigger idea that might or might not evolve into something outside the scope of marathon. However spinning this off (and signing it off) as a separate project requires more effort (ex. release plan & integration as an external dependency). Keeping it in the project reduces the bootstrap effort and importing it as a sub-tree allows us to be able to push/pull from/to the upstream.

(b) Yes, that's an in-joke. "Small and cute" is self-explanatory : A PR with just a few lines, "Fluffy" means slighter bigger in size, yet clean and organised. This PR doesn't really fall into this ("small") category, thus the uber-fluffy bunny above (because I think it's kind of organised 😄 )

(c) You can find the discussion in `MARATHON_EE-1707` that evolved a bit after the whole team was together the past two weeks, and it also merged with an idea that I started as a spare-time project : to provide some foundation for normalising the API across the different DC/OS projects.

jdef:
Thanks for the context. Relevant information from the MARATHON_EE ticket
should be surfaced into a public JIRA and linked to this PR so that the
public can follow along.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+        Option o_help = Option.builder("h")
+                .longOpt("help")
+                .desc("Print this help message")
+                .argName("help")
+                .build();
+
+        Option o_project = Option.builder("p")
+                .longOpt("project")
+                .required()
+                .desc("The project name to target")
```
jeschkies: What does this mean? Is this a target folder?

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+                .hasArg()
+                .argName("project")
+                .build();
+
+        Option o_version = Option.builder("v")
+                .longOpt("version")
+                .required()
+                .desc("The project version to use")
```
jeschkies: How are the versions defined? Do we agreed on versioning? I think the new API strives for backwards compatibility. @timcharper has more insight.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+                .hasArg()
+                .argName("version")
+                .build();
+
+        Option o_repo = Option.builder("r")
+                .longOpt("repo")
+                .required()
+                .desc("The repository URI to use")
```
jeschkies: Is this just a folder? Could we postpone the support for repositories? This seems like a door that says open even if we won't support repositories for now. Let's keep things simple.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+                        ps.println("");
+
+                        ps.close();
+                        fos.close();
+                    }
+                }
+            }
+
+        } catch (UnrecognizedOptionException e) {
+            System.err.println("ERROR: Unrecognized option " + e.getOption());
```
jeschkies: Why not use a logger?

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+            );
+
+            // Load project
+            RAMLParser raml = new RAMLParser();
+            GlobalScope globalScope = new GlobalScope();
+            raml.loadProjectTypes(project, globalScope);
+
+            // Initialize template
+            RepositoryView compilerView = repo.openProjectCompiler(
+                cmd.getOptionValue("project"),
```
jeschkies: What is the difference between `openProjectTypes` and `openProjectCompiler`?

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+            CompiledFragment frag = engine.getCompiledFragment(CompiledFragmentType.FILE_CONTENTS);
+
+            // Display type registry
+            for (TypeScope scope: globalScope.childScopesOfType(FileScope.class)) {
+                for (Map.Entry<String, Type> t: scope.flatTypeMap().entrySet()) {
+                    Type type = t.getValue();
+                    StructuralType sType = type.getStructural();
+
+                    if (!(sType instanceof ScalarType) && !(sType instanceof ArrayType)) {
+                        String baseDir = "/Users/icharala/Develop/playground-scala/src/main/scala/types/";
```
jeschkies: I don't think we should hard code this.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+                        FileOutputStream fos = new FileOutputStream(baseDir + type.getId() + ".scala");
+                        PrintStream ps = new PrintStream(fos, false, "UTF-8");
+                        ps.println("package types");
+                        ps.println("");
+                        ps.println("import java.time.OffsetDateTime");
```
jeschkies: Hm, should this be done by the template?

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
```diff
@@ -0,0 +1,156 @@
...
+            // Initialize template
+            RepositoryView compilerView = repo.openProjectCompiler(
+                cmd.getOptionValue("project"),
+                cmd.getOptionValue("version")
+            );
+            CompilerEngine engine = new StringTemplateEngine(compilerView);
+            CompiledFragment frag = engine.getCompiledFragment(CompiledFragmentType.FILE_CONTENTS);
+
+            // Display type registry
+            for (TypeScope scope: globalScope.childScopesOfType(FileScope.class)) {
```
jeschkies: Why do we have to iterator over the scopes? Couldn't the compiler do so?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/TypeCompiler.java
```diff
@@ -0,0 +1,30 @@
...
+     * The repository view from which to load the compiler templates
+     */
+    private RepositoryView compilerView;
+
+    /**
+     * Initialize the type compiler view
+     * @param engine
+     * @param compilerView
+     */
+    public TypeCompiler(CompilerEngine engine, RepositoryView compilerView) {
```
jeschkies: Why is this compiler not iterating over the types?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+/**
+ * A StringTemplateEngine engine is using the StringTemplateEngine template engine to
+ * compile a user expression into a buffer.
+ */
+public class StringTemplateEngine implements CompilerEngine {
+
+    /**
+     * The zone ID to use when parsing/writing dates
+     */
+    private ZoneId dateZone;
```
jeschkies: UTC it is 🙂.

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+
+    /**
+     * Mapping of scalar types
+     */
+    private Map<String, String> scalarTypes;
```
jeschkies: What does that mean?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+        this.group = new STGroupRepository(compilerView, "compiler.stg", '<', '>');
+
+        // Register <escape()> formatter
+        this.group.registerRenderer(String.class, new FormattersRenderer());
+        this.group.defineTemplate("escape", "value", "<value; format=\"_escapeFormatter\">");
+        this.group.defineTemplate("capitalize", "value", "<value; format=\"_capitalizeFormatter\">");
+
+        // Load the language settings
+        Ini ini = new Ini();
+        ini.load(compilerView.openFile("config/language.ini"));
```
jeschkies: Is the ini specific to StringTemplate?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+        this.scalarTypes = ini.get("scalar_types");
+
+        // Locate date expression formatters
+        Map<String, String> dateSection = ini.get("date");
+        if (dateSection == null) dateSection = new HashMap<>();
+        this.dateParsers = new HashMap<>();
+        this.dateExpressions = new HashMap<>();
+        this.dateZone = ZoneId.of(dateSection.getOrDefault("timezone", "Europe/Berlin"));
+
+        // Prepare defaults
```
jeschkies: I don't quite get why w define date parsers here. When are we parsing dates? In the deserialization?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+     * @return The contents of the template
+     */
+    private String templateRender(String templateName, BlockArguments arguments) throws IOException {
+        ST tpl = group.getInstanceOf(templateName);
+        if (tpl == null) {
+            throw new IOException("Unable to locate template " + templateName);
+        }
+
+        // Import public getter methods as fields from BlockArguments
+        for (Method method : arguments.getClass().getDeclaredMethods()) {
```
jeschkies: Isn't StringTemplate doing this by default?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+            } catch (IllegalAccessException e) {
+                e.printStackTrace();
+            }
+        }
+
+        // Write template to the output stream
+        return tpl.render();
+    }
+
+//    /**
```
jeschkies: Can this go?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+    /**
+     * Render a value template
+     *
+     * @param template
+     * @param type
+     * @param kind
+     * @param value
+     * @return
+     */
+    private String renderValueTemplate(String template, Type type, TypeScope scope, String kind, Object value) {
```
jeschkies: What is a value template?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+        // Try the template specified
+        ST tpl = group.getInstanceOf(template);
+
+        // Fallback to "valueof" if the template is missing
+        if (tpl == null) {
+            System.err.println("WARNING: Could not find template `" + template + "`, falling back to `toString`");
```
jeschkies: Why not use a logging library?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+            // Render scalar template
+            return renderValueTemplate("valueof", type, scope, ((ScalarType) type).storageType.toName(), value);
+
+        }
+
+        // We should never reach this point
+        return "<unsupported>";
+    }
+
+//    /**
```
jeschkies: Can this go?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
```diff
@@ -0,0 +1,777 @@
...
+            case DEBUG_ENTRYPOINT:
+                templateName = "debug";
+                break;
+        }
+
+        // Instantiate and return new compiled fragment
+        return new StringTemplateCompiledFragment(templateName);
+    }
+
+//    /**
```
jeschkies: Can this go?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/internal/BlockArguments.java
```diff
@@ -0,0 +1,25 @@
+package io.mesosphere.types.dtr.compiler.internal;
+
+import java.util.Map;
+
+/**
+ * A set of parameters that are passed down to a code block,
+ * defining the template fields and scope.
+ */
+public interface BlockArguments {
```
jeschkies: Why is this empty?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/internal/CompiledFragment.java
```diff
@@ -0,0 +1,29 @@
+package io.mesosphere.types.dtr.compiler.internal;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+/**
+ * A compiled fragment is instantiated from the compiler engine and can be
+ * evaluated at any time.
+ */
+public interface CompiledFragment {
```
jeschkies: Could you give an example?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/internal/CompilerEngine.java
```diff
@@ -0,0 +1,46 @@
...
+     * </p>
+     * For example `StringValue` would be `String` in Java
+     * </p>
+     * Note that array types or other more complicated structures are constructed
+     * through the engine templates and we don't need to provide such functionality here.
+     *
+     * @param storageType The internal storage type for which to get a name
+     * @return Returns a string with the name of the storage type in the user-defined implementation
+     */
+    String getStorageTypeName(Value storageType);
```
jeschkies: Hm, `StorageType` is an odd name. Could this be done in StringTemplate? I mean the internal model would know which type it is so the template could just decide how to name it.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: Both of them return a view to different folders.

* `openProjectTypes` gives access to `/types` folder
* `openProjectCompiler ` gives access to the `/compiler` folder

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: Yeah, I could replace the `LocalRepository` with something much simpler for now.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: This comes from the type registry repository idea, in order to keep things versioned (this assumes that the types reside outside from marathon). The path components are:

`<repository dir>/<project>/<version>/{types|compiler|migrations}`

* `-r` defines the repository as a URI (ex. file:/path/to/repository) for LocalRepository
* `-p` defines the project (pointless in our case) --> `marathon`
* `-v` defines the version of the API to generate (ex. `1`). It's just the folder name
* `/types` : Contains the RAML types
* `/compiler` : Contains the compiler templates
* `/migrations` : (will) contain the migration rules between API versions

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: Check my comment above

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: True, this needs a bit re-structuring. 👍  What happens there is the following:

* We create a `GlobalScope`
* For every RAML file we have we create a `FileScope`
* Technically, calling `GlobalScope.flatTypeMap` will suffice.

In a previous implementation the naming of the types was wrongly prefixed with the filename, thus I iterated over the child "FileScope"s. But I think we can remove this now.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: If you are referring to the `baseDir`, totally! (Just pushed this change)

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: Yes, moved this on template 👍

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
wavesoft: True. Left over from the debugging times... 👍

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/TypeCompiler.java
wavesoft: I wanted to add a bit more flexibility on the compiler interface, allowing the user to perform fine-grained, per-item compilation.

But I would guess that adding the support to additionally compiling the entire scope would make sense :+1:

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Haha, true. Each scalar type (well. `Value` instances to be precise) has an internal name, that needs to be mapped to the language-specific type name. For instance:

```ini
any=Object
binary=byte[]
boolean=Boolean
datetime=java.time.OffsetDateTime
date_only=java.time.OffsetDateTime
time_only=java.time.OffsetDateTime
datetime_only=java.time.OffsetDateTime
integer=Int
number=Double
string=String
void=None
```

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Yes. It provides useful arguments to the compiler. For instance:

* The mapping between the internal types and the language-specific type (check the comment above)
* Date/Time-specific configuration

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: This is used when parsing/writing down timezone-unaware dates. But you have a point, fixing this on UTC should suffice 😄

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: This defines how the date literals reach the compiler. Let me elaborate:

* Consider a type of `type: datetime`
* The engine is going to define a `defaultValue` expression, as a `ValueArguments` instance
    * If the date is an expression that can parsed by snakeyaml, this value is going to be stored internally as a date object
    * If the date is a string that cannot be parsed, the date parser is used instead. The parser depends on the type of the date, so **this** is where the date parsers are used.
* Through the `ValueArguments` interface the user has access to:
    * The raw date instance and it's properties
    * A stringified expression of the date. For **this** value you need the date formatters

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Definitely 😄

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Definitely 😄

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Definitely 😄

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Yes, but it does this on the values you define with `.add()`, but it cannot take all the properties you have in "arguments" and expose them as template arguments. 

As a matter of fact I can now see that the `arguments` argument could be simply changed into a `Map<String, BlockArgument>`. (The idea of the `BlockArgument` is to annotate whatever can be passed as an argument to the compiler)

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: Definitely. Will be fixed 👍

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/internal/BlockArguments.java
wavesoft: The idea of the `BlockArgument` is to annotate whatever can be passed as an argument to the compiler.

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/internal/CompiledFragment.java
wavesoft: Initially, I thought of creating something that can be evaluated in a getter, allowing templates to render other templates. Therefore you could pass a "CompiledFragment" to a template argument, and render it when needed.

But I then realised that something like this can be easily achieved with StringTemplate if given values with enough information (and that was the beginning of the `*Arguments` classes).

Therefore I agree that this could be replaced with something like `engine.render(template, arguments)`

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/internal/CompilerEngine.java
wavesoft: As a matter of fact yes. This could be removed. This is a remainder of the initial moustache implementation ...

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
wavesoft: It's a template that constructs the language-specific expression that carries the given value. Useful for creating `defaultValue` expressions.

Namely, we have two kinds of templates, that being `inst` or `valueof` and they both accept a very similar set of arguments.

type-generator/src/main/java/io/mesosphere/types/dtr/Main.java
jeschkies: Hm, ok, this should be documented somewhere. However, I'd drop the versioning for now. That seems quite out of scope.

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
jeschkies: I think this should be done in the string template itself. It does support maps, right?

type-generator/src/main/java/io/mesosphere/types/dtr/compiler/impl/StringTemplateEngine.java
jeschkies: > If the date is a string that cannot be parsed
If I understand the RAML spec correctly this should not happen and it should be fine to throw an error for unparsable default value.

type-generator/src/main/java/io/mesosphere/types/gen/Main.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Entry point
+     * @param args
+     */
+    public static void main(String[] args) throws ParseException {
+        Options options = new Options();
+        CommandLineParser parser = new DefaultParser();
+
+        // Prepare options
+        Option o_help = Option.builder("h")
```
zen-dog: suggestion: `JCommander` gives an arguably nicer parameter definition (with annotations) and parsing: http://jcommander.org/

jeschkies:
I was looking at the architecture a little bit. I think it would help if we would follow the naming scheme of common compiler architectures:
1. Lexing / Tokenization
2. Parsing / Syntax analysis which produces the AST
3. Semantic analysis which converts the AST to an intermediate representation.
5. Code generation.

I've omitted the optimization steps.

Lexing and parsing is done by the YAML parser. One could argue that Checking some RAML specs should be part of this.

The semantic analysis is producing the types in our case. This is the intermediate representation.

The code generation is done with StringTemplate. So it seems what is called `CompilereEngine` here is the code generator.

What do you think about following these naming schemes?

zen-dog:
Great work however 8000 slocs is a lot for the first commit. How about we strip it to "20% code that is doing 80% work"? Repository, versioning, migration and other concerns that are above of what we have now should be addressed in separate PRs.

jeschkies:
I refreshed my memory a little bit yesterday and looked into compiler architectures. We should not go that wild but it helps to agree on common terms. As it turns out I was not quite right with

> Lexing and parsing is done by the YAML parser. One could argue that Checking some RAML specs should be part of this.

The YAML parser is rather out tokenizer and we miss a proper parser. E.g. The YAML parsers allows

```yaml
properties:
    properties:
        properties:
            foo: string
```

however that is not a valid RAML expression AFIK. We could go insane and define a RAML grammar for [ANTLR](http://www.antlr.org/) or [FastParse](https://github.com/lihaoyi/fastparse). However, that is not trivial. This brings me to the question: Why are we not using the [official RAML parser](https://github.com/raml-org/raml-java-parser)? What am I missing?

The RAML parser also comes with its own internal representation. What was the motivation for our own internal types, ie intermediate representation? Do we need this for now?

type-generator/src/main/java/io/mesosphere/types/gen/Main.java
```diff
@@ -0,0 +1,154 @@
...
+
+        // Parse options
+        try {
+            CommandLine cmd = parser.parse(options, args);
+
+            // Handle and bail early on hep argument
+            if (cmd.hasOption("help")) {
+                HelpFormatter formatter = new HelpFormatter();
+                formatter.printHelp( "dtr", options );
+                return;
```
timcharper: should it sys.exit(1) ?

timcharper:
Wow, gave it a read through. On the one hand, yes, we've gone from 1,000 lines of Scala (including SBT integration) to 6,500 lines of Java, plus some odd tests. This isn't a fair comparison, directly, for multiple reasons: Java is much more verbose than Java, this work has substantially better code documentation, and the Scala implementation should've had tests to begin with. (are validations implemented, too?)

Still, that accounted for, it does feel like a significant jump in complexity.

At the end of the day, though, I don't know really how much I care, if this is going to be split out and maintained as a separate tool. Maintained by who / which team, though? 🤔 

I'll try and fiddle with it more.

timcharper:
Could you include some instruction on how to run this for the Marathon project?

type-generator/src/main/java/io/mesosphere/types/gen/Main.java
wavesoft: Good point 👍

wavesoft:
On my defence it's `3,929` lines of code plus `2,346` comments, plus `1,028` empty 😆 
But you also need to account the StringTemplate sources from #5765

I have addressed most of @jeschkies comments and I am fixing-up the README. Will update this PR in a while.

wavesoft:
Ps. I already know that this could be used by the DC/OS UI team to replace https://github.com/dcos/raml-validator-loader and possibly even generate the API interface to marathon.

wavesoft:
@jeschkies I decided not to use the official RAML because of my hard experience the last time I did this with [the raml-validation-loader](https://github.com/dcos/raml-validator-loader) . In principle, traversing the type tree through the RAML API or walking the YAML definition is roughly the same thing. (Meaning that even if I use the official RAML parser I still need to to the same tricks you see in `createType`), however:

* The RAML API tries to be smart and high-level, since the intended usage is to validate a given input. Accessing the internal type representation is trickier (and in some cases with the Typescript version of RAML library -- even impossible, leading into [ugly hacks](https://github.com/dcos/raml-validator-loader/blob/master/src/generators/FacetValidators.js#L254)). Since the Typescript version is a carbon-copy of the official Java version, I thought that the Java API interface would be similar.

* My RAML parser implementation is using asynchronous type definitions, allowing types to be referred to, without being yet defined. This allows my parser to walk through the directory tree only once. On the other hand, the official RAML parser needs to resolve all the includes and validate the tree for every file it's asked to parse, synchronously. This leads into multiple parsing of the same files again and again, slowing down the parsing time. As a comparison:
    * Parsing AND Generating the sources with `dcos-type-generator` takes about 3 seconds.
    * Just loading `rest-api/public/api/api.raml` with the official RAML parser takes about 8 seconds.
 
* (Minor) The way the RAML builder [loads the the raml files](https://jar-download.com/javaDoc/org.raml/raml-parser-2/1.0.6/org/raml/v2/api/RamlModelBuilder.html#buildApi-java.io.File-) is not compatible with the streaming approach that I had in mind. This means that I cannot use the `RepositoryView` abstraction. 

On the other hand, I agree that if we use the official RAML parser we would get syntax validation of the RAML file for free. But again, you also mentioned, we could just include Yagi in the project...

jeschkies:
>  In principle, traversing the type tree through the RAML API or walking the YAML definition is roughly the same thing. (Meaning that even if I use the official RAML parser I still need to to the same tricks you see in createType).

I don't think so. Traversing through the parsed YAML is akin to iterating over the output of a lexer, ie the tokens (maybe the syntactic analysis), traversing through the parsed RAML is akin to iterating over the output of the semantic analysis, ie the intermediate representation, as the RAML parser defines a grammar.

> The RAML API tries to be smart and high-level, since the intended usage is to validate a given input. Accessing the internal type representation is trickier (and in some cases with the Typescript version of RAML library -- even impossible, leading into ugly hacks). Since the Typescript version is a carbon-copy of the official Java version, I thought that the Java API interface would be similar.

I think Typescript is out of scope for now. Didn't we use the RAML parser for the current generator? What issues did we encounter there?

As @timcharper mentions someone will have to maintain this. So far I don't see a reason why we would not use the RAML parser. You've raised some points but they do not seem to exceed the added value of an open source project such as the official parser which has [at least twenty contributors](https://github.com/raml-org/raml-java-parser/graphs/contributors).

If we want to think long term going with your version and have a short term results, I hardly see any other way than:
1. Follow a common compiler pipeline pattern. Note, we could come up with something better but the likelihood of outsmarting sixty plus years of best practices is very low ;).
2. Rely on existing projects and adapt them if needed. I had to do the same for the Swagger generator which was in much worse shape than the RAML parser but still did not block the project from going forward and going towards the long term goal.

Overall I do like your vision. Unfortunately, it is not a high priority yet. The current generator has bugs and is not tested. It is hard to make changes. If we can tackle these issues *and* take a step closer to your version, I'm quite happy.

wavesoft:
> I don't think so. Traversing through the parsed YAML is akin to iterating over the output of a lexer, ie the tokens (maybe the syntactic analysis), traversing through the parsed RAML is akin to iterating over the output of the semantic analysis, ie the intermediate representation, as the RAML parser defines a grammar.

In an ideal world yes. Sadly RAML does not really represent types in a compiler-friendly format. So don't expect to get something like the "Internal Models" from the RAML library for free.

As a simple example, "enums" are annotations on top of other types. This is quite incompatible with the PoV of the "Compiler AST" that we discussed. _(And that's why I introduced an `EnumType` in the suggested "internal types" in the first place)_.

 [We already do](https://github.com/mesosphere/marathon/blob/master/project/RamlTypeGenerator.scala#L145) "heuristics" and/or apply constraints in order reduce the RAML description into something usable. For example:
1. [We deny declaration of arrays as `Type[]`s](https://github.com/mesosphere/marathon/blob/master/project/RamlTypeGenerator.scala#L145)
3. [We accept only string enums](https://github.com/mesosphere/marathon/blob/master/project/RamlTypeGenerator.scala#L157) even though the RAML spect does not explicitly describes so.
2. [We check if there are regex fields and we generate maps instead of objects](https://github.com/mesosphere/marathon/blob/master/project/RamlTypeGenerator.scala#L725)

.. and we more or less do exactly what I do in the RAML parser, but through the additional layer of the RAML library. So the only benefit that I see from using the official RAML library is just the syntactical validation. 

> As @timcharper mentions someone will have to maintain this. So far I don't see a reason why we would not use the RAML parser. You've raised some points but they do not seem to exceed the added value of an open source project such as the official parser which has at least twenty contributors.

We already maintain our `RamlTypeGenerator`, right? As I mentioned above, at least one extra team (the UI) can benefit from this tool. Keep in mind that I already maintain the https://github.com/dcos/raml-validator-loader , so if the UI team switches to the `dcos-type-generator` and I take care of maintaining this project, my load would remain the same 😄 

> 1. Follow a common compiler pipeline pattern. Note, we could come up with something better but the likelihood of outsmarting sixty plus years of best practices is very low ;).

I completely agree. I am going to clean-up the logic and separate it into 3 distinct PRs, trying to follow some compiler principles (even though I can't see them matching 100% this case).

> 2. Rely on existing projects and adapt them if needed. I had to do the same for the Swagger generator which was in much worse shape than the RAML parser but still did not block the project from going forward and going towards the long term goal.

Regarding the RAML parser, I wish I could use it. Honestly. But the more I play with it, the more restrictive it feels... 

However I did like the `StringTemplate4` approach and went with it 😄

type-generator/src/main/java/io/mesosphere/types/gen/compiler/internal/arguments/README.md
```diff
@@ -0,0 +1,8 @@
+# Argument Classes
```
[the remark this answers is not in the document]
wavesoft: Sorry, I didn't notice this comment before 😄 They are arguments to the template. They encapsulate the "type" fields, but they also provide access to some additional meta-information. (For example for the literal "foo" the template might need access to the underlaying type).

But I kind of agree that they are also "models". I just don't want to mix them with the internal types.

zen-dog:
This giant PR has been open for a while now - what do we do with it? I'll accept it *only* if Karsten is 100% happy with it.

jeschkies:
Hodor!

wavesoft:
I added the `wip` label because there is an ongoing work according to the discussion.

PR #5761, opened 2017-11-22
 Retrieve status for pods in pods controller.

jeschkies:
Summary:
Retrieve the status for one pod. It seems that the current implementation as resource
is wrong.

JIRA issues: MARATHON-7878

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
@@ -4,14 +4,14 @@ package api.akkahttp.v2
 import java.time.Clock
 
 import akka.event.EventStream
-import akka.http.scaladsl.model.{StatusCodes, Uri}
+import akka.http.scaladsl.model.{ StatusCodes, Uri }
 import akka.http.scaladsl.model.headers.Location
 import akka.http.scaladsl.server.Route
-import mesosphere.marathon.api.akkahttp.{Controller, Headers, Rejections}
-import mesosphere.marathon.api.akkahttp.PathMatchers.{PodsPathIdLike, forceParameter}
+import mesosphere.marathon.api.akkahttp.{ Controller, Headers, Rejections }
+import mesosphere.marathon.api.akkahttp.PathMatchers.{ PodsPathIdLike, forceParameter }
 import mesosphere.marathon.core.group.GroupManager
 import mesosphere.marathon.core.instance.Instance
-import mesosphere.marathon.plugin.auth.{Authenticator, Authorizer, CreateRunSpec, ViewRunSpec}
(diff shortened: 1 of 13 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -147,7 +149,23 @@ class PodsController(
...
+  def status(podId: PathId): Route =
+    authenticated.apply { implicit identity =>
+      podManager.find(podId) match {
+        case None =>
+          reject(Rejections.EntityNotFound.noPod(podId))
+        case Some(pod) =>
+          authorized(ViewRunSpec, pod).apply {
+            onSuccess(podStatusService.selectPodStatus(podId)) { maybeStatus =>
+              // If selectPodStatus returns None this is a bug since find(podId) already verifies that the pod exists.
+              // We don't filter the pods with an authentication since we check for authorization before.
```
alenkacz: authentication or authorization? I think you wanted to use word authorization in both cases in this sentence, right?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -147,7 +149,23 @@ class PodsController(
...
+  @SuppressWarnings(Array("all")) // async/await
+  def status(podId: PathId): Route =
+    authenticated.apply { implicit identity =>
+      podManager.find(podId) match {
+        case None =>
+          reject(Rejections.EntityNotFound.noPod(podId))
+        case Some(pod) =>
+          authorized(ViewRunSpec, pod).apply {
+            onSuccess(podStatusService.selectPodStatus(podId)) { maybeStatus =>
+              // If selectPodStatus returns None this is a bug since find(podId) already verifies that the pod exists.
```
alenkacz: is it possible that the pod might be deleted by another request in between find and select calls?

jeschkies: I think so. However, IMHO that is still a bug in our setup. It's actually an issue with `selectPodStatus` which does not distinguish between pod not found or access rejected.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
jeschkies: It should be "authorization". Thanks.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -147,7 +149,23 @@ class PodsController(
...
+              // We don't filter the pods with an authentication since we check for authorization before.
+              val status: raml.PodStatus = maybeStatus.get
```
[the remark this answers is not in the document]

PR #5762, opened 2017-11-22
 Retrieve all status in pods controller. 

jeschkies:
Summary:
Fetch and filter statuses for all pods. Depends on #5761.

JIRA issues: MARATHON-7880

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
@@ -152,6 +152,7 @@ class PodsController(
       }
     }
 
+  @SuppressWarnings(Array("OptionGet"))
   def status(podId: PathId): Route =
     authenticated.apply { implicit identity =>
       podManager.find(podId) match {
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -171,7 +174,19 @@ class PodsController(
 
   def version(podId: PathId, v: String): Route = ???
 
-  def allStatus(): Route = ???
+  def allStatus(): Route =
+    authenticated.apply { implicit identity =>
+      def isAuthorized(pod: PodDefinition): Boolean = authorizer.isAuthorized(identity, ViewRunSpec, pod)
+
+      val filteredPods = Source(podManager.ids())
+        .mapAsync(Int.MaxValue) { id =>
```
meln1k: Do you think it's worth limiting amount of concurrent calls here?

jeschkies: I'd say so but we don't have a common way to configure this yet. So I'd rather follow up.

meln1k: Well, let's just hardcode the limit for now :)

PR #5766, opened 2017-11-22
Use correct v1.5.1 URL in index page.

tgockel:
Summary:
The "Install Marathon" index page advised users to curl an invalid URL. Change
the URL to point to the current (1.5.1) tarball.

JIRA issues:


kensipe:
Thanks!!!

PR #5767, opened 2017-11-23
 Define version and versions route in pods controller. 

jeschkies:
Summary:
This implements and tests the version endpoint for pods. I also introduced a test behaviour for unknown entities which should simplify testing. It has been refactored for leader and queue controller tests but another refactoring may be required for the apps controller test.

Depends on #5762.

JIRA issues: MARATHON-7879

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
@@ -20,16 +20,16 @@ object Rejections {
     private def readableVersion(version: Option[Timestamp]) = version.fold("")(v => s" in version $v")
 
     def noApp(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
-      EntityNotFound(Message(s"App '$id' does not exist ${readableVersion(version)}"))
+      EntityNotFound(Message(s"App '$id' does not exist${readableVersion(version)}"))
     }
     def noGroup(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
-      EntityNotFound(Message(s"Group '$id' does not exist ${readableVersion(version)}"))
+      EntityNotFound(Message(s"Group '$id' does not exist${readableVersion(version)}"))
     }
     def noLeader(): EntityNotFound = {
       EntityNotFound(Message("There is no leader"))
     }
-    def noPod(id: PathId): EntityNotFound = {
-      EntityNotFound(Message(s"Pod '$id' does not exist"))
+    def noPod(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
+      EntityNotFound(Message(s"Pod '$id' does not exist${readableVersion(version)}"))
     }
 
     def queueApp(appId: PathId): EntityNotFound = {
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
@@ -31,6 +31,9 @@ object Rejections {
     def noPod(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
       EntityNotFound(Message(s"Pod '$id' does not exist${readableVersion(version)}"))
     }
+    def noPod(id: PathId, version: String): EntityNotFound = {
+      EntityNotFound(Message(s"Pod '$id' does not exist in version $version"))
+    }
 
     def queueApp(appId: PathId): EntityNotFound = {
       EntityNotFound(Message(s"Application $appId not found in tasks queue."))
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
@@ -170,7 +170,6 @@ class PodsController(
       }
     }
 
-  @SuppressWarnings(Array("all")) // async/await
   def status(podId: PathId): Route =
     authenticated.apply { implicit identity =>
       podManager.find(podId) match {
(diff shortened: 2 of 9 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
```diff
@@ -20,16 +20,16 @@ object Rejections {
     private def readableVersion(version: Option[Timestamp]) = version.fold("")(v => s" in version $v")
 
     def noApp(id: PathId, version: Option[Timestamp] = None): EntityNotFound = {
-      EntityNotFound(Message(s"App '$id' does not exist ${readableVersion(version)}"))
+      EntityNotFound(Message(s"App '$id' does not exist${readableVersion(version)}"))
```
zen-dog: why was the space removed (here and below)?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -204,9 +202,32 @@ class PodsController(
...
-  def versions(podId: PathId): Route = ???
+  def versions(podId: PathId): Route =
+    authenticated.apply { implicit identity =>
+      podManager.find(podId) match {
+        case None =>
+          reject(Rejections.EntityNotFound.noPod(podId))
+        case Some(pod) =>
+          authorized(ViewRunSpec, pod).apply {
+            val versions = podManager.versions(podId).runWith(Sink.seq)
+            complete(versions)
```
zen-dog: `pods.raml` says the return type here is `type: string[]` where `podManager.versions` gives you a `Source[Timestamp...]`. hmm...

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
```diff
@@ -33,6 +31,22 @@ class GroupsControllerTest extends UnitTest with ScalatestRouteTest with Inside
...
+      val f = new FixtureWithRealGroupManager(authorized = false)
+      behave like unknownEntity(forRoute = f.groupsController.route, withRequest = Delete("/groupname"), withMessage = "Group '/groupname' does not exist")
+
+      val infoService = mock[GroupInfoService]
+      infoService.selectGroup(any, any, any, any) returns Future.successful(None)
+      infoService.selectGroupVersion(any, any, any, any) returns Future.successful(None)
+      val g = new Fixture(infoService = infoService)
+
+      behave like unknownEntity(forRoute = g.groupsController.route, withRequest = Get("/unknown-group"), withMessage = "Group '/unknown-group' does not exist")
+      behave like unknownEntity(forRoute = g.groupsController.route, withRequest = Get("/groupname/versions/2017-10-30T16:08:53.852Z"), withMessage = "Group '/groupname' does not exist in version 2017-10-30T16:08:53.852Z")
```
zen-dog: Where does the magical timestamp of `2017-10-30T16:08:53.852Z` come from? `SettableClock` already has `2015-04-09T12:30:00` as a magical constant.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/QueueControllerTest.scala
```diff
@@ -380,14 +376,14 @@ class QueueControllerTest extends UnitTest with ScalatestRouteTest with Inside w
...
-    "access without authorization is denied if the app is in the queue" in new Fixture {
+    "access without authorization is denied if the app is in the queue" in {
       Given("an authenticated but not authorized request")
       val f = new Fixture(authenticated = true, authorized = false)
 
       And("an app in the queue")
       f.launchQueue.listAsync returns Future.successful(Seq(
         LaunchQueue.QueuedInstanceInfo(AppDefinition("app".toRootPath), inProgress = false, 0, 0,
-          backOffUntil = clock.now() + 100.seconds, startedAt = clock.now())))
+          backOffUntil = f.clock.now() + 100.seconds, startedAt = f.clock.now())))
```
zen-dog: Good catch.

src/main/scala/mesosphere/marathon/api/akkahttp/Rejections.scala
jeschkies: This is confusing, I know. So `readableVersion(version)` already has a space prepended. We would end up with two. If `readableVersion(version)` is empty we would have an extra space.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
jeschkies: If all versions are datetimes then I can adapt. I always thought that versions are strings but internally datetimes.

src/test/scala/mesosphere/marathon/api/akkahttp/v2/GroupsControllerTest.scala
jeschkies: `2017-10-30T16:08:53.852Z` is an unknown versions. That's what the test is testing 🙂.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/PodsController.scala
```diff
@@ -204,9 +202,32 @@ class PodsController(
...
+          }
+      }
+    }
 
-  def version(podId: PathId, v: String): Route = ???
+  def version(podId: PathId, v: String): Route =
+    authenticated.apply { implicit identity =>
+      val version = Timestamp(v) // Original Resource is not validating.
```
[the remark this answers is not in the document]
jeschkies: So, it could be handled like other invalid paths or just as an unknown entity. I'd opt for the later. What do you think?

PR #5769, opened 2017-11-23
Add IPv6 support for mesos health checks

alenkacz:
Summary:
Phase 1 of adding ipv6 support that is needed by our customers. Currently their infra is running on ipv4, only containers will be switched to ipv6.

This PR adds the possibility to specify that Mesos health check should be run on ipv6.

JIRA issues: MARATHON_EE-1713

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -77,7 +77,7 @@ object Dependency {
     val Alpakka  = "0.8"
     val Chaos = "0.10.0"
     val Guava = "19.0"
-    val Mesos = "1.4.0"
+    val Mesos = "1.5.0-health-check-ipv6"
     // Version of Mesos to use in Dockerfile.
     val MesosDebian = "1.4.0-2.0.1"
     val OpenJDK = "openjdk:8u121-jdk"
--- a/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
@@ -407,6 +407,7 @@ trait AppValidation {
     app.portDefinitions is optional(portDefinitionsValidator)
     app is containsCmdArgsOrContainer
     app.healthChecks is every(portIndexIsValid(portIndices(app)))
+    app.healthChecks is every(complyWithIpProtocolRules(app.container))
     app must haveAtMostOneMesosHealthCheck
     app.fetch is every(valid)
     app.secrets is { secrets: Map[String, SecretDef] =>
@@ -494,6 +495,20 @@ trait AppValidation {
     }
   }
 
+  private def complyWithIpProtocolRules(container: Option[Container]): Validator[AppHealthCheck] =
+    isTrue(AppValidationMessages.HealthCheckIpProtocolLimitation) { healthCheck =>
+      val allowedProtocols = Set(
+        AppHealthCheckProtocol.MesosHttp,
+        AppHealthCheckProtocol.MesosHttps,
+        AppHealthCheckProtocol.MesosTcp)
+
+      def isMesosHttpTcpHealthCheck: Boolean = allowedProtocols.contains(healthCheck.protocol)
+      def isDockerContainer = container.exists(c => c.`type` == EngineType.Docker)
+      val hasIpProtocol = healthCheck.ipProtocol.isDefined
+
+      !hasIpProtocol || (isMesosHttpTcpHealthCheck && isDockerContainer)
+    }
+
   private val haveAtMostOneMesosHealthCheck: Validator[App] = {
     val mesosProtocols = Set(
       AppHealthCheckProtocol.Command,
@@ -537,4 +552,7 @@ object AppValidationMessages {
 
   val DockerEngineLimitedToSingleContainerNetwork =
     "may only specify a single container network when using the Docker container engine"
+
+  val HealthCheckIpProtocolLimitation =
+    "Health check ipProtocol can be used only for docker container and mesos http/https/tcp healthcheck"
 }
--- a/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
@@ -8,6 +8,7 @@ import com.wix.accord._
 import com.wix.accord.dsl._
 import mesosphere.marathon.api.v2.Validation.{ featureEnabled, _ }
 import mesosphere.marathon.core.externalvolume.ExternalVolumes
+import mesosphere.marathon.core.health.IPv4
 import mesosphere.marathon.raml._
 import mesosphere.marathon.state.{ AppDefinition, PathId, ResourceRole }
 import mesosphere.marathon.stream.Implicits._
@@ -502,11 +503,11 @@ trait AppValidation {
         AppHealthCheckProtocol.MesosHttps,
         AppHealthCheckProtocol.MesosTcp)
 
-      def isMesosHttpTcpHealthCheck: Boolean = allowedProtocols.contains(healthCheck.protocol)
+      def isMesosHttpHealthCheck: Boolean = allowedProtocols.contains(healthCheck.protocol)
       def isDockerContainer = container.exists(c => c.`type` == EngineType.Docker)
-      val hasIpProtocol = healthCheck.ipProtocol.isDefined
+      val hasDefaultIpProtocol = healthCheck.ipProtocol == IpProtocol.Ipv4
 
-      !hasIpProtocol || (isMesosHttpTcpHealthCheck && isDockerContainer)
+      hasDefaultIpProtocol || (isMesosHttpHealthCheck && isDockerContainer)
     }
 
   private val haveAtMostOneMesosHealthCheck: Validator[App] = {
--- a/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
@@ -554,5 +554,5 @@ object AppValidationMessages {
     "may only specify a single container network when using the Docker container engine"
 
   val HealthCheckIpProtocolLimitation =
-    "Health check ipProtocol can be used only for docker container and mesos http/https/tcp healthcheck"
+    "Health check ipProtocol can only be used for docker container and mesos http/https/tcp healthcheck"
 }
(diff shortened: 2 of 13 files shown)


src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
```diff
@@ -277,4 +277,65 @@ class AppValidationTest extends UnitTest with ValidationTestLike {
...
+              protocol = protocol,
+              port = Some(80),
+              ipProtocol = Some(IpProtocol.Ipv6))))
+
+          basicValidator(dockerAppWithMesosHttpHealthCheck) should be(aSuccess)
+        }
+      }
+
+      "fail for Marathon HTTP health check with ip protocol" in {
+        val dockerAppWithMesosHttpHealthCheck = dockerApp.copy(healthChecks =
```
zen-dog: you mean `val dockerAppWithMarathonHealthCheck`?

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
```diff
@@ -277,4 +277,65 @@ class AppValidationTest extends UnitTest with ValidationTestLike {
...
+      val ucrApp = App(
+        id = "/foo",
+        container = Some(Container(
+          `type` = EngineType.Mesos,
+          docker = Some(DockerContainer(
+            image = "xyz")))))
+
+      "should fail even for otherwise supported types" in {
+        allowedProtocols.foreach { protocol =>
+          val dockerAppWithMesosHttpHealthCheck = ucrApp.copy(healthChecks =
```
zen-dog: `val ucrApp...`?

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
```diff
@@ -277,4 +277,65 @@ class AppValidationTest extends UnitTest with ValidationTestLike {
...
+              protocol = protocol,
+              port = Some(80),
+              ipProtocol = Some(IpProtocol.Ipv6))))
+
+          basicValidator(dockerAppWithMesosHttpHealthCheck) should haveViolations(
+            "/healthChecks(0)" -> AppValidationMessages.HealthCheckIpProtocolLimitation)
+        }
+      }
+    }
+  }
```
zen-dog: One more test without setting the protocol?

project/Dependencies.scala
```diff
@@ -77,7 +77,7 @@ object Dependency {
     val Alpakka  = "0.8"
     val Chaos = "0.10.0"
     val Guava = "19.0"
-    val Mesos = "1.4.0"
+    val Mesos = "1.5.0-health-check-ipv6"
```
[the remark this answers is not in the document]
alenkacz: not mesos, but mesos JVM library - yes. Basically 1.4.1 does not contain the changes I need and 1.5.0 will be ready in 2 weeks

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -342,7 +377,8 @@ object MesosTcpHealthCheck {
       maxConsecutiveFailures = proto.getMaxConsecutiveFailures,
       delay = proto.getDelaySeconds.seconds,
       portIndex = PortReference.fromProto(proto),
-      port = if (proto.hasPort) Some(proto.getPort) else None
+      port = if (proto.hasPort) Some(proto.getPort) else None,
+      ipProtocol = if (proto.hasIpProtocol) Some(ipProtocolFromProto(proto.getIpProtocol)) else None
```
meichstedt: `else IPv4`?

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
```diff
@@ -277,4 +277,65 @@ class AppValidationTest extends UnitTest with ValidationTestLike {
...
+
+    "is UCR app" should {
+      val ucrApp = App(
+        id = "/foo",
+        container = Some(Container(
+          `type` = EngineType.Mesos,
+          docker = Some(DockerContainer(
+            image = "xyz")))))
+
+      "should fail even for otherwise supported types" in {
```
meichstedt: `"fail even for otherwise supported types"`

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
```diff
@@ -277,4 +277,65 @@ class AppValidationTest extends UnitTest with ValidationTestLike {
     }
 
   }
+
+  "health check validation" when {
+    val allowedProtocols = Set(
```
meichstedt: Shouldn't we make sure to test behavior for all diallowed protocols as well?
`AppHealthCheckProtocol` is a sealed trait, so this could be done with a little help from the compiler :)

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
alenkacz: you mean IP protocol? Hopefully - added...

alenkacz:
@zen-dog @meln1k ready for re-review...

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -242,13 +244,17 @@ case class MesosHttpHealthCheck(
   port: Option[Int] = HealthCheckWithPort.DefaultPort,
   path: Option[String] = MarathonHttpHealthCheck.DefaultPath,
   protocol: Protocol = MesosHttpHealthCheck.DefaultProtocol,
+  ipProtocol: Option[IpProtocol] = None,
```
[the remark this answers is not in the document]
zen-dog: I thought about that too, but since mesos protos are backwards compatible `ipProtocol` is defined as `optional`. I guess we can either omit setting it or adhere to the provided default.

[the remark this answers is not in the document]
meichstedt: The Marathon model doesn't need to reflect the proto. If `None` equals `IPv4` and doesn't lead to problems elsewhere, I would make it explicit and make it mandatory. This effectively reduces the number of combinations we need to test.

src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
```diff
@@ -494,6 +495,20 @@ trait AppValidation {
...
+      val allowedProtocols = Set(
+        AppHealthCheckProtocol.MesosHttp,
+        AppHealthCheckProtocol.MesosHttps,
+        AppHealthCheckProtocol.MesosTcp)
+
+      def isMesosHttpTcpHealthCheck: Boolean = allowedProtocols.contains(healthCheck.protocol)
+      def isDockerContainer = container.exists(c => c.`type` == EngineType.Docker)
+      val hasIpProtocol = healthCheck.ipProtocol.isDefined
+
+      !hasIpProtocol || (isMesosHttpTcpHealthCheck && isDockerContainer)
```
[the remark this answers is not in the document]
alenkacz: good point, thank you. I will adjust the logic...

timcharper:
As for automated tests, asserting that we send the right thing to Mesos does seem sufficient. However, I think we should write (and execute) a manual test plan for this.

src/main/scala/mesosphere/marathon/core/health/HealthCheck.scala
```diff
@@ -282,6 +290,19 @@ case class MesosHttpHealthCheck(
 object MesosHttpHealthCheck {
   val DefaultPath = None
   val DefaultProtocol = Protocol.MESOS_HTTP
+  val DefaultIpProtocol = IPv4
+
+  def ipProtocolConversion(ipProtocol: IpProtocol): Protos.HealthCheckDefinition.IpProtocol = ipProtocol match {
+    case IPv4 => Protos.HealthCheckDefinition.IpProtocol.IPv4
+    case IPv6 => Protos.HealthCheckDefinition.IpProtocol.IPv6
+    case _ => throw new IllegalStateException(s"Unsupported protocol in health check: $ipProtocol")
```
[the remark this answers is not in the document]
alenkacz: well the only reason I did it like this is because compiler will give us just warning (by the way I tend to think that it is useful to have match errors escalated as build errors but that is a different story...)

So if someone ignores the warning (and I think it is possible it happens) in runtime it will give much more cryptic exception "IPv6 (of class mesosphere.marathon.core.health.IPv6$)
scala.MatchError: IPv6 (of class mesosphere.marathon.core.health.IPv6$)" than the one I put there...

But I agree that with the solution as it is now, we don't even have a chance to get that warning, so I'll remove it. Thanks...

alenkacz:
@timcharper I can't test it manually until we have mesos version supporting this. Or am I missing something?

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
alenkacz: Yep, I am not sure either ... I can definitely create manual TableDriven test...

src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala
```diff
@@ -537,4 +551,7 @@ object AppValidationMessages {
 
   val DockerEngineLimitedToSingleContainerNetwork =
     "may only specify a single container network when using the Docker container engine"
+
+  val HealthCheckIpProtocolLimitation =
+    "Health check ipProtocol can only be used for docker container and mesos http/https/tcp healthcheck"
```
meichstedt: I think this should now be "IPv6 can only be used for container type DOCKER and Mesos http/https/tcp health checks"

src/test/scala/mesosphere/marathon/api/v2/validation/AppValidationTest.scala
meichstedt: Sorry – I was referring to the map but I wasn't aware it's private.
The compiler help would probably be little over engineered: a partial function that triggers the tests, and a verification based on the map to check that the PF is defined for all available values :)

alenkacz:
@timcharper nvm, I am talking to the core guys, they will have something merged in two hours, then I will create the manual test plan... Good point, thanks!

PR #5771, opened 2017-11-24
Network name with underscore is accepted by Marathon MARATHON-7851

alenkacz:
Summary:
Verify that network name with underscore is accepted by marathon - we had a breaking change in on of the previous versions of Marathon and some of our customers rely on this behavior.

JIRA issues: MARATHON-7851

### DIFF
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -1077,6 +1077,20 @@ def test_vip_docker_bridge_mode(marathon_service_name):
 
     http_output_check()
 
+def test_network_name_with_underscore_is_deployed():
+    """Tests that application containing network name with underscore is accepted
+        by Marathon service.  We do not assert that the app is deployed because that
+        would require to create that network prior to running tests."""
+
+    app_def = apps.network_name_underscore_app()
+    client = marathon.create_client()
+    # add_app call already asserts that marathon returned response with new deployment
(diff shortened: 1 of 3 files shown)


tests/system/marathon_common_tests.py
```diff
@@ -1078,6 +1078,22 @@ def http_output_check():
     http_output_check()
 
 
+@shakedown.dcos_1_10
+def test_network_name_with_underscore_is_deployed():
+    """Tests that application containing network name with underscore is accepted
+        by Marathon service.  We do not assert that the app is deployed because that
+        would require to create that network prior to running tests."""
```
zen-dog: Wouldn't that also crash the agents when you try to deploy an app with a non-existing cni network?

alenkacz: this is docker, not UCR so no CNI and thus hopefully no crash...

but thanks for reminding me, I need to push that stacktrace to core-oncall...

alenkacz:
> Ad. If we just test whether or not marathon accepts the definition (and don't start the actual task) then wouldn't this test make more sense as a simple unit test in AppResource?

yeah, I discussed this with Ken ... he was proposing SI test, I also think controller test MIGHT be enough - although you mock many things in the resource/controller test so this is more end to end approach...

Does someone else have an opinion on this? I don't want to rewrite for third time... :)

PR #5779, opened 2017-11-29
Perf optimize /v2/apps json writes. (~20% faster)

htmldoug:
Perf optimize /v2/apps json writes. (~20% faster)

## Problem
marathon-lb likes to hammer `GET /v2/apps?embed=apps.tasks`, which is driving the load average of our cluster up to 25.

This is driven by 3 factors.
1. We have 840 apps on the cluster, resulting in a 7 MB json response body.
2. We run 20 marathon-lb instances, far more than the recommended 2 (https://github.com/mesosphere/marathon-lb/issues/282).
3. play-json writes are expensive.

Our Ops team manages 1 and 2 which are beyond my control, so this change optimizes 3 (the play-json writes) a bit.

## Optimizing Json Writes
Profiling shows that json writes dominate CPU time, particularly generating scala Maps.

The worst offender is `EnvVarConversion.envVarRamlWrites`, which makes an unnecessary copy of the `"env"` map. Replacing `.map` with `.mapValues` produces a view instead of a copy and avoids rehashing the keys. This nets a cool 20% improvement over our typical use case.

hot methods

### Results
Before:
```
marathon(before)> benchmark/jmh:run -t1 -f3 -wi 10 -i 30 .*JsonFormatBenchmark
[info] Benchmark                     Mode  Cnt   Score   Error   Units
[info] JsonFormatBenchmark.appInfo  thrpt   90  32.456 ± 0.688  ops/ms
```

After:
```
marathon(perf-optimize-json)> benchmark/jmh:run -t1 -f3 -wi 10 -i 30 .*JsonFormatBenchmark
[info] Benchmark                     Mode  Cnt   Score   Error   Units
[info] JsonFormatBenchmark.appInfo  thrpt   90  41.079 ± 0.613  ops/ms
```

## jmh "asm"
This also reverts a change in https://github.com/mesosphere/marathon/commit/94efcee727612559ca227af4adce3a458281ee2c to set the jmh generator from "asm" back to the default. When set to "asm", sbt fails with https://gist.github.com/htmldoug/cbf820af93f89c4516893eb8ba4a163a and I can't get the benchmarks to run at all. Happy to set it back if someone can get it working.

## JIRA issues
None filed or searched for. I'm unfamiliar with your process. Holler if you need me to create some.

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/json/JsonFormatBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/json/JsonFormatBenchmark.scala
@@ -1,0 +1,116 @@
+package mesosphere.marathon
+package json
+
+import java.util.concurrent.TimeUnit
+
+import mesosphere.marathon.api.v2.json.AppAndGroupFormats
+import mesosphere.marathon.core.appinfo.{AppInfo, EnrichedTask}
+import mesosphere.marathon.core.condition.Condition
+import mesosphere.marathon.core.instance.Instance.AgentInfo
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.Task.{LaunchedOnReservation, Reservation, Status}
+import mesosphere.marathon.core.task.state.NetworkInfo
+import mesosphere.marathon.state.{AppDefinition, EnvVarString, PathId, Timestamp}
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+import play.api.libs.json.Json
+
+/**
+  * This benchmark focuses on the CPU time of [[AppAndGroupFormats.ExtendedAppInfoWrites]].
+  *
+  * GET /v2/apps?embed=apps.tasks is slow over a large number of apps (840 in our case).
+  *
+  * marathon-lb hammers this endpoint relentlessly. Json writing dominates CPU.
+  *
+  * Run with: {{{
+  *   benchmark/jmh:run -t1 -f1 -wi 5 -i 10 .*JsonFormatBenchmark
+  * }}}
+  *
+  * Or with Java Flight Recorder profiling: {{{
+  *   benchmark/jmh:run -prof jmh.extras.JFR -t1 -f1 -wi 5 -i 20 .*JsonFormatBenchmark
+  * }}}
+  */
+@State(Scope.Benchmark)
+object JsonFormatBenchmark extends AppAndGroupFormats {
+
+  /**
+    * Attempts to represent a real world-ish [[AppInfo]],
+    * with particular focus on accurately representing sizes of json objects,
+    * which translate to expensive scala Maps.
+    */
+  val realisticAppInfo: AppInfo = {
+
+    /**
+      * Taken from our dev cluster with: {{{
+      *   curl /v2/apps | jq '[.apps[].env | length] | add / length'
+      * }}}
+      */
+    val numEnvVars = 70
+
+    /**
+      * Taken from our dev cluster with: {{{
+      *   curl /v2/apps | jq '[.apps[].labels | length] | add / length'
+      * }}}
+      */
+    val numLabels = 11
+
+    val appId = PathId("benchmark/app/definition")
+
+    AppInfo(
+      app = AppDefinition(
+        id = appId,
+        env = 0.to(numEnvVars).map(i => ("KEY_" * 10) + i -> EnvVarString(("VALUE_" * 10) + i)).toMap,
+        labels = 0.to(numLabels).map(i => ("KEY_" * 10) + i -> (("VALUE_" * 10) + i)).toMap
+      ),
+      maybeTasks = Some(
+        Seq(
+          EnrichedTask(
+            appId,
+            task = LaunchedOnReservation(
+              taskId = Task.Id("benchmark_app_definition.2e251a11-af74-11e7-8e35-12ebe3d150b4"),
+              runSpecVersion = Timestamp.now(),
+              status = Status(
+                stagedAt = Timestamp.now(),
+                startedAt = Some(Timestamp.now()),
+                mesosStatus = None,
+                condition = Condition.Running,
+                networkInfo = NetworkInfo(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -321,6 +321,5 @@ lazy val benchmark = (project in file("benchmark"))
   .dependsOn(marathon % "compile->compile; test->test")
   .settings(
     testOptions in Test += Tests.Argument(TestFrameworks.JUnit),
-    libraryDependencies ++= Dependencies.benchmark,
-    generatorType in Jmh := "asm"
```
htmldoug: Tagging @unterstein @jeschkies of https://github.com/mesosphere/marathon/commit/94efcee727612559ca227af4adce3a458281ee2c.

When set to "asm", sbt fails with https://gist.github.com/htmldoug/cbf820af93f89c4516893eb8ba4a163a and I can't get the benchmarks to run at all. Maybe you know of a better alternative than removing this.

src/main/scala/mesosphere/marathon/raml/EnvVarConversion.scala
```diff
@@ -8,17 +8,17 @@ import scala.collection.immutable.Map
 trait EnvVarConversion {
   implicit val envVarRamlWrites: Writes[Map[String, state.EnvVarValue], Map[String, EnvVarValueOrSecret]] =
     Writes {
-      _.map {
-        case (name, state.EnvVarString(v)) => name -> EnvVarValue(v)
-        case (name, state.EnvVarSecretRef(secret: String)) => name -> EnvVarSecret(secret)
+      _.mapValues {
```
jeschkies: Nice catch! What do you think about  using the builder from [Map.newBuilder](http://www.scala-lang.org/api/2.12.3/scala/collection/immutable/Map$.html#newBuilder[A,B]:scala.collection.mutable.Builder[(A,B),CC[A,B]]) or `breakOut` instead?

```scala
implicit val envVarRamlWrites: Writes[Map[String, state.EnvVarValue], Map[String, EnvVarValueOrSecret]] = new Writes {
  override def write(in: Map[String, state.EnvVarValue]): Map[String, EnvVarValueOrSecret] = {
    val builder = Map.newBuilder[String, EnvVarValueOrSecret]
    in.foreach {
      case (name, state.EnvVarString(v))=> builder += key -> EnvVarValue(v)
      case (name, state.EnvVarSecretRef(secret: String)) => builder += name -> EnvVarSecret(secret)
    }
    builder.result()
  }
}
```

or

```scala
implicit val envVarRamlWrites: Writes[Map[String, state.EnvVarValue], Map[String, EnvVarValueOrSecret]] = new Writes {
  override def write(in: Map[String, state.EnvVarValue]): Map[String, EnvVarValueOrSecret] = {
    in.iterator.map {
      case (name, state.EnvVarString(v))=> key -> EnvVarValue(v)
      case (name, state.EnvVarSecretRef(secret: String)) => name -> EnvVarSecret(secret)
    }.breakOut
  }
}
```

benchmark/src/main/scala/mesosphere/marathon/json/JsonFormatBenchmark.scala
```diff
@@ -0,0 +1,116 @@
...
+  /**
+    * Attempts to represent a real world-ish [[AppInfo]],
+    * with particular focus on accurately representing sizes of json objects,
+    * which translate to expensive scala Maps.
+    */
+  val realisticAppInfo: AppInfo = {
+
+    /**
+      * Taken from our dev cluster with: {{{
+      *   curl /v2/apps | jq '[.apps[].env | length] | add / length'
```
jeschkies: Cool!

src/main/scala/mesosphere/marathon/raml/EnvVarConversion.scala
htmldoug: Either `Map.newBuilder` or `breakOut` would construct a new Map from scratch. That'd involve rehashing all the keys and the additional memory allocation we're aiming to avoid.

The important bits of .mapValues are here: https://github.com/scala/scala/blob/v2.12.4/src/library/scala/collection/MapLike.scala#L252-L265. Only downside would be if the function is expensive or invoked for the same key many times--neither happens here.

jeschkies: Good point. I'd like to see some benchmarks. We did see issue with `mapValues` being lazy in the past. It is nicer for the GC to make copies.

jeschkies: I've done a quick test with `breakOut` and a builder. Both are slightly slower than this solution. So it's good with me. We probably could achieve more speed in the serialization as well. Further more a lighter response should help as well.

htmldoug: > We probably could achieve more speed in the serialization as well.

I have some ideas around this, but they're not the low hanging fruit that this PR is.

If we could skip the AST step of case class => AST => String (e.g. scala-pickling or sbt-serialization), that'd cut out a good chunk of computation, but it'd involve rewriting that 1,000-line RAML code generator. I'm afraid to touch that.

For remaining optimizations of the current approach, play.json...JsObject uses a mutable.LinkedHashMap internally. I think we could get decent performance gains with a more optimized Map implementation (e.g. guava), but that'd involve hooking into the JsObject package private constructor. If we wanted to avoid touching the Raml generator, we could hook in the replacement with AspectJ since you're already using it with kamon.

If you think performance gains could justify some dark magic, I could through up another PR and benchmarks. Otherwise, if using package private constructors is a deal-breaker, I'll leave it be.

Let me know how you feel, or if you have other optimizations in mind. We're feeling the performance pain, so I wouldn't mind donating some time to it, especially with how generous you've all been with PR reviews and merges!

htmldoug: Correction: The JsObject [*field* is private, not the constructor](https://github.com/playframework/play-json/blob/2.6.7/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala#L125). This may be less dark and dirty than I previously thought.

PR #5782, opened 2017-11-29
 Unify responses from `/apps/{app_id}/tasks`. 

jeschkies:
Summary:
`GET /v2/apps/{app_id}/tasks` would return enriched tasks while
`-X DELETE /v2/apps/{app_id}/tasks/{task_id}` returned something.
This diff is unifying the end points to always return enriched tasks.

I've also prepended all RAML models in `AppControllerTest` with `raml` to
make debugging simpler.

JIRA issues: MARATHON-7811


src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -331,12 +330,29 @@ class AppsController(
               complete((StatusCodes.OK, List(Headers.`Marathon-Deployment-Id`(plan.id)), DeploymentResult(plan.id, plan.version.toOffsetDateTime)))
             }
           case TaskKillingMode.Wipe =>
-            onSuccess(taskKiller.kill(appId, findToKill, wipe = true)) { instances =>
-              complete(raml.InstanceList(instances.map(_.toRaml)))
+            val killInstances = async {
+              val instances = await(taskKiller.kill(appId, findToKill, wipe = true))
+              // TODO: Does it make sense to set health for a killed task?
```
jeschkies: What do you think, @meichstedt?

docs/docs/rest-api/public/api/v2/types/enrichedTask.raml
```diff
@@ -37,6 +37,11 @@ types:
       version?: string
       localVolumes?: LocalVolumeId[]
 
+  EnrichedTaskSingle:
```
alenkacz: do we need the single one? Isn't it enough to keep tasks property, with array and just a single item in it? Then I as a client can use the same deserializer to read it...

jeschkies: Yes. The current API has the `task` field for a single task and `tasks` for multiple. We can change this in a future version.

alenkacz: oh, ok ... that's unfortunate... I thought that we are already making a breaking change by switching from the previous task serialization to this one so it would be ok to break this as well... (since people will have to change the way they deserialize responses anyway)

timcharper: I submit that we just call this "Task" and "TaskList" (Task list, task being singular, is the common grammar).

`EnrichedTask` is an internal concern that API consumers can safely ignore.

timcharper: I submit that we just call this "Task" and "TaskList" (Task list, task being singular, is the common grammar).

EnrichedTask is an internal concern that API consumers can safely ignore.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
timcharper: @jeschkies would it be useful to see why a task was killed? (failed health check?)

I'm inlined towards no, not useful to include health check with killed tasks.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -114,14 +115,26 @@ class AppTasksResource @Inject() (
       val deploymentF = taskKiller.killAndScale(pathId, findToKill, force)
       deploymentResult(result(deploymentF))
     } else {
-      reqToResponse(taskKiller.kill(pathId, findToKill, wipe)) {
-        tasks => ok(jsonObjString("tasks" -> tasks))
+      val response: Future[Response] = async {
+        val instances = await(taskKiller.kill(pathId, findToKill, wipe))
+        val healthStatuses = await(healthCheckManager.statuses(pathId))
+        val enrichedTasks: Seq[EnrichedTask] = instances.map { instance =>
+          val killedTask = instance.appTask // Should this rather be instance.taskMap[taskId]?
```
timcharper: It seems correct to me

docs/docs/rest-api/public/api/v2/types/enrichedTask.raml
meichstedt: Agree with @timcharper. `EnrichedTask` is an internal construct meant to wrap information that needs to be serialized; to the user, it's just `Task` information. `Task/TaskList` makes sense to me. I'd love to see the naming adjusted.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meichstedt: @jeschkies AFAICT the result is meant to provide the information which tasks have been killed. I don't think we need to add health information here and in other locations where we kill tasks via the API, and I think the related logic should be removed.

@timcharper `TaskKiller` will set `KillReason.KillingTasksViaApi`. If the task is killed because it failed health checks too often, `KillReason.FailedHealthChecks` will be logged. IMO, when tasks are killed via the AppsController, `KillingTasksViaApi` is always correct.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -331,17 +329,35 @@ class AppsController(
...
+              complete((StatusCodes.OK, raml.EnrichedTasksList(enrichedTasks.toRaml)))
             }
           case TaskKillingMode.KillWithoutWipe =>
-            onSuccess(taskKiller.kill(appId, findToKill, wipe = false)) { instances =>
-              complete(raml.InstanceList(instances.map(_.toRaml)))
+            val killInstances = async {
+              val instances = await(taskKiller.kill(appId, findToKill, wipe = false))
+              val healthStatuses = await(healthCheckManager.statuses(appId))
+              instances.map { instance =>
+                EnrichedTask(appId, instance.appTask, instance.agentInfo, healthStatuses.getOrElse(instance.instanceId, Nil))
```
meichstedt: I don't have an overview right now, but my initial inclination wrt to refactoring this was to remove `EnrichedTask` and add a serializer from Instance -> Task information. Basically, the serializer drops information about the instance and presents it as if it were a task, more or less.

Maybe that's a follow up PR, not sure about the impact: instead of creating an `appinfo.EnrichedTask` in the controller (that's API layer already), you could either
- directly create a `raml.EnrichedTask` from the instance here and get rid of the conversion OR
- create a conversion from instance -> `raml.EnrichedTask` (or: `raml.Task` as per the above suggestion by @timcharper)

I'd prefer the latter – I'm OK with a follow-up as I assume it's a bigger change (~40 usages of `appInfo.EnrichedTask`)

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
meichstedt: "Fun with TaskIds". Question: if the user requests a resident task with simplified id `app.task.uuid.6` to be killed (i.e. incarnation no 6), but the instance is at incarnation 7 (simplified id `app.task.uuid.7`), will `app.task.uuid.7` be killed?

:trollface: 

(I think this is something we don't need to fix in this API version)

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -146,20 +159,22 @@ class AppTasksResource @Inject() (
...
-      reqToResponse(taskKiller.kill(pathId, findToKill, wipe)) {
-        tasks => tasks.headOption.fold(unknownTask(id))(task => ok(jsonObjString("task" -> task)))
+      val response: Future[Response] = async {
+        val instances = await(taskKiller.kill(pathId, findToKill, wipe))
+        val healthStatuses = await(healthCheckManager.statuses(pathId))
+        instances.headOption match {
+          case None =>
+            unknownTask(id)
+          case Some(instance) =>
+            val killedTask = instance.appTask // Should this rather be instance.taskMap[taskId]?
```
meichstedt: The `appTask` helper was introduced to prevent accessing the map (and helps identifying the locations where we do that). I'd either stick to the helper or get rid of it; a mixture of both is unfortunate.

src/test/scala/mesosphere/marathon/api/v2/SpecInstancesResourceTest.scala
```diff
@@ -72,19 +72,56 @@ class SpecInstancesResourceTest extends UnitTest with GroupCreation {
...
-      val response = appsTaskResource.deleteMany(appId, host, scale = false, force = false, wipe = false, auth.request)
+      val response = appsTaskResource.deleteMany(appId.toString, host, scale = false, force = false, wipe = false, auth.request)
       response.getStatus shouldEqual 200
+
+      val expected =
+        s"""
+           |{ "tasks": [
+           |  {
+           |    "appId" : "/my/app",
+           |    "healthCheckResults" : [ ],
```
meichstedt: According to the raml, `healthCheckResults`, `ipAddresses`, `ports`, `servicePorts` and `localVolumes` are optional fields. It would be consecutive to add `pragma.omitEmpty` so that the fields are left out if empty instead of being rendered as empty arrays.

I'd favor the pragma to be added so the output is less cluttered with empty arrays.

meichstedt:
@meln1k actually backwards compatibility was broken incidentally. It was never intended to render instances as instances in the Apps/AppTasks controllers.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -114,14 +115,26 @@ class AppTasksResource @Inject() (
...
-        tasks => ok(jsonObjString("tasks" -> tasks))
+      val response: Future[Response] = async {
+        val instances = await(taskKiller.kill(pathId, findToKill, wipe))
+        val healthStatuses = await(healthCheckManager.statuses(pathId))
+        val enrichedTasks: Seq[EnrichedTask] = instances.map { instance =>
+          val killedTask = instance.appTask // Should this rather be instance.taskMap[taskId]?
+          val enrichedTask = EnrichedTask(pathId, killedTask, instance.agentInfo, healthStatuses.getOrElse(instance.instanceId, Nil))
+          enrichedTask
+        }
+        ok(jsonObjString("tasks" -> enrichedTasks.toRaml))
```
zen-dog: Why not another (!) raml type with `tasks: EnrichedTasksList`?

docs/docs/rest-api/public/api/v2/types/enrichedTask.raml
jeschkies: So this should be `TaskSingle` then, right?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
jeschkies: Well, there is already a type [`Task`](https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/v2/types/task.raml#L62). But it's not used to I'm replacing it.

src/test/scala/mesosphere/marathon/api/v2/SpecInstancesResourceTest.scala
jeschkies: Isn't that breaking the API again? Currently the RAML is not our ground truth.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
jeschkies: Ok, I stick with the old.

src/test/scala/mesosphere/marathon/api/v2/SpecInstancesResourceTest.scala
timcharper: Yeah, it is breaking. We can't omit empty fields without a major version bump.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
timcharper: Sounds like we can omit the health check information here, then.

src/test/scala/mesosphere/marathon/api/v2/SpecInstancesResourceTest.scala
meichstedt: If the actual API contract is mandatory, the raml shouldn't state otherwise. If this is the case,
I would be inclined to change the raml. The schema shouldn't say fields are optional, if it is **not** safe to omit them.

However, from what I can see and test, `healthCheckResults`, `ports`, `servicePorts` and `localVolumes` are in fact omitted in 1.5.x if they're not set/empty. I am currently not sure whether it is possible to get a task w/o `ipAddresses` but this shouldn't be hard to test.

So, what is breaking if we use `pragma.omitEmpty` on those that have always been omitted (verified: 1.3.10, 1.4.5, 1.5.2)?

PR #5783, opened 2017-11-29
 Pods do not respect hostname: UNIQUE constraint

meln1k:
Pods do not respect hostname: UNIQUE constraint

Summary: Fix for the bug where UNIQUE constraint was not respected if we scale pods.

JIRA issues: MARATHON_EE-1701


src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -120,7 +120,8 @@ class GroupManagerImpl(
             Left(left)
           case Right(changed) =>
             val unversioned = assignDynamicServicePorts(from, changed)
-            val to = GroupVersioningUtil.updateVersionInfoForChangedApps(version, from, unversioned)
+            val toApps = GroupVersioningUtil.updateVersionInfoForChangedApps(version, from, unversioned)
```
jdef: nit: `toApps` and `to` are bad names. suggest: `withUpdatedApps` and `withUpdatedAppsPods`

src/main/scala/mesosphere/marathon/raml/ConstraintConversion.scala
```diff
@@ -47,7 +47,7 @@ trait ConstraintConversion {
       case Protos.Constraint.Operator.UNLIKE => ConstraintOperator.Unlike
       case Protos.Constraint.Operator.MAX_PER => ConstraintOperator.MaxPer
     }
-    Constraint(c.getField, operator, Option(c.getValue))
+    Constraint(c.getField, operator, Option(c.getValue).map(_.trim).filter(_.nonEmpty))
```
jdef: how is this change related to this PR? did leak from another PR or debugging session?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -40,6 +40,8 @@ class Group(
   lazy val transitiveApps: Set[AppDefinition] = transitiveAppsById.values.toSet
   lazy val transitiveAppIds: Set[PathId] = transitiveAppsById.keySet
 
+  lazy val transitivePods: Set[PodDefinition] = transitivePodsById.values.toSet
```
jdef: I think I understand the need for this. I'm wondering how it'll impact benchmarks

src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
```diff
@@ -59,4 +60,53 @@ object GroupVersioningUtil {
...
+            log.info(s"${newPod.id}: upgrade detected for Pod (oldVersion ${oldPod.versionInfo})")
+            oldPod.versionInfo.withConfigChange(newVersion = version)
+          } else if (oldPod.isOnlyScaleChange(newPod)) {
+            log.info(s"${newPod.id}: scaling op detected for Pod (oldVersion ${oldPod.versionInfo})")
+            oldPod.versionInfo.withScaleOrRestartChange(newVersion = version)
+          } else if (oldPod.versionInfo != newPod.versionInfo && newPod.versionInfo == VersionInfo.NoVersion) {
+            log.info(s"${newPod.id}: restart detected for Pod (oldVersion ${oldPod.versionInfo})")
+            oldPod.versionInfo.withScaleOrRestartChange(newVersion = version)
+          } else {
+            oldPod.versionInfo
```
jdef: unfortunate that we still execute the copy a few lines later, even if this doesn't change. only to later throw out the object later on when the versions are detected as unchanged

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -372,5 +377,41 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+      val updateResult = marathon.updatePod(pod.id, scaledPod, force = true)
+
+      Then("The pod is scaled")
+      updateResult should be(OK)
+      waitForDeployment(updateResult)
+
+      And("Size of the pod should still be 1")
+      val status2 = marathon.status(pod.id)
+      status2 should be(OK)
+      status2.value.instances should have size 1
```
jdef: why still 1, because there's only one agent?

src/main/scala/mesosphere/marathon/raml/ConstraintConversion.scala
meln1k: Since the default protobuf value for strings is empty string, we have to manually check and filter them to produce a correct result.

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: Yep.

src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
zen-dog: It's unfortunate that we have this logic twice - for apps and pods. What do you think about moving `versionInfo` into `RunSpec`?

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -372,5 +377,41 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+      waitForDeployment(createResult)
+      waitForPod(pod.id)
+
+      When("The pod config is updated")
+      val scaledPod = pod.copy(instances = 2)
+      val updateResult = marathon.updatePod(pod.id, scaledPod, force = true)
+
+      Then("The pod is scaled")
+      updateResult should be(OK)
+      waitForDeployment(updateResult)
```
zen-dog: Wait, since the pod can't be scaled up this deployment should never end, correct?

src/main/scala/mesosphere/marathon/raml/ConstraintConversion.scala
jdef: are we using proto3? i thought we were still using proto2 and could rely on hasXXX methods to determine the presence of a field

src/main/scala/mesosphere/marathon/upgrade/GroupVersioningUtil.scala
jdef: I kind of hate `versionInfo` and I also despise code duplication. If `versionInfo` is refactored into RunSpec then maybe tackle that in a separate PR? I'd love to see `versionInfo` disappear (and be replaced by a cleaner data structure) because it conflates "object version" (static metadata) with status-related things (the last time this was scaled, or some other config change was made).

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
jdef: ok. that wasn't immediately obvious to me. maybe document it? or not, if other marathoners all understand that there's only one agent (or else if its documented higher up and I missed it)

src/main/scala/mesosphere/marathon/raml/ConstraintConversion.scala
meln1k: According to the [spec](https://developers.google.com/protocol-buffers/docs/proto#optional) for proto2:
> If the default value is not specified for an optional element, a type-specific default value is used instead: for strings, the default value is the empty string.

jdef:
OK, but your code is doing more string work than it needs to. Instead, what
about checking hasValue?
```java

    public static final int VALUE_FIELD_NUMBER = 3;
    private volatile java.lang.Object value_;
    /**
     * <code>optional string value = 3;</code>
     */
    public boolean hasValue() {
      return ((bitField0_ & 0x00000004) == 0x00000004);
    }

```

(see https://raw.githubusercontent.com/mesosphere/marathon/master/src/main/java/mesosphere/marathon/Protos.java)

```scala

Constraint(c.getField, operator, if (c.hasValue()) Option(c.getValue) else None)

```

src/main/scala/mesosphere/marathon/state/Group.scala
meln1k: It seems we can safely make it a `def`.

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -372,5 +378,52 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+      val scaledPod = pod.copy(instances = 2)
+      val updateResult = marathon.updatePod(pod.id, scaledPod, force = true)
+
+      Then("The pod is not scaled")
+      updateResult should be(OK)
+      val queueResult = marathon.launchQueue()
+      val jsQueueResult = queueResult.entityJson
+
+      val queuedRunspecs = (jsQueueResult \ "queue").as[Seq[JsObject]]
+      val jsonPod = queuedRunspecs.find { spec => (spec \ "pod" \ "id").as[String] == s"/$podName" }.get
```
zen-dog: We have `MarathonFacade.launchQueueForAppId` method for this

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -372,5 +378,52 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+
+      val unfulfilledConstraintRejectSummary = (jsonPod \ "processedOffersSummary" \ "rejectSummaryLastOffers").as[Seq[JsObject]]
+        .find { e => (e \ "reason").as[String] == "UnfulfilledConstraint" }.get
+
+      And("unique constraint reject must happen")
+      (unfulfilledConstraintRejectSummary \ "declined").as[Int] should be >= 1
```
zen-dog: You might be too fast with this request (if no offer was matched yet). I would put an `eventually` around it.

src/main/scala/mesosphere/marathon/state/RootGroup.scala
```diff
@@ -408,12 +408,12 @@ class RootGroup(
       Group(
```
alenkacz: please update the comment for this method, it still says `Timestamp(0)`

btw. why do we have this method here anyway? It seems to be used in tests only...

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: Unfortunately It doesn't contain the right object :(

src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -372,5 +378,52 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
       status2 should be(OK)
       status2.value.instances.filter(_.status == raml.PodInstanceState.Stable) should have size 2
     }
+
+    "deploy a simple pod with unique constraint and then " taggedAs WhenEnvSet(envVar, default = "true") in {
```
jeschkies: Why is this tagged?

meln1k: It's a Mesos App test, we don't run them on mac (https://github.com/mesosphere/marathon/#testing)

timcharper: @meln1k huh, that's strange. The Mesos (not docker) tests run just fine on a mac for me. The majority of our integration tests run against a Mesos cluster.

PR #5787, opened 2017-12-01
Benchmark and speed up service port assignment by up to three magnitudes.

jeschkies:
Summary:
As @zen-dog found a big performance degradation when Marathon has already 5000 apps and a new app is added. He found that `GroupManagerImpl.assignDynamicServicePorts` had an exponential runtime.

The benchmark and flamegraph showed that most time was spend in copying and building the `RootGroup` after the update. It also seems that we update all apps even when the get the same port assigned again. This change introduces a filter so that only apps that are new or changed are updated in the `RootGroup`.

```
```

This is only a fix for small app updates. If many apps are updated at once we will still hit this bottleneck. We should look into the creation of `transitiveAppsById` and `transitivePodsById`
which take up most of the runtime.

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
@@ -2,10 +2,11 @@ package mesosphere.marathon
 package upgrade
 
 import java.util.concurrent.TimeUnit
-import mesosphere.marathon.core.pod.{ MesosContainer, BridgeNetwork }
+
+import mesosphere.marathon.core.group.impl.AssignDynamicServiceLogic
+import mesosphere.marathon.core.pod.{ BridgeNetwork, MesosContainer }
 import mesosphere.marathon.raml.{ Endpoint, Image, ImageType, Resources }
 import mesosphere.marathon.state.Container
-
 import mesosphere.marathon.state.PathId._
 import mesosphere.marathon.state._
 import mesosphere.marathon.core.pod.PodDefinition
@@ -15,7 +16,7 @@ import org.openjdk.jmh.infra.Blackhole
 import scala.collection.breakOut
 
 @State(Scope.Benchmark)
-class GroupVersioningUtilBenchmark {
+class GroupBenchmark {
 
   val version = VersionInfo.forNewConfig(Timestamp(1))
 
@@ -48,7 +49,7 @@ class GroupVersioningUtilBenchmark {
               Some(0),
               Seq("tcp"))))))
 
-  @Param(value = Array("100", "5000"))
+  @Param(value = Array("100", "500", "1000", "2500", "5000"))
   var numberOfSavedApps: Int = _
   lazy val ids = 0 until numberOfSavedApps
 
@@ -74,32 +75,34 @@ class GroupVersioningUtilBenchmark {
     tmpGroup
   }
 
-  //  lazy val podPaths: Vector[PathId] = ids.map { podId =>
-  //    val groupPath = childGroupPaths(podId % numberOfGroups)
-  //    groupPath / s"pod-${podId}"
-  //  }(breakOut)
-
-  //  lazy val podDefs: Map[PathId, PodDefinition] = podPaths.map { path =>
-  //    path -> makePod(path)
-  //  }(breakOut)
-
   def upgraded = {
     val appId = childGroupPaths(0) / s"app-$numberOfSavedApps"
     rootGroup.updateApp(appId, (maybeApp) => makeApp(appId))
   }
+
+  object AssignServicePorts extends AssignDynamicServiceLogic {
+    val servicePortsRange: Range = 10000 until 20000
+    val config = AllConf.withTestConfig("--local_port_min", servicePortsRange.start.toString, "--local_port_max", (servicePortsRange.end + 1).toString)
+  }
 }
 
-@OutputTimeUnit(TimeUnit.MICROSECONDS)
-@BenchmarkMode(Array(Mode.Throughput, Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.AverageTime))
 @Fork(1)
-class GroupVersioningUtilBenchmark1 extends GroupVersioningUtilBenchmark {
+class GroupManagerBenchmark extends GroupBenchmark {
+
+  //  @Benchmark
+  //  def updateVersionInfoForChangedApps(hole: Blackhole): Unit = {
+  //    println(rootGroup.groupsById.keys)
+  //    val newRootGroup = GroupVersioningUtil.updateVersionInfoForChangedApps(
+  //      Timestamp (2),
+  //      rootGroup, upgraded)
+  //    hole.consume(newRootGroup)
+  //  }
 
   @Benchmark
-  def updateVersionInfoForChangedApps(hole: Blackhole): Unit = {
-    println(rootGroup.groupsById.keys)
-    val newRootGroup = GroupVersioningUtil.updateVersionInfoForChangedApps(
-      Timestamp (2),
-      rootGroup, upgraded)
+  def assignDynamicServicePorts(hole: Blackhole): Unit = {
+    val newRootGroup = AssignServicePorts.assignDynamicServicePorts(rootGroup, upgraded)
     hole.consume(newRootGroup)
   }
 }
--- a/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
@@ -91,14 +91,14 @@ class GroupBenchmark {
 @Fork(1)
 class GroupManagerBenchmark extends GroupBenchmark {
 
-  //  @Benchmark
-  //  def updateVersionInfoForChangedApps(hole: Blackhole): Unit = {
-  //    println(rootGroup.groupsById.keys)
-  //    val newRootGroup = GroupVersioningUtil.updateVersionInfoForChangedApps(
-  //      Timestamp (2),
-  //      rootGroup, upgraded)
-  //    hole.consume(newRootGroup)
-  //  }
+  @Benchmark
+  def updateVersionInfoForChangedApps(hole: Blackhole): Unit = {
+    println(rootGroup.groupsById.keys)
+    val newRootGroup = GroupVersioningUtil.updateVersionInfoForChangedApps(
+      Timestamp (2),
+      rootGroup, upgraded)
+    hole.consume(newRootGroup)
+  }
 
   @Benchmark
   def assignDynamicServicePorts(hole: Blackhole): Unit = {
--- a/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
@@ -35,7 +35,7 @@ class GroupBenchmark {
       id = path,
       networks = Seq(BridgeNetwork()),
       labels = Map("ID" -> path.toString),
-      version = version.lastConfigChangeAt,
+      versionInfo = version,
 
       containers = Seq(
         MesosContainer(
--- a/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
@@ -79,11 +79,6 @@ class GroupBenchmark {
     val appId = childGroupPaths(0) / s"app-$numberOfSavedApps"
     rootGroup.updateApp(appId, (maybeApp) => makeApp(appId))
   }
-
-  object AssignServicePorts extends AssignDynamicServiceLogic {
-    val servicePortsRange: Range = 10000 until 20000
-    val config = AllConf.withTestConfig("--local_port_min", servicePortsRange.start.toString, "--local_port_max", (servicePortsRange.end + 1).toString)
-  }
 }
 
 @OutputTimeUnit(TimeUnit.MILLISECONDS)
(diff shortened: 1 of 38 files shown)


jeschkies:
**Note**: This change should also include

* Tutorial for the flamegraph creation.
* Documentation of `assignDynamicServicePorts`

jdef:
looked at changes made to `assignDynamicServicePorts` and the assumption made by the optimization appears to be: once service ports are assigned to an app, and for as long as that app isn't changed, then service ports don't need to be reassigned for said app when handling bulk updates. I think this is OK because I think Marathon never returns to the user an app def w/ servicePort = 0 (though someone may want to double-check me on this)

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -28,11 +28,103 @@ import scala.concurrent.{ Await, ExecutionContext, Future }
...
+      case None => true // new app
+      case Some(oldApp) => oldApp.isUpgrade(newApp)
+    }
+  }
+
+  def assignDynamicServicePorts(from: RootGroup, to: RootGroup): RootGroup = {
+    val portRange = Range(config.localPortMin(), config.localPortMax())
+    var taken = from.transitiveApps.flatMap(_.servicePorts) ++ to.transitiveApps.flatMap(_.servicePorts)
+
+    def nextGlobalFreePort: Int = {
```
meln1k: What do you think about changing the method signature to`nextGlobalFreePort()` since this method has side effects?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -28,11 +28,103 @@ import scala.concurrent.{ Await, ExecutionContext, Future }
...
+  def changedOrNew(originalApps: Map[AppDefinition.AppKey, AppDefinition], newApp: AppDefinition): Boolean = {
+    originalApps.get(newApp.id) match {
+      case None => true // new app
+      case Some(oldApp) => oldApp.isUpgrade(newApp)
+    }
+  }
+
+  def assignDynamicServicePorts(from: RootGroup, to: RootGroup): RootGroup = {
+    val portRange = Range(config.localPortMin(), config.localPortMax())
+    var taken = from.transitiveApps.flatMap(_.servicePorts) ++ to.transitiveApps.flatMap(_.servicePorts)
```
meln1k: Since we're trying to get this code fast, what do you think about using a mutable set here and initialize it with a while loop?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -28,11 +28,103 @@ import scala.concurrent.{ Await, ExecutionContext, Future }
...
+      portDefinitions: Seq[PortDefinition],
+      servicePorts: Seq[Int]): Seq[PortDefinition] =
+      if (portDefinitions.nonEmpty)
+        portDefinitions.zipAll(servicePorts, AppDefinition.RandomPortDefinition, AppDefinition.RandomPortValue).map {
+          case (portDefinition, servicePort) => portDefinition.copy(port = servicePort)
+        }
+      else Seq.empty
+
+    def assignPorts(newApp: AppDefinition): AppDefinition = {
+      //all ports that are already assigned in old app definition, but not used in the new definition
```
alenkacz: spaces are missing after //

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -28,11 +28,103 @@ import scala.concurrent.{ Await, ExecutionContext, Future }
 import scala.util.control.NonFatal
 import scala.util.{ Failure, Success }
 
+trait AssignDynamicServiceLogic extends StrictLogging {
```
zen-dog: Why do we need an extra trait just to cover this method?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -32,6 +32,19 @@ trait AssignDynamicServiceLogic extends StrictLogging {
...
+    * @return true if app is new or an updated, false otherwise.
+    */
+  def changedOrNew(originalApps: Map[AppDefinition.AppKey, AppDefinition], newApp: AppDefinition): Boolean = {
+    originalApps.get(newApp.id) match {
+      case None => true // new app
+      case Some(oldApp) => oldApp.isUpgrade(newApp)
+    }
+  }
+
   def assignDynamicServicePorts(from: RootGroup, to: RootGroup): RootGroup = {
```
zen-dog: This whole method needs a loooooong comment explaining what it does and how.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
zen-dog: With the changes below the updates to this map should be done for changed apps/changed ports. I wonder if a mutable structure would make any difference here. But then again - we have a benchmark now so we can easily test - data beats opinion!

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -32,6 +32,19 @@ trait AssignDynamicServiceLogic extends StrictLogging {
...
     val portRange = Range(config.localPortMin(), config.localPortMax())
     var taken = from.transitiveApps.flatMap(_.servicePorts) ++ to.transitiveApps.flatMap(_.servicePorts)
```
zen-dog: Please add types for both variables.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
jeschkies: I think the improvement would be minimal since 30% of the time is spent in the RootGroup creation.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
jeschkies: Good point.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
unterstein: +1

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
```diff
@@ -0,0 +1,108 @@
...
+          resources = Resources(1.0),
+          image = Some(Image(ImageType.Docker, "alpine")),
+          endpoints = List(
+            Endpoint(
+              "service",
+              Some(2015),
+              Some(0),
+              Seq("tcp"))))))
+
+  @Param(value = Array("100", "500", "1000", "2500", "5000")) //  @Param(value = Array("5000"))
```
unterstein: why the comment?

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
```diff
@@ -0,0 +1,108 @@
...
+  var numberOfSavedApps: Int = _
+  lazy val ids = 0 until numberOfSavedApps
+
+  @Param(value = Array("5", "15")) //  @Param(value = Array("5"))
```
unterstein: why the commented?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -28,11 +28,103 @@ import scala.concurrent.{ Await, ExecutionContext, Future }
...
+      newApp.copy(
+        portDefinitions = mergeServicePortsAndPortDefinitions(newApp.portDefinitions, servicePorts),
+        container = updatedContainer.orElse(newApp.container)
+      )
+    }
+
+    val dynamicApps: Iterator[AppDefinition] =
+      to.transitiveApps
+        .iterator
+        .filter{ newApp => changedOrNew(from.transitiveAppsById, newApp) }
```
unterstein: 1. 🎉 !
2. `.filter{` -> `.filter {` ?

timcharper:
@jdef this sounds right to me; I've never seen Marathon run port 0.

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
jeschkies: It's easier to test this way. We don't need mocked repositories etc for testing this. I'd love a better pattern but cannot think of one.

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
```diff
@@ -0,0 +1,108 @@
...
+}
+
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(1)
+class GroupManagerBenchmark extends GroupBenchmark {
+
+  @Benchmark
+  def updateVersionInfoForChangedApps(hole: Blackhole): Unit = {
+    println(rootGroup.groupsById.keys)
```
timcharper: Intentional to leave this println statement?

timcharper:
@jeschkies please see my patch in which I further separate concerns, and increase the overall performance of the method by an additional 20%

https://github.com/mesosphere/marathon/tree/tharper/benchmark-service-port-assignment

Before:

```
[info] Benchmark                                        (numberOfGroups)  (numberOfSavedApps)  Mode  Cnt   Score   Error  Units
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                  100  avgt   10   0.121 ± 0.009  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                  500  avgt   10   0.614 ± 0.041  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                 1000  avgt   10   1.331 ± 0.129  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                 2500  avgt   10   4.037 ± 0.216  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                 5000  avgt   10  10.611 ± 0.313  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                  100  avgt   10   0.122 ± 0.005  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                  500  avgt   10   0.576 ± 0.055  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                 1000  avgt   10   1.220 ± 0.107  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                 2500  avgt   10   3.597 ± 0.255  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                 5000  avgt   10   9.350 ± 0.479  ms/op
```

After:

```
[info] Benchmark                                        (numberOfGroups)  (numberOfSavedApps)  Mode  Cnt  Score   Error  Units
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                  100  avgt   10  0.089 ± 0.024  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                  500  avgt   10  0.538 ± 0.078  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                 1000  avgt   10  1.070 ± 0.073  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                 2500  avgt   10  3.443 ± 0.421  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                 5                 5000  avgt   10  8.230 ± 0.341  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                  100  avgt   10  0.084 ± 0.006  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                  500  avgt   10  0.429 ± 0.009  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                 1000  avgt   10  0.932 ± 0.031  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                 2500  avgt   10  2.527 ± 0.111  ms/op
[info] GroupManagerBenchmark.assignDynamicServicePorts                15                 5000  avgt   10  6.582 ± 0.373  ms/op
```

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
jeschkies: No 😞

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
jeschkies: No. It was just  a fast fix. I merged your patch.

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
jeschkies: Leftovers from testing.

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
jeschkies: I won't tell you 😉.

zen-dog:
To give you an idea what this patch means. Here are 2 graphs (before and after patch) for sending marathon 1.6k "create app" requests (each one creating 1 app):
Before
After
`x-axis:` number of requests
`y-axis:` response time in `ms` per request
Note: this is just time-to-response (which is not the same as actually finishing the deployment). 

Total time to send all 1.6k requests went from `442,262 ms` to `93,491 ms`, so almost a *5x* improvement!

PR #5788, opened 2017-12-01
Collect diagnostics bundle for failed SI job MARATHON-7855

alenkacz:
Summary:
Collecting dcos diagnostics bundle will allow us to debug failures more effectively.

JIRA issues: MARATHON-7855

### DIFF
--- a/Jenkinsfile.mom.si
+++ b/Jenkinsfile.mom.si
@@ -47,6 +47,7 @@ node('shakedown') {
                """
           } finally {
             junit allowEmptyResults: true, testResults: 'shakedown.xml'
+            archive includes: "diagnostics.zip"
           }
         }
     }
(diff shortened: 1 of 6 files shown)


alenkacz:
You can verify the result here MARATHON-7855

Jenkinsfile.mom.si
```diff
@@ -47,6 +47,7 @@ node('shakedown') {
                """
           } finally {
             junit allowEmptyResults: true, testResults: 'shakedown.xml'
+            archive includes: "diagnostics.zip"
```
zen-dog: Here shakedown is called directly (not `si_pipeline`) - the bundle will not be created. Besides I don't think Jenkinsfile is used anywhere.

PR #5789, opened 2017-12-04
Add documentation on fault domain awareness

sascala:
I added some docs for fault domain awareness in marathon app and pod definitions. I added one question regarding whether the defaults I document are dc/os-specific or not.

Summary:

JIRA issues:

### DIFF
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -16,7 +16,7 @@ All Marathon operators are supported when the field name is `@hostname`.
 
 ### Region and zone as field names
 
-Use the `@region` and `@zone` field names to configure [fault domain awareness and capacity extension]({{ site.baseurl }}/docs/fault-domain-awareness.html).
+Use @region and @zone as field names to configure [fault domain awareness]({{ site.baseurl }}/docs/fault-domain-awareness.html).
 
 ### Attribute as field name
 
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -16,7 +16,7 @@ All Marathon operators are supported when the field name is `@hostname`.
 
 ### Region and zone as field names
 
-Use @region and @zone as field names to configure [fault domain awareness]({{ site.baseurl }}/docs/fault-domain-awareness.html).
+Use `@region` and `@zone` as field names to configure [fault domain awareness]({{ site.baseurl }}/docs/fault-domain-awareness.html).
 
 ### Attribute as field name
 
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -8,15 +8,19 @@ Constraints control where apps run to allow optimizing for either fault toleranc
 
 ## Field Names
 
-### Hostname as field name
+### Hostname
 
-Entering `hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
+Entering `@hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
 
-All Marathon operators are supported when the field name is `hostname`.
+All Marathon operators are supported when the field name is `@hostname`.
+
+### Region and zone
+
+Use `@region` and `@zone` to configure [fault domain awareness and capacity extension]({{ site.baseurl }}/docs/fault-domain-awareness.html).
 
 ### Attribute as field name
 
-If hostname is not specified as the field name, then the field name is interpreted as a Mesos agent node attribute. A Mesos agent node attribute allows you to tag an agent node. See `mesos-slave --help` to learn how to set the attributes.
+If `@hostname`, `@region`, or `@zone` are not specified as field names, then the field name is interpreted as a Mesos agent node attribute. A Mesos agent node attribute allows you to tag an agent node. See `mesos-slave --help` to learn how to set the attributes.
 
 If the specified attribute is not defined on the agent node, most operators will refuse to run tasks on it. In fact, only the `UNLIKE` operator will (and always will) accept this offer for now, while other operators will always refuse it.
 
--- a/docs/docs/constraints.md
+++ b/docs/docs/constraints.md
@@ -8,15 +8,15 @@ Constraints control where apps run to allow optimizing for either fault toleranc
 
 ## Field Names
 
-### Hostname
+### Hostname as field name
 
 Entering `@hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
 
 All Marathon operators are supported when the field name is `@hostname`.
 
-### Region and zone
+### Region and zone as field names
 
-Use `@region` and `@zone` to configure [fault domain awareness and capacity extension]({{ site.baseurl }}/docs/fault-domain-awareness.html).
+Use the `@region` and `@zone` field names to configure [fault domain awareness and capacity extension]({{ site.baseurl }}/docs/fault-domain-awareness.html).
 
 ### Attribute as field name
 
--- a/docs/docs/fault-domain-awareness.md
+++ b/docs/docs/fault-domain-awareness.md
@@ -19,7 +19,7 @@ For on-premise deployments, a zone would be a physical data center rack.
 
 For public cloud deployments, a zone would be the "availability zone” concept provided by most cloud providers.
 	
-Spread your services across zones if you are latency-sensitive or if you need a moderate degree of high availability: you can place your apps in one region, but balance across zones.
+Spread your apps or pods across zones if you are latency-sensitive or if you need a moderate degree of high availability: you can place your apps in one region, but balance across zones.
 
 ## Region fault domains
 
@@ -29,7 +29,7 @@ For on-premise deployments, a region might be a data center.
 	 
 For public cloud deployments, most cloud providers expose a “region” concept.
 	 
-You can deploy your services in a specific region region based on the available capacity. <!-- can you say more about what you mean here? -->
+You can deploy your apps or pods in a specific region region based on the available capacity.
 
 ### Local and remote regions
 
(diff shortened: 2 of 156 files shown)


docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Adding fault domain awareness to your app or pod definition
+
+In your Marathon app or pod definition, you can use placement constraints to:
+
+- Specify a region and zone for your app or pod, so that all instances of the app or pod will be scheduled only in that region and zone.
+
+- Specify a region without a specific zone, so that all instances of a given app or pod will be scheduled in that region (but not necessarily in the same zone).
+
+### Placement constraint guidelines
```
sascala: are these constraints DC/OS-specific?

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
+---
+title: Fault Domain Awareness
+---
+
+# Fault Domain Awareness
+
+## Overview
+
+A fault domain is a set of nodes that share similar failure and latency characteristics. Two nodes in the same fault domain are both affected by failure events within the domain. Placing nodes in more than one fault domain reduces the risk that a failure will affect both nodes.
```
timcharper: nodes usually refer to the server or VM running an instance. Would it be more clear to say:

> Placing _instances_ in more than one fault domain...

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+Marathon supports fault domain awareness, so that you can mitigate the risk of fault domain failure for high availability and configure your apps and pods to take advantage of region awareness to allow for increased capacity when needed.
```
timcharper: Sentence feels run on; split it up?

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+Marathon currently offers hierarchical 2-level fault domain: zone and region.
```
timcharper: > Marathon currently supports Mesos's 2-level hierarchical fault domains

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+	
+## Zone fault domains
+Zone fault domains offer a moderate degree of fault isolation because they share the same region. However, network latency between zones in the same region is moderately low (typically < 10ms).
+	
+For on-premise deployments, a zone would be a physical data center rack. 
+
+For public cloud deployments, a zone would be the "availability zone” concept provided by most cloud providers.
+	
+Spread your apps or pods across zones if you are latency-sensitive or if you need a moderate degree of high availability: you can place your apps in one region, but balance across zones.
```
timcharper: It seems like the conditions should come first?

> If you are latency-sensitive, or if your goal is high-availability, place your instances in a one region and balance them across zones.

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Region fault domains
+
+Region fault domains offer the most fault isolation, though inter-region network latency is high (typically 50-100ms ). 
+	 
+For on-premise deployments, a region might be a data center.
+	 
+For public cloud deployments, most cloud providers expose a “region” concept.
+	 
+You can deploy your apps or pods in a specific region region based on the available capacity.
```
timcharper: I might suggest that we consistently just call them "instances" instead of apps or pods (it _is_, after all, actually the instances we balance)

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Local and remote regions
+
+- A **local region** is a region that contains the master nodes.
```
timcharper: > The **local region** is the region running the Mesos master nodes.

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+- A **remote region** contains only agent nodes. There is high latency between a remote region and the local region.
```
timcharper: > contains only _Mesos_ agent nodes

> There is _usually_ high latency

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Mesos installation considerations
+
+- Masters nodes must be in the same region because otherwise the latency between them will be too high. They should, however, be spread across zones for fault-tolerance.
```
timcharper: > _Mesos Master_ nodes must be in the same region

(also, we should note that Marathon should run in the local region, also)

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+}
+```
+
+- Instances will all be launched in the `aws-us-east1` region and evenly divided between `aws-us-east1`‘s zones `a`,`b`,`c`,`d`.
+
+### Use Case: Use Remote Regions to Increase Capacity
+
+You can configure apps and pods to use remote regions when extra capacity is needed. Your cluster must consist of one local region, which contains the master agents, system services, and agents, and one or more remote regions, which will contain only agents.
+
+To increase capacity, you will add new agents to a remote region or regions of your cluster, and then update your apps or pods to be launched in that region or those regions appropriately.
```
timcharper: I might suggest that we take out any reference to launch in multiple regions. Sadly, due to time constraints and needing to sync with other products supporting the constraint language, we did not implement the IN constraint operator. Launching an app in multiple regions, right now, requires that the operator write a regular expression. I don't know if we want to even mention that, or provide a guide to do so. Regular require a high degree of expertise to get correct.

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+### Local and remote regions
+
+- A **local region** is a region that contains the master nodes.
+- A **remote region** contains only agent nodes. There is high latency between a remote region and the local region.
+
+## Mesos installation considerations
+
+- Masters nodes must be in the same region because otherwise the latency between them will be too high. They should, however, be spread across zones for fault-tolerance.
+	
+- You must have less than 100 ms latency between regions.
```
alenkacz: maybe it is just myself but ... why? Or was it supposed to be between zones, not regions?

docs/docs/fault-domain-awareness.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Adding fault domain awareness to your app or pod definition
+
+In your Marathon app or pod definition, you can use placement constraints to:
```
alenkacz: maybe add link to http://mesosphere.github.io/marathon/docs/constraints.html ?

docs/docs/fault-domain-awareness.md
kensipe: I'm not sure where we got the definition of fault domain but it seemed narrow.   There are different types of fault domains... at a node level (if a node is lost all services on it are lost), at a rack level (if a rack is lost we lose all nodes on the rack), at a cluster level (if a router goes out all services are lost), at a region level.

from wikipedia:  a failure domain encompasses a section of a network that is negatively affected when a critical device or network service experiences problems. The size of a failure domain and its potential impact depends on the device or service that is malfunctioning

docs/docs/fault-domain-awareness.md
kensipe: we should state the specific type of fault domain we are referring to.
Marathon previously supported node and rack fault domains... Marathon now supports region.   
Although now as I'm writing it... I'm not sure we REALLY do.   We do support workload bursting to other regions which is part of the equation.

docs/docs/fault-domain-awareness.md
kensipe: +1

docs/docs/fault-domain-awareness.md
kensipe: I would prefer that we DO mention it... that it is an advanced feature that requires regex at the moment.   As a user, I would rather have transparency from the product team and understand what is possible.   Perhaps it will drive the "IN" feature request externally as well.

docs/docs/fault-domain-awareness.md
sascala: that makes a lot of sense.

docs/docs/fault-domain-awareness.md
sascala: In case I do end up mentioning this functionality, can either of you tell me more about what the regex needs to do?

docs/docs/fault-domain-awareness.md
sascala: can do

docs/docs/fault-domain-awareness.md
sascala: @vinodkone said it's just to be safe; they haven't yet tested what latency can a master/agent communication can tolerate

docs/docs/fault-domain-awareness.md
sascala: Latest commit addresses all comments except this one -- I'm checking with @vinodkone later to see how to proceed.

docs/docs/constraints.md
```diff
@@ -8,15 +8,15 @@ Constraints control where apps run to allow optimizing for either fault toleranc
...
 
 Entering `@hostname` as the field name matches the agent node hostnames. See `UNIQUE operator`, below, for a usage example.
 
 All Marathon operators are supported when the field name is `@hostname`.
 
-### Region and zone
+### Region and zone as field names
 
-Use `@region` and `@zone` to configure [fault domain awareness and capacity extension]({{ site.baseurl }}/docs/fault-domain-awareness.html).
+Use the `@region` and `@zone` field names to configure [fault domain awareness and capacity extension]({{ site.baseurl }}/docs/fault-domain-awareness.html).
```
timcharper: Hmm, this seems off. (it was off before)

> Use the `@region` and `@zone` field names to configure fault domain awareness and capacity extension

A) The grammar is suffering. Should there be a "the" in there?
B) I don't think it's an extension. Support for Mesos fault domains is built-in to Marathon. It's on by default.
C) I just can't make sense of the sentence. The more I read it, the more confused I am.

Maybe this would be better?

> The `@region` and `@zone` fields refer to the respective [fault domain properties]({{ site.baseurl }}/docs/fault-domain-awareness.html). Specifying a constraint on `@region` will enable an instance to be placed in remote regions.

docs/docs/fault-domain-awareness.md
kensipe: @sascala it doesn't seem like any of my comments were addressed

PR #5792, opened 2017-12-05
Properly verify that the leader election is finished

alenkacz:
Summary:
After multi-master switch in SI tests these tests were running, but were extremely flaky. The reason for that is that shakedown.marathon_leader_ip() asks the mesosDNS for the current leader, but mesosDNS is updated only every 30s. So it might happen that shakedown.marathon_leader_ip() still returns old leader ip for some time after the successful election.
In the end I just ended up retrying on mesos dns call instead of using v2/info endpoint because of the comment that @jdef made here https://github.com/mesosphere/marathon/pull/5792#pullrequestreview-81203937

JIRA issues: MARATHON-7942

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -711,7 +711,7 @@ def deployment_wait(timeout=120, service_id=None):
     shakedown.time_wait(lambda: deployment_predicate(service_id), timeout)
 
 
-@retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=ignore_exception)
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=60, retry_on_exception=ignore_exception)
 def __marathon_leadership_changed_in_mesosDNS(original_leader):
     """ This method uses mesosDNS to verify that the leadership changed.
         We have to retry because mesosDNS checks for changes only every 30s.
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -366,7 +366,7 @@ def is_enterprise_cli_package_installed():
     try:
         result_json = json.loads(stdout)
     except JSONDecodeError as error:
-        raise DCOSException('Could not parse: "{}"'.format(stdout)) from error
+        raise DCOSException('Could not parse: "{}"'.format(stdout))(error)
     return any(cmd['name'] == 'dcos-enterprise-cli' for cmd in result_json)
 
 
@@ -709,6 +709,7 @@ def deployment_wait(timeout=120, service_id=None):
         the dcos-cli and remove this method later.
     """
     shakedown.time_wait(lambda: deployment_predicate(service_id), timeout)
+
 
 @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=ignore_exception)
 def marathon_leadership_changed(original_leader):
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -6,6 +6,7 @@ import shlex
 import time
 import uuid
 import sys
+import retrying
 
 from datetime import timedelta
 from dcos import http, mesos
@@ -333,9 +334,8 @@ def get_marathon_leader_not_on_master_leader_node():
     if marathon_leader == master_leader:
         delete_marathon_path('v2/leader')
         shakedown.wait_for_service_endpoint('marathon', timedelta(minutes=5).total_seconds())
-        new_leader = shakedown.marathon_leader_ip()
-        assert new_leader != marathon_leader, "A new Marathon leader has not been elected"
-        marathon_leader = new_leader
+        assert_marathon_leadership_changed(marathon_leader)
+        new_leader = marathon_leader()
         print('switched leader to: {}'.format(marathon_leader))
 
     return marathon_leader
@@ -709,3 +709,17 @@ def deployment_wait(timeout=120, service_id=None):
         the dcos-cli and remove this method later.
     """
     shakedown.time_wait(lambda: deployment_predicate(service_id), timeout)
+
+def get_marathon_leader():
+    client = marathon.create_client()
+    about = client.get_about()
+    return about.get("leader")
+
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=ignore_exception)
+def assert_marathon_leadership_changed(original_leader):
+    """ We have to retry here because leader election takes time and what might happen is that some nodes might
+        not be aware of the new leader being elected resulting in HTTP 502.
+    """
+    current_leader = get_marathon_leader()
+    print('leader: {}'.format(current_leader))
+    assert original_leader != current_leader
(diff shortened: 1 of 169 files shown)


tests/system/common.py
```diff
@@ -709,3 +709,17 @@ def deployment_wait(timeout=120, service_id=None):
...
     """
     shakedown.time_wait(lambda: deployment_predicate(service_id), timeout)
+
+def get_marathon_leader():
+    client = marathon.create_client()
+    about = client.get_about()
+    return about.get("leader")
+
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=ignore_exception)
+def assert_marathon_leadership_changed(original_leader):
```
jdef: querying marathon for the information is interesting. because this is for a marathon test that's probably trying to test whether marathon is doing the right thing, but it's also *relying* on marathon to do the right thing (and report it as such in the /v2/info endpoint). in other words, if marathon went split-brain, how would your test know that /v2/info was reliable?

using DNS might have been flaky (because of stale data) but at least forced the test to query a 3rd party to validate who the leader was. there's value in asking a 3rd party because if there's a split-brain bug in marathon then it might not be generating reliable information in /v2/info.

an alternative 3rd party source of leader information would be the mesos v1 master/operator API (see GET_FRAMEWORKS) https://github.com/apache/mesos/blob/7beea9cbe083caf738fb02b9d58dadf3da900f4b/include/mesos/v1/master/master.proto#L61

tests/system/common.py
```diff
@@ -709,3 +709,13 @@ def deployment_wait(timeout=120, service_id=None):
...
+
+
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=ignore_exception)
+def marathon_leadership_changed(original_leader):
+    """ This method uses mesosDNS to verify that the leadership changed.
+        We have to retry because mesosDNS checks for changes only every 30s.
+    """
+    current_leader = shakedown.marathon_leader_ip()
+    print('leader: {}'.format(current_leader))
+    assert original_leader != current_leader
```
kensipe: YES!!!

tests/system/common.py
zen-dog: Interesting argument. However I would argue that the opposite it true. In case of split-brain zookeeper would:

> After the partition, all the servers in the minority region will get shutdown and moves to leader election phase. All the client sessions connected to these servers will be disconnected and will receive “KeeperState.Disconnected” event to their watchers, if any registered. A leader election will happen in both partitions and minority partition won’t be able to elect a leader. Therefore, will shutdown. Majority partition will elect a new leader and continue operation.

Hence marathon, connected to a minority partition should also shutdown, restart and won't be able to operate because there is no leader. Hence `v2/info` will not contain any leader information. 

I would argue that relying on `v2/info` is much closer to what "marathon should do", accurately mirrors what should happen during a network partition and doesn't bring in "laggy dependencies" like MesosDNS.

jdef:
Your argument is basically trusting marathon to do the right thing (and
asking marathon to confirm it) vs. asking a 3rd party "hey, did marathon
actually re-register a new leader?"

Given that this is a SI test (and probably shouldn't be making assumptions
about the underlying implementation of Marathon leader election) I think
asking for 3rd party verification is more appropriate. Another benefit of
using the mesos v1 master/operator API (get_frameworks) to confirm marathon
leadership is this: if multiple marathon frameworks register with the same
name (which I think is still allowed by mesos) then the v1 operator API
will report it (I don't know if mesos-dns even takes this into account).

zen-dog:
> Your argument is basically trusting marathon to do the right thing (and
asking marathon to confirm it) vs. asking a 3rd party "hey, did marathon
actually re-register a new leader?"

Rather my argument is that zookeeper is already the 3rd party we relying on. Asking MesosDNS would be asking a 4th party. 

> Another benefit of using the mesos v1 master/operator API (get_frameworks) to confirm marathon
leadership is this: if multiple marathon frameworks register with the same
name (which I think is still allowed by mesos) then the v1 operator API
will report it (I don't know if mesos-dns even takes this into account).

Asking mesos as the ultimate source of truth can be a good safety net.

alenkacz:
@zen-dog I am actually convinced by @jdef arguments and I think that we should stick with using mesos-dns for that. With the up-to 30s retry it should now work reliably...

zen-dog:
> I am actually convinced by @jdef arguments and I think that we should stick with using mesos-dns for that.

I'm not convinced yet. I think we already have our 3rd party in form of Zookeeper. And we would want to rely on smth. else I would rather rely on the admin router (than on MesosDNS) since it is *the* discovery service for all other dc/os components.

jdef:
doesn't adminrouter ping the marathon /leader endpoint to discover the
leader? which means .. trusting the thing that you're trying to test :(

tests/system/common.py
```diff
@@ -333,9 +334,8 @@ def get_marathon_leader_not_on_master_leader_node():
     if marathon_leader == master_leader:
         delete_marathon_path('v2/leader')
         shakedown.wait_for_service_endpoint('marathon', timedelta(minutes=5).total_seconds())
```
zen-dog: If you want to be explicit and test admin router *and* mesosDNS I would make it explicit. Smth. along the lines of:
```
// Check if changes propagated to admin router
shakedown.wait_for_service_endpoint('marathon', timedelta(minutes=5).total_seconds())
with service_endpoint:
   GET /v2/leader
   assert leader != old_leader
// Check mesosDNS too:
marathon_leadership_changed(marathon_leader)
```

alenkacz: done

tests/system/common.py
```diff
@@ -709,3 +709,31 @@ def deployment_wait(timeout=120, service_id=None):
         the dcos-cli and remove this method later.
     """
     shakedown.time_wait(lambda: deployment_predicate(service_id), timeout)
+
+
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=ignore_exception)
```
zen-dog: Make it `1min` to be on the safe side

tests/system/common.py
```diff
@@ -709,3 +709,31 @@ def deployment_wait(timeout=120, service_id=None):
...
+    current_leader = shakedown.marathon_leader_ip()
+    print('leader according to MesosDNS: {}'.format(current_leader))
+    assert original_leader != current_leader
+
+
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=10, retry_on_exception=ignore_exception)
+def __marathon_leadership_changed_in_marathon_api(original_leader):
+    """ This method uses Marathon API to figure out that leadership changed.
+        We have to retry here because leader election takes time and what might happen is that some nodes might
+        not be aware of the new leader being elected resulting in HTTP 502.
```
zen-dog: you mean admin router could not be aware of new marathon leader resulting in 502

PR #5795, opened 2017-12-05
Fix for event stream tests with multimaster

alenkacz:
Summary:
Event stream tests were failing after switch to multimaster. The reason for that is that all the commands were talking to (mesos) master but there are multiple masters now so the commands might be executed on different nodes. Now I do the curl always on marathon leader to make that deterministic.

JIRA issues: MARATHON-7943

### DIFF
--- a/tests/system/fixtures/__init__.py
+++ b/tests/system/fixtures/__init__.py
@@ -42,7 +42,7 @@ def events_to_file():
         shakedown.run_command(leader_ip,
             '(curl --compressed -H "Cache-Control: no-cache" -H "Accept: text/event-stream" ' +
             '-H "Authorization: token={}" '.format(shakedown.dcos_acs_token()) +
-            '-o events.txt -k https://marathon.mesos:8443/v2/events; echo $? > events.exitcode) &'.format(leader_ip))
+            '-o events.txt -k https://marathon.mesos:8443/v2/events; echo $? > events.exitcode) &')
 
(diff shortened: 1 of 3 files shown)


kensipe:
doesn't pass flake8

```
tests/system/fixtures/__init__.py:43:13: E128 continuation line under-indented for visual indent
tests/system/fixtures/__init__.py:44:13: E128 continuation line under-indented for visual indent
tests/system/fixtures/__init__.py:45:13: E128 continuation line under-indented for visual indent
tests/system/fixtures/__init__.py:50:13: E128 continuation line under-indented for visual indent
tests/system/fixtures/__init__.py:51:13: E128 continuation line under-indented for visual indent
```

kensipe:
@alenkacz fixed the python syntax issues and rebased

zen-dog:
> I'm ok with this... but I would MUCH rather have a run_command_on_leader function in shakedown.
we could move this method into our `common.py * Methods to move to shakedown in the future*` and migrate them eventually. However I'm more concerned about making marathon SI tests green now.

kensipe:
https://github.com/dcos/shakedown/pull/178

zen-dog:
> Is it hard to extend shakedown? Or why not do it right ahead?

you'll need to extend shakedown (PR) and release a new version. Both can take some time. It's usually faster to fix it in `common.py` and move a bunch of methods to shakedown eventually.

PR #5796, opened 2017-12-05
Benchmark and speed up service port assignment by up to three magnitu…

jeschkies:
…des. (#5787)

Summary:
As @zen-dog found a big performance degradation when Marathon has already 5000
apps and a new app is added. He found that `GroupManagerImpl.assignDynamicServicePorts`
had an exponential runtime.

The benchmark and flamegraph showed that most time was spend in copying and
building the `RootGroup` after the update. It also seems that we update all
apps even when the get the same port assigned again. This change introduces a
filter so that only apps that are new or changed are updated in the `RootGroup`.

```
```

This is only a fix for small app updates. If many apps are updated at once we
will still hit this bottleneck. We should look into the creation of
`transitiveAppsById` and `transitivePodsById` which take up most of the runtime.

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
@@ -1,0 +1,104 @@
+package mesosphere.marathon
+package upgrade
+
+import java.util.concurrent.TimeUnit
+
+import mesosphere.marathon.core.group.impl.AssignDynamicServiceLogic
+import mesosphere.marathon.core.pod.{ BridgeNetwork, MesosContainer }
+import mesosphere.marathon.raml.{ Endpoint, Image, ImageType, Resources }
+import mesosphere.marathon.state.Container
+import mesosphere.marathon.state.PathId._
+import mesosphere.marathon.state._
+import mesosphere.marathon.core.pod.PodDefinition
+import org.openjdk.jmh.annotations.{ Group => _, _ }
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.collection.breakOut
+
+@State(Scope.Benchmark)
+class GroupBenchmark {
+
+  val version = VersionInfo.forNewConfig(Timestamp(1))
+
+  def makeApp(path: PathId) =
+    AppDefinition(
+      id = path,
+      labels = Map("ID" -> path.toString),
+      versionInfo = version,
+      networks = Seq(BridgeNetwork()),
+      container = Some(
+        Container.Docker(Nil, "alpine", List(Container.PortMapping(2015, Some(0), 10000, "tcp", Some("thing")))))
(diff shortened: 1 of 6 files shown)


src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -111,7 +111,10 @@ class GroupManagerImpl(
         logger.info(s"Upgrade root group version:$version with force:$force")
 
         val from = rootGroup()
-        val unversioned = assignDynamicServicePorts(from, change(from))
+        val unversioned = AssignDynamicServiceLogic.assignDynamicServicePorts(
+          Range.inclusive(config.localPortMin(), config.localPortMax()),
+          from,
+          change(from))
```
jeschkies: is this correct?

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
```diff
@@ -0,0 +1,104 @@
...
+}
+
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.AverageTime))
+@Fork(1)
+class GroupManagerBenchmark extends GroupBenchmark {
+
+  @Benchmark
+  def updateVersionInfoForChangedApps(hole: Blackhole): Unit = {
+    println(rootGroup.groupsById.keys)
```
alenkacz: is the println kept here intentional?

src/main/scala/mesosphere/marathon/core/group/impl/AssignDynamicServiceLogic.scala
```diff
@@ -0,0 +1,108 @@
...
+      }
+    val dynamicApps: Iterator[AppDefinition] =
+      to.transitiveApps
+        .iterator
+        .filter { newApp => changedOrNew(from.transitiveAppsById, newApp) }
+        .map {
+          // assign values for service ports that the user has left "blank" (set to zero)
+          case app: AppDefinition if app.hasDynamicServicePorts =>
+            assignPorts(app, from.app(app.id), portRange, unassignedPortsIterator)
+          case app: AppDefinition =>
```
alenkacz: this is hurting readability a bit because it looks like mergeServicePortsAndPortDefinitions is only done for app that does not have dynamic service ports - but that is not true, because it is also hidden in the assignPorts before returning the value...

But I guess doing it like this we save ourselves one more extra copy so we will probably have to live with this..

benchmark/src/main/scala/mesosphere/marathon/upgrade/GroupManagerBenchmark.scala
jeschkies: I removed it. Thanks.

PR #5798, opened 2017-12-06
fix IS constraint for pods

timcharper:
Pods use a different serialization mechanism for constraints. When implementing the IS operator, this conversion method wasn't updated.

Add additional tests to improve coverage

JIRA Issues: MARATHON_EE-1667


timcharper:
Doh! Good catch. Code is correct, description not. I'll fix the latter.

PR #5800, opened 2017-12-06
Upgrade Akka, Scala, and other libs

timcharper:
Upgrade Akka, Scala, and other libs

JIRA Issues: MARATHON-7917

Please merge #5799 and #5801 first

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/election/CuratorElectionStreamTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/election/CuratorElectionStreamTest.scala
@@ -58,9 +58,8 @@ class CuratorElectionStreamTest extends AkkaUnitTest with Inside with ZookeeperS
         override lazy val zkTimeoutDuration = 250.milliseconds
       }
 
-      inside(Try(CuratorElectionStream.newCuratorConnection(conf))) {
-        case Failure(ex) =>
-          ex shouldBe a[UnknownHostException]
+      a[Throwable] shouldBe thrownBy {
+        CuratorElectionStream.newCuratorConnection(conf)
       }
     }
   }
--- a/build.sbt
+++ b/build.sbt
@@ -116,7 +116,7 @@ lazy val commonSettings = testSettings ++
   coverageFailOnMinimum := true,
 
   fork in run := true,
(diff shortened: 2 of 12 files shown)


timcharper:
Strange, these two tests have failed twice in a row, but when run individually, locally, they pass:

Test Name | Duration | Age
-- | -- | --
mesosphere.marathon.integration.RestartIntegrationTest.Restarting Marathon when readiness should deployment with 1 ready and 1 not ready instance is continued properly after a restart | 2 min 5 sec | 2
mesosphere.marathon.integration.RestartIntegrationTest.Restarting Marathon when health checks should deployment with 2 unhealthy instances is continued properly after master abdication | 1 min 40 sec | 2

src/test/scala/mesosphere/marathon/core/election/CuratorElectionStreamTest.scala
```diff
@@ -58,9 +58,8 @@ class CuratorElectionStreamTest extends AkkaUnitTest with Inside with ZookeeperS
         override lazy val zkTimeoutDuration = 250.milliseconds
       }
 
-      inside(Try(CuratorElectionStream.newCuratorConnection(conf))) {
-        case Failure(ex) =>
-          ex shouldBe a[UnknownHostException]
+      a[Throwable] shouldBe thrownBy {
```
jeschkies: I think it should be `an[UnknownHostException] shouldBe thrownBy`.

timcharper: No, the logic changed. The Zookeeper client logs the UnknownHostException, but it just swallows it. It's kind of frustrating. Not really anything we can do.

timcharper: https://github.com/apache/zookeeper/blob/release-3.4.11/src/java/main/org/apache/zookeeper/client/StaticHostProvider.java#L82

timcharper: The end result is an illegal argument exception is thrown because it thinks nothing was passed.

I suppose we could open up an issue.

PR #5804, opened 2017-12-07
 Compute transitive apps by id index lazily. 

jeschkies:
Summary:
This is the first step to remove the app and pods index alltorgether.
For now it is caluclated only at one spot.

JIRA issues: MARATHON-7945

PR #5805, opened 2017-12-07
 Remove app and pod transitive index from groups. 

jeschkies:
Summary:
This is another speed up for groups. Currently we build an index over
transitive apps and pods which is costly. This index is not required
since we have an index over all transitive groups and each group has an
index of its apps and pods.

Depends on #5804.

JIRA issues: MARATHON-7945

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/state/RootGroupBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/state/RootGroupBenchmark.scala
@@ -1,0 +1,76 @@
+package mesosphere.marathon
+package state
+
+import java.util.concurrent.TimeUnit
+
+import mesosphere.marathon.core.pod.BridgeNetwork
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.util.Random
+
+@State(Scope.Benchmark)
+class GroupBenchmark {
+
+  val version = VersionInfo.forNewConfig(Timestamp(1))
+
+  def makeApp(path: PathId) =
+    AppDefinition(
+      id = path,
+      labels = Map("ID" -> path.toString),
+      versionInfo = version,
+      networks = Seq(BridgeNetwork()),
+      container = Some(
+        Container.Docker(Nil, "alpine", List(Container.PortMapping(2015, Some(0), 10000, "tcp", Some("thing")))))
+    )
+
+  //  @Param(value = Array("100", "500", "1000", "2500", "5000"))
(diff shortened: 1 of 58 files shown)


src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
```diff
@@ -187,13 +187,13 @@ class MarathonSchedulerActorTest extends AkkaUnitTest with ImplicitSender with G
 
     "ScaleApp" in withFixture() { f =>
       import f._
-      val app = AppDefinition(id = "test-app".toPath, instances = 1, cmd = Some("sleep"))
+      val app = AppDefinition(id = "test-app-scale".toPath, instances = 1, cmd = Some("sleep"))
```
jeschkies: I'm not sure this is a valid path. It should start with `/` otherwise the parent is null.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -20,31 +20,42 @@ class Group(
     val dependencies: Set[PathId] = defaultDependencies,
     val version: Timestamp = defaultVersion) extends IGroup {
 
-  lazy val transitiveAppsById: Map[AppDefinition.AppKey, AppDefinition] = apps ++ groupsById.values.flatMap(_.transitiveAppsById)
-  lazy val transitivePodsById: Map[PathId, PodDefinition] = pods ++ groupsById.values.flatMap(_.transitivePodsById)
+  def app(appId: PathId): Option[AppDefinition] = {
+    apps.get(appId) orElse group(appId.parent).flatMap(_.apps.get(appId))
```
jeschkies: We check the apps first in case the parent is null. We also avoid building the transitive group index.

benchmark/src/main/scala/mesosphere/marathon/state/RootGroupBenchmark.scala
```diff
@@ -0,0 +1,82 @@
...
+      container = Some(
+        Container.Docker(Nil, "alpine", List(Container.PortMapping(2015, Some(0), 10000, "tcp", Some("thing")))))
+    )
+
+  //  @Param(value = Array("100", "500", "1000", "2500", "5000"))
+  @Param(value = Array("2", "10", "100"))
+  var appsPerGroup: Int = _
+  lazy val appIds = 0 until appsPerGroup
+
+  //  @Param(value = Array("5", "15", "100", "1000"))
```
jeschkies: We cannot event build groups with a depth of 1000 :/.

PR #5806, opened 2017-12-07
Collect diagnostics on all exit codes greater than 0

alenkacz:
Summary:
Currently when the code was e.g. 3 then diagnostics bundle was not collected.


alenkacz:
close your eyes, @zen-dog I am breaking the 24h rule 🙈

PR #5807, opened 2017-12-07
 Replace transitive apps and pods set with an iterable. 

jeschkies:
Summary:
This change builds on top of #5805 and cuts the execution time
of the assign dynamic service ports.

The current master takes ~18ms for 15 groups and 10,000 apps 
in the dynamic port assignment benchmark. This change speeds
it up to <5ms.

I also added a benchmark for building the root group which implies
a lot of `updateApp` calls. Here, the current master takes >97000ms
for building a root group with 20 nested child groups and 1000 apps
each, i.e. building a group with 20,000 apps. This patch speeds up
the building time by two magnitudes to <230ms.

The app, pod and child group access time stays constant while the
master as a linear time in terms of number of groups.

The speed is achieved by removing the build up of an apps and pods
index (`transitiveAppsById` and `transitivePodsById`). The look up
stays constant by leveraging the constant lookup of child groups.

JIRA issues: MARATHON-7946

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/group/impl/AssignDynamicServiceLogic.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/AssignDynamicServiceLogic.scala
@@ -79,16 +79,15 @@ object AssignDynamicServiceLogic extends StrictLogging {
      * app in the case that servicePorts are re-posted all as 0's.
      */
     val usedServicePorts: Set[Int] =
-      (from.transitiveApps.iterator ++ to.transitiveApps.iterator).flatMap(_.servicePorts).toSet
+      (from.transitiveApps ++ to.transitiveApps).flatMap(_.servicePorts).toSet
     val unassignedPortsIterator = portRange.iterator
       .filter { p => !usedServicePorts.contains(p) }
       .map { port =>
         logger.debug(s"Take next configured free port: $port")
         port
       }
-    val dynamicApps: Iterator[AppDefinition] =
+    val dynamicApps: Iterable[AppDefinition] =
       to.transitiveApps
-        .iterator
         .filter { newApp => changedOrNew(from, newApp) }
         .map {
           // assign values for service ports that the user has left "blank" (set to zero)
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -46,13 +46,13 @@ class Group(
 
   private def transitiveAppsIterator(): Iterator[AppDefinition] = apps.valuesIterator ++ groupsById.valuesIterator.flatMap(_.transitiveAppsIterator())
   private def transitiveAppIdsIterator(): Iterator[PathId] = apps.keysIterator ++ groupsById.valuesIterator.flatMap(_.transitiveAppIdsIterator())
-  lazy val transitiveApps: Iterable[AppDefinition] = transitiveAppsIterator().toIterable
-  lazy val transitiveAppIds: Iterable[PathId] = transitiveAppIdsIterator().toIterable
+  lazy val transitiveApps: Iterable[AppDefinition] = transitiveAppsIterator().toVector
+  lazy val transitiveAppIds: Iterable[PathId] = transitiveAppIdsIterator().toVector
 
   private def transitivePodsIterator(): Iterator[PodDefinition] = pods.valuesIterator ++ groupsById.valuesIterator.flatMap(_.transitivePodsIterator())
   private def transitivePodIdsIterator(): Iterator[PathId] = pods.keysIterator ++ groupsById.valuesIterator.flatMap(_.transitivePodIdsIterator())
-  lazy val transitivePods: Iterable[PodDefinition] = transitivePodsIterator().toIterable
-  lazy val transitivePodIds: Iterable[PathId] = transitivePodIdsIterator().toIterable
+  lazy val transitivePods: Iterable[PodDefinition] = transitivePodsIterator().toVector
+  lazy val transitivePodIds: Iterable[PathId] = transitivePodIdsIterator().toVector
 
   lazy val transitiveRunSpecs: Iterable[RunSpec] = transitiveApps ++ transitivePods
   lazy val transitiveRunSpecIds: Iterable[PathId] = transitiveAppIds ++ transitivePodIds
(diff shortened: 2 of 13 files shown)


src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -20,42 +20,65 @@ class Group(
...
+    * @return None if no group was found or non empty option with group.
     */
   def group(gid: PathId): Option[Group] = transitiveGroupsById.get(gid)
 
   private def transitiveAppsIterator(): Iterator[AppDefinition] = apps.valuesIterator ++ groupsById.valuesIterator.flatMap(_.transitiveAppsIterator())
   private def transitiveAppIdsIterator(): Iterator[PathId] = apps.keysIterator ++ groupsById.valuesIterator.flatMap(_.transitiveAppIdsIterator())
-  lazy val transitiveApps: Set[AppDefinition] = transitiveAppsIterator().toSet
-  lazy val transitiveAppIds: Set[PathId] = transitiveAppIdsIterator().toSet
+  lazy val transitiveApps: Iterable[AppDefinition] = transitiveAppsIterator().toVector
+  lazy val transitiveAppIds: Iterable[PathId] = transitiveAppIdsIterator().toVector
```
jeschkies: Building the vector is crucial here. Iterating over vectors is much faster than iterating over maps. as these [benchmarks on Scala's collections](http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#lists-vs-vectors) show that @meln1k sent around. In cases where we have to iterate it is better to make a copy.

src/test/scala/mesosphere/marathon/state/RootGroupTest.scala
```diff
@@ -31,6 +31,18 @@ class RootGroupTest extends UnitTest with GroupCreation {
       current.app(path) should be('defined)
     }
 
+    "find an app without a parent" in {
+      Given("an existing root group with an app without a parent")
+      val app = AppDefinition("app".toPath, cmd = Some("sleep"))
```
jeschkies: This came out of a different test case. I think apps without a parent should be considered invalid.

src/main/scala/mesosphere/marathon/core/group/impl/AssignDynamicServiceLogic.scala
```diff
@@ -79,16 +79,15 @@ object AssignDynamicServiceLogic extends StrictLogging {
      * app in the case that servicePorts are re-posted all as 0's.
      */
     val usedServicePorts: Set[Int] =
-      (from.transitiveApps.iterator ++ to.transitiveApps.iterator).flatMap(_.servicePorts).toSet
+      (from.transitiveApps ++ to.transitiveApps).flatMap(_.servicePorts).toSet
```
jeschkies: We do not have to use an iterator here since concat of two vectors is fast enough.

timcharper: 👏

jeschkies:
## Master
## This Patch
| Benchmark                                       | (numberOfGroups) |(numberOfSavedApps) | Mode | Cnt| Score| Error |  Units
|---|---|---|---|---|---|-----|---|
| assignDynamicServicePorts | 5 |  100 | avgt | 10 | 0.036| ±0.002|  ms/op
| assignDynamicServicePorts | 5 |  500 | avgt | 10 | 0.141| ±0.006|  ms/op
| assignDynamicServicePorts | 5 | 1000 | avgt | 10 | 0.271| ±0.018|  ms/op
| assignDynamicServicePorts | 5 | 2500 | avgt | 10 | 0.647| ±0.036|  ms/op
| assignDynamicServicePorts | 5 | 5000 | avgt | 10 | 1.515| ±0.406|  ms/op
| assignDynamicServicePorts | 5 |10000 | avgt | 10 | 3.585| ±0.148|  ms/op
| assignDynamicServicePorts |15 |  100 | avgt | 10 | 0.036| ±0.001|  ms/op
| assignDynamicServicePorts |15 |  500 | avgt | 10 | 0.135| ±0.003|  ms/op
| assignDynamicServicePorts |15 | 1000 | avgt | 10 | 0.262| ±0.006|  ms/op
| assignDynamicServicePorts |15 | 2500 | avgt | 10 | 0.591| ±0.046|  ms/op
| assignDynamicServicePorts |15 | 5000 | avgt | 10 | 1.591| ±0.222|  ms/op
| assignDynamicServicePorts |15 |10000 | avgt | 10 | 3.304| ±0.175|  ms/op

jeschkies:
# Benchmark `buildRootGroup`
`sbt "benchmark/jmh:run -i 10 -wi 3 -f1 -t1 .*RootGroupBenchmark.buildRootGroup"`
## Master

## This Patch
|(appsPerGroup)| (groupDepth) | Mode | Cnt | Score | Error | Units
|---|---|---|---|---|-----|---|
|   2|  5| avgt | 10|  0.054| ± 0.005| ms/op
|   2| 10| avgt | 10|  0.172| ± 0.021| ms/op
|   2| 20| avgt | 10|  0.461| ± 0.013| ms/op
|  10|  5| avgt | 10|  0.264| ± 0.031| ms/op
|  10| 10| avgt | 10|  0.736| ± 0.085| ms/op
|  10| 20| avgt | 10|  1.949| ± 0.055| ms/op
| 100|  5| avgt | 10|  2.318| ± 0.214| ms/op
| 100| 10| avgt | 10|  6.948| ± 1.484| ms/op
| 100| 20| avgt | 10| 20.625| ± 2.231| ms/op
|1000|  5| avgt | 10| 27.452| ± 1.402| ms/op
|1000| 10| avgt | 10| 76.132| ± 4.700| ms/op
|1000| 20| avgt | 10|320.854| ±20.372| ms/op

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -20,42 +20,65 @@ class Group(
...
 
   private def transitivePodsIterator(): Iterator[PodDefinition] = pods.valuesIterator ++ groupsById.valuesIterator.flatMap(_.transitivePodsIterator())
   private def transitivePodIdsIterator(): Iterator[PathId] = pods.keysIterator ++ groupsById.valuesIterator.flatMap(_.transitivePodIdsIterator())
-  lazy val transitivePods: Set[PodDefinition] = transitivePodsIterator().toSet
-  lazy val transitivePodIds: Set[PathId] = transitivePodIdsIterator().toSet
+  lazy val transitivePods: Iterable[PodDefinition] = transitivePodsIterator().toVector
```
meln1k: Why we want to use `Iterable` here when we already know that the type is `Vector`?

jeschkies: That is a good question. The transition to `Iterable` was easier. `Vector` seemed a little more opinionated to me. I'd say we go for vector all the way.

jeschkies: But then again, I don't think these are bottlenecks anymore. I really would love to start measuring execution times with Kamon or Metrics.

jeschkies:
@zen-dog could you check the dependent PRs #5804 and #5805 as well. Thanks 🙂.

PR #5808, opened 2017-12-07
Pass killGracePeriodSeconds value to Mesos for pods

timcharper:
JIRA Issues: MARATHON-7952

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -16,8 +16,8 @@ import mesosphere.mesos.protos.Implicits._
 import org.apache.mesos.Protos.{ DurationInfo, KillPolicy }
 import org.apache.mesos.{ Protos => mesos }
 
+import java.util.concurrent.TimeUnit.SECONDS
 import scala.collection.immutable.Seq
-import scala.concurrent.duration._
 
 object TaskGroupBuilder extends StrictLogging {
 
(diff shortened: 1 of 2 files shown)


kensipe:
nice work in short time frame

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -183,6 +185,15 @@ object TaskGroupBuilder extends StrictLogging {
       computeHealthCheck(healthCheck, portAssignments).foreach(builder.setHealthCheck)
     }
 
+    for {
+      lc <- container.lifecycle
+      killGracePeriodSeconds <- lc.killGracePeriodSeconds
+    } {
+      val durationInfo = DurationInfo.newBuilder.setNanoseconds((killGracePeriodSeconds * 1000000000).toLong)
```
jeschkies: Nit: I like to have Java do the transformation `Duration.ofSeconds(killGracePeriodSeconds).getNano()`. This avoids magic numbers.

src/test/scala/mesosphere/mesos/TaskGroupBuilderTest.scala
```diff
@@ -1055,5 +1056,29 @@ class TaskGroupBuilderTest extends UnitTest with Inside {
...
+      val (_, taskGroupInfo, _, _) = TaskGroupBuilder.build(
+        podSpec,
+        offer,
+        s => Instance.Id.forRunSpec(s),
+        defaultBuilderConfig,
+        RunSpecTaskProcessor.empty,
+        resourceMatch.asInstanceOf[ResourceMatchResponse.Match].resourceMatch
+      )
+
+      taskGroupInfo.getTasks(0).getKillPolicy.getGracePeriod.getNanoseconds shouldBe (killDuration.toNanos)
```
jeschkies: Here you use `toNanos` 😉

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
timcharper: Yeah, I thought about this and anticipated somebody might bring this up. I totally prefer to use durations. In this case, the input duration was a double and already unit-less. Didn't seem worth the extra allocation. And the conversion was covered by a test.

kensipe:
@timcharper while not the norm... I would support a merge.    @jeschkies rejection was addressed and I wouldn't want to wait all weekend to get another review.     There isn't a "rush" except for the customer that needs this... needs a back port.

PR #5809, opened 2017-12-08
AppId parent lookup is inefficient

meln1k:
Summary: caching of the parent path id for the sake of the performance

JIRA issues: MARATHON-7947

Before:
```
[info] Benchmark                         Mode  Cnt    Score     Error   Units
[info] PathIdBenchmark.allParentsSpeed  thrpt   10    0.004 ?   0.001  ops/us
[info] PathIdBenchmark.allParentsSpeed   avgt   10  274.137 ? 105.434   us/op
```

After:
```
[info] Benchmark                         Mode  Cnt  Score   Error   Units
[info] PathIdBenchmark.allParentsSpeed  thrpt   10  1.795 ? 0.083  ops/us
[info] PathIdBenchmark.allParentsSpeed   avgt   10  0.563 ? 0.015   us/op
```


benchmark/src/main/scala/mesosphere/marathon/upgrade/PathIdBenchmark.scala
```diff
@@ -0,0 +1,32 @@
...
+import mesosphere.marathon.state.PathId
+import org.openjdk.jmh.annotations.{ Group => _, _ }
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.util.Random
+
+@State(Scope.Benchmark)
+object PathIdBenchmark {
+
+  lazy val pathIdSegments = (1 to 100)
```
jeschkies: Could you parameterize this? E.g. 

```scala
@Param(value = Array("2", "10", "100", "1000"))
   var numberOfPaths: Int = _
   lazy val pathIdSegments = (1 to numberOfPaths)
```

benchmark/src/main/scala/mesosphere/marathon/upgrade/PathIdBenchmark.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  lazy val pathId = PathId(pathIdSegments.map(_ => Random.alphanumeric.filter(_.isLetter).take(10).mkString).toList)
+
+}
+
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@BenchmarkMode(Array(Mode.Throughput, Mode.AverageTime))
```
jeschkies: I think time is enough 🙂.

PR #5810, opened 2017-12-08
specify region to prevent sbt error

timcharper:
JIRA Issues: MARATHON-7917

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -108,6 +108,7 @@ lazy val commonSettings = testSettings ++
     s3("downloads.mesosphere.io/maven")
   )),
   s3credentials := new EnvironmentVariableCredentialsProvider() | InstanceProfileCredentialsProvider.getInstance(),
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -108,6 +108,7 @@ lazy val commonSettings = testSettings ++
     s3("downloads.mesosphere.io/maven")
   )),
   s3credentials := new EnvironmentVariableCredentialsProvider() | InstanceProfileCredentialsProvider.getInstance(),
+  s3region := com.amazonaws.services.s3.model.Region.US_West_2,
```
meln1k: Is it the right region?

meln1k: It used to be a different one:

https://github.com/ohnosequences/sbt-s3-resolver/blob/v0.8.0/src/main/scala/SBTS3Resolver.scala#L51

kensipe:
#5811 seems to be the same change... oddly these 2 PRs reference different JIRAs.

timcharper:
Dude, I give up.

PR #5811, opened 2017-12-08
Fix of the broken region setting for s3 resolution

meln1k:
Summary: added missing setting to the build file

JIRA issues: MARATHON-7955


timcharper:
@kensipe there is no US_East_1, apparently.

PR #5812, opened 2017-12-08
Improve the exception thrown for incorrect timestamp string

alenkacz:
Summary:
With the joda time to java time switch we changed the exception message published through API. And apparently some tests rely on that message.
By changing that message, we in a way did a breaking change in our API but let's hope no more people rely on that exact exception message.
By this patch, I tried to make the exception thrown more descriptive and detached from the actual message that our time library is printing out.

JIRA issues: MARATHON-7954

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/state/RootGroupBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/state/RootGroupBenchmark.scala
@@ -1,0 +1,80 @@
+package mesosphere.marathon
+package state
+
+import java.util.concurrent.TimeUnit
+
+import mesosphere.marathon.core.pod.BridgeNetwork
+import org.openjdk.jmh.annotations._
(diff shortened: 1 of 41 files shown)


ichernetsky:
It seems `PodsControllerTest` should be updated. Otherwise looks good. Thank you.

```
*** 1 TEST FAILED ***
PodsControllerTest:
PodsController should 
- reject GET of /mypod::versions/unparsable-datetime for entity not found: Pod 'mypod' does not exist in version unparsable-datetime *** FAILED *** (18 milliseconds)
  Request was not rejected, response was HttpResponse(500 Internal Server Error,List(),HttpEntity.Strict(text/plain; charset=UTF-8,There was an internal server error.),HttpProtocol(HTTP/1.1)) (DynamicVariable.scala:58)
```

alenkacz:
I've fixed the test. @jeschkies can you check it out, since you are author of that pods controller? Thanks

PR #5813, opened 2017-12-08
An incomplete implementation of pod persistent volumes and app/pod persistent volume profiles

ichernetsky:
## Things that are done so far

- Volume and Residency RAML types
- Conversion to and from RAML, JSON, and Protocol Buffers
- App and pod volume refactoring
- Some changes to task and task group builders
- Reserving and creating volumes
- Launching pods on reservation
- Supporting pod restarts
- Persistent volume profiles
- System integration testing

## Things to do yet

- Move reservation info from the task level to the instance level
- Once reservation info is moved, probably there won't be any need in task polymorphism (Reserved, Ephemeral etc)
- Documentation
- Unit- and integration testing
- More system integration testing
- Instance update care only about last task status update, which doesn't really work well with pods with persistent volumes, because if one of pod tasks exits normally, this is considered worthy of checking and scaling the pod
- Migration
- Remove `LegacyAppInstance`

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON_EE-1735
https://jira.mesosphere.com/browse/MARATHON_EE-1677
https://jira.mesosphere.com/browse/MARATHON_EE-1683


alenkacz:
Is there any chance we can split this PR into smaller ones or is it unrealistic? @ichernetsky

ichernetsky:
@alenkacz, unfortunately it is impossible due to the deadline. I would be possible otherwise, but it took too much time to get it working.

ichernetsky:
The hardest part was to identify bits to be changed and to wire them properly. Speaking of the actual diff, I would say, about 3/4 of it is about renaming and moving bits around. So you should not be intimidated by its size, I guess. Only one fourth (at most) does require your attention.

build.sbt
```diff
@@ -17,6 +17,9 @@ def formattingTestArg(target: File) = Tests.Argument("-u", target.getAbsolutePat
 credentials ++= loadM2Credentials(streams.value.log)
 resolvers ++= loadM2Resolvers(sLog.value)
 
+lazy val myRepo = "My Repo" at "https://s3.us-east-2.amazonaws.com/ichernetsky/maven/"
```
alenkacz: why do we need this? was this commited by mistake?

alenkacz: is it because of mesos JVM lib? Can we upload it to the mesosphere maven repo as we did for the special ipv6 build... ?

src/main/scala/mesosphere/marathon/raml/VolumeConversion.scala
```diff
@@ -33,106 +45,140 @@ trait VolumeConversion extends ConstraintConversion with DefaultConversions {
...
-    val info = ExternalVolumeInfo(
-      size = vol.external.size,
-      name = vol.external.name.getOrElse(throw SerializationFailedException("external volume requires a name")),
-      provider = vol.external.provider.getOrElse(throw SerializationFailedException("external volume requires a provider")),
-      options = vol.external.options
+  implicit val volumeExternalReads: Reads[AppExternalVolume, state.VolumeWithMount] = Reads { volumeRaml =>
+    val info = state.ExternalVolumeInfo(
+      size = volumeRaml.external.size,
+      name = volumeRaml.external.name.getOrElse(
+        throw SerializationFailedException("external volume requires a name")),
```
alenkacz: shouldn't this be handled through validation layer?

also the indentation seems weird (meaning a bit different) to me...

build.sbt
ichernetsky: Yes, it is. Once it is available from the Mesos repo, I will remove this.

src/main/scala/mesosphere/marathon/raml/VolumeConversion.scala
ichernetsky: It is. But since `scapegoat` complains about `.get` calls. Also, it is better to double-check than to be sorry :)

The indentation looks fine to me. What do you dislike about it?

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -80,11 +80,15 @@ object ContainerSerializer {
         builder.setMesos(MesosAppCSerializer.toMesos(ma))
     }
 
-    container.volumes.foreach {
-      case _: PersistentVolume => // PersistentVolumes are handled differently
-      case ev: ExternalVolume => ExternalVolumes.build(builder, ev) // this also adds the volume
-      case dv: DockerVolume => builder.addVolumes(VolumeSerializer.toMesos(dv))
-      case _: SecretVolume => // SecretVolumes are handled differently
+    container.volumes.foreach { volumeWithMount =>
```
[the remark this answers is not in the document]
ichernetsky: Done. Thanks for suggesting it.

ichernetsky:
Addressed a comment and rebased.

docs/docs/rest-api/public/api/v2/types/residency.raml
```diff
@@ -0,0 +1,22 @@
...
+uses:
+  strings: stringTypes.raml
+types:
+  Residency:
+    type: object
+    description: |
+      When using local persistent volumes that pin tasks onto agents,
+      these values define how Marathon handles terminal states of these tasks.
+    properties:
+      relaunchEscalationTimeoutSeconds:
```
timcharper: I'm about 95% confident this field currently doesn't do anything. (for apps). And, it seems like it directly overlaps in functionality with `UnreachableStrategy.inactiveAfterSeconds` (which does work, however, is currently forced to be disabled, much to the dismay of @unterstein ).

As far as I can tell, it didn't do anything in as far back as Marathon 1.1.

I'd be inclined to just remove the field since it's been broken that long, and overlaps in functionality with UnreachableStrategy.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -120,48 +123,55 @@ object ContainerSerializer {
...
-        .build()
+  def toProto(volumeWithMount: VolumeWithMount): Protos.Volume = {
+    val volume = volumeWithMount.volume
+    val mount = volumeWithMount.mount
+    val mode = if (mount.readOnly) mesos.Protos.Volume.Mode.RO else mesos.Protos.Volume.Mode.RW
+
+    volume match {
+      case p: PersistentVolume =>
+        Protos.Volume.newBuilder()
+          .setContainerPath(mount.mountPath)
```
timcharper: I think I'd prefer to see the common fields extracted above the match statement.

IE:

```scala
  def toProto(volumeWithMount: VolumeWithMount): Protos.Volume = {
    val volume = volumeWithMount.volume
    val mount = volumeWithMount.mount
    val mode = if (mount.readOnly) mesos.Protos.Volume.Mode.RO else mesos.Protos.Volume.Mode.RW

    val volumeBuilder = Protos.Volume.newBuilder()
      .setContainerPath(mount.mountPath)
      .setMode(mode)

    volume match {
      case p: PersistentVolume =>
        volumeBuilder
          .setPersistent(PersistentVolumeInfoSerializer.toProto(p.persistent))

      case e: ExternalVolume =>
        Protos.Volume.newBuilder()
          .setExternal(ExternalVolumeInfoSerializer.toProto(e.external))

      case d: HostVolume =>
        Protos.Volume.newBuilder()
          .setHostPath(d.hostPath)

      case s: SecretVolume =>
        Protos.Volume.newBuilder()
          .setSecret(Protos.Volume.SecretVolumeInfo.newBuilder().setSecret(s.secret))
    }

    volumeBuilder.build()
  }

```

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
```diff
@@ -52,11 +52,11 @@ object InstanceUpdateOperation {
     */
   case class LaunchOnReservation(
       instanceId: Instance.Id,
-      newTaskId: Task.Id,
+      newTaskIds: Map[Task.Id, Task.Id], // mapping of old task IDs to new ones
```
timcharper: I had the impression this was a map of instanceId to taskId. Should it be? Having a taskId -> taskId mapping seems odd.

src/main/scala/mesosphere/marathon/raml/VolumeConversion.scala
timcharper: indentation should just be handled by our formatter :)

I think I'd like to make it clear that this error message should never be seen. Maybe throw an IllegalStateException?

src/main/scala/mesosphere/marathon/core/externalvolume/ExternalVolumes.scala
```diff
@@ -14,14 +14,15 @@ import org.apache.mesos.Protos.ContainerInfo
 object ExternalVolumes {
   private[this] lazy val providers: ExternalVolumeProviderRegistry = StaticExternalVolumeProviderRegistry
```
timcharper: I want to avoid scope creep here, so feel free to ignore / address in a separate commit.

Does this trait / object look like a pretty classic case of speculative generality to you? The more I look at it, the more confused I am. It seems like `ExternalVolumeProviderRegistry`, and `StaticExternalVolumeProviderRegistry` could just be removed, and turned in to, at worst, a map. And, there's only one volume provide. So it could just be a single method with an assertion.

docs/docs/rest-api/public/api/v2/types/residency.raml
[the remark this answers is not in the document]
ichernetsky: It is not fully leveraged as of now, but it doesn't overlap in functionality with anything of `UnreachableStrategy`. Currently, it is enforced that if an app/pod has a PV, `taskLostBehavior` has to be set to `WAIT_FOREVER`. Both fields make sense, but the implementation is not complete, here I agree.

docs/docs/rest-api/public/api/v2/types/residency.raml
```diff
@@ -0,0 +1,22 @@
...
+        type: integer
+        format: int64
+        minimum: 0
+        default: 3600
+        description: |
+          When a task using persistent local volumes cannot be restarted on the
+          agent it's been pinned to, Marathon will try to launch this task on
+          another node after this timeout. Defaults to 3600 (one hour).",
+      taskLostBehavior:
```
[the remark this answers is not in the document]
ichernetsky: Please refer to my comment to your message above.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -81,10 +81,13 @@ object ContainerSerializer {
...
 
     container.volumes.foreach {
-      case _: PersistentVolume => // PersistentVolumes are handled differently
-      case ev: ExternalVolume => ExternalVolumes.build(builder, ev) // this also adds the volume
-      case dv: DockerVolume => builder.addVolumes(VolumeSerializer.toMesos(dv))
-      case _: SecretVolume => // SecretVolumes are handled differently
+      case VolumeWithMount(volume, mount) =>
+        volume match {
+          case _: PersistentVolume => // PersistentVolumes are handled differently
+          case ev: ExternalVolume => ExternalVolumes.build(builder, ev, mount) // this also adds the volume
```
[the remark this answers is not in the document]
ichernetsky: Yes, `ExternalVolumeProvider.build` expects one. Because providers can add volumes. This is what `DVDIProvider` does.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
ichernetsky: Will do.

src/main/scala/mesosphere/marathon/core/externalvolume/ExternalVolumes.scala
ichernetsky: I think this was done having some idea about possible extensions in the future. Added to the description of https://jira.mesosphere.com/browse/MARATHON_EE-1735

src/main/scala/mesosphere/marathon/raml/VolumeConversion.scala
ichernetsky:
```
/**
  * Thrown for errors during [[Normalization]]. Validation should normally be checking for invalid data structures
  * that would lead to errors during normalization, so these exceptions are very unexpected.
  * @param msg provides details
  */
case class NormalizationException(msg: String) extends Exception(msg)

case class SerializationFailedException(message: String) extends Exception(message)
```

These exceptions are said to be very unexpected, whatever it means :)

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
ichernetsky: It is needed to resident tasks in order to have somewhat stable task IDs, stable here means that only the UUID part changes upon relaunch. It am not crazy about it too, probably it could be done in a better way, but it would mean yet another refactoring which I cannot afford right now. Should I add this to the to-do list?

alenkacz: maybe the field then should be named oldToNewTaskIds? (then we can lose the comment too)

Or we can explain in the comment why is the mapping needed in the first place. What do you think?

docs/docs/rest-api/public/api/v2/types/volumes.raml
```diff
@@ -7,42 +7,29 @@ uses:
...
     properties:
-      name:
-        type: strings.Name
-        description: The name of the volume to reference.
-      host:
+      size?:
+        type: integer
+        format: int64
+        minimum: 0
+        description: The size of the external volume in GiB
```
meln1k: In my view GiB looks a bit coarse. What do you think about MiBs?

ichernetsky: Why? I think it is what it is. Changing it would mean a breaking change.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
ichernetsky: Makes sense. Will do.

ichernetsky:
Comrades, thanks for your reviews. Needless to say, they are greatly appreciated!

docs/docs/rest-api/public/api/v2/types/volumes.raml
```diff
@@ -142,5 +143,20 @@ types:
...
+        type: strings.Path
+        description: The path inside the container at which the volume is mounted.
+      readOnly?:
+        type: boolean
+
+  AppVolume:
+    type: AppPersistentVolume | AppExternalVolume | AppSecretVolume | AppHostVolume
   PodVolume:
-    type: PodSecretVolume | HostVolume | EphemeralVolume
+    type: PodPersistentVolume | PodSecretVolume | PodHostVolume | PodEphemeralVolume
```
zen-dog: Nice cleanup 👏

plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
```diff
@@ -114,53 +119,18 @@ trait PodSpec extends RunSpec {
...
+  val name: Option[String]
 }
 
-/**
-  * Application volume with a secret
-  */
-trait AppSecretVolumeSpec extends AppVolumeSpec {
-
+trait SecretVolumeSpec extends VolumeSpec {
   val secret: String
```
zen-dog: Nit: add field comments here and below (I know they're self-explanatory but we do anyway since it's plugin-interface)

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
```diff
@@ -68,18 +67,19 @@ private[impl] case object DVDIProvider extends ExternalVolumeProvider {
       // containerizer. the docker containerizer simply ignores them.
       applyOptions(volBuilder, dockerVolumeParameters(volume))
 
+      val mode = if (mount.readOnly) MesosVolume.Mode.RO else MesosVolume.Mode.RW
```
zen-dog: This conversion happens a few times in code. May be make it a `def` in `object VolumeMount`?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
```diff
@@ -39,10 +39,13 @@ object InstanceUpdateOperation {
 
   /**
     * @param instanceId Designating the instance that shall be launched.
-    * @param newTaskId The id of the task that will be launched via Mesos
+    * @param oldToNewTaskIds Mapping from old task IDs to new ones. It is required in order to maintain stable task IDs
+    *                        for resident tasks (currently, the ones, which have persistent volumes), which are launched
+    *                        upon a resource reservation. Stable here means that only the UUID and attempt part change,
```
zen-dog: Only attemps changes, not the UUID part.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -84,28 +84,37 @@ object InstanceUpdater extends StrictLogging {
 
   private[marathon] def launchOnReservation(instance: Instance, op: LaunchOnReservation): InstanceUpdateEffect = {
     if (instance.isReserved) {
-      require(instance.tasksMap.size == 1, "Residency is not yet implemented for task groups")
+      val currentTasks = instance.tasksMap
+      val taskEffects = currentTasks.map {
+        case (taskId, task) =>
+          val newTaskId = op.oldToNewTaskIds.getOrElse(taskId, throw new IllegalStateException("reached"))
```
zen-dog: "reached" what?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -84,28 +84,37 @@ object InstanceUpdater extends StrictLogging {
...
-              since = op.timestamp
-            ),
-            tasksMap = Map(updatedTask.taskId -> updatedTask),
-            runSpecVersion = op.runSpecVersion,
-            // The AgentInfo might have changed if the agent re-registered with a new ID after a reboot
-            agentInfo = op.agentInfo
-          )
-          val events = eventsGenerator.events(updated, task = None, op.timestamp, previousCondition = Some(instance.state.condition))
-          InstanceUpdateEffect.Update(updated, oldState = Some(instance), events)
+      val nonUpdates = taskEffects.filter {
```
zen-dog: What do you think about `taskEffects.collect{ case e: TaskUpdateEffect.Update => e}`?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -84,28 +84,37 @@ object InstanceUpdater extends StrictLogging {
...
+        case _: TaskUpdateEffect.Update => false
+        case _ => true
+      }
 
-        case _ =>
-          InstanceUpdateEffect.Failure(s"Unexpected taskUpdateEffect $taskEffect")
+      val allUpdates = nonUpdates.isEmpty
```
zen-dog: You could just: `val allUpdates = taskEffects.forall(_.isInstanceOf[TaskUpdateEffect.Update])`. But that's a nit.

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -81,9 +80,11 @@ class OfferOperationFactory(
           val persistence = Mesos.Resource.DiskInfo.Persistence.newBuilder().setId(vol.id.idString)
           principalOpt.foreach(persistence.setPrincipal)
 
+          val name = vol.persistentVolume.name.getOrElse(vol.mount.mountPath)
```
zen-dog: Using `vol.mount.mountPath` as volume name or just throw "reached" like everywhere else?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -137,12 +132,15 @@ object Task {
...
     // Regular expression for matching taskIds since instance-era
     private val TaskIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)$""".r
+    private val ResidentTaskIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)\.(\d+)$""".r
 
     private val uuidGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
 
     def runSpecId(taskId: String): PathId = {
       taskId match {
-        case TaskIdWithInstanceIdRegex(runSpecId, prefix, instanceId, maybeContainer) => PathId.fromSafePath(runSpecId)
+        case ResidentTaskIdWithInstanceIdRegex(runSpecId, prefix, uuid, container, attempt) =>
```
zen-dog: `(runSpecId, _, _, _, _)`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -151,6 +149,8 @@ object Task {
 
     def containerName(taskId: String): Option[String] = {
       taskId match {
+        case ResidentTaskIdWithInstanceIdRegex(runSpecId, prefix, uuid, container, attempt) =>
```
zen-dog: `(_, _, _, container, _)`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -161,20 +161,16 @@ object Task {
 
     def attempt(taskId: String): Option[Long] = {
       taskId match {
+        case ResidentTaskIdWithInstanceIdRegex(runSpecId, prefix, uuid, container, attempt) => Some(attempt.toLong)
```
zen-dog: `(_...)`?

plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
ichernetsky: Done.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
ichernetsky: Done.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
ichernetsky: Done.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
ichernetsky: We use `nonUpdates` in a failure message below.

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
ichernetsky: It is different here. I will add a comment elaboration why it is done the way it is done.

src/main/scala/mesosphere/marathon/core/task/Task.scala
ichernetsky: It is less-readable. I intentionally kept it the way it was and still is.

src/main/scala/mesosphere/marathon/core/task/Task.scala
ichernetsky: It is less-readable. I intentionally kept it the way it was and still is.

src/main/scala/mesosphere/marathon/core/task/Task.scala
ichernetsky: It is less-readable. I intentionally kept it the way it was and still is.

src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala
ichernetsky: Done.

ichernetsky:
Rebased.

src/main/scala/mesosphere/marathon/core/task/Task.scala
zen-dog: Doesn't scala (or linter) complain about not used parameters?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -120,12 +120,24 @@ case class AppDefinition(
...
+
+  override val persistentVolumeMounts: Seq[VolumeMount] =
+    container.map(_.volumes.collect { case VolumeWithMount(_: PersistentVolume, m) => m }).getOrElse(Seq.empty)
+
+  override val externalVolumes: Seq[ExternalVolume] =
+    container.map(_.volumes.map(_.volume).collect { case pv: ExternalVolume => pv }).getOrElse(Seq.empty)
 
   override val diskForPersistentVolumes: Double = persistentVolumes.map(_.persistent.size).sum.toDouble
 
+  private[state] val persistentVolumesAndMounts: Seq[VolumeWithMount] =
```
zen-dog: nit: `persistentVolumesWITHMounts`?

ichernetsky: Will do.

src/main/scala/mesosphere/marathon/core/task/Task.scala
ichernetsky: Scala doesn't complain. By linter you mean Scapegoat? If yes, then it is fine too.

ichernetsky:
Thanks a lot, @zen-dog

docs/docs/rest-api/public/api/v2/types/residency.raml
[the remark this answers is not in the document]
timcharper: Worthy of mention, @meichstedt and I talked also and agrees that the residency fields should be deprecated and that it should not be added to the pods API. Also, here was an issue filed a year ago saying that the residency fields were superseded: https://jira.mesosphere.com/browse/MARATHON-2615

In light of this, do you feel okay removing the fields from the pods public API?

docs/docs/rest-api/public/api/v2/types/volumes.raml
timcharper: Oh geez. 🤦‍♂️ ExternalVolume size is expressed in GiB, PersistentVolume size in MB.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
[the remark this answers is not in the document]
timcharper: Yes, this was my point :). DVDIProvider doesn't need to. This code suffers greatly from speculative generality.

Change:

```
/Users/tim/src/m8e/marathon/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala

  81 |   override def build(builder: ContainerInfo.Builder, ev: ExternalVolume): Unit =
  82 |     builder.addVolumes(Builders.toUnifiedContainerVolume(ev))
  83 | 
```

To

```
/Users/tim/src/m8e/marathon/src/main/scala/mesosphere/marathon/core/externalvolume/impl/providers/DVDIProvider.scala

  81 |   override def build(builder: ContainerInfo.Builder, ev: ExternalVolume): Array[MesosVolume] =
  82 |     Array(Builders.toUnifiedContainerVolume(ev)
  83 | 
```

(and thread through, etc., etc.)

src/main/scala/mesosphere/marathon/core/externalvolume/ExternalVolumes.scala
timcharper: Yeah, I can understand the motive. It's speculative generality to write code that would possibly allow us to write other extensions in the future. Code is already extensible. We can just change the code. :) All this extra general code is just noise and provides no actual value.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -84,28 +84,41 @@ object InstanceUpdater extends StrictLogging {
...
-            runSpecVersion = op.runSpecVersion,
-            // The AgentInfo might have changed if the agent re-registered with a new ID after a reboot
-            agentInfo = op.agentInfo
-          )
-          val events = eventsGenerator.events(updated, task = None, op.timestamp, previousCondition = Some(instance.state.condition))
-          InstanceUpdateEffect.Update(updated, oldState = Some(instance), events)
+      val nonUpdates = taskEffects.filter {
+        case _: TaskUpdateEffect.Update => false
+        case _ => true
+      }
```
timcharper: consider using collect?

```
taskEffects.collect { case update: TaskUpdateEffect.Update => update }
```

docs/docs/rest-api/public/api/v2/types/pod.raml
```diff
@@ -12,6 +12,7 @@ uses:
   volumes: volumes.raml
   unreachableStrategy: unreachableStrategy.raml
   killSelection: killSelection.raml
+  residency: residency.raml
```
timcharper: Please strongly consider omitting this from the pods API.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -137,12 +132,15 @@ object Task {
 
     // Regular expression for matching taskIds since instance-era
     private val TaskIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)$""".r
+    private val ResidentTaskIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)\.(\d+)$""".r
```
timcharper: (scope creep)

Periods are not allowed in runSpecs and therefore will not appear in instanceId or taskIds. The regex should be better specified as:

```
"""^([^.]+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)\.(\d+)$""".r
```

Aside from being more correct, my intuition says that replacing `.+` with `[^.]+` will perform better, too, as there are many less branches to consider for the regex matcher.

timcharper: Further, nothing reads better (or performs) than this: :)

```
def runSpecId(taskId: String): PathId = {
  PathId.fromSafePath(taskId.takeWhile(_ != '.'))
}
```

Slight disadvantage that we won't detect invalid taskIds. But... ¯\_(ツ)_/¯ this is probably not the right place to validate that.

src/main/scala/mesosphere/marathon/raml/VolumeConversion.scala
timcharper: I remember @jdef and I having this same conversation :)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -591,11 +603,11 @@ object AppDefinition extends GeneralPurposeCombinators {
   def residentUpdateIsValid(from: AppDefinition): Validator[AppDefinition] = {
     val changeNoVolumes =
       isTrue[AppDefinition]("Persistent volumes can not be changed!") { to =>
-        val fromVolumes = from.persistentVolumes
-        val toVolumes = to.persistentVolumes
+        val fromVolumes = from.persistentVolumesAndMounts
+        val toVolumes = to.persistentVolumesAndMounts
```
timcharper: Are we checking more data than before? (mounts?) We should be able to allow the user to change mounts. This does not result in a reservation change.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
ichernetsky: Why? It would collect update ops, whereas here it has to collect non-update ops. Both `nonUpdates` and `allUpdates` are needed.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -591,11 +603,11 @@ object AppDefinition extends GeneralPurposeCombinators {
...
         def sameSize = fromVolumes.size == toVolumes.size
-        def noChange = from.persistentVolumes.forall { fromVolume =>
-          toVolumes.find(_.containerPath == fromVolume.containerPath).contains(fromVolume)
+        def noChange = fromVolumes.forall { fromVolume =>
+          toVolumes.find(_.mount.mountPath == fromVolume.mount.mountPath).contains(fromVolume)
```
timcharper: Can you speak some words to this, too? What was the motivation for this change?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
ichernetsky: We are checking the same data as before.

docs/docs/rest-api/public/api/v2/types/pod.raml
ichernetsky: Done. Will be updated by the very next push.

docs/docs/rest-api/public/api/v2/types/residency.raml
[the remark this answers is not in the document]
ichernetsky: I tend to adhere to the principle of least surprise. I feel that this difference will be surprising to some extent to our users. On the other hand, it definitely won't be their bigger surprise.

The pod residency will be removed shortly.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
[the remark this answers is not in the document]
ichernetsky: Done. `build` now returns just `MesosVolume`.

src/main/scala/mesosphere/marathon/core/externalvolume/ExternalVolumes.scala
ichernetsky: Fixed.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
ichernetsky: It is going to be true in case of pods. In case of apps, `mountPath` (known as `containerPath` previously) is used when creating volumes. Please refer to `OfferOperationFactory.createVolumes`.

docs/docs/rest-api/public/api/v2/types/volumes.raml
jdef: there is a JIRA that's tracking this problem. It's consistently a source of confusion for everyone and should be fixed. https://jira.mesosphere.com/browse/MARATHON_EE-1666

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -268,40 +329,86 @@ class InstanceOpFactoryImpl(
...
+            condition = Condition.Reserved,
+            since = now,
+            activeSince = None,
+            healthy = None
+          ),
+          tasksMap = tasks.map(t => t.taskId -> t)(collection.breakOut),
+          runSpecVersion = runSpec.version,
+          unreachableStrategy = runSpec.unreachableStrategy
+        )
+        val stateOp = InstanceUpdateOperation.Reserve(instance)
```
unterstein: the part above looks pretty similar to the app section, can we generalize things here?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -336,23 +443,38 @@ object InstanceOpFactoryImpl {
...
       agentInfo = agentInfo,
       state = InstanceState(Condition.Created, since, activeSince = None, healthy = None),
       tasksMap = taskIDs.map { taskId =>
         // the task level host ports are needed for fine-grained status/reporting later on
-        val taskHostPorts: Seq[Int] = taskId.containerName.map { ctName =>
-          allocPortsByCTName.withFilter{ case (name, port) => name == ctName }.map(_._2)
-        }.getOrElse(Seq.empty[Int])
-
-        val networkInfo = NetworkInfo(agentInfo.host, taskHostPorts, ipAddresses = Nil)
+        val networkInfo = taskNetworkInfos.getOrElse(taskId, throw new Exception("should not have happened"))
```
unterstein: Please provide a reasonable error message.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -336,23 +443,38 @@ object InstanceOpFactoryImpl {
...
+    taskIDs.map { taskId =>
+      // the task level host ports are needed for fine-grained status/reporting later on
+      val taskHostPorts: Seq[Int] = taskId.containerName.map { ctName =>
+        allocPortsByCTName.withFilter { case (name, port) => name == ctName }.map(_._2)
+      }.getOrElse(Seq.empty[Int])
+
+      val networkInfo = NetworkInfo(agentInfo.host, taskHostPorts, ipAddresses = Nil)
+      taskId -> networkInfo
+    }(collection.breakOut)
+  }
```
unterstein: Is this shift of code necessary?

plugin-interface/src/main/scala/mesosphere/marathon/plugin/RunSpec.scala
```diff
@@ -25,6 +25,16 @@ trait RunSpec {
...
+
+  /**
+    * Volume definitions
+    */
+  val volumes: Seq[VolumeSpec]
+
+  /**
+    * Volume mounts.
+    */
+  val volumeMounts: Seq[VolumeMountSpec]
```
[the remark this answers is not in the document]
ichernetsky: Because it is just the beginning of unifying apps and pods. Otherwise, in order to get rid of all tiny idiosyncrasies it would be necessary to finishing their unification in addition to the volume refactoring and the pod PV implementation, which is way out of scope.

docs/docs/rest-api/public/api/v2/types/residency.raml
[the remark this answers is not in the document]
ichernetsky: `residency` is not used with pods anymore.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -144,13 +145,27 @@ object ResourceMatcher extends StrictLogging {
     // that means if the resources that are matched are still unreserved.
     def needToReserveDisk = selector.needToReserve && runSpec.diskForPersistentVolumes > 0
 
-    val diskMatch = if (needToReserveDisk)
+    val diskMatch = if (needToReserveDisk) {
+      val volumes = runSpec.persistentVolumes
+      val mounts = runSpec.persistentVolumeMounts
+      val (namedVolumes, namedMounts) =
+        if (volumes.exists(_.name.isEmpty) && volumes.length == mounts.length) {
```
unterstein: I`m not sure if I can follow this lines...what is the intention to behind this condition above and the separate logic below?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -309,10 +324,22 @@ object ResourceMatcher extends StrictLogging {
...
+    volumeMounts: Seq[VolumeMount],
     scope: ScalarMatchResult.Scope = ScalarMatchResult.Scope.NoneDisk): Seq[ScalarMatchResult] = {
 
+    def matchesProfileName(profileName: Option[String], resource: Protos.Resource): Boolean = {
+      profileName.forall { specifiedProfileName =>
+        if (resource.hasDisk && resource.getDisk.hasSource && resource.getDisk.getSource.hasProfile) {
+          specifiedProfileName == resource.getDisk.getSource.getProfile
+        } else {
+          false
+        }
```
unterstein:
```
specifiedProfileName =>
    resource.hasDisk && resource.getDisk.hasSource && resource.getDisk.getSource.hasProfile && 
        specifiedProfileName == resource.getDisk.getSource.getProfile
```
?

docs/docs/rest-api/public/api/v2/types/volumes.raml
```diff
@@ -58,31 +45,36 @@ types:
         minimum: 0
         description: |
           For `mount` mesos disk resources, the optional maximum size of an exclusive mount volume to be considered.
+      profileName?:
```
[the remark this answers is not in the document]
ichernetsky: Fixed.

Not sure about when we can remove them.

ichernetsky:
@timcharper, @jdef, thanks a ton for your reviews.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -220,23 +224,25 @@ object TaskGroupBuilder extends StrictLogging {
...
 
-      mesosNetworks.foreach(containerInfo.addNetworkInfos(_))
+      mesosNetworks.foreach(containerInfo.addNetworkInfos)
 
-      podDefinition.volumes.collect{
+      podDefinition.volumes.collect {
         // see the related code in computeContainerInfo
         case e: EphemeralVolume =>
           mesos.Volume.newBuilder()
+            .setContainerPath(ephemeralVolumePathPrefix + e.name.getOrElse(""))
```
unterstein: what is this semantically if we fall back to default `""` ?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -220,23 +224,25 @@ object TaskGroupBuilder extends StrictLogging {
...
             .setMode(mesos.Volume.Mode.RW) // if not RW, then how do containers plan to share anything?
             .setSource(mesos.Volume.Source.newBuilder()
               .setType(mesos.Volume.Source.Type.SANDBOX_PATH)
               .setSandboxPath(mesos.Volume.Source.SandboxPath.newBuilder()
                 .setType(mesos.Volume.Source.SandboxPath.Type.SELF)
-                .setPath(ephemeralVolPathPrefix + e.name) // matches the path in computeContainerInfo
+                .setPath(ephemeralVolumePathPrefix + e.name.getOrElse("")) // matches the path in computeContainerInfo
```
unterstein: see above

docs/docs/rest-api/public/api/v2/types/residency.raml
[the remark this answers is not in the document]
ichernetsky: Let's call it a collateral improvement and keep it as is then.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -191,62 +197,116 @@ class InstanceOpFactoryImpl(
...
+          containerNameToTaskId.getOrElse(container.name, throw new IllegalStateException(
+            s"failed to get a task ID for the given container name: ${container.name}"))
+        }
+
+        val (executorInfo, groupInfo, hostPorts) = TaskGroupBuilder.build(pod, offer,
+          instanceId, podContainerTaskIds, builderConfig, runSpecTaskProc, resourceMatch, Some(volumeMatch))
+
+        val networkInfos = podTaskNetworkInfos(pod, agentInfo, podContainerTaskIds, hostPorts)
+        val now = clock.now()
+        val statuses = networkInfos.map {
```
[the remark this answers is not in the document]
ichernetsky: `podTaskNetworkInfos` always returns `networkInfo`s all for pod tasks/containers.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
ichernetsky: I do not see any obvious advantage of generalizing things here. In my opinion, it will save  a couple of lines, but will hurt readability.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
ichernetsky: What do you mean? A common code got factored out into a separate function, AFAICT.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
ichernetsky: Done.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
ichernetsky: I will elaborate on it.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
ichernetsky: Adjusted.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: Adjusted.

ichernetsky:
@unterstein, thank you for the review.

docs/docs/rest-api/public/api/v2/types/pod.raml
timcharper: 🙇

PR #5814, opened 2017-12-08
Pass killGracePeriodSeconds value to Mesos for pods

kensipe:
Pass killGracePeriodSeconds value to Mesos for pods

Summary:
Cherry pick wasn't possible... There are difference from the original work on #5808 in the test.

JIRA Issues: MARATHON-7952


kensipe:
release 1.5.4 to follow this merge

PR #5815, opened 2017-12-11
Using MarathonPollerObserver on perf tests

wavesoft:
This PR introduces a few fixes to the performance tests:

1. Uses [`MarathonPollerObserver`](http://dcos-performance-test-driver.readthedocs.io/en/latest/classes/Observers.html#marathonpollerobserver) instead of subscribing to the SSE event stream. This is creating synthetic "Deployment Success" events deep-diffing the `/groups` endpoint.
2. Adds support for running the tests in a container instead of in the host machine. The main problem here is that marathon is no longer on `127.0.0.1` but on a particular named host. To accommodate this:
    1. The `jmx_host` is now configurable
    2. The `DCLUSTER_ARGS` is introduced to allow arbitrary flags to be passed to the [`marathon-dcluster`](https://github.com/wavesoft/marathon-dcluster#configuration) script, namely the `--network` configuration. You can check the [launch script](https://github.com/mesosphere/marathon-perf-testing/blob/master/run-marathon-perf-tests.sh#L61) for more details on this.

_The performance infrastructure is currently using this branch (`ic/fix/perf-tests`) for testing. Until we merge this branch I am going to periodically rebase this PR_

### DIFF
--- a/tests/performance/config/perf-driver/test-1-apps-n-instances.yml
+++ b/tests/performance/config/perf-driver/test-1-apps-n-instances.yml
@@ -21,7 +21,8 @@ include:
 
   # Use marathon's SSE endpoint to measure our metrics
   - observers/observer-marathon-http-timing.yml
-  - observers/observer-marathon-sse.yml
+  # - observers/observer-marathon-sse.yml
+  - observers/observer-marathon-poller.yml
   - observers/observer-jmx.yml
 
   # Measure the given metrics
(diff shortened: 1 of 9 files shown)


tests/performance/config/perf-driver/test-1-apps-n-instances.yml
```diff
@@ -21,7 +21,8 @@ include:
 
   # Use marathon's SSE endpoint to measure our metrics
   - observers/observer-marathon-http-timing.yml
-  - observers/observer-marathon-sse.yml
+  # - observers/observer-marathon-sse.yml
```
meln1k: Should we remove the comment too?

ichernetsky: 👍

PR #5816, opened 2017-12-11
Benchmark and speed up service port assignment by up to three magnitu…

unterstein:
…des. (#5787)

Summary:
As @zen-dog found a big performance degradation when Marathon has already 5000 apps and a new app is added. He found that `GroupManagerImpl.assignDynamicServicePorts` had an exponential runtime.

The benchmark and flamegraph showed that most time was spend in copying and building the `RootGroup` after the update. It also seems that we update all apps even when the get the same port assigned again. This change introduces a filter so that only apps that are new or changed are updated in the `RootGroup`.

```
```

This is only a fix for small app updates. If many apps are updated at once we will still hit this bottleneck. We should look into the creation of `transitiveAppsById` and `transitivePodsById`
which take up most of the runtime.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
@@ -224,6 +224,17 @@ private[impl] class GroupManagerActor(
       }
   }
 
+  /**
+    * Checks whether newApp is new or changed.
+    *
+    * @param originalApps A map of the original apps form before the update.
+    * @param newApp       The new app that is tested.
+    * @return true if app is new or an updated, false otherwise.
+    */
+  def changedOrNew(originalApps: Map[AppDefinition.AppKey, AppDefinition], newApp: AppDefinition): Boolean = {
+    originalApps.get(newApp.id).forall { _.isUpgrade(newApp) }
+  }
+
   private[impl] def assignDynamicServicePorts(from: RootGroup, to: RootGroup): RootGroup = {
     val portRange = Range(config.localPortMin(), config.localPortMax())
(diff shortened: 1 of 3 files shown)


unterstein:
This is a back port of #5787

I decided against refactoring the service port assignment logic to a dedicated class `AssignDynamicServiceLogic` as it was done in the initial PR. The logic of the GroupManager was refactored from an Actor to a Implementation from 1.4 to 1.5 and a lot of logic changed during the networking overhaul. Therefore the logic should stay in the actor in marathon 1.4 and only the two places should change, where the performance was improved. The tests which are in the `AssignDynamicServiceLogicTest` in releases/1.5 were previously in the `GroupManagerTest`, therefore they stay also in the `GroupManagerActorTest`.

WDYT?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerActor.scala
```diff
@@ -276,22 +287,25 @@ private[impl] class GroupManagerActor(
...
-        // assign values for service ports that the user has left "blank" (set to zero)
-        case app: AppDefinition if app.hasDynamicServicePorts => assignPorts(app)
-        case app: AppDefinition =>
-          // Always set the ports to service ports, even if we do not have dynamic ports in our port mappings
-          app.copy(
-            portDefinitions = mergeServicePortsAndPortDefinitions(app.portDefinitions, app.servicePorts)
-          )
-      }
+    val dynamicApps: Iterator[AppDefinition] =
+      to.transitiveApps
```
meln1k: I just wander how big is the difference if you would use `.view` instead of iterator :)

PR #5818, opened 2017-12-11
More accurate parsing of the Mesos task info in SI tests MARATHON-7958

alenkacz:
Summary:
Some parts of mesos task info were queries out of first task status but that is not possible anymore because Mesos started sending TASK_STARTING status. We need to explicitly ask for STATUS_RUNNING for this to work.

JIRA issues: MARATHON-7958, MARATHON-7959

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -753,3 +753,13 @@ def running_task_status(task_statuses):
             return task_status
 
     assert False, "Did not find a TASK_RUNNING status in task statuses: %s" % (task_statuses,)
+
+
+def task_by_name(tasks, name):
+    """ Find mesos task by its name
+    """
+    for task in tasks:
+        if task['name'] == name:
+            return
+
+    assert False, "Did not find task with name %s in this list of tasks: %s" % (name, tasks,)
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -742,7 +742,7 @@ def marathon_leadership_changed(original_leader):
 def running_status_network_info(task_statuses):
     """ From a given list of statuses retrieved from mesos API it returns network info of running task.
     """
-    return running_task_status['container_status']['network_infos'][0]
+    return running_task_status(task_statuses)['container_status']['network_infos'][0]
 
 
 def running_task_status(task_statuses):
@@ -759,7 +759,9 @@ def task_by_name(tasks, name):
     """ Find mesos task by its name
     """
     for task in tasks:
+        print(task['name'])
+        print(name)
         if task['name'] == name:
-            return
+            return task
 
     assert False, "Did not find task with name %s in this list of tasks: %s" % (name, tasks,)
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -759,8 +759,6 @@ def task_by_name(tasks, name):
     """ Find mesos task by its name
     """
     for task in tasks:
-        print(task['name'])
-        print(name)
         if task['name'] == name:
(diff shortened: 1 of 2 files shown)


tests/system/marathon_pods_tests.py
```diff
@@ -425,7 +425,7 @@ def test_pod_with_container_network():
 
     tasks = common.get_pod_tasks(pod_id)
 
-    network_info = tasks[0]['statuses'][0]['container_status']['network_infos'][0]
+    network_info = common.running_status_network_info(tasks[0]['statuses'])
```
zen-dog: Same problem in line 457?

alenkacz: Good point, thanks. Fixed

tests/system/common.py
```diff
@@ -737,3 +737,13 @@ def marathon_leadership_changed(original_leader):
...
+
+
+def running_status_network_info(task_statuses):
+    """ From a given list of statuses retrieved from mesos API it returns network info of running task.
+    """
+    for task_status in task_statuses:
+        if task_status['state'] == "TASK_RUNNING":
+            return task_status['container_status']['network_infos'][0]
+
+    pytest.fail("did not found a TASK_RUNNING status in task statuses: %s" %(task_statuses,))
```
zen-dog: I would just `assert False, "Haven't found..."` here

alenkacz: I don't see much difference but changed...

tests/system/common.py
```diff
@@ -737,3 +737,13 @@ def marathon_leadership_changed(original_leader):
...
+
+
+def running_status_network_info(task_statuses):
+    """ From a given list of statuses retrieved from mesos API it returns network info of running task.
+    """
+    for task_status in task_statuses:
+        if task_status['state'] == "TASK_RUNNING":
+            return task_status['container_status']['network_infos'][0]
+
+    assert False, "Did not found a TASK_RUNNING status in task statuses: %s" % (task_statuses,)
```
jeschkies: Nit: "Did not find a TASK_RUNNING"

tests/system/common.py
```diff
@@ -737,3 +737,13 @@ def marathon_leadership_changed(original_leader):
...
     __marathon_leadership_changed_in_marathon_api(original_leader)
     __marathon_leadership_changed_in_mesosDNS(original_leader)
+
+
+def running_status_network_info(task_statuses):
+    """ From a given list of statuses retrieved from mesos API it returns network info of running task.
+    """
+    for task_status in task_statuses:
+        if task_status['state'] == "TASK_RUNNING":
+            return task_status['container_status']['network_infos'][0]
```
jeschkies: Somehow it seems to me that we make a lot of assumptions here. E.g. we retrieve just the first network info of the first task that is task_running. Shouldn't identify the task by name?

alenkacz: I am querying tasks by name now, good point

tests/system/common.py
```diff
@@ -737,3 +737,31 @@ def marathon_leadership_changed(original_leader):
...
+
+    assert False, "Did not find a TASK_RUNNING status in task statuses: %s" % (task_statuses,)
+
+
+def task_by_name(tasks, name):
+    """ Find mesos task by its name
+    """
+    for task in tasks:
+        print(task['name'])
+        print(name)
```
zen-dog: Forgot to remove `print`s?

PR #5820, opened 2017-12-12
Task IDs ignored if Mesos Task ID does not match expected Marathon Task ID

meln1k:
Task IDs ignored if Mesos Task ID does not match expected Marathon Task ID

Summary: mesos updates with unknown task ids will result in killing the task instead of silently ignoring the update.

JIRA issues: MARATHON-7926


src/test/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImplTest.scala
```diff
@@ -294,6 +293,26 @@ class TaskStatusUpdateProcessorImplTest extends AkkaUnitTest {
       Then("not do anything else")
       verifyNoMoreInteractions()
     }
+
+    "receiving an running update for known instnace but unknown task" in new Fixture {
```
jeschkies: 1. Nit: typo.
2. Nit: The test start with ` "The TaskStatusUpdateProcessor implementation" should`. Scalatest is converting it to names such as `The TaskStatusUpdateProcessor implementation should receiving an running update for known instance but unknown task`.  I believe it should be `The TaskStatusUpdateProcessor implementation should not kill unknown tasks but ignore them`.

src/test/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImplTest.scala
```diff
@@ -294,6 +293,26 @@ class TaskStatusUpdateProcessorImplTest extends AkkaUnitTest {
...
+      val appId = PathId("/app")
+      val instance = TestInstanceBuilder.newBuilder(appId).addTaskResidentLaunched().getInstance()
+      val incrementedTaskId = Task.Id.forResidentTask(Task.Id(instance.instanceId.idString))
```
jeschkies: Wow. This is an interesting fix for crafting the instance id 🙂.

src/test/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImplTest.scala
```diff
@@ -294,6 +293,26 @@ class TaskStatusUpdateProcessorImplTest extends AkkaUnitTest {
...
+      val status = MesosTaskStatusTestHelper.running(incrementedTaskId)
+
+      taskTracker.instance(instance.instanceId) returns Future.successful(Some(instance))
+
+      When("publish the status")
+      updateProcessor.publish(status).futureValue
+      Then("load the task in the task tracker")
+      verify(taskTracker).instance(instance.instanceId)
+      Then("initiate the task kill")
+      verify(killService).killUnknownTask(incrementedTaskId, KillReason.Unknown)
```
jeschkies: Wait, I thought we do not to kill the task.

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -47,7 +47,20 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
 
+    def taskIsUnknown(instance: Instance, taskId: Task.Id) = {
+      instance.tasksMap.get(taskId).isEmpty
+    }
+
     instanceTracker.instance(taskId.instanceId).flatMap {
+      case Some(instance) if taskIsUnknown(instance, taskId) =>
+        if (killWhenUnknown(taskCondition)) {
+          killUnknownTaskTimer {
+            logger.warn(s"Kill unknown ${taskId} because instance doesn't contain it")
```
alenkacz: I am not sure what does this mean: "instance does not contain it"... maybe marathon does not know about it or what is it trying to say ... ?

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -47,7 +47,20 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
 
+    def taskIsUnknown(instance: Instance, taskId: Task.Id) = {
+      instance.tasksMap.get(taskId).isEmpty
+    }
+
     instanceTracker.instance(taskId.instanceId).flatMap {
+      case Some(instance) if taskIsUnknown(instance, taskId) =>
+        if (killWhenUnknown(taskCondition)) {
+          killUnknownTaskTimer {
+            logger.warn(s"Kill unknown ${taskId} because it's not tracked by marathon")
```
meichstedt: I think we could be very explicit here, since this is an exceptional case where we might want to have more info in the logs if it occurs. Something like
> `Kill ${taskId} because it's unknown to marathon. The related instance ${instanceId} is associated with ${instance.tasksMap.keys}`

src/test/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImplTest.scala
```diff
@@ -294,6 +293,26 @@ class TaskStatusUpdateProcessorImplTest extends AkkaUnitTest {
...
       verifyNoMoreInteractions()
     }
+
+    "kill the orphaned task when receiving an running update for the known instance but unknown task" in new Fixture {
+      val appId = PathId("/app")
+      val instance = TestInstanceBuilder.newBuilder(appId).addTaskResidentLaunched().getInstance()
+      val incrementedTaskId = Task.Id.forResidentTask(Task.Id(instance.instanceId.idString))
+      val status = MesosTaskStatusTestHelper.running(incrementedTaskId)
+
+      taskTracker.instance(instance.instanceId) returns Future.successful(Some(instance))
```
meichstedt: Leftover: can you please rename the val to `instanceTracker`?

src/test/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImplTest.scala
meichstedt: The ticket says
> [...]
> When Mesos sends a task running update for a task ID with a launchCount counterpart newer than that which Marathon expects,
> **Marathon kills the task**, and eventually replaces it

meichstedt: Should we add a new `KillReason.Outdated`?

meichstedt:
@jeschkies @timcharper AFAICT the acceptance criteria are written in that way because this is one way how people can actually end up in this situation.

Background: Marathon saves instance updated before it accepts offers with `LaunchTask` operations. If there's a zk write failure, we expect Marathon to abdicate, so it's rather unlikely that a task is launched even though we couldn't update the instance state.

I don't think it's worth writing an integration test that involves restoring zk state from a backup. I'm OK with a unit test that verifies correct behavior when handling an update for `task.7`when the instance expects `task.6`. However, that could be expressed in the JIRA as a test plan :)

src/main/scala/mesosphere/marathon/core/task/termination/KillReason.scala
```diff
@@ -34,6 +34,9 @@ object KillReason {
   /** The task is killed because it is unknown */
   case object Unknown extends KillReason
 
+  /** The task is killed because marathon has a different launchCount for the task of the respective instance */
+  case object Outdated extends KillReason
```
meichstedt: Glad to see my suggestion be implemented so fast, but it was really a question. Thinking about it now, `Diverged` might be better – in the case we encountered, the update was actually for a _newer_ attempt, so it was the taskId in the persistence layer that was outdated, not the task that should have been killed.

@meln1k, @timcharper, does `Diverged` make sense to you?

meln1k: @meichstedt how about `NotInSync`?

PR #5822, opened 2017-12-12
Introduce wait (and retries) before hitting v2/info MARATHON-7960

alenkacz:
Summary:
The scenario that happened is described in JIRA. Basically it might happen, that the cluster is not fully up-to-date after starting up and before this cluster_info method is hitted.

JIRA issues: MARATHON-7960


alenkacz:
@zen-dog to convince you this is necessary, dcos integration tests are doing this too as I found out today... https://github.com/dcos/dcos-test-utils/blob/8d019c99059c960ec9a304aa0843cfdf660356cc/dcos_test_utils/dcos_api.py#L193

tests/system/common.py
```diff
@@ -126,6 +126,18 @@ def cluster_info(mom_name='marathon-user'):
...
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=5, retry_on_exception=ignore_exception)
+def wait_for_marathon_about():
+    """ When the cluster is starting, it might happen that there is some delay in:
+    - marathon leader registration with mesos
+    - admin router refreshing cache (every 30s)
+
+    To improve our chances everything happened before we hit the v2/info we retry here.
+    """
+    client = marathon.create_client()
+    return client.get_about()
```
kensipe: according to the docs, this function exists to ensure that the 1 marathon leader is registered and admin router cache is updated...   questions:
1. the about is a hit against `v2/info` which will return regardless of status or registration.   Should we have an endpoint that does what is documented?   or is there something in the /info JSON response we can wait on?

2. I'm not sure why admin router refresh is mentioned here... the admin router on a master is going to route to it's local marathon.   I don't believe it takes 30s for an end point to materialize.  I don't understand what this comment means.

kensipe: and how is it different than https://github.com/mesosphere/marathon/blob/4c222b13bc73718e250e8ddafdb1627f4fd9773f/tests/system/fixtures/__init__.py#L14-L20

kensipe: which is call for all functions:  https://github.com/mesosphere/marathon/blob/a5caab6ac148aa4a12e082123b15839d184c2a65/tests/system/test_marathon_root.py#L54

tests/system/test_marathon_root.py
```diff
@@ -60,6 +60,7 @@ def marathon_service_name():
 
 
 def setup_module(module):
+    common.wait_for_marathon_about()
     common.cluster_info()
```
kensipe: add the `cluster_info` already calls the marathon about.
https://github.com/mesosphere/marathon/blob/master/tests/system/common.py#L109-L114

It think if we just call a wait for marathon function at setup prior to cluster_info we are good.   We used to have just a wait for marathon... perhaps it was refactored out.

alenkacz: well the cluster info prints out marathon version as a part of cluster info so I think it is still useful...

I refactored the wait to the setup phase because it was suggested by @zen-dog and I actually liked that suggestion. So now we call it twice but I guess it does not matter

tests/system/common.py
zen-dog: That's a good point. `wait_for_service_endpoint` there is already waiting for marathon.

alenkacz: 1. we are calling marathon through service/marathon so through admin router
2. admin router returns HTTP 404 on service/marathon until it receives info from mesos that marathon registered - this is done every 30s, it is even visible in the AR logs

But yeah, I guess we CAN probably use wait_for_service_endpoint instead, good point.

alenkacz: btw. it is called for all test functions, right? but not for the setup function...

alenkacz:
@kensipe I am not waiting for service endpoint - see if it makes more sense to you...

alenkacz:
@zen-dog @kensipe so can I get your approvals now?

zen-dog:
You got my blessing ;)

PR #5825, opened 2017-12-15
Expose common Marathon dependencies as transitive dependencies

timcharper:
Even though marathon-plugin-interface doesn't directly use Akka, there's
little reason why plugins should have to include their separate version
of Akka.

JIRA Issues: MARATHON-7917


kensipe:
🚢

PR #5826, opened 2017-12-15
Use Marathon event stream directly for SI tests.

jeschkies:
Summary:
Some system integration tests SSH into the cluster and call the Marathon
event API from there. This is not necessary. We can call it from the
outside and assert each event.

JIRA issues: MARATHON-7964


tests/system/sseclient/README.md
```diff
@@ -0,0 +1,2 @@
+# Server Side Events (SSE) client for Python
+This is bacially a clone form [mpetazzoni/sseclient](https://github.com/mpetazzoni/sseclient).
```
alenkacz: basically

alenkacz: and from not form

tests/system/common.py
```diff
@@ -763,3 +763,11 @@ def task_by_name(tasks, name):
             return task
 
     assert False, "Did not find task with name %s in this list of tasks: %s" % (name, tasks,)
+
+
+@retrying.retry(wait_fixed=100, stop_max_attempt_number=30, retry_on_exception=ignore_exception, stop_max_delay=30000)
```
alenkacz: is 3s enough?

tests/system/common.py
```diff
@@ -763,3 +763,11 @@ def task_by_name(tasks, name):
...
+def assert_event(event_type, events):
```
alenkacz: assert_event_type maybe? Because we are not asserting any contents of the event...

tests/system/common.py
```diff
@@ -763,3 +763,11 @@ def task_by_name(tasks, name):
...
+    """ Skip events until event with given type was found."""
```
alenkacz: btw. this can be potentially dangerous. If the order of the events we are expecting is not guaranteed...

if we do
assert_event(a)
assert_event(b)

but the events arrive as:
b
a

we will never get success out of this...

tests/system/common.py
jeschkies: It's 30 seconds.

tests/system/common.py
jeschkies: That is true but I think for the events we check the ordering is guaranteed.

tests/system/common.py
alenkacz: if we do up to 30 retries every 100 ms it might be just 3s, right?

tests/system/sseclient/__init__.py
```diff
@@ -0,0 +1,139 @@
+"""
```
alenkacz: btw. what is the advantage of pulling in a library code like this?

I mean the code in the original repo is unit tested and we lose any future bugfixes...

jeschkies: I want to limit the dependencies as we don't use virtualenv or Pipenv and the like to provision and environment with all dependencies. I tried to use Pipenv but failed. See https://github.com/pypa/pipenv/issues/931 for details.

alenkacz: ah, get it... thanks for the explanation!

jeschkies: #5831 is a proposal for Pipenv.

tests/system/common.py
jeschkies: Ah. You a right. I looked at the max wait time.

jeschkies:
The job is stuck again at `test_event_channel_for_pods`:

```
'Connection': 'keep-alive', 'X-Marathon-Leader': 'http://10.0.7.242:8080', 'Cache-Control': 'no-cache, no-store, must-revalidate', 'Pragma': 'no-cache', 'Expires': '0', 'Vary': 'Accept-Encoding, User-Agent'}
┊ util.py                    606 DEBUG    duration: dcos.http._request: 0.01s
┊ connectionpool.py          208 DEBUG    Starting new HTTP connection (1): 34.210.34.200
┊ connectionpool.py          396 DEBUG    http://34.210.34.200:80 "GET /service/marathon/v2/events HTTP/1.1" 200 None
```
I wonder why it's querying the wrong endpoint.

jeschkies:
The tests timed out. Weird that it's reported as passed.

tests/system/common.py
kensipe: `assert_event_type` sounds good

tests/system/marathon_pods_tests.py
```diff
@@ -102,11 +101,13 @@ def test_create_pod_with_private_image():
         common.delete_secret(secret_name)
 
 
-@shakedown.dcos_1_9
-@pytest.mark.usefixtures("wait_for_marathon_and_cleanup", "events_to_file")
-def test_event_channel_for_pods():
+@shakedown.dcos_1_9 # NOQA
```
kensipe: why NOQA for this?

jeschkies: Good question. Flake8 was complaining. I should ignore a specific rule. The issue was about a method parameter.

jeschkies: It would be a redefinition pf the fixture.

jeschkies:
Superseded by #5850

PR #5827, opened 2017-12-16
Upgrade to sbt 1.0.4

timcharper:
Bump plugins to latest. Fix resulting compile errors and deprecation warnings.

JIRA issues: MARATHON-7917


project/PhabricatorTestReportListener.scala
```diff
@@ -65,14 +65,14 @@ class PhabricatorTestReportListener(outputDir: File) extends TestsListener {
     endGroup(name, TestResult.Failed)
   }
 
-  override def endGroup(name: String, result: sbt.TestResult.Value): Unit = {
+  override def endGroup(name: String, result: sbt.TestResult): Unit = {
```
kensipe: why do we need the PhabricatorTestReportListener?   we don't use Phab any longer

timcharper: Yeah; thought the same thing.

It is referenced here: https://github.com/mesosphere/marathon/blob/fabba407034f5a3a55428a0d2555c9ae503e64bb/ci/githubClient.sc#L71

But it's possible that this part of the script isn't even called, or, the file isn't even used.

Getting rid of it would be ideal. It is quite a bit of complexity for our SBT project.

alenkacz: We even have ticket for that! :D https://jira.mesosphere.com/browse/MARATHON-7780

alenkacz:
@meln1k don't you have branch updating to the sbt 1.0.x too from before?

project/PhabricatorTestReportListener.scala
ichernetsky: Are you going to delete it in another PR?

kensipe: lets delete please

PR #5829, opened 2017-12-18
 Transfer all bytes with Guava instead of parts with IOUtils. 

jeschkies:
A user report that the event stream would stutter. He found that the
change to Apache IOUtils is not reading the complete input stream and
also buffers the input before it's copied. We switch to Guava
ByteStreams.copy which copies all bytes.

JIRA issues: MARATHON-7919

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
+++ b/src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
@@ -10,12 +10,12 @@ import javax.servlet.http.{ HttpServletRequest, HttpServletResponse }
 
 import akka.Done
 import akka.http.scaladsl.model.StatusCodes._
+import com.google.common.io.{ ByteStreams, Closeables }
 import com.google.inject.Inject
 import mesosphere.chaos.http.HttpConf
 import mesosphere.marathon.core.election.ElectionService
 import mesosphere.marathon.io.IO
 import mesosphere.marathon.stream.Implicits._
-import org.apache.commons.io.IOUtils
 import org.slf4j.LoggerFactory
 
 import scala.annotation.tailrec
@@ -294,13 +294,13 @@ class JavaUrlConnectionRequestForwarder @Inject() (
             () => cloneResponseEntity(leaderConnection, response)
           )
         } finally {
-          Try(leaderConnection.getInputStream.close())
-          Try(leaderConnection.getErrorStream.close())
+          Closeables.closeQuietly(leaderConnection.getInputStream())
+          Closeables.closeQuietly(leaderConnection.getErrorStream())
         }
       }
     } finally {
-      Try(request.getInputStream.close())
-      Try(response.getOutputStream.close())
+      Closeables.closeQuietly(request.getInputStream())
+      Closeables.close(response.getOutputStream(), true)
     }
 
   }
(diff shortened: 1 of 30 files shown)


jeschkies:
I tested the functionality manually.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
```diff
@@ -294,32 +292,29 @@ class JavaUrlConnectionRequestForwarder @Inject() (
...
-          Try(leaderConnection.getErrorStream.close())
+          Closeables.closeQuietly(leaderConnection.getInputStream())
+          Closeables.closeQuietly(leaderConnection.getErrorStream())
         }
       }
     } finally {
-      Try(request.getInputStream.close())
-      Try(response.getOutputStream.close())
+      Closeables.closeQuietly(request.getInputStream())
+      Closeables.close(response.getOutputStream(), true)
```
alenkacz: why not closeQuietly here? If I see it correctly in the code, it does the same (close quietly calls close with swallow exception)

src/test/scala/mesosphere/marathon/integration/setup/MarathonTest.scala
```diff
@@ -878,8 +878,10 @@ trait MarathonClusterTest extends Suite with StrictLogging with ZookeeperServerT
     super.afterAll()
   }
 
-  def nonLeader(leader: ITLeaderResult): MarathonFacade = {
-    marathonFacades.find(!_.url.contains(leader.port.toString)).get
+  def nonLeader(): MarathonFacade = {
```
alenkacz: idea is not finding any usages of this method... and since you did not have to alter any usages here and the code still compiles, do we need this method ... ?

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
jeschkies: This is confusing, I know. There is no `closeQuietly` for output streams.

src/test/scala/mesosphere/marathon/integration/setup/MarathonTest.scala
jeschkies: I was wondering the same. I did use it for an integration test that did not work. Since it is not used at all I'll just remove it.

src/main/scala/mesosphere/marathon/api/LeaderProxyFilter.scala
alenkacz: oh I see... that's weird :) ok thanks for the explanation

timcharper:
@jeschkies This seems like an improvement and the blocking does not seem any worse than we do today :)

And, Jetty 9 does support non-blocking responses. See my working implementation here which propagates through flushes:

https://github.com/mesosphere/marathon/tree/tharper/multi-master-event-stream-non-blocking-poc

I mostly wanted to see if it was possible. If you think we should go this route then we just need to replace the blocking java URL client from `createAndConfigureConnection` with an akka-http client connection, and then get rid of the `sillyThreadBlockingReader`.

jeschkies:
> I mostly wanted to see if it was possible. If you think we should go this route then we just need to replace the blocking java URL client from createAndConfigureConnection with an akka-http client connection, and then get rid of the sillyThreadBlockingReader.

Thanks a lot! I'll take a look.

PR #5831, opened 2017-12-18
 Provision node and run SI tests with pipenv. 

jeschkies:
Summary:
This introduces pipenv to manage our system integration test dependencies
and environment. This way we can introduces and remove dependencies
without changing the Jenkins Docker images and maintain reproducible runs.

An environment is initialized with `make init`.
The tests are run with with `make test`.

You can also jump into the shell of the environment with `pipenv shell`.


jeschkies:
[This](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-open/job/karsten%252Fpipenv/14/) is an example job. Bundling does not work yet.

ci/launch_cluster.sh
```diff
@@ -24,8 +24,9 @@ fi
 echo "Workspace: ${WORKSPACE}"
 echo "Using: ${TEMPLATE}"
 
-apk update
-apk --upgrade add gettext wget
+# apk update
```
alenkacz: why did we just commented out this?

jeschkies: Ah, this should go.

jeschkies:
I've also triggered a test run against a [strict cluster](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-strict/job/karsten%252Fpipenv/).

PR #5832, opened 2017-12-18
Docs: Add multiple protocol specification

dirkjonker:
Summary:
This specific option did not exist in the documentation, however it does work.

It is available in the DC/OS documentation though: https://docs.mesosphere.com/1.10/deploying-services/service-ports/#definitions

Note that without spaces it seems not to work.


alenkacz:
@dirkjonker thank you for the fix!

PR #5833, opened 2017-12-18
Avoid querying mesosDNS multiple times

alenkacz:
Summary:
What I've seen in the logs is the following:
┊ leader according to marathon API: 10.0.6.88:8080
┊ leader according to MesosDNS: 10.0.6.88
┊ switched leader to: 10.0.4.59

Second and third queries mesos DNS. But that runs on all masters so it probably might happen, that we hit not updated not on the next query. This change does not do that unnecessary extra call to mesos DNS that is also not 100% reliable in terms of the data it returns.

Also the marathon method did not work properly because it was returning API with port.

JIRA issues: MARATHON-7965


tests/system/common.py
```diff
@@ -328,14 +328,13 @@ def assert_app_tasks_healthy(client, app_def):
...
-    print('marathon: {}'.format(marathon_leader))
-    print('leader: {}'.format(master_leader))
+    print('marathon leader: {}'.format(marathon_leader))
+    print('mesos leader: {}'.format(master_leader))
 
     if marathon_leader == master_leader:
         delete_marathon_path('v2/leader')
         shakedown.wait_for_service_endpoint('marathon', timedelta(minutes=5).total_seconds())
-        marathon_leadership_changed(marathon_leader)
-        marathon_leader = shakedown.marathon_leader_ip()
```
alenkacz: this is where the extra call to mesos DNS was made and is no longer there...

jeschkies:
> mesos DNS that is also not 100% reliable in terms of the data it returns

Is this a bug or does it just take time until the data is correct?

tests/system/common.py
```diff
@@ -727,16 +726,19 @@ def __marathon_leadership_changed_in_marathon_api(original_leader):
...
     """
-    current_leader = marathon.create_client().get_leader()
+    # Leader is returned like this 10.0.6.88:8080 - we want just the IP
+    current_leader = marathon.create_client().get_leader().split(':', 1)
     print('leader according to marathon API: {}'.format(current_leader))
     assert original_leader != current_leader
+    return current_leader
 
 
 def marathon_leadership_changed(original_leader):
```
jeschkies: Nit: Could we name this either `assert_marathon_leadership_changed` or have it return a boolean and then `assert marathon_leadership_changed(original_leader)`.

tests/system/common.py
jeschkies: This is a little confusing code. Isn't `marathon_leadership_changed` calling the DNS as well?

alenkacz: well the thing is there are multiple masters and multiple mesosDNS instances. By querying multiple times we might hit different masters. On every master, mesosDNS is updated every 30s but this does not have to be synchronized in between masters...

another approach might be to explicitly wait for the information being propagated to all masters... 

but we are just trying to verify that the election is finished so maybe the fact that it got propagated to one master is enough from the test perspective... (but still different masters might return different info at that point...)

tests/system/common.py
alenkacz: good point, I changed that, thank you

alenkacz:
please do not merge this yet... locally shakedown is telling me this https://jira.mesosphere.com/browse/MARATHON-7980 so I have to figure out what it means...

PR #5834, opened 2017-12-18
Contravariance Security fix

kensipe:
Contravariance Security fix

Summary:
Please read the desc of https://jira.mesosphere.com/browse/MARATHON-7974

I don't know if we will keep the Contravariance `[-R]` on Auth... however this definitely showed 5 locations in Marathon where security just plain didn't work.   The typing wasn't changed and there were no tests to confirm.     This PR fixes all but 1 issue which requires someone with more context of the code to fix.

*Needs* Fix needed for TaskController.  Line 129 `authorizer.isAuthorized(identity, UpdateRunSpec, id)`   can NOT be correct.   The "resource" is a `PathId` and needs to be a `RunSpec`.    Also it appears that it should have the `DeleteRunSpec` permission to me.   @alenkacz you were the last to write this code... could you that a look and update this branch? 

JIRA issues: MARATHON-7974

### DIFF
--- a/plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/AuthorizedAction.scala
+++ b/plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/AuthorizedAction.scala
@@ -9,7 +9,7 @@ import mesosphere.marathon.plugin.{ Group, RunSpec }
   *
   * @tparam R the type of the resource.
   */
-sealed trait AuthorizedAction[+R]
+sealed trait AuthorizedAction[-R]
 
 /**
   * The following objects will be passed to the Authorizer when an action affects an application, in order to identify
--- a/src/main/scala/mesosphere/marathon/api/GroupApiService.scala
+++ b/src/main/scala/mesosphere/marathon/api/GroupApiService.scala
@@ -33,11 +33,9 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       case Some(version) =>
         val targetVersion = Timestamp(version)
         groupManager.group(group.id, targetVersion)
-          .filter(isAllowed(ViewGroup, _))
-          .map(g => {
-            val existingGroup = g.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion"))
-            Some(rootGroup.putGroup(existingGroup, newVersion))
-          })
+            .map(_.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion")))
+          .filter(checkAuthorizationOrThrow(ViewGroup, _))
+          .map(g => Some(rootGroup.putGroup(g, newVersion)))
       case None => Future.successful(None)
     }
 
@@ -60,10 +58,6 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
     await(revertToOlderVersion)
       .orElse(scaleChange)
       .getOrElse(createOrUpdateChange)
-  }
-
-  def isAllowed[Resource](action: AuthorizedAction[Resource], mayResource: Option[Resource])(implicit identity: Identity, authorizer: Authorizer): Boolean = {
-    checkAuthorizationOrThrow(action, mayResource.get)(identity, authorizer)
   }
 
   private def checkAuthorizationOrThrow[Resource](action: AuthorizedAction[Resource], resource: Resource)(implicit identity: Identity, authorizer: Authorizer): Boolean = {
--- a/src/main/scala/mesosphere/marathon/api/GroupApiService.scala
+++ b/src/main/scala/mesosphere/marathon/api/GroupApiService.scala
@@ -33,7 +33,7 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       case Some(version) =>
         val targetVersion = Timestamp(version)
         groupManager.group(group.id, targetVersion)
-            .map(_.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion")))
+          .map(_.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion")))
           .filter(checkAuthorizationOrThrow(ViewGroup, _))
           .map(g => Some(rootGroup.putGroup(g, newVersion)))
       case None => Future.successful(None)
--- a/src/main/scala/mesosphere/marathon/api/GroupApiService.scala
+++ b/src/main/scala/mesosphere/marathon/api/GroupApiService.scala
@@ -33,7 +33,7 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       case Some(version) =>
         val targetVersion = Timestamp(version)
         groupManager.group(group.id, targetVersion)
-          .filter(checkAuthorizationOrThrow(ViewGroup, _))
+          .filter(isAllowed(ViewGroup, _))
           .map(g => {
             val existingGroup = g.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion"))
             Some(rootGroup.putGroup(existingGroup, newVersion))
@@ -52,7 +52,7 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       val updatedGroup: Group = Raml.fromRaml(
         GroupConversion(groupUpdate, group, newVersion) -> appConversionFunc)
 
-      if (maybeExistingGroup.isEmpty) checkAuthorizationOrThrow(CreateRunSpec, updatedGroup)
+      if (maybeExistingGroup.isEmpty) checkAuthorizationOrThrow(CreateGroup, updatedGroup)
 
       rootGroup.putGroup(updatedGroup, newVersion)
     }
@@ -60,6 +60,10 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
     await(revertToOlderVersion)
       .orElse(scaleChange)
       .getOrElse(createOrUpdateChange)
+  }
+
+  def isAllowed[Resource](action: AuthorizedAction[Resource], mayResource: Option[Resource])(implicit identity: Identity, authorizer: Authorizer): Boolean = {
+    checkAuthorizationOrThrow(action, mayResource.get)(identity, authorizer)
   }
 
   private def checkAuthorizationOrThrow[Resource](action: AuthorizedAction[Resource], resource: Resource)(implicit identity: Identity, authorizer: Authorizer): Boolean = {
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -425,7 +425,7 @@ class AppsController(
   private def getVersion(appId: PathId, version: Timestamp)(implicit identity: Identity): Route = {
     onSuccess(groupManager.appVersion(appId, version.toOffsetDateTime)) {
       case Some(app) =>
-        authorized(ViewRunSpec, app, Rejections.EntityNotFound.noApp(appId)).apply {
+        authorized(ViewRunSpec, app).apply {
           complete(app.toRaml)
         }
       case None =>
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
@@ -36,6 +36,7 @@ import mesosphere.marathon.state._
 import mesosphere.marathon.stream.Sink
 import play.api.libs.json.Json
 import PathMatchers.forceParameter
+import mesosphere.marathon.plugin.auth.AuthorizedResource.SystemConfig
 
 import scala.async.Async._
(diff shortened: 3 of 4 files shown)


src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -52,7 +52,7 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       val updatedGroup: Group = Raml.fromRaml(
         GroupConversion(groupUpdate, group, newVersion) -> appConversionFunc)
 
-      if (maybeExistingGroup.isEmpty) checkAuthorizationOrThrow(CreateRunSpec, updatedGroup)
+      if (maybeExistingGroup.isEmpty) checkAuthorizationOrThrow(CreateGroup, updatedGroup)
```
kensipe: This was just previously wrong... it is a Group not a RunSpec.

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -62,6 +62,10 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       .getOrElse(createOrUpdateChange)
   }
 
+  def isAllowed[Resource](action: AuthorizedAction[Resource], mayResource: Option[Resource])(implicit identity: Identity, authorizer: Authorizer): Boolean = {
+    checkAuthorizationOrThrow(action, mayResource.get)(identity, authorizer)
+  }
```
kensipe: this may seem unusual.. however the original code below `checkAuthorizationOrThrow` was called by 2 functions... 1 was line 55 which is for a Group.   The other was line 36 for an `Option[Group]' which didn't work

kensipe: there is probably a more "clever" way to solve this... whatever it is... lets do it

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -144,7 +145,7 @@ class AppsController(
         case None =>
           reject(Rejections.EntityNotFound.noApp(appId))
         case Some(info) =>
-          authorized(ViewResource, info.app).apply {
+          authorized(ViewResource, SystemConfig).apply {
```
kensipe: This is not tied to the app... this is permission of `/info` which is part of SystemConfig.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -410,7 +411,7 @@ class AppsController(
 
   private def listVersions(appId: PathId)(implicit identity: Identity): Route = {
     val versions = groupManager.appVersions(appId).runWith(Sink.seq)
-    authorized(ViewRunSpec, groupManager.app(appId), Rejections.EntityNotFound.noApp(appId)).apply {
+    authorized(ViewRunSpec, groupManager.runSpec(appId).get, Rejections.EntityNotFound.noApp(appId)).apply {
```
kensipe: has to be on the runspec.. not the app

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
alenkacz: Thanks for fixing this!! Not sure my way is more "clever" but check it out and tell me what you think... :)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
alenkacz: I think this should be "authorized(ViewRunSpec, info.app)" - I think it is tight to an app.

Correct me if I am wrong...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
alenkacz: appdefinition is a subtype of runspec so I think we can use appDefinition here...

I will edit this one and commit and you can review if you think it is OK...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
```diff
@@ -411,10 +411,14 @@ class AppsController(
 
   private def listVersions(appId: PathId)(implicit identity: Identity): Route = {
     val versions = groupManager.appVersions(appId).runWith(Sink.seq)
-    authorized(ViewRunSpec, groupManager.runSpec(appId).get, Rejections.EntityNotFound.noApp(appId)).apply {
-      onSuccess(versions) { versions =>
-        complete(VersionList(versions))
-      }
+    groupManager.app(appId) match {
+      case Some(app) =>
+        authorized(ViewRunSpec, app, Rejections.EntityNotFound.noApp(appId)).apply {
```
kensipe: @alenkacz  this is easier to read my opinion!

Question tho:   `case None` is entity not found... how do we have `case Some(app) with an `EntityNotFound`?  Should that be a not authorized?

kensipe:
reviewing now...

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
alenkacz: Good question, I don't know. I think it is like this because this is just a rewrite of old resources and I see this logic there as well. https://github.com/mesosphere/marathon/blob/e07df868b0b620e355ffdc498296c06a3f3b8a94/src/main/scala/mesosphere/marathon/api/v2/AppVersionsResource.scala#L45

maybe @meln1k will know more since he is author of this...

I personally would be for changing this to return not authorized here instead of not found to be consistent with the rest...

kensipe:
@alenkacz lgtm

src/main/scala/mesosphere/marathon/api/akkahttp/v2/AppsController.scala
meln1k: I think that `NotFound` case will be covered by `PathMatcher`. So we probably should use not authorized in this case.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/AuthorizedAction.scala
```diff
@@ -9,7 +9,7 @@ import mesosphere.marathon.plugin.{ Group, RunSpec }
   *
   * @tparam R the type of the resource.
   */
-sealed trait AuthorizedAction[+R]
+sealed trait AuthorizedAction[-R]
```
jeschkies: Why do we want that?

This means given `A <: B` that `AuthorizedAction[A] :> AuthorizedAction[B]`. So If `A` is a sub resource of `B` any action of `A` is allowed on `B` but not any action on `B` is allowed on `A`. Did I get this correct?

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
```diff
@@ -33,11 +33,9 @@ class GroupApiService(groupManager: GroupManager)(implicit authorizer: Authorize
       case Some(version) =>
         val targetVersion = Timestamp(version)
         groupManager.group(group.id, targetVersion)
+          .map(_.getOrElse(throw new IllegalArgumentException(s"Group ${group.id} not available in version $targetVersion")))
```
jeschkies: What motivated this change?

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -100,7 +100,8 @@ class TasksController(
         tryParseTaskIds(taskIds) match {
           case Left(rejection) => reject(rejection)
           case Right(instanceIdsToAppId) =>
-            if (!isAuthorized(instanceIdsToAppId.values)) {
+            val affectedApps = instanceIdsToAppId.values.flatMap(appId => groupManager.app(appId))(collection.breakOut)
```
jeschkies: What is the type of `affectedApps`? I don't quite get what `latMap(appId => groupManager.app(appId))` is doing. Also, I'm unsure if `collection.breakOut` works if there is no type on `affectedApps`. I'm probably missing something.

plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/AuthorizedAction.scala
kensipe: The biggest issue is that we saw a lot of code that was failing (Ex: https://github.com/dcos/metronome/pull/161/files) 
The issue with that code is `case Some(run) => request.authorized(ViewRunSpec, run, Ok(run))` passed `run` which was of type `StartedJobRun`... which is incorrect.  When I investigated... any value was possible to pass in here... because with `[+R]` the `AuthorizationAction[A]` will look for an `A` type in the superclass hierarchy...  AND type inferencing... it went all the way up to `Any`.      We had a 1 place in metronome and several across several versions of Marathon where the wrong type of `R` was passed in for the AuthorizedAction type.   `-R` will allow that type or a subtype.  This is useful in that a compile time failure will occur if the wrong type is used (we had several places looking for ViewRunSpec auth or a Group action which is obviously incorrect).

src/main/scala/mesosphere/marathon/api/GroupApiService.scala
kensipe: this is consistent with the previously expected code but that code was incorrect.
AuthorizedAction[R] for `checkAuthorizationOrThrow[Resource](action: AuthorizedAction[Resource], resource: Resource)` is `Resource`.   `Resource` is expected to be of type `Group` but in this case prior to the change it was of type `Option[Group]`.   Once Auth was enforced properly it was required to map to the type from the option.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
kensipe: @alenkacz is probably best to answer that..

kensipe: well.. I expect her to give an answer to the breakOut... the type is `Seq[AppDefinition]`
as it is used in private def kill function on L204

plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/AuthorizedAction.scala
timcharper: Yeah, Covariance definitely seems wrong. An `AuthorizedAction[RunSpec]` can be used with a `Group` (RunSpec and Group both implement Any). The likely intention was the other way: an `AuthorizedAction[RunSpec]` can be used for anything that implements `RunSpec` (which the contra-variance allows)

Relevant is the type spec for the method used here:

```
authorizer.isAuthorized(identity, UpdateRunSpec, app)
// isAuthorized[Resource](Identity, AuthorizedAction[Resource], Resource)
```

(hopefully this helps shed light on why covariance is wrong)

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
timcharper: `collection.breakOut` is for when you want the result of some operation to turn in to some other type. 

Recommend:
```
val affectedApps: Seq[AppDefinition] = instanceIdsToAppId.values.flatMap(appId => groupManager.app(appId))(collection.breakOut)
```

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
```diff
@@ -125,7 +126,12 @@ class TasksController(
       .groupBy(instance => instance.instanceId.runSpecId)
       .map { case (appId, instances) => appId -> instances.to[Seq] }(collection.breakOut))
   }
-  private def isAuthorized(appIds: Iterable[PathId])(implicit identity: Identity): Boolean = appIds.forall(id => authorizer.isAuthorized(identity, UpdateRunSpec, id))
+
+  private def isAuthorized(apps: Iterable[AppDefinition])(implicit identity: Identity): Boolean =
+    apps
+      .map(app => authorizer.isAuthorized(identity, UpdateRunSpec, app))
```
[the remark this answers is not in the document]
timcharper: Oh... the method is used just once? Potentially consider moving inline?

Replace:
```
if (!isAuthorized(affectedApps)) {
```

with

```
if (affectedApps.forall { app => authorizer.isAuthorized(identity, UpdateRunSpec, app)) })
```

plugin-interface/src/main/scala/mesosphere/marathon/plugin/auth/AuthorizedAction.scala
jeschkies: Not sure I get it. I understand the type inference but I'd rather start at the types themselves. Just changing the type hierarchy because the type inference was wrong seems not the right reason to change this as one can always help the inference by providing types explicitly.

src/main/scala/mesosphere/marathon/api/akkahttp/v2/TasksController.scala
jeschkies: I know what `breakOut` does but without a type annotation for `affectedApps` this becomes very hard to read. Unfortunately this has been merged. `val affectedApps: Seq[AppDefinition]` would have been cleaner IMHO.

PR #5837, opened 2017-12-19
[WIP] Remove JSON Schema in favor of RAML

timcharper:
JIRA issues: MARATHON-7978


timcharper:
@jdef anything else of which you're aware that we need to do for this?

jdef:
isn't there a unit test (or more than one) that attempts to load the schema file and test various properties?

jdef:
for example: https://github.com/mesosphere/marathon/blob/15e85e5f30c549aa975ac852597cef9de00daa67/src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionSchemaJSONTest.scala

jdef:
some things to double-check:
- [ ] DC/OS CLI dependency on schema (if any)
- [ ] DC/OS UI dependency on schema (if any)
- [ ] Marathon Java Client dependency on schema (if any)

jdef:
To which endpoint are you referring?

On Wed, Dec 20, 2017 at 8:28 AM, Karsten Jeschkies <notifications@github.com

meln1k:
I agree with @jeschkies regarding the backwards compatibility.

meichstedt:
I agree that removing the endpoint and the schemas should be considered a breaking change, we don't know who's relying on it. We could mitigate that by redirecting to the 1.5 content on GH? 😬 

However, the raml is the source of truth and AFAIK we don't even have tests that verify the compatibility between the two, so it is also questionable how much value the endpoint and the schema are actually providing.

timcharper:
Yes, I had all of the same concerns. It's doubtful that it is being used. In the JIRA I raised this question.

After sleeping on it for a night I feel worse about it. We should deprecate the endpoint with a flag, and then remove in a subsequent major version.

timcharper:
I have a follow up PR that uses the newly introduced deprecation mechanism, which will allow us to phase this out more gracefully:

https://github.com/mesosphere/marathon/pull/6280

PR #5844, opened 2017-12-20
AppResidency fields relaunchEscalationTimeoutSeconds and taskLostBehavior have no effect

meln1k:
Summary: appResidency is deprecated and no longer used

JIRA issues: MARATHON-7968


mesos-simulation/src/test/scala/mesosphere/mesos/scale/Metric.scala
```diff
@@ -23,18 +23,18 @@ case class Counter(name: String, count: Int) extends Metric {
...
-  mean: Double,
-  min: Double,
-  p50: Double,
-  p75: Double,
-  p95: Double,
-  p98: Double,
-  p99: Double,
-  p999: Double,
-  stddev: Double) extends Metric
+    name: String,
```
jeschkies: What happened here? 🙂

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -36,14 +36,13 @@ object AppHelpers {
     * "update" operation into a "create" operation. This helper func facilitates that.
     */
   def withoutPriorAppDefinition(update: raml.AppUpdate, appId: PathId): raml.App = {
-    val selectedStrategy = AppConversion.ResidencyAndUpgradeStrategy(
-      residency = update.residency.map(Raml.fromRaml(_)),
+    val selectedStrategy = AppConversion.UpgradeStrategyConverter(
```
jeschkies: So `UpgradeStrategyConverter` is equivalent to `ResidencyAndUpgradeStrategy`?

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -124,8 +121,7 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
     * has a `versionInfo` constructed from `OnlyVersion(app.version)`.
     */
   implicit val appRamlReader: Reads[App, AppDefinition] = Reads[App, AppDefinition] { app =>
-    val selectedStrategy = ResidencyAndUpgradeStrategy(
-      app.residency.map(Raml.fromRaml(_)),
+    val selectedStrategy = UpgradeStrategyConverter(
```
jeschkies: Could you add a type to `selectedStrategy`?

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -308,6 +303,20 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       ))
       else None
 
+    val unreachableStrategy = service.when(_.hasUnreachableStrategy, _.getUnreachableStrategy.toRaml).orElse(App.DefaultUnreachableStrategy)
+
+    val hasPersistentVolumes = service.hasContainer && service.getContainer.getVolumesList.exists(_.hasPersistent)
```
jeschkies: Where is this coming from?

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -308,6 +303,20 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
...
+
+    val residency = if (hasPersistentVolumes || service.hasResidency) {
+      unreachableStrategy.flatMap {
```
jeschkies: Why not match as you've done above? Is `unreachableStrategy` an `Option`? That seems weird.

Also, why not factor this out since it's used in several places?

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -308,6 +303,20 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       ))
       else None
 
+    val unreachableStrategy = service.when(_.hasUnreachableStrategy, _.getUnreachableStrategy.toRaml).orElse(App.DefaultUnreachableStrategy)
```
jeschkies: What is the type?

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -413,25 +422,19 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
 
 object AppConversion extends AppConversion {
 
-  case class ResidencyAndUpgradeStrategy(residency: Option[Residency], upgradeStrategy: state.UpgradeStrategy)
-
-  object ResidencyAndUpgradeStrategy {
+  object UpgradeStrategyConverter {
```
jeschkies: Ah, now I understand. Great change.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -115,8 +113,6 @@ case class AppDefinition(
 
   val portNumbers: Seq[Int] = portDefinitions.map(_.port)
 
-  override val isResident: Boolean = residency.isDefined
```
jeschkies: 🤨 don't we still have resident instances?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -253,7 +245,7 @@ case class AppDefinition(
       if (proto.hasUnreachableStrategy)
         UnreachableStrategy.fromProto(proto.getUnreachableStrategy)
       else
-        UnreachableStrategy.default(residencyOption.isDefined)
+        UnreachableStrategy.default(isResident)
```
jeschkies: Where is this coming from since you've removed the original. Is it because the original was overridden?

src/main/scala/mesosphere/marathon/state/RunSpec.scala
```diff
@@ -42,8 +42,8 @@ trait RunSpec extends plugin.RunSpec {
   val version: Timestamp
   val resources: Resources
   val backoffStrategy: BackoffStrategy
-  val upgradeStrategy: UpgradeStrategy
   def isResident: Boolean = persistentVolumes.nonEmpty
+  val upgradeStrategy: UpgradeStrategy
```
jeschkies: what is the reason for changing the order?

src/main/scala/mesosphere/marathon/stream/UriIO.scala
```diff
@@ -128,7 +128,7 @@ object UriIO extends StrictLogging {
           .map(creds => new BasicAWSCredentials(creds.getAWSAccessKeyId, creds.getAWSSecretKey))
       }
       fromURL.orElse(fromProviderChain).getOrElse {
-        S3Settings().credentialsProvider.getCredentials
+        S3Settings()(actorSystem).credentialsProvider.getCredentials
```
jeschkies: What's the reason for this change?

src/test/scala/mesosphere/marathon/api/v2/json/AppUpdateTest.scala
```diff
@@ -408,40 +380,6 @@ class AppUpdateTest extends UnitTest with ValidationTestLike {
       )), createdViaUpdate.container)
     }
 
-    "empty app persists existing residency" in {
```
jeschkies: Hm, I think we still want to test whether the API accepts specs that do have the fields set.

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
meln1k: After residency became a property that depends on existence of persistent volumes, we no longer need to pass it to the ResidencyAndUpgradeStrategy. The name became a bit obsolete, that's why I renamed it.

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
meln1k: That's how we can find if we have persistent volumes. You can check the persistent volumes field in the RunSpec for more details.

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
meln1k: The problem here is that different classes are used. Above we use Raml, and here we use classes from the state package.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
meln1k: well, now we have this at the RunSpec level.

src/main/proto/marathon.proto
```diff
@@ -132,7 +132,7 @@ message ServiceDefinition {
   optional int64 last_scaling_at = 23;
   optional int64 last_config_change_at = 24;
   optional ObsoleteIpAddress OBSOLETE_ipAddress = 25; // deprecated since 1.5
-  optional ResidencyDefinition residency = 26;
+  optional ResidencyDefinition residency = 26 [deprecated = true];
```
ichernetsky: I remember that the way fields of `marathon.proto` are way different from the approach Protobuf docs suggest. @jdef is probably the one to talk to. Basically, it is about avoiding `[deprecated = true]`, and renaming fields instead (something like `residency -> DEPRECATED_residency`). @timcharper it would be nice to have your take on this.

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -64,7 +60,15 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       ports = None, // deprecated field
       portDefinitions = if (app.networks.hasNonHostNetworking) None else Some(app.portDefinitions.toRaml),
       readinessChecks = app.readinessChecks.toRaml,
-      residency = app.residency.toRaml,
+      residency = if (app.isResident)
+        app.unreachableStrategy match {
+        case state.UnreachableDisabled => None
```
ichernetsky: I think that is the unreachable strategy is disabled, then `residency.taskLostBehavior` should be to `WAIT_FOREVER`, shouldn't it?

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -308,6 +303,20 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       ))
       else None
 
+    val unreachableStrategy = service.when(_.hasUnreachableStrategy, _.getUnreachableStrategy.toRaml).orElse(App.DefaultUnreachableStrategy)
+
+    val hasPersistentVolumes = service.hasContainer && service.getContainer.getVolumesList.exists(_.hasPersistent)
+
+    val residency = if (hasPersistentVolumes || service.hasResidency) {
+      unreachableStrategy.flatMap {
+        case raml.UnreachableDisabled(_) => None
```
ichernetsky: The same here, shouldn't it be `residency.taskLostBehavior = WAIT_FOREVER`, so that we stay backward compatible, until `residency` is removed, not just deprecated?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -482,7 +472,7 @@ object AppDefinition extends GeneralPurposeCombinators {
 
   private val complyWithResidencyRules: Validator[AppDefinition] =
     isTrue("AppDefinition must contain persistent volumes and define residency") { app =>
-      !(app.residency.isDefined ^ app.persistentVolumes.nonEmpty)
+      !(app.isResident ^ app.persistentVolumes.nonEmpty)
```
ichernetsky: AFAIU, `isResident = persistentVolumes.nonEmpty`, therefore, the whole expression always evaluates to `true`, doesn't it?

jdef:
For consistency, protobuf fields no longer in use should renamed
OBSOLETE-xxx instead of being marked deprecated. Obsolete fields are
generally only referred to in migration code.

PR #5849, opened 2017-12-21
Changes to offer and reservation operations to support CSI volumes

ichernetsky:
- Reservation fields like `DiskInfo.Source.{id,metadata}` and `Resource.provider_id` have to be preserved when dealing with offer operations.
- Parametrize `VolumeWithMount` type in order to simplify code in a few places.
- Offer operations must be grouped by resource provider ID, otherwise operations get dropped by Mesos.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
@@ -69,6 +69,14 @@ object InstanceOp {
       resources: Seq[MesosProtos.Resource],
       oldInstance: Option[Instance] = None) extends InstanceOp {
 
+    private def getProviderId(resource: MesosProtos.Resource): Option[String] = {
+      if (resource.hasProviderId && resource.getProviderId.hasValue) {
+        Some(resource.getProviderId.getValue)
+      } else {
+        None
+      }
+    }
+
     override lazy val offerOperations: Seq[MesosProtos.Offer.Operation] = {
       val (withDisk, withoutDisk) = resources.partition(_.hasDisk)
       val reservationsForDisks = withDisk.map { resource =>
@@ -86,36 +94,33 @@ object InstanceOp {
         resourceBuilder.build()
       }
 
-      val maybeDestroyVolumes: Option[MesosProtos.Offer.Operation] =
-        if (withDisk.nonEmpty) {
+      val maybeDestroyVolumes: Seq[MesosProtos.Offer.Operation] =
+        withDisk.groupBy(getProviderId).values.map { withDiskResources =>
           val destroyOp =
             MesosProtos.Offer.Operation.Destroy.newBuilder()
-              .addAllVolumes(withDisk.asJava)
+              .addAllVolumes(withDiskResources.asJava)
 
-          val op =
-            MesosProtos.Offer.Operation.newBuilder()
-              .setType(MesosProtos.Offer.Operation.Type.DESTROY)
-              .setDestroy(destroyOp)
+          MesosProtos.Offer.Operation.newBuilder()
+            .setType(MesosProtos.Offer.Operation.Type.DESTROY)
+            .setDestroy(destroyOp)
+            .build()
+        }.to[Seq]
+
+      val maybeUnreserve: Seq[MesosProtos.Offer.Operation] =
+        if (withDisk.nonEmpty || reservationsForDisks.nonEmpty) {
+          (withoutDisk ++ reservationsForDisks).groupBy(getProviderId).values.map { resources =>
+            val unreserveOp = MesosProtos.Offer.Operation.Unreserve.newBuilder()
+              .addAllResources(resources.asJava)
               .build()
 
-          Some(op)
-        } else None
-
-      val maybeUnreserve: Option[MesosProtos.Offer.Operation] =
-        if (withDisk.nonEmpty || reservationsForDisks.nonEmpty) {
-          val unreserveOp = MesosProtos.Offer.Operation.Unreserve.newBuilder()
-            .addAllResources(withoutDisk.asJava)
-            .addAllResources(reservationsForDisks.asJava)
-            .build()
-          val op =
             MesosProtos.Offer.Operation.newBuilder()
               .setType(MesosProtos.Offer.Operation.Type.UNRESERVE)
               .setUnreserve(unreserveOp)
               .build()
-          Some(op)
-        } else None
+          }.to[Seq]
+        } else Seq.empty
 
-      Seq(maybeDestroyVolumes, maybeUnreserve).flatten
+      maybeDestroyVolumes ++ maybeUnreserve
     }
 
     override def applyToOffer(offer: MesosProtos.Offer): MesosProtos.Offer =
--- a/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
@@ -6,6 +6,7 @@ import mesosphere.marathon.core.instance.update.InstanceUpdateOperation
 import mesosphere.marathon.stream.Implicits._
 import mesosphere.marathon.tasks.ResourceUtil
 import mesosphere.marathon.tasks.ResourceUtil.RichResource
+import mesosphere.mesos.protos.ResourceProviderID
 import org.apache.mesos.{ Protos => MesosProtos }
 
 /**
@@ -69,14 +70,6 @@ object InstanceOp {
       resources: Seq[MesosProtos.Resource],
       oldInstance: Option[Instance] = None) extends InstanceOp {
 
-    private def getProviderId(resource: MesosProtos.Resource): Option[String] = {
-      if (resource.hasProviderId && resource.getProviderId.hasValue) {
-        Some(resource.getProviderId.getValue)
-      } else {
-        None
-      }
-    }
-
     override lazy val offerOperations: Seq[MesosProtos.Offer.Operation] = {
       val (withDisk, withoutDisk) = resources.partition(_.hasDisk)
       val reservationsForDisks = withDisk.map { resource =>
@@ -94,8 +87,10 @@ object InstanceOp {
         resourceBuilder.build()
       }
 
-      val maybeDestroyVolumes: Seq[MesosProtos.Offer.Operation] =
-        withDisk.groupBy(getProviderId).values.map { withDiskResources =>
+      val maybeDestroyVolumes: Seq[MesosProtos.Offer.Operation] = {
+        // Mesos requires that there is an operation per each resource provider ID
+        val volumesToDestroyGroupedByProviderId = withDisk.groupBy(ResourceProviderID.fromResourceProto).values
+        volumesToDestroyGroupedByProviderId.map { withDiskResources =>
           val destroyOp =
             MesosProtos.Offer.Operation.Destroy.newBuilder()
               .addAllVolumes(withDiskResources.asJava)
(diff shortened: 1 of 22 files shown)


src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -47,78 +47,99 @@ class OfferOperationFactory(
...
-
-      Mesos.Resource.newBuilder(resource)
-        .setRole(role)
-        .setReservation(reservation)
-        .build()
+  private def getProviderId(resource: Mesos.Resource): Option[String] = {
+    if (resource.hasProviderId && resource.getProviderId.hasValue) {
+      Some(resource.getProviderId.getValue)
+    } else {
+      None
```
kensipe: this pattern seems to be repeated...  should we have a common function?

ichernetsky: Yes, I mentioned it in the slack channel. I will do it, but I just don't have time for it right, in hope that we land it today.

ichernetsky: I won't forget about it :)

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryHelper.scala
```diff
@@ -78,13 +78,11 @@ class InstanceOpFactoryHelper(
     reservationLabels: ReservationLabels,
     newState: InstanceUpdateOperation.Reserve,
     resources: Seq[Mesos.Resource],
-    localVolumes: Seq[(DiskSource, LocalVolume)]): InstanceOp.ReserveAndCreateVolumes = {
+    localVolumes: Seq[(Option[String], DiskSource, LocalVolume)]): InstanceOp.ReserveAndCreateVolumes = {
```
alenkacz: I would personally go with a case class here since we already have a tuple of three items and it is hard to get what is what...

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -315,13 +315,15 @@ class InstanceOpFactoryImpl(
     offer: Mesos.Offer,
     resourceMatch: ResourceMatcher.ResourceMatch): InstanceOp = {
 
-    val localVolumes: Seq[(DiskSource, Task.LocalVolume)] =
+    val localVolumes: Seq[(Option[String], DiskSource, Task.LocalVolume)] =
       resourceMatch.localVolumes.map {
-        case (source, volume, mount) =>
-          (source, Task.LocalVolume(Task.LocalVolumeId(runSpec.id, volume, mount), volume, mount))
+        case (providerId, source, volumeWithMount) =>
```
alenkacz: for better readability I would go with this

case (providerId, source, VolumeWithMount(volume, mount))

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -315,13 +315,15 @@ class InstanceOpFactoryImpl(
...
+          (providerId, source, Task.LocalVolume(
+            Task.LocalVolumeId(runSpec.id, volumeWithMount.volume, volumeWithMount.mount),
+            volumeWithMount.volume, volumeWithMount.mount))
       }
 
-    val persistentVolumeIds = localVolumes.map { case (_, localVolume) => localVolume.id }
+    val persistentVolumeIds = localVolumes.map { case (_, _, localVolume) => localVolume.id }
```
alenkacz: if you go with case class even this will be much more readable from my point of view :)

src/main/scala/mesosphere/marathon/core/launcher/InstanceOp.scala
```diff
@@ -86,36 +94,33 @@ object InstanceOp {
...
 
-      val maybeUnreserve: Option[MesosProtos.Offer.Operation] =
+      val maybeUnreserve: Seq[MesosProtos.Offer.Operation] =
         if (withDisk.nonEmpty || reservationsForDisks.nonEmpty) {
-          val unreserveOp = MesosProtos.Offer.Operation.Unreserve.newBuilder()
-            .addAllResources(withoutDisk.asJava)
-            .addAllResources(reservationsForDisks.asJava)
-            .build()
-          val op =
+          (withoutDisk ++ reservationsForDisks).groupBy(getProviderId).values.map { resources =>
```
alenkacz: it took me some time to actually understand what is going on here, that we actually need to group resources per provider id because mesos required that...

I actually understood after reading your comment in OfferOperationFactory.scala

Maybe we can change this code to use some named variables or at least add some comments to make it more obvious for the reader?

Maybe we can separate the first step (up to .values) assign it to a variable and add a comment...

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -47,78 +47,99 @@ class OfferOperationFactory(
...
+        .build()
+    }.to[Seq]
+  }
+
+  def createVolumes(
+    reservationLabels: ReservationLabels,
+    localVolumes: Seq[(Option[String], DiskSource, LocalVolume)]): Seq[Mesos.Offer.Operation] = {
+
+    // Mesos requires that there is an operation per each resource provider ID
+    localVolumes.groupBy(_._1).values.map { localVolumes =>
```
alenkacz: I personally don't like _._1 - try to use case instead to make it more readable?

Well actually if you get rid of the tuple as I wrote in a different comment this would be solved :)

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -107,7 +106,9 @@ object HostVolume {
   }
 }
 
-case class DiskSource(diskType: DiskType, path: Option[String]) {
+case class DiskSource(
+    diskType: DiskType, path: Option[String],
+    id: Option[String], metadata: Map[String, String], profileName: Option[String]) {
   if (diskType == DiskType.Root)
     require(path.isEmpty, "Path is not allowed for diskType")
   else
```
alenkacz: maybe we should extend the toString method to use also the new fields added?

alenkacz: maybe we should extend the toString method to use also the new fields added?

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -129,24 +130,33 @@ case class DiskSource(diskType: DiskType, path: Option[String]) {
...
 
 object DiskSource {
-  val root = DiskSource(DiskType.Root, None)
+  val root = DiskSource(DiskType.Root, None, None, Map.empty, None)
+
   @SuppressWarnings(Array("OptionGet"))
   def fromMesos(source: Option[Source]): DiskSource = {
     val diskType = DiskType.fromMesosType(source.map(_.getType))
+    val id = source.flatMap(s => if (s.hasId) Some(s.getId) else None)
+    val metadata = source.map {
```
alenkacz: what do you think about

````
val metadata = source match {
      case Some(s) if s.hasMetadata => s.getMetadata.fromProto
      case _ => Map.empty[String, String]
    }
````

File: src/main/scala/mesosphere/mesos/ResourceMatcher.scala [hunk 9]
```diff
@@ -148,27 +148,25 @@ object ResourceMatcher extends StrictLogging {
...
-          val indexedVolumes = volumes.zipWithIndex.map { case (v, i) => v.copy(name = Some(i.toString)) }
-          val indexedMounts = mounts.zipWithIndex.map { case (m, i) => m.copy(volumeName = Some(i.toString)) }
-          (indexedVolumes, indexedMounts)
-        } else {
-          (volumes, mounts)
-        }
+      val volumesWithMounts = runSpec match {
+        case _: AppDefinition =>
+          volumes.zip(mounts).map {
+            case (volume, mount) => VolumeWithMount(volume, mount)
```
alenkacz: why don't we need the temporary names anymore?

src/test/scala/mesosphere/marathon/api/validation/RunSpecValidatorTest.scala
```diff
@@ -676,19 +676,19 @@ class RunSpecValidatorTest extends UnitTest with ValidationTestLike {
       )
 
       // scalastyle:off magic.number
-      def hostVolume(hostPath: String = "/etc/foo", mountPath: String = "/test", readOnly: Boolean = false): VolumeWithMount =
+      def hostVolume(hostPath: String = "/etc/foo", mountPath: String = "/test", readOnly: Boolean = false): VolumeWithMount[Volume] =
```
alenkacz: shouldn't this be HostVolume not Volume?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryHelper.scala
meln1k: I agree, case class would be really nice here :)

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -72,24 +73,22 @@ object VolumeMount {
   def readOnlyToProto(readOnly: Boolean): Mode = if (readOnly) Mode.RO else Mode.RW
 }
 
-case class VolumeWithMount(volume: Volume, mount: VolumeMount)
+case class VolumeWithMount[+A <: Volume](volume: A, mount: VolumeMount)
```
meln1k: Is it possible to simplify the type here?

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
ichernetsky: Done.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
ichernetsky: Done.

ichernetsky:
@alenkacz this is initial support for things explained to some extent in the following docs:

- https://docs.google.com/document/d/18OE7dDJAOvqS1kBRmya8VyyS8C6hhOoGXs77AkTpRYA
- https://docs.google.com/document/d/125YWqg_5BB5OY9a6M7LZcby5RSqBwo2PZzpVLuxYXh4
- https://docs.google.com/document/d/1WtaGoO5vkpL7Kk0Uxor0q10ySV4Y0Xx11PEjjnT00P0
- https://docs.google.com/document/d/1weGOCKQSkwtsXQS9nUEjUZqei4W8N87XmSXAI_hM-qM

ichernetsky:
More tests are coming.

src/main/scala/mesosphere/marathon/state/Volume.scala
alenkacz: yep, makes sense. Let's not put metadata there... thanks!

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -355,13 +352,13 @@ object ResourceMatcher extends StrictLogging {
         case nextAllocation :: restAllocations =>
           val (matcher, nextAllocationSize) = nextAllocation match {
             case Left(size) => ({ _: Protos.Resource => true }, size)
-            case Right(v) =>
+            case Right(vm) =>
```
alenkacz: I would do

```
case Right(VolumeWithMount(volume, _)) =>
```

PR #5850, opened 2017-12-21
Use Marathon event stream directly for SI tests.

jeschkies:
Summary:
Some system integration tests SSH into the cluster and call the Marathon
event API from there. This is not necessary. We can call it from the
outside and assert each event.

Depends on #5860.

JIRA issues: MARATHON-7964


jeschkies:
The info endpoint is failing.

ichernetsky:
@jeschkies it is `Caused by: $file.ci.utils$BuildException: 'flake8 --count --max-line-length=120 tests/system src/test/python' exited with 1`, AFAICT.

jeschkies:
> @jeschkies it is Caused by: $file.ci.utils$BuildException: 'flake8 --count --max-line-length=120 tests/system src/test/python' exited with 1, AFAICT.

Yep. The Python and thus flake8 version is out of date.

tests/Makefile
```diff
@@ -17,7 +17,7 @@ init:
 	pipenv install --dev --skip-lock
 
 build:
-	flake8 --count --max-line-length=120
+	flake8 --count --max-line-length=120 system
```
alenkacz: hood catch! 👍

Jenkinsfile.pr.si
```diff
@@ -21,6 +21,7 @@ node('py35') {
                 sh """git fetch origin pull/$params.Pull_Request_Id/head:$params.Pull_Request_Id"""
                 sh """git checkout $params.Pull_Request_Id"""
             }
+            sh """pip3 install requests"""
```
wavesoft: Is `requests` needed anywhere?

ami/install.bash
```diff
@@ -1,46 +1,52 @@
...
-        npm \
-        python3-pip \
-        rpm \
-        sbt
+        openjdk-8-jdk \
+        libssl-dev \
+        sbt \
+        zlib1g-dev
+
+# Download, compile and install Python 3.6.2
```
wavesoft: How come install python from sources and not using a binary release? (not that it's changing that much, just curious).

Jenkinsfile.pr.si
jeschkies: Yes. For Github statuses.

ami/install.bash
jeschkies: There is no Python 3.6 backport for Stretch at least I did not find any.

PR #5854, opened 2017-12-21
HOST networking breaks ports for applications created via PUT

meln1k:
Summary: change in the appConversion class for correct transformation of empty seq to option

JIRA issues: MARATHON-7969

### DIFF
--- a/src/test/scala/mesosphere/marathon/raml/AppConversionTest.scala
+++ b/src/test/scala/mesosphere/marathon/raml/AppConversionTest.scala
@@ -94,6 +94,15 @@ class AppConversionTest extends UnitTest with ValidationTestLike {
     behave like convertToRamlAndBack(hostApp)
     behave like convertToProtobufThenToRAML(hostApp)
 
+    "correctly handle empty port definitions" in {
+      Given("An app definition")
+      val ramlApp = hostApp.copy(portDefinitions = Seq.empty).toRaml[App]
+
+      When("The app definition is converted to an app")
+      Then("Port definitions must be parsed correctly")
+      ramlApp.portDefinitions shouldEqual None
+    }
+
     "convert legacy service definitions to RAML" in {
       val legacy = Protos.ServiceDefinition.newBuilder()
         .setId("/legacy")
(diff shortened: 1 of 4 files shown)


jdef:
I'd actually like to see a test case that exercises PUT in the manner suggested by the JIRA report, so that I understand the resulting application that Marathon ends up creating. The original JIRA talks about `PORT` and `portMappings` - it's not immediately clear to me how either of those are related to this PR.

jdef:
The sample JSON from the JIRA (when PUT'ing as a new app):
```json
{
  "id":"tomcat",
  "mem":512,
  "cpus":1.0,
  "instances":1,
  "container": {
    "type":"DOCKER",
    "docker": {
      "image":"tomcat:8.0",
      "network":"HOST"
    }
  }
}
```
... should yield an app w/ a single port definition: `{ "port": 0 }` -- that's the test that I'd like to see

meln1k:
@jdef the problem that we can only test it in integration test, because unit tests with PUT uses mocked implementation. I tested it manually and it works.

@timcharper do you think we should have an integration test here?

src/test/scala/mesosphere/marathon/raml/AppConversionTest.scala
```diff
@@ -94,6 +94,15 @@ class AppConversionTest extends UnitTest with ValidationTestLike {
     behave like convertToRamlAndBack(hostApp)
     behave like convertToProtobufThenToRAML(hostApp)
 
+    "correctly handle empty port definitions" in {
+      Given("An app definition")
```
timcharper: With such a small test, I'm inclined to believe that these given, then, when statements add very little value. Let's remove them?

timcharper: Also, in the links JIRA, it seems that port definitions wasn't specified at all. I think that the JIRA could be potentially clarified.

What we should really be testing is AppUpdate, right?

alenkacz:
@meln1k just out of curiosity why did we close this? I did not find info not here not in JIRA... :)

meln1k:
@alenkacz it turned out that fixing this bug will consume much more time that was initially planned.

jdef:
> it turned out that fixing this bug will consume much more time that was initially planned.

Can you please write up an RCA somewhere to help the next person that tackles this?

alenkacz:
@jdef @meln1k yeah that would be nice - I think JIRA is the right place for that...

PR #5855, opened 2017-12-22
Mark SI pull requests job as failed.

jeschkies:
Summary:
In some cases the launch would fail but we would not report it as such.

### DIFF
--- a/ci/si_pipeline.sh
+++ b/ci/si_pipeline.sh
@@ -63,5 +63,5 @@ case $CLUSTER_LAUNCH_CODE in
       ;;
   2) exit-as-unstable "Cluster launch failed.";;
   3) exit-as-unstable "Cluster did not start in time.";;
-  *) exit-as-unstalbe "Unknown error in cluster launch: $CLUSTER_LAUNCH_CODE";;
+  *) exit-as-unstable "Unknown error in cluster launch: $CLUSTER_LAUNCH_CODE";;
 esac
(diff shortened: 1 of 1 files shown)


ci/si_pipeline.sh
```diff
@@ -63,5 +63,5 @@ case $CLUSTER_LAUNCH_CODE in
       ;;
   2) exit-as-unstable "Cluster launch failed.";;
   3) exit-as-unstable "Cluster did not start in time.";;
-  *) echo "Unknown error in cluster launch"; exit "$CLUSTER_LAUNCH_CODE";;
+  *) exit-as-unstalbe "Unknown error in cluster launch: $CLUSTER_LAUNCH_CODE";;
```
alenkacz: typo here :)

ci/si_pipeline.sh
```diff
@@ -29,7 +29,7 @@ function create-junit-xml {
 }
 
 function exit-as-unstable {
-    echo "Cluster launch failed."
+    echo "$1"
     create-junit-xml "dcos-launch" "cluster.create" "$1"
     ./dcos-launch delete
     exit 0
```
alenkacz: why do we exit with 0 here in case of error?

jeschkies: Ah, because a launch failure should be an unstable build not a broken build.

alenkacz: ok, makes sense! :) thanks

PR #5856, opened 2017-12-22
Block and restore iptables rules on one command

alenkacz:
Summary:
Doing blocking and restoration in multiple command can cause problems because one of the SSH command might fail. Also if we manage to partition out the master whose IP we use as DCOS_URL, then all subsequent API requests (as well as SSH commands) fail.

This should make it more reliable.

JIRA issues: MARATHON-7965

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -7,7 +7,6 @@ import time
 import uuid
 import sys
 import retrying
-import contextlib
 
 from datetime import timedelta
 from dcos import http, mesos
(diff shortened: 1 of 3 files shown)


tests/system/common.py
```diff
@@ -226,38 +226,21 @@ def systemctl_master(command='restart'):
...
-        host,
-        'if [ ! -e {} ] ; then sudo iptables-save > {} ; fi'.format(filename, filename))
-
-
-def restore_iptables(host, filename='iptables.rules'):
-    """ Reconnect a previously partitioned node to the network
-        :param hostname: host or IP of the machine to partition from the cluster
-    """
-
+def block_iptable_rules(host, port_number, sleep_seconds, input=True, output=True):
```
jeschkies: Could you add some docs? This implementation is not trivial.

Some nit: How about `block_iptable_rules_for_seconds`.

tests/system/common.py
```diff
@@ -226,38 +226,21 @@ def systemctl_master(command='restart'):
...
-@contextlib.contextmanager
-def iptable_rules(host):
-    filename = 'iptables-{}.rules'.format(uuid.uuid4().hex)
-    save_iptables(host, filename)
-    try:
-        yield
-    finally:
-        # return config to previous state
-        restore_iptables(host, filename)
+        """if [ ! -e {} ] ; then sudo iptables-save > {} ; fi;
```
jeschkies: A little hard to read. How about 

```python
cmd = """
  if [ ! -e {backup} ] ; then sudo iptables-save > {backup} ; fi;
  sudo iptables -I INPUT -p tcp --dport {input_port} -j DROP;
  sudo iptables -I OUTPUT -p tcp --dport {output_port} -j DROP;
  sleep {seconds};
  if [ -e {backup} ]; then sudo iptables-restore < {backup} && sudo rm {backup} ; fi
""".format(backup=filename, seconds=sleep_seconds, input_port=input, output_port=output)
```

tests/system/test_marathon_root.py
```diff
@@ -158,14 +158,11 @@ def test_marathon_zk_partition_leader_change(marathon_service_name):
...
 
-    # blocking zk on marathon leader (not master leader)
-    with common.iptable_rules(original_leader):
-        common.block_port(original_leader, 2181, direction='INPUT')
-        common.block_port(original_leader, 2181, direction='OUTPUT')
-        #  Wait for a leader change before restoring  iptables rules
-        common.marathon_leadership_changed(original_leader)
-        # Make sure marathon is available
-        shakedown.wait_for_service_endpoint(marathon_service_name, timedelta(minutes=5).total_seconds())
+    common.block_iptable_rules(original_leader, 2181, 30)
```
jeschkies: Nit: `common.block_iptable_rules(original_leader, 2181, sleep_seconds=30)`

tests/system/test_marathon_root.py
```diff
@@ -174,12 +171,11 @@ def test_marathon_master_partition_leader_change(marathon_service_name):
     original_leader = common.get_marathon_leader_not_on_master_leader_node()
 
     # blocking outbound connection to mesos master
-    with common.iptable_rules(original_leader):
-        common.block_port(original_leader, 5050, direction='OUTPUT')
-        #  Wait for a leader change before restoring  iptables rules
-        common.marathon_leadership_changed(original_leader)
-        # Make sure marathon is available
-        shakedown.wait_for_service_endpoint(marathon_service_name, timedelta(minutes=5).total_seconds())
+    common.block_iptable_rules(original_leader, 5050, 60, False, True)
```
jeschkies: Nit: Please name the parameters as it's not clear what they mean.

tests/system/common.py
alenkacz: nice, I did not know you can do that!

tests/system/common.py
```diff
@@ -226,38 +225,26 @@ def systemctl_master(command='restart'):
...
-        restore_iptables(host, filename)
+    cmd = """
+          if [ ! -e {backup} ] ; then sudo iptables-save > {backup} ; fi;
+          {block}
+          sleep {seconds};
+          if [ -e {backup} ]; then sudo iptables-restore < {backup} && sudo rm {backup} ; fi
+        """.format(backup=filename, seconds=sleep_seconds,
+                   block=iptables_block_string(block_input, block_output, port_number))
+
+    shakedown.run_command_on_agent(host, cmd)
```
jeschkies: Isn't this a blocking call?

alenkacz: I think it is... I think right now it is not a problem. If we move to parallel testing than it might be problem...

jeschkies:
🎄

PR #5857, opened 2017-12-22
 Ease up on the toString for Group and log clashing ids.

jeschkies:
Summary:
This is a forward port of @timcharper's hot fix for 1.5.2.

Group toString serializes every app definition in its entirety and it is
not cheap to render to logs in large clusters.

JIRA issues: MARATHON-7990

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -3,14 +3,15 @@ package state
 
 import java.util.Objects
 
+import com.typesafe.scalalogging.StrictLogging
 import com.wix.accord._
 import com.wix.accord.dsl._
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.plugin.{ Group => IGroup }
-import mesosphere.marathon.state.Group._
-import mesosphere.marathon.state.PathId._
+import mesosphere.marathon.state.Group.{ defaultApps, defaultPods, defaultGroups, defaultDependencies, defaultVersion }
+import mesosphere.marathon.state.PathId.{ validPathWithBase, StringPathId }
 
 class Group(
     val id: PathId,
@@ -111,7 +112,7 @@ class Group(
   override def toString = s"Group($id, ${apps.values}, ${pods.values}, ${groupsById.values}, $dependencies, $version)"
 }
 
-object Group {
+object Group extends StrictLogging {
   type GroupKey = PathId
 
   def apply(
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -108,10 +109,26 @@ class Group(
...
+    s ++= ")"
+    s.toString
+  }
+  override def toString = {
+    val summarizedApps = summarize(apps.valuesIterator.map(_.id))
+    val summarizedPods = summarize(pods.valuesIterator.map(_.id))
+    val summarizedGroups = summarize(groupsById.valuesIterator.map(_.id))
+    val summarizedDependencies = summarize(dependencies.iterator)
+
+    s"Group($id, $summarizedApps, $summarizedPods, $summarizedGroups, $summarizedDependencies, $version)"
```
kensipe: This needs to be more clear... it's current output would be something like:
`Group(/, Seq(/marathon-slack, /marathon-lb), Seq(), Seq(/development, /integration))`

It is currently not possible from the logs to know what is apps, what is pods, what is groups, etc.
lets make this easier and more parsable.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -3,14 +3,15 @@ package state
...
 import com.wix.accord._
 import com.wix.accord.dsl._
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.plugin.{ Group => IGroup }
-import mesosphere.marathon.state.Group._
-import mesosphere.marathon.state.PathId._
+import mesosphere.marathon.state.Group.{ defaultApps, defaultPods, defaultGroups, defaultDependencies, defaultVersion }
+import mesosphere.marathon.state.PathId.{ validPathWithBase, StringPathId }
```
kensipe: none of these imports seem to be used... are they implicit?  if they are can we call them out?

src/main/scala/mesosphere/marathon/state/Group.scala
ichernetsky: 👍

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -108,10 +109,26 @@ class Group(
 
   override def hashCode(): Int = Objects.hash(id, apps, pods, groupsById, dependencies, version)
 
-  override def toString = s"Group($id, ${apps.values}, ${pods.values}, ${groupsById.values}, $dependencies, $version)"
+  private def summarize[T](iterator: Iterator[T]): String = {
```
meln1k: Should we move this to some other method?

src/main/scala/mesosphere/marathon/state/Group.scala
timcharper: It would be easy enough to put it, "apps = ...", Etc. agreed.

src/main/scala/mesosphere/marathon/state/Group.scala
jeschkies: `validPathWithBase` is used in line 137. `StringPathId` is indeed an explicit import of an implicit. It's used in line 195 for `group.id.map(_.toPath)`.

PR #5858, opened 2017-12-27
Fix network_infos retrieval after Mesos update

alenkacz:
Summary:
Mesos bump changes the format of the statuses returned - it started returning status STARTING which does not contain network infos. So wherever we access statuses[0] we don't get the running status as we are expecting.

JIRA issues:

### DIFF
--- a/tests/system/marathon_pods_tests.py
+++ b/tests/system/marathon_pods_tests.py
@@ -519,7 +519,7 @@ def test_pod_with_persistent_volume():
 
     tasks = common.get_pod_tasks(pod_id)
 
-    host = tasks[0]['statuses'][0]['container_status']['network_infos'][0]['ip_addresses'][0]['ip_address']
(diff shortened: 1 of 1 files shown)


tests/system/marathon_pods_tests.py
```diff
@@ -519,7 +519,7 @@ def test_pod_with_persistent_volume():
 
     tasks = common.get_pod_tasks(pod_id)
 
-    host = tasks[0]['statuses'][0]['container_status']['network_infos'][0]['ip_addresses'][0]['ip_address']
+    host = common.running_status_network_info(tasks[0]['statuses'])['ip_addresses'][0]['ip_address']
```
kensipe: we need to know if this is based on recent breaking changes for networks... if it is, we need to know if it works for older marathon versions.   We need a solution that can be run on all supported cluster versions.

alenkacz: this solution will work with old as well as new versions...

PR #5866, opened 2018-01-03
Rename to AgentMaintenance, add it to queue stats

alenkacz:
Summary:
AgentMaintenance seems like a better name for the offer rejection.
Also it is not included in rejection stats in queue - I think the reason for that is that it was missing in reasonFunnel.

JIRA issues: MARATHON-7999


alenkacz:
@timcharper I did not find any but it was not very client-facing so far.

But I have altered the example json for queue so thank you for pointing that out!

PR #5867, opened 2018-01-03
Replace break statement with flag.

jeschkies:
We used to break the while loop of a buffer read. This exception would
propagate through to Jetty. A simple flag should be fine.

JIRA issues: MARATHON-8000

### DIFF
--- a/src/main/scala/mesosphere/marathon/io/IO.scala
+++ b/src/main/scala/mesosphere/marathon/io/IO.scala
@@ -7,7 +7,6 @@ import com.typesafe.scalalogging.StrictLogging
 import org.apache.commons.io.IOUtils
 
 import scala.util.{ Failure, Success, Try }
-import scala.util.control.Breaks.break
 
 object IO extends StrictLogging {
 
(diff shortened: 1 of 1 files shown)


timcharper:
I created a JIRA issue and updated the description.

src/main/scala/mesosphere/marathon/io/IO.scala
```diff
@@ -47,21 +45,19 @@ object IO extends StrictLogging {
...
+  def transfer(maybeFrom: Option[InputStream], maybeTo: Option[OutputStream]): Long = {
+    (maybeFrom, maybeTo) match {
       case (Some(from), Some(to)) =>
-        val buf = createBuffer();
-        var total = 0
-        while (true) {
-          val r = from.read(buf)
-          if (r == -1) {
-            break
+        @tailrec def iter(buf: Array[Byte], total: Long): Long =
```
meln1k: Why do we need to pass `buf: Array[Byte]` as an argument?

PR #5868, opened 2018-01-04
Replace break statement with flag. (#5867)

jeschkies:
We used to break the while loop of a buffer read. This exception would
propagate through to Jetty. A simple flag should be fine.

JIRA issues: MARATHON-8000

4043a2e

### DIFF
--- a/src/main/scala/mesosphere/marathon/io/IO.scala
+++ b/src/main/scala/mesosphere/marathon/io/IO.scala
@@ -5,13 +5,11 @@ import java.io.{ Closeable, File, FileNotFoundException, InputStream, OutputStre
 
 import com.typesafe.scalalogging.StrictLogging
 import org.apache.commons.io.IOUtils
+import scala.annotation.tailrec
 
 import scala.util.{ Failure, Success, Try }
-import scala.util.control.Breaks.break
 
 object IO extends StrictLogging {
-
-  def createBuffer() = new Array[Byte](8192)
 
   def listFiles(file: String): Array[File] = listFiles(new File(file))
   def listFiles(file: File): Array[File] = {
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/io/IO.scala
```diff
@@ -47,21 +45,19 @@ object IO extends StrictLogging {
...
+              to.flush()
+              iter(buf, total + r)
           }
-          to.write(buf, 0, r)
-          to.flush()
-          total += r
-        }
-        total
+
+        iter(new Array[Byte](8192), 0L)
```
ichernetsky: Would it make sense to actually make the buffer size larger?

src/main/scala/mesosphere/marathon/io/IO.scala
```diff
@@ -47,21 +45,19 @@ object IO extends StrictLogging {
...
-        while (true) {
-          val r = from.read(buf)
-          if (r == -1) {
-            break
+        @tailrec def iter(buf: Array[Byte], total: Long): Long =
+          from.read(buf) match {
+            case -1 => total
+            case r =>
+              to.write(buf, 0, r)
+              to.flush()
```
ichernetsky: What's the advantage of flushing on every write call, instead of calling it after it is done writing?

jeschkies: That was a bug. When we copy an SSE stream the write will never be finished. So we have to flush after each chunk.

src/main/scala/mesosphere/marathon/io/IO.scala
jeschkies: This is the default from Guava and Apache Commons IO. So I'm not sure if it makes sense to differ from them.

src/main/scala/mesosphere/marathon/io/IO.scala
ichernetsky: Got it. Thank you!

src/main/scala/mesosphere/marathon/io/IO.scala
ichernetsky: Thank you!

PR #5869, opened 2018-01-05
Benchmark and speed up group validation.

jeschkies:
Summary:
Traversing the groups recursively is slower than going over the flattened id.

This also includes a bug fix. The validation of
```
app.id is PathId.validPathWithBase(base)
```
as the based base was always the parent. The same holds up for
```
path is childOf(path.parent)
```

The invalid app definitions in the unit tests have been fixed as well.

Additionally this change introduces more expressive violations paths. See the unit test changes
for details.

JIRA issues: MARATHON_EE-883

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -165,7 +165,7 @@ object Group extends StrictLogging {
       group.transitiveGroupValues as "groups" is every(
         noAppsAndPodsWithSameId and noAppsAndGroupsWithSameName and noPodsAndGroupsWithSameName
           and isTrue("Group has to be child of groups with parent id") { childGroup =>
-            if (childGroup.id.parent.isRoot) group.groupsById.contains(childGroup.id)
+            if (childGroup.id.parent == group.id) group.groupsById.contains(childGroup.id)
             else {
               group.group(childGroup.id.parent) match {
                 case None => false
@@ -196,7 +196,7 @@ object Group extends StrictLogging {
 
   private def isChildOfParentId(group: Group): Validator[AppDefinition] = {
     isTrue("App has to be child of group with parent id") { app =>
-      if (app.id.parent.isRoot) group.apps.contains(app.id)
+      if (app.id.parent == group.id) group.apps.contains(app.id)
       else {
         group.group(app.id.parent) match {
           case None => false
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -88,6 +88,7 @@ class Group(
     builder ++= transitiveGroups()
     builder.result()
   }
+  lazy val transitiveGroupValues: Iterable[Group] = transitiveGroups().map(_._2).toVector
 
   /** @return true if and only if this group directly or indirectly contains app definitions. */
   def containsApps: Boolean = apps.nonEmpty || groupsById.exists { case (_, group) => group.containsApps }
@@ -152,12 +153,23 @@ object Group extends StrictLogging {
   def validGroup(base: PathId, enabledFeatures: Set[String]): Validator[Group] =
     validator[Group] { group =>
       group.id is validPathWithBase(base)
-      group.apps.values as "apps" is every(
-        AppDefinition.validNestedAppDefinition(group.id.canonicalPath(base), enabledFeatures))
+      //      group.apps.values as "apps" is every(
+      //        AppDefinition.validNestedAppDefinition(group.id.canonicalPath(base), enabledFeatures))
+      group.transitiveApps as "apps" is every(
+        validator[AppDefinition] { app =>
+          group.group(app.id.parent) match {
+            case None => Failure
+            case Some(parentGroup) => validPathWithBase(parentGroup.id).apply(app.id)
+          }
+        })
       group is noAppsAndPodsWithSameId
       group is noAppsAndGroupsWithSameName
       group is noPodsAndGroupsWithSameName
-      group.groupsById.values as "groups" is every(validGroup(group.id.canonicalPath(base), enabledFeatures))
+      //group.groupsById.values as "groups" is every(validGroup(group.id.canonicalPath(base), enabledFeatures))
+      group.transitiveGroupValues as "groups" is every(
+        // TODO: Validate path with base
+        noAppsAndPodsWithSameId and noAppsAndGroupsWithSameName and noPodsAndGroupsWithSameName
+      )
     }
 
   private def noAppsAndPodsWithSameId: Validator[Group] =
@@ -167,8 +179,8 @@ object Group extends StrictLogging {
 
   private def noAppsAndGroupsWithSameName: Validator[Group] =
     isTrue("Groups and Applications may not have the same identifier.") { group =>
-      val groupIds = group.groupsById.keySet
-      val clashingIds = groupIds.intersect(group.apps.keySet)
+      val groupIds = group.groupsById.keys
+      val clashingIds = groupIds.filter(group.apps.contains(_))
       if (clashingIds.nonEmpty)
         logger.info(s"Found the following clashingIds in group ${group.id}: ${clashingIds}")
       clashingIds.isEmpty
@@ -176,8 +188,10 @@ object Group extends StrictLogging {
 
   private def noPodsAndGroupsWithSameName: Validator[Group] =
     isTrue("Groups and Pods may not have the same identifier.") { group =>
-      val groupIds = group.groupsById.keySet
-      val clashingIds = groupIds.intersect(group.pods.keySet)
+      val groupIds = group.groupsById.keys
+      val clashingIds = groupIds.filter(group.pods.contains(_))
+      if (clashingIds.nonEmpty)
+        logger.info(s"Found the following clashingIds in group ${group.id}: ${clashingIds}")
       clashingIds.isEmpty
     }
 
--- a/src/main/scala/mesosphere/marathon/state/Group.scala
+++ b/src/main/scala/mesosphere/marathon/state/Group.scala
@@ -10,8 +10,8 @@ import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.plugin.{ Group => IGroup }
-import mesosphere.marathon.state.Group.{ defaultApps, defaultPods, defaultGroups, defaultDependencies, defaultVersion }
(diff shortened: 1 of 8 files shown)


jeschkies:
# Benchmark `validateRootGroup`

sbt "benchmark/jmh:run -i 10 -wi 3 -f1 -t1 .*RootGroupBenchmark.validateRootGroup"

## Master
|(appsPerGroup)|(groupDepth) | Mode | Cnt | Score | Error | Units
|---:|---:|---:|---:|---:|----:|---|
|   2|  5| avgt| 10|   0.406| ± 0.040|ms/op
|   2| 10| avgt| 10|   0.782| ± 0.056|ms/op
|   2| 20| avgt| 10|   1.844| ± 0.150|ms/op
|  10|  5| avgt| 10|   1.677| ± 0.090|ms/op
|  10| 10| avgt| 10|   4.034| ± 0.294|ms/op
|  10| 20| avgt| 10|   8.236| ± 0.650|ms/op
| 100|  5| avgt| 10|  17.563| ± 1.376|ms/op
| 100| 10| avgt| 10|  37.673| ± 2.930|ms/op
| 100| 20| avgt| 10|  86.548| ± 4.552|ms/op
|1000|  5| avgt| 10| 179.341| ± 9.539|ms/op
|1000| 10| avgt| 10| 355.218| ±12.458|ms/op
|1000| 20| avgt| 10| 941.220| ±31.937|ms/op

## This Patch
|(appsPerGroup)|(groupDepth) | Mode | Cnt | Score | Error | Units
|---:|---:|---:|---:|---:|----:|---|
|   2|  5| avgt| 10|   0.333| ± 0.038|ms/op
|   2| 10| avgt| 10|   0.699| ± 0.207|ms/op
|   2| 20| avgt| 10|   1.284| ± 0.137|ms/op
|  10|  5| avgt| 10|   1.453| ± 0.210|ms/op
|  10| 10| avgt| 10|   3.079| ± 0.277|ms/op
|  10| 20| avgt| 10|   5.936| ± 0.468|ms/op
| 100|  5| avgt| 10|  14.520| ± 0.872|ms/op
| 100| 10| avgt| 10|  30.836| ± 2.073|ms/op
| 100| 20| avgt| 10|  60.739| ± 3.540|ms/op
|1000|  5| avgt| 10| 152.573| ± 9.179|ms/op
|1000| 10| avgt| 10| 320.610| ±22.012|ms/op
|1000| 20| avgt| 10| 631.478| ±60.635|ms/op

That is roughly a 30% speed up for 20,000 apps distributed over twenty groups.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -152,13 +153,58 @@ object Group extends StrictLogging {
...
+              group.group(childGroup.id.parent) match {
+                case None => false
+                case Some(parentGroup) => parentGroup.groupsById.contains(childGroup.id)
+              }
+            }
+          }
+      )
+    }
+
+  implicit def everyApp(validator: Validator[AppDefinition]): Validator[Iterable[AppDefinition]] = {
```
jeschkies: `Validation.every` outputs the wrong paths. Also, I think it is better to show the actual app id.

src/test/scala/mesosphere/marathon/core/deployment/DeploymentPlanTest.scala
```diff
@@ -96,12 +96,12 @@ class DeploymentPlanTest extends UnitTest with GroupCreation {
...
-      val app3 = AppDefinition("app3".toPath, Some("cmd3"))
-      val updatedApp1 = AppDefinition("app".toPath, Some("sleep 30"))
-      val updatedApp2 = AppDefinition("app2".toPath, Some("cmd2"), instances = 10)
-      val app4 = AppDefinition("app4".toPath, Some("cmd4"))
+      val app1 = AppDefinition("/group/app".toPath, Some("sleep 10"))
+      val app2 = AppDefinition("/group/app2".toPath, Some("cmd2"))
+      val app3 = AppDefinition("/group/app3".toPath, Some("cmd3"))
+      val updatedApp1 = AppDefinition("/group/app".toPath, Some("sleep 30"))
+      val updatedApp2 = AppDefinition("/group/app2".toPath, Some("cmd2"), instances = 10)
+      val app4 = AppDefinition("/group/app4".toPath, Some("cmd4"))
```
jeschkies: These apps *did* pass the old validation.

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -152,13 +153,58 @@ object Group extends StrictLogging {
   def validGroup(base: PathId, enabledFeatures: Set[String]): Validator[Group] =
     validator[Group] { group =>
       group.id is validPathWithBase(base)
-      group.apps.values as "apps" is every(
-        AppDefinition.validNestedAppDefinition(group.id.canonicalPath(base), enabledFeatures))
```
jeschkies: Is there a reason that we did not validate the pods?

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -152,32 +153,79 @@ object Group extends StrictLogging {
...
+        }
+      }
+    }
+  }
+
+  private def isChildOfParentId(group: Group): Validator[AppDefinition] = {
+    isTrue("App has to be child of group with parent id") { app =>
+      if (app.id.parent.isRoot) group.apps.contains(app.id)
+      else {
+        group.group(app.id.parent) match {
```
zen-dog: Shouldn't it be smth. like `group.group(app.id.parent).id == group.id`? (just checking the ids)

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -152,32 +153,79 @@ object Group extends StrictLogging {
...
+    }
+  }
+
+  private def isChildOfParentId(group: Group): Validator[AppDefinition] = {
+    isTrue("App has to be child of group with parent id") { app =>
+      if (app.id.parent == group.id) group.apps.contains(app.id)
+      else {
+        group.group(app.id.parent) match {
+          case None => false
+          case Some(parentGroup) => parentGroup.apps.contains(app.id)
```
zen-dog: Wdyt about just checking the ids e.g. smth. like `group.group(app.id.parent).id == group.id`?

src/main/scala/mesosphere/marathon/state/Group.scala
jeschkies: I've changed it to `(app.id.parent == group.id) group.apps.contains(app.id)`

src/test/scala/mesosphere/marathon/api/v2/ModelValidationTest.scala
```diff
@@ -87,13 +87,9 @@ class ModelValidationTest extends UnitTest with GroupCreation {
...
 
-      validate(rootGroup)(RootGroup.rootGroupValidator(Set())) match {
-        case Success => fail()
-        case f: Failure =>
-          val errors = (Json.toJson(f) \ "details").as[Seq[JsObject]]
-          errors should have size 1
-          (errors.head \ "path").as[String] should be("/groups(0)/groups(0)/apps(1)")
-      }
+      validate(rootGroup)(RootGroup.rootGroupValidator(Set())) should haveViolations(
+        "/apps//test/group1/invalid" -> "AppDefinition must either contain one of 'cmd' or 'args', and/or a 'container'."
```
zen-dog: 👍

src/main/scala/mesosphere/marathon/state/Group.scala
zen-dog: ...or just group equality: `group.group(app.id.parent) == group` since `equals` already checking everything?

src/main/scala/mesosphere/marathon/state/Group.scala
zen-dog: On the second thought shouldn't we check the equality of the whole group e.g. `app.id.parent == group`?

src/main/scala/mesosphere/marathon/state/Group.scala
jeschkies: I'm not sure this is the same. Basically we want to avoid this case

```json
{
    "id": "/my_group",
    "apps": { "/different_group/my_app": {"cmd": "sleep"}}
}
{
    "id":"/different_group",
    "apps": {"/different_group/another_app": {"cmd":"sleep"}}
}
```
So for app `my_app` we fetch it's parent `different_group` and check that it does actually hold the app. As I've mentioned we probably should do this check during the creation of a group or an `updateApp` and `updatePod` call.

zen-dog: Ah, I get it now. `parentGroup` confused me a bit - I'd rename it to `childGroup`

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -88,6 +88,7 @@ class Group(
     builder ++= transitiveGroups()
     builder.result()
   }
+  lazy val transitiveGroupValues: Iterable[Group] = transitiveGroups().map(_._2).toVector
```
alenkacz: I would personally do this but I guess it is not that important in this case...

```
{ case (_, group) => group }
```

src/main/scala/mesosphere/marathon/state/Group.scala
alenkacz: do we know why is the pod validation missing here? It seems to me too that it should be here...

src/main/scala/mesosphere/marathon/state/Group.scala
```diff
@@ -152,13 +153,64 @@ object Group extends StrictLogging {
...
+
       group is noAppsAndPodsWithSameId
       group is noAppsAndGroupsWithSameName
       group is noPodsAndGroupsWithSameName
-      group.groupsById.values as "groups" is every(validGroup(group.id.canonicalPath(base), enabledFeatures))
+
+      group.transitiveGroupValues as "groups" is every(
+        noAppsAndPodsWithSameId and noAppsAndGroupsWithSameName and noPodsAndGroupsWithSameName
+          and isTrue("Group has to be child of groups with parent id") { childGroup =>
+            if (childGroup.id.parent == group.id) group.groupsById.contains(childGroup.id)
```
alenkacz: oh, I get it now... this row is just an optimization to not do lookout when we are dealing with parent...

that do you think about moving this optimization directly to the group method (it can be it's implementation detail). Then we can take advantage of this speedup everywhere...

src/main/scala/mesosphere/marathon/state/Group.scala
jeschkies: I don't know why. I guess it was just missed.

PR #5874, opened 2018-01-08
Ease up on the toString for Group and log clashing ids.

timcharper:
Group toString serializes every app definition in its entirety and it is
not cheap to render to logs in large clusters.

Backport of 2da552f

JIRA issues: MARATHON-7990


ichernetsky:
Restart the build job three times, and it failed three times for unrelated reasons, AFAICT.

PR #5877, opened 2018-01-09
An akka-stream based mesos v1 scheduler implementation

zen-dog:
An akka-stream based mesos v1 scheduler implementation:
- implemented as a submodule `mesos-client`
- used _scalapb_ to compile protobufs
- provides _Source[Event]_ to output mesos events and a _Sink[Call]_ to send calls to mesos
- added basic integration tests
- has mesos protobuf files (`mesos.proto` and `scheduler.proto`) in `mesos-client/src/main/protobuf`
- uses `scalapb` plugin and set it up to compile mesos protobuf files
- uses backpressure overflow strategy for mesos source
- handles mesos leader redirectes correctly

JIRA issues: MARATHON-7657


src/test/scala/mesosphere/marathon/integration/facades/MesosFacade.scala
```diff
@@ -83,7 +83,7 @@ object MesosFacade {
   case class ITFrameworks(frameworks: Seq[ITFramework])
 }
 
-class MesosFacade(url: String, implicit val waitTime: FiniteDuration = 30.seconds)(implicit val system: ActorSystem, materializer: Materializer)
+class MesosFacade(val url: String, implicit val waitTime: FiniteDuration = 30.seconds)(implicit val system: ActorSystem, materializer: Materializer)
```
timcharper: Scala allows the mix of implicits like this? (even if it does... probably shouldn't :) :) )

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+      .via(log("Received mesos Event: "))
+  }
+
+  override lazy val (killSwitch: UniqueKillSwitch, mesosSource: Source[Event, NotUsed]) = source
+    .viaMat(KillSwitches.single)(Keep.right)
+    .toMat(BroadcastHub.sink)(Keep.both)
+    .run()
+
+
+  /**
```
timcharper: awesome docs! 👏  👏  👏

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+    * }
+    * ```
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#subscribe-1
+    */
+  def mesosSource: Source[Event, NotUsed]
+
+  /**
+    * Sink for mesos calls. Multiple publishers can materialize this sink to send mesos `Call`s. Every `Call` is sent
+    * using a new HTTP connection.
```
timcharper: There's no way to pipeline these then? :(

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+    */
+  def mesosSink: Sink[Call, NotUsed]
+
+  /**
+    * A kill switch for the mesos source. Calling `shutdown()` or `abort()` on it will close the connection to mesos.
+    * Note that depending on `failoverTimeout` provided with SUBSCRIBED call mesos could start killing tasks and
+    * executors started by the framework. Make sure to set `failoverTimeout` appropriately. See `teardown()` method
+    * for another way to shutdown a framework.
+    */
+  val killSwitch: UniqueKillSwitch
```
timcharper: What is the advantage of exposing two ways to tear down?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+      declineInverseOffers = Some(Call.DeclineInverseOffers(inverseOfferIds = offers, filters = filters))
+    )
+  }
+
+}
+
+// TODO: Add more integration tests
+
+object MesosClient {
+  case class MesosRedicrectException(leader: URI) extends Exception(s"New mesos leader available at $leader")
```
timcharper: spelling

timcharper: spelling

mesos-client/README.md
```diff
@@ -0,0 +1,129 @@
...
+5. **Event Deserializer** Currently mesos-v1-client only supports protobuf encoded events/calls. Event deserializer uses [scalapb](https://scalapb.github.io/) library to parse the extracted RecordIO frame from the previous stage into a mesos [Event](https://github.com/apache/mesos/blob/master/include/mesos/scheduler/scheduler.proto#L36)
+6. **Subscribed Handler** parses the `SUBSCRIBED` event and updates the connection context with the returned `frameworkId`.
+7. **BroadcastHub** ad the end the flow is going through a broadcast hub. This allows for a dynamic "fan-out" streaming of events and avoids making multiple upstream connections to mesos when the source is materialized by multiple event consumers
```
timcharper: "at the end"

I would motion that we should leave the usage of BroadcastHub up to the user. It is problematic in that the subscribers may not see the first element, and, back-pressure is applied for the slowest of subscribers.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+      .idleTimeout(conf.idleTimeout().seconds)
+      .buffer(conf.sourceBufferSize(), overflowStrategy)
+      .via(dataBytesExtractor)
+      .via(recordIoScanner)
+      .via(eventDeserializer)
+      .via(subscribedHandler)
+      .via(log("Received mesos Event: "))
+  }
+
+  override lazy val (killSwitch: UniqueKillSwitch, mesosSource: Source[Event, NotUsed]) = source
```
timcharper: as seen above ^ I'm not sure that using Broadcast hub by default is a good idea.

Perhaps we can just expose the `Source`, and then leave it up to the user to use BroadcastHub or not? And, perhaps we should just connect to Mesos when the user materializes said Source?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+      `type` = Some(Call.Type.DECLINE),
+      decline = Some(Decline(offerIds = offerIds, filters = filters))
+    )
+  }
+
+  /** Sent by the scheduler to remove any/all filters that it has previously set via ACCEPT or DECLINE calls.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#revive
+    */
+  def revive(roles: Option[String] = None): Call = {
```
timcharper: `roles: Option[String]`? Should be a collection?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+      revive = Some(Revive(role = roles))
+    )
+  }
+
+  /** Suppress offers for the specified roles. If `roles` is empty the `SUPPRESS` call will suppress offers for all
+    * of the roles the framework is currently subscribed to.
+    *
+    * http://mesos.apache.org/documentation/latest/upgrades/#1-2-x-revive-suppress
+    */
+  def suppress(roles: Option[String] = None): Call = {
```
timcharper: ditto :)

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+      kill = Some(Kill(taskId = taskId, agentId = agentId, killPolicy = killPolicy))
+    )
+  }
+
+  /** Sent by the scheduler to shutdown a specific custom executor. When an executor gets a shutdown event, it is
+    * expected to kill all its tasks (and send TASK_KILLED updates) and terminate.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#shutdown
+    */
+  def shutdown(executorId: ExecutorID, agentId: AgentID):Call = {
```
timcharper: Where does the ExecutorID name come from... Mesos? Can we make it ExecutorId? Preference towards consistence capitalization of Id, and, id is not an acronym.

mesos-client/src/main/scala/mesosphere/mesos/conf/MesosConf.scala
```diff
@@ -0,0 +1,48 @@
+package mesosphere.mesos.conf
+
+import org.rogach.scallop.ScallopConf
+
+/**
+  * Configuring mesos v1 client
+  */
+class MesosConf(args: Seq[String]) extends ScallopConf(args) {
```
timcharper: Perhaps this config should be exposed via a config object, so the user of the client can expose CLI arguments as such?

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
```diff
@@ -0,0 +1,62 @@
...
+import akka.stream.scaladsl.{Sink, Source}
+import com.typesafe.scalalogging.StrictLogging
+import mesosphere.mesos.client.MesosClient
+import mesosphere.mesos.conf.MesosConf
+import org.apache.mesos.v1.mesos.{Filters, FrameworkInfo}
+import org.apache.mesos.v1.scheduler.scheduler.Event
+
+import scala.util.{Failure, Success}
+
+object UselessFramework extends App with StrictLogging {
```
timcharper: Should call it ProudFramework AMIRITE? Too proud to accept any offers.

mesos-client/README.md
zen-dog: I built it in because otherwise every materialisation of the `Source` would lead to a *new connection* to mesos being created. And we definitely don't want that. `BroadcastHub` is one way to avoid it - but I'm open to others!

src/test/scala/mesosphere/marathon/integration/facades/MesosFacade.scala
zen-dog: Now that I'm looking at it - it does look strange :) But my only change was to make `url` public.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+    1. MergeHub allows dynamic "fan-in" junction point for mesos calls from multiple producers.
+    2. Call Enhancer updates the call with the framework Id from the connection context
+    3. Event Serializer serializes calls to byte array
+    4. Build a HTTP request from the data using `mesosStreamId` header from the context
+    5. Http Sink creates a new connection using akka's `Http().singleRequest` and sends the data
+
+    Note: Merge hub will wait for the _connection context_ object to be fully initialized first meaning that we have current leader's
+    `host`, `port` and `Mesos-Stream-Id` to send the events to.
+    */
+  private val sink: Sink[(ConnectionContext, HttpRequest), Future[Done]] = Sink.foreach[(ConnectionContext, HttpRequest)]{
```
wavesoft: #nit, but perhaps rename to `httpSink` for clarity?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+    case (_, request) =>
+      Http().singleRequest(request).map { response =>
+          logger.info(s"Response: $response")
+          response.discardEntityBytes()
```
wavesoft: Shouldn't we check if the request was successful? What about throwing an exception if the response code was not `2xx`?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
jeschkies: +1

mesos-client/src/main/scala/mesosphere/mesos/conf/MesosConf.scala
jeschkies: How about [Config](https://github.com/lightbend/config) from Typesafe/Lightbend?

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  /** Run Foo framework that:
+    *  - successfully subscribes
+    *  - declines all offers.
+    *
+    *  Not much, but shows the basic idea. Good to test against local mesos.
+    *
+    */
+  override def main(args: Array[String]): Unit = {
```
jeschkies: Where is this called? How about and integration test?

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
```diff
@@ -0,0 +1,62 @@
...
+    implicit val system = ActorSystem()
+    implicit val materializer = ActorMaterializer()
+    implicit val executionContext = system.dispatcher
+
+    val frameworkInfo = FrameworkInfo(
+      user = "foo",
+      name = "Example FOO Framework",
```
jeschkies: nit: `Userless Framework` ;)

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
```diff
@@ -0,0 +1,62 @@
...
+
+      if (event.`type`.get == Event.Type.SUBSCRIBED) {
+        logger.info("Successfully subscribed to mesos")
+      }
+      else if (event.`type`.get == Event.Type.OFFERS) {
+
+        val offerIds = event.offers.get.offers.map(_.id).toList
+
+        Source(offerIds)
+          .map{ oId => logger.info(s"Declining offer with id = ${oId.value}"); oId }  // Decline all offers
```
jeschkies: Why not reuse the `log` flow?

mesos-client/src/test/scala/mesosphere/mesos/client/MesosClientIntegrationTest.scala
```diff
@@ -0,0 +1,145 @@
...
+        mesosStreamId shouldBe defined
+        frameworkId shouldBe defined
+    }
+  }
+
+  "Mesos client should successfully subscribe to mesos with framework Id" in {
+    val frameworkID = FrameworkID(UUID.randomUUID().toString)
+
+    When("a framework subscribes with a framework Id")
+    withFixture(Some(frameworkID)) { f =>
```
jeschkies: Out of curiosity, why is the `withFixture` here and not at the beginning of the test?

mesos-client/src/test/scala/mesosphere/mesos/client/MesosClientIntegrationTest.scala
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  "Mesos client should successfully receive heartbeats" in withFixture() { f =>
+    When("a framework subscribes")
+    val res: Future[Seq[Event]] = f.client.mesosSource.take(2).runWith(Sink.seq)
+
+    Then("a heartbeat event should arrive")
+    inside(res.futureValue) {
+      case events =>
+        events.count(_.`type`.contains(Event.Type.HEARTBEAT)) should be > 0
```
jeschkies: Does this check whether one or two heartbeats are received?

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
meln1k: NarcissisticFramework should also work :)

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+    3. Event Serializer serializes calls to byte array
+    4. Build a HTTP request from the data using `mesosStreamId` header from the context
+    5. Http Sink creates a new connection using akka's `Http().singleRequest` and sends the data
+
+    Note: Merge hub will wait for the _connection context_ object to be fully initialized first meaning that we have current leader's
+    `host`, `port` and `Mesos-Stream-Id` to send the events to.
+    */
+  private val sink: Sink[(ConnectionContext, HttpRequest), Future[Done]] = Sink.foreach[(ConnectionContext, HttpRequest)]{
+    case (_, request) =>
+      Http().singleRequest(request).map { response =>
```
meln1k: It's extremely dangerous to use Sink.foreach here. It will fire http requests with a light-speed rate, quickly overflowing a small http pool (the pool is shared for the entire actor system) and will possible cause unexpected failures. I had this in production and it's definitely not something you'd like to have.

Much better solution is to use a Flow-based variant of the http pool

build.sbt
```diff
@@ -312,3 +312,19 @@ lazy val benchmark = (project in file("benchmark"))
...
     libraryDependencies ++= Dependencies.benchmark
   )
+
+// see also mesos-client/README.md
+lazy val `mesos-client` = (project in file("mesos-client"))
+  .configs(IntegrationTest)
+  .enablePlugins(GitBranchPrompt, BasicLintingPlugin, TestWithCoveragePlugin)
+  .settings(commonSettings: _*)
+  .settings(formatSettings: _*)
+  .dependsOn(marathon % "compile->compile; test->test")
```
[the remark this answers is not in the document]
zen-dog: Hm.  right now `mesos-client` isn't used anywhere in marathon however it already depends on marathon for some basic test classes (e.g. `MesosClusterTest`). Later both would depend on each other...

mesos-client/src/main/scala/mesosphere/mesos/client/ConnectionContext.scala
```diff
@@ -0,0 +1,19 @@
+package mesosphere.mesos.client
+
+import mesosphere.mesos.conf.MesosConf
+import org.apache.mesos.v1.mesos.FrameworkID
+
+case class ConnectionContext(host: String,
```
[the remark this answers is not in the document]
zen-dog: Good idea

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+import akka.http.scaladsl.Http
+import akka.http.scaladsl.model._
+import akka.stream._
+import akka.stream.alpakka.recordio.scaladsl.RecordIOFraming
+import akka.stream.scaladsl._
+import akka.util.ByteString
+import akka.{Done, NotUsed}
+import com.google.protobuf
+import com.typesafe.scalalogging.StrictLogging
+import mesosphere.mesos.client.MesosClient.MesosRedicrectException
```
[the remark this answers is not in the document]
zen-dog: My Intellij rule is to make it `_` when > 5 imports.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,514 @@
...
+    7. BroadcastHub ad the end the flow is going through a broadcast hub. This allows for a dynamic "fan-out" streaming of events and avoids
+       making multiple upstream connections to mesos when the source is materialized by multiple event consumers.
+
+    Note: _Connection Handler_ updates the _connection context_ object with current mesos leader `host` and `port` along with `Mesos-Stream-Id`
+    header value. These settings are used by mesos sink when sending calls to mesos. The same applies to _Subscribed Handler_ which saves
+    `frameworkId` from the `SUBSCRIBED` event in the connection context.
+
+    It is declared lazy to decouple instantiation of the MesosClient instance from connection initialization.
+    */
+  def log[T](prefix: String): Flow[T, T, NotUsed] = Flow[T].map{ e => logger.info(s"$prefix$e"); e }
```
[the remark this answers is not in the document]
zen-dog: I dislike utility files (for tending to accumulate a lot of garbage over time). Also akka-stream has a `log` method that does mostly the same, however it doesn't use our `StrictLogging` and using it there wasn't trivial.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: -1 on that. Main reason why I inserted broadcast huyb was that otherwise a new mesos connection (SUBSCRIBE) would be established on every `mesosSource` materialization. That is not what we want. We want 1 connection in (and multiple requests out). But may be there is a better way to achieve this (without using a `BroadcastHub`)?

zen-dog: See here: https://doc.akka.io/docs/akka-http/current/client-side/connection-level.html#opening-http-connections

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: Very good point. I played with `Http().cachedHostConnectionPool` and `Http().superPool()` but it turns out that they can't guarantee delivery of the messages in the same order they were sent. I'm using `Http().outgoingConnection` that holds one long-living connection to mesos (probably an improvement by itself) and sends calls over it. I also left an explaining comment in code.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: Good point. I decided to fail the stream for now.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: See the changes above ;)

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,540 @@
...
+  // mesos in the same order they were sent.
+  //
+  // `Http().outgoingConnection` will create one long-living connection and use it to send events to mesos. Backpressure is applied
+  // automatically when mesos is slow and events are in order.
+  private val httpConnection: Flow[HttpRequest, HttpResponse, NotUsed] = Flow[HttpRequest]
+      .via(Http().outgoingConnection(context.get().host, context.get().port))
+
+  private val responseHandler: Sink[HttpResponse, Future[Done]] = Sink.foreach[HttpResponse] { response =>
+    response status match {
+      case status if status.isFailure() =>
```
meln1k: we should also do `response.discardEntityBytes()` here :)

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: I'm exposing the `killSwitch` here and it offers 2 ways of tearing down (`shutdown` completes the downstream, `abort` fails it).

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
timcharper: Can we not just make that clear in the documentation? "A new Mesos connection will be established for every source materialization. If you require multiple consumers, please consider using BroadcastHub"

There are multiple other ways to avoid using BroadcastHub. You can use the Akka stream graph DSL, you can use the alsoTo DSL operator on flows/sources, or you can use some other broadcast mechanism.

I think the ideal is that the user of this library makes the decision that makes the most sense. Personally, I would not choose to use BroadcastHub because I lose the guarantee that might consumers will receive the first messages.

mesos-client/README.md
```diff
@@ -0,0 +1,129 @@
...
+A kill switch can be used to forcefully shutdown the mesos source. Calling `shutdown()` or `abort()` on it will close the connection to mesos. Note that depending on `failoverTimeout` provided with `SUBSCRIBED` call mesos could start killing tasks and executors started by the framework. Make sure to set `failoverTimeout` appropriately. See `teardown()` method for another way to shutdown a framework.
+
+## Mesos API Trait
+Mesos API trait has three main methods: `mesosSource` for inbound events, `mesosSink` for outbound calls and `killSwitch` to forcefully disconnect from mesos.
+
+```
+trait MesosApi {
+
+  /**
+    * First call to this method will initialize the connection to mesos and return a `Source[String, NotUser]` with
```
timcharper: Typo: NotUsed

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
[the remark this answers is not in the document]
timcharper: Please make this private :-)

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
timcharper: Are you aware of mapMaterializedValue? It seems like we could expose the killSwitch as a materialized value (whereas right now the materialized value for MesosSource is NotUsed)

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
timcharper: I think he is probably just using it to test the code. (In the end, it should probably be a separate project, but I don't have a problem with it during the prototyping phase :-))

mesos-client/src/test/scala/mesosphere/mesos/client/MesosClientIntegrationTest.scala
zen-dog: it gets the `frameworkID` passed as parameter

mesos-client/src/test/scala/mesosphere/mesos/client/MesosClientIntegrationTest.scala
zen-dog: ok `s/receive heartbeats/receive heartbeat` ;)

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,540 @@
...
+          response
+        case StatusCodes.TemporaryRedirect =>
+          val leader = new URI(response.header[headers.Location].get.value())
+          logger.warn(s"New mesos leader available at $leader")
+          // Update the context with the new leader's host and port and throw an exception that is handled in the
+          // next `recoverWith` stage.
+          context.updateAndGet(c => c.copy(url = leader))
+          throw new MesosRedicrectException(leader)
+        case _ =>
+          throw new IllegalArgumentException(s"Mesos server error: $response")
```
meln1k: I think we may need to drain an http entity source

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
zen-dog: What Tim said ;)

build.sbt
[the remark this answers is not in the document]
zen-dog: I agree with everybody. But in practice that means separating marathon's IT classes into it's own project and that's a pandora box I don't want to open right now.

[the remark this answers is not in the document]
jeschkies: Ok, so once we move to the client we have to refactor our test code.

mesos-client/src/main/scala/mesosphere/mesos/conf/MesosConf.scala
jeschkies: What is the decision on this?

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
jeschkies: Shouldn't it be part of `src/test` then?

project/Dependencies.scala
```diff
@@ -55,6 +62,8 @@ object Dependencies {
     alpakkaS3 % "compile",
     commonsCompress % "compile", // used for tar flow
     commonsIO % "compile",
+    alpakkaCodes % "compile",
```
jeschkies: Do we use Alpakka in Marathon as well?

project/Dependencies.scala
```diff
@@ -117,11 +126,13 @@ object Dependency {
     val JUnitBenchmarks = "0.7.2"
     val JMH = "1.19"
     val ScalaCheck = "1.13.5"
+    val ScalaPb = "0.6.6"
```
jeschkies: Why is this a test dependency?

mesos-client/src/main/scala/mesosphere/mesos/conf/MesosConf.scala
zen-dog: I'm not sure what the fascination with Typesafe conf is - my idea was that we can mix `MesosConf` with `MarathonConf` similar to how we do it now. But I understand that it makes it harder to use this client separately. 
I made it a simple case class with the option to move to Typesafe/whatever later.

project/Dependencies.scala
zen-dog: Yes

mesos-client/src/main/scala/mesosphere/mesos/examples/UselessFramework.scala
zen-dog: Let eat bee, let eat bee...

zen-dog:
> There are many ways to fanout events, BroadcastHub is just one of them. 

As I mentioned previously - the major advantage of the broadcast is that multiple source materialization will *not* lead to new mesos connections. I don't see any scenario where a framework user would need that. Imho it makes sense to prevent this in the client.

> Additionally, I feel strongly that the mechanism for closing the Mesos client should be exposed via a materialized value (Instead of Source[Event, NotUsed], Source[Event, KillSwitch]).

I agree. Hasn't come to changing it yet but I will.

janisz:
IMO mesos-client should be external repo published as a JAR so other implementations such as Jeknins, Spark or Chronos can get benefits from V1 API.

jeschkies:
> IMO mesos-client should be external repo published as a JAR so other implementations such as Jeknins, Spark or Chronos can get benefits from V1 API.

It doesn't have to be an extra repository since we could still publish it to Maven. I do agree though. Once it's stable we could publish it.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -0,0 +1,535 @@
...
+
+    def connectionSource(host: String, port: Int) =
+      Source.single(request)
+      .via(log(s"Connecting to the new leader: $host:$port"))
+      .via(httpConnection(host, port))
+      .via(log("HttpResponse: "))
+      .via(connectionHandler)
+
+    val source = connectionSource(context.get().host, context.get().port)
+      .recoverWithRetries(conf.redirectRetires, {
```
ichernetsky: Spelling: `Retires`

jeschkies:
@timcharper and @zen-dog, what's the state? Is this PR superseded by the new branch? Should we merge and continue?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: Yep, it's mesos type

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
zen-dog: No, revive is only defined for one role in v1. `roles` (plural) is whats confusing here. Fixed

PR #5878, opened 2018-01-09
ChangeLog updates for 1.4.10 and 1.4.11 versions

ichernetsky:
JIRA issues: https://jira.mesosphere.com/browse/MARATHON-7996

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,7 +9,7 @@
 
 ### Fixed issues
 
-- [MARATHON-7974](https://jira.mesosphere.com/browse/MARATHON-7974) - Fixes a security concern for a non super user. It was discovered that permissions for GroupResources was not honored and it is now resolved.
+- [MARATHON-7974](https://jira.mesosphere.com/browse/MARATHON-7974) - Fixes a security concern for a non super user. It was discovered that permissions for GroupResources were not honored in a few places. Specifically, if the user is NOT super user and HAS access, access would be refused anyway. It is now resolved.
 - [MARATHON-7962](https://jira.mesosphere.com/browse/MARATHON-7962) - Enable killGracePeriodSeconds for Pods.
 - [MARATHON_EE-1770](https://jira.mesosphere.com/browse/MARATHON_EE-1770) - Optimized port assignments and performance enhancements.
 - [MARATHON_EE-1764](https://jira.mesosphere.com/browse/MARATHON_EE-1764) - Optimized port assignments and performance enhancements.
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -1,6 +1,24 @@
...
+### Fixed issues
+
+- [MARATHON-7992](https://jira.mesosphere.com/browse/MARATHON-7992) - Validation errors on root group in large clusters can lead Marathon to crash.
+
+
+## Changes from 1.4.9 to 1.4.10
+
+### Fixed issues
+
+- [MARATHON-7974](https://jira.mesosphere.com/browse/MARATHON-7974) - Fixes a security concern for a non super user. It was discovered that permissions for GroupResources was not honored and it is now resolved.
```
timcharper: I think that this is more alarming than it needs to be. It wasn't that we have a security vulnerability, but rather, only super users could access things. So, it was the opposite of a vulnerability (security was too tight).

timcharper: @kensipe can you please confirm that my understanding here is correct?

ichernetsky: That's how it is phrased on the releases page: https://github.com/mesosphere/marathon/releases
Should it be rephrased anyway?

timcharper: Yes, I think it should be. Also, I think we should fix the release notes.

changelog.md
```diff
@@ -1,6 +1,24 @@
+## Changes from 1.4.10 to 1.4.11
+
+### Fixed issues
+
+- [MARATHON-7992](https://jira.mesosphere.com/browse/MARATHON-7992) - Validation errors on root group in large clusters can lead Marathon to crash.
```
kensipe: I'm looking at https://github.com/mesosphere/marathon/compare/v1.4.10...releases/1.4

I see MARATHON-7990 but not MARATHON-7992

ichernetsky: It is for 1.4.11

kensipe: looks like slack is down :( .   The issue is in the tracking (and not on your end).   Marathon-7992 is Unassigned and in review however it has #5874 in the comments.   #5874 is merged (so not in review) and lists MARATHON-7990 as it's JIRA issue.

I'll approve but under duress :)

ichernetsky:
https://jira.mesosphere.com/browse/MARATHON-7992 is the correct one. It looks like the one was just copy and pasted.

ichernetsky:
*the other on* was copy and pasted.

changelog.md
timcharper: Yeah, my bad. At least I added a PR link in the comments, though? :)

ichernetsky: It is a tiny thing, not even worth mentioning. Merging :)

PR #5884, opened 2018-01-10
Document how to pass boolean option using environment variables

eLvErDe:


alenkacz:
@eLvErDe please see also #5903 - by reviewing your PR I found that our docs are obsolete

PR #5887, opened 2018-01-10

zen-dog:
when using docker user networking (and not `hostPorts` exist). This part of code went missing from 1.3 to 1.4.

Also had to fix a bunch of scalalint complains - otherwise it wouldn't compile.

JIRA: COPS-2072

### DIFF
--- a/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
+++ b/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
@@ -87,10 +87,11 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.map(_.name).zip(effectivePorts).foreach {
-        case (Some(portName), Some(effectivePort)) =>
+      requestedPorts.zip(effectivePorts).foreach {
+        case (PortRequest(Some(portName), _), Some(effectivePort)) =>
           env += (s"PORT_${portName.toUpperCase}" -> effectivePort.toString)
-        // TODO(jdef) port name envvars for generated container ports
+        case (PortRequest(Some(portName), port), None) =>
+          env += (s"PORT_${portName.toUpperCase}" -> port.toString)
         case _ =>
       }
 
--- a/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
+++ b/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
@@ -87,11 +87,13 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.zip(effectivePorts).foreach {
-        case (PortRequest(Some(portName), _), Some(effectivePort)) =>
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
```diff
@@ -87,10 +87,11 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.map(_.name).zip(effectivePorts).foreach {
-        case (Some(portName), Some(effectivePort)) =>
+      requestedPorts.zip(effectivePorts).foreach {
+        case (PortRequest(Some(portName), _), Some(effectivePort)) =>
           env += (s"PORT_${portName.toUpperCase}" -> effectivePort.toString)
-        // TODO(jdef) port name envvars for generated container ports
+        case (PortRequest(Some(portName), port), None) =>
```
jdef: should consider generated container ports here (e.g. user declared a container port of 0) otherwise you might end up with some envvar PORT_FOO=0

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -1614,6 +1615,36 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
+  // Regression COPS-2072
+  test("PortsEnvWithOnlyMappingsAndUserNetworking") {
+    val command =
+      TaskBuilder.commandInfo(
+        runSpec = AppDefinition(
+          id = runSpecId,
+          container = Some(Docker(
+            network = Some(DockerInfo.Network.USER),
+            portMappings = Seq(
+              PortMapping(containerPort = 8080, hostPort = None, servicePort = 9000, protocol = "tcp", name = Some("http")),
```
jdef: we also want a test case that sets containerPort=0 so that the "generated container port" logic is tested

src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
jdef: if the intent is NOT to generate envvars for generated containers ports, then ... please reinstate the deleted TODO

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
zen-dog: Good catch. Done.

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -1614,6 +1615,38 @@ class TaskBuilderTest extends MarathonSpec with Matchers {
...
+        hostPorts = Seq(None, None),
+        envPrefix = None
+      )
+    val env: Map[String, String] =
+      command.getEnvironment.getVariablesList.toList.map(v => v.getName -> v.getValue).toMap
+
+    assert("8080" == env("PORT_8080"))
+    assert("8080" == env("PORT_HTTP"))
+
+    val port1 = env("PORT1")
```
jdef: maybe include a sanity check that `port1 != 0`?

PR #5888, opened 2018-01-10

zen-dog:
when using docker user networking (and not `hostPorts` exist). This part of code went missing from 1.3 to 1.4.

JIRA: COPS-2072

### DIFF
--- a/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
+++ b/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
@@ -87,10 +87,11 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.map(_.name).zip(effectivePorts).foreach {
-        case (Some(portName), Some(effectivePort)) =>
+      requestedPorts.zip(effectivePorts).foreach {
+        case (PortRequest(Some(portName), _), Some(effectivePort)) =>
           env += (s"PORT_${portName.toUpperCase}" -> effectivePort.toString)
-        // TODO(jdef) port name envvars for generated container ports
+        case (PortRequest(Some(portName), port), None) =>
+          env += (s"PORT_${portName.toUpperCase}" -> port.toString)
         case _ =>
       }
 
--- a/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
+++ b/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
@@ -87,11 +87,13 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.zip(effectivePorts).foreach {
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
```diff
@@ -87,10 +87,11 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.map(_.name).zip(effectivePorts).foreach {
-        case (Some(portName), Some(effectivePort)) =>
+      requestedPorts.zip(effectivePorts).foreach {
+        case (PortRequest(Some(portName), _), Some(effectivePort)) =>
           env += (s"PORT_${portName.toUpperCase}" -> effectivePort.toString)
-        // TODO(jdef) port name envvars for generated container ports
+        case (PortRequest(Some(portName), port), None) =>
```
jdef: same comments as for #5887

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -1630,6 +1630,36 @@ class TaskBuilderTest extends UnitTest {
...
+          envPrefix = None
+        )
+
+      val env: Map[String, String] =
+        command.getEnvironment.getVariablesList.toList.map(v => v.getName -> v.getValue).toMap
+
+      assert("8080" == env("PORT_8080"))
+      assert("8080" == env("PORT_HTTP"))
+
+      val port1 = env("PORT1")
```
jdef: ditto re: sanity check that `port1 ! = 0`

src/test/scala/mesosphere/mesos/TaskBuilderTest.scala
```diff
@@ -1,25 +1,25 @@
...
+import mesosphere.marathon.state.{AppDefinition, Container, PathId, Timestamp, _}
 import mesosphere.marathon.stream.Implicits._
-import mesosphere.marathon.test.{ MarathonTestHelper, SettableClock }
-import mesosphere.marathon.{ Protos, _ }
-import mesosphere.mesos.protos.{ Resource, _ }
+import mesosphere.marathon.test.{MarathonTestHelper, SettableClock}
+import mesosphere.marathon.{Protos, _}
+import mesosphere.mesos.protos.{Resource, _}
 import org.apache.mesos.Protos.TaskInfo
-import org.apache.mesos.{ Protos => MesosProtos }
```
meln1k: Looks like intelij has different opinion on how to format imports compared to SBT :)

jdef:
lgtm

the port-by-name docs could use some clarification/editing as well: https://github.com/mesosphere/marathon/blob/master/docs/docs/networking.md#per-task-environment-variables

for example, PORT_name doesn't always refer to a host port. these variables are actually pretty confusing because their meaning changes depending upon how ports are allocated by an app. It might be worth a table that clearly shows how the value of the PORTxxx variables changes depending on the app configuration.

PR #5889, opened 2018-01-10
Instance.isReserved must return true if there is at least one Reserved task

ichernetsky:
There is a race condition in task status updates handling in case of pods.
If a pod has two or more containers and terminal states are sent to Marathon
not one right after another, but with some delay, Marathon launches a new pod
without waiting till all the previous tasks reach a terminal state. It is somewhat
benign in case of ephemeral pods, but it is severe in case of pods with
persistent volumes.

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON_EE-1817
https://jira.mesosphere.com/browse/MARATHON_EE-1820

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -34,7 +34,7 @@ case class Instance(
   val runSpecId: PathId = instanceId.runSpecId
   val isLaunched: Boolean = state.condition.isActive
 
-  def isReserved: Boolean = state.condition == Condition.Reserved
+  def isReserved: Boolean = tasksMap.values.exists(_.status.condition == Condition.Reserved)
   def isCreated: Boolean = state.condition == Condition.Created
   def isError: Boolean = state.condition == Condition.Error
   def isFailed: Boolean = state.condition == Condition.Failed
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -34,7 +34,9 @@ case class Instance(
   val runSpecId: PathId = instanceId.runSpecId
   val isLaunched: Boolean = state.condition.isActive
 
+  // An instance has to be considered as Reserved if at least one of its tasks is Reserved.
   def isReserved: Boolean = tasksMap.values.exists(_.status.condition == Condition.Reserved)
+
   def isCreated: Boolean = state.condition == Condition.Created
   def isError: Boolean = state.condition == Condition.Error
   def isFailed: Boolean = state.condition == Condition.Failed
--- a/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
@@ -3,7 +3,7 @@ package core.instance.update
 
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.condition.Condition
-import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.core.instance.{ Instance, Reservation }
 import mesosphere.marathon.core.instance.update.InstanceUpdateOperation.{ LaunchEphemeral, LaunchOnReservation, MesosUpdate, Reserve }
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.update.{ TaskUpdateEffect, TaskUpdateOperation }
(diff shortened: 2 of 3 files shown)


timcharper:
Can we get some context on this? JIRA issue? What is broken without this? Why do we have duplicate reserved state signal?

timcharper:
@meln1k can you please create a JIRA sub-task for this backport, and set a fix version?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -34,7 +34,9 @@ case class Instance(
   val runSpecId: PathId = instanceId.runSpecId
   val isLaunched: Boolean = state.condition.isActive
 
-  def isReserved: Boolean = state.condition == Condition.Reserved
+  // An instance has to be considered as Reserved if at least one of its tasks is Reserved.
+  def isReserved: Boolean = tasksMap.values.exists(_.status.condition == Condition.Reserved)
```
alenkacz: does this change mean that instance might exist where it's condition is not reserved BUT isReserved now returns true?

Wouldn't a cleaner solution be to change the logic in a way that such instance will still have state condition set to reserved?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -45,7 +45,16 @@ object InstanceUpdater extends StrictLogging {
             logger.info("all tasks of {} are terminal, requesting to expunge", updated.instanceId)
             InstanceUpdateEffect.Expunge(updated, events)
           } else {
-            InstanceUpdateEffect.Update(updated, oldState = Some(instance), events)
+            // If the updated task is Reserved, it means that the real task reached a Terminal state,
+            // which in turn means that the task managed to get up and running, which means that
+            // its persistent volume(s) had been created, and therefore they must never be destroyed/unreserved.
+            if (updatedTask.status.condition == Condition.Reserved) {
+              val suspendedState = Reservation.State.Suspended(timeout = None)
+              val suspended = updated.copy(reservation = updated.reservation.map(_.copy(state = suspendedState)))
```
alenkacz: I would personally instead of such a deep object copying go more in this direction
instance.suspendReservation()
and move the rest of the code somewhere to the appropriate case classes

Btw. if ALL tasks are terminal, we can destroy the volume?
Because if (updated.tasksMap.values.forall(_.isTerminal)) -> then we do not suspend the reservation and from the comment you wrote I am not sure it is a bug or a correct behavior...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
ichernetsky: Yes. Please refer to the tickets. Probably `Instance.state.condition` will be dropped soon, though I am not sure yet. It is the last big refactoring related to PVs for pods, that's to be done yet. As it is implemented now, instance/task status updates handling is wrong, if an instance has two or more tasks. We could live with the current implementation in case of ephemeral pods, but since now we introduce support for PVs pods it's become serious, and due to some kind of regression on Mesos side, which actually made this race condition happened, I am posting here just a quick fix before starting the long-time-ago-planned task handling overhaul.

Basically, in case of stateful pods, Last-Write-Wins strategy is just wrong when it comes to handling terminal status updates.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
ichernetsky: Now, if all tasks are terminal, we do expunge an instance, but all tasks can be terminal here, only if the instance doesn't have PVs. Otherwise it is guaranteed that at least one of them is reserved.

Noted. I will avoid `.copy()`, when I will work on overhauling this part, but for now I will keep it as is in order to avoid testing one more time, because the fix has to be landed as soon as possible in order to include it into beta3

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: makes sense, thanks 👍

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
ichernetsky: BTW, please refer to `Task.update` method to the details on why it is guaranteed.

alenkacz: oh ok, can you maybe add a note that all tasks being terminal cannot happen for tasks with PV? It was not obvious to me at the first glance.

PR #5892, opened 2018-01-11

zen-dog:
for randomly generated ports correctly.

JIRA: COPS-2072

### DIFF
--- a/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
+++ b/src/main/scala/mesosphere/mesos/EnvironmentHelper.scala
@@ -87,11 +87,13 @@ object EnvironmentHelper {
           env += (s"PORT_$generatedPort" -> generatedPort.toString)
       }
 
-      requestedPorts.zip(effectivePorts).foreach {
-        case (PortRequest(Some(portName), _), Some(effectivePort)) =>
+      requestedPorts.zip(effectivePorts).zipWithIndex.foreach {
+        case ((PortRequest(Some(portName), _), Some(effectivePort)), _) =>
           env += (s"PORT_${portName.toUpperCase}" -> effectivePort.toString)
-        case (PortRequest(Some(portName), port), None) =>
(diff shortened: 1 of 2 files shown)


jdef:
lgtm

timcharper:
Should we sanitize these? Port names can contain hyphens. Container may fail to launch if we specify a port name containing a hyphen

zen-dog:
Port names are of type `strings.LegacyName` so `pattern: ^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`

jdef:
yes @timcharper the port names should be sanitized (if we're not already doing it), probably converting hyphens to underscore

jdef:
.. on second thought, I know we have customers that depend on **spaces** in envvar names. so sanitizing the port names in this case might constitute a breaking change if some customers are relying on hyphens in the port names

zen-dog:
I didn't see the hyphen in the regex. But as @jdef said - changing it means breaking existing app defs. I'll create a ticket to track this.

timcharper:
@zen-dog - there is definitely a hyphen allowed in the port name regex. Do we know for sure that Mesos won't fail if you try to launch a task with a hyphen in an env-var?

Maybe it'll be fine. But using spaces and hyphens in env vars seems like a pretty insane thing to want.

jdef:
apps don't use strict envvar name validation:
https://github.com/mesosphere/marathon/blob/3539856caa666a44702a1cbf41a285c9a321d014/src/main/scala/mesosphere/marathon/api/v2/validation/AppValidation.scala#L317

apps only require that envvar names are not empty:
https://github.com/mesosphere/marathon/blob/3539856caa666a44702a1cbf41a285c9a321d014/src/main/scala/mesosphere/marathon/api/v2/validation/EnvVarValidation.scala#L26

we went through some customer pain when I tried to increase strictness of
envvar names for apps. hence
https://github.com/mesosphere/marathon/commit/2b81cb4bfecdcceaa96ab4a09c69a3b503e25caf#diff-27cd0a75da12670feb9970f3e2cc20e5

PR #5893, opened 2018-01-11
Revert "use mesos nodes for auth step"

jeschkies:
Summary:
Since pull requests can override the Jenkinsfile the original commit is not adding any
security. Thus I propose to revert it.


jeschkies:
Dropping because we changed the trust policy of forked Jenkinsfiles.

PR #5894, opened 2018-01-11
Use block_iptable_rules_for_seconds.

jeschkies:
Summary:
One system integration test was still using the removed `save_iptables` and
`block_port` logic.

### DIFF
--- a/tests/system/marathon_pods_tests.py
+++ b/tests/system/marathon_pods_tests.py
@@ -487,7 +487,7 @@ def test_pod_health_failed_check():
     tasks = common.get_pod_tasks(pod_id)
     for new_task in tasks:
         new_task_id = new_task['id']
-        assert new_task_id != initial_id1, f"Task {task_id} has not been restarted"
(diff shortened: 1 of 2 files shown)


tests/system/marathon_pods_tests.py
```diff
@@ -481,16 +481,14 @@ def test_pod_health_failed_check():
...
-    common.restore_iptables(host)
+    common.block_iptable_rules_for_seconds(host, port, 7, block_input=True, block_output=False)
     common.deployment_wait(service_id=pod_id)
 
     tasks = common.get_pod_tasks(pod_id)
-    for task in tasks:
-        assert task['id'] != initial_id1, "One of the tasks has not been restarted"
-        assert task['id'] != initial_id2, "One of the tasks has not been restarted"
+    for new_task in tasks:
+        task_id = new_task['id']
```
alenkacz: since we renamed task to new_task this should also be new_task_id...

jeschkies: ✅

PR #5896, opened 2018-01-12
Make test clusters easier to be identified.

jeschkies:
Summary:
The current clusters use a unix timestamp as an identifier

`system-integration-tests-marathon-si-dcos-permissive-master--1515744121`

this change will use the build number instead

`system-integration-tests-marathon-si-dcos-permissive-master--42`


jeschkies:
Please note that the system integration test failed with three known issues.

PR #5899, opened 2018-01-12
Add documentation for Mesos maintenance mode

unterstein:
Add documentation for Mesos maintenance mode

Summary:
Pull request #5435 introduces simple support for mesos maintenance mode. This pull request adds more information about the implementation, the limitations and configuration instructions.

JIRA issues:
MARATHON-8006

### DIFF
--- a/docs/docs/maintenance-mode.md
+++ b/docs/docs/maintenance-mode.md
@@ -1,0 +1,24 @@
+---
+title: Maintenance Mode
+---
+
+# Maintenance Mode
+
+As of version 1.6, Marathon has simple built-in support for [Mesos Maintenance Primitives](http://mesos.apache.org/documentation/latest/maintenance/).
+
+Operators regularly need to perform maintenance tasks on machines that comprise a Mesos cluster. Most Mesos upgrades can be done without affecting running tasks, but there are situations where maintenance may affect running tasks. In order to meet Service Level Agreements or to ensure uninterrupted services for their end users, Marathon is respecting configured maintenance windows. 
+If this feature is enabled, Marathon is not scheduling tasks to agents currently within a maintenance window. Furthermore Marathon can decline offers before a maintenance window, if the according parameter is configured via command line argument.
+
+## How it works
+The current implementation is respecting configured maintenance windows from Mesos agents. If Marathon receives an offer with an included maintenance window, Marathon will check this agent is currently within this period or if the configured time before the window is reached. If one of this checks is true, Marathon will not use this offer to start a new task.
+
+## Limitations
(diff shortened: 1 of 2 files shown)


docs/docs/maintenance-mode.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Maintenance Mode
+
+As of version 1.6, Marathon has simple built-in support for [Mesos Maintenance Primitives](http://mesos.apache.org/documentation/latest/maintenance/).
+
+Operators regularly need to perform maintenance tasks on machines that comprise a Mesos cluster. Most Mesos upgrades can be done without affecting running tasks, but there are situations where maintenance may affect running tasks. In order to meet Service Level Agreements or to ensure uninterrupted services for their end users, Marathon is respecting configured maintenance windows. 
+If this feature is enabled, Marathon is not scheduling tasks to agents currently within a maintenance window. Furthermore Marathon can decline offers before a maintenance window, if the according parameter is configured via command line argument.
+
+## How it works
+The current implementation is respecting configured maintenance windows from Mesos agents. If Marathon receives an offer with an included maintenance window, Marathon will check this agent is currently within this period or if the configured time before the window is reached. If one of this checks is true, Marathon will not use this offer to start a new task.
```
alenkacz: I would link the mesos docs here again to make it obvious how to set up a maintenance window...

docs/docs/maintenance-mode.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Limitations
+As described, Marathon will not start tasks on certain offers. But Marathon will also not migrate tasks from agents close to a maintenance window to other agents. In case of an agent maintenance, Marathon would detect the unreachable agent after receiving the according Mesos task status updates. Marathon would then restart these unreachable tasks on other agents. If your application is not allowed to have downtime, you need to manually migrate the tasks.
```
alenkacz: Maybe we can make this part more explicit?

"In case of an agent maintenance" - In case an agent goes down for a planned maintenance ... ?
Maybe we can use more the names from mesos docs - so talk about draining phase (during which we just ignore offers but don't do an actual draining) and then down state during which we restart apps after we detect the agent down...

Just a suggestions...

docs/docs/maintenance-mode.md
```diff
@@ -0,0 +1,24 @@
...
+
+The efforts to enhance the current implementation are tracked in [this JIRA issue](https://jira.mesosphere.com/browse/MARATHON-3216).
+
+## How to configure
+This feature needs to be activated via command line argument. You need to add `maintenance_mode` to the set of `--enable_features` in your marathon startup arguments.
```
jeschkies: So it is not enabled by default?

unterstein: Thanks. No, I added docs, that it`s off by default.

docs/docs/maintenance-mode.md
unterstein: thanks, changed the docs in favor of your suggestion!

docs/docs/maintenance-mode.md
unterstein: yes!

PR #5901, opened 2018-01-14
ELK - Support loading a single file

timcharper:


tools/elk/README.md
```diff
@@ -13,8 +13,19 @@ Logstash config can be generated for a DCOS bundle using the following command:
...
 ```
 
-Command will attempt to detect log line format (include date, etc.), and generate the appropriate Logstash input
-directives such that the hostname field will be populated in Elasticsearch.
+Command will detect masters in the bundle, attempt to detect log line format (include date, etc.), and generate the
+appropriate Logstash input directives such that the hostname field will be populated in Elasticsearch.
+
+### Single file
+
+Logstash config can be generate for a single log file using the following command:
```
alenkacz: generated

tools/elk/lib/main.sc
```diff
@@ -46,7 +56,13 @@ def warningLineSplitter(file: Path, warnLength: Int) =
     line
   }
 
-def unzipAndStripLogs(masters: Map[String, Path])(implicit mat: Materializer): Map[String, Path] = {
+def stripAnsiFlow(inputPath: Path): Flow[ByteString, ByteString, NotUsed] =
+  Flow[ByteString]
+    .via(warningLineSplitter(inputPath, 128000))
```
alenkacz: what is this number?

timcharper: warns if line is longer than that

PR #5902, opened 2018-01-15
Verify SSL connection for events integration test with CA from DC/OS.

jeschkies:
Summary:
All requests that require an SSL verification fail because the CA file is not loaded. With this
change we'll download the certificate and provide it to the event stream client for strict clusters.

### DIFF
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -51,4 +51,4 @@ fi
 jq -r .ssh_private_key cluster_info.json > "$CLI_TEST_SSH_KEY"
 
 # Return dcos_url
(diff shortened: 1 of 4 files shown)


jeschkies:
>I'm wondering if any test failed because of that?
Yes. Checkout [test_event_channel_for_pods](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-strict/job/master/412/testReport/junit/system/test_marathon_root/test_event_channel_for_pods/).

Note, the system integration tests fail due to a dcos cli package download error. It is also related to SSL issues. Somehow the CLI is verifying the connection even though it's switched off.

jeschkies:
Sweet. All strict tests pass with this change!

PR #5903, opened 2018-01-15
Document proper usage of env cmd vars

alenkacz:
Summary:
In version 1.4 a breaking change was introduced for the way how env vars for CMD parameters are named and this was not documented so far. This PR adds this documentation. See this old ticket https://jira.mesosphere.com/browse/MARATHON_EE-569 and this code in `Main.scala` https://github.com/mesosphere/marathon/commit/f6fccc35f4bd6ce4dad2b43aab9b0f862299643a#diff-d9fbeb18eb73de0292d5767e79e5da8e


alenkacz:
if this is OK, I guess it makes sense to backport it to 1.5 branch as well? @unterstein

PR #5904, opened 2018-01-15
Block Mesos port on Marathon master for 100 seconds.

jeschkies:
Summary:
Currently a partition test blocks the Mesos port on the Marathon leader
host for 60 seconds. That is not enough to trigger 5 Mesos heartbeat
failures and a following leader election. Thus we set it to 100 seconds.

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -365,8 +365,7 @@ def install_enterprise_cli_package():
        command to create secrets, manage service accounts etc.
     """
     print('Installing dcos-enterprise-cli package')
-    stdout, stderr, return_code = shakedown.run_dcos_command('package install dcos-enterprise-cli --cli --yes')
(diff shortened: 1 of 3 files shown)


tests/system/common.py
```diff
@@ -365,8 +365,7 @@ def install_enterprise_cli_package():
        command to create secrets, manage service accounts etc.
     """
     print('Installing dcos-enterprise-cli package')
-    stdout, stderr, return_code = shakedown.run_dcos_command('package install dcos-enterprise-cli --cli --yes')
-    assert return_code == 0, "Failed to install dcos-enterprise-cli package"
+    stdout, stderr, return_code = shakedown.run_dcos_command('package install dcos-enterprise-cli --cli --yes', raise_on_error=True)
```
alenkacz: did you mean to commit this? it's not related to this issue... I guess it does not break anything either...

PR #5906, opened 2018-01-15
Script for manually running perf tests against DC/OS

wavesoft:
This commit introduces the `manual_run_dcos.sh` that can be used to run
the performance tests against an already deployed DC/OS cluster.

_Part of JIRA: DCOS-20295_

### DIFF
--- a/tests/performance/config/perf-driver/environments/target-cluster-custom.yml
+++ b/tests/performance/config/perf-driver/environments/target-cluster-custom.yml
@@ -1,10 +1,10 @@
 # ----------------------------------------------------------- #
-# Configuration Fragment : Cluster Environment                #
+# Configuration Fragment : Custom DC/OS Cluster Target        #
 # ----------------------------------------------------------- #
 # This fragment instructs the driver to use the marathon      #
-# exposed from an enterprise cluster. As part of it's setup   #
-# sequence, the driver will authenticate to the cluster with  #
-# the default superuser account.                              #
+# exposed from an enterprise cluster. In contrast to other    #
+# cluster targets it requires the auth token to be given by   #
+# the user.                                                   #
 # ----------------------------------------------------------- #
 
 # Global test configuration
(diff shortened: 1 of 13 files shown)


tests/performance/manual_run_dcos.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
jeschkies: Out of curiosity, did you run [shellcheck](https://github.com/koalaman/shellcheck) against this script?

tests/performance/config/perf-driver/environments/target-cluster-custom.yml
```diff
@@ -0,0 +1,36 @@
...
+# ----------------------------------------------------------- #
+# This fragment instructs the driver to use the marathon      #
+# exposed from an enterprise cluster. In contrast to other    #
+# cluster targets it requires the auth token to be given by   #
+# the user.                                                   #
+# ----------------------------------------------------------- #
+
+# Global test configuration
+# ===========================
+config:
```
jeschkies: Is this already interpreted by the per driver? Where is its implementation?

jeschkies:
@wavesoft, what is the state of this diff?

tests/performance/manual_run_dcos.sh
wavesoft: Nope, but did now and applied some interesting fixes 😄

tests/performance/config/perf-driver/environments/target-cluster-custom.yml
wavesoft: Yes, this is part of the [standard driver configuration](http://dcos-performance-test-driver.readthedocs.io/en/latest/config/Readme.html). 

The driver itself is supposed to be available in the user's environment before running the manual script.

wavesoft:
Key @jeschkies , for some reason your comments slipped through my e-mail filters 😄 I addressed your comments so it should be ready to merge.

I have used this script to run the tests on my local infrastructure, so the actual PR is not needed. I just thought it would be nice to have it on our toolbox.

PR #5907, opened 2018-01-15
Added clues for `ResidentTaskIntegrationTest`

zen-dog:
to help debug the flakes.

### DIFF
--- a/src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
@@ -147,7 +147,8 @@ class ResidentTaskIntegrationTest extends AkkaIntegrationTest with EmbeddedMarat
       scaleToSuccessfully(PathId(app.id), 5)
 
       Then("exactly 5 tasks have been created")
-      launchedTasks(PathId(app.id)).size shouldBe 5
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/marathon/integration/ResidentTaskIntegrationTest.scala
```diff
@@ -147,7 +147,8 @@ class ResidentTaskIntegrationTest extends AkkaIntegrationTest with EmbeddedMarat
       scaleToSuccessfully(PathId(app.id), 5)
 
       Then("exactly 5 tasks have been created")
-      launchedTasks(PathId(app.id)).size shouldBe 5
+      val all = allTasks(PathId(app.id))
+      all.count(_.launched) shouldBe 5 withClue (s"Found ${all.size}/5 tasks: ${all}")
```
alenkacz: don't you want to use launched count in the clue since that is what you are asserting?

zen-dog: Nope, that was on purpose. I actually want to see all tasks info.

alenkacz: yeah but you are expecting 5 launched tasks and then you are saying Found ${all.size}/5 while I think it should be ${all.count(_.launched)}/5

PR #5909, opened 2018-01-16
Backport a bunch of IT fixes around `PortAllocator`

zen-dog:
Summary:
A bunch of 1.5 commits were used here. It wasn't possible to cherry-pick them directly because of other changes to base IT classes missing so I had to add them manually:
- every test suite gets it's own port range
- `app_mock.py` got revisited with handling SIGTERM properly and reusing socket address
- `ForwardService` is killed gracefully


zen-dog:
I give up on this PR :( It just doesn't wont to become green. Feel free to pick up the flag @jeschkies @meln1k

PR #5910, opened 2018-01-16
 OffsetDateTime json serialization is different form Timestamp serialization

meln1k:
Summary: added a formatting to every usage of OffsetDateTime to keep string representation consistent across Timestamp and OffsetDateTime

JIRA issues: MARATHON-8005

### DIFF
--- a/src/main/scala/mesosphere/marathon/raml/DefaultConversions.scala
+++ b/src/main/scala/mesosphere/marathon/raml/DefaultConversions.scala
@@ -1,10 +1,13 @@
 package mesosphere.marathon
 package raml
 
+import java.time.OffsetDateTime
+
 import mesosphere.marathon.core.instance
 import mesosphere.marathon.state.{ PathId, Timestamp }
 import mesosphere.marathon.stream.Implicits._
 import org.apache.mesos.{ Protos => mesos }
+import play.api.libs.json.JsString
 
 import scala.collection.breakOut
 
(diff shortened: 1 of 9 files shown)


ichernetsky:
The failure is unrelated. Rebuilding...

src/main/scala/mesosphere/marathon/raml/DefaultConversions.scala
```diff
@@ -60,4 +63,9 @@ trait DefaultConversions {
       .getOrElse(throw new IllegalArgumentException(s"$name is an unknown protocol"))
     IpAddr(ipAddress.getIpAddress, protocol)
   }
+
+  implicit object OffsetDateTimeWrite extends play.api.libs.json.Writes[OffsetDateTime] {
+    def writes(o: OffsetDateTime) = JsString(o.format(Timestamp.formatter))
```
jeschkies: What about reads? Or is this only used for API models?

meln1k: Only for the API. I want to keep `Reads` default since they're not strict and therefore backwards compatible.

alenkacz:
btw. if I understand this correctly this is potentially a breaking change for someone who has some integration on top of our API and is asserting versions in a certain format or with a certain value (like SI tests or whatever)... just noting that :)

PR #5911, opened 2018-01-17
Make reservation ID generation backward-compatible with Marathon versions < 1.6

ichernetsky:
JIRA issues: https://jira.mesosphere.com/browse/MARATHON-8032

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Reservation.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Reservation.scala
@@ -12,6 +12,34 @@ import play.api.libs.json._
 case class Reservation(volumeIds: Seq[LocalVolumeId], state: Reservation.State)
 
 object Reservation {
+
+  /**
+    * Reservation.Id represents a reservation ID which is used in a task reservation labels.
+    */
+  case class Id(runSpecId: String, separator: String, prefix: Option[String], uuid: String) {
+    override def toString: String = prefix match {
+      case Some(prefixStr) => runSpecId + separator + prefixStr + uuid
+      case None => runSpecId + separator + uuid
+    }
+
+    def instanceId: Instance.Id = prefix match {
+      case Some(prefixStr) => Instance.Id(runSpecId + separator + prefixStr + uuid)
+      case None => Instance.Id(runSpecId + separator + "marathon-" + uuid)
+    }
+  }
+
+  object Id {
+    private val ReservationIdRegex = """^(.+)([\._])([^_\.]+)$""".r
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/core/instance/Reservation.scala
```diff
@@ -12,6 +12,34 @@ import play.api.libs.json._
...
+    * Reservation.Id represents a reservation ID which is used in a task reservation labels.
+    */
+  case class Id(runSpecId: String, separator: String, prefix: Option[String], uuid: String) {
+    override def toString: String = prefix match {
+      case Some(prefixStr) => runSpecId + separator + prefixStr + uuid
+      case None => runSpecId + separator + uuid
+    }
+
+    def instanceId: Instance.Id = prefix match {
+      case Some(prefixStr) => Instance.Id(runSpecId + separator + prefixStr + uuid)
```
alenkacz: I don't think that reservation id class should have such a deep knowledge of how instance id is concatenated from various string. I think this should definitely be somewhere near Instance.Id implementation

ichernetsky: It is implied by the definition of `Instance.Id`.
```
case class Id(idString: String) extends Ordered[Id]
```
It is scattered all over the code. Though I agree, it probably warrants a refactoring.

src/main/scala/mesosphere/marathon/core/instance/Reservation.scala
```diff
@@ -12,6 +12,34 @@ import play.api.libs.json._
...
+      case None => Instance.Id(runSpecId + separator + "marathon-" + uuid)
+    }
+  }
+
+  object Id {
+    private val ReservationIdWithInstanceIdRegex = """^(.+)\.(instance-|marathon-)([^_\.]+)$""".r
```
jeschkies: Why not reuse the pattern from `Task.Id`?

meln1k:
I wasn't able to test it locally due to my misbehaving mesos. If someone could try this test case, it would be awesome.

- Start the app from C.1 using Marathon 1.5
- Upgrade Marathon.
- Verify that the app is up and running, and Marathon has not taken any action towards it, and that the file contains only one line.

ichernetsky:
I verified all the C.* test cases yesterday.

src/main/scala/mesosphere/marathon/core/instance/Reservation.scala
ichernetsky: It is different here.

PR #5912, opened 2018-01-17
A few SI tests for apps and pods with persistent volumes

ichernetsky:
JIRA issues: https://jira.mesosphere.com/browse/MARATHON_EE-1792

### DIFF
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -761,7 +761,7 @@ def test_app_with_persistent_volume_recovers():
         run, data = shakedown.run_command_on_master(cmd)
 
         assert run, "{} did not succeed".format(cmd)
-        assert data == target_data, "'{}' was not equal to {}".format(data, target_data)
+        assert target_data in data, "'{}' not found in {}".format(target_data, data)
 
     check_task(cmd, target_data='hello\n')
 
@@ -775,7 +775,6 @@ def test_app_with_persistent_volume_recovers():
 
     check_task_recovery()
 
-    tasks = client.get_tasks(app_id)
     port = tasks[0]['ports'][0]
     host = tasks[0]['host']
     cmd = "curl {}:{}/data/foo".format(host, port)
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -700,18 +700,23 @@ def test_restart_container_with_persistent_volume():
     client = marathon.create_client()
     client.add_app(app_def)
 
-    shakedown.deployment_wait()
+    common.deployment_wait(service_id=app_id)
 
     tasks = client.get_tasks(app_id)
     assert len(tasks) == 1, "The number of tasks is {} after deployment, but 1 was expected".format(len(tasks))
 
-    port = tasks[0]['ports'][0]
     host = tasks[0]['host']
+    port = tasks[0]['ports'][0]
     cmd = "curl {}:{}/data/foo".format(host, port)
-    run, data = shakedown.run_command_on_master(cmd)
 
-    assert run, "{} did not succeed".format(cmd)
-    assert data == 'hello\n', "'{}' was not equal to hello\\n".format(data)
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
+    def check_task(cmd, target_data):
+        run, data = shakedown.run_command_on_master(cmd)
+
+        assert run, "{} did not succeed".format(cmd)
+        assert data == target_data, "'{}' was not equal to {}".format(data, target_data)
+
+    check_task(cmd, target_data='hello\n')
 
     client.restart_app(app_id)
     shakedown.deployment_wait()
@@ -726,10 +731,8 @@ def test_restart_container_with_persistent_volume():
     port = tasks[0]['ports'][0]
     host = tasks[0]['host']
     cmd = "curl {}:{}/data/foo".format(host, port)
-    run, data = shakedown.run_command_on_master(cmd)
 
-    assert run, "{} did not succeed".format(cmd)
-    assert data == 'hello\nhello\n', "'{}' was not equal to hello\\nhello\\n".format(data)
+    check_task(cmd, target_data='hello\nhello\n')
 
 
 @shakedown.dcos_1_8
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -747,7 +747,7 @@ def test_app_with_persistent_volume_recovers():
     client = marathon.create_client()
     client.add_app(app_def)
 
-    shakedown.deployment_wait()
+    common.deployment_wait(service_id=app_id)
 
     tasks = client.get_tasks(app_id)
     assert len(tasks) == 1, "The number of tasks is {} after deployment, but 1 was expected".format(len(tasks))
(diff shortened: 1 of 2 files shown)


jeschkies:
I triggered system integration tests for [open](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-pr/83/) and [strict](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-pr/84/) for this PR.

tests/system/marathon_common_tests.py
```diff
@@ -732,6 +732,50 @@ def check_task_recovery():
...
+       it recovers on the node it was launched on, and it gets attached
+       to the same persistent-volume."""
+
+    app_def = apps.persistent_volume_app()
+    app_id = app_def['id']
+
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    shakedown.deployment_wait()
```
jeschkies: Please wait for the specific app.

tests/system/marathon_common_tests.py
```diff
@@ -732,6 +732,50 @@ def check_task_recovery():
...
+
+    port = tasks[0]['ports'][0]
+    host = tasks[0]['host']
+    cmd = "curl {}:{}/data/foo".format(host, port)
+    run, data = shakedown.run_command_on_master(cmd)
+
+    assert run, "{} did not succeed".format(cmd)
+    assert data == 'hello\n', "'{}' was not equal to hello\\n".format(data)
+
+    shakedown.kill_process_on_host(host, '[h]ttp.server')
```
jeschkies: What is this doing? Is this killing the `persistent_volume_app` and triggers as `TASK_FAILED` event?

tests/system/marathon_common_tests.py
```diff
@@ -732,6 +732,50 @@ def check_task_recovery():
...
+    shakedown.deployment_wait()
+
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
+    def check_task_recovery():
+        tasks = client.get_tasks(app_id)
+        assert len(tasks) == 1, "The number of tasks is {} after recovery, but 1 was expected".format(len(tasks))
```
jeschkies: Should we rather trigger the state?

tests/system/marathon_pods_tests.py
```diff
@@ -505,18 +506,52 @@ def test_pod_with_persistent_volume():
...
     port2 = tasks[1]['discovery']['ports']['ports'][0]["number"]
-    dir1 = tasks[0]['container']['volumes'][0]['container_path']
-    dir2 = tasks[1]['container']['volumes'][0]['container_path']
-    print(host, port1, port2, dir1, dir2)
+    path1 = tasks[0]['container']['volumes'][0]['container_path']
+    path2 = tasks[1]['container']['volumes'][0]['container_path']
+    print(host, port1, port2, path1, path2)
 
-    time.sleep(1)
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
```
zen-dog: 👍  was about to do the same

tests/system/marathon_common_tests.py
zen-dog: I've added retrying for the state below.

tests/system/marathon_common_tests.py
zen-dog: Yes

jeschkies:
Two [tests fail](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-pr/83/) for open
* system.test_marathon_root.test_pod_with_persistent_volume_restarts
* system.test_marathon_root.test_app_with_persistent_volume_recovers

zen-dog:
I've pushed a few changes so don't forget to pull @ichernetsky (should there be a need for it). 
Another problem with the tests: `http.server` sometimes fails to restart because it can'b bind to the port (OS thinks the port is still in use) so you end up with 3x `hello` (`hello\nhello\nhello\n`) instead of 2x.

tests/system/marathon_pods_tests.py
```diff
@@ -505,18 +506,52 @@ def test_pod_with_persistent_volume():
...
+    def check_http_endpoint(port, path):
```
jeschkies: This is very similar to `check_task`. Should we unify then?

tests/system/marathon_pods_tests.py
```diff
@@ -505,18 +506,52 @@ def test_pod_with_persistent_volume():
...
+
+    tasks = common.get_pod_tasks(pod_id)
+    port1 = tasks[0]['discovery']['ports']['ports'][0]["number"]
+    port2 = tasks[1]['discovery']['ports']['ports'][0]["number"]
+    path1 = tasks[0]['container']['volumes'][0]['container_path']
+    path2 = tasks[1]['container']['volumes'][0]['container_path']
+    print(host, port1, port2, path1, path2)
+
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
+    def check_http_endpoint(port, path):
```
jeschkies: See above.

tests/system/marathon_common_tests.py
```diff
@@ -726,10 +731,56 @@ def check_task_recovery():
...
+    def check_task(cmd, target_data):
+        run, data = shakedown.run_command_on_master(cmd)
+
+        assert run, "{} did not succeed".format(cmd)
+        assert data == target_data, "'{}' was not equal to {}".format(data, target_data)
+
+    check_task(cmd, target_data='hello\n')
+
+    shakedown.kill_process_on_host(host, '[h]ttp.server')
+    common.deployment_wait(service_id=app_id)
```
jeschkies: I'm not sure the `TASK_FAILED` triggers a new deployment. Also, it is a race condition. The `deployment_wait` can succeed before the deployment even started. How about using the event stream instead? @zen-dog was really enthusiastic about this ;).

tests/system/marathon_pods_tests.py
zen-dog: Too soon.

tests/system/marathon_common_tests.py
zen-dog: Indeed, killing a task/instance does not trigger a deployment.

zen-dog: And this is a hacky way to kill tasks(instances in pod test): we could achieve the same with `DELETE v2/apps/{app_id}/tasks/{task_id}` and `DELETE /v2/pods/{id}::instances/{instance}` and there are appropriate `client` methods for it. However even that way there is still no deployment triggered.

ichernetsky: Yeah, you are right, guys, it doesn't. I will address it. On the other hand, killing a task using the Marathon API is not the same a killing a task process. So, it is not a hacky way. It is just what needs to be done.

zen-dog: How do you mean? Marathon will send a task kill request to mesos so what's the difference? 

The problem with current approach: you kill *all* `http.server`s on the agent *(actually you kill task's executor and then fail to kill the task itself since it's gone already)* which means this test can't run parallel to any other test. Not that we're doing this right now but a man can dream.

ichernetsky: What I mean is that if there is a pod of two tasks, and if Marathon asks Mesos to kill one of them, both of the tasks will transition to `TASK_KILLING` state, and then `TASK_KILLED`. On the other hand, if one of the tasks gets killed not by Mesos or just fails, it will transition to `TASK_FAILED`, and the other one will be killed by Mesos, and therefore the state transition for the other task will be `TASK_KILLING -> TASK_KILLED`. I understand the flaws of the current approach, but this is what I would like to have tested.

zen-dog: I understand the reasoning but still don't like the idea of killing all `http.server` on that machine. how about putting some uuid into the cmd and grepping for it? e.g. smth. like `f"echo {uuid4().hex} && python http.server..."`?

jeschkies:
Open is still failing `system.test_marathon_root.test_pod_with_persistent_volume_recovers`. Permissive has 38 errors!

ichernetsky:
@jeschkies there is a known issue, which causes the test to fail. For some reason, sometimes after recovery pod tasks disappear from `/v2/tasks` output. I will work on it either today or on Monday. For the time being, I am adding `wip` label to this PR.

jeschkies:
@ichernetsky, thanks a ton. I ran the test locally without issues. Are you tracking your efforts somewhere?

ichernetsky:
@jeschkies, yes, please refer to https://jira.mesosphere.com/browse/MARATHON-8045

jeschkies:
@ichernetsky, thanks. I marked this JIRA as being blocked.

ichernetsky:
It is stupid me. It turns out that `common.get_pod_tasks` function fetches tasks from Mesos, not Marathon. It was complaining about missing task statuses, and I was checking Marathon's `/v2/tasks`, which was obviously empty, because it returns only app tasks, but I realized it too late. I just added a function call to refresh statuses to `wait_for_status_network_info`. I re-ran SI tests for both open and permissive clusters. Also I ran the test a few dozen times manually. All green. My New Year Resolution is stop being stupid. @jeschkies, please approve it and merge it on Monday.

PR #5914, opened 2018-01-17
Introduce SI test for ipv6 healthchecks

alenkacz:
Summary:
DC/OS 1.11 introduces ipv6 healthchecks. This SI test verifies that these healthchecks work.

JIRA issues: MARATHON_EE-1766


tests/Pipfile
```diff
@@ -12,7 +12,7 @@ python_version = "3.6"
 
 [packages]
 
-shakedown = {git = "https://github.com/dcos/shakedown.git", ref = "1.4.11"}
+shakedown = {git = "https://github.com/dcos/shakedown.git", ref = "1.4.12"}
```
jeschkies: The `Pipfile.lock` has to be updated as well.

tests/system/apps/ipv6-healthcheck.json
```diff
@@ -0,0 +1,25 @@
+{
+  "id": "/test-ipv6",
+  "instances": 1,
+  "cmd": "nc -kl -p 8080 -e sh -c $'sleep 1; echo -e \"HTTP/1.1 200 OK\\r\\nContent-Length: 3\\r\\n\\r\\nHi\"'",
```
jeschkies: What is this script doing?

tests/system/marathon_common_tests.py
```diff
@@ -1133,3 +1133,31 @@ def http_output_check():
...
     http_output_check()
+
+
+@shakedown.dcos_1_11
+def test_ipv6_healthcheck():
+    """ There is new feature in DC/OS 1.11 that allows containers running on IPv6 network to be healthchecked from
+        Marathon. This tests verifies executing such healthcheck.
+    """
+    agents = shakedown.get_agents()
+    network_cmd = "sudo docker network create --driver=bridge --ipv6 --subnet=fd01::/64 mesos-docker-ipv6-test"
```
jeschkies: What is this doing?

tests/system/apps/ipv6-healthcheck.json
alenkacz: that's a super awesome netcat listening on port 8080. This took me the most time from the whole PR :D support for webservers with ipv6 is very poor in general so this ended up to be the easiest way to do so.

Do you have any idea how to clarify that for the readers?

tests/system/marathon_common_tests.py
alenkacz: to be able to run container on ipv6 I have to have ipv6 network. This creates the network. Then in the app json you can see that the container I am launching uses this network...

I'll add comment there

tests/Pipfile
alenkacz: I did commit that too after we talked yesterday

tests/system/apps/ipv6-healthcheck.json
zen-dog: A comment at `def ipv6_healthcheck()`?

tests/system/marathon_common_tests.py
```diff
@@ -1133,3 +1133,33 @@ def http_output_check():
...
+
+@shakedown.dcos_1_11
+def test_ipv6_healthcheck():
+    """ There is new feature in DC/OS 1.11 that allows containers running on IPv6 network to be healthchecked from
+        Marathon. This tests verifies executing such healthcheck.
+    """
+    agents = shakedown.get_agents()
+    # to test ipv6 healthcheck we need ipv6 network before launching the container
+    # the container launch then uses this network in the app definition json
+    network_cmd = "sudo docker network create --driver=bridge --ipv6 --subnet=fd01::/64 mesos-docker-ipv6-test"
```
zen-dog: This (and removing the network afterwards) should be done as a fixture. Otherwise the cleanup might not be executed.

alenkacz: good point, done

tests/system/apps/ipv6-healthcheck.json
alenkacz: done

tests/system/apps/__init__.py
```diff
@@ -81,3 +81,7 @@ def fake_framework():
 
 def external_volume_mesos_app():
     return load_app('external-volume-mesos-app')
+
+def ipv6_healthcheck():
+    """ The app uses netcat to listen on port. It uses alpine image which has netcat with ipv6 support by default. """
```
zen-dog: ... I meant a comment explaining what the `nc...` command does :)

tests/system/common.py
```diff
@@ -785,3 +786,15 @@ def task_by_name(tasks, name):
 
 async def assert_event(event_type, event_stream, within=10):
     await asyncio.wait_for(find_event(event_type, event_stream), within)
+
+
+@contextlib.contextmanager
+def docker_ipv6_network(network_name):
```
zen-dog: definitely should have `_fixture` in the name (e.g. `docker_ipv6_network_fixture`)

tests/system/marathon_common_tests.py
```diff
@@ -1133,3 +1133,25 @@ def http_output_check():
         assert 'Relay from {}'.format(relay_app["id"]) in output
 
     http_output_check()
+
+
+@shakedown.dcos_1_11
+def test_ipv6_healthcheck():
```
zen-dog: I would pass the fixture directly to the test e.g. `def test_ipv6_healthcheck(docker_ipv6_network_fixture)`

tests/system/apps/ipv6-healthcheck.json
jeschkies: Ah. Now I get it :).

PR #5918, opened 2018-01-20
 Report execution times of update requests.

jeschkies:
Summary:
We want to have a better understand where time is psent when the root
group is updated. This patch introduces measurements for most sub-steps.

The log line looks like

```
DeploymentPlanId=... ChangeRootGroup=0ms,..., MakeDeploymentPlan=0ms, StoreRootGroup=20ms
```


jeschkies:
This is just a proposal. Kamon should be able to give us the same reporting but I fail to see how. Also, I'd like to get this into 1.6 as it should ease debugging quite a bit.

jeschkies:
Let's drop it for now and revisit later.

PR #5919, opened 2018-01-22
Abdicate leadership via API after a delay

timcharper:
This brings back behavior we previously had with Marathon 1.5.

When abdication is requested via the API, if marathon exits too quickly, it will shut down before the HTTP response is delivered to the client, leading in the appearance of a failed request.

JIRA Issues: MARATHON-8037

### DIFF
--- a/docs/docs/rest-api/public/api/v2/leader.raml
+++ b/docs/docs/rest-api/public/api/v2/leader.raml
@@ -50,6 +50,10 @@ delete:
       required: false
       description: URI to restore zk state.
       example: file:///var/backup/marathon/backup
+    delay?:
+      required: false
+      description: Delay, in milliseconds, after which to abdicate leadership (default: 500)
+      example: "1000"
   responses:
     200:
       description: The abdication message from the current leader.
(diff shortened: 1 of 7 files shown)


docs/docs/rest-api/public/api/v2/leader.raml
```diff
@@ -50,14 +50,18 @@ delete:
       required: false
       description: URI to restore zk state.
       example: file:///var/backup/marathon/backup
+    delay?:
```
alenkacz: wouldn't seconds be better for this?

Also I would go with delaySeconds instead of delay. Looking at HealthCheck raml definition, this is how we named properties in the past (and from the user perspective it is actually convenient too, because you need to know the unit anyway)

Also shouldn't the default be 0 to be backward compatible?

timcharper: The value for 1.5 was 500ms

https://github.com/mesosphere/marathon/blob/releases/1.5/src/main/scala/mesosphere/marathon/core/election/impl/CuratorElectionService.scala#L50

After some consideration, I decided that modifying the API was unnecessary complexity. So, I just reverted that part.

PR #5920, opened 2018-01-22
Adding JSON serialise/deserialise benchmarks

wavesoft:
This commit introduces the mesosphere.marathon.json.JsonSerializeDeserializeBenchmark
that measures the JSON serialisation & de-serialisation performance using real-world
datasets.

To run this use:
```
sbt "benchmark/jmh:run mesosphere.marathon.json.JsonSerializeDeserializeBenchmark"
```


benchmark/src/main/scala/mesosphere/marathon/json/JsonSerializeDeserializeBenchmark.scala
```diff
@@ -0,0 +1,97 @@
...
+  def jsonParse(hole: Blackhole): Unit = {
+    val value: JsValue = Json.parse(jsonMockContents)
+    hole.consume(value)
+  }
+
+  @Benchmark
+  def jsonParseDeserialise(hole: Blackhole): Unit = {
+    val value: JsValue = Json.parse(jsonMockContents)
+    val groupUpdate: raml.GroupUpdate = Json.fromJson[raml.GroupUpdate](value).get
+    hole.consume(groupUpdate)
```
jeschkies: If I'm not mistaken `value` should be consumed as well.

wavesoft: AFAIK `consume` is used to avoid JVM run-time and compile-time optimisations. Therefore it's enough to just consume the last dangling variable, since `value` is already used in order to produce a value for `groupUpdate`

jeschkies: Ah, thanks for clarifying.

wavesoft:
@unterstein that's actually a good idea for a test utility function. I will work on adding this. If this PR is still open by the time I am done, I will add it here and remove the mocks 👍

PR #5928, opened 2018-01-24
Add license when launching SI tests

alenkacz:
Summary:
License was probably made mandatory for EE versions of DC/OS as was announced by email.
Strict tests passing here https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-strict/job/av%252Flicense/3/console

JIRA issues: MARATHON_EE-1830

### DIFF
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -33,6 +33,8 @@ deployment_name: $DEPLOYMENT_NAME
 provider: aws
 aws_region: us-west-2
 key_helper: true
+dcos_config:
+    license_key_contents: $DCOS_LICENSE
 template_parameters:
     DefaultInstanceType: m4.large
     AdminLocation: 0.0.0.0/0
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -41,6 +41,11 @@ template_parameters:
     LicenseKey: $DCOS_LICENSE
 EOF
 
+if [[ -z `curl $TEMPLATE | grep 'LicenseKey'` ]]; then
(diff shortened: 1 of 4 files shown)


ci/launch_cluster.sh
```diff
@@ -38,9 +38,12 @@ template_parameters:
     AdminLocation: 0.0.0.0/0
     PublicSlaveInstanceCount: 1
     SlaveInstanceCount: 5
+    LicenseKey: $DCOS_LICENSE
 EOF
 
-if ! ./dcos-launch create; then
+./dcos-launch create
```
alenkacz: I did this because dcos-launch was failing for me with validation error but the output of dcos-launch was not in the output of jenkins run

I saw only this:
./dcos-launch create -L debug
+ exit 2

zen-dog: strange...

ci/launch_cluster.sh
```diff
@@ -38,9 +38,12 @@ template_parameters:
     AdminLocation: 0.0.0.0/0
     PublicSlaveInstanceCount: 1
     SlaveInstanceCount: 5
+    LicenseKey: $DCOS_LICENSE
```
jeschkies: What is this is empty, e.g. with open?

alenkacz: empty string should be ok, that is actually the default, but the open template might not contain this param - I hope it will just ignore it

jeschkies:
`Parameters: [LicenseKey] do not exist in the template`. Hm, could you fix that @alenkacz?

PR #5931, opened 2018-01-25
Create the migration flag only while migrations are performed

ichernetsky:
Currently, the flag is created unconditionally for the duration of
checking the current version in ZK and performing migrations if necessary.
There is no need to create the flag for the duration of checking
the current version, it should be created only if and when there is
a need to perform migrations.

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-8044

### DIFF
--- a/src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
+++ b/src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
@@ -132,9 +132,10 @@ class Migration(
   def runMigrations(version: Protos.StorageVersion, backupCreated: Boolean = false): Future[Seq[StorageVersion]] =
     async {
       if (!backupCreated && config.backupLocation.isDefined) {
-        logger.info("Backup current state")
+        logger.info("Making a backup of the current state")
         await(backup.backup(config.backupLocation.get))
-        logger.info("Backup finished. Apply migration.")
+        logger.info(s"The backup has been saved to ${config.backupLocation}")
+        logger.info("Going to apply the migration steps.")
       }
       val result = await(applyMigrationSteps(version))
       await(storeCurrentVersion())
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -119,32 +119,28 @@ class Migration(
...
   def migrate(): Seq[StorageVersion] =
     Await.result(migrateAsync(), Duration.Inf)
 
+  @SuppressWarnings(Array("all")) // async/await
+  def runMigrations(version: Protos.StorageVersion, backupCreated: Boolean = false): Future[Seq[StorageVersion]] =
+    async {
+      if (!backupCreated && config.backupLocation.isDefined) {
+        logger.info("Backup current state")
+        await(backup.backup(config.backupLocation.get))
+        logger.info("Backup finished. Apply migration.")
```
zen-dog: I'd definitely add where the backup was created e.g. `logger.info("Backup successfully saved to ${config.backupLocation}...`.

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -161,20 +157,30 @@ class Migration(
...
-          val result = await(applyMigrationSteps(version))
-          await(storeCurrentVersion())
-          result
         case Some(version) if version == currentBuildVersion =>
           logger.info("No migration necessary, already at the current version")
           Nil
         case _ =>
           logger.info("No migration necessary, no version stored")
+          // mark migration as started
+          await(persistenceStore.startMigration())
```
zen-dog: Since storing storage version is part of the migration?

ichernetsky: Since the day one, I guess. Since the day this was placed into `Migration.scala`. The flag serves as a guard for any write operation related to versioning and data layouts.

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
ichernetsky: Done.

PR #5932, opened 2018-01-25
First step in making launch cluster run on mac

alenkacz:
Summary:
Do not use apt on mac, use homebrew instead.

JIRA issues:


alenkacz:
the SI test failure is because of failed test, not because of cluster launch failure "system.test_marathon_universe.test_install_marathon"

PR #5933, opened 2018-01-25
Serialize full and light events only once.

jeschkies:
Summary:
Currently each SSE handle is serializing the events. This
is a small overhead. It should be enough to serialize them
once and then broadcast to each handle actor.

The next step is to replace the HttpEventStreamHandleActor
with an Akka stream so that we do not have to deal with
slow consumers.

JIRA issues: MARATHON-8035


jeschkies:
This supersedes #5927.

PR #5937, opened 2018-01-25
Fix multiple jenkins failure reports

zen-dog:
by removing the unnecessary reporter.


zen-dog:
See the last 2 failed builds for how the test results look with this change

PR #5939, opened 2018-01-26
Complete deployment after it was removed to avoid race condition.

jeschkies:
Summary:
The `DeploymentActor` is sending a `DeploymentFinishd` messaged to the
`DeploymentManagerActor` *and* fulfills a promis for the
`MarathonSchedulerActor` which triggers events. This can lead to a race
condition. The scheduler might anounce a finished deployment while it's
still in the queue of active deployments. This change has the manager
fulfill the promise *after* the deployment was removed from the queue.
Thus we avoid the race condition.

JIRA issues: MARATHON-8039

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentManagerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentManagerActor.scala
@@ -150,7 +150,7 @@ class DeploymentManagerActor(
       runningDeployments.remove(plan.id).map { deploymentInfo =>
         logger.info(s"Removing ${plan.id} for ${plan.targetIdsString} from list of running deployments")
         deploymentStatus -= plan.id
-        deploymentRepository.delete(plan.id) // TODO: wait for future?
+        deploymentRepository.delete(plan.id)
         deploymentInfo.promise.complete(result)
       }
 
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentManagerActor.scala
```diff
@@ -145,11 +146,12 @@ class DeploymentManagerActor(
...
       sender() ! cancelDeployment(plan.id)
 
-    case DeploymentFinished(plan) =>
-      runningDeployments.remove(plan.id).map { _ =>
+    case DeploymentFinished(plan, result) =>
+      runningDeployments.remove(plan.id).map { deploymentInfo =>
         logger.info(s"Removing ${plan.id} for ${plan.targetIdsString} from list of running deployments")
         deploymentStatus -= plan.id
-        deploymentRepository.delete(plan.id)
+        deploymentRepository.delete(plan.id) // TODO: wait for future?
```
jeschkies: Should it stay as a fire and forget?

zen-dog: No, I've built `DeploymentManagerActor` to *not* block. Let's keep it this way pls

jeschkies: I would not block but fulfill the promise the the deployment was deleted. Let's keep it as it is for now.

PR #5949, opened 2018-01-29
Adding SI tests for Hybrid Cloud

wavesoft:
This PR introduces hybrid-cloud System Integration tests:

- `marathon_common_tests.py::test_faultdomains_default` : Checks the behaviour of launching an app without fault domain constraints
- `marathon_common_tests.py::test_faultdomains_region_only` : Checks the behaviour of launching an app with only `region` fixed 
- `marathon_common_tests.py::test_faultdomains_region_and_zone` : Checks the behaviour of launching an app with `region` and `zone` fixed
- `marathon_common_tests.py::test_faultdomains_distribute` : Checks the behaviour of the `GROUP_BY` constraints

These tests are discovering the zones and regions of the cluster by asking the cluster's mesos leader. When launching a DC/OS cluster make sure that there is at least 1 remote region, bigger than the local region, having at least 2 zones each. For example:

```yaml
fault_domain_helper:
  LocalZone:
    num_zones: 2
    num_private_agents: 2
    local: true
  RemoteZone:
    num_zones: 3
    num_public_agents: 1
    num_private_agents: 2
```

## WIP Items

* [ ] Modify `dcos-launch` to start a multi-zone cluster on EE by default
* [ ] Figure out other constraints that needs to be

### DIFF
--- a/tests/system/apps/__init__.py
+++ b/tests/system/apps/__init__.py
@@ -89,3 +89,34 @@ def ipv6_healthcheck():
         For more information about the nc options just run `docker run alpine nc --help`
     """
     return load_app('ipv6-healthcheck')
+
+def faultdomain_app(region=None, zone=None, instances=1, constraints=[]):
+    """
+    This is a dynamic app definition based on the faultdomain-base-app. It modifies it by appending
+    the name, zone and region configuration as given
+
+    :param region: The region to start this app in
+    :param zone: The zone to start this app in
+    :param instances: The number of instances in this app
+    :param constraints: Other constraints to append
+    :return: Returns the App Definition
+    """
+    app = load_app('faultdomain-base-app')
+    app['instances'] = instances
+
+    # Append region constraint
+    if not region is None:
+        app['constraints'] += [
+            ["@region", "IS", str(region)]
+        ]
+
+    # Append zone constraint
+    if not zone is None:
+        app['constraints'] += [
+            ["@zone", "IS", str(zone)]
+        ]
+
+    # Append misc constraints
+    app['constraints'] += constraints
+
+    return app
--- a/tests/system/apps/faultdomain-base-app.json
+++ b/tests/system/apps/faultdomain-base-app.json
@@ -1,0 +1,8 @@
+{
+  "id": "/fault-domain",
+  "cmd": "sleep 1200",
+  "instances": 1,
+  "cpus": 0.01,
+  "mem": 32,
+  "constraints": []
+}
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -16,6 +16,7 @@ from distutils.version import LooseVersion
 from json.decoder import JSONDecodeError
 from shakedown import marathon
 from urllib.parse import urljoin
+from utils import get_cluster_slave_domains, get_used_regions_and_zones, get_app_domains
 
 
 marathon_1_3 = pytest.mark.skipif('marthon_version_less_than("1.3")')
@@ -780,6 +781,71 @@ def task_by_name(tasks, name):
 
     assert False, "Did not find task with name %s in this list of tasks: %s" % (name, tasks,)
 
+def assert_app_in_all_domains(app, regions=None, zones=None):
+    """Asserts that all tasks of the given app are within all of the given region and/or zone
+    :param app_id: The app instance, as obtained by `marathon_client.get_app(...)`
+    :param region: The regions to test (can be None, String, List or Tuple)
+    :param zone: The zones to test (can be None, String, List or Tuple)
+    """
+
+    if type(regions) is str:
+        regions = [regions]
+    if type(zones) is str:
+        zones = [zones]
+
+    # Sanitize agent domains and user input
+
+    slave_domains = get_cluster_slave_domains()
+
+    assert len(slave_domains) > 0, "Did not find any agents in the DC/OS cluster" % (app['id'],)
+
+    if not regions is None:
+        slave_regions = set(map(lambda x: x.region, slave_domains.values()))
+        for _region in regions:
+            if not _region in slave_regions:
+                assert False, "Region %s was not found in the cluster (expecting one of %s)" \
+                              % (_region, ', '.join(slave_regions))
+    if not zones is None:
+        slave_zones = set(map(lambda x: x.zone, slave_domains.values()))
+        for _zone in zones:
+            if not _zone in slave_zones:
+                assert False, "Zone %s was not found in the cluster (expecting one of %s)" \
+                              % (_zone, ', '.join(slave_zones))
+
+    # Check if regions and/or zones overlap completely
+
+    (used_regions, used_zones) = get_used_regions_and_zones(get_app_domains(app))
+
+    if not regions is None:
+        assert used_regions == set(regions), \
+            "Application %s is not running on all the exepected regions ({%s} instead of {%s})" % (
+                app['id'], ', '.join(used_regions), ', '.join(regions)
+            )
+
+    if not zones is None:
+        assert used_zones == set(zones), \
+            "Application %s is not running on all the exepected zones ({%s} instead of {%s})" % (
+                app['id'], ', '.join(used_zones), ', '.join(zones)
+            )
+
+
+def assert_app_in_some_domains(app, region=None, zone=None):
+    """Asserts that any of the tasks of the given app are within all of the given region and/or zone
+    """
+
+    for task in tasks:
+        slave_domain = slave_domains[task['slaveId']]
+
+        if regions is not None:
+            assert slave_domain.region in regions, \
+                "Application %s has a task (%s) that runs in the wrong region (%s), expecting: %s" % (
+                    app['id'], task['id'], slave_domain.region, ', '.join(regions)
+                )
+        if zones is not None:
+            assert slave_domain.zone in zones, \
+                "Application %s has a task (%s) that runs in the wrong zone (%s), expecting: %s" % (
+                    app['id'], task['id'], slave_domain.zone, ', '.join(zones)
+                )
 
 async def find_event(event_type, event_stream):
     async for event in event_stream:
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -13,6 +13,8 @@ import time
 
 from datetime import timedelta
 from dcos import http, marathon
+from utils import get_cluster_local_domain, count_agents_in_faultdomains, get_biggest_cluster_region, get_app_domains, \
+    get_used_regions_and_zones
 from shakedown import dcos_version_less_than, marthon_version_less_than, required_private_agents # NOQA
 
 
@@ -1153,4 +1155,143 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
     assert app['tasksHealthy'] == target_instances_count, \
         "The number of healthy tasks is {}, but {} was expected".format(app['tasksHealthy'], target_instances_count)
 
+    client.remove_app(app['id'], True)
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+def test_faultdomains_default():
+    """Tests if the applications that do not have a fault domain specified are launched on the same
+       fault domain as the master.
+    """
+
+    # Discover which is the "local" cluster domain
+    local_domain = get_cluster_local_domain()
+
+    # Get how many agents do we have in the local domain (zone and region) and try to start
+    # that many applications. Mesos should round-robbin the instances in the region
+    region_agents = count_agents_in_faultdomains(regions=local_domain.region)
+
+    # Create an app with default configuration
+    app_def = apps.faultdomain_app(instances=region_agents)
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
+
+    # Ensure that all instances are running in the same region (it's ok if they belong on different zones)
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(local_domain.region,))
+
+    # Remove app
+    client.remove_app(app['id'], True)
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+def test_faultdomains_region_only():
+    """Tests if the applications that only have a `region` specified are launched on all the different
+       zones available in that region on the cluster
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region (%s)" % (
+        region,
+    )
+
+    # Create an app pinned on the given region, with at least double the number of instances than
+    # the zones in the cluster, to ensure that they are distributed among all of them
+    app_def = apps.faultdomain_app(region=region, instances=len(zones) * 2)
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
+
+    # Ensure that all instances are running in the same region, and on *all* the region zones
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(region,), zones=zones)
+
+    # Remove app
+    client.remove_app(app['id'], True)
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+def test_faultdomains_region_and_zone():
+    """Tests if the applications with a `region` and a `zone` defined are always starting on the
+       same region and zone
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region (%s)" % (
+        region,
+    )
+
+    single_zone = zones[0]
+
+    # Create an app pinned on the given region and zone.
+    #
+    # (We are starting more instances than the number of zones in the region in order to increase
+    #  the chance of a faulty instance to appear on a different zone)
+    #
+    app_def = apps.faultdomain_app(region=region, zone=single_zone, instances=len(zones) * 2)
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
+
+    # Ensure that all instances are running in the same region, and on *all* the region zones
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(region,), zones=(single_zone,))
+
+    # Remove app
+    client.remove_app(app['id'], True)
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+def test_faultdomains_distribute():
+    """Tests if the applications with a `region` and ["@zone", "GROUP_BY", "n"] specified are correctly
+       distributed evenly across "n" zones
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region (%s)" % (
+        region,
+    )
+
+    # Pick a number smaller than the total number of zones
+    group_num = round(len(zones) / 2)
+
+    # Create an app pinned on the given region, yet grouping by the number above
+    # We are also using instances 4 * group_num, effectively having 4 apps running in each of the above zones
+    # and therefore increase the chance of something going wrong.
+    app_def = apps.faultdomain_app(
+        region=region,
+        instances=group_num * 4,
+        constraints=[["@zone", "GROUP_BY", str(group_num)]]
+    )
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
+
+    # Check if the application was launched correctly
+    app = client.get_app(app_def['id'])
+    (regions, zones) = get_used_regions_and_zones(get_app_domains(app))
+
+    assert regions == set([region]), "App %s was launced in the wrong region(s): %s instead of %s" % (
+            app['id'], ', '.join(regions), region
+        )
+    assert len(zones) == group_num, "App %s was launced more zones than expected: %d (%s) instead of %d" % (
+        app['id'], len(zones), ', '.join(zones), group_num
+    )
+
+    # Remove app
     client.remove_app(app['id'], True)
--- a/tests/system/utils.py
+++ b/tests/system/utils.py
@@ -1,7 +1,7 @@
 import os
 import uuid
 
-from dcos import http, util
+from dcos import http, util, mesos
 from dcos.errors import DCOSException
 
 
@@ -39,3 +39,114 @@ def get_resource(resource):
                 raise Exception
         except Exception:
             raise DCOSException("Can't read from resource: {0}. Please check that it exists.".format(resource))
+
+class FaultDomain:
+    """
+    High-level representation of a fault domain
+    """
+    def __init__(self, config):
+        # Make sure config is a dict
+        if not type(config) is dict:
+            config = {}
+
+        # Extract fault domain
+        fault_domain = config.get('fault_domain', {})
+
+        # Extract zone/region
+        self.zone = fault_domain.get('zone', {}).get('name', 'default')
+        self.region = fault_domain.get('region', {}).get('name', 'default')
+
+
+def get_cluster_local_domain():
+    """Contacts the DC/OS mesos master and returns it's faultDomain configuration (aka "local domain")
+    """
+    master = mesos.get_master()
+    return FaultDomain(master.state().get('domain', {}))
+
+def get_cluster_slave_domains():
+    """Returns a dictionary with the slave IDs in the cluster and their corresponding
+       fault domain information
+    """
+    slave_domains = {}
+
+    # Populate slave_domains with the ID and the corresponding domain for each slave
+    master = mesos.get_master()
+    for slave in master.slaves():
+        slave_domains[slave['id']] = FaultDomain(slave._short_state.get('domain', None))
+    return slave_domains
+
+def get_all_cluster_regions():
+    """Returns a dictionary with all the regions and their zones in the cluster
+    """
+    domain_regions = {}
+
+    # Populate all the domain zones and regions found in the cluster
+    for domain in get_cluster_slave_domains().values():
+        if not domain.region in domain_regions:
+            domain_regions[domain.region] = []
+        if not domain.zone in domain_regions[domain.region]:
+            domain_regions[domain.region].append(domain.zone)
+
+    # Return dictionary
+    return domain_regions
+
+def get_biggest_cluster_region():
+    """Returns a tuple with the name of the biggest region in the cluster and the zones in it
+    """
+    biggest_region = None
+    biggest_region_zones = []
+
+    for (region, zones) in get_all_cluster_regions().items():
+        if len(zones) > len(biggest_region_zones):
+            biggest_region_zones = zones
+            biggest_region = region
+
+    return (biggest_region, biggest_region_zones)
+
+def get_app_domains(app):
+    """Returns a list of all te fault domains used by the tasks of the specified app
+    """
+    tasks = app.get('tasks', [])
+    slave_domains = get_cluster_slave_domains()
+
+    assert len(tasks) > 0, "App %s did not launch any tasks on mesos" % (app['id'],)
+
+    # Collect the FaultDomain objects from all the agents where the tasks are running
+    domains = []
+    for task in tasks:
+        domains.append(slave_domains[task['slaveId']])
+
+    return domains
+
+def get_used_regions_and_zones(domains):
+    """Returns tuple with the sets of the used regions and zones from the given list
+       of FaultDomain object instances (ex. obtained by get_app_domains())"""
+    used_regions = set()
+    used_zones = set()
+
+    for domain in domains:
+        used_regions.add(domain.region)
+        used_zones.add(domain.zone)
+
+    return (used_regions, used_zones)
+
+def count_agents_in_faultdomains(regions=None, zones=None):
+    """Return the number of agents that belong on the specified region and/or zone
+    """
+    counter = 0
+
+    # Make sure to always operate on iterable
+    if type(regions) is str:
+        regions = [regions]
+    if type(zones) is str:
+        zones = [zones]
+
+    # Increment counter for the agents that pass the checks
+    for domain in get_cluster_slave_domains().values():
+        if not regions is None and not domain.region in regions:
+            continue
+        if not zones is None and not domain.zone in zones:
+            continue
+        counter += 1
+
+    return counter


tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
         "The number of healthy tasks is {}, but {} was expected".format(app['tasksHealthy'], target_instances_count)
 
     client.remove_app(app['id'], True)
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
```
alenkacz: do faultdomains really work in EE only?

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+
+    # Create an app with default configuration
+    app_def = apps.faultdomain_app(instances=region_agents)
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
+
+    # Ensure that all instances are running in the same region (it's ok if they belong on different zones)
```
alenkacz: belong to

tests/system/marathon_common_tests.py
wavesoft: I think I read it somewhere. Let me verify.

tests/system/marathon_common_tests.py
wavesoft: 👍

Jenkinsfile.disabled.si
```diff
@@ -27,7 +27,9 @@ node('py36') {
             [$class: 'FileBinding', credentialsId: '11fcc957-5156-4470-ae34-d433da88248a', variable: 'DOT_SHAKEDOWN'],
             [$class: 'StringBinding', credentialsId: 'ca159ad3-7323-4564-818c-46a8f03e1389', variable: 'DCOS_LICENSE'],
             [$class: 'StringBinding', credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d', variable: 'DOCKER_HUB_USERNAME'],
-            [$class: 'StringBinding', credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD']
+            [$class: 'StringBinding', credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD'],
+            [$class: 'StringBinding', credentialsId: 'ca159ad3-7323-4564-818c-46a8f03e1389', variable: 'DCOS_LICENSE'],
```
jeschkies: This seems like a duplicate.

Jenkinsfile.permissive.si
```diff
@@ -27,7 +27,9 @@ node('py36') {
             [$class: 'FileBinding', credentialsId: '11fcc957-5156-4470-ae34-d433da88248a', variable: 'DOT_SHAKEDOWN'],
             [$class: 'StringBinding', credentialsId: 'ca159ad3-7323-4564-818c-46a8f03e1389', variable: 'DCOS_LICENSE'],
             [$class: 'StringBinding', credentialsId: '7bdd2775-2911-41ba-918f-59c8ae52326d', variable: 'DOCKER_HUB_USERNAME'],
-            [$class: 'StringBinding', credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD']
+            [$class: 'StringBinding', credentialsId: '42f2e3fb-3f4f-47b2-a128-10ac6d0f6825', variable: 'DOCKER_HUB_PASSWORD'],
+            [$class: 'StringBinding', credentialsId: 'ca159ad3-7323-4564-818c-46a8f03e1389', variable: 'DCOS_LICENSE'],
```
jeschkies: This seems like a duplicate.

ci/launch_cluster.sh
```diff
@@ -13,13 +17,22 @@ CHANNEL="$1"
 VARIANT="$2"
 
 if [ "$VARIANT" == "open" ]; then
-  TEMPLATE="https://s3.amazonaws.com/downloads.dcos.io/dcos/${CHANNEL}/cloudformation/multi-master.cloudformation.json"
+  INSTALLER="https://downloads.dcos.io/dcos/${CHANNEL}/dcos_generate_config.sh"
+  EXTRA_DCOS_CONFIG=""
 else
-  TEMPLATE="https://s3.amazonaws.com/downloads.mesosphere.io/dcos-enterprise-aws-advanced/${CHANNEL}/${VARIANT}/cloudformation/ee.multi-master.cloudformation.json"
+  if [ -z "$DCOS_INSTALLER_TEMPLATE" ]; then
+      echo "Expecting DCOS_INSTALLER_TEMPLATE environment variable"
```
jeschkies: Should we fail here?

ci/launch_cluster.sh
```diff
@@ -13,13 +17,22 @@ CHANNEL="$1"
...
+  fi
+  if [ -z "$DCOS_LICENSE" ]; then
+      echo "Expecting $DCOS_LICENSE environment variable"
+  fi
+  # Replace {} in the template with the channel
+  INSTALLER=${$DCOS_INSTALLER_TEMPLATE/{\}/$CHANNEL}
```
jeschkies: How about using sed?

ci/launch_cluster.sh
```diff
@@ -33,24 +46,36 @@ wget 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' && chmod
 envsubst <<EOF > config.yaml
 ---
 launch_config_version: 1
-template_url: $TEMPLATE
 deployment_name: $DEPLOYMENT_NAME
-provider: aws
+platform: aws
+instance_type: m4.xlarge
```
jeschkies: Why `m4.xlarge` and not `m4.large`?

ci/launch_cluster.sh
```diff
@@ -33,24 +46,36 @@ wget 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' && chmod
...
+os_name: cent-os-7-dcos-prereqs
 aws_region: us-west-2
+aws_key_name: default
+provider: onprem
+installer_url: $INSTALLER
```
jeschkies: How about using `$INSTALLER_URL` instead of `$INSTALLER`?

ci/launch_cluster.sh
```diff
@@ -33,24 +46,36 @@ wget 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' && chmod
...
-    LicenseKey: $DCOS_LICENSE
+num_masters: 3
+fault_domain_helper:
+  LocalZone:
+    num_zones: 2
+    num_public_agents: 1
+    num_private_agents: 2
+    local: true
+  RemoteZone:
+    num_zones: 3
```
jeschkies: Could we go with two zones instead?

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
         "The number of healthy tasks is {}, but {} was expected".format(app['tasksHealthy'], target_instances_count)
 
     client.remove_app(app['id'], True)
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+def test_faultdomains_default():
```
jeschkies: How about a `skipif("fault_domais < 2")`

tests/system/apps/__init__.py
```diff
@@ -89,3 +89,34 @@ def ipv6_healthcheck():
...
+    :param zone: The zone to start this app in
+    :param instances: The number of instances in this app
+    :param constraints: Other constraints to append
+    :return: Returns the App Definition
+    """
+    app = load_app('faultdomain-base-app')
+    app['instances'] = instances
+
+    # Append region constraint
+    if not region is None:
```
jeschkies: Hm, `if region is not None` reads a little better.

tests/system/common.py
```diff
@@ -780,6 +781,71 @@ def task_by_name(tasks, name):
...
+    if type(zones) is str:
+        zones = [zones]
+
+    # Sanitize agent domains and user input
+
+    slave_domains = get_cluster_slave_domains()
+
+    assert len(slave_domains) > 0, "Did not find any agents in the DC/OS cluster" % (app['id'],)
+
+    if not regions is None:
```
jeschkies: `if region is not None`

tests/system/common.py
```diff
@@ -780,6 +781,71 @@ def task_by_name(tasks, name):
...
+        slave_regions = set(map(lambda x: x.region, slave_domains.values()))
```
jeschkies:
```python
slave_regions = set([x.region for x in slave_domains.values()])
```

`map` is slightly faster but harder to read IMHO.

tests/system/common.py
```diff
@@ -780,6 +781,71 @@ def task_by_name(tasks, name):
...
+        for _region in regions:
+            if not _region in slave_regions:
+                assert False, "Region %s was not found in the cluster (expecting one of %s)" \
+                              % (_region, ', '.join(slave_regions))
```
jeschkies:
```python
unknown_regions = set(regions) - slave_regions
assert len(unknown_regions) == 0, ...
```

tests/system/common.py
```diff
@@ -780,6 +781,71 @@ def task_by_name(tasks, name):
...
+    if not zones is None:
+        slave_zones = set(map(lambda x: x.zone, slave_domains.values()))
+        for _zone in zones:
+            if not _zone in slave_zones:
+                assert False, "Zone %s was not found in the cluster (expecting one of %s)" \
+                              % (_zone, ', '.join(slave_zones))
```
jeschkies: Same as above.

Jenkinsfile.disabled.si
wavesoft: Ah, true 👍

Jenkinsfile.permissive.si
wavesoft: True 👍

ci/launch_cluster.sh
wavesoft: I was so sure I added an exit statement there :D Definitely

ci/launch_cluster.sh
wavesoft: I have nothing against sed, however I generally prefer to avoid invoking an external tool for something bash has a built-in. (again, `bash`, not `zsh` or other shells, but we do have a bash hashbang in the beginning 😉 )

ci/launch_cluster.sh
wavesoft: ah, true, I just copy-pasted my reference config. We can definitely work with `m4.large`

ci/launch_cluster.sh
wavesoft: Yup. Good idea 👍

ci/launch_cluster.sh
wavesoft: We could give it a try. It was just a hunch that 2 might not show some corner-case errors due to fewer choices that could be accidentally picked.

wavesoft:
Ok, I am going to separate the cluster code into a separate PR to make things easier to digest.

tests/system/apps/__init__.py
wavesoft: 👍

tests/system/common.py
wavesoft: :+1:

tests/system/common.py
wavesoft: Sure, I am still more used to the functional approach to pythonic approaches, but that looks more expressive 👍

tests/system/common.py
wavesoft: :+1:

tests/system/common.py
wavesoft: 👍

tests/system/marathon_common_tests.py
wavesoft: Again, I am not sure if we should silently ignore this set of tests, instead of making sure they run in the correct environment..

tests/system/apps/__init__.py
```diff
@@ -89,3 +89,34 @@ def ipv6_healthcheck():
...
+    :return: Returns the App Definition
+    """
+    app = load_app('faultdomain-base-app')
+    app['instances'] = instances
+
+    # Append region constraint
+    if region is not None:
+        app['constraints'] += [
+            ["@region", "IS", str(region)]
+        ]
```
jeschkies: Isn't this just `app['constraints'].append(["@region", "IS", str(region)])`?

tests/system/apps/faultdomain-base-app.json
```diff
@@ -0,0 +1,8 @@
+{
+  "id": "/fault-domain",
```
jeschkies: Hm, this is not much different than other app definitions. Should we just use the sleep app and change the id?

tests/system/common.py
```diff
@@ -16,6 +16,7 @@
 from json.decoder import JSONDecodeError
 from shakedown import marathon
 from urllib.parse import urljoin
+from utils import get_cluster_slave_domains, get_used_regions_and_zones, get_app_domains
```
jeschkies: I'm pretty sure flake8 does not like this 😉.

tests/system/common.py
```diff
@@ -780,6 +781,69 @@ def task_by_name(tasks, name):
 
     assert False, "Did not find task with name %s in this list of tasks: %s" % (name, tasks,)
 
+def assert_app_in_all_domains(app, regions=None, zones=None):
+    """Asserts that all tasks of the given app are within all of the given region and/or zone
```
jeschkies: Can a task be in several regions? Or do you mean the tasks are distributed across the regions?

tests/system/common.py
```diff
@@ -780,6 +781,69 @@ def task_by_name(tasks, name):
...
+
+    slave_domains = get_cluster_slave_domains()
+
+    assert len(slave_domains) > 0, "Did not find any agents in the DC/OS cluster" % (app['id'],)
+
+    if regions is not None:
+        slave_regions = set([x.region for x in slave_domains.values()])
+        unknown_regions = set(regions) - slave_regions
+        assert len(unknown_regions) == 0, "Region(s) %s was not found in the cluster (expecting one of %s)" \
+                              % (', '.join(unknown_regions), ', '.join(slave_regions))
```
jeschkies: Use `"expecting one of {}".format(...)` instead of `%` syntax.

tests/system/common.py
```diff
@@ -780,6 +781,69 @@ def task_by_name(tasks, name):
...
+
+    if regions is not None:
+        assert used_regions == set(regions), \
+            "Application %s is not running on all the exepected regions ({%s} instead of {%s})" % (
+                app['id'], ', '.join(used_regions), ', '.join(regions)
+            )
+
+    if zones is not None:
+        assert used_zones == set(zones), \
+            "Application %s is not running on all the exepected zones ({%s} instead of {%s})" % (
```
jeschkies: See above.

tests/system/common.py
```diff
@@ -780,6 +781,69 @@ def task_by_name(tasks, name):
...
+                app['id'], ', '.join(used_zones), ', '.join(zones)
+            )
+
+
+def assert_app_in_some_domains(app, region=None, zone=None):
+    """Asserts that any of the tasks of the given app are within all of the given region and/or zone
```
jeschkies: Not sure what this means. Can a task be in all regions at the same time?

tests/system/marathon_common_tests.py
jeschkies: We do so for multi cluster tests. Same for EE tests. However, I guess we could split up the tests and just run the once that are supposed to pass.

tests/system/utils.py
```diff
@@ -39,3 +39,114 @@ def get_resource(resource):
...
+        self.region = fault_domain.get('region', {}).get('name', 'default')
+
+
+def get_cluster_local_domain():
+    """Contacts the DC/OS mesos master and returns it's faultDomain configuration (aka "local domain")
+    """
+    master = mesos.get_master()
+    return FaultDomain(master.state().get('domain', {}))
+
+def get_cluster_slave_domains():
```
jeschkies: Nit: `get_cluster_agent_domains`

tests/system/utils.py
```diff
@@ -39,3 +39,114 @@ def get_resource(resource):
...
+    """Returns a dictionary with the slave IDs in the cluster and their corresponding
```
jeschkies: Nit: `agent`

tests/system/utils.py
```diff
@@ -39,3 +39,114 @@ def get_resource(resource):
...
+
+def get_all_cluster_regions():
+    """Returns a dictionary with all the regions and their zones in the cluster
+    """
+    domain_regions = {}
+
+    # Populate all the domain zones and regions found in the cluster
+    for domain in get_cluster_slave_domains().values():
+        if not domain.region in domain_regions:
+            domain_regions[domain.region] = []
```
jeschkies: How about `defaultdict`?

```python
domain_regions = defaultdict(set)
for domain in get_cluster_slave_domains().values():
    domain_regions.append(domain.zone)
```

tests/system/utils.py
```diff
@@ -39,3 +39,114 @@ def get_resource(resource):
...
+def get_app_domains(app):
+    """Returns a list of all te fault domains used by the tasks of the specified app
+    """
+    tasks = app.get('tasks', [])
+    slave_domains = get_cluster_slave_domains()
+
+    assert len(tasks) > 0, "App %s did not launch any tasks on mesos" % (app['id'],)
+
+    # Collect the FaultDomain objects from all the agents where the tasks are running
+    domains = []
```
jeschkies: Should this be a set?

tests/system/utils.py
```diff
@@ -39,3 +39,114 @@ def get_resource(resource):
...
+
+    # Make sure to always operate on iterable
+    if type(regions) is str:
+        regions = [regions]
+    if type(zones) is str:
+        zones = [zones]
+
+    # Increment counter for the agents that pass the checks
+    for domain in get_cluster_slave_domains().values():
+        if not regions is None and not domain.region in regions:
```
jeschkies: Nit: `if regions is not None`

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@pytest.mark.skipif("shakedown.ee_version() is None")
+def test_faultdomains_region_and_zone():
+    """Tests if the applications with a `region` and a `zone` defined are always starting on the
+       same region and zone
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region (%s)" % (
```
jeschkies: See my comment on string interpolation.

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    app_def = apps.faultdomain_app(region=region, zone=single_zone, instances=len(zones) * 2)
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
+
+    # Ensure that all instances are running in the same region, and on *all* the region zones
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(region,), zones=(single_zone,))
```
jeschkies: What does `(region,)` mean?

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    # Check if the application was launched correctly
+    app = client.get_app(app_def['id'])
+    (used_regions, used_zones) = get_used_regions_and_zones(get_app_domains(app))
+
+    assert used_regions == set(regions), "App %s was not launched in all regions: %s instead of %s" % (
+            app['id'], ', '.join(used_regions), ', '.join(regions)
+        )
+
+    # Remove app
+    client.remove_app(app['id'], True)
```
jeschkies: I thinks we have a fixture/cleanup for this.

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    assert len(regions) > 1, "The tests require at least 2 regions in the cluster"
+
+    # Create an app that should be deployed on a unique region
+    # The number of instances should be the number of regions available in the system
+    app_def = apps.faultdomain_app(
+        instances=len(regions),
+        constraints=[["@region", "UNIQUE"]]
+    )
+    client = marathon.create_client()
+    client.add_app(app_def)
```
jeschkies: Please define an app id that is unique, e.g. `app-unique-fault-domain`. This makes debugging easier.

tests/system/marathon_common_tests.py
```diff
@@ -1154,3 +1157,173 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region (%s)" % (
+        region,
+    )
+
+    # Create an app that should be deployed on a unique region
+    # The number of instances should be the number of regions available in the system
+    app_def = apps.faultdomain_app(
```
jeschkies: Please change the app id.

tests/system/apps/__init__.py
wavesoft: Yes, this is a bit more pythonic way of doing this, plus it is visually closer to the the `constraints` field format. I have no problem to switch this to `.append` though.

tests/system/common.py
wavesoft: I see no complain from `flake8` on this 😄

tests/system/apps/faultdomain-base-app.json
wavesoft: Yes, it's very close to the sleep app. I created a new app definition purely for separation of concerns.

tests/system/common.py
wavesoft: "All the tasks of the given app are distributed across the regions" let me rephrase it 😄

tests/system/common.py
wavesoft: 👍 I used `%` for uniformity since we are using it in other places too

tests/system/common.py
wavesoft: 👍

tests/system/marathon_common_tests.py
wavesoft: It's a tuple of one item. Like `[region]`, but lighter.

tests/system/common.py
wavesoft: This function was not used. I removed it.

tests/system/utils.py
wavesoft: Why not 👍

tests/system/marathon_common_tests.py
wavesoft: So, are we deleting all apps at teardown? Or is there something we can annotate the test with? Because I only saw tests deploying apps with `add_app` and nothing cleans them 🤔

tests/system/marathon_common_tests.py
wavesoft: That's a good idea 👍

tests/system/marathon_common_tests.py
wavesoft: Done 👍

wavesoft:
@jeschkies I should have addressed all your comments. I am working on resolving the conflicts now...

wavesoft:
Rebased, let's see if the tests pass now 😄

ci/launch_cluster.sh
```diff
@@ -29,33 +29,57 @@ VARIANT="$2"
...
+  if [ -z "$DCOS_INSTALLER_TEMPLATE" ]; then
+      echo "Expecting DCOS_INSTALLER_TEMPLATE environment variable"
+      exit 1
+  fi
+  if [ -z "$DCOS_LICENSE" ]; then
+      echo "Expecting $DCOS_LICENSE environment variable"
+      exit 1
+  fi
+  # Replace {} in the template with the channel
+  INSTALLER_URL=$(echo ${DCOS_INSTALLER_TEMPLATE} | sed "s|{}|${CHANNEL}|g")
```
jeschkies: Why | and not / in sed?

wavesoft: Because `/` will b present on the expanded channel 😉

wavesoft:
We should merge this only when the `system-integration-tests`/ `marathon-si-dcos-strict` passes

kensipe:
@wavesoft looks like we need a rebase

tests/system/marathon_common_tests.py
```diff
@@ -1237,3 +1240,184 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+
+    # Discover which is the "local" cluster domain
+    local_domain = get_cluster_local_domain()
+
+    # Get how many agents do we have in the local domain (zone and region) and try to start
+    # that many applications. Mesos should round-robbin the instances in the region
+    region_agents = count_agents_in_faultdomains(regions=local_domain.region)
+
+    # Create an app with default configuration
+    app_def = apps.faultdomain_app(instances=region_agents, suffix="-defaults")
```
kensipe: we prefer to not have constants as app ids... we should be consistent.   we should be able to launch multiple fault domain apps and not have collisions.   see other app launches as an example.

tests/system/marathon_common_tests.py
```diff
@@ -1237,3 +1240,184 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
```
kensipe: is there a reason to shorten the default wait time?   if not... lets remove it.

tests/system/marathon_common_tests.py
```diff
@@ -1237,3 +1240,184 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    )
+
+    # Create an app pinned on the given region, with at least double the number of instances than
+    # the zones in the cluster, to ensure that they are distributed among all of them
+    app_def = apps.faultdomain_app(region=region, instances=len(zones) * 2, suffix="-region-only")
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(timeout=timedelta(minutes=1).total_seconds(), app_id=app_def['id'])
```
kensipe: lets use defaults

tests/system/marathon_common_tests.py
kensipe: this should definitely NOT be in the test itself... if it isn't already handled by a test fixture... a test fixture should be created.

tests/system/marathon_common_tests.py
```diff
@@ -1237,3 +1240,184 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+
+    assert used_regions == set([region]), "App %s was not launched in the expected region(s): %s instead of %s" % (
+        app['id'], ', '.join(used_regions), region
+    )
+    assert used_zones == set(zones), "App %s was not launched in all zones: %s instead of %s" % (
+        app['id'], ', '.join(used_zones), ', '.join(zones)
+    )
+
+    # Remove app
+    client.remove_app(app['id'], True)
```
kensipe: test clean should never be in the test

tests/system/marathon_common_tests.py
wavesoft: This is not creating a constant app ID...
1. `apps.faultdomain_app` calls 
2. `apps.load_app` that uses
3. `utils.make_id` to create a unique ID
4. The `suffix` is appended to help debugging

tests/system/marathon_common_tests.py
wavesoft: I think I copy-pasted this from somewhere. Removed the timeout adjustment...

tests/system/marathon_common_tests.py
wavesoft: 👍

tests/system/utils.py
wavesoft: Good point 👍

ci/launch_cluster.sh
```diff
@@ -38,25 +38,34 @@ envsubst <<EOF > "$CONFIG_PATH"
...
 dcos_config:
-    cluster_name: $DEPLOYMENT_NAME
-    resolvers:
-        - 8.8.4.4
-        - 8.8.8.8
-    dns_search: mesos
-    master_discovery: static
-    exhibitor_storage_backend: static
-    rexray_config_preset: aws
+  cluster_name: $DEPLOYMENT_NAME
```
jeschkies: Why the change of spacing? This makes reviewing hard.

ci/launch_cluster.sh
```diff
@@ -38,25 +38,34 @@ envsubst <<EOF > "$CONFIG_PATH"
 ---
 launch_config_version: 1
 deployment_name: $DEPLOYMENT_NAME
-installer_url: $INSTALLER
-provider: onprem
-platform: aws
 aws_region: us-west-2
-os_name: cent-os-7-dcos-prereqs
-key_helper: true
+installer_url: $INSTALLER_URL
```
jeschkies: Please move up to limit the number of line changes.

ci/launch_cluster.sh
```diff
@@ -38,25 +38,34 @@ envsubst <<EOF > "$CONFIG_PATH"
 ---
 launch_config_version: 1
 deployment_name: $DEPLOYMENT_NAME
-installer_url: $INSTALLER
-provider: onprem
-platform: aws
 aws_region: us-west-2
-os_name: cent-os-7-dcos-prereqs
```
jeschkies: Was this removed?

tests/system/apps/__init__.py
jeschkies: I think it should be `append`.

tests/system/apps/__init__.py
```diff
@@ -96,3 +96,39 @@ def ipv6_healthcheck(app_id=None, parent_group="/"):
...
+    :param instances: The number of instances in this app
+    :param constraints: Other constraints to append
+    :return: Returns the App Definition
+    """
+    app = load_app('faultdomain-base-app')
+    app['instances'] = instances
+
+    # Append name suffix if needed
+    if suffix:
+        app['id'] += suffix
```
jeschkies: please see how we set the app ids now.

tests/system/common.py
```diff
@@ -800,7 +801,53 @@ def task_by_name(tasks, name):
...
+    """
+
+    if type(regions) is str:
+        regions = [regions]
+    if type(zones) is str:
+        zones = [zones]
+
+    # Sanitize agent domains and user input
+
+    slave_domains = get_cluster_agent_domains()
```
jeschkies: Replace `slave_domains` with `agent_domains`.

tests/system/common.py
```diff
@@ -800,7 +801,53 @@ def task_by_name(tasks, name):
...
+
+    assert len(slave_domains) > 0, "Did not find any agents in the DC/OS cluster"
```
jeschkies: Please use the matchers from Precisely:

```python
assert_that(agent_domains, has_len(greater_than(0)))
```

This will ease debugging.

tests/system/common.py
```diff
@@ -800,7 +801,53 @@ def task_by_name(tasks, name):
...
+
+    if regions is not None:
+        slave_regions = set([x.region for x in slave_domains.values()])
+        unknown_regions = set(regions) - slave_regions
+        assert len(unknown_regions) == 0, "Region(s) {} was not found in the cluster (expecting one of {})".format(
+            ', '.join(unknown_regions), ', '.join(slave_regions))
```
jeschkies: What do you think about

```python
assert_that(regions, is_(subset_of(agent_regions)))
```

tests/system/common.py
```diff
@@ -800,7 +801,53 @@ def task_by_name(tasks, name):
...
+    (used_regions, used_zones) = get_used_regions_and_zones(get_app_domains(app))
+
+    if regions is not None:
+        assert used_regions == set(regions), \
+            "Application {} is not running on all the exepected regions ([{}] instead of [{}])".format(
+                app['id'], ', '.join(used_regions), ', '.join(regions)
+            )
+
+    if zones is not None:
+        assert used_zones == set(zones), \
```
jeschkies:
```python
assert_that(used_zones, contains_exactly(zones))
```

tests/system/marathon_common_tests.py
jeschkies: Please see @zen-dog's changes on app ids.

tests/system/utils.py
```diff
@@ -39,3 +40,121 @@ def get_resource(resource):
                 raise Exception
         except Exception:
             raise DCOSException("Can't read from resource: {0}. Please check that it exists.".format(resource))
+
+
+class FaultDomain:
+    """
+    High-level representation of a fault domain
```
jeschkies: What does that mean? What do you think about using a `namedtuple` instead?

tests/system/utils.py
```diff
@@ -39,3 +40,121 @@ def get_resource(resource):
...
+    """
+    master = mesos.get_master()
+    return FaultDomain(master.state().get('domain', {}))
+
+
+def get_cluster_agent_domains():
+    """Returns a dictionary with the slave IDs in the cluster and their corresponding
+       fault domain information
+    """
+    slave_domains = {}
```
jeschkies: Please use `agent_domains`.

tests/system/utils.py
```diff
@@ -39,3 +40,121 @@ def get_resource(resource):
...
+            domain_regions[domain.region] = []
+        if domain.zone not in domain_regions[domain.region]:
+            domain_regions[domain.region].append(domain.zone)
+
+    # Return dictionary
+    return domain_regions
+
+
+def get_biggest_cluster_region():
+    """Returns a tuple with the name of the biggest region in the cluster and the zones in it
```
jeschkies: How is bigger defined?

ci/launch_cluster.sh
wavesoft: It's not removed, it's moved down.

I had a weird conflict in this file so I sorted the lines alphabetically to make it easy to compare.

ci/launch_cluster.sh
wavesoft: Done

tests/system/apps/__init__.py
wavesoft: I still see no reasonable explanation, but ok 👍

tests/system/common.py
wavesoft: Done

tests/system/common.py
wavesoft: I like how simpler it makes the expression, but I don't like how less verbose the error message is, so honestly, I am baffled 🤔

tests/system/utils.py
wavesoft: _Not my fault that the API uses `master.slaves` as name 🤣_
Done

tests/system/utils.py
wavesoft: Number of zones. Updated the documentation.

tests/system/common.py
wavesoft: How is `"agent_domains must be grater than 0"` is more verbose than `"Did not find any agents in the DC/OS cluster"` 😄

tests/system/utils.py
wavesoft: I am indifferent on that. However I personally like a fully described type definition more than using generators like `namedtuple` .. but that might be more my C++ background 😄

ci/launch_cluster.sh
```diff
@@ -42,21 +42,30 @@ installer_url: $INSTALLER
 provider: onprem
 platform: aws
 aws_region: us-west-2
-os_name: cent-os-7-dcos-prereqs
 key_helper: true
 instance_type: m4.large
-num_public_agents: 1
-num_private_agents: 3
 num_masters: 3
+os_name: cent-os-7-dcos-prereqs
```
jeschkies: Nit: this should be line 45.

ci/launch_cluster.sh
```diff
@@ -42,21 +42,30 @@ installer_url: $INSTALLER
...
+ssh_user: centos
 dcos_config:
     cluster_name: $DEPLOYMENT_NAME
     resolvers:
-        - 8.8.4.4
         - 8.8.8.8
-    dns_search: mesos
+        - 8.8.4.4
+    dns_search: us-west-2.compute.internal
```
jeschkies: Why is this required?

ci/launch_cluster.sh
```diff
@@ -42,21 +42,30 @@ installer_url: $INSTALLER
...
     master_discovery: static
-    exhibitor_storage_backend: static
+    exhibitor_storage_backend: zookeeper
```
jeschkies: Why is this required?

ci/launch_cluster.sh
```diff
@@ -42,21 +42,30 @@ installer_url: $INSTALLER
...
 num_masters: 3
+os_name: cent-os-7-dcos-prereqs
+ssh_user: centos
 dcos_config:
     cluster_name: $DEPLOYMENT_NAME
     resolvers:
-        - 8.8.4.4
         - 8.8.8.8
-    dns_search: mesos
+        - 8.8.4.4
```
jeschkies: Nit: This should be switched with line 53.

ci/launch_cluster.sh
wavesoft: I don't really remember. I only remember that I had to go through some try and error until I got something working... But let's try what we are doing right now on `master` and if it fails we try my defaults 👍

tests/system/utils.py
wavesoft: Oh, now I remember why. The fault domain is obtained by doing `master.state().get('domain')` on the mesos API. And the response is quite irregular (ex. it can be `None` if there is no fault domain, or it could be an empty object etc.)

That's why I created the `FaultDomain` class, as a wrapper to the response. So, by doing:
```
FaultDomain(master.state().get('domain'))
```
you always get a well-known reference to the fault domain

tests/system/marathon_common_tests.py
wavesoft: I think we have resolved this. I am using `make_app` and passing the `suffix` as part of the app prefix 👍

wavesoft:
All new comments from @jeschkies should be addressed. Please check and ✅

wavesoft:
@zen-dog agreed. Added it on Jenkins.

ci/launch_cluster.sh
```diff
@@ -45,9 +45,17 @@ aws_region: us-west-2
 os_name: cent-os-7-dcos-prereqs
 key_helper: true
 instance_type: m4.large
-num_public_agents: 1
-num_private_agents: 3
 num_masters: 3
+ssh_user: centos
```
jeschkies: Is this really required?

ci/launch_cluster.sh
```diff
@@ -45,9 +45,17 @@ aws_region: us-west-2
...
+fault_domain_helper:
+    LocalRegion:
+      num_zones: 2
+      num_public_agents: 1
+      num_private_agents: 2
+      local: true
+    RemoteRegion:
+      num_zones: 2
+      num_private_agents: 2
```
jeschkies: Thanks, this is so much easier to review.

tests/system/apps/__init__.py
```diff
@@ -8,15 +8,15 @@ def apps_dir():
     return os.path.dirname(os.path.abspath(__file__))
 
 
-def load_app(app_def_file, app_id=None, parent_group="/"):
+def load_app(app_def_file, app_id=None, parent_group="/",  app_suffix=""):
```
jeschkies: Why the suffix? Would it be enough to have it part of `app_id`?

tests/system/apps/__init__.py
```diff
@@ -96,3 +96,36 @@ def ipv6_healthcheck(app_id=None, parent_group="/"):
...
+    the name, zone and region configuration as given
+
+    :param region: The region to start this app in
+    :param zone: The zone to start this app in
+    :param instances: The number of instances in this app
+    :param constraints: Other constraints to append
+    :return: Returns the App Definition
+    """
+    app = load_app('faultdomain-base-app',
+                   app_suffix=("" if suffix is None else "-{}".format(suffix)))
```
jeschkies: I'm not sure we need the suffix.

tests/system/common.py
```diff
@@ -800,7 +801,49 @@ def task_by_name(tasks, name):
...
+    if type(regions) is str:
+        regions = [regions]
+    if type(zones) is str:
+        zones = [zones]
+
+    # Sanitize agent domains and user input
+
+    agent_domains = get_cluster_agent_domains()
+
+    assert len(agent_domains) > 0, "Did not find any agents in the DC/OS cluster"
```
jeschkies: Please use the new matchers. It is not a hard requirement but eases debugging.

tests/system/common.py
```diff
@@ -800,7 +801,49 @@ def task_by_name(tasks, name):
         if task['name'] == name:
             return task
 
-    assert False, "Did not find task with name %s in this list of tasks: %s" % (name, tasks,)
+    assert False, "Did not find task with name {} in this list of tasks: {}".format(name, tasks)
+
+
+def assert_app_in_all_domains(app, regions=None, zones=None):
+    """Asserts that all tasks of the given app are distributed in all of the given region and/or zone
```
jeschkies: Hm, does this mean we assert that all zones and regions are unique of an app deployment are unique?

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    # that many applications. Mesos should round-robbin the instances in the region
+    region_agents = count_agents_in_faultdomains(regions=local_domain.region)
+
+    # Create an app with default configuration
+    app_def = apps.faultdomain_app(instances=region_agents, suffix="-defaults")
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(app_id=app_def['id'])
```
jeschkies: Please use `common.deployment_wait`

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+def test_faultdomains_region_only():
+    """Tests if the applications that only have a `region` specified are launched on all the different
+       zones available in that region on the cluster
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region ({})".format(
```
jeschkies: Use a `pytest.mark.skipif` instead.

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+def test_faultdomains_region_and_zone():
+    """Tests if the applications with a `region` and a `zone` defined are always starting on the
+       same region and zone
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region ({})".format(
```
jeschkies: See above

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+        region,
+    )
+
+    single_zone = zones[0]
+
+    # Create an app pinned on the given region and zone.
+    #
+    # (We are starting more instances than the number of zones in the region in order to increase
+    #  the chance of a faulty instance to appear on a different zone)
```
jeschkies: Super nit. Why the `(` and `)`?

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+def test_faultdomains_region_unique():
+    """Tests if the applications with a ["@region", "UNIQUE"] constraint correctly starts the apps
+       across all regions in the cluster
+    """
+
+    # Find out all the regions in the cluster
+    regions = set(get_all_cluster_regions().keys())
+
+    assert len(regions) > 1, "The tests require at least 2 regions in the cluster"
```
jeschkies: See above

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(app_id=app_def['id'])
+
+    # Check if the application was launched correctly
+    app = client.get_app(app_def['id'])
+    (used_regions, used_zones) = get_used_regions_and_zones(get_app_domains(app))
+
+    assert used_regions == set(regions), "App {} was not launched in all regions: {} instead of {}".format(
+            app['id'], ', '.join(used_regions), ', '.join(regions)
```
jeschkies: Please use `assert_that` and matchers.

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+        region=region,
+        instances=len(zones),
+        constraints=[["@zone", "UNIQUE"]],
+        suffix="-zone-unique"
+    )
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(app_id=app_def['id'])
```
jeschkies: Please use `common.deployment_wait`.

tests/system/matcher/__init__.py
```diff
@@ -25,7 +25,8 @@ def has_len(matcher):
 ]
 
 
-def assert_that(value, matcher):
+def assert_that(value, matcher, customDescription=""):
     result = matcher.match(value)
     if not result.is_match:
-        raise AssertionError("\nExpected: {0}\nbut: {1}".format(matcher.describe(), result.explanation))
+        raise AssertionError("{0}\nExpected: {1}\nbut: {2}".format(
+            customDescription, matcher.describe(), result.explanation))
```
jeschkies: I do like it but please use `snake_case` and omit the newline when the description is empty.

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+
+        # Extract fault domain
+        fault_domain = config.get('fault_domain', {})
+
+        # Extract zone/region
+        self.zone = fault_domain.get('zone', {}).get('name', 'default')
+        self.region = fault_domain.get('region', {}).get('name', 'default')
+
+
+def get_cluster_local_domain():
```
jeschkies: What do you think about passing the cluster state instead?

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+
+    # Populate agent_domains with the ID and the corresponding domain for each slave
+    master = mesos.get_master()
+    for slave in master.slaves():
+        agent_domains[slave['id']] = FaultDomain(slave._short_state.get('domain', None))
+    return agent_domains
+
+
+def get_all_cluster_regions():
+    """Returns a dictionary with all the regions and their zones in the cluster
```
jeschkies: So it's `region -> list[zone]`?

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+    """
+    domain_regions = {}
+
+    # Populate all the domain zones and regions found in the cluster
+    for domain in get_cluster_agent_domains().values():
+        if domain.region not in domain_regions:
+            domain_regions[domain.region] = []
```
jeschkies: How about using the `defaultdict` instead?

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+def test_faultdomains_zone_unique():
+    """Tests if the applications with a region fixed and a ["@zone", "UNIQUE"] constraint correctly starts
+       the apps across all zones in the given region
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
```
jeschkies: Why do we have to find the biggest region?

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+    """Returns a tuple with the name of the region with the most zones in the
+       cluster and a list with the actual zones in it
+    """
+    biggest_region = None
+    biggest_region_zones = []
+
+    for (region, zones) in get_all_cluster_regions().items():
+        if len(zones) > len(biggest_region_zones):
+            biggest_region_zones = zones
+            biggest_region = region
```
jeschkies: I'm not sure if we really need `get_biggest_cluster_region` but what do you think about
```python
def get_biggest_cluster_region():
   return max(get_all_cluster_regions().items(), key=lambda kv: len(kv[1]))
```

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+
+    return (biggest_region, biggest_region_zones)
+
+
+def get_app_domains(app):
+    """Returns a list of all te fault domains used by the tasks of the specified app
```
jeschkies: What do you mean with "te"?

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+    """
+    tasks = app.get('tasks', [])
+    agent_domains = get_cluster_agent_domains()
+
+    assert len(tasks) > 0, "App %s did not launch any tasks on mesos" % (app['id'],)
+
+    # Collect the FaultDomain objects from all the agents where the tasks are running
+    domains = set()
+    for task in tasks:
+        domains.append(agent_domains[task['slaveId']])
```
jeschkies: Do be more pythonic
```python
set(agent_domains[task['slaveId']] for task in tasks)
```

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+    return (biggest_region, biggest_region_zones)
+
+
+def get_app_domains(app):
+    """Returns a list of all te fault domains used by the tasks of the specified app
+    """
+    tasks = app.get('tasks', [])
+    agent_domains = get_cluster_agent_domains()
+
+    assert len(tasks) > 0, "App %s did not launch any tasks on mesos" % (app['id'],)
```
jeschkies: I don't think this assertion is required.

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+    used_zones = set()
+
+    for domain in domains:
+        used_regions.add(domain.region)
+        used_zones.add(domain.zone)
+
+    return (used_regions, used_zones)
+
+
+def count_agents_in_faultdomains(regions=None, zones=None):
```
jeschkies: Are we really checking for the fault domain as one can pass regions and zones.

tests/system/utils.py
```diff
@@ -39,3 +40,122 @@ def get_resource(resource):
...
+
+    # Increment counter for the agents that pass the checks
+    for domain in get_cluster_agent_domains().values():
+        if regions is not None and domain.region not in regions:
+            continue
+        if zones is not None and domain.zone not in zones:
+            continue
+        counter += 1
+
+    return counter
```
jeschkies: If I understand this correctly you want something like

```python
len([d for d in get_cluster_agent_domains().values()
     if d.region in regions and d.zone in zones])
```

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    )
+
+    # Create an app pinned on the given region, with at least double the number of instances than
+    # the zones in the cluster, to ensure that they are distributed among all of them
+    app_def = apps.faultdomain_app(region=region, instances=len(zones) * 2, suffix="-region-only")
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(app_id=app_def['id'])
```
zen-dog: `common.deployment_wait`

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    app_def = apps.faultdomain_app(
+        region=region,
+        zone=single_zone,
+        instances=len(zones) * 2,
+        suffix="-region-and-zone")
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    shakedown.deployment_wait(app_id=app_def['id'])
```
zen-dog: ^.^

tests/system/marathon_common_tests.py
```diff
@@ -1241,3 +1243,174 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+def test_faultdomains_zone_unique():
+    """Tests if the applications with a region fixed and a ["@zone", "UNIQUE"] constraint correctly starts
+       the apps across all zones in the given region
+    """
+
+    # Find out the biggest region in the cluster
+    (region, zones) = get_biggest_cluster_region()
+
+    assert len(zones) > 1, "The tests require at least 2 zones in the biggest region ({})".format(
```
zen-dog: `skipif`

tests/system/apps/__init__.py
wavesoft: @jeschkies nope, because if you define an `app_id`, no additional random part will be generated. Hence the `suffix`.

tests/system/apps/__init__.py
wavesoft: Because depending on the `region`, `zone` and `constraints` you can create different variations of the app definition. The `suffix` is a symbolic name for the region/zone/constraints combination so it's easier debuggable.

tests/system/common.py
wavesoft: Nope, more than one task can be started in the same region/zone. We just need to make sure that no tasks are spilled in regions/zones outside of the ones configured...

ci/launch_cluster.sh
wavesoft: I think I added it for ensuring that we have a consistent SSH user, regardless of the installation method or target OS. But I can remove it.

tests/system/marathon_common_tests.py
wavesoft: I thought of it, but I have a feeling we are hiding something if I just silently skip this test.
That's because we are testing a crucial feature, and if the cluster is not able to run this kind of test then we should *not* ignore it, but fail proud and loud.

Hence I don't think that this is a classic `skipif` case. For instance, consider the other cases:
* `skipif("shakedown.ee_version() is None")` : We run the tests in a matrix of cluster configurations, and this particular test is testing only the EE feature. Totally legit use.
* `skipif("shakedown.ee_version() == 'strict'")` : Like before, we are running the test in the very specific case of "strict" clusters. Again, totally legit use.

But consider this:
* `skipif("len(get_biggest_cluster_region()[1]) < 2")` : This means that this test cannot run if we hare less than 2 regions in the cluster. But do we have such cases? Because after we merge this PR, *all* of our test matrix is going to run on a 2-region cluster. 

And consider the case we _do_ use `skipif`, but in the future we decide to reduce the cluster size due to some other reason. You are never going to figure out that these tests never run...

I don't know. Both cases have it's positive and negative aspects. I am open on both solutions, I just need some justification, to ease my mind 😄

tests/system/marathon_common_tests.py
wavesoft: It's a guilty comment. `( )` denotes something we _shouldn't_ be doing, but we are doing nonetheless :trollface:

tests/system/utils.py
wavesoft: All the other `*_cluste_*` functions in the `utils.py` are not using this approach. What's your reasoning for breaking the interface with this one?

tests/system/marathon_common_tests.py
wavesoft: Because some tests require at least 2 zones in a region. For these tests, this function makes sure we pick the correct one.

tests/system/utils.py
wavesoft: Yes, but in some cases you also want the name of that zone, hence the tuple approach.
Honestly, if you want to make things more verbose you could do:
```
region = get_biggest_region_name()
zones = get_region_zones(region)
```

This is just compacting the two methods in one:
```
(region, zones) = get_biggest_region()
```

tests/system/utils.py
wavesoft: Almost.. this function has multiple signatures and this one-liner is not handling it:
* `count_agents_in_faultdomains(regions=[...])` count agents in these region(s)
* `count_agents_in_faultdomains(zones=[...])` count agents in these zone(s)
* `count_agents_in_faultdomains(regions=[], zones=[])` count agents in these regions AND zones.

tests/system/marathon_common_tests.py
jeschkies: That's actually a good point. The test should fail if we accidentally change the environment. I'd argue it it a prerequisite, though. What do you think about running a check before as we do for Marathon being reachable?

We could have something like
```
def multiple_zones(test_func):
    def wrapped():
         assert len(zones) > 1, "The tests ..."
         return test_func()

    return wrapped

@require(multiple_zones())
def test_faultdomains_default():
    ....
```

wavesoft: That's actually a nice approach to it. Let me give it a try.

tests/system/apps/__init__.py
jeschkies: You could still use `make_app_id`. The suffix is just odd to me. It basically states includes this in my random id.

tests/system/common.py
```diff
@@ -39,6 +40,19 @@
 marathon_1_7 = pytest.mark.skipif('marathon_version_less_than("1.7")')
 
 
+def require(requirement, message=None):
```
jeschkies: Nice! I think we want to include this in the shakedown library. Unfortunately, it is not structured yet.

tests/system/marathon_common_tests.py
jeschkies: Who know that?

tests/system/marathon_common_tests.py
```diff
@@ -1268,3 +1271,169 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+
+    # Ensure that all instances are running in the same region, and on *all* the region zones
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(region,), zones=zones)
+
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+@common.require(lambda: len(get_biggest_cluster_region()[1]) > 1,
```
jeschkies: Why the lambda?

tests/system/utils.py
jeschkies: Hm, it feels this is actually a filter. The count should be done afterwards.

tests/system/common.py
jeschkies: Ok, I still don't quite get this assertion. Maybe we want to split them up. I see four assertions below. I'm sure these could be their own methods.

tests/system/marathon_common_tests.py
```diff
@@ -1268,3 +1271,169 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    app_def = apps.faultdomain_app(instances=region_agents, suffix="-defaults")
+    client = marathon.create_client()
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    common.deployment_wait(app_id=app_def['id'])
+
+    # Ensure that all instances are running in the same region (it's ok if they belong to different zones)
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(local_domain.region,))
```
jeschkies: As mentioned above. What are we asserting here? If I understand correctly you want the app only running in the defined region and all zones? What's the zone/region hierarchy again?

tests/system/utils.py
```diff
@@ -38,3 +40,108 @@ def get_resource(resource):
...
+        # Make sure config is a dict
+        if not type(config) is dict:
+            config = {}
+
+        # Extract fault domain
+        fault_domain = config.get('fault_domain', {})
+
+        # Extract zone/region
+        self.zone = fault_domain.get('zone', {}).get('name', 'default')
+        self.region = fault_domain.get('region', {}).get('name', 'default')
```
jeschkies: Ok, so the domain could have a filed `zone` that itself is an object with a property `name`. And we default to `default`.

tests/system/utils.py
```diff
@@ -38,3 +40,108 @@ def get_resource(resource):
...
             raise DCOSException("Can't read from resource: {0}. Please check that it exists.".format(resource))
+
+
+class FaultDomain:
+    """
+    High-level representation of a fault domain
+    """
+    def __init__(self, config):
+        # Make sure config is a dict
+        if not type(config) is dict:
```
jeschkies: This should go. Python uses duck typing. As long as `config` offers a proper `get` method it's fine. If not the caller will see the proper exception.

tests/system/utils.py
```diff
@@ -38,3 +40,108 @@ def get_resource(resource):
...
+        # Extract zone/region
+        self.zone = fault_domain.get('zone', {}).get('name', 'default')
+        self.region = fault_domain.get('region', {}).get('name', 'default')
+
+
+def get_cluster_local_domain():
+    """Contacts the DC/OS mesos master and returns it's faultDomain configuration (aka "local domain")
+    """
+    master = mesos.get_master()
+    return FaultDomain(master.state().get('domain', {}))
```
jeschkies: The `None` case should be handled in the `FaultDomain` constructor.

tests/system/utils.py
```diff
@@ -38,3 +40,108 @@ def get_resource(resource):
...
+def get_cluster_agent_domains():
+    """Returns a dictionary with the slave IDs in the cluster and their corresponding
+       fault domain information
+    """
+    agent_domains = {}
+
+    # Populate agent_domains with the ID and the corresponding domain for each slave
+    master = mesos.get_master()
+    for slave in master.slaves():
+        agent_domains[slave['id']] = FaultDomain(slave._short_state.get('domain', None))
```
jeschkies: No need to default to `None`.

tests/system/marathon_common_tests.py
```diff
@@ -1268,3 +1271,169 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+@common.require(lambda: len(get_all_cluster_regions()) > 1,
+                "Cluster must have at least 2 regions")
+def test_faultdomains_region_unique():
+    """Tests if the applications with a ["@region", "UNIQUE"] constraint correctly starts the apps
+       across all regions in the cluster
+    """
+
+    # Find out all the regions in the cluster
+    regions = set(get_all_cluster_regions().keys())
```
jeschkies: `keys` should be a set.

tests/system/utils.py
```diff
@@ -38,3 +40,108 @@ def get_resource(resource):
...
+    # Populate all the domain zones and regions found in the cluster
+    for domain in get_cluster_agent_domains().values():
+        if domain.zone not in domain_regions[domain.region]:
+            domain_regions[domain.region].append(domain.zone)
+
+    # Return dictionary
+    return domain_regions
+
+
+def get_biggest_cluster_region():
```
jeschkies: Come to think of it I would pass the regions as a parameter.

tests/system/utils.py
```diff
@@ -38,3 +40,108 @@ def get_resource(resource):
...
+
+
+def get_all_cluster_regions():
+    """Returns a dictionary with all the regions and their zones in the cluster
+    """
+    domain_regions = defaultdict(list)
+
+    # Populate all the domain zones and regions found in the cluster
+    for domain in get_cluster_agent_domains().values():
+        if domain.zone not in domain_regions[domain.region]:
```
jeschkies: If you use a set we could avoid this check.

tests/system/marathon_common_tests.py
```diff
@@ -1268,3 +1271,169 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    # Ensure that all instances are running in the same region (it's ok if they belong to different zones)
+    app = client.get_app(app_def['id'])
+    common.assert_app_in_all_domains(app, regions=(local_domain.region,))
+
+
+@shakedown.dcos_1_11
+@pytest.mark.skipif("shakedown.ee_version() is None")
+@pytest.mark.usefixtures("wait_for_marathon_and_cleanup")
+@common.require(lambda: len(get_biggest_cluster_region()[1]) > 1,
+                "Cluster must have at least 2 zones in the biggest region")
```
jeschkies: Now that I see the use case, I feel we have to iterate of `FaultDomain`. What I would expect here is

```python
len(get_biggest_cluster_region().zones) > 1
```

tests/system/utils.py
jeschkies: Reading the code a little more I see how we can improve this. The `FaultDomain` has a region and zones. That should ease your use cases a lot.

tests/system/marathon_common_tests.py
wavesoft: Sorry, leftover :angel:

tests/system/apps/__init__.py
wavesoft: That sounds reasonable :+1:

tests/system/marathon_common_tests.py
wavesoft: True. Since we are not using `python2.x` that's redundant...

tests/system/utils.py
wavesoft: I think it's purpose was not very clear. I updated the comment. 
Please let me know if you still have questions on this.

wavesoft: Also, do you know any concrete example were we are using something that is not a `dict` but uses the `.get` method in our codebase?

tests/system/utils.py
wavesoft: Yup, updated comment block.

tests/system/marathon_common_tests.py
wavesoft: So, we have:
* `Regions` that contain ...
* `Zones` and ...
* The tuple (region, zone) is called  `Fault Domain`

I separated the checks in region / zone. This might make it a bit more clear.

tests/system/marathon_common_tests.py
wavesoft: I don't really get what you are saying here :smile: 
Can you elaborate a bit more?

tests/system/marathon_common_tests.py
wavesoft: Leftovers...

tests/system/utils.py
wavesoft: I didn't get the last part. Can you elaborate on _how_ it can ease the use cases?

tests/system/utils.py
wavesoft: If I use a set, I cannot create the nested `region`->`domains` dict structure.
What do you mean?

tests/system/utils.py
wavesoft: That sounds like premature optimisation. Where are you planning on re-using this function?

tests/system/utils.py
wavesoft: Whatever you feel more comfortable with. :man_shrugging: 
I personally don't like creating an entire array just to count it's items, but if you think it makes the code cleaner, I could have a look.

tests/system/utils.py
jeschkies: That's not a premature optimization simply best practice.

tests/system/marathon_common_tests.py
```diff
@@ -1268,3 +1271,167 @@ def test_ipv6_healthcheck(docker_ipv6_network_fixture):
...
+    client.add_app(app_def)
+
+    # Wait for the app to be deployed
+    common.deployment_wait(app_id=app_def['id'])
+
+    # Check if the application was launched correctly
+    app = client.get_app(app_def['id'])
+    (used_regions, used_zones) = get_used_regions_and_zones(get_app_domains(app))
+
+    assert_that(used_regions, equal_to(regions), "App {} was not launched in all regions: {} instead of {}".format(
```
jeschkies: Isn't this the same as `common.assert_app_in_regions`?

tests/system/utils.py
jeschkies:
```python
class FaultDomain:

    def __init__(self, ...):
        self.region = '...'
        self.zones = list(...)

def get_biggest_cluster_region():
    max(get_all_cluster_domains(), key=lambda d: len(d.zones))

@common.require(len(get_biggest_cluster_region().zones) > 1, "Cluster must have at least 2 zones in the biggest region")
def test_faultdomains_zone_unique():
    """Tests if the applications with a region fixed and a ["@zone", "UNIQUE"] constraint correctly starts
       the apps across all zones in the given region
    """
    # Find out the biggest region in the cluster
    domain = get_biggest_cluster_region()
    zones = domain.zones
    ...
```

Maybe I'm a little confused about the relation between domain, region and zone. When you say `get_biggest_cluster_region()` don't you mean a domain which has the most zones?

tests/system/utils.py
timcharper: This is Python 3.5; why not use optional typing?

Generally I prefer methods that are strict in what they accept. Can we push the concern to make sure they're always iterable upstream? IE passing a string to this method should be an error, not something we just fix and sweep under the rug.

Also, does it make sense to have it accept a python set, and default regions / zones default to an empty set, rather than `None`?

tests/system/utils.py
timcharper: Yeah, seems like we should have a generic count function which receives a predicate, right?

```
def count(arr, predicate):
    return sum(1 for x in arr if predicate(x))
```

tests/system/utils.py
timcharper: Inclined to believe that this is to find the remote zone for testing.

Couldn't we just find and use any region that:

- has the requisite number of zones
- isn't the default region

?

tests/system/utils.py
wavesoft: That sounds like a good idea. Let me try it.

tests/system/marathon_common_tests.py
wavesoft: Actually I can't find any document that explicitly states that this is an EE feature. 
So I can remove this condition.

wavesoft: I think I had in mind the related _Cluster Linking_ that is an EE feature.

tests/system/marathon_common_tests.py
wavesoft: Ps. That particular test `test_faultdomains_default` passes even without any regions defined. So there is no need for any particular pre-conditions.

tests/system/marathon_common_tests.py
wavesoft: Indeed :)

tests/system/utils.py
wavesoft: So, I still don't get the full idea, but let me elaborate:
1. The cluster is separated in `regions` 
2. Each region is further separated in `zones`
3. Each agent is launched in a `zone`
4. The "domain" of an agent is the tuple `(region, domain)` where it's launched.
5. This domain "tuple" is represented with the `FaultDomain` class. (that's why I don't get your idea of using `zones = list(...)`)

regions-zones

tests/system/utils.py
wavesoft: You mean the cluster leading master's state? Or is there a "cluster state" somewhere that I am not aware of? :thinking: 

This function is too small, and the 50% of it is just to fetch the master state. Therefore I don't think it makes sense to fragment it even further. But that's just my personal opinion.

tests/system/utils.py
wavesoft: I am assuming there are no other comments here?

tests/system/utils.py
wavesoft: I am indifferent on this, so why not :+1:  (even if it makes the `require` condition a teensy bit uglier)

zen-dog:
I'll approve immediately after I see 3 green SI runs (_strict, permissive, open_)

zen-dog:
In _Open_ and _Permissive_ `test_install_universe_package[cassandra]` failed in the last run. Seems strange since this test doesn't fail on current _master_.

meichstedt:
I don't think this one's gonna make it.

zen-dog:
> I don't think this one's gonna make it.

Oh, no you gave up on this one! 😂

PR #5952, opened 2018-01-29
Skip serializing docker default network.

gupta-ak:
Summary:
The Mesos agent determines the default docker network mode by checking
if the `ContainerInfo::DockerInfo.network()` field is set and setting
the OS specific default. So, we don't serialize the network if it's
the default (`HOST` in this case).

JIRA issues:
[MARATHON-8056](https://jira.mesosphere.com/browse/MARATHON-8056)

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
+++ b/src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
@@ -238,6 +238,14 @@ object DockerSerializer {
       case HostNetwork => DockerInfo.Network.HOST // it's the default, but we include here for posterity
       case unsupported => throw SerializationFailedException(s"unsupported docker network type $unsupported")
     }.foreach(builder.setNetwork)
+
+    // The default network mode is OS specific for the mesos agent (`HOST` on Linux and `BRIDGE` on Windows), so
+    // we need to clear the network field and let the agent decide what mode to use. Since the default is `HOST` in
+    // in marathon, if we didn't clear this field, we would by default send `HOST`, which is unsupported on Windows.
+    if (DockerInfo.getDefaultInstance.getNetwork == builder.getNetwork) {
+      builder.clearNetwork
+    }
+
     builder.build
   }
 }
--- a/src/test/scala/mesosphere/marathon/api/serialization/ContainerSerializerTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/serialization/ContainerSerializerTest.scala
@@ -6,7 +6,7 @@ import mesosphere.marathon.state.Container.PortMapping
 import mesosphere.marathon.state.Container
 
 import scala.collection.JavaConverters._
-import mesosphere.marathon.core.pod.{ BridgeNetwork, ContainerNetwork }
+import mesosphere.marathon.core.pod.{ HostNetwork, BridgeNetwork, ContainerNetwork }
 import org.scalatest.Inside
 import org.apache.mesos.{ Protos => Mesos }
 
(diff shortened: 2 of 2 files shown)


ichernetsky:
@gupta-ak, thanks a lot for the patch. Could you please add a unit-test?

gupta-ak:
@ichernetsky I added a unit test to `mesosphere.marathon.api.serialization.ContainerSerializerTest`.

src/main/scala/mesosphere/marathon/api/serialization/ContainerSerializer.scala
```diff
@@ -238,6 +238,14 @@ object DockerSerializer {
       case HostNetwork => DockerInfo.Network.HOST // it's the default, but we include here for posterity
       case unsupported => throw SerializationFailedException(s"unsupported docker network type $unsupported")
     }.foreach(builder.setNetwork)
+
+    // The default network mode is OS specific for the mesos agent (`HOST` on Linux and `BRIDGE` on Windows), so
+    // we need to clear the network field and let the agent decide what mode to use. Since the default is `HOST` in
+    // in marathon, if we didn't clear this field, we would by default send `HOST`, which is unsupported on Windows.
+    if (DockerInfo.getDefaultInstance.getNetwork == builder.getNetwork) {
```
alenkacz: can we really do this? What if I deliberately set the host network in the app definition and we here clear that before sending it to mesos. Even though my choice might have been invalid...

WDYT? @zen-dog @ichernetsky

alenkacz:
@gupta-ak thanks for the PR!! Let me just sort out the question I had and then we can merge this! 🎉

jdef:
@alenkacz @zen-dog Strongly disagree with this patch. Trying to second guess the user's intent by translating HOST networking to "use the agent default network" is completely inappropriate. While I can empathize with the operator in the scenario illustrated by the OP, I'm very disinclined to approve this commit.

Instead, let's consider an alternative approach: maybe Marathon could support a `--default-network-mode` flag that allowed a user to override the currently hard-coded default of HOST networking. This flag would, of course, apply for both UCR and Docker containers. However, this approach could prove tricky to get right because of assumptions scattered throughout the code base. And it could be argued that this also *breaks* API because the API currently states that the default network mode is HOST -- and many app and pod definitions have been written to assume that.

I don't have a good solution for the problem described by the OP. This PR isn't a good solution, sorry.

zen-dog:
Like jdef said, translating HOST networking to agent default network automatically is an extreme assumption which would break a lot of existing customer services. Luckily marathon can be easily extended via plugins to achieve what you want. Take a look for example at this [plugin](https://github.com/mesosphere/marathon-example-plugins/blob/master/env/src/main/scala/mesosphere/marathon/example/plugin/env/EnvVarExtenderPlugin.scala) - you an update the resulting proto setting the networking to what you need it to be.

gupta-ak:
Thanks for the comments. I do think something needs to be done even if this PR isn't the correct solution. At the very least, it needs to be documented that the default marathon settings do not work on Windows, and users need to explicitly set the docker networking mode to `container` or `container/bridge` for Windows workloads. 

@zen-dog Could you give a concrete example on how it would actually break any existing services? The code doesn't check for `HOST` mode at all; it simply doesn't distinguish "default" in the protobuf from "unserialized" (similar to how proto3 works). The only case that it would ever break if `mesos-agent` does not honor the default specified in `mesos.proto`, which never happens on Linux. The only weird thing is that if a user deliberately sends `HOST` to a Windows container (an invalid configuration), it would end up working due to the `HOST` -> deserialized behavior.

jdef:
@gupta-ak I think the right things to do here are:
(a) close out this PR; asking Marathon internals to subvert the intent of the API is not the right approach.
(b) file a JIRA (so Mesosphere and community can track) or else a PR (ditto) for a docs update that calls out potential problems for Windows users and redirects them to the Windows documentation for Mesos (http://mesos.apache.org/documentation/latest/windows/). I suspect that this isn't the only "wall" that Windows users will hit.
(c) file a Mesos JIRA to update the known issues documented by Mesos (http://mesos.apache.org/documentation/latest/windows/). I don't see the HOST networking problem mentioned on this page, and it probably should be.

gupta-ak:
Sounds good.

PR #5953, opened 2018-01-30
Fix package uploading and improvements to release pipeline

timcharper:
* Augment ci/pipeline script to have parameterized release targets
* Upgrade-in-place to Ammonite 1.0.3 if not already done

JIRA Issues: MARATHON-7983

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -42,6 +42,7 @@ ansiColor('xterm') {
             sh """echo "${PKG_SSH_HOST} ${PKG_SSH_HOSTKEY_TYPE} ${PKG_SSH_HOSTKEY}" | sudo -E tee -a /etc/ssh/ssh_known_hosts"""
 
             withDockerRegistry([credentialsId: 'docker-hub-credentials']) {
+              # Once MARATHON-7986 is done and we update the Jenkins node to a version that has Ammonite 1.x.x or later, we can get rid of this line
               sh """if [ \$(md5sum /usr/local/bin/amm | awk '{print \$1}') = '07143cdf89e94b0a957f5656193d88d7' ]; then sudo curl -L -o /usr/local/bin/amm https://github.com/lihaoyi/Ammonite/releases/download/1.0.3/2.12-1.0.3 && sudo chmod +x /usr/local/bin/amm; fi"""
 
               if (params.dryRun == false)
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -1,7 +1,7 @@
 #!/usr/bin/env groovy
 
 ansiColor('xterm') {
-  node('JenkinsMarathonCI-Debian8-2017-10-23') {
+  node('JenkinsMarathonCI-Debian8-2017-10-06') {
 
     properties([
       parameters([
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -11,7 +11,7 @@ ansiColor('xterm') {
         ),
         string(name: 'version',
           defaultValue: '',
-          description: 'Version that should be released. You will need to tag and push a tag after this job succeeds. MUST start with a number. If unspecified, version is inferred from ref.'
+          description: 'Community Release Version (ex. 1.5.2). Version that should be released. You will need to tag and push a tag after this job succeeds. Must start with a number.'
         ),
         booleanParam(name: 'runTests',
           description: 'Whether unit/integration and package tests should be run',
@@ -45,8 +45,14 @@ ansiColor('xterm') {
               # Once MARATHON-7986 is done and we update the Jenkins node to a version that has Ammonite 1.x.x or later, we can get rid of this line
               sh """if [ \$(md5sum /usr/local/bin/amm | awk '{print \$1}') = '07143cdf89e94b0a957f5656193d88d7' ]; then sudo curl -L -o /usr/local/bin/amm https://github.com/lihaoyi/Ammonite/releases/download/1.0.3/2.12-1.0.3 && sudo chmod +x /usr/local/bin/amm; fi"""
 
+              targets = if (params.dryRun == false) {
+                "docker-tag linux-packages jar-artifact s3-package """ + (params.latest ? "docker-latest" : "")
+              } else {
+                ""
+              }
+
               if (params.dryRun == false)
-                sh """sudo -E env MARATHON_VERSION=$params.version ci/pipeline release --checkout $params.ref --run-tests=$params.runTests docker-tag linux-packages jar-artifact s3-package """ + (params.latest ? "docker-latest" : "")
+                sh """sudo -E env MARATHON_VERSION=$params.version ci/pipeline release --checkout $params.ref --run-tests=$params.runTests $targets"""
             }
           }
         }
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -5,19 +5,23 @@ ansiColor('xterm') {
 
     properties([
       parameters([
-        string(name: 'gitsha',
-          defaultValue: '',
-          description: 'Git Commit SHA to Build and Release (ex. 167d7fb86 or 167)'
+        string(name: 'ref',
+          defaultValue: 'origin/releases/1.5',
+          description: 'Git Commit SHA or Branch to Build and Release (ex. 167d7fb86 or 167)'
         ),
         string(name: 'version',
           defaultValue: '',
-          description: 'Community Release Version (ex. v1.5.2)'
+          description: 'Version that should be released. You will need to tag and push a tag after this job succeeds. MUST start with a number. If unspecified, version is inferred from ref.'
+        ),
+        booleanParam(name: 'runTests',
+          description: 'Whether unit/integration and package tests should be run',
+          defaultValue: true
         ),
         booleanParam(name: 'latest',
           defaultValue: false,
           description: 'True if docker mesosphere/marathon:latest should be pushed'
-        )]
-      )
+        )
+      ])
     ])
 
     stage("Run Pipeline") {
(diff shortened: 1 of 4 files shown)


ci/pipeline
```diff
@@ -246,36 +279,78 @@ def jenkins(): Unit = {
...
+    sys.exit(1)
+    ???
+  }
+
+  config.checkout.foreach { gitSha =>
+    %('git, "checkout", gitSha)
+  }
 
-  (tagVersion, requestVersion)
+  // if (pushTag) {
```
jeschkies: Why is this commented out?

Jenkinsfile.release
```diff
@@ -1,23 +1,31 @@
...
-          defaultValue: '',
-          description: 'Git Commit SHA to Build and Release (ex. 167d7fb86 or 167)'
+        string(name: 'ref',
+          defaultValue: 'origin/releases/1.5',
+          description: 'Git Commit SHA or Branch to Build and Release (ex. 167d7fb86 or 167)'
         ),
         string(name: 'version',
           defaultValue: '',
-          description: 'Community Release Version (ex. v1.5.2)'
+          description: 'Version that should be released. You will need to tag and push a tag after this job succeeds. MUST start with a number. If unspecified, version is inferred from ref.'
```
jeschkies: Do we want to be extra clever and infer the version?

Jenkinsfile.release
```diff
@@ -1,23 +1,31 @@
...
         ),
         booleanParam(name: 'latest',
           defaultValue: false,
           description: 'True if docker mesosphere/marathon:latest should be pushed'
-        )]
-      )
+        ),
+        booleanParam(name: 'dryRun',
+          defaultValue: false,
+          description: 'Set to true to only run the compile phase.'
```
jeschkies: Nice!

Jenkinsfile.release
```diff
@@ -31,10 +39,13 @@ ansiColor('xterm') {
           sshagent (credentials: ['0f7ec9c9-99b2-4797-9ed5-625572d5931d']) {
             // add the ssh host key for the package uploader step
             // These environment variables are defined in the "Properties Content" section for the job.
```
jeschkies: What is the reason for not using `withCredentials` for these?

Jenkinsfile.release
```diff
@@ -31,10 +39,13 @@ ansiColor('xterm') {
...
-            sh """echo "${PKG_SSH_HOST} ${PKG_SSH_HOSTKEY_TYPE} ${PKG_SSH_HOSTKEY}" | sudo -E tee -a sudo -E tee -a /etc/ssh/ssh_known_hosts"""
+            sh """echo "${PKG_SSH_HOST} ${PKG_SSH_HOSTKEY_TYPE} ${PKG_SSH_HOSTKEY}" | sudo -E tee -a /etc/ssh/ssh_known_hosts"""
 
             withDockerRegistry([credentialsId: 'docker-hub-credentials']) {
-              sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
+              sh """if [ \$(md5sum /usr/local/bin/amm | awk '{print \$1}') = '07143cdf89e94b0a957f5656193d88d7' ]; then sudo curl -L -o /usr/local/bin/amm https://github.com/lihaoyi/Ammonite/releases/download/1.0.3/2.12-1.0.3 && sudo chmod +x /usr/local/bin/amm; fi"""
```
jeschkies: What is the reason for provisioning Ammonite here?

Jenkinsfile.release
```diff
@@ -31,10 +39,13 @@ ansiColor('xterm') {
...
+
+              if (params.dryRun == false)
+                sh """sudo -E env MARATHON_VERSION=$params.version ci/pipeline release --checkout $params.ref --run-tests=$params.runTests docker-tag linux-packages jar-artifact s3-package """ + (params.latest ? "docker-latest" : "")
```
jeschkies: Why is the version passed as an environment parameter? Also, what is the reason for not triggering `linux-packages`, `jar-artifact` and `s3-package` with `release`?

build.sbt
```diff
@@ -292,6 +292,14 @@ lazy val marathon = (project in file("."))
   .settings(formatSettings: _*)
   .settings(packagingSettings: _*)
   .settings(
+    version := {
+      val v = sys.env.get("MARATHON_VERSION").filter(_ != "git").getOrElse(version.value)
```
jeschkies: Ah, got it. I should have read on. Do we want to allow this? For 1.6 onward we do not require this, do we?

ci/pipeline
```diff
@@ -246,36 +279,78 @@ def jenkins(): Unit = {
  * @return Version and artifact description of Marathon build.
  */
 @main
-def release(requestVersion: String, gitSha: String, latest: Boolean = false): Unit = {
-  val releaseVersion = withoutVersionPrefix(requestVersion)
-  val tagVersion = s"v$releaseVersion"
+def release(args: String*): Unit = {
+  case class Config(
```
jeschkies: I like the introduction of scopt. I thought Ammonite could already parse arguments, though.

Jenkinsfile.release
timcharper: No :)

timcharper: fail loud and proud.

ci/pipeline
timcharper: Did not change this; Jenkins does not have access to push the tag. Some day we would like to have jenkins tag and push on success.

Jenkinsfile.release
timcharper: Oh, I see what you mean now, you're asking if we want to allow the version to be inferred.

Yeah, we should probably make it required. I don't see of a valid reason to allow it to be inferred.

Jenkinsfile.release
timcharper: Splitting up the various published phases into various targets allows one to more easily rerun a part of the job without needing to be worried about affecting other already published artifacts. It enables more rapid iteration to fix something if it is broken.

We hardcode, in the Jenkins file, that all targets should be run (minus, of course, docker-latest).

build.sbt
timcharper: I think it is nice, for 1.5. I have a preference to not create a tag until we are sure that the packages built past tests, the integration tests pass, etc. It avoids the chicken and egg scenario.

ci/pipeline
timcharper: You're right, it can already parse arguments. Support is pretty lacking.

Take the following example:

```

13:05 $ cat args.sc
@main def main(run_tests: Boolean, rest: String*): Unit = {
  println(run_tests)
  println(rest.mkString(","))
}
 ~
13:06 $ amm args.sc --flag true lol 1 2 3
The following argument failed to parse:

--run_tests: Boolean = "lol" failed to parse with java.lang.IllegalArgumentException: 'lol' is not a boolean.

expected signature:

main
  --run_tests  Boolean
  --rest       String*
```

In this case, since I provided the wrong argument name, it assumed the next argument, "LOL", was for the perimeter `–-run_tests`.

This can be worse if you accidentally mistype a parameter:

```
13:09 $ amm args.sc true --run_tests false 1 2 3 --lol false
true
--run_tests,false,1,2,3,--lol,false
 ~
```

Ammonite just completely ignores it.

So, while it would be possible to use Ammonite's argument parsing, there are more pitfalls and the potential for surprise is much greater.

Other limitations of Ammonite's argument pursing abilities is that you cannot specify default values for arguments.

Jenkinsfile.release
timcharper: Because the version in the AMI image is insufficient and I cannot update the AMI image right now. Blocked on credentials.

timcharper: https://jira.mesosphere.com/browse/MARATHON-8057

Jenkinsfile.release
```diff
@@ -31,10 +39,13 @@ ansiColor('xterm') {
...
             // add the ssh host key for the package uploader step
             // These environment variables are defined in the "Properties Content" section for the job.
-            sh """echo "${PKG_SSH_HOST} ${PKG_SSH_HOSTKEY_TYPE} ${PKG_SSH_HOSTKEY}" | sudo -E tee -a sudo -E tee -a /etc/ssh/ssh_known_hosts"""
+            sh """echo "${PKG_SSH_HOST} ${PKG_SSH_HOSTKEY_TYPE} ${PKG_SSH_HOSTKEY}" | sudo -E tee -a /etc/ssh/ssh_known_hosts"""
 
             withDockerRegistry([credentialsId: 'docker-hub-credentials']) {
-              sh """sudo -E ci/pipeline release $params.version $params.gitsha $params.latest"""
+              sh """if [ \$(md5sum /usr/local/bin/amm | awk '{print \$1}') = '07143cdf89e94b0a957f5656193d88d7' ]; then sudo curl -L -o /usr/local/bin/amm https://github.com/lihaoyi/Ammonite/releases/download/1.0.3/2.12-1.0.3 && sudo chmod +x /usr/local/bin/amm; fi"""
+
+              if (params.dryRun == false)
```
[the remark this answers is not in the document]
timcharper: It currently happens here:

```
/Users/tim/src/m8e/marathon-1.5/ci/pipeline

 329 |   val version = build(config.runTests)
 330 |   println(s"Built version ${version}; releasing")
```

ci/pipeline
```diff
@@ -162,30 +163,33 @@ def provisionHost(): Unit = utils.stage("Provision") {
 /**
  * Run the main build pipeline.
  *
- * @return Version of Marathon build.
+ * @return Version of Marathon build (without leading v)
```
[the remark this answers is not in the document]
timcharper: Yeah, that would be nice. In another PR? I have a SemVer class that would be useful: https://github.com/mesosphere/marathon-stuff/blob/master/marathon-jira/lib/util.sc#L16

ci/pipeline
```diff
@@ -246,36 +279,78 @@ def jenkins(): Unit = {
...
+        case None =>
+          throw new RuntimeException(s"${target} is not a known release target; valid options are:\n* ${ReleaseTarget.all.map(_.name).mkString("\n* ")}")
+      }
+    }
+  }
 
-  require(ValidVersionRegex.findFirstMatchIn(tagVersion).nonEmpty,
-    s"tagName should match regex ${ValidVersionRegex}")
+  val config = parser.parse(args, Config()).getOrElse {
+    sys.exit(1)
```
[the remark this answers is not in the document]
timcharper: `???` has a return type of `Nothing`, so it does not affect the inferred value of parser.parse.

Jenkinsfile.release
kensipe: it should be required and checked

ci/pipeline
jeschkies: Thanks for clarifying!

ci/pipeline
jeschkies: I think this could easily be done with the Github API.

Jenkinsfile.release
timcharper: Why would we use withCredentials ? :) For ssh credentials, my understanding is that using an sshagent is the way to provide ssh credentials in Jenkins

PR #5954, opened 2018-01-30
Expose region and zone task info in API

timcharper:
JIRA Issues: MARATHON_EE-1849


alenkacz:
@timcharper I will just note the concern @meichstedt had to open the discussion - do we want to have those items on the top level? Maybe having them nested in faultDomain property would make more sense ... ?

alenkacz:
Also do we want to put it to the pods status endpoint too? Or pods cannot be used with placement constraints?

alenkacz:
@timcharper can someone please re-review the whole thing including the pods and then merge?

PR #5956, opened 2018-01-30
Use three agents for SI test clusters.

jeschkies:
Summary:
Our system integration test cluster would sometimes not launch because
the AWS account ran out of resources. Currently a multi-node cluster requires
ten EC2 instances and three Elastic Load Balancers. With this change we
should only require eight instances per cluster.


jeschkies:
The system integration test is failing for a known flaky test `system.test_marathon_universe.test_neo4j_universe_package_install[neo4j]`.

jeschkies:
> @jeschkies I would just verify we are not skipping more tests that we did before. Otherwise 👍

I did a quick search `rg agents` and it should only two required agents.

PR #5961, opened 2018-01-30
Document disable_http_compression flag

timcharper:
JIRA Issues: MARATHON-8058

### DIFF
--- a/docs/docs/command-line-flags.md
+++ b/docs/docs/command-line-flags.md
@@ -317,6 +317,8 @@ The Web Site flags control the behavior of Marathon's web site, including the us
 * `--leader_proxy_ssl_ignore_hostname` (Optional. Default: false): Do not
     verify that the hostname of the Marathon leader matches the one in the SSL
     certificate when proxying API requests to the current leader.
+* `--[disable_]http_compression` (Optional. Default: enabled): Specifies whether to perform gzip compression on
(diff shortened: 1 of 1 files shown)


docs/docs/command-line-flags.md
```diff
@@ -317,6 +317,8 @@ The Web Site flags control the behavior of Marathon's web site, including the us
 * `--leader_proxy_ssl_ignore_hostname` (Optional. Default: false): Do not
     verify that the hostname of the Marathon leader matches the one in the SSL
     certificate when proxying API requests to the current leader.
+* `--[disable_]http_compression` (Optional. Default: enabled): Specifies whether to perform gzip compression on
+    responses if the client requests as much via the `Accept-Encoding` parameter.
```
wavesoft: > if the client requests as much via the `Accept-Encoding` parameter

I would rephrase this a bit, since this is a standard and expected behaviour. So I could suggest to replace this part with _"... if the client supports it"_. 

According to the [HTTP/1.1 specs](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.3) the server should respond with an encoding the client understands and the `Accept-Encoding` is a way of handling this negotiation process. 

Therefore a lack of an `Accept-Encoding: gzip` header does not really imply that "I am not interested in compression", rather that I don't support it.

wavesoft:
Perfect

PR #5963, opened 2018-01-30
Unreachable Tests

kensipe:
Unreachable Tests

Summary:
SI tests for testing the unreachable strategy for Marathon.

To run:
1) create and setup new  cluster
2) `shakedown -q tests/system/test_unreachable.py::test_unreachable_within_inactive_time[1200-1200]`

This test is currently failing expectations.   1200 seconds is 20 mins.   The new task should not be created until 25 mins after the agent is shutdown.   Currently it is:
Inactive Time from unreachable: 0.215 seconds
Expunge Time from unreachable: 8.207 seconds

The app deployed is:
```
{'id': '/unreachable-9b00a701cf9443dd95549dee80084b07', 'cpus': 0.1, 'mem': 32, 'disk': 0, 'instances': 1, 'container': {'type': 'DOCKER', 'docker': {'image': 'redis', 'network': 'HOST', 'privileged': False, 'forcePullImage': False}}, 'portDefinitions': [{'port': 10013, 'protocol': 'tcp'}], 'unreachableStrategy': {'inactiveAfterSeconds': 1200, 'expungeAfterSeconds': 1200}}
```
JIRA issues:


kensipe:
looks like the tests might be bad... looking into it

tests/system/common.py
```diff
@@ -107,6 +107,10 @@ def command_health_check(command='true', failures=1, timeout=2):
         'maxConsecutiveFailures': failures
     }
 
+def unreachableStrategy(app_def, inactive=100, expunge=200):
```
jeschkies: Hm, what is the motivation for this method if there is a `unreachable` fixture?

tests/system/common.py
```diff
@@ -771,6 +775,35 @@ def running_task_status(task_statuses):
...
+    """
+    client = mesos.DCOSClient()
+    frameworks = client.get_master_state()['frameworks']
+    for framework in frameworks:
+        if framework["name"] == framework_name:
+            return framework['unreachable_tasks']
+    return []
+
+
+def unreachable_task_predicate(service):
```
jeschkies: Could you document this method.

tests/system/common.py
```diff
@@ -771,6 +775,35 @@ def running_task_status(task_statuses):
...
+    return len(unreachable_tasks(service)) > 0
+
+
+def wait_for_unreachable_task(service='marathon', timeout_sec=10*60, inverse=False):
+    return shakedown.time_wait(lambda:unreachable_task_predicate(service), timeout_seconds=timeout_sec, inverse_predicate=inverse, sleep_seconds=1)
+
+
+def marathon_task_predicate(service, app_id, count):
```
jeschkies: What do you mean by predicate?

tests/system/common.py
```diff
@@ -771,6 +775,35 @@ def running_task_status(task_statuses):
...
+        if framework["name"] == framework_name:
+            return framework['unreachable_tasks']
+    return []
+
+
+def unreachable_task_predicate(service):
+    return len(unreachable_tasks(service)) > 0
+
+
+def wait_for_unreachable_task(service='marathon', timeout_sec=10*60, inverse=False):
```
jeschkies: I'd much rather work with retry and assert.

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
+""" Test using root marathon.
```
jeschkies: Our of curiosity, why is this not part of `test_marathon_root` or `marathon_common_tests`?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def teardown_module(module):
+    common.clean_up_marathon()
+
+# This can not be run currently as `shakedown -q tests/system/test_unreachable.py`
+# They currently need a 20 min delay prior to running each one separately or you get
+# inappropriate results.
+# TO RUN:  shakedown -q tests/system/test_unreachable.py::test_unreachable_within_inactive_time[360-360]
+@pytest.mark.parametrize("inactive_sec, expunge_sec", [
```
jeschkies: What is the reason for the different parameters? What are we trying to verify?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
...
+  (600, 600),
+  (1200, 1200)])
+def test_unreachable_within_inactive_time(inactive_sec, expunge_sec):
+
+    common.clean_up_marathon()
+    app_def = apps.unreachable()
+    app_id = app_def['id']
+
+    app_def = common.unreachableStrategy(app_def, inactive=inactive_sec, expunge=expunge_sec)
+    print(app_def)
```
jeschkies: Will this go?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
...
+
+    client = marathon.create_client()
+    client.add_app(app_def)
+    shakedown.deployment_wait()
```
jeschkies: Please add the app id when waiting.

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
...
+    tasks = client.get_tasks(app_id)
+
+    # confirm task and get task_id and host details.
+    assert len(tasks) == 1, "The number of tasks is {} after deployment, but only 1 was expected".format(len(tasks))
```
jeschkies: Why is this necessary here? We are not testing deployments.

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
...
+    original_task_id = tasks[0]['id']
+    original_host = tasks[0]['host']
+    print('Starting Unreachable Test IP: {}'.format(original_host))
+    print("Testing 'inactiveAfterSeconds':{},'expungeAfterSeconds':{}".format(inactive_sec, expunge_sec))
+    start = time.time()
+    # losing an agent
+    shakedown.stop_agent(original_host)
```
jeschkies: How about blocking ports instead?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,91 @@
...
+    time.sleep(240)
+    common.wait_for_unreachable_task()
```
jeschkies: I'd much rather see a retrying/assert combo instead of sleep/wait. The former is more robust against flakiness.

jeschkies: What about using the event stream instead?

tests/system/common.py
kensipe: used to created a variety of different unreachable configurations used by the test_unreachable.pyL65 for the parameterized testing.

tests/system/common.py
kensipe: uses shakedown as a standard with the added benefit of providing stdout feedback for long wait periods.   I prefer to stick with this.

tests/system/test_unreachable.py
kensipe: This test in it's current form is more for data gathering than it is a "test".   There are assertions for certain points... but the data output is more important and the goal.

It is not in `test_marathon_root` because it isn't exactly a test and it isn't something we want to run in the same way.   We will eventually create a test (which will also not be in the traditional test modules) for testing a 24hr timeout.    This will be separated out for what I hope is logical reasons.

tests/system/test_unreachable.py
kensipe: unreachable timings from the customers perspective do not work the way I would expect them to....   This is gather a number of different combinations to evaluate the details.

tests/system/test_unreachable.py
kensipe: I prefer it to.. yeah

tests/system/test_unreachable.py
kensipe: it asserts the valid start of the tests... it can be removed if you have strong objections.

kensipe:
sleeps in this case are not flaky... and actually may be necessary.

tests/system/test_unreachable.py
jeschkies: Ah, ok. Thanks for clarifying 🙂.

tests/system/test_unreachable.py
jeschkies: Got it.

tests/system/common.py
jeschkies: Hm, I actually find the shakedown logs rather "polluting". retry with assert gives a little cleaner feedback in the test reports. See https://github.com/dcos/dcos/pull/2426 vs https://github.com/dcos/dcos/pull/2421.

It's not a big deal but I preferr errors such as `assert 'DEGRADED' == 'STABLE' failed` to `TimeoutError`.

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
     assert False, "Did not find a TASK_RUNNING status in task statuses: %s" % (task_statuses,)
 
 
+def unreachable_tasks(framework_name='marathon'):
+    """ Returns the unreachable tasks for a given framework.
+    """
+    client = mesos.DCOSClient()
+    frameworks = client.get_master_state()['frameworks']
```
jeschkies: Do we have this more often? If so we might want to add a helper method

```python
def get_framework_state(framework_name):
    frameworks = client.get_master_state()['frameworks']
    for framework in frameworks:
        if framework["name"] == framework_name:
            return framework
    return None

def unreachable_tasks(framework_name='marathon'):
    framework = get_framework_state(framework_name):
    if framework is not None:
        return framework['unreachable_tasks']
    else:
        return []
```

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+def mesos_tasks(framework_name='marathon'):
+    client = mesos.DCOSClient()
+    frameworks = client.get_master_state()['frameworks']
+    for framework in frameworks:
+        if framework["name"] == framework_name:
+            return framework['tasks']
+    return []
+
+
+def unreachable_task_predicate(service):
```
jeschkies: Could you add docs. I'm also not sure what `unreachable_task_predicate` means.

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+    @retrying.retry(wait_fixed=1000, stop_max_delay=timeout_sec*1000)
+    def __wait_for_unreachable_task(start=time.time(), msg='unreachable tasks'):
+        elapsed = elapse_time(start, time.time())
+        if (elapsed % report_freq == 0):
+            print("Waiting on {} at {}s".format(msg, elapsed))
+
+        if inverse:
+            assert not unreachable_task_predicate(service)
+        else:
+            assert unreachable_task_predicate(service)
```
jeschkies: Why not `assert len(unreachable_tasks(service))`? That seems easier to read to me.

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+
+
+def wait_for_unreachable_task(service='marathon', timeout_sec=10*60, inverse=False):
+
+    report_freq = 5 if timeout_sec < 600 else 15
+    print("Starting wait for unreachable tasks")
+    @retrying.retry(wait_fixed=1000, stop_max_delay=timeout_sec*1000)
+    def __wait_for_unreachable_task(start=time.time(), msg='unreachable tasks'):
+        elapsed = elapse_time(start, time.time())
+        if (elapsed % report_freq == 0):
```
jeschkies: Hm, this is a little confusing. What is the reason for not reporting this with each try?

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+
+def unreachable_task_predicate(service):
+    return len(unreachable_tasks(service)) > 0
+
+
+def elapse_time(start, end, precision=0):
+    return round(end - start, precision)
+
+
+def wait_for_unreachable_task(service='marathon', timeout_sec=10*60, inverse=False):
```
jeschkies: Hm, how about defining the inverse as `wait_for_reachable_task`?

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+
+        if inverse:
+            assert not unreachable_task_predicate(service)
+        else:
+            assert unreachable_task_predicate(service)
+
+    __wait_for_unreachable_task()
+
+
+def marathon_task_predicate(app_id, task_id):
```
jeschkies: This name would suggest to me that there is a Marathon task. What do you think about `app_contains_task(app_id, task_id)`?

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+    """ Returns true if there is a task that is NOT the task_id
+    """
+    client = marathon.create_client()
+    tasks = client.get_tasks(app_id)
+    task = task_other_than_task_id(tasks, task_id)
```
jeschkies:
```python
task_ids = set([t.id for t in tasks])
return task_id in task_ids
```

or 

```python
return task_id in [t.id for t in tasks]
```

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+        elapsed = elapse_time(start, time.time())
+        if (elapsed % 5 == 0):
+            print("Waiting on {} at {}s".format(msg, elapsed))
+
+        assert marathon_task_predicate(app_id, task_id)
+
+    __wait_for_marathon_task()
+
+
+def task_remove_predicate(app_id, task_id):
```
jeschkies: I'm not sure that this is tested in the method. It seems to test that there is only one task. Not that the task has been removed.

tests/system/common.py
```diff
@@ -771,6 +778,99 @@ def running_task_status(task_statuses):
...
+
+def wait_for_unreachable_task_kill(app_id, task_id, timeout_sec=6*60):
+    print("Starting wait for expunge of task")
+    @retrying.retry(wait_fixed=1000, stop_max_delay=timeout_sec*1000)
+    def __wait_for_unreachable_task_kill(start=time.time(), msg='expunge of task'):
+        elapsed = elapse_time(start, time.time())
+        if (elapsed % 5 == 0):
+            print("Waiting on {} at {}s".format(msg, elapsed))
+
+        assert task_remove_predicate(app_id, task_id)
```
jeschkies: I think we had this in another review. What do you think about using the event stream for this?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,233 @@
...
+    assert len(tasks) == 1, "The number of tasks is {} after deployment, but only 1 was expected".format(len(tasks))
+    original_task_id = tasks[0]['id']
+    original_host = tasks[0]['host']
+    print('Starting Unreachable Test IP: {}'.format(original_host))
+    print("Testing 'inactiveAfterSeconds':{},'expungeAfterSeconds':{}".format(inactive_sec, expunge_sec))
+    print("task id: {}".format(original_task_id))
+
+    start = time.time()
+    # losing an agent
+    shakedown.stop_agent(original_host)
```
jeschkies: Why not `with agents_shutdown`?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,233 @@
...
+    print("Sleeping for {}".format(initial_unreachable_delay()))
+    time.sleep(initial_unreachable_delay())
```
jeschkies: Why the sleep if we wait for the unreachable task?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,233 @@
...
+    # this test ignores node rate limiting
+    common.wait_for_unreachable_task()
+    unreachable_time = time.time()
+    print("Unreachable Time: {} seconds".format(elapse_time(start, unreachable_time)))
+    print('Agent and Task Reported Unreachable.  Wait for inactiveAfterSeconds to engage.')
+
+    # recovery based on inactive time (len(tasks) goes from 1 to 0 based on unreachable.  now we wait for 1.)
+    # task recovery is dependent on inactive_sec and the reconcilation time window
+    print("Sleeping for {}".format(inactive_sec*0.7))
+    time.sleep(inactive_sec*0.7)
```
jeschkies: What are we waiting for? I'd much rather wait for a condition.

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,233 @@
...
+    print("new task id: {}".format(new_task_id))
+    print("Inactive Time from start: {} seconds".format(elapse_time(start, inactive_time)))
+    print("Inactive Time from unreachable: {} seconds".format(elapse_time(unreachable_time, inactive_time)))
+
+    # recovery
+    shakedown.start_agent(original_host)
+
+    # first the unreachables go away (because they become reachable)
+    # this will result in the app indicating 2 of 1 until the unreachable is killed
+    common.wait_for_unreachable_task(inverse=True)
```
jeschkies: I'm not really sure what this method is waiting for given the comment above.

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,233 @@
...
+    print("Remaining task id: {}".format(task_remaining))
+
+
+def initial_unreachable_delay():
+    """ Sleep the first 75% of the time.
+    """
+    return 0.75 * DCOS_AGENT_UNREACHABLE_TIME
+
+
+def elapse_time(start, end, precision=0):
```
jeschkies: Is this the right spot for such a method?

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,233 @@
...
+    print("Expunge Time from unreachable: {} seconds".format(elapse_time(unreachable_time, expunge_time)))
+    tasks = client.get_tasks(app_id)
+    task_remaining = tasks[0]['id']
+    print("Remaining task id: {}".format(task_remaining))
+
+
+def initial_unreachable_delay():
+    """ Sleep the first 75% of the time.
+    """
+    return 0.75 * DCOS_AGENT_UNREACHABLE_TIME
```
jeschkies: Where is `0.75` coming from?

tests/system/apps/__init__.py
```diff
@@ -93,3 +93,8 @@ def ipv6_healthcheck():
         For more information about the nc options just run `docker run alpine nc --help`
     """
     return load_app('ipv6-healthcheck')
+
+def unreachable():
+    """ The app use to test the unreachable strategy
```
alenkacz: used

tests/system/common.py
```diff
@@ -108,6 +109,12 @@ def command_health_check(command='true', failures=1, timeout=2):
     }
 
 
+def unreachableStrategy(app_def, inactive=100, expunge=200):
```
alenkacz: inactiveSeconds, expungeSeconds

ci/launch_cluster.sh
```diff
@@ -27,6 +27,12 @@ fi
 CHANNEL="$1"
 VARIANT="$2"
 DEPLOYMENT_NAME="$3"
+EXPIRATION="2hours"
```
alenkacz: what is the default expiration in dcos_launch if we don't specify anything? Is it 2h?

tests/system/common.py
```diff
@@ -771,6 +778,105 @@ def running_task_status(task_statuses):
     assert False, "Did not find a TASK_RUNNING status in task statuses: %s" % (task_statuses,)
 
 
+def unreachable_tasks(framework_name='marathon', app_id=None):
+    """ Returns the unreachable tasks for a given framework.
+    """
+    client = mesos.DCOSClient()
+    frameworks = client.get_master_state()['frameworks']
+    for framework in frameworks:
```
alenkacz: we might be able to put this if "if framework["name"] == framework_name:" directly on this line to reduce the nesting?

tests/system/common.py
```diff
@@ -771,6 +778,105 @@ def running_task_status(task_statuses):
...
+                for unreachable_task in framework['unreachable_tasks']:
+                    if unreachable_task['name'] == app_id:
+                        return unreachable_task
+
+    return []
+
+def mesos_tasks(framework_name='marathon'):
+    client = mesos.DCOSClient()
+    frameworks = client.get_master_state()['frameworks']
+    for framework in frameworks:
```
alenkacz: this can be done on one line using generators?

tests/system/common.py
```diff
@@ -771,6 +778,105 @@ def running_task_status(task_statuses):
...
+        if framework["name"] == framework_name:
+            return framework['tasks']
+    return []
+
+
+def unreachable_task_predicate(service, app_id=None):
+    return len(unreachable_tasks(service, app_id)) > 0
+
+
+def elapse_time(start, end, precision=0):
```
alenkacz: elapsed

tests/system/common.py
```diff
@@ -771,6 +778,105 @@ def running_task_status(task_statuses):
...
+    @retrying.retry(wait_fixed=1000, stop_max_delay=timeout_sec*1000)
+    def __wait_for_unreachable_task(start=time.time(), msg='unreachable tasks'):
+        elapsed = elapse_time(start, time.time())
+        if (elapsed % report_freq == 0):
+            print("Waiting on {} at {}s".format(msg, elapsed))
+
+        if inverse:
+            assert not unreachable_task_predicate(service, app_id)
+        else:
+            assert unreachable_task_predicate(service, app_id)
```
alenkacz: maybe we can add some custom message to this assert to ease the debugging?

tests/system/common.py
```diff
@@ -771,6 +778,105 @@ def running_task_status(task_statuses):
...
+    """ Returns true if there is a task that is NOT the task_id
+    """
+    client = marathon.create_client()
+    tasks = client.get_tasks(app_id)
+    task = task_other_than_task_id(tasks, task_id)
+    return task is not None
+
+
+def task_other_than_task_id(tasks, task_id):
+    for task in tasks:
```
alenkacz: I would use generators here again :)

tests/system/test_unreachable.py
alenkacz: btw. we don't have to use format anymore since we are on python 3.6 and can use interpolation 🎉

tests/system/test_unreachable.py
```diff
@@ -0,0 +1,373 @@
...
+DCOS_MAX_AGENT_PING_TIMEOUTS=20
+# DCOS default is 15*20 = 300s == 5m
+DCOS_AGENT_UNREACHABLE_TIME=DCOS_AGENT_PING_TIMEOUT * DCOS_MAX_AGENT_PING_TIMEOUTS
+
+# node rate limit assume per 1 node in X mins.  The current default is 1/20 mins
+# this is in case we want to automate the delay of each test.
+DCOS_NODE_RATE_LIMIT=20
+
+# marathon default reconcilation window is 600s == 10m
+MARATHON_RECONCILATION_INTERNAL=600
```
alenkacz: did you mean interval?

tests/system/common.py
alenkacz: but it is more app_not_contains_task right?

tests/system/common.py
```diff
@@ -771,6 +778,105 @@ def running_task_status(task_statuses):
...
+        if task['id'] != task_id:
+            return task
+    return None
+
+
+def wait_for_marathon_task(app_id, task_id, timeout_sec=6*60):
```
alenkacz: can we find a better name? wait_for_app_without_task ?

PR #5966, opened 2018-01-31
documentation update for the 1.5.6 release

meln1k:


meln1k:
Well, we can just copy the same tar with a different path, right? Or do you think it might cause some problems?

ichernetsky:
LGTM, but let's wait for Tim's input.

timcharper:
Yeah, let's copy the jar file, and keep it consistent.

PR #5970, opened 2018-02-01
Version Marathon and name its artifacts.

jeschkies:
Summary:
With this change Marathon will be version with <major>.<minor>.<build-no>.
The build number is the number of commits since the last branch point of
a release branch. E.g. `releases/1.5` was branched at `aa89e89`. This is the
common ancestor for 1.5 and the current release. If master or `releases/1.6`
is 42 commits ahead of `aa89e89`, the build number is 42.

The build artifacts are named as follows
* `marathon-1.6.42-3e42d0d.tgz` for Tarball
* `marathon-docs-1.6.42-3e42d0d.tgz` for dcouments Tarball
* `marathon_2.12-1.6.42.jar` for Jar
* `marathon_1.6.42_all.deb` for Debian
* `marathon-1.6.42.3e42d0d-1.noarch.rpm` for RPM
* `mesosphere/marathon:v1.6.42` for Docker images

That means the tarball packages include the commit hash in their names
so that it is easier to find the commit when we integrate a build into
DC/OS.

JIRA issues: MARATHON_EE-1759


build.sbt
```diff
@@ -282,6 +296,10 @@ lazy val marathon = (project in file("."))
   .settings(formatSettings: _*)
   .settings(packagingSettings: _*)
   .settings(
+    version := {
+      import sys.process._
+      ("./version" !!).trim
```
jeschkies: The version script is the definite source of Marathon's version.

ci/pipeline
```diff
@@ -88,26 +88,9 @@ def uploadTarballPackagesToS3(version: String, buildLocation: String): Option[aw
  */
 @main
 def createTarballPackages(): String = utils.stage("Package Tarballs") {
```
jeschkies: I also want to change the type to Tim's `SemVer`.

ci/pipeline
```diff
@@ -189,32 +171,6 @@ def buildAndTestDockerAndLinuxPackages(): Unit = {
   }
 }
 
-/**
```
jeschkies: We can introduce it again if it is needed. However, as I understand the tags will be defined by the build version.

ci/pipeline
```diff
@@ -162,15 +145,16 @@ def provisionHost(): Unit = utils.stage("Provision") {
...
 
   val logFileName = s"ci-${sys.env.getOrElse("BUILD_TAG", "run")}.log"
-  try {
-    compileAndTest(logFileName)
-  } finally {
-    zipLogs(logFileName)    // Try to archive ci and sandbox logs in any case
-  }
+// Fast iteration.
+//  try {
+//    compileAndTest(logFileName)
```
jeschkies: I'll enable it again once everything built properly.

jeschkies:
The rpm package name is off. Not sure where this comes from but it's called `marathon-1.6.301.8143440-SNAPSHOT.noarch.rpm`.

jeschkies:
Two outstanding issues:
[ ] The wrong docker image is tested
[ ] The git hash seems to be wrong

jeschkies:
The git hash is of the merge since we test `pr-merge`. @timcharper, would you have an idea why the Docker test is picking a different version?

build.sbt
```diff
@@ -153,10 +150,28 @@ lazy val packagingSettings = Seq(
   mappings in (Compile, packageDoc) := Seq(),
   debianChangelog in Debian := Some(baseDirectory.value / "changelog.md"),
 
+  (packageName in Universal) := {
+    git.gitHeadCommit.value match {
+      case Some(commit) =>
+        val shortCommit = commit.take(7)
```
meln1k: why 7?

ichernetsky: It is a convention. Git SHAa are usually shortened to 7 chars. Though, I agree it would be nice to have it as a variable.

version
```diff
@@ -0,0 +1,16 @@
...
+#
+# Change MINOR to x+1 and BRANCH_POINT to commit hash of common ancestor of master and releases/1.x
+# after a releases/1.x was cut
+MAJOR=1
+MINOR=6
+BRANCH_POINT=aa89e89
+
+# Infer version
+# Number of commits since branch point
+BUILD_NUMBER="$(git rev-list --count --first-parent $BRANCH_POINT..HEAD)"
```
ichernetsky: Why not `git rev-list --count HEAD ^master`?

version
```diff
@@ -0,0 +1,16 @@
...
+
+# E.g. 1.7.42
+echo "$MAJOR.$MINOR.$BUILD_NUMBER"
```
ichernetsky: Why is a patch number a build number? Is it as intended that each commit increments the patch number?

version
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Version script for Marathon.
+#
+# Change MINOR to x+1 and BRANCH_POINT to commit hash of common ancestor of master and releases/1.x
+# after a releases/1.x was cut
+MAJOR=1
```
ichernetsky: What is the advantage of getting the version from `version` file instead of getting it from `git`?

ci/pipeline
timcharper: Is the plan for us to create tags for versions that we deem as stable?

ci/utils.sc
```diff
@@ -136,3 +136,20 @@ def escapeCmdArg(cmd: String): String = {
   val subbed = cmd.replace("'", "\\'").replace("\n", "\\n")
   s"""$$'${subbed}'"""
 }
+
+case class SemVer(major: Int, minor: Int, build: Int, commit: String) {
+  def toStringtWithCommit(): String = s"$major.$minor.$build-$commit"
```
timcharper: No prefix here? Perhaps this should just be the default `toString`?

ci/utils.sc
```diff
@@ -136,3 +136,20 @@ def escapeCmdArg(cmd: String): String = {
   val subbed = cmd.replace("'", "\\'").replace("\n", "\\n")
   s"""$$'${subbed}'"""
 }
+
+case class SemVer(major: Int, minor: Int, build: Int, commit: String) {
```
timcharper: Do you think we should rename this version, since it's no longer really a semantic version? Perhaps MarathonVersion?

version
timcharper: Preference to name this variable "COMMIT_NUMBER"

I like the script. The divergence point isn't always discernible by the master branch, right?

build.sbt
timcharper: Is there a strong reason to not get this variable from the version script?

ci/pipeline
jeschkies: As I understood we only tag community releases. A community release was soaked for two weeks and thus is considered stable. @meichstedt and @kensipe what are your takes on this?

build.sbt
jeschkies: Not really. I wanted to keep the version script simple. I'll add a `version commit`.

version
jeschkies: `git` is not so flexible at least I did not find it too be so. Also, major and minor version have to be persisted somewhere. Matthias, Ken and I agreed that it should not be a tag since tags should be reserved for community releases.

version
jeschkies: > Why not `git rev-list --count HEAD ^master`?

We don't always run this on master. And Tim is correct, it is not always trivial to find divergence points from branches.

version
jeschkies: I'll change it to commit number. And yes, it increments with each commit.

ci/utils.sc
jeschkies: Good point.

jeschkies:
> I believe, native packaging guidelines should be taken into account too. For instance, I cannot recall a distro in which `SNAPSHOT` are allowed in package names/versions, speaking of the example `marathon-1.6.42.3e42d0d-SNAPSHOT.noarch.rpm`

You are right. It should be `marathon-1.6.42.3e42d0d.noarch.rpm`.

jeschkies:
@ichernetsky, the RPM build still attaches `SNAPSHOT` to the artifact. I think this is by convention but don't know where that is defined.

ci/utils.sc
jeschkies: I'll rename it with the forward port of your changes.

ci/pipeline
meichstedt: @jeschkies tagging community releases seems sufficient to me. As stated in the scope doc
> Promoting the above build to a community release means tagging the related commit [...] and publishing relevant artifacts [...].

Depending on how often/how many builds we soak, we could tag every build that has been soaked and is considered stable; Once every 4-6 weeks should be enough though, so we don't end up with too many versions.

(In a later improvement I'd like to integrate with JIRA and not only tag these commits, but also create the according versions in JIRA).

PR #5973, opened 2018-02-02
Optimize `v2/groups` request performance

zen-dog:
Improved `v2/groups?embed...` requests performance by introducing dedicated thread pools for fetching info instead of using the `ForkJoinPool`. 
Additionally introduces dedicated execution context for `GroupManagerModule`, `StorageModule` and `SchedulerActions` components.

[Benchmark](https://github.com/wg/wrk) results master:
```
Running 1m test @ http://localhost:8080/v2/groups?embed=...
  10 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.17s   354.93ms   2.00s    70.33%
    Req/Sec     0.12      0.33     1.00     87.76%
  237 requests in 1.00m, 217.12MB read
  Socket errors: connect 0, read 0, write 0, timeout 146
Requests/sec:      3.95
Transfer/sec:      3.62MB
``` 

This branch:
```
Running 1m test @ http://localhost:8080/v2/groups?embed=...
  10 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   249.39ms   62.67ms 580.40ms   65.82%
    Req/Sec     4.01      1.82    10.00     85.60%
  2403 requests in 1.00m, 2.15GB read
Requests/sec:     40.00
Transfer/sec:     36.64MB
```
**Summary**: 10x throughput, 5x latency and 5x std. deviation improvement.

JIRA: COPS-2224

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -35,9 +35,10 @@ import mesosphere.marathon.core.task.termination.TaskTerminationModule
 import mesosphere.marathon.core.task.tracker.InstanceTrackerModule
 import mesosphere.marathon.core.task.update.TaskStatusUpdateProcessor
 import mesosphere.marathon.storage.StorageModule
+import mesosphere.util.NamedExecutionContext
 import mesosphere.util.state.MesosLeaderInfo
-import scala.concurrent.ExecutionContext
 
+import scala.concurrent.ExecutionContext
 import scala.util.Random
 
 /**
@@ -81,7 +82,7 @@ class CoreModuleImpl @Inject() (
   )
 
   // TASKS
-  val storageExecutionContext = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(8))
+  val storageExecutionContext = NamedExecutionContext.fixedThreadPoolExecutionContext(8, "storage-module")
   override lazy val taskTrackerModule =
     new InstanceTrackerModule(clock, marathonConf, leadershipModule,
       storageModule.instanceRepository, instanceUpdateSteps)(actorsModule.materializer)
@@ -199,7 +200,7 @@ class CoreModuleImpl @Inject() (
 
   // GROUP MANAGER
 
-  val groupManagerExecutionContext = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(8))
+  val groupManagerExecutionContext = NamedExecutionContext.fixedThreadPoolExecutionContext(8, "group-manager-module")
   override lazy val groupManagerModule: GroupManagerModule = new GroupManagerModule(
     marathonConf,
     scheduler,
@@ -261,7 +262,7 @@ class CoreModuleImpl @Inject() (
   //
   // TODO: this can be removed when MarathonSchedulerActor becomes a core component
 
-  val schedulerActionsExecutionContext = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(8))
+  val schedulerActionsExecutionContext = NamedExecutionContext.fixedThreadPoolExecutionContext(8, "scheduler-actions")
   override lazy val schedulerActions: SchedulerActions = new SchedulerActions(
     storageModule.groupRepository,
     healthModule.healthCheckManager,
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -81,7 +81,7 @@ class CoreModuleImpl @Inject() (
   )
 
   // TASKS
-
+  val storageExecutionContext = ExecutionContext.fromExecutor(Executors.newFixedThreadPool(8))
   override lazy val taskTrackerModule =
     new InstanceTrackerModule(clock, marathonConf, leadershipModule,
       storageModule.instanceRepository, instanceUpdateSteps)(actorsModule.materializer)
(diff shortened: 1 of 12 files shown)


src/main/scala/mesosphere/marathon/core/appinfo/AppInfoModule.scala
```diff
@@ -21,8 +21,11 @@ class AppInfoModule @Inject() (
     healthCheckManager: HealthCheckManager,
     marathonSchedulerService: MarathonSchedulerService,
     taskFailureRepository: TaskFailureRepository) {
+
+  val ec = NamedExecutionContext.fixedThreadPoolExecutionContext(8, "app-info-module")
```
kensipe: `ec`?   are we trying to stay within a linter line length rule :) 

could we make this more readable pls?

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
```diff
@@ -17,15 +18,14 @@ import scala.concurrent.Future
 
 private[appinfo] class DefaultInfoService(
     groupManager: GroupManager,
-    newBaseData: () => AppInfoBaseData) extends AppInfoService with GroupInfoService with PodStatusService {
-  import mesosphere.marathon.core.async.ExecutionContexts.global
+    newBaseData: () => AppInfoBaseData) extends AppInfoService with GroupInfoService with PodStatusService with StrictLogging {
 
-  private[this] val log = LoggerFactory.getLogger(getClass)
+  implicit val ec = NamedExecutionContext.fixedThreadPoolExecutionContext(8, "default-info-service")
```
kensipe: ditto

src/main/scala/mesosphere/marathon/core/appinfo/AppInfoModule.scala
zen-dog: There are a few akka/scala specific and accepted "short names" e.g. `ec` for `ExecutionContext`, `mat` for `Materializer` etc. Falls into the same category as  loop counter `for(i=0,..)`

timcharper: Should this thread pool be ideally related to the number of threads available?

Should we make it configurable? (even if via an undocumented environment variable)?

It seems like we should ideally create the execution context in the module and share it with the pods status controller?

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -199,10 +200,11 @@ class CoreModuleImpl @Inject() (
 
   // GROUP MANAGER
 
+  val groupManagerExecutionContext = NamedExecutionContext.fixedThreadPoolExecutionContext(8, "group-manager-module")
```
meichstedt: Why 8 threads? Did we measure that 8 is a meaningful number? I'm not trying to overcomplicate things, but it seems to me that making these configurable would allow us to also easily run different tests to verify that the number makes sense.

src/main/scala/mesosphere/marathon/core/appinfo/AppInfoModule.scala
meichstedt: `ec` is fine for me for the reason @zen-dog pointed out.
I agree with @timcharper that the number of threads could as well be configurable.

src/main/scala/mesosphere/util/ThreadPoolContext.scala
```diff
@@ -18,5 +19,21 @@ object ThreadPoolContext {
...
   )
+}
+
+object NamedExecutionContext {
 
+  /**
+    * Returns an execution context backed by a fixed thread pool. All threads in the pool are prefixed with `namePrefix`
+    * e.g. `slow-io-pool-thread-1`.
+    *
+    * @param numThreads number of threads in the pol
```
meichstedt: nit: s/pol/pool

src/main/scala/mesosphere/marathon/core/appinfo/impl/DefaultInfoService.scala
jeschkies: I wonder if this really has an impact if app info and the group manager have their own thread pools. I'd like to understand where we benefit from a fixed worker thread pool and where we should be using the fork join thread pool.

zen-dog: I tested the scenario with one thread pool and it adds roughly 20% response time. Not that much but I think dividing them still makes sense - Dīvide et Imperā!

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
zen-dog: I did measurements with the different number of threads. Having more than 8 threads does not produce better numbers.

zen-dog:
> Should this thread pool be ideally related to the number of threads available?
> Should we make it configurable? (even if via an undocumented environment variable)?
I don't think there is much value in setting it to `availableProcessors` or whatever.=. We already have chaos creating a thread for each connection, akka dispatcher pools and a bunch of other threads. And when we used to block everywhere our thread count would grow > 1k threads at times. 
But I do agree that it makes sense to make it configurable.

> It seems like we should ideally create the execution context in the module and share it with the pods status controller?
Shared execution context showed 20% bigger response times (for whatever reason).

src/main/scala/mesosphere/marathon/core/appinfo/AppInfoConfig.scala
```diff
@@ -0,0 +1,21 @@
+package mesosphere.marathon
+package core.appinfo
+
+import org.rogach.scallop.ScallopConf
+
+trait AppInfoConfig extends ScallopConf {
+
+  lazy val appInfoModuleExecutionContextSize = opt[Int](
```
timcharper: Let's base the default as `Runtime.getRuntime().availableProcessors() * 2`? (or... just # of availableProcessors?)

PR #5971, opened 2018-02-02
Fix package uploading and improvements to release pipeline

timcharper:
Forwardport of a673f2e9

* Fix package uploading and improvements to release pipeline 
* Augment ci/pipeline script to have parameterized release targets
* Upgrade-in-place to Ammonite 1.0.3 if not already done
* Dry run mode
* Use JenkinsMarathonCI-Debian8-2017-10-06 node as releases/1.5.
* Use SemVer class to reduce guesswork of 'v' prefix or not

PR #5974, opened 2018-02-02
Script for generating versioned documentation for marathon

meln1k:
Summary: Ammonite script for docs generation

JIRA issues: MARATHON-3213


scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env amm
```
jeschkies: I wonder if this should be in `ci`. What do you think?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
...
+  */
+
+def makeTmpDir(): Path = {
+  val timestamp = Instant.now().toString.replace(':', '-')
+  val path = root/"tmp"/s"marathon-docs-build-$timestamp"
+  mkdir! path
+  path
+}
+
+val latestReleaseVersion = "1.5"
```
alenkacz: can we somehow make this not manual? I mean now we have to think that with every release, we must update this script...

scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env amm
+
+import java.time.Instant
+import ammonite.ops._
+import ammonite.ops.ImplicitWd._
+
+/**
+  * Docs generation process:
+  *
+  * 1. copy docs/docs and docs/_layouts/docs.html part from 1.3, 1.4 and 1.5 branches to the temp dirs
```
jeschkies: Hm. Why is this step necessary? Sorry for my ignorance.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
...
+def makeTmpDir(): Path = {
+  val timestamp = Instant.now().toString.replace(':', '-')
+  val path = root/"tmp"/s"marathon-docs-build-$timestamp"
+  mkdir! path
+  path
+}
+
+val latestReleaseVersion = "1.5"
+
+val marathonVersions = List("1.1", "1.3", "1.4", latestReleaseVersion)
```
alenkacz: there was no marathon 1.2? 🤔

scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
...
+  *
+  * 1. copy docs/docs and docs/_layouts/docs.html part from 1.3, 1.4 and 1.5 branches to the temp dirs
+  * 2. copy the latest release (1.5_ branch doc to the temp dir
+  * 3. generate the latest release docs
+  * 4. replace docs/docs with docs/docs from 1.3, 1.4, 1.5 and generate docs for the _sites/docs_version
+  */
+
+def makeTmpDir(): Path = {
+  val timestamp = Instant.now().toString.replace(':', '-')
+  val path = root/"tmp"/s"marathon-docs-build-$timestamp"
```
jeschkies: Where is `root` coming from?

scripts/generate_docs.sc
jeschkies: Hm. This would be another version to hardcode. I'd opt for using the version script from #5970.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
...
+
+val marathonVersions = List("1.1", "1.3", "1.4", latestReleaseVersion)
+
+val buildDir = makeTmpDir()
+
+val docsDir = pwd/up/"docs"
+
+%ls docsDir
+
+%git('checkout, "master")
```
jeschkies: Maybe I'm missing something here. Where and when is this run? Should the caller decide which commit is built?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,101 @@
...
+def generateDocsForVersion(docsPath: Path, version: String, outputPath: String = "_site"): Unit = {
+  %('bundle, "exec", s"jekyll build --config _config.yml,_config.$version.yml -d $outputPath/$version/")(docsPath)
+}
+
+def branchNameForVersion(version: String) = {
+  s"releases/$version"
+}
+
+
+println(buildDir)
```
alenkacz: is this intentional?

scripts/generate_docs.sc
meln1k: I don't want to mess with git branches too much, this step makes a clear separation between docs fetching from branches and following docs generation processes

scripts/generate_docs.sc
meln1k: It's predefined in ammonite

scripts/generate_docs.sc
meln1k: well, it's not exactly version, more a name of the release branch

scripts/generate_docs.sc
meln1k: Good catch, this is a dead code

scripts/generate_docs.sc
meln1k: It's just a name of the release branch. I don't think it will be changed too often

scripts/generate_docs.sc
meln1k: Good point :)

scripts/generate_docs.sc
meln1k: not really

scripts/generate_docs.sc
```diff
@@ -0,0 +1,95 @@
...
+def makeTmpDir(): Path = {
+  val timestamp = Instant.now().toString.replace(':', '-')
+  val path = root/"tmp"/s"marathon-docs-build-$timestamp"
+  mkdir! path
+  path
+}
+
+val latestReleaseVersion = "1.5"
+
+val marathonVersions = List("1.3", "1.4", latestReleaseVersion)
```
meichstedt: While this is not automated: why not just `List("1.3", "1.4", "1.5")`? If we update the `latestReleaseVersion` we also have to adjust this list to include 1.5. I'd prefer to only have one place to edit, or even none :)

scripts/generate_docs.sc
meichstedt: I agree that this shouldn't be hardcoded. It should be rather simple to compute the latest version.

However I think for the first step it's ok to have it manual. Let's iterate from there and improve in the next step.

What I'd suggest to get rid of the hardcoded values and prevent us from publishing docs based on unreleased commits:
1. list all `vX.Y.Z` **tags**, sort them and take the 3 latest _minor_ tags (currently this should result in 1.3.13, 1.4.11, 1.5.7). We don't plan so bump the major version any time soon.
2. Add `master`. We should always generate docs for master.
3. Generate docs for the resulting list of ("1.3.13", "1.4.11", "1.5.7", "master").

Note: It makes sense to me that the docs are contained in _major.minor_  folders. It would be great to add information about which patch release is the documented one (e.g. "1.5.7"), but maintain those docs in a "1.5" folder.

scripts/generate_docs.sc
meichstedt: It should eventually be a jenkins job :)

scripts/generate_docs.sc
meichstedt: As written above, I'd go for tags instead.

zen-dog: `SemVer` class from this PR will be helpful once merged:
https://github.com/mesosphere/marathon/pull/5971/files#diff-06a37bc5ba23ab5621562efe4a70a0e1R141

scripts/generate_docs.sc
```diff
@@ -0,0 +1,95 @@
...
+  */
+
+def makeTmpDir(): Path = {
+  val timestamp = Instant.now().toString.replace(':', '-')
+  val path = root/"tmp"/s"marathon-docs-build-$timestamp"
+  mkdir! path
+  path
+}
+
+val latestReleaseVersion = "1.5"
```
zen-dog: If we want to keep it manual then `val latestReleaseVersion = marathonVersions.tail`. But I'd rather have what Matthias suggested: generate list of tags, use Tim's `SemVer` to pick 3 last major/minor versions, take the docs from those tags.

scripts/generate_docs.sc
jeschkies: Ah. Thanks.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,109 @@
...
+  mkdir! path
+  path
+}
+
+val ignoredVersions = Set("1.6.0")
+
+/**
+  * getting 3 latest releases tags to build docs for
+  */
+val docsTargetVersions = %%('git, "tag", "-l", "--sort=version:refname").out.lines.filter(_.matches("""v[1-9]+\.\d+\.\d+""")).groupBy(v => v.drop(1).takeWhile {
```
jeschkies: Hm, could you give an example? Should we just use the version script?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,109 @@
...
+}
+
+def tagNameForVersion(version: String) = {
+  s"tags/v$version"
+}
+
+
+// step 1: copy docs/docs to the respective dirs
+
+docsTargetVersions foreach { version =>
```
jeschkies: What do you think about making these steps main methods as we've done in `ci/pipeline`?

meln1k: I think it's a good idea, but I'd rather do it in the next iteration. This script depends on other branches that aren't merged yet, and it needs to be improved after those branches will be merged.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env amm
+
+/**
+  *
+  * DO NOT RUN THIS SCRIPT FROM THE SHELL!
```
kensipe: This comment makes no sense..   
first if not a shell then what?
second... if you give a warning like this why needs to be explained.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,147 @@
...
+  *
+  * run ./doc_generation_launcher.sh instead
```
kensipe: this doesn't make sense

scripts/generate_docs.sc
```diff
@@ -0,0 +1,147 @@
...
+    println(s"Copying docs for $tagVersion to the docs/docs folder")
+    cp.into(buildDir / releaseBranchVersion / 'docs, rootDocsDir)
+    cp.over(buildDir / releaseBranchVersion / "docs.html", rootDocsDir / '_layouts / "docs.html")
+    println(s"Generation docs for $tagVersion")
+    write.over(rootDocsDir / s"_config.$releaseBranchVersion.yml", s"baseurl : /marathon/$releaseBranchVersion")
+    generateDocsForVersion(rootDocsDir, releaseBranchVersion)
+  }
+}
+
+def buildDocs() = {
```
kensipe: I would rather see a `@main` and make this script runnable like our other scripts.

scripts/generate_docs.sc
kensipe: for this change to run `./doc_generation_launcher.sh` 
add that preconditions for launching this script are in the bash file.

I would much prefer you fail this script if run without some pre-condition.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,147 @@
...
+/**
+  * get the major.minor version from a [v]major.minor.patch
+  *
+  * v1.4.11 -> 1.4
+  * 1.45.983 -> 1.45
+  *
+  * @param tag
+  * @return
+  */
+def releaseBranchVersion(tag: String) = tag.dropWhile(_.isLetter).takeWhile {
```
jeschkies: Hm, now that I think of it you might be able to leverage `utils.SemVer`. What do you think?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,147 @@
...
+
+  println("Cleaning previously generated docs")
+  rm! rootDocsDir / '_site
+
+  println(s"Generating root docs for $latestReleaseVersion")
+  %("bundle", "install", "--path", s"vendor/bundle")(rootDocsDir)
+  %('bundle, "exec", s"jekyll build --config _config.yml -d _site")(rootDocsDir)
+}
+// step 3: generate docs for other versions
+def generateVersionedDocs() {
```
jeschkies: Thanks for introducing the methods. Shouldn't the tmp dir be in this method somewhere?

I'm sorry. I think I get the general idea but I have a hard time following the code. So we 1. checkout each release branch and copy the docs to a temporary folder. We then generate to docs for the latest release in the main repo. 3. states that we generate the docs for the other version we've copied in step 2. However, I don't really see this happening here. Or rather I don't see how the versions are referenced. What do you think about something like this:

```scala
val versions: Seq[utils.SemVer] = Seq(...)

versions.map { version =>
   checkout(version)
   val tmpDir: Path = copyToTmpDir(...)
   (version, tmpDir)
}.map { case (version, docsPath) =>
   val docsDir: Path = generateDocs(version, docsPath)
   (version, docsDir) 
}.fold(targetDir) { (targetDir, (version, docsDir)) =>
   copy(docsDir, targetDir, version)
   targetDir
}
```

something in that manor.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+def buildDocs() = {
+  checkoutDocsToTempFolder()
+  generateTopLevelDocs()
+  generateVersionedDocs()
+}
+
+buildDocs()
```
jeschkies: Please add a main method.

scripts/generate_docs.sc
jeschkies: So, `buildDocs` is doing this. However, methods `checkoutDocsToTempFolder`, `generateTopLevelDocs` and `generateVersionedDocs` assume some side effects. It might be helpful to express these side effects by passing the directories as I suggested above.

scripts/generate_docs.sc
meln1k: Sounds good

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+
+import java.time.Instant
+import ammonite.ops._
+import ammonite.ops.ImplicitWd._
+
+/**
+  * Docs generation process:
+  *
+  * 1. copy docs/docs and docs/_layouts/docs.html part from 1.3, 1.4 and 1.5 branches to the temp dirs
+  * 2. copy the latest release (1.5_ branch doc to the temp dir
```
ichernetsky: Is there a type: `(1.5_`?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+  * 3. generate the latest release docs
+  * 4. replace docs/docs with docs/docs from 1.3, 1.4, 1.5 and generate docs for the _sites/docs_version
+  */
+
+@main
+def main(tracked_by_git: Boolean = true) = {
+  if(tracked_by_git) {
```
ichernetsky: nit: `if (`

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+    //If the script is being tracked by git, we should create and run an untracked copy of it in order
```
ichernetsky: nit: `// If`

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+    // to avoid errors during branch checkout
+    cp.over(pwd/"generate_docs.sc", pwd/"generate_docs.temp")
+    %('amm, "generate_docs.temp", "--tracked_by_git", "false")
+    rm! pwd/"generate_docs.temp"
+  } else {
+    //check that repo is clean
```
ichernetsky: nit: `// check`

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+def isReleaseTag(tag: String) = tag.matches("""v[1-9]+\.\d+\.\d+""")
+
+/**
+  * get the major.minor version from a [v]major.minor.patch
+  *
+  * v1.4.11 -> 1.4
+  * 1.45.983 -> 1.45
+  *
+  * @param tag
+  * @return
```
ichernetsky: Please either remove it or document it.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+    println(s"Docs folder for $releaseBranchVersion is copied to ${tagBuildDir / "docs"}")
+  }
+}
+
+// step 2: generate the default version (latest release tag)
+
+def generateTopLevelDocs(buildDir: Path, docsDir: Path) = {
+  val rootDocsDir = buildDir / 'docs
+
+
```
ichernetsky: nit: extra line

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+  cp.into(docsDir, buildDir)
+
+  println("Cleaning previously generated docs")
+  rm! rootDocsDir / '_site
+
+  println(s"Generating root docs for $latestReleaseVersion")
+  %("bundle", "install", "--path", s"vendor/bundle")(rootDocsDir)
+  %('bundle, "exec", s"jekyll build --config _config.yml -d _site")(rootDocsDir)
+}
+// step 3: generate docs for other versions
```
ichernetsky: There should be empty lines around this one in order to have it consistent with the previous step comments.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+  * 4. replace docs/docs with docs/docs from 1.3, 1.4, 1.5 and generate docs for the _sites/docs_version
+  */
+
+@main
+def main(tracked_by_git: Boolean = true) = {
+  if(tracked_by_git) {
+    //If the script is being tracked by git, we should create and run an untracked copy of it in order
+    // to avoid errors during branch checkout
+    cp.over(pwd/"generate_docs.sc", pwd/"generate_docs.temp")
+    %('amm, "generate_docs.temp", "--tracked_by_git", "false")
```
jeschkies: What is the reason for this script? I don't quite understand it, sorry.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+    rm! pwd/"generate_docs.temp"
+  } else {
+    //check that repo is clean
+    val possiblyChangedLines: Vector[String] = %%('git, "diff-files", "--name-only").out.lines
+    if (possiblyChangedLines.nonEmpty) {
+      val msg = s"Git repository isn't clean, aborting docs generation. Changed files:\n${possiblyChangedLines.mkString("\n")}"
+      println(msg)
+      exit()
```
jeschkies: Nit: `exit(1)`.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    val buildDir = makeTmpDir()
+    val docsDir = pwd/up/"docs"
+
+    buildDocs(buildDir, docsDir)
```
jeschkies: Nice!

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+def makeTmpDir(): Path = {
+  val timestamp = Instant.now().toString.replace(':', '-')
+  val path = root/"tmp"/s"marathon-docs-build-$timestamp"
+  mkdir! path
+  path
+}
+
+val ignoredReleaseBranchesVersions = Set("1.6")
+
+val currentGitBranch = %%('git, "rev-parse", "--abbrev-ref", "HEAD").out.lines.head
```
jeschkies: Could you move this up? IT's hard to follow the main method otherwise.

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+}
+// step 3: generate docs for other versions
+def generateVersionedDocs(buildDir: Path) {
+  val rootDocsDir = buildDir / 'docs
+  println(s"Generating docs for versions ${docsTargetVersions.map(_._2).mkString(", ")}")
+  docsTargetVersions foreach { case (releaseBranchVersion, tagVersion) =>
+    println("Cleaning docs/docs")
+    rm! rootDocsDir / 'docs
+    println(s"Copying docs for $tagVersion to the docs/docs folder")
+    cp.into(buildDir / releaseBranchVersion / 'docs, rootDocsDir)
```
jeschkies: How are `rootDocsDir` and `docsDir` from `buildDocs` related?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,157 @@
...
+  s"tags/v$version"
+}
+
+// step 1: copy docs/docs to the respective dirs
+
+def checkoutDocsToTempFolder(buildDir: Path, docsDir: Path) = {
+  docsTargetVersions foreach { case (releaseBranchVersion, tagVersion) =>
+    val tagName = branchForTag(tagVersion)
+    %git('checkout, s"$tagName")
+    val tagBuildDir = buildDir / releaseBranchVersion
```
jeschkies: How about returning this directory so you can use the value in `generateVersionedDocs` and not rely on convention?

scripts/generate_docs.sc
meln1k: so, we want to run this script when it's not being tracked by git because it breaks if we switch branches. Do do this, we create a temporary untracked copy and run it.

scripts/generate_docs.sc
meln1k: unfortunately it's not an error code but just an exit message.

scripts/generate_docs.sc
meln1k: We can do it, but I'm not sure if it will make the code cleaner.

scripts/generate_docs.sc
jeschkies: This could be `exit(msg)` then 🙂 .

scripts/generate_docs.sc
```diff
@@ -0,0 +1,154 @@
...
+def releaseBranchVersion(tag: String) = tag.dropWhile(_.isLetter).takeWhile {
+  var reachedMinorVersion = false
+  elem => elem match {
+    case c if c.isDigit => true
+    case c if c == '.' && !reachedMinorVersion =>
+      reachedMinorVersion = true
+      true
+    case _ => false
+  }
+}
```
jeschkies: This should really be `utils.SemVer`

scripts/generate_docs.sc
```diff
@@ -0,0 +1,154 @@
...
+
+def notIgnoredBranch(branchVersion: String) = !ignoredReleaseBranchesVersions.contains(branchVersion)
+
+def getTheLastTagVersion(tags: Seq[String]) = tags.last.drop(1)
```
jeschkies: What is dropped?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,154 @@
...
+def isReleaseTag(tag: String) = tag.matches("""v[1-9]+\.\d+\.\d+""")
+
+/**
+  * get the major.minor version from a [v]major.minor.patch
+  *
+  * v1.4.11 -> 1.4
+  * 1.45.983 -> 1.45
+  *
+  */
+def releaseBranchVersion(tag: String) = tag.dropWhile(_.isLetter).takeWhile {
```
jeschkies: Nit: `minorMayor`

scripts/generate_docs.sc
```diff
@@ -0,0 +1,154 @@
...
+def getTheLastTagVersion(tags: Seq[String]) = tags.last.drop(1)
+
+/**
+  * getting 3 latest release versions and corresponding latest tags
+  *
+  * example: List[(String, String)] = List(("1.3", "1.3.14"), ("1.4", "1.4.11"), ("1.5", "1.5.6"))
+  */
+val docsTargetVersions = listAllTagsInOrder
+  .filter(isReleaseTag)
+  .groupBy(releaseBranchVersion)
```
jeschkies: If the tags would return a SemVer this would be `groupBy(version => s"${version.major}.${version.minor})`

scripts/generate_docs.sc
```diff
@@ -0,0 +1,154 @@
...
+  .filterKeys(notIgnoredBranch)
+  .mapValues(getTheLastTagVersion)
```
jeschkies: This is probably `getLatestPatch`, right?

scripts/generate_docs.sc
jeschkies: IMHO this is the build artifact of this method. What is the benefit of relying on convention here when the compiler can ensure that the correct dir is used?

scripts/generate_docs.sc
```diff
@@ -0,0 +1,154 @@
...
+    println(s"Generation docs for $tagVersion")
+    write.over(rootDocsDir / s"_config.$releaseBranchVersion.yml", s"baseurl : /marathon/$releaseBranchVersion")
+    generateDocsForVersion(rootDocsDir, releaseBranchVersion)
+  }
+}
+
+def buildDocs(buildDir: Path, docsDir: Path) = {
+  checkoutDocsToTempFolder(buildDir, docsDir)
+  generateTopLevelDocs(buildDir, docsDir)
+  generateVersionedDocs(buildDir)
```
jeschkies: How about 

```scala
docsTargetVersions foreach { case (releaseBranchVersion, tagVersion) =>
    val tagBuildDir = buildDir / releaseBranchVersion
    checkoutDocsToTempFolder(tagBuildDir, docsDir, tagVersion)
    generateVersionedDocs(tagBuildDir)
}
```

Or event
```scala
docsTargetVersions.map { case (releaseBranchVersion, tagVersion) =>
    checkoutDocsToTempFolder(buildDir / releaseBranchVersion, docsDir, tagVersion)
}.foreach { tagBuildDir =>
    generateVersionedDocs(tagBuildDir)
}
```

scripts/generate_docs.sc
meln1k: `v` in the start of the version :)

scripts/generate_docs.sc
meln1k: I tried that, and the output formatting doesn't look super nice

ci/generate_docs.sc
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+  * get the major.minor version from a [v]major.minor.patch
+  *
+  * v1.4.11 -> 1.4
+  * 1.45.983 -> 1.45
+  *
+  */
+def toSemanticVersion(tag: String): SemVer = {
+  val verionOnly = tag.filterNot(c => c.isLetter || c == '-')
```
timcharper: I don't think this method does what the description says it does. Above, it indicates that it will only take the first two digits, dropping the "V" if it starts with one, and only taking the first two numbers (major, minor).

Perhaps you meant to use  `tagWithoutV.takeWhile(_ != '-') `

ci/generate_docs.sc
```diff
@@ -0,0 +1,150 @@
...
+def getLatestPatch(tags: Seq[SemVer]): SemVer = {
+  tags.last
+}
+
+/**
+  * getting 3 latest release versions and corresponding latest tags
+  *
+  * example: List[(String, String)] = List(("1.3", "1.3.14"), ("1.4", "1.4.11"), ("1.5", "1.5.6"))
+  */
+val docsTargetVersions: List[(String, SemVer)] = listAllTagsInOrder.filter(isReleaseTag).map(toSemanticVersion).groupBy(version => s"${version.major}.${version.minor}").filterKeys(notIgnoredBranch).mapValues(getLatestPatch).toList.sortBy(_._1).takeRight(3)
```
timcharper: Seems like maybe we could add a method, dropRight, to SemVer?

ci/generate_docs.sc
```diff
@@ -0,0 +1,150 @@
...
+
+val (latestReleaseBranch, latestReleaseVersion) = docsTargetVersions.last
+
+def generateDocsForVersion(docsPath: Path, version: String, outputPath: String = "_site"): Unit = {
+  %('bundle, "exec", s"jekyll build --config _config.yml,_config.$version.yml -d $outputPath/$version/")(docsPath)
```
timcharper: Use the docker image? Ain't nobody got time to install the Ruby dependencies.

ci/generate_docs.sc
timcharper: Also, instead of doing this, why don't we add a method to the SemVer companion object that can parse a version, dropping the prefix "V" if necessary? Then, we can use the `.copy `method on SemVer to remove the extension?

jeschkies: +1

It should be possible to call `SemVer(tag)`.

ci/generate_docs.sc
jeschkies: It's not easy to follow this. So we just want to know the highest patch version, right? We can always get "major.minor".

```scala
val docsTargetVersions: List[SemVer] = {
   val minors: Map[(Int, Int), List[SemVer]] = listAllTagsInOrder
      .filter(isReleaseTag)
      .map(toSemanticVersion)
      .groupBy(v => (v.major, v.minor))

    minors.values.map(_.max)
}
```

ci/generate_docs.sc
```diff
@@ -0,0 +1,150 @@
...
+  s"tags/${version.toTagString}"
+}
+
+// step 1: copy docs/docs to the respective dirs
+
+def checkoutDocsToTempFolder(buildDir: Path, docsDir: Path): Seq[(String, Path)] = {
+  docsTargetVersions map { case (releaseBranchVersion, tagVersion) =>
+    val tagName = branchForTag(tagVersion)
+    %git('checkout, s"$tagName")
+    val tagBuildDir = buildDir / releaseBranchVersion
```
jeschkies: I still think `tagBuildDir` should be injected but that's not a reason to reject this 🙂.

docs/Dockerfile
```diff
@@ -13,7 +13,6 @@ RUN apt-get update && \
 
 COPY entrypoint.sh /
 
-EXPOSE 4000
```
timcharper: Why remove this port expose? Running the docker image and using the Jekyll Web server is sometimes useful.

meln1k: Well, jekyll web view is limited to viewing only one version of the docs. I'll update the docs generation manual later.

PR #5981, opened 2018-02-02
Improve the docs for command line flags.

meichstedt:
Summary:
The two timeouts `task_launch_timeout` and `task_launch_confirm_timeout` are closely related and should be configured together. This change adds a note to each of the flags to also see the respective other.

JIRA issues: MARATHON-3603


ichernetsky:
There is no Jenkins job for this PR for some reason. Since it has only docs changes, it is safe to merge it without running a CI job.

PR #5982, opened 2018-02-05
Hybrid Cloud: Always launch a multi-zone cluster

wavesoft:
---
ℹ️ This PR will update #5949 , so review that first before merging

---

This commit contains only the changes to the scripts that configures how the cluster is launched in order to always launch a multi-zone cluster.

### DIFF
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -36,7 +36,7 @@ fi
 echo "Workspace: ${WORKSPACE}"
 echo "Using: ${INSTALLER_URL}"
 
-PLATFORM=`uname`
+PLATFORM=$(uname)
 if [ "$PLATFORM" == 'Darwin' ]; then
     brew install gettext
 else
@@ -77,8 +77,7 @@ dcos_config:
 $EXTRA_DCOS_CONFIG
 EOF
 
-./dcos-launch create
-if [ $? -ne 0 ]; then
+if [ ./dcos-launch create -ne 0 ]
   echo "Failed to launch a cluster via dcos-launch"
   exit 2
 fi
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -4,6 +4,10 @@ set -x -e -o pipefail
 
 # Two parameters are expected: CHANNEL and VARIANT where CHANNEL is the respective PR and
 # VARIANT could be one of four custer variants: open, strict, permissive and disabled
+#
+# Also, when non-open variants are selected, the following environment variables must
+# be available: DCOS_INSTALLER_TEMPLATE and DCOS_LICENSE
+#
 if [ "$#" -ne 2 ]; then
     echo "Expected 2 parameters: <channel> and <variant> e.g. launch_cluster.sh testing/pull/1739 open"
     exit 1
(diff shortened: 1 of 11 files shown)


ci/launch_cluster.sh
```diff
@@ -13,13 +17,24 @@ CHANNEL="$1"
 VARIANT="$2"
 
 if [ "$VARIANT" == "open" ]; then
-  TEMPLATE="https://s3.amazonaws.com/downloads.dcos.io/dcos/${CHANNEL}/cloudformation/multi-master.cloudformation.json"
+  INSTALLER_URL="https://downloads.dcos.io/dcos/${CHANNEL}/dcos_generate_config.sh"
```
jeschkies: I wonder if we could ge pre-baked AWS images at some point.

ci/launch_cluster.sh
```diff
@@ -33,24 +48,35 @@ wget 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' && chmod
...
-    LicenseKey: $DCOS_LICENSE
+num_masters: 3
+fault_domain_helper:
+  LocalZone:
+    num_zones: 2
+    num_public_agents: 1
+    num_private_agents: 2
+    local: true
+  RemoteZone:
+    num_zones: 2
```
jeschkies: Could you help me understand. What does `num_zones: 2` mean. Two zones or two instances for this zone? How is the remote zone named?

ci/launch_cluster.sh
wavesoft: That would be the best, but this means that we are releasing new DC/OS images on every release.

ci/launch_cluster.sh
wavesoft: Ha, I think the name got you mixed-up. Yeah, I need to change the name from `RemoteZone` to `RemoteRegion` 😄 

Effectively you have two regions. A `Local` region (that becomes the default if there are no region/zone constraints) and a `Remote` region. Now for both of these regions, you define in how many zones to divide it.

My key takeway from this is : Rename `RemoteZone` to `RemoteRegion`

wavesoft:
> Also, why is the merge target `ic/tests/hybrid-cloud` and not `master`?

I wanted to make sure the configuration works, by testing it against the other PR, before we merge it on master eventually.

ci/launch_cluster.sh
jeschkies: That'd be fine IMHO. Anyhow, that's out of scope.

ci/launch_cluster.sh
```diff
@@ -13,13 +17,24 @@ CHANNEL="$1"
...
+  # Replace {} in the template with the channel
+  INSTALLER_URL=${$DCOS_INSTALLER_TEMPLATE/{\}/$CHANNEL}
+  EXTRA_DCOS_CONFIG="  security: ${VARIANT}"$'\n'"  license_key_contents: ${DCOS_LICENSE}"
 fi
 
 echo "Workspace: ${WORKSPACE}"
-echo "Using: ${TEMPLATE}"
+echo "Using: ${INSTALLER_URL}"
 
 PLATFORM=`uname`
```
jeschkies: I know it was not your change but shellcheck is saying

```
PLATFORM=`uname`
         ^-- SC2006: Use $(..) instead of legacy `..`.
```

ci/launch_cluster.sh
```diff
@@ -33,24 +48,35 @@ wget 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' && chmod
...
+$EXTRA_DCOS_CONFIG
 EOF
 
-if [[ -z `curl $TEMPLATE | grep 'LicenseKey'` ]]; then
-    sed -i '/LicenseKey/d' 'config.yaml'
-    echo 'Removed LicenseKey parameter as it was not in template'
-fi
-
 ./dcos-launch create
 if [ $? -ne 0 ]; then
```
jeschkies: According to shellcheck we could do `if [ ./dcos-launch create -ne 0 ]; then`. I'm fine with not cleaning up though.

ci/launch_cluster.sh
wavesoft: Fixed

ci/launch_cluster.sh
wavesoft: Since we are doing something similar with the wait below, let's change it too.

PR #5983, opened 2018-02-05
Fixed a bunch of scala compile warning

zen-dog:
mostly with unused imports, parameters, and variables.


src/main/scala/mesosphere/marathon/api/akkahttp/AkkaHttpMarathonService.scala
```diff
@@ -73,7 +73,7 @@ class AkkaHttpMarathonService(
       async {
         val oldHandler = await(oldHandlerF)
         logger.info(s"Shutting down Akka HTTP service on ${config.httpPort()}")
-        val unbound = await(oldHandler.unbind())
+        val _ = await(oldHandler.unbind())
```
meln1k: Do you think we can just write `await(oldHandler.unbind())`?

zen-dog: Sadly, no. This will eat up the exceptions in the `await` block.

zen-dog:
> bummer about CI flake. LGTM!

that wasn't a flake - I actually forgot to remove a deleted option from the test suite.

PR #5986, opened 2018-02-06
Enhance App API to reflect resources for executor

janisz:
Add option to configure custom executor resources.

Fixes: [MARATHON-8063](https://jira.mesosphere.com/browse/MARATHON-8063)

### DIFF
--- a/src/main/scala/mesosphere/marathon/raml/AppConversion.scala
+++ b/src/main/scala/mesosphere/marathon/raml/AppConversion.scala
@@ -79,7 +79,8 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       versionInfo = app.versionInfo.toRaml,
       unreachableStrategy = Some(app.unreachableStrategy.toRaml),
       killSelection = app.killSelection.toRaml,
-      tty = app.tty
+      tty = app.tty,
+      executorResources = app.executorResources.toRaml
     )
   }
 
@@ -162,7 +163,8 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       secrets = Raml.fromRaml(app.secrets),
       unreachableStrategy = app.unreachableStrategy.map(_.fromRaml).getOrElse(AppDefinition.DefaultUnreachableStrategy),
       killSelection = app.killSelection.fromRaml,
-      tty = app.tty
+      tty = app.tty,
+      executorResources = app.executorResources.map(_.fromRaml)
     )
     result
   }
@@ -210,7 +212,8 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       taskKillGracePeriodSeconds = update.taskKillGracePeriodSeconds.orElse(app.taskKillGracePeriodSeconds),
       unreachableStrategy = update.unreachableStrategy.orElse(app.unreachableStrategy),
       killSelection = update.killSelection.getOrElse(app.killSelection),
-      tty = update.tty.orElse(app.tty)
+      tty = update.tty.orElse(app.tty),
+      executorResources = update.executorResources
     )
   }
 
--- a/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
+++ b/src/main/scala/mesosphere/marathon/state/AppDefinition.scala
@@ -14,7 +14,7 @@ import mesosphere.marathon.core.plugin.PluginManager
 import mesosphere.marathon.core.pod.{ HostNetwork, Network }
 import mesosphere.marathon.core.readiness.ReadinessCheck
 import mesosphere.marathon.plugin.validation.RunSpecValidator
-import mesosphere.marathon.raml.{ App, Apps, Resources }
+import mesosphere.marathon.raml.{ App, Apps, ExecutorResources, Resources }
 import mesosphere.marathon.state.Container.{ Docker, MesosAppC, MesosDocker }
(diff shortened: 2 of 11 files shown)


meichstedt:
@timcharper, @kensipe please take a look. Also see the previous discussion on https://github.com/mesosphere/marathon/pull/5675

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -180,6 +182,10 @@ case class AppDefinition(
       .setUnreachableStrategy(unreachableStrategy.toProto)
       .setKillSelection(killSelection.toProto)
 
+    executorResources.foreach(r => builder.addExecutorResources(ScalarResource(Resource.CPUS, r.cpus)))
+    executorResources.foreach(r => builder.addExecutorResources(ScalarResource(Resource.MEM, r.mem)))
+    executorResources.foreach(r => builder.addExecutorResources(ScalarResource(Resource.DISK, r.disk)))
+
```
timcharper: Any reason this isn't combined into a single for each statement?

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -249,12 +260,25 @@ case class AppDefinition(
...
+    else
+      Some(Resources(
+        cpus = executorResourcesMap.getOrElse(
+          Resource.CPUS, this.executorResources.getOrElse(AppDefinition.MinimalExecutorResources).cpus),
+        mem = executorResourcesMap.getOrElse(
+          Resource.MEM, this.executorResources.getOrElse(AppDefinition.MinimalExecutorResources).mem),
+        disk = executorResourcesMap.getOrElse(
+          Resource.DISK, this.executorResources.getOrElse(AppDefinition.MinimalExecutorResources).disk)
+      ))
+
```
kensipe: this whole `executorResourcesMap` thing...  lines 228-232 and 263-274...   There is no external reference to it that I see...  As this is all internal... this seems like a complicated way of managing these details.   The nested Somes with defaults ... 

why don't we just default the map with defaults at creation, then add what is provided by the protob?
it would simplify this to just getting the values.

kensipe: I also don't care for the shorten name `executorRes`

it is an easy replacement for `executorResourcesMap.isEmpty` to `proto.getExecutorResourcesList.isEmpty`

I don't see the value of `this.executorResources.isEmpty` as this is in `mergeFromProto` and there is no way `executorResources` could have been set... which makes it 1 logical jump away from `DefaultExecutorResources` which is `None`

kensipe: looking closer it looks like you are trying to maintain the same code as `resourcesMap`... which totally makes sense... however It seems more challenging from a readability standpoint to read this:
```
   cpus = executorResourcesMap.getOrElse(
          Resource.CPUS, executorResources.getOrElse(AppDefinition.MinimalExecutorResources).cpus),
```
and this:
```
        cpus = resourcesMap.getOrElse(Resource.CPUS, this.resources.cpus),
```

if we can have the "Else" as the default vs another "Else"... it would really improve readability IMO.

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
```diff
@@ -413,6 +437,10 @@ object AppDefinition extends GeneralPurposeCombinators {
 
   val DefaultIsResident = false
 
+  val DefaultExecutorResources: Option[Resources] = None
+
+  val MinimalExecutorResources: Resources = ExecutorResources.Default.fromRaml
```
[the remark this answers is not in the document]
janisz: I agree, this is a mess but I have no idea how to clean it up. We can move `ExecutorResources.Default` from raml to pods. I can add a comment here to describe why it's done in that way. Basically, I tried to have backward compatibility so set executor to none. On the other hand, maybe we should explicitly define executor resources.

src/main/scala/mesosphere/marathon/raml/AppConversion.scala
```diff
@@ -79,7 +79,8 @@ trait AppConversion extends DefaultConversions with ConstraintConversion with En
       versionInfo = app.versionInfo.toRaml,
       unreachableStrategy = Some(app.unreachableStrategy.toRaml),
       killSelection = app.killSelection.toRaml,
-      tty = app.tty
+      tty = app.tty,
+      executorResources = app.executorResources.toRaml
```
jdef: I think you need a similar update for `appProtoRamlWriter` (used for migrations)

src/main/scala/mesosphere/marathon/state/AppDefinition.scala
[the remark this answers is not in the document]
jdef: Application of "defaults" should probably be applied in `AppNormalization` somewhere. However in this case it looks like you're also using "MinimalExecutorResources" in protobuf deserialization to .. compensate for lack of a migration strategy.

I think if you patch up AppNormalization to deal with defaults, as well as make the suggested AppConversion change, AND write a migration that forces all service definitions through app normalization (like we did to convert everything in 1.5 for the networking API) then you'd have something a little cleaner? (and could remove MinimalExecutorResources here)

PR #5976, opened 2018-02-02
Improve documentation for GROUP_BY operator with respect to the value parameter.

meichstedt:
Improve documentation for GROUP_BY operator with respect to the value parameter.

JIRA issues: MARATHON-1782

PR #5989, opened 2018-02-07
Improve documentation for GROUP_BY operator with respect to the value parameter. (#5976)

meichstedt:
Improve documentation for GROUP_BY operator with respect to the value parameter. (#5976)

JIRA issues: MARATHON-1782


ichernetsky:
This was merged into master already.

PR #5996, opened 2018-02-08
Copy tarballs from builds to releases

jeschkies:
Summary:
Instead of rebuilding releases we want to copy master builds to the release folder.
This is part of the forward port of a673f2e.

Depends on #5970.

JIRA issues: MARATHON_EE-1840


ci/pipeline
```diff
@@ -122,7 +128,8 @@ def asPullRequest(run: => (utils.SemVer, Option[awsClient.Artifact])): Unit = {
  * @param artifactUrl Url for the saved Marathon artifact.
  * @param sha1 The sah1 checksum of the artifact.
  */
-def updateDcosImage(version: utils.SemVer, artifactUrl: String, sha1: String): Unit = utils.stage("Patch DC/OS Branch") {
+@main
```
zen-dog: nit: since it's a main target shouldn't we move it below the `//////////////////// MAIN TARGETS ////////////////////`?

ci/pipeline
```diff
@@ -284,38 +251,26 @@ def jenkins(): Unit = {
...
 
-  println(s"Releasing version: $releaseVersion")
-  %('git, "checkout", gitSha)
-
-  val version = build()
-  val tagVersion = version.toTagString
-  %('git, "tag", "-a", tagVersion, "-m", tagVersion)
+  val config = releases.parser.parse(args, releases.Config()).getOrElse {
+    sys.exit(1)
+    ???
```
zen-dog: please add a `println(s"Couldn't parse....")`

jeschkies: This should never be reached. Since Scopt will show this error

```
Error: Argument targets... failed when given 'foo'. foo is not a known release target; valid options are:
* docker-tag
* docker-latest
* s3-package
* linux-packages
* jar-artifact
* plugin-interface
Error: Argument targets... failed when given 'bar'. bar is not a known release target; valid options are:
* docker-tag
* docker-latest
* s3-package
* linux-packages
* jar-artifact
* plugin-interface
Error: Missing option --version
Error: Missing option --run-tests
pipeline release
Usage: scopt [options] [targets...]

  --version <value>    build that should be released, e.g. 1.7.42-deadbeef
  --run-tests <value>  whether to run tests for build
  targets...           targets to release: docker-tag, docker-latest, s3-package, linux-packages, jar-artifact, plugin-interface
Could not parse arguments 'foo, bar'.
```

PR #5997, opened 2018-02-08
Stabilize scheduler test

jeschkies:
Summary:
`MarathonSchedulerActorTest` has been flaky lately. This patch tries to stabilize it.
The loop does not show this issue anymore.


jeschkies:
Somehow, an integration test starts to fail. We should wait for the stats on Monday.

jeschkies:
Drop as last 100 runs on master showed only 3 failures.

PR #5998, opened 2018-02-09
Do not set unconditionally DiskInfo.metadata when dealing with volumes

ichernetsky:
Summary:
Marathon set metadata to an empty map even if metadata was not present
in resource offers at all.

JIRA issues: MARATHON-8073

### DIFF
--- a/src/main/scala/mesosphere/marathon/state/Volume.scala
+++ b/src/main/scala/mesosphere/marathon/state/Volume.scala
@@ -108,7 +108,7 @@ object HostVolume {
 
 case class DiskSource(
     diskType: DiskType, path: Option[String],
-    id: Option[String], metadata: Map[String, String], profileName: Option[String]) {
+    id: Option[String], metadata: Option[Map[String, String]], profileName: Option[String]) {
   if (diskType == DiskType.Root)
     require(path.isEmpty, "Path is not allowed for diskType")
   else
@@ -136,7 +136,7 @@ case class DiskSource(
       else
         bld.setPath(Source.Path.newBuilder().setRoot(p))
       id.foreach(bld.setId)
-      bld.setMetadata(metadata.toMesosLabels)
+      metadata.foreach(metadata => bld.setMetadata(metadata.toMesosLabels))
       profileName.foreach(bld.setProfile)
       Some(bld.build)
     case (_, _) =>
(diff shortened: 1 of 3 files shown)


src/test/scala/mesosphere/marathon/state/VolumeTest.scala
```diff
@@ -148,17 +148,17 @@ class VolumeTest extends UnitTest {
     }
 
     "validating that DiskSource asMesos converts to an Option Mesos Protobuffer" in {
-      DiskSource(DiskType.Root, None, None, Map.empty, None).asMesos shouldBe None
-      val Some(pathDisk) = DiskSource(DiskType.Path, Some("/path/to/folder"), None, Map.empty, None).asMesos
+      DiskSource(DiskType.Root, None, None, None, None).asMesos shouldBe None
+      val Some(pathDisk) = DiskSource(DiskType.Path, Some("/path/to/folder"), None, None, None).asMesos
```
jdef: for these tests that use `None` for metadata .. shouldn't they verify that source.hasMetadata returns false?

src/main/scala/mesosphere/marathon/state/Volume.scala
```diff
@@ -136,7 +136,7 @@ case class DiskSource(
       else
         bld.setPath(Source.Path.newBuilder().setRoot(p))
       id.foreach(bld.setId)
-      bld.setMetadata(metadata.toMesosLabels)
+      metadata.foreach(metadata => bld.setMetadata(metadata.toMesosLabels))
```
meichstedt: This doesn't safely improve the situation IMO. If metadata is passed as `Some(Map.empty)`, it will still lead to the incompatible behavior. That might not be the case now, but it could be introduced as a regression. Using `metadata: Map[String, String]` and checking for `.nonEmpty` would prevent that.

Otherwise we should prevent that `DiskSource` ever provides a `Some(Map.empty)`. From what I can see, `DiskSource` doesn't need to be a case class. Private constructor and an `apply` that makes sure the metadata is populated correctly?

ichernetsky: @meichstedt, the thing is that if `metadata` is empty in a resource offer, it should be empty in an offer operation; if `metadata` is absent in a resource offer, it should be empty in an offer operation too. That's why its type has been changed to `Option[Map[String, String]]`. Does it address your concern? I am not sure if we refer to the same thing here.

src/test/scala/mesosphere/marathon/state/VolumeTest.scala
ichernetsky: Done.

src/main/scala/mesosphere/marathon/state/Volume.scala
jdef: @meichstedt I don't follow. Marathon should reliably reproduce what Mesos encodes in the originally offered resource.disk.source.metadata. `metadata` is of type `Labels`, which is a `message`, not `repeated` or `map`.

1. no `metadata` protobuf <--> `None`
2. an empty `metadata` protobuf <--> `Some(Map.empty)`
3. non-empty `metadata` protobuf <--> `Some(Map[String,String] { ... entries ... })`

meichstedt: @jdef I agree – when reviewing, I was under the impression that there's a chance Marathon mutates those values. It doesn't so that's fine. @ichernetsky created a JIRA a while back related to this ser/deser being a problem in general, since we potentially lack information afterwards.

PR #5999, opened 2018-02-09
Publish Docker tags, native packages and jar.

jeschkies:
Summary:
This patch enables the new release pipeline to publish tags of the
Docker image, native Linux packages and the JAR file.

Depends on #5996.

JIRA issues: MARATHON_EE-1842, MARATHON_EE-1846, MARATHON_EE-1844

### DIFF
--- a/ci/pipeline
+++ b/ci/pipeline
@@ -263,15 +263,29 @@ def release(args: String*): Unit = {
 
   %('git, "checkout", config.version.commit)
 
-  buildDockerAndLinuxPackages()
-  if (config.runTests)
-    testDockerAndLinuxPackages()
+  val targets = config.targets.toSet
 
-  config.targets.distinct.foreach {
+  if (targets.contains(ReleaseTarget.LinuxPackages) ||
+      targets.contains(ReleaseTarget.DockerTag) ||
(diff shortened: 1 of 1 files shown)


ci/pipeline
```diff
@@ -263,14 +263,28 @@ def release(args: String*): Unit = {
 
   %('git, "checkout", config.version.commit)
 
-  buildDockerAndLinuxPackages()
-  if (config.runTests)
-    testDockerAndLinuxPackages()
+  val targets = config.targets.toSet
+
+  if (targets.contains(ReleaseTarget.LinuxPackages) ||
```
timcharper: I think this will break the dry run feature? The intention with the dry run feature is that we would have a way to run the compile and test phases without producing any external side effects.

May be, instead, we could add a dependencies field to each of the release targets, and add a release target of "Compile", "Test", and, "PackageTest", with respective dependencies.

Then, below, instead of "targets.foreach", we could invoke something like:

```
targets.toSeq.flatMap { t => t.recursiveDependencies :+ t }.distinct.foreach { t =>
  ...
}
```

(We could not use a set as those are unordered)

This would cause the graph execution order to be worked out, naturally, without too much code.

jeschkies: Good point.

PR #6009, opened 2018-02-12
Update docs with valid examples for fetch URIs

meln1k:
Summary: Updated the docs to use the fetch field instead of uris, and provided comprehensive examples.

JIRA issues: MARATHON-8078

### DIFF
--- a/docs/docs/application-basics.md
+++ b/docs/docs/application-basics.md
@@ -36,7 +36,7 @@ When you define and launch an application, Marathon hands over execution to Meso
 
 To run any non-trivial application, you typically depend on a collection of resources: files and/or archives of files. 
 To manage resource allocation, Marathon uses the [Mesos fetcher](http://mesos.apache.org/documentation/latest/fetcher/) 
-do the legwork in terms of downloading (and potentially) extracting resources.
+to do the legwork in terms of downloading (and potentially) extracting resources.
 
 Before we dive into this topic, let's have a look at an example:
 
@@ -58,10 +58,10 @@ Before we dive into this topic, let's have a look at an example:
 
 What the example above does: before executing the `cmd`, download the resource `https://example.com/app/cool-script.sh` 
 (via Mesos) and make it available in the application task's sandbox. You can check that these have been downloaded by 
-visiting the Mesos UI and clicking into a Mesos worker node's sandbox where you should now find `cool-script.sh`.
+visiting the Mesos UI and clicking into a Mesos agent node's sandbox where you should now find `cool-script.sh`.
 
 As of Mesos v0.22 and above, the fetcher code does not make downloaded files executable by default.
- In the example above, we should provide the field `"executable": true` to tell Mesos change the fetch result to 
+ In the example above, we provided the field `"executable": true` to tell Mesos change the fetch result to 
  executable for every user. If the `"executable"` field is `"true"`, the `"extract"` field is ignored and has no effect.
 
 As already mentioned above, Marathon also [knows how to handle](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/mesos/TaskBuilder.scala) 
@@ -79,8 +79,7 @@ first attempt to unpack/extract resources with the following file extensions:
 To prevent this behaviour, field `"extract": false` must be provided.
 
 And how this looks in practice shows you the following example: let's assume you have an application executable in a zip
- file at `https://example.com/app.zip`. This zip file contains the script `cool-script.sh` and that's what you 
- want to execute. Here's how:
+ file at `https://example.com/app.zip`. This zip file contains the script `cool-script.sh` and that's what you want to execute. Here's how:
 
 ```json
 {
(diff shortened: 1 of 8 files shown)


docs/docs/application-basics.md
```diff
@@ -34,28 +34,39 @@ When you define and launch an application, Marathon hands over execution to Meso
 
 ## Using Resources in Applications
 
-To run any non-trivial application, you typically depend on a collection of resources: files and/or archives of files. To manage resource allocation, Marathon has the concept of URIs. URIs use the Mesos fetcher to do the legwork in terms of downloading (and potentially) extracting resources.
+To run any non-trivial application, you typically depend on a collection of resources: files and/or archives of files. 
+To manage resource allocation, Marathon uses the [Mesos fetcher](http://mesos.apache.org/documentation/latest/fetcher/) 
+do the legwork in terms of downloading (and potentially) extracting resources.
```
ichernetsky: nit: `to do the legwork`.

docs/docs/application-basics.md
```diff
@@ -74,21 +89,78 @@ And how this looks in practice shows you the following example: let's assume you
...
+    "fetch": [
+        {
+          "uri": "https://example.com/app.zip",
+          "cache": true
+        }
+    ]
+}
+```
+
+It's also possible to specify path and name of the destination where fetched resource will be stored:
```
ichernetsky: nit: `the fetched resource`

docs/docs/application-basics.md
```diff
@@ -74,21 +89,78 @@ And how this looks in practice shows you the following example: let's assume you
...
+    "fetch": [
+        {
+          "uri": "https://example.com/some-script.sh",
+          "destPath": "cool-script.sh"
+        }
+    ]
+}
+```
+
+Note also that you can specify many resources, not only one. So, for example, you could provide a git repository
```
ichernetsky: nit: `not just one`.

docs/docs/application-basics.md
```diff
@@ -34,28 +34,39 @@ When you define and launch an application, Marathon hands over execution to Meso
...
+          "executable": true
+        }
     ]
 }
 ```
 
-What the example above does: before executing the `cmd`, download the resource `https://example.com/app/cool-script.sh` (via Mesos) and make it available in the application task's sandbox. You can check that these have been downloaded by visiting the Mesos UI and clicking into a Mesos worker node's sandbox where you should now find `cool-script.sh`.
+What the example above does: before executing the `cmd`, download the resource `https://example.com/app/cool-script.sh` 
+(via Mesos) and make it available in the application task's sandbox. You can check that these have been downloaded by 
+visiting the Mesos UI and clicking into a Mesos worker node's sandbox where you should now find `cool-script.sh`.
```
alenkacz: I think the worker is oficially called agent?

docs/docs/application-basics.md
```diff
@@ -34,28 +34,39 @@ When you define and launch an application, Marathon hands over execution to Meso
...
 
-**Note:** As of Mesos v0.22 and above, the fetcher code does not make downloaded files executable by default. In the example above, `cmd` first makes the file executable.
+As of Mesos v0.22 and above, the fetcher code does not make downloaded files executable by default.
+ In the example above, we should provide the field `"executable": true` to tell Mesos change the fetch result to
```
alenkacz: we provided maybe... ?

docs/docs/application-basics.md
```diff
@@ -65,7 +76,11 @@ As already mentioned above, Marathon also [knows how to handle](https://github.c
 * `.tar.xz`
 * `.zip`
 
-And how this looks in practice shows you the following example: let's assume you have an application executable in a zip file at `https://example.com/app.zip`. This zip file contains the script `cool-script.sh` and that's what you want to execute. Here's how:
+To prevent this behaviour, field `"extract": false` must be provided.
+
+And how this looks in practice shows you the following example: let's assume you have an application executable in a zip
+ file at `https://example.com/app.zip`. This zip file contains the script `cool-script.sh` and that's what you 
+ want to execute. Here's how:
```
alenkacz: the newline here is intentional?

docs/docs/application-basics.md
```diff
@@ -74,21 +89,78 @@ And how this looks in practice shows you the following example: let's assume you
...
 }
 ```
 
 Note that in contrast to the example `basic-1` we now have a `cmd` that looks as follows: `app/cool-script.sh`. This stems from the fact that when the zip file gets downloaded and extracted, a directory `app` according to the file name `app.zip` is created where the content of the zip file is extracted into.
 
-Note also that you can specify many resources, not only one. So, for example, you could provide a git repository and some resources from a CDN as follows:
+Additionally, if you want to use a fetcher cache, `"cache": true` field must be specified. If a URI is encountered 
+for the first time (for the same user), it is first downloaded into the cache, then copied to the sandbox directory 
+from there. If the same URI is encountered again, and a corresponding cache file is resident in the cache or 
+still en route into the cache, then downloading is omitted and the fetcher proceeds directly to copying from the cache.
```
zen-dog: I think it's worth mentioning that the cache is "per agent" and so effectively if the task is restarted on the new agent each time, the file will be downloaded each time.

PR #6010, opened 2018-02-13
Add support for Graphite reporter interval parameter

timcharper:
Also, fix documentation for graphite reporter: tcp is not allowed. Log a warning if a protocol other than udp is specified.

JIRA issues: MARATHON-8080, MARATHON-8079

### DIFF
--- a/src/main/scala/mesosphere/marathon/Main.scala
+++ b/src/main/scala/mesosphere/marathon/Main.scala
@@ -3,6 +3,7 @@ package mesosphere.marathon
 import java.net.URI
 
 import akka.actor.ActorSystem
+import akka.http.scaladsl.model.Uri
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.{ Guice, Module }
 import com.typesafe.config.{ Config, ConfigFactory }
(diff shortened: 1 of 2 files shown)


timcharper:
I tested this change, manually.

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -61,11 +62,17 @@ class MarathonApp(args: Seq[String]) extends AutoCloseable with StrictLogging {
...
+          logger.warn(s"Graphite reporter protocol ${url.scheme} is not supported; using UDP")
+        }
+        val params = url.query()
         s"""
            |kamon.statsd {
-           |  hostname: ${url.getHost}
-           |  port: ${if (url.getPort == -1) 8125 else url.getPort}
+           |  hostname: ${url.authority.host}
+           |  port: ${if (url.authority.port == 0) 8125 else url.authority.port}
+           |  flush-interval: ${params.collectFirst { case ("interval", iv) => iv.toInt }.getOrElse(10)} seconds
```
alenkacz: is it documented somewhere that 10 is a default interval?

timcharper: This default is specified in the upstream Kamon StatsD dependency

timcharper: (And our documentation already specifies that the default is 10)

PR #6011, opened 2018-02-13
Add deployment failure reason to event.

jeschkies:
Summary:
Currently we do not send the reason for a deployment failure along with
the event. This makes debugging hard. This change introduces an optional
reason message.

### DIFF
--- a/docs/docs/rest-api/public/api/v2/events.raml
+++ b/docs/docs/rest-api/public/api/v2/events.raml
@@ -16,6 +16,11 @@ get:
         description:
             Specify subscribed event types.
             You can specify this parameter multiple times with different values.
+    plan-format:
+        required: false
+        description:
+            Indicate whether to receive "light" events or full events including
+            deployment plans. Defaults to full events.
 
   responses:
     200:
--- a/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
@@ -293,12 +293,19 @@ trait EventFormats {
     )
   }
   implicit lazy val LightDeploymentFailedWrites: Writes[DeploymentFailed] = Writes { event =>
-    Json.obj(
+    val serializedEvent = Json.obj(
       "id" -> event.id,
       "plan" -> LightDeploymentPlanWrites.writes(event.plan),
       "eventType" -> "deployment_failed",
       "timestamp" -> Timestamp.now().toString
     )
+
+    event.reason match {
+      case Some(reason) =>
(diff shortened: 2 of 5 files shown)


docs/docs/rest-api/public/api/v2/events.raml
```diff
@@ -16,6 +16,11 @@ get:
         description:
             Specify subscribed event types.
             You can specify this parameter multiple times with different values.
+    plan-format:
```
wavesoft: Is this part of this PR? Could it be that you meant to include the `reason` field?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -293,12 +293,19 @@ trait EventFormats {
...
+    val serializedEvent = Json.obj(
       "id" -> event.id,
       "plan" -> LightDeploymentPlanWrites.writes(event.plan),
       "eventType" -> "deployment_failed",
       "timestamp" -> Timestamp.now().toString
     )
+
+    event.reason match {
+      case Some(reason) =>
+        serializedEvent ++ Json.obj("reason" -> reason)
```
wavesoft: Please document the `reason` field in the RAML docs

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentManagerActor.scala
```diff
@@ -147,7 +147,7 @@ class DeploymentManagerActor(
       sender() ! cancelDeployment(plan.id)
 
     case DeploymentFinished(plan, result) =>
-      runningDeployments.remove(plan.id).map { deploymentInfo =>
+      runningDeployments.remove(plan.id).foreach { deploymentInfo =>
```
wavesoft: ✨

docs/docs/rest-api/public/api/v2/events.raml
jeschkies: This is a clean up. Boy Scout rule. We never followed up.

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
jeschkies: I don't think we have the event types in RAML anywhere. Do we?

jeschkies:
I created a DC/OS build with this build of Marathon https://github.com/dcos/dcos/pull/2457.

docs/docs/rest-api/public/api/v2/events.raml
wavesoft: It could be a separate PR, but I am totally fine with it being here 👍

PR #6013, opened 2018-02-13
Download diagnostic bundle after failed Si test.

jeschkies:
Summary:
This change not only enables the download of diagnostic bundles agian
but also makes `ci/launch_cluster.sh` runnable from macOS.


ci/launch_cluster.sh
```diff
@@ -2,34 +2,56 @@
...
-echo "Workspace: ${WORKSPACE}"
 echo "Using: ${TEMPLATE}"
 
-PLATFORM=`uname`
-if [ "$PLATFORM" == 'Darwin' ]; then
-    brew install gettext
-else
-    apt-get update && apt-get install -y -t jessie-backports gettext-base wget
+# Ensure envsubst is available.
+if ! command -v envsubst >/dev/null 2>&1; then
```
wavesoft: I would vote against on installing missing dependencies in this file. I am in favour of checking the environment and failing early, like the `jq` requirement above.

For this purpose you could introduce a `install_deps.sh` file that the CI will run first to check the env **and** install missing deps.

ci/launch_cluster.sh
```diff
@@ -2,34 +2,56 @@
...
+    PLATFORM=$(uname)
+    if [ "$PLATFORM" == 'Darwin' ]; then
+        if [ ! -f /usr/local/opt/gettext/bin/envsubst ]; then
+            echo "Installing gettext to get envsubst."
+            brew install gettext
```
wavesoft: Yeah, don't do this without asking the user first 😄

ci/launch_cluster.sh
```diff
@@ -2,34 +2,56 @@
...
+        apt-get update && apt-get install -y -t jessie-backports gettext-base wget
+    fi
+fi
+
+# Ensure dcos-launch is available.
+mkdir -p "$(pwd)/bin"
+PATH=$PATH:"$(pwd)/bin"
+if ! command -v dcos-launch >/dev/null 2>&1; then
+    echo "dcos-launch was not found. Downloading '$(pwd)/bin' ..."
+    wget 'https://downloads.dcos.io/dcos-test-utils/bin/linux/dcos-launch' \
```
wavesoft: What about the OSX binary? Again, this could go on an `install_deps.sh` file.

ci/si_pipeline.sh
```diff
@@ -70,10 +69,10 @@ case $CLUSTER_LAUNCH_CODE in
...
       (cd tests && make init test)
       SI_CODE=$?
-      # if [ ${SI_CODE} -gt 0 ]; then
-      #  download-diagnostics-bundle
-      # fi
-      ./dcos-launch delete
+      if [ ${SI_CODE} -gt 0 ]; then
+        download-diagnostics-bundle
+      fi
+      ./bin/dcos-launch delete
```
wavesoft: What if `dcos-launch` is already available in your system? 

I would recommend here to just modify the PATH and append the `./bin` subfolder.

jeschkies:
Open fails with know SI test issue.

ci/launch_cluster.sh
jeschkies: Yeah. That's kind of wild.

ci/si_pipeline.sh
jeschkies: So far this script is only run by the CI and it assumes a specific installation path for dcos-launch.

ci/launch_cluster.sh
jeschkies: Done.

PR #6022, opened 2018-02-15
Use variant when deciding about license key

alenkacz:
JIRA issues: MARATHON-8086


alenkacz:
The cluster started so I am going to merge this...

PR #6025, opened 2018-02-16
Do not fail script after error.

jeschkies:
Summary:
Sourcing `si_install_deps.sh` set `-e` which means the pipeline script
would abort on an error. However, we have to delete the cluster first
and then exit.

### DIFF
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -71,4 +71,6 @@ fi
 jq -r .ssh_private_key cluster_info.json > "$CLI_TEST_SSH_KEY"
 
 # Return dcos_url
-echo "$(dcos-launch describe | jq -r ".masters[0].public_ip")"
+CLUSTER_IP="$(dcos-launch describe | jq -r ".masters[0].public_ip")"
+echo "Launched cluster with IP $CLUSTER_IP"
+echo "$CLUSTER_IP"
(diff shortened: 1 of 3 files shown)


ci/launch_cluster.sh
```diff
@@ -71,4 +71,6 @@ fi
 jq -r .ssh_private_key cluster_info.json > "$CLI_TEST_SSH_KEY"
 
 # Return dcos_url
-echo "$(dcos-launch describe | jq -r ".masters[0].public_ip")"
+CLUSTER_IP="$(dcos-launch describe | jq -r ".masters[0].public_ip")"
+echo "Launched cluster with IP $CLUSTER_IP"
```
alenkacz: 👍

ci/si_pipeline.sh
```diff
@@ -75,10 +74,16 @@ case $CLUSTER_LAUNCH_CODE in
...
       fi
-      dcos-launch delete
+      dcos-launch delete || true
       exit "$SI_CODE" # Propagate return code.
       ;;
   2) exit-as-unstable "Cluster launch failed.";;
-  3) exit-as-unstable "Cluster did not start in time.";;
-  *) exit-as-unstable "Unknown error in cluster launch: $CLUSTER_LAUNCH_CODE";;
+  3)
+      dcos-launch delete || true
```
alenkacz: I dunno whether we want to keep this or not - I guess I don't have strong opinion, although I would probably remove these...

jeschkies:
Cleanup succeeded despite the failing tests.

ci/si_pipeline.sh
jeschkies: I've removed the `|| true`. We do have to keep the delete, though, since a rollback does not cleanup everything.

PR #6026, opened 2018-02-16
Run system integration tests at specific time.

jeschkies:
Summary:
We start the SI clusters a specific times so that they won't overlap
with each other and with clusters for Metronome tests.

JIRA issues: MARATHON-8094


timcharper:
No JIRA issue for this?

Where do we go to determine whether or not the schedules are overlapping?

jeschkies:
> No JIRA issue for this?
> Where do we go to determine whether or not the schedules are overlapping?

I've create a ticket that tracks the times. A little overlapping is fine.

PR #6028, opened 2018-02-16
Retry on DCOSException.

jeschkies:
Summary:
Fixes `TypeError: unsupported operand type(s) for |=: 'bool' and 'DCOSException'`.

### DIFF
--- a/tests/system/test_marathon_universe.py
+++ b/tests/system/test_marathon_universe.py
@@ -36,7 +36,8 @@ def test_install_marathon():
     """
 
     # Install
-    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=DCOSException)
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30,
+                    retry_on_exception=lambda e: isinstance(e, DCOSException))
     def install_marathon():
         shakedown.install_package_and_wait(PACKAGE_NAME)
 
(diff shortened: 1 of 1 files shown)


tests/system/test_marathon_universe.py
```diff
@@ -36,15 +36,17 @@ def test_install_marathon():
     """
 
     # Install
-    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=DCOSException)
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30,
+                    retry_on_exception=lambda e: isinstance(e, DCOSException))
```
ichernetsky: Should it be just `common.ignore_exception`?

tests/system/test_marathon_universe.py
```diff
@@ -36,15 +36,17 @@ def test_install_marathon():
...
     def install_marathon():
         shakedown.install_package_and_wait(PACKAGE_NAME)
 
     install_marathon()
     assert shakedown.package_installed(PACKAGE_NAME), 'Package failed to install'
 
     # 5000ms = 5 seconds, 5 seconds * 60 attempts = 300 seconds = WAIT_TIME_IN_SECS
-    @retrying.retry(wait_fixed=5000, stop_max_attempt_number=60, retry_on_exception=DCOSException)
+    @retrying.retry(wait_fixed=5000, stop_max_attempt_number=60,
+                    retry_on_exception=lambda e: isinstance(e, DCOSException))
```
ichernetsky: The same here.

tests/system/test_marathon_universe.py
jeschkies: I am actually not sure, @kensipe?

kensipe: the `common.ignore_exception` will ignore all exceptions... the way this is written it only retry on a `DCOSException`... I can see cases where that is what is desired.    Is there a reason to be specific here?   should we modify our common code to allow for this case?

jeschkies: Lol, originally I thought you introduced the retry on `DCOSException` here but git blame says it was me 🙈.

PR #6029, opened 2018-02-19
Document how to run Marathon Docker image on Mac.

jeschkies:
Summary:
This change explains how to run the Marathon Docker image on Mac with a local
running ZooKeeper and Mesos master. No ZooKeeper or Mesos Docker
container is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,7 +201,7 @@ Or run it without zookeeper container:
 When using Docker on Mac you have to use a [special address](https://docs.docker.com/docker-for-mac/networking/#there-is-no-docker0-bridge-on-macos) instead of `localhost`:
 
     docker run -p 8080:8080 --hostname $(hostname) -P mesosphere/marathon:v1.6.335 \
-        --master docker.for.mac.host.internal:505 \
+        --master docker.for.mac.host.internal:5050 \
         --zk zk://docker.for.mac.host.internal:2181/marathon
 
 If you want to inspect the contents of the Docker container:
--- a/README.md
+++ b/README.md
@@ -200,8 +200,8 @@ Or run it without zookeeper container:
 
 When using Docker on Mac you have to use a [special address](https://docs.docker.com/docker-for-mac/networking/#there-is-no-docker0-bridge-on-macos) instead of `localhost`:
 
-    docker run --network host --hostname $(hostname) mesosphere/marathon:v1.6.335 \
-        --master zk://docker.for.mac.host.internal:2181/mesos \
+    docker run -p 8080:8080 --hostname $(hostname) -P mesosphere/marathon:v1.6.335 \
+        --master docker.for.mac.host.internal:505 \
         --zk zk://docker.for.mac.host.internal:2181/marathon
 
 If you want to inspect the contents of the Docker container:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -198,6 +198,12 @@ Or run it without zookeeper container:
 
     docker run marathon:{version} --master local --zk zk://localhost:2181/marathon
 
+On Mac you have to use a special address instead of `localhost`:
```
alenkacz: Maybe "when using docker for Mac" would be more specific?

alenkacz: and we might link https://docs.docker.com/docker-for-mac/networking/#there-is-no-docker0-bridge-on-macos

README.md
```diff
@@ -198,6 +198,12 @@ Or run it without zookeeper container:
...
+
+    docker run --network host --hostname $(hostname) mesosphere/marathon:v1.6.335 \
+        --master zk://docker.for.mac.host.internal:2181/mesos \
+        --zk zk://docker.for.mac.host.internal:2181/marathon
```
kensipe: why not?   ` docker run --network host --hostname $(hostname) mesosphere/marathon:v1.6.335 --master $(hostname):5050 --zk zk://$(hostname):2181/marathon`
or
` docker run --network host --hostname $(hostname) mesosphere/marathon:v1.6.335 --master zk://$(hostname):2181/mesos --zk zk://$(hostname):2181/marathon`

kensipe: for running with local zk and mesos-local... this configuration has the following issues:
1. it registers with the master but goes inactive immediately
2. I can't hit marathon port 8080

Has this been tested?  or you confident this works? or is there missing information?

jeschkies: Yes, I did test it. `--master $(hostname):5050` did not work on my machine. I think this has to do with the VM the Docker image runs in.

kensipe:
sort of... deployments aren't deploying...  needs more work.. I'm sure it is the network connection of not being local... but some combination of this seems necessary.

jeschkies:
I adapted the changes.

README.md
```diff
@@ -198,6 +198,12 @@ Or run it without zookeeper container:
 
     docker run marathon:{version} --master local --zk zk://localhost:2181/marathon
 
+When using Docker on Mac you have to use a [special address](https://docs.docker.com/docker-for-mac/networking/#there-is-no-docker0-bridge-on-macos) instead of `localhost`:
+
+    docker run -p 8080:8080 --hostname $(hostname) -P mesosphere/marathon:v1.6.335 \
+        --master docker.for.mac.host.internal:505 \
```
kensipe: this is port `:5050`

however I don't think this completely works as mentioned in the previous comment.   Did you get it to work?

kensipe:
'docker run -p 8080:8080  mesosphere/marathon:v1.6.335 --master docker.for.mac.host.internal:5050 --zk zk://docker.for.mac.host.internal:2181/marathon`
seems to register marathon in mesos... but it doesn't seem to get any offers.   It also seems to always register with a new framework id... so it is unclear if it is reading zk :(

ichernetsky:
@kensipe, could you please re-review it?

ichernetsky:
Oops, never mind, I had the tab open since yesterday, hence missed that you did it already yesterday.

jeschkies:
This is now exactly as @kensipe proposed. Could you please review? Feel free to edit it.

jeschkies:
Dropping as won't do since it does not have a big priority. Devs can still use an all Docker approach if they really want to.

PR #6033, opened 2018-02-20
Fix flaky tests for apps and pods with persistent volumes

ichernetsky:
They were flaky because every once in a while `kill` managed to
finish before an app/pod task(s) got to start, and hence not killing
anything, and therefore there was no need to launch a replacement
task.

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-8045

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -790,3 +790,26 @@ async def find_event(event_type, event_stream):
 
 async def assert_event(event_type, event_stream, within=10):
     await asyncio.wait_for(find_event(event_type, event_stream), within)
+
+
+def kill_process_on_host(hostname, pattern):
+    """ Kill the process matching pattern at ip
+
+        :param hostname: the hostname or ip address of the host on which the process will be killed
+        :param pattern: a regular expression matching the name of the process to kill
+        :return: True if at least one process got killed or terminated on its own, and False otherwise
+    """
+
+    status, stdout = shakedown.run_command_on_agent(hostname, "ps aux | grep -v grep | grep '{}'".format(pattern))
+    pids = [p.strip().split()[1] for p in stdout.splitlines()]
+    killed_some = False
+
+    for pid in pids:
+        status, stdout = shakedown.run_command_on_agent(hostname, "sudo kill -9 {}".format(pid))
+        if status:
+            killed_some = True
+            print("Killed pid: {}".format(pid))
+        else:
+            print("Unable to killed pid: {}".format(pid))
+
+    return killed_some
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -797,7 +797,7 @@ def kill_process_on_host(hostname, pattern):
 
         :param hostname: the hostname or ip address of the host on which the process will be killed
         :param pattern: a regular expression matching the name of the process to kill
-        :return: True if at least one process gets killed, or False otherwise
+        :return: IDs of processes that got either killed or terminated on their own
     """
 
     cmd = "ps aux | grep -v grep | grep '{}' | awk '{{ print $2 }}' | tee >(xargs sudo kill -9)".format(pattern)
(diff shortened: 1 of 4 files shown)


tests/system/common.py
```diff
@@ -790,3 +790,26 @@ def task_by_name(tasks, name):
...
+        :param pattern: a regular expression matching the name of the process to kill
+        :return: True if at least one process gets killed, or False otherwise
+    """
+
+    status, stdout = shakedown.run_command_on_agent(hostname, "ps aux | grep -v grep | grep '{}'".format(pattern))
+    pids = [p.strip().split()[1] for p in stdout.splitlines()]
+    killed_some = False
+
+    for pid in pids:
+        status, stdout = shakedown.run_command_on_agent(hostname, "sudo kill -9 {}".format(pid))
```
jeschkies: Ideally the `ps` and the kills should run within one command to reduce the number of SSH requests.

tests/system/marathon_pods_tests.py
```diff
@@ -545,7 +545,12 @@ def wait_for_status_network_info():
     task_id1 = tasks[0]['id']
     task_id2 = tasks[1]['id']
 
-    shakedown.kill_process_on_host(host, '[h]ttp.server')
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
```
jeschkies: Nice!

tests/system/common.py
```diff
@@ -790,3 +790,26 @@ def task_by_name(tasks, name):
 
 async def assert_event(event_type, event_stream, within=10):
     await asyncio.wait_for(find_event(event_type, event_stream), within)
+
+
+def kill_process_on_host(hostname, pattern):
```
jeschkies: Should we use this method in other places as well?

zen-dog: No, ideally not at all or as little as possible. `kill_task(host, '[h]ttp.server')` will kill *all http.server`s on the host - that makes running tests in parallel impossible.

tests/system/common.py
timcharper: Yep, the should totally just be a one-liner. It seems that Python doesn't need to be involved.

tests/system/marathon_pods_tests.py
```diff
@@ -545,7 +545,12 @@ def wait_for_status_network_info():
...
+    def kill_task(host, pattern):
+        killed = common.kill_process_on_host(host, pattern)
+        assert killed, "no task got killed on {} for pattern {}".format(host, pattern)
+
+    kill_task(host, '[h]ttp.server')
```
timcharper: `'[h]ttp.server' ` is the same as `'http.server' `, right? What is the intention with the square brackets here?

Also, `.` will match any character? Should we escape that?

tests/system/common.py
ichernetsky: It is an amended version of `shakedown.kill_process_on_host`, which returns `None`, so there is no way to tell if anything has been killed at all. I agree that it can be improved.

tests/system/common.py
jeschkies: Yes. But one can define the pattern. E.g. how does it related to `shakedown.kill_process_on_host(host, '[s]leep 1000')` in `marathon_common_tests.py`?

tests/system/common.py
```diff
@@ -790,3 +790,22 @@ def task_by_name(tasks, name):
...
+    """ Kill the process matching pattern at ip
+
+        :param hostname: the hostname or ip address of the host on which the process will be killed
+        :param pattern: a regular expression matching the name of the process to kill
+        :return: True if at least one process gets killed, or False otherwise
+    """
+
+    cmd = "ps aux | grep -v grep | grep '{}' | awk '{{ print $2 }}' | tee >(xargs sudo kill -9)".format(pattern)
```
jeschkies: Sweet!

tests/system/common.py
```diff
@@ -790,3 +790,22 @@ def task_by_name(tasks, name):
...
+    status, stdout = shakedown.run_command_on_agent(hostname, cmd)
+    pids = [p.strip() for p in stdout.splitlines()]
+    if pids:
+        print("Killed pids: {}".format(", ".join(pids)))
+        return True
```
jeschkies: Nit: I'd return which pids where killed and then `assert len(killed) > 0`.

PR #6037, opened 2018-02-21
Update docs for PATCH mode

meln1k:
Summary: more precise description of the patch mode for apps

JIRA issues: MARATHON-2670

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -143,8 +143,9 @@
   patch:
     description:
       Change multiple existing applications by applying a patch.
-      Instances of these applications will be started or restarted according to `needsRestart` logic to meet the
-      requirements of the new app definitions.
+      Instances of these applications will be started or restarted according to the ususal logic to meet the
+      requirements of the new app definitions. Usually, restart is required if the application configuration was changed
+      (except for scaling changes).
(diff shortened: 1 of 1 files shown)


docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -143,7 +143,8 @@
   patch:
     description:
       Change multiple existing applications by applying a patch.
-      All instances of these applications get replaced by the new version.
+      Instances of these applications will be started or restarted according to `needsRestart` logic to meet the
```
alenkacz: according to *the* ... ?

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -370,7 +371,8 @@
     #
     patch:
       description:
-        Replaces parameters of a running application. All running instances get upgraded to the new definition.
+        Replaces parameters of a running application. Instances of this application will be started or restarted
+        according to `needsRestart` logic to meet the requirements of the new definition.
```
zen-dog: If you're mentioning it you should also specify what the logic currently does.

docs/docs/rest-api/public/api/v2/apps.raml
alenkacz: also does everyone know what's 'needsRestart' logic?

PR #6038, opened 2018-02-22
Do not start more instances than needed

ichernetsky:
When determining how many instances to launch during deployments
and scale checks, do account for `isReserved` instances, otherwise
Marathon starts extra instances in case of apps and pods with
persistent volumes.

JIRA issues: MARATHON-7751


ichernetsky:
Come on, Karsten, why a reject? It is work-in-progress. The PR is created just to have a build in order to verify the fix using DC/OS clusters. :)
Yes, it happens after a deployment finishes, but a scale check initiates before it finishes, this is what makes Marathon start more instances.

ichernetsky:
Karsten, if ran `test_restart_container_with_persistent_volume`, `test_app_with_persistent_volume_recovers`, and `test_pod_with_persistent_volume_recovers` thousands, if not tens of thousands, times. I have just fired open and permissive SI tests. Is it OK to merge it?

`test_pod_with_persistent_volume_recovers` was failing once in a thousand times on overage due to https://jira.mesosphere.com/browse/MARATHON-8110 which will be addressed in a separate PR.

jeschkies:
I just dug into the logs of my loop and happened to [find](https://jira.mesosphere.com/browse/MARATHON-7751?focusedCommentId=142898&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-142898) the same lines. Awesome work @ichernetsky!

PR #6039, opened 2018-02-22
Trying to bring #5909 back to life 

meln1k:


jeschkies:
@meln1k, should we close this or are you going to keep trying 🙂

jeschkies:
Won't do for now.

PR #6040, opened 2018-02-22
Enhance logs in TaskLauncherActor.

jeschkies:
Summary:
We sometimes do not know when the `TaskLauncherActor` accepts or rejects
offers. This change should help us to debug MARATHON-7751.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -146,7 +146,7 @@ class InstanceOpFactoryImpl(
      */
 
     def maybeLaunchOnReservation: Option[OfferMatchResult] = if (needToLaunch) {
-      logger.info("Need to launch on reservation.")
+      logger.debug(s"Need to launch on reservation for ${runSpec.id}, version ${runSpec.version}")
       val maybeVolumeMatch = PersistentVolumeMatcher.matchVolumes(offer, request.reserved)
 
       maybeVolumeMatch.map { volumeMatch =>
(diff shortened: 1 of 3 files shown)


jeschkies:
I also created a [loop](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-karsten/).

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -146,6 +146,7 @@ class InstanceOpFactoryImpl(
      */
 
     def maybeLaunchOnReservation: Option[OfferMatchResult] = if (needToLaunch) {
+      logger.info("Need to launch on reservation.")
```
alenkacz: wouldn't it help to also say for which runspec? When there are many apps on the cluster it might not be concrete enough... ?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -179,6 +180,7 @@ class InstanceOpFactoryImpl(
     } else None
 
     def maybeReserveAndCreateVolumes: Option[OfferMatchResult] = if (needToReserve) {
+      logger.info("Need to reserve.")
```
alenkacz: the same here

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jeschkies: Well, yeah

PR #6041, opened 2018-02-22
Update docs for PATCH mode (#6037)

meln1k:
Summary: more precise description of the patch mode for apps

JIRA issues: MARATHON-2670

(cherry picked from commit ebd2765)

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -143,8 +143,8 @@
   patch:
     description:
       Change multiple existing applications by applying a patch.
-      Instances of these applications will be started or restarted according to the ususal logic to meet the
-      requirements of the new app definitions. Usually, restart is required if the application configuration was changed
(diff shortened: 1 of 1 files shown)


docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -143,7 +143,9 @@
   patch:
     description:
       Change multiple existing applications by applying a patch.
-      All instances of these applications get replaced by the new version.
+      Instances of these applications will be started or restarted according to the ususal logic to meet the
```
alenkacz: typo in usual

docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -143,7 +143,9 @@
...
+      requirements of the new app definitions. Usually, restart is required if the application configuration was changed
```
alenkacz: Most of the time instead of usually here? So that we don't repeat the same word...

PR #6042, opened 2018-02-22
MAWS Enabled Publishing

kensipe:
MAWS Enabled Publishing

Summary:
maws works with profiles instead of keys and secrets.   
Using the Default provider chain enables maws to work

JIRA issues:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import java.time.{LocalDate, ZoneOffset}
 import java.time.format.DateTimeFormatter
 
-import com.amazonaws.auth.{AWSCredentialsProviderChain, EnvironmentVariableCredentialsProvider, InstanceProfileCredentialsProvider}
+import com.amazonaws.auth.DefaultAWSCredentialsProviderChain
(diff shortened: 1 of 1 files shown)


alenkacz:
how does MAWS publishing works? Why don't we need env variable credentials? Can someone help me understand? Thanks :)

jeschkies:
> how does MAWS publishing works? Why don't we need env variable credentials? Can someone help me understand? Thanks :)

The AWS SDK has a certain search path to find credentials. Check the [docs](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) for the order. When all tools abide to this chain it is fairly easy to authenticate a tool that uses the SDK.

When you login with MAWS it creates or updates an AWS profile under `~/.aws/credentials`. As the docs say, this is the third step in the credential chain.

With this change one just has to enable a profile that has access to the S3 bucket for Marathon Maven releases.

alenkacz:
@jeschkies @kensipe and this will still work from jenkins because we expose the right env variables ... ?

PR #6045, opened 2018-02-23
Send Instance health changed event when the overall status changed

clems4ever:


clems4ever:
We are currently using this patch in v1.5.6. I'd be glad to put it in master if need be.

kensipe:
test

kensipe:
our norm is to add features to master and back port... it would be great if you would add to master.   I will review soon(ish).    
you are currently using this code in our production env?   that is fantastic!

clems4ever:
Got it, I'll do the patch in master then.

clems4ever:
done: https://github.com/mesosphere/marathon/pull/6046

PR #6046, opened 2018-02-23
Send Instance health changed event when the overall status changed

clems4ever:

Backport in branch v1.5: https://github.com/mesosphere/marathon/pull/6045

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
@@ -7,7 +7,7 @@ import mesosphere.marathon.core.health.{Health, MarathonHttpHealthCheck, PortRef
 import mesosphere.marathon.core.health.impl.AppHealthCheckActor.{ApplicationKey, InstanceKey}
 import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.state.PathId
-import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.annotations.{Mode, OutputTimeUnit, Scope, State, BenchmarkMode, Fork, Benchmark}
 import org.openjdk.jmh.infra.Blackhole
 import mesosphere.marathon.state.Timestamp
 
@@ -28,16 +28,16 @@ object AppHealthCheckBenchmark {
     MarathonHttpHealthCheck(portIndex = Some(PortReference(443))),
     MarathonHttpHealthCheck(portIndex = Some(PortReference(8080)))
   )
-  val applicationKeys = 0.to(NB_APPLICATIONS).map(appId => {
+  val applicationKeys = 0.to(NB_APPLICATIONS).flatMap(appId => {
     0.to(NB_VERSIONS_PER_APPLICATION).map(version => Timestamp(version.toLong))
       .map(version => ApplicationKey(PathId(s"$appId"), version))
-  }).flatten
+  })
 
   val shuffledApplicationKeys = scala.util.Random.shuffle(applicationKeys)
 
-  val instanceKeys = applicationKeys.map(appKey => {
+  val instanceKeys = applicationKeys.flatMap(appKey => {
     0.to(NB_INSTANCES_PER_APPLICATION).map(instanceId => InstanceKey(appKey, Instance.Id(s"$instanceId")))
-  }).flatten
+  })
 
   val randomlySelectedInstanceKeysWithStatus = 0.to(NB_STATUS_UPDATES).map(idx => {
     val instanceIdx = randomGenerator.nextInt(NB_INSTANCES)
--- a/benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
@@ -1,0 +1,105 @@
+package mesosphere.marathon
+package core.health.impl
+
+import java.util.concurrent.TimeUnit
+
+import mesosphere.marathon.core.health.{Health, MarathonHttpHealthCheck, PortReference, impl}
+import mesosphere.marathon.core.health.impl.AppHealthCheckActor.{ApplicationKey, InstanceKey}
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.state.PathId
+import org.openjdk.jmh.annotations._
+import org.openjdk.jmh.infra.Blackhole
+import mesosphere.marathon.state.Timestamp
+
+object AppHealthCheckBenchmark {
+  private final val NB_APPLICATIONS = 1000
+  private final val NB_VERSIONS_PER_APPLICATION = 3
+
+  // this number is used for head revision of the app only
+  private final val NB_INSTANCES_PER_APPLICATION = 10
+  private final val NB_INSTANCES = NB_APPLICATIONS * NB_INSTANCES_PER_APPLICATION
+
+  private final val NB_STATUS_UPDATES = 10000
+
+  val randomGenerator = scala.util.Random
+
+  val healthChecks = Seq(
+    MarathonHttpHealthCheck(portIndex = Some(PortReference(80))),
+    MarathonHttpHealthCheck(portIndex = Some(PortReference(443))),
+    MarathonHttpHealthCheck(portIndex = Some(PortReference(8080)))
+  )
+  val applicationKeys = 0.to(NB_APPLICATIONS).map(appId => {
+    0.to(NB_VERSIONS_PER_APPLICATION).map(version => Timestamp(version.toLong))
+      .map(version => ApplicationKey(PathId(s"$appId"), version))
+  }).flatten
+
+  val shuffledApplicationKeys = scala.util.Random.shuffle(applicationKeys)
+
+  val instanceKeys = applicationKeys.map(appKey => {
+    0.to(NB_INSTANCES_PER_APPLICATION).map(instanceId => InstanceKey(appKey, Instance.Id(s"$instanceId")))
+  }).flatten
+
+  val randomlySelectedInstanceKeysWithStatus = 0.to(NB_STATUS_UPDATES).map(idx => {
+    val instanceIdx = randomGenerator.nextInt(NB_INSTANCES)
+    val hcIdx = instanceIdx % 3
+    val healthSelector = instanceIdx % 2
+    val instanceKey = instanceKeys(instanceIdx)
+    val failingHealth = Health(instanceKey.instanceId, lastSuccess = Some(Timestamp(5)), lastFailure = Some(Timestamp(10)))
+    val successHealth = Health(instanceKey.instanceId, lastSuccess = Some(Timestamp(5)), lastFailure = Some(Timestamp(0)))
+    val health = healthSelector match {
+      case 0 => successHealth
+      case _ => failingHealth
+    }
+    (instanceKey.applicationKey, healthChecks(hcIdx), health)
+  })
+  val doNothingNotifier = (h: Option[Boolean]) => {}
+
+  val appHealthCheckProxy = new impl.AppHealthCheckActor.AppHealthCheckProxy
+  val appHealthCheckProxyWithAlreadyRegisteredHealthChecks = {
+    val proxy = new impl.AppHealthCheckActor.AppHealthCheckProxy
+    for(
+      applicationKey <- applicationKeys;
+      healthCheck <- healthChecks
+    ) {
+      proxy.addHealthCheck(applicationKey, healthCheck)
+    }
+    proxy
+  }
+}
+
+@State(Scope.Benchmark)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@BenchmarkMode(Array(Mode.Throughput, Mode.AverageTime))
+@Fork(1)
+class AppHealthCheckBenchmark {
+  import AppHealthCheckBenchmark._
+
+  /**
+    * Simulate Marathon starting and registering all health checks
+    * @param hole
+    */
+  @Benchmark
+  def addHealthChecks(hole: Blackhole): Unit = {
+    for(
+      applicationKey <- shuffledApplicationKeys;
+      healthCheck <- healthChecks
+    ) {
+      appHealthCheckProxy.addHealthCheck(applicationKey, healthCheck)
+    }
+  }
+
+  /**
+    * Simulate all health checks reporting at the same time, which is unlickely
+    * to happen but give an idea on the performance of this method.
+    * @param hole
+    */
+  @Benchmark
+  def updateHealthCheckStatuses(hole: Blackhole): Unit = {
+    for(
+      update <- randomlySelectedInstanceKeysWithStatus
+    ) {
+      appHealthCheckProxyWithAlreadyRegisteredHealthChecks
+        .updateHealthCheckStatus(update._1, update._2, update._3, doNothingNotifier)
+    }
+  }
+}
--- a/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
@@ -1,0 +1,112 @@
+package mesosphere.marathon
+package core.health.impl
+
+import akka.actor.{ Actor, Props }
+import akka.event.EventStream
+import com.typesafe.scalalogging.StrictLogging
+import mesosphere.marathon.core.event.InstanceHealthChanged
+import mesosphere.marathon.core.health.impl.AppHealthCheckActor._
+import mesosphere.marathon.core.health.{ Health, HealthCheck }
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.state.{ PathId, Timestamp }
+
+/**
+  * This actor aggregates the statuses of health checks at the application level
+  * in order to maintain a global healthiness status for each instance.
+  *
+  * @param eventBus The eventStream to publish status changed events to
+  */
+class AppHealthCheckActor(eventBus: EventStream) extends Actor with StrictLogging {
+  /**
+    * Map of health check definitions of all applications
+    */
+  private var healthChecks: Map[ApplicationKey, Set[HealthCheck]] = Map.empty
+
+  /**
+    *  Map of results of all health checks for all applications.
+    *  Results are optional, therefore the global health status of an instance
+    *  is either:
+    *    unknown (if some results are still missing)
+    *    healthy (if all results are known and healthy)
+    *    not healthy (if all results are known and at least one is unhealthy)
+    */
+  private var healthCheckStates: Map[InstanceKey, Map[HealthCheck, Option[Health]]] =
+    Map.empty
+
+
+  private def computeGlobalHealth(instanceHealthResults: Map[HealthCheck, Option[Health]]): Option[Boolean] = {
+    val instanceHealthResultsSet = instanceHealthResults.values.toSet
+    val healthy = instanceHealthResultsSet.forall(x => x.fold(false)(_.alive))
+    val unhealthy = instanceHealthResultsSet.forall(_.nonEmpty) && !healthy
+    (healthy, unhealthy) match {
+      case (true, _) => Some(true)
+      case (_, true) => Some(false)
+      case _ => Option.empty[Boolean]
+    }
+  }
+
+  private def notifyHealthChanged(applicationKey: ApplicationKey, instanceId: Instance.Id,
+    healthiness: Option[Boolean]): Unit = {
+    logger.debug(s"Instance global health status changed to healthiness=$healthiness " +
+      s"for instance appId:$applicationKey instanceId:$instanceId")
+    eventBus.publish(InstanceHealthChanged(
+      instanceId, applicationKey.version, applicationKey.appId, healthiness))
+  }
+
+  private def healthCheckExists(applicationKey: ApplicationKey, healthCheck: HealthCheck): Boolean =
+    healthChecks.contains(applicationKey) && healthChecks(applicationKey).contains(healthCheck)
+
+  override def receive: Receive = {
+    case AddHealthCheck(appKey, healthCheck) =>
+      healthChecks = healthChecks +
+        (appKey -> (healthChecks.getOrElse(appKey, Set.empty) + healthCheck))
+      logger.debug(s"Add health check $healthCheck to instance appId:${appKey.appId} version:${appKey.version}")
+
+    case RemoveHealthCheck(appKey, healthCheck) =>
+      logger.debug(s"Remove health check $healthCheck from instance appId:${appKey.appId} version:${appKey.version}")
+
+      healthChecks = healthChecks +
+        (appKey -> (healthChecks.getOrElse(appKey, Set.empty) - healthCheck))
+
+      healthCheckStates = healthCheckStates.map(kv => {
+        val newHealthChecks = kv._2.filter(x => x._1 != healthCheck)
+        (kv._1, newHealthChecks)
+      })
+
+    case HealthCheckStatusChanged(appKey, healthCheck, health) =>
+      if (healthCheckExists(appKey, healthCheck)) {
+        logger.debug(s"Status changed to $health for health check $healthCheck of " +
+          s"instance appId:${appKey.appId} version:${appKey.version} instanceId:${health.instanceId}")
+
+        val instanceKey = InstanceKey(appKey, health.instanceId)
+        val currentInstanceHealthResults = healthCheckStates.getOrElse(instanceKey, {
+          healthChecks.getOrElse(appKey, Set.empty).map(x => (x, Option.empty[Health])).toMap
+        })
+
+        val newInstanceHealthResults = currentInstanceHealthResults + (healthCheck -> Some(health))
+
+        val currentInstanceGlobalHealth = computeGlobalHealth(currentInstanceHealthResults)
+        val newInstanceGlobalHealth = computeGlobalHealth(newInstanceHealthResults)
+
+        // only notifies on transitions between statuses
+        if (currentInstanceGlobalHealth != newInstanceGlobalHealth)
+          notifyHealthChanged(appKey, health.instanceId, newInstanceGlobalHealth)
+
+        healthCheckStates = healthCheckStates +
+          (instanceKey -> newInstanceHealthResults)
+      }
+  }
+}
+
+object AppHealthCheckActor {
+  case class ApplicationKey(appId: PathId, version: Timestamp)
+  case class InstanceKey(applicationKey: ApplicationKey, instanceId: Instance.Id)
+
+  def props(eventBus: EventStream): Props = Props(new AppHealthCheckActor(eventBus))
+
+  case class AddHealthCheck(appKey: ApplicationKey, healthCheck: HealthCheck)
+  case class RemoveHealthCheck(appKey: ApplicationKey, healthCheck: HealthCheck)
+  case class HealthCheckStatusChanged(
+    appKey: ApplicationKey,
+    healthCheck: HealthCheck, health: Health)
+}
--- a/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
@@ -5,7 +5,14 @@ import akka.actor.{ Actor, Props }
 import akka.event.EventStream
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.event.InstanceHealthChanged
-import mesosphere.marathon.core.health.impl.AppHealthCheckActor._
+import mesosphere.marathon.core.health.impl.AppHealthCheckActor.{
+  AddHealthCheck,
+  RemoveHealthCheck,
+  PurgeHealthCheckStatuses,
+  HealthCheckStatusChanged,
+  AppHealthCheckProxy,
+  ApplicationKey
+}
 import mesosphere.marathon.core.health.{ Health, HealthCheck }
 import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.state.{ PathId, Timestamp }
--- a/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
@@ -13,25 +13,59 @@ import mesosphere.marathon.state.{ PathId, Timestamp }
 import scala.collection.generic.Subtractable
 import scala.collection.mutable
 
+/**
+  * Actor computing instance healthiness of all instances of each application
+  * in order to send `InstanceHealthChanged` events to the event bus when the healthiness
+  * of a given instance changes. It basically aggregates all statuses of a given instance.
+  *
+  * There are 3 possible healthiness states:
+  * - `Unknown` if at least one health check status of the instance is unknown
+  * - `Unhealthy` if all statuses of the instance are known but at least one is failing
+  * - `Healthy` if all statuses of the instance are known and all are healthy
+  */
 object AppHealthCheckActor {
   case class ApplicationKey(appId: PathId, version: Timestamp)
   case class InstanceKey(applicationKey: ApplicationKey, instanceId: Instance.Id)
 
   def props(eventBus: EventStream): Props = Props(new AppHealthCheckActor(eventBus))
 
+  /**
+    * Actor command adding an health check definition for a given application so that the actor knows which health check
+    * statuses are know and which are not for a given instance.
+    * @param appKey The application key of the application for which the health check is defined
+    * @param healthCheck The definition of the health check to add
+    */
   case class AddHealthCheck(appKey: ApplicationKey, healthCheck: HealthCheck)
+
+  /**
+    * Actor command removing a health check definition for a given application.
+    * This method also purges the statuses of health checks of all instances that ran it.
+    * @param appKey The application key of the application for which the health check is removed
+    * @param healthCheck The definition of the health check to remove
+    */
   case class RemoveHealthCheck(appKey: ApplicationKey, healthCheck: HealthCheck)
 
-  case class PurgeHealthCheckStatuses(hc: Seq[(InstanceKey, HealthCheck)])
+  /**
+    * Actor command purging several health check statuses.
+    * @param toPurge The list of health check statuses to purge.
+    */
+  case class PurgeHealthCheckStatuses(toPurge: Seq[(InstanceKey, HealthCheck)])
+
+  /**
+    * Actor command updating the status of an health check of a given instance of an application. If the instance status
+    * has changed, it sends a `InstanceHealthChanged` event to the event bus.
+    * @param appKey The application key for which the health check status is updated
+    * @param healthCheck The health check for which the status is updated
+    * @param healthState The new state of the health check
+    */
   case class HealthCheckStatusChanged(
       appKey: ApplicationKey,
-      healthCheck: HealthCheck, health: Health)
+      healthCheck: HealthCheck, healthState: Health)
 
-  trait InstanceHealthChangedNotifier {
-    def notify(applicationKey: ApplicationKey, instanceId: Instance.Id,
-      healthiness: Option[Boolean]): Unit
-  }
-
+  /**
+    * This proxy class is the implementation of the actor. It has been created to be used in performance benchmarks.
+    * Beware this class is NOT thread safe.
+    */
   class AppHealthCheckProxy extends StrictLogging {
     /**
       * Map of health check definitions of all applications
@@ -39,8 +73,8 @@ object AppHealthCheckActor {
     private[impl] val healthChecks: mutable.Map[ApplicationKey, Set[HealthCheck]] = mutable.Map.empty
 
     /**
-      *  Map of results of all health checks for all applications.
-      *  Results are optional, therefore the global health status of an instance
+      *  Map of statuses of all health checks for all instances of all applications.
+      *  Statuses are optional, therefore the global health status of an instance
       *  is either:
       *    unknown (if some results are still missing)
       *    healthy (if all results are known and healthy)
@@ -49,6 +83,11 @@ object AppHealthCheckActor {
     private[impl] val healthCheckStates: mutable.Map[InstanceKey, Map[HealthCheck, Option[Health]]] =
       mutable.Map.empty
 
+    /**
+      * This method derives the instance healthiness from a list of health check statuses
+      * @param instanceHealthResults The map of health checks statuses per health check definition
+      * @return The instance global healthiness
+      */
     private def computeGlobalHealth(instanceHealthResults: Map[HealthCheck, Option[Health]]): Option[Boolean] = {
       val isHealthAlive = (health: Option[Health]) => health.fold(false)(_.alive)
       val isHealthUnknown = (health: Option[Health]) => health.isEmpty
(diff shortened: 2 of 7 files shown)


pierresouchay:
This is a really important change for us as we have apps that have several healthchecks and some of them take time to be up, especially for applications with a warming up phase.

It really cause us some troubles on some of our apps, and it would benefits to many to have this kind of sane behaviour, especially when restarting applications or upgrading applications with new versions.

clems4ever:
Hello @alenkacz, no problem. It has been a pain in our company for the restart of apps, thus we decided to tackle it.
Regarding this PR, I simplified a bit the code on our end. I'm rebasing right now.

clems4ever:
The rebase is done.

clems4ever:
Do you know why it takes forever to build the new change?

kensipe:
@clems4ever apologies... we have it setup so that PRs require a committer to bump the build.   this prevents malicious executions... if everyone is paying attention... let me help with that

kensipe:
hit "proceed" button!   unicorns and gerbils released!

clems4ever:
thank you @kensipe!

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,92 @@
...
+import akka.actor.{ Actor, Props }
+import akka.event.EventStream
+import com.typesafe.scalalogging.StrictLogging
+import mesosphere.marathon.core.event.InstanceHealthChanged
+import mesosphere.marathon.core.health.impl.AppHealthCheckActor._
+import mesosphere.marathon.core.health.{ Health, HealthCheck }
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.state.{ PathId, Timestamp }
+
+class AppHealthCheckActor(eventBus: EventStream) extends Actor with StrictLogging {
```
kensipe: we could use some class documentation on the purpose of this case

kensipe: add detail on why it is here... not what it does...  provide some detail around the intent of global health

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,92 @@
...
+
+      healthCheckStates = healthCheckStates.map(kv => {
+        val newHealthChecks = kv._2.filter(x => x._1 != healthCheck)
+        (kv._1, newHealthChecks)
+      })
+
+    case HealthCheckStatusChanged(appKey, healthCheck, health) =>
+      if (healthCheckExists(appKey, healthCheck)) {
+        logger.debug(s"Healthcheck status changed to $health for health check $healthCheck of " +
+          s"instance ${appKey.appId} ${appKey.version} ${health.instanceId}")
```
kensipe: lets add more detail here... I don't want to need to remember order of values in the logs... could we prefix the `${appKey.appId}` with `appId: ${appKey.appId}`.   same for version and instanceId

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: Sure, I will add some documentation. This will indeed help people understand the purpose of this class.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: You are right, I will add those prefixes.

clems4ever:
Thank you very much for the review @kensipe! This one was really critical for us since we have more and more apps deployed in our platform. I guess a lot of users will be happy to have it in the next release :).

clems4ever:
@kensipe , I added some documentation and updated the logs.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,112 @@
...
+    *    unknown (if some results are still missing)
+    *    healthy (if all results are known and healthy)
+    *    not healthy (if all results are known and at least one is unhealthy)
+    */
+  private var healthCheckStates: Map[InstanceKey, Map[HealthCheck, Option[Health]]] =
+    Map.empty
+
+
+  private def computeGlobalHealth(instanceHealthResults: Map[HealthCheck, Option[Health]]): Option[Boolean] = {
+    val instanceHealthResultsSet = instanceHealthResults.values.toSet
```
timcharper: Is the allocation of this Set needed?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,112 @@
...
+    val healthy = instanceHealthResultsSet.forall(x => x.fold(false)(_.alive))
+    val unhealthy = instanceHealthResultsSet.forall(_.nonEmpty) && !healthy
+    (healthy, unhealthy) match {
+      case (true, _) => Some(true)
```
timcharper: It seems like the unhealthy computation is not needed if healthy is true. Perhaps it would be better to convert this to a series of if statements?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,112 @@
...
+      healthChecks = healthChecks +
+        (appKey -> (healthChecks.getOrElse(appKey, Set.empty) - healthCheck))
+
+      healthCheckStates = healthCheckStates.map(kv => {
+        val newHealthChecks = kv._2.filter(x => x._1 != healthCheck)
+        (kv._1, newHealthChecks)
+      })
+
+    case HealthCheckStatusChanged(appKey, healthCheck, health) =>
+      if (healthCheckExists(appKey, healthCheck)) {
```
timcharper: Seems like perhaps we should log a debug message if the health check status changes and we don't have a record of this health check?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,112 @@
...
+  override def receive: Receive = {
+    case AddHealthCheck(appKey, healthCheck) =>
+      healthChecks = healthChecks +
+        (appKey -> (healthChecks.getOrElse(appKey, Set.empty) + healthCheck))
+      logger.debug(s"Add health check $healthCheck to instance appId:${appKey.appId} version:${appKey.version}")
+
+    case RemoveHealthCheck(appKey, healthCheck) =>
+      logger.debug(s"Remove health check $healthCheck from instance appId:${appKey.appId} version:${appKey.version}")
+
+      healthChecks = healthChecks +
```
timcharper: When do we prune unused map entries when an appKey is no longer used? (I.e. we deploy a new version of the app, and all the old instances of that app are killed)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,112 @@
...
+    *    healthy (if all results are known and healthy)
+    *    not healthy (if all results are known and at least one is unhealthy)
+    */
+  private var healthCheckStates: Map[InstanceKey, Map[HealthCheck, Option[Health]]] =
+    Map.empty
+
+
+  private def computeGlobalHealth(instanceHealthResults: Map[HealthCheck, Option[Health]]): Option[Boolean] = {
+    val instanceHealthResultsSet = instanceHealthResults.values.toSet
+    val healthy = instanceHealthResultsSet.forall(x => x.fold(false)(_.alive))
```
timcharper: Perhaps this is better?
```
instanceHealthResults.values.forall { health => health.alive.contains(true) }
```

src/test/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActorTest.scala
```diff
@@ -0,0 +1,133 @@
...
+      actor ! AddHealthCheck(f.appKey, f.hcPort80)
+      actor ! AddHealthCheck(f.appKey, f.hcPort443)
+
+      actor ! HealthCheckStatusChanged(f.appKey, f.hcPort80,
+        Health(f.instances.head, lastSuccess = Some(Timestamp(5)), lastFailure = Some(Timestamp(0))))
+      actor ! HealthCheckStatusChanged(f.appKey, f.hcPort443,
+        Health(f.instances.head, lastSuccess = Some(Timestamp(5)), lastFailure = Some(Timestamp(0))))
+
+      actor ! RemoveHealthCheck(f.appKey, f.hcPort80)
+      actor ! RemoveHealthCheck(f.appKey, f.hcPort443)
```
timcharper: Can we verify that the actor has cleaned up the entries for this appKey? Use akka testkit underlyingActor method to access the internal map directly?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: I will do that.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: You are right, the key space is growing indefinitely. I will fix that.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: Indeed! I'll update it. Thanks.

src/test/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActorTest.scala
clems4ever: I'll check how I can do it with what you propose.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: Not needed, just convenient. I'll optimize with what you propose.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: I agree, I'll use if else statements.

clems4ever:
@timcharper , nice catch! Thank you for the review.
@kensipe , I think there's an easy way to avoid the leak, I will implement it.

Regarding the benchmark, do you have a some sort of test suite you'd like me to use or should I build one from scratch?

alenkacz:
@clems4ever see this part of the repo https://github.com/mesosphere/marathon/tree/master/benchmark if you need any help with that, ping me on slack :-)

clems4ever:
Thank you @alenkacz , I'll have a look.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: Actually it does not compile and for a good reason: health is the option and alive is a bool.

clems4ever:
I've generated the benchmarks but not have the time to analyze yet.

mesosphere marathon core health impl apphealthcheckbenchmark addhealthchecks-averagetime-2 jfr
mesosphere marathon core health impl apphealthcheckbenchmark addhealthchecks-throughput-0 jfr
mesosphere marathon core health impl apphealthcheckbenchmark updatehealthcheckstatuses-averagetime-3 jfr
mesosphere marathon core health impl apphealthcheckbenchmark updatehealthcheckstatuses-throughput-1 jfr

clems4ever:
Hey, I've removed the graphs since they are not really useful here. I've done some optimizations and benchmarked them. Here are the the two tests:
1/ 1000 apps having 3 versions and 3 health checks. (addHealthCheck)
2/ 10000 randomly selected updates of health check statuses in bulk (updateHealthCheckStatus)

And the results
```
*Without optim*

[info] # Run complete. Total time: 00:02:42
[info] Benchmark                                           Mode  Cnt  Score   Error   Units
[info] AppHealthCheckBenchmark.addHealthChecks            thrpt   20  0.606 ± 0.005  ops/ms
[info] AppHealthCheckBenchmark.updateHealthCheckStatuses  thrpt   20  0.118 ± 0.004  ops/ms
[info] AppHealthCheckBenchmark.addHealthChecks             avgt   20  1.680 ± 0.017   ms/op
[info] AppHealthCheckBenchmark.updateHealthCheckStatuses   avgt   20  9.708 ± 0.594   ms/op

*With optim and immutable map*

[info] # Run complete. Total time: 00:02:42
[info] Benchmark                                           Mode  Cnt  Score   Error   Units
[info] AppHealthCheckBenchmark.addHealthChecks            thrpt   20  0.590 ± 0.017  ops/ms
[info] AppHealthCheckBenchmark.updateHealthCheckStatuses  thrpt   20  0.118 ± 0.003  ops/ms
[info] AppHealthCheckBenchmark.addHealthChecks             avgt   20  1.727 ± 0.032   ms/op
[info] AppHealthCheckBenchmark.updateHealthCheckStatuses   avgt   20  8.528 ± 0.214   ms/op


*With optim and mutable map*

[info] # Run complete. Total time: 00:02:42
[info] Benchmark                                           Mode  Cnt  Score   Error   Units
[info] AppHealthCheckBenchmark.addHealthChecks            thrpt   20  0.738 ± 0.019  ops/ms
[info] AppHealthCheckBenchmark.updateHealthCheckStatuses  thrpt   20  0.143 ± 0.027  ops/ms
[info] AppHealthCheckBenchmark.addHealthChecks             avgt   20  1.285 ± 0.009   ms/op
[info] AppHealthCheckBenchmark.updateHealthCheckStatuses   avgt   20  4.985 ± 0.118   ms/op
```

Here is my configuration:  8 core CPU and 32GB of RAM

Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    2
Core(s) per socket:    4
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 94
Model name:            Intel(R) Xeon(R) CPU E3-1270 v5 @ 3.60GHz
Stepping:              3
CPU MHz:               3600.000
CPU max MHz:           4000,0000
CPU min MHz:           800,0000
BogoMIPS:              7200.00
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              8192K
NUMA node0 CPU(s):     0-7
Flags:                 fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb invpcid_single pti intel_pt tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm mpx rdseed adx smap clflushopt xsaveopt xsavec xgetbv1 xsaves dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp

And 2 slots of RAM with the following specs:

Handle 0x000A, DMI type 17, 40 bytes
Memory Device
	Array Handle: 0x0009
	Error Information Handle: Not Provided
	Total Width: 64 bits
	Data Width: 64 bits
	Size: 16384 MB
	Form Factor: DIMM
	Set: None
	Locator: DIMM1
	Bank Locator: Not Specified
	Type: DDR4
	Type Detail: Synchronous Unbuffered (Unregistered)
	Speed: 2400 MHz
	Manufacturer: 01980000802C
	Serial Number: E8391463
	Asset Tag: 04171700
	Part Number: KYXC0V-MIB          
	Rank: 2
	Configured Clock Speed: 2133 MHz
	Minimum Voltage: Unknown
	Maximum Voltage: Unknown
	Configured Voltage: 1.2 V

clems4ever:
Can someone tell me why the tests end red at head since I cannot check the logs?

timcharper:
Hey @clems4ever , I'll get the tests running today and will investigate this, sorry for the delay! It has taken me a while to go through all of my notifications.

timcharper:
We are still getting used to the new pull request authorization system, we missed the error message. I have kicked off the build again.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,157 @@
...
+  trait InstanceHealthChangedNotifier {
+    def notify(applicationKey: ApplicationKey, instanceId: Instance.Id,
+      healthiness: Option[Boolean]): Unit
+  }
+
+  class AppHealthCheckProxy extends StrictLogging {
+    /**
+      * Map of health check definitions of all applications
+      */
+    var healthChecks: mutable.Map[ApplicationKey, Set[HealthCheck]] = mutable.Map.empty
```
timcharper: Mutable variables should be `private [this]`

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+      *  Map of results of all health checks for all applications.
+      *  Results are optional, therefore the global health status of an instance
+      *  is either:
+      *    unknown (if some results are still missing)
+      *    healthy (if all results are known and healthy)
+      *    not healthy (if all results are known and at least one is unhealthy)
+      */
+    var healthCheckStates: mutable.Map[InstanceKey, Map[HealthCheck, Option[Health]]] =
```
timcharper: `private [this]`

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
timcharper: Also, double mutable (var + mutable datastructure). Prefer:

```
val healthChecks: mutable.Map[...]
```

or

```
var healthChecks: immutable.Map[...]
```

But not both :-)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
timcharper: Double mutable, also

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,157 @@
...
+    def removeHealthCheck(applicationKey: ApplicationKey, healthCheck: HealthCheck): Unit = {
+      logger.debug(s"Remove health check $healthCheck from instance appId:${applicationKey.appId} version:${applicationKey.version}")
+
+      val healthChecksAfterRemoval = healthChecks.getOrElse(applicationKey, Set.empty) - healthCheck
+      if (healthChecksAfterRemoval.isEmpty)
+        healthChecks.remove(applicationKey)
+      else
+        healthChecks.update(applicationKey, healthChecksAfterRemoval)
+
+      healthCheckStates = healthCheckStates.map(kv => {
```
timcharper: Would be probably best to avoid creating a new copy of the map here, especially since this is a mutable map.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,157 @@
...
+        val newHealthChecks = kv._2.filter(x => x._1 != healthCheck)
+        (kv._1, newHealthChecks)
+      }).filter(kv => kv._2.nonEmpty)
```
timcharper:
```
healthCheckStates.keysIterator.foreach { k =>
  val newValue = healthCheckStates(k).filter { case (_, state) => state != healthCheck }
  if (newValue.isEmpty)
    healthCheckStates.remove(k)
  else
    healthCheckStates(k) = newValue
}
```

timcharper: You could also potentially implement a generic garbage collection method, which is called after `purgeHealthCheckStatuses`?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    def purgeHealthCheckStatuses(toPurge: Seq[(InstanceKey, HealthCheck)]): Unit = {
+      toPurge.foreach(hc => {
+        val instanceKey = hc._1
```
timcharper: Preference to use pattern matching/deconstructor

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,157 @@
...
+          healthCheckStates.remove(instanceKey)
+        else
+          healthCheckStates.update(instanceKey, newInstanceHealthChecks)
+      })
+    }
+
+    def updateHealthCheckStatus(appKey: ApplicationKey, healthCheck: HealthCheck, health: Health,
+      notifier: (Option[Boolean] => Unit)): Unit = {
+      val hcDefinitions = healthChecks.getOrElse(appKey, Set.empty)
+      if (hcDefinitions.contains(healthCheck)) {
```
timcharper: Perhaps this reads little better?

```
healthChecks.get(appKey) match {
  case Some(hcDefinitions) if hcDefinitions.contains(healthCheck) =>
    ...
  case _ =>
    logger.warn(s"Status of $healthCheck health check changed but it does not exist in inventory")
}
```

(It is honestly fine is written, though :-))

src/test/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActorTest.scala
```diff
@@ -0,0 +1,221 @@
...
+        Health(f.instances(2), lastSuccess = Some(Timestamp(5)), lastFailure = Some(Timestamp(0))))
+
+      systemLog.expectMsg(InstanceHealthChanged(
+        f.instances(2), f.appKey.version, f.appKey.appId, Some(true)))
+
+      actor ! RemoveHealthCheck(f.appKey, f.hcPort80)
+      actor ! RemoveHealthCheck(f.appKey, f.hcPort443)
+    }
+
+    "health checks inventories is cleaned when health checks of a given version are removed" in {
```
timcharper: 👏 👏 👏

timcharper:
Jenkins cannot run the tests because the following file conflicts:

- src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala

I will try and watch more carefully. Feel free to flag us in the [Marathon community slack channel](https://mesos.slack.com/messages/C1L7D22KY) if we miss the build approval notification.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: I changed `var` to `val` but I cannot make the variables private otherwise we will lose a part of the test coverage. Some tests are checking if the purge works as expected.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: see comment above

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: I used the `retains` method of mutable maps.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: I've implemented a generic purge method that is used to purge both HC definitions and HC results.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: Did it.

src/test/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActorTest.scala
clems4ever: :)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: Well, I eventually made them private and end up creating methods to check existence of objects in the maps so that I can use them in the tests.

clems4ever:
I rebased the PR after the last commit.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
timcharper: You can make something private such that the test can access is, `private [impl]` (package scoped privates)

clems4ever: Ok, I'm trying now. Can you tell me what's wrong with the build in the meantime?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+  trait InstanceHealthChangedNotifier {
+    def notify(applicationKey: ApplicationKey, instanceId: Instance.Id,
+      healthiness: Option[Boolean]): Unit
+  }
+
+  class AppHealthCheckProxy extends StrictLogging {
+    /**
+      * Map of health check definitions of all applications
+      */
+    val healthChecks: mutable.Map[ApplicationKey, Set[HealthCheck]] = mutable.Map.empty
```
timcharper: `private [impl]` should mark this is private and still allow the test cases to access the map

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+  case class HealthCheckStatusChanged(
+      appKey: ApplicationKey,
+      healthCheck: HealthCheck, health: Health)
+
+  trait InstanceHealthChangedNotifier {
+    def notify(applicationKey: ApplicationKey, instanceId: Instance.Id,
+      healthiness: Option[Boolean]): Unit
+  }
+
+  class AppHealthCheckProxy extends StrictLogging {
```
timcharper: Can we document this?:

- Used internally by …
- NOT thread safe

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done. Tests are green on my side.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+import scala.collection.generic.Subtractable
+import scala.collection.mutable
+
+object AppHealthCheckActor {
+  case class ApplicationKey(appId: PathId, version: Timestamp)
+  case class InstanceKey(applicationKey: ApplicationKey, instanceId: Instance.Id)
+
+  def props(eventBus: EventStream): Props = Props(new AppHealthCheckActor(eventBus))
+
+  case class AddHealthCheck(appKey: ApplicationKey, healthCheck: HealthCheck)
```
timcharper: Should we document these Akka commands?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+  case class RemoveHealthCheck(appKey: ApplicationKey, healthCheck: HealthCheck)
+
+  case class PurgeHealthCheckStatuses(hc: Seq[(InstanceKey, HealthCheck)])
+  case class HealthCheckStatusChanged(
+      appKey: ApplicationKey,
+      healthCheck: HealthCheck, health: Health)
+
+  trait InstanceHealthChangedNotifier {
```
timcharper: This looks unused, is it?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+
+      if (instanceHealthResults.values.forall(isHealthAlive))
+        Some(true)
+      else if (instanceHealthResults.values.exists(isHealthUnknown))
+        Option.empty[Boolean]
+      else
+        Some(false)
+    }
+
+    def addHealthCheck(applicationKey: ApplicationKey, healthCheck: HealthCheck): Unit = {
```
timcharper: Docs :)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+      logger.debug(s"Add health check $healthCheck to instance appId:${applicationKey.appId} version:${applicationKey.version}")
+      healthChecks.update(applicationKey, healthChecks.getOrElse(applicationKey, Set.empty) + healthCheck)
+    }
+
+    def removeHealthCheck(applicationKey: ApplicationKey, healthCheck: HealthCheck): Unit = {
```
timcharper: Docs :)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+              if (newHcContainer.isEmpty)
+                healthChecksContainers.remove(key)
+              else
+                healthChecksContainers.update(key, newHcContainer)
+            case _ =>
+          }
+      })
+    }
+
+    def purgeHealthCheckDefinition(applicationKey: ApplicationKey, healthCheck: HealthCheck): Unit = {
```
timcharper: Docs :)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+      purgeHealthChecks[ApplicationKey, HealthCheck, Set[HealthCheck]](healthChecks, Seq(applicationKey -> healthCheck))
+    }
+
+    def purgeHealthChecksStatuses(toPurge: Seq[(InstanceKey, HealthCheck)]): Unit = {
```
timcharper: Docs :)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,172 @@
...
+      purgeHealthChecks[InstanceKey, (HealthCheck, Option[Health]), Map[HealthCheck, Option[Health]]](healthCheckStates, toPurge)
+    }
+
+    def updateHealthCheckStatus(appKey: ApplicationKey, healthCheck: HealthCheck, health: Health,
```
timcharper: Docs :)

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: removed

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
clems4ever: done

clems4ever:
I've added more documentation as suggested.

clems4ever:
Awesome! Thank you very much @alenkacz , @timcharper and @kensipe .

timcharper:
Just waiting for a second reviewer to approve, then we can merge.

kensipe:
back from vacation... quick scan of changes looks great!   I will make time to review today.

benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
```diff
@@ -0,0 +1,105 @@
+package mesosphere.marathon
+package core.health.impl
+
+import java.util.concurrent.TimeUnit
+
+import mesosphere.marathon.core.health.{Health, MarathonHttpHealthCheck, PortReference, impl}
+import mesosphere.marathon.core.health.impl.AppHealthCheckActor.{ApplicationKey, InstanceKey}
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.state.PathId
+import org.openjdk.jmh.annotations._
```
ichernetsky: In general we try to avoid `_` imports in the recent past. Could you please address here and in the other places of this patch?

src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala
```diff
@@ -2,26 +2,27 @@ package mesosphere.marathon
...
 import akka.stream.Materializer
 import akka.util.Timeout
 import scala.concurrent.ExecutionContext.Implicits.global
-import mesosphere.marathon.core.event.{ AddHealthCheck, RemoveHealthCheck }
+import mesosphere.marathon.core.event.{AddHealthCheck, RemoveHealthCheck}
 import mesosphere.marathon.core.group.GroupManager
 import mesosphere.marathon.core.health._
-import mesosphere.marathon.core.health.impl.HealthCheckActor.{ AppHealth, GetAppHealth, GetInstanceHealth }
+import mesosphere.marathon.core.health.impl.AppHealthCheckActor.ApplicationKey
+import mesosphere.marathon.core.health.impl.HealthCheckActor.{AppHealth, GetAppHealth, GetInstanceHealth}
```
ichernetsky: It seems that the code had not been compiled before updating the PR last time. Could you please fix formatting by running `compile` and `test:compile`?

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
```diff
@@ -0,0 +1,236 @@
...
+
+  /**
+    * This proxy class is the implementation of the actor. It has been created to be used in performance benchmarks.
+    * Beware this class is NOT thread safe.
+    */
+  class AppHealthCheckProxy extends StrictLogging {
+    /**
+      * Map of health check definitions of all applications
+      */
+    private[impl] val healthChecks: mutable.Map[ApplicationKey, Set[HealthCheck]] = mutable.Map.empty
```
ichernetsky: Do we care about "thread safety" here? I think we should. The docs say:

> if an actor makes changes to its internal state while processing a message, and accesses that state while processing another message moments later. It is important to realize that with the actor model you don't get any guarantee that the same thread will be executing the same actor for different messages.

ichernetsky: On the other hand, the docs say:

> In layman's terms this means that changes to internal fields of the actor are visible when the next message is processed by that actor. So fields in your actor need not be volatile or equivalent.

ichernetsky: It seems it is addressed by the "happens-before" rules. In this case, please ignore me.

benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
timcharper: I think it's fine here. The other benchmarks (that we wrote) do the same.

src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckActor.scala
timcharper: There is a comment at the top of this class, "BEWARE this class is NOT thread safe". This, plus the fact that this class is used by an actor, resolves all concerns I have.

src/main/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManager.scala
clems4ever: I've fixed it.

benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
clems4ever: I've removed _ imports in files where I've introduced it.

timcharper:
Looks like a compile failure:

```
[warn]                 ^
[error] /home/admin/workspace/marathon-pipelines_PR-6046-QIWMZLBIQDDBWOONUI6CTWJHBIM72YAVAKYVMQRZHABNTDRAPMIQ/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala:91:45: value isLaunched is not a member of mesosphere.marathon.core.instance.Instance
[error]     val hcToPurge = instances.withFilter(!_.isLaunched).map(instance => {
[error]                                             ^
```

timcharper:
(might have happened with a rebase... @meichstedt just changed that and removed that method, I think.)

timcharper:
Here is the commit:

4d20e105d72f5b690e09d206dccc4c595f0b68ec

```
commit 4d20e105d72f5b690e09d206dccc4c595f0b68ec
Author: Matthias Eichstedt <matthias.eichstedt@mesosphere.io>
Date:   Thu Mar 29 21:34:11 2018 +0200

    Removed the redundant Instance.isLaunched value in favor of Instance.isActive (#6127)

    JIRA issues: MARATHON-2547
```

clems4ever:
thank you for the fix @timcharper!

kensipe:
@clems4ever this is on master which means it is on 1.6+.     Were you looking for this to be on 1.5?  (which is DCOS 1.10)

clems4ever:
thank you @kensipe .

kensipe:
Also... I spent last week in Paris... it is a fantastic and beautiful city you live in!

clems4ever:
Well, I think I will try to rebase it in 1.5.6 tomorrow. If it's trivial, I think it's best to merge it so that other users can take advantage of it if needed.
Also, do you have an ETA for 1.6? It can help us decide if we can wait for the new version or if the release is too far from now to afford waiting.

Yes, Paris is great. Too bad we were not able to meet and take a beer.

kensipe:
the question of when 1.6 is tricky... we have a general policy now of have exhaustive testing before a release goes out which general is ~ 2 weeks in soak tests.    That means we would need to push a new 1.6 release internal first then a community release is 2 weeks away.   I will check to see if pushing a new build makes sense tomorrow.

clems4ever:
Oh, I have not seen that there is a 1.6 released already. This is great!
Regarding 1.5, I propose we wait the result of the rebase (hopefully tomorrow) to take the decision. As I said, if it's trivial it's just a matter of pushing a button I guess, isn't it?

kensipe:
our preference on backport to 1.5 is to cherry-pick the commit... I just did a dry run and it picks cleanly.   I'll just push a cherry-pick PR to 1.5.   it has the additionally benefit of maintaining your authorship on the commit.

PR #6047, opened 2018-02-23
Remove Ammonite from library dependencies

timcharper:
This adds about 30mb to our build artifact (ammonite + the scala compiler)

We shouldn't be distributing this with our production artifact


timcharper:
@meln1k if you need this, perhaps one of the solutions would work?

* Add it in to ~/.sbt/1.0/ammonite.sbt

```
libraryDependencies ++= Seq(
  ...
)
```

* Make it a test dependency, so it doesn't get distributed with our main artifact?

PR #6048, opened 2018-02-23
Allow to kill tasks during a deployment

meichstedt:
Killing tasks via the API is now always directly processed by the KillService. Before, when the `wipe` flag was not specified as `true`, killing tasks was indirected via the MarathonSchedulerService, and silently failed if the associated app was locked by a deployment, even if `force=true` was specified. With this patch, it is possible to kill tasks even if they are subject of a running deployment.

JIRA issues: MARATHON-1686

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -167,31 +167,6 @@ class MarathonSchedulerActor private (
     case cmd @ Deploy(plan, force) =>
       deploy(sender(), cmd)
 
-    case cmd @ KillTasks(runSpecId, tasks) =>
-      @SuppressWarnings(Array("all")) /* async/await */
-      def killTasks(): Future[Event] = {
-        logger.debug("Received kill tasks {} of run spec {}", tasks, runSpecId)
-        async {
-          await(killService.killInstances(tasks, KillReason.KillingTasksViaApi))
-          await(schedulerActions.scale(runSpecId))
-          self ! cmd.answer
-          cmd.answer
-        }.recover {
-          case t: Throwable =>
-            CommandFailed(cmd, t)
-        }
-      }
-
-      withLockFor(Set(runSpecId)) {
-        killTasks().pipeTo(sender)
-      } match {
-        case None =>
-          // KillTasks is user initiated. If we don't process it, then we should make it obvious as to why.
-          logger.warn(
-            s"Could not acquire lock while killing tasks ${tasks.map(_.instanceId).toList} for ${runSpecId}")
-        case _ =>
-      }
-
     case DeploymentFinished(plan) =>
       removeLocks(plan.affectedRunSpecIds)
       deploymentSuccess(plan)
@@ -201,8 +176,6 @@ class MarathonSchedulerActor private (
       deploymentFailed(plan, reason)
 
     case RunSpecScaled(id) => removeLock(id)
-
-    case TasksKilled(runSpecId, _) => removeLock(runSpecId)
 
     case msg => logger.warn(s"Received unexpected message from ${sender()}: $msg")
   }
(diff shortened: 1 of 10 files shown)


meichstedt:
Let's see what the CI says :trollface:

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -46,8 +45,8 @@ class TaskKiller @Inject() (
           if (wipe) {
             val expunged = await(expunge(foundInstances))
             val killed = await(killService.killInstances(launchedInstances, KillReason.KillingTasksViaApi))
-          } else {
-            if (launchedInstances.nonEmpty) service.killInstances(runSpecId, launchedInstances)
+          } else if (launchedInstances.nonEmpty) {
+            val killed = await(killService.killInstances(launchedInstances, KillReason.KillingTasksViaApi))
```
meichstedt: awaiting the future here means we actually wait until the task is reportedly killed. AFAICT, the future will be completed rather immediately if the task in question is unreachable. However, when the task is unreachable but Mesos doesn't report it as such (e.g. due to agent re-registration rate limit), waiting here will result in a request timeout.

It might make more sense to just trigger the kill and not wait.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -167,31 +167,6 @@ class MarathonSchedulerActor private (
     case cmd @ Deploy(plan, force) =>
       deploy(sender(), cmd)
 
-    case cmd @ KillTasks(runSpecId, tasks) =>
-      @SuppressWarnings(Array("all")) /* async/await */
-      def killTasks(): Future[Event] = {
-        logger.debug("Received kill tasks {} of run spec {}", tasks, runSpecId)
-        async {
-          await(killService.killInstances(tasks, KillReason.KillingTasksViaApi))
-          await(schedulerActions.scale(runSpecId))
```
meichstedt: A scale check will be initiated for each task that is reportedly terminal so the `schedulerActions.scale(runSpecId)` call here was pretty redundant.

meichstedt:
Nobody requested it yet, but I think there should be an explicit integration test to verify a deployment runs fine with a task being killed in between.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
zen-dog: The original code in `MarathonSchedulerActor` also triggered a scale check (`await(schedulerActions.scale(runSpecId))`) resulting in new tasks being started in most cases almost immediately.
Do you want to wait for the periodic scale check here?

zen-dog: Also definitely missing explicit logging on what instance Ids are killed here.

zen-dog:
> a) starts a never-ending-deployment (1 agent, hostname UNIQUE, 2 instances)

on the second thought - I don't think this step is necessary. We're testing there a code path that doesn't exist anymore, so why should we? 🤔

timcharper:
The change looks great to me; I would like to see a couple of manual tests spelled out (and, run) to check this change in behavior. Can we add a few to the linked JIRA ticket?

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
meichstedt: As mentioned above (on the old code):
> A scale check will be initiated for each task that is reportedly terminal so the schedulerActions.scale(runSpecId) call here was pretty redundant.

To be precide: after the instance has been killed and is expunged, the `ScaleAppUpdateStepImpl` will send a `ScaleRunSpec` to the schedulerActor.

timcharper:
Commandeering this.

meichstedt:
Should be reconsidered once we have instance goals

PR #6051, opened 2018-02-27
Implement gpu_scheduling_behavior parameter and basic functionality

meln1k:
Summary: Implemented a new GPU-aware offer matching logic, command line flags, RAML examples and unit test

JIRA issues: MARATHON-8088

### DIFF
--- a/docs/docs/rest-api/public/api/v2/examples/queue.json
+++ b/docs/docs/rest-api/public/api/v2/examples/queue.json
@@ -145,9 +145,9 @@
             "processed": 23
           },
           {
-            "reason": "InsufficientGpus",
+            "reason": "DeclinedScarceResources",
             "declined": 0,
-            "processed": 23
+            "processed": 11
           }
         ],
         "rejectSummaryLaunchAttempt": [
@@ -192,9 +192,9 @@
             "processed": 46
           },
           {
-            "reason": "InsufficientGpus",
+            "reason": "DeclinedScarceResources",
             "declined": 0,
-            "processed": 46
+            "processed": 31
           }
         ]
       },
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -335,6 +335,7 @@ trait MarathonConf
     name = "gpu_scheduling_behavior",
     descr = "Defines how offered GPU resources should be treated. Possible settings are `undefined`, `restricted` and " +
       "`unrestricted`",
+    noshort = true,
     default = Some(GpuSchedulingBehavior.Undefined),
     validate = validateGpuSchedulingBehavior
   )
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -317,5 +317,26 @@ trait MarathonConf
       "window start time. This is only evaluated if `maintenance_mode` is in the set of `enable_features`!",
     default = Some(0)
   )
+
+  private[this] def validateGpuSchedulingBehavior(setting: String): Boolean = {
+    val allowedSettings = Set(GpuSchedulingBehavior.Undefined, GpuSchedulingBehavior.Restricted, GpuSchedulingBehavior.Unrestricted)
+    require(
+      isFeatureSet(Features.GPU_RESOURCES) || setting == GpuSchedulingBehavior.Undefined,
+      "gpu_resources must be set in order to use gpu_scheduling_behavior"
+    )
+    require(
+      allowedSettings.contains(setting),
+      s"Setting $setting is invalid. Valid settings are ${allowedSettings.mkString(", ")}"
+    )
+    true
+  }
+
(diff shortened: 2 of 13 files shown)


src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -317,5 +317,26 @@ trait MarathonConf
...
+      "gpu_resources must be set in order to use gpu_scheduling_behavior"
+    )
+    require(
+      allowedSettings.contains(setting),
+      s"Setting $setting is invalid. Valid settings are ${allowedSettings.mkString(", ")}"
+    )
+    true
+  }
+
+  lazy val gpuSchedulingBehavior = opt[String](
```
alenkacz: should we maybe do also `noshort = true` here?

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -236,13 +248,40 @@ object ResourceMatcher extends StrictLogging {
...
+          }
+          true
+
+        case GpuSchedulingBehavior.Restricted =>
+          val noPersistentVolumeToMatch = PersistentVolumeMatcher.matchVolumes(offer, reservedInstances).isEmpty
+          if (gpuResourcesAreWasted && noPersistentVolumeToMatch) {
+            noOfferMatchReasons += NoOfferMatchReason.DeclinedScarceResources
+            false
+          } else {
+            true
```
alenkacz: should we maybe log something here? That we are launching on GPU node while restricted but we are doing that because of persistent volume?

src/test/scala/mesosphere/marathon/raml/PodStatusConversionTest.scala
```diff
@@ -407,6 +407,30 @@ class PodStatusConversionTest extends UnitTest {
           "persistent#volume#5425cbaa-8fd3-45f0-afa4-74ef4fcc594b")))
     }
 
+    "persistent pod launched, received RUNNING status from Mesos, task endpoint health looks great" in {
```
alenkacz: how is this test related to GPUs?

src/main/scala/mesosphere/mesos/NoOfferMatchReason.scala
```diff
@@ -14,6 +14,7 @@ object NoOfferMatchReason {
   case object UnfulfilledConstraint extends NoOfferMatchReason
   case object AgentMaintenance extends NoOfferMatchReason
   case object NoCorrespondingReservationFound extends NoOfferMatchReason
+  case object DeclinedScarceResources extends NoOfferMatchReason
```
alenkacz: you have to add this to the reasonfunnel otherwise it won't appear in the API. I've been there and it sucks and is easy to miss...

Maybe it would be nice to add a test that verifies that it appears in API, then we would see this...

src/test/scala/mesosphere/marathon/raml/PodStatusConversionTest.scala
meln1k: It's something from another changelist :) good catch!

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
meln1k: Design doc says nothing about it. @timcharper what do you think?

alenkacz: I did not mean warn or error, I would log just info... But @timcharper feel free to comment :)

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -236,13 +248,42 @@ object ResourceMatcher extends StrictLogging {
...
+          }
+          true
+
+        case GpuSchedulingBehavior.Restricted =>
+          val noPersistentVolumeToMatch = PersistentVolumeMatcher.matchVolumes(offer, reservedInstances).isEmpty
+          if (gpuResourcesAreWasted && noPersistentVolumeToMatch) {
+            noOfferMatchReasons += NoOfferMatchReason.DeclinedScarceResources
+            false
+          } else {
+            addOnMatch(() => logger.warn(s"Runspec [${runSpec.id}] doesn't require any GPU resources but " +
```
alenkacz: I would prefer to log this as info

docs/docs/rest-api/public/api/v2/examples/queue.json
```diff
@@ -145,9 +145,9 @@
             "processed": 23
           },
           {
-            "reason": "InsufficientGpus",
+            "reason": "DeclinedScarceResources",
```
ichernetsky: Should the examples with `InsufficientGpus` be kept instead of being deleted?

alenkacz: they were in the example two times before this change ... I actually wondered the same when reviewing :)

meln1k: yep, it seems it was a copy paste error before

PR #6052, opened 2018-02-27
 Application specific override for GPU scheduling behavior

meln1k:
https://github.com/mesosphere/marathon/pull/6051 needs to be merged first!

 Application specific override for GPU scheduling behavior

Summary: Implemented an application-specific GPU scheduling behavior override

JIRA issues: MARATHON-8089

### DIFF
--- a/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
+++ b/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
@@ -250,7 +250,7 @@ object ResourceMatcher extends StrictLogging {
 
     val checkGpuSchedulingBehaviour: Boolean = {
       val applicationSpecificGpuBehavior = runSpec.labels.get("GPU_SCHEDULING_BEHAVIOR")
-        .filter(behavior => Set(GpuSchedulingBehavior.Restricted, GpuSchedulingBehavior.Unrestricted).contains(behavior))
+        .filter(behavior => validBehaviors.contains(behavior))
       val availableGPUs = groupedResources.getOrElse(Resource.GPUS, Nil).foldLeft(0.0)(_ + _.getScalar.getValue)
       val gpuResourcesAreWasted = availableGPUs > 0 && runSpec.resources.gpus == 0
       applicationSpecificGpuBehavior.getOrElse(conf.gpuSchedulingBehavior()) match {
@@ -576,4 +576,6 @@ object ResourceMatcher extends StrictLogging {
           s"Not all basic resources satisfied: $basicResourceString")
     }
   }
+
+  private val validBehaviors = Set(GpuSchedulingBehavior.Restricted, GpuSchedulingBehavior.Unrestricted)
 }
--- a/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
+++ b/src/main/scala/mesosphere/mesos/ResourceMatcher.scala
@@ -249,9 +249,11 @@ object ResourceMatcher extends StrictLogging {
     }
 
     val checkGpuSchedulingBehaviour: Boolean = {
+      val applicationSpecificGpuBehavior = runSpec.labels.get("GPU_SCHEDULING_BEHAVIOR")
+        .filter(behavior => Set(GpuSchedulingBehavior.Restricted, GpuSchedulingBehavior.Unrestricted).contains(behavior))
       val availableGPUs = groupedResources.getOrElse(Resource.GPUS, Nil).foldLeft(0.0)(_ + _.getScalar.getValue)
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
```diff
@@ -1179,6 +1179,60 @@ class ResourceMatcherTest extends UnitTest with Inside {
...
+        "--draining_seconds", "300",
+        "--gpu_scheduling_behavior", "unrestricted",
+        "--enable_features", "gpu_resources")
+      val offer = MarathonTestHelper.makeBasicOffer(gpus = 4)
+        .build()
+      val app = AppDefinition(
+        id = "/test".toRootPath,
+        resources = Resources(cpus = 1.0, mem = 128.0, disk = 0.0, gpus = 0),
+        portDefinitions = PortDefinitions(0, 0),
+        labels = Map("GPU_SCHEDULING_BEHAVIOR" -> "restricted")
```
jeschkies: I think we are missing tests for no label. What do you think about using a property table?

```scala
Table(
    ("gpu_scheduling_behavior", "GPU_SCHEDULING_BEHAVIOR", expected),
    ("unrestricted", Some("restricted"), Match),
    ("unrestricted", None, Match),
    ("restricted", Some("unrestricted"), NoMatch),
    ("restricted", None, NoMatch),
)
```

This seems easier to comprehend to me.

src/main/scala/mesosphere/mesos/ResourceMatcher.scala
```diff
@@ -249,9 +249,11 @@ object ResourceMatcher extends StrictLogging {
     }
 
     val checkGpuSchedulingBehaviour: Boolean = {
+      val applicationSpecificGpuBehavior = runSpec.labels.get("GPU_SCHEDULING_BEHAVIOR")
+        .filter(behavior => Set(GpuSchedulingBehavior.Restricted, GpuSchedulingBehavior.Unrestricted).contains(behavior))
```
alenkacz: do we want to create this set for every check? Maybe we should create that on the ResourceMatcher level at least?

alenkacz: do we want to create this set for every check? Maybe we should create that on the ResourceMatcher level at least?

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
```diff
@@ -1179,6 +1180,46 @@ class ResourceMatcherTest extends UnitTest with Inside {
...
+          "--gpu_scheduling_behavior", gpuSchedulingBehavior,
+          "--enable_features", "gpu_resources")
+        val offer = MarathonTestHelper.makeBasicOffer(gpus = 4)
+          .build()
+        val app = AppDefinition(
+          id = "/test".toRootPath,
+          resources = Resources(cpus = 1.0, mem = 128.0, disk = 0.0),
+          portDefinitions = PortDefinitions(0, 0),
+
+          labels = overrideLabel.map(label => Map("GPU_SCHEDULING_BEHAVIOR" -> label)).getOrElse(Map.empty)
```
jeschkies: Nit: Is `overrideLabel.map( ... Map)` not returning a `Option[Map]`?

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
```diff
@@ -1179,6 +1180,46 @@ class ResourceMatcherTest extends UnitTest with Inside {
...
+
+      val overrideCases = Table(
+        ("gpu_scheduling_behavior", "GPU_SCHEDULING_BEHAVIOR", "expected"),
+        ("unrestricted", Some("restricted"), "NoMatch"),
+        ("unrestricted", None, "Match"),
+        ("restricted", Some("unrestricted"), "Match"),
+        ("restricted", None, "NoMatch")
+      )
+
+      forAll(overrideCases) { (gpuSchedulingBehavior, overrideLabel, expected) =>
```
jeschkies: Nit: I'd like to use these cases outside of the `.. should ... in` pattern so we create new cases. It's not a blocker though.

src/test/scala/mesosphere/mesos/ResourceMatcherTest.scala
meln1k: it does!

jeschkies: Oh, I missed the `getOeElse` part.

PR #6054, opened 2018-02-28
Do not start more instances than needed

ichernetsky:
When determining how many instances to launch during deployments
and scale checks, do account for `isReserved` instances, otherwise
Marathon starts extra instances in case of apps and pods with
persistent volumes.

It is a cherry-pick of 311b176 with fixes of flaky tests.

JIRA issues: MARATHON-7751


ichernetsky:
Hmm. Two test timeouts in a row.

ichernetsky:
Rebased.

PR #6056, opened 2018-03-01
Use given matched resources when building Mesos protobufs for pods

ichernetsky:
Previously, passed resources were ignored, and we just constructed new
resource protobufs based on a pod definition.

JIRA issues: [MARATHON-8110](https://jira.mesosphere.com/browse/MARATHON-8110)

### DIFF
--- a/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
+++ b/src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
@@ -43,7 +43,7 @@ object TaskGroupBuilder extends StrictLogging {
     resourceMatch: ResourceMatcher.ResourceMatch,
     volumeMatchOption: Option[PersistentVolumeMatcher.VolumeMatch]
   ): (mesos.ExecutorInfo, mesos.TaskGroupInfo, Seq[Option[Int]]) = {
-    verifyResources(podDefinition, resourceMatch.resources)
+    val packedResources = packResources(podDefinition, resourceMatch.resources)
 
     val allEndpoints = podDefinition.containers.flatMap(_.endpoints)
 
@@ -52,7 +52,7 @@ object TaskGroupBuilder extends StrictLogging {
 
     val executorInfo = computeExecutorInfo(
       podDefinition,
-      resourceMatch.resources,
+      packedResources,
       resourceMatch.portsMatch,
       mesosNetworks,
       volumeMatchOption,
@@ -74,7 +74,7 @@ object TaskGroupBuilder extends StrictLogging {
           val portAssignments = computePortAssignments(podDefinition, endpoints)
 
           val task = computeTaskInfo(container, podDefinition, offer, instanceId, taskId,
-            resourceMatch.resources, resourceMatch.hostPorts, config, portAssignments)
+            packedResources, resourceMatch.hostPorts, config, portAssignments)
             .setDiscovery(taskDiscovery(podDefinition, endpoints))
           task.build
       }.asJava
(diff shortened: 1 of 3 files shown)


ichernetsky:
I ran `test_pod_with_persistent_volume_recovers` in a loop for ~12 hours. No single failure.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -83,6 +86,31 @@ object TaskGroupBuilder extends StrictLogging {
...
   }
 
+  // This method just double-checks that the matched resources are exactly
+  // what is needed and in right quantities.
+  @SuppressWarnings(Array("ComparingFloatingPointTypes"))
+  private[this] def verifyResources(pod: PodDefinition, matchedResources: Seq[mesos.Resource]): Unit = {
+    assert(matchedResources.count(_.getName == "cpus") <= 1)
+    assert(matchedResources.count(_.getName == "mem") <= 1)
+    assert(matchedResources.count(_.getName == "disk") <= 1)
+    assert(matchedResources.count(_.getName == "gpus") <= 1)
```
jeschkies: I wonder if we could define resources that it's impossible to have multiple definitions. I'm having a similar issue with the Rust client for Mesos.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -43,13 +43,16 @@ object TaskGroupBuilder extends StrictLogging {
...
+    verifyResources(podDefinition, resourceMatch.resources)
+
     val allEndpoints = podDefinition.containers.flatMap(_.endpoints)
 
     val mesosNetworks = buildMesosNetworks(
       podDefinition.networks, allEndpoints, resourceMatch.hostPorts, config.mesosBridgeName)
 
     val executorInfo = computeExecutorInfo(
       podDefinition,
+      resourceMatch.resources,
```
jeschkies: Ah, so this is the fix?

ichernetsky: The fix is how `scalarResource` is changed and how it is used now.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: I don't really know :(

timcharper: In the case of disk resources, it is definitely possible to have multiple matched resources (there are different types of disk resources: root, path, mount)

ichernetsky: This is not about persistent volumes.

ichernetsky: I talked to Greg about it, and you are right. If more disks are specified using `--resources` flag, then Mesos can send an offer with two or more disks.

timcharper: Could this code be run against a launch attempt for a pod with persistent volumes? Sorry, probably lacking some context here, but I don't fully see how persistent volumes aren't potentially a factor here.

ichernetsky: Persistent volume resources are passed separately.
```
    volumeMatchOption: Option[PersistentVolumeMatcher.VolumeMatch]
```

ichernetsky:
Rebased.

ichernetsky:
An unrelated failure.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -83,6 +86,189 @@ object TaskGroupBuilder extends StrictLogging {
...
+      }
+    }
+
+    private def consume(
+      resources: List[(BigDecimal, mesos.Resource)],
+      quantity: BigDecimal): Option[(List[(BigDecimal, mesos.Resource)], mesos.Resource)] =
+      if (quantity <= BigDecimal(0)) {
+        None
+      } else {
+        resources.partition(_._1 != quantity) match {
```
jeschkies: What is `_._1`? How about making this explicit?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -83,6 +86,189 @@ object TaskGroupBuilder extends StrictLogging {
...
+    // Find a GPU resource for the given quantity if possible, and remove the resource from the resource list.
+    def consumeGpus(quantity: Double): Option[mesos.Resource] = {
+      consume(gpus, BigDecimal(quantity)).map {
+        case (resources, resource) =>
+          gpus = resources
+          resource
+      }
+    }
+
+    private def consume(
```
jeschkies: I'm not really sure what this method is doing.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -83,6 +86,189 @@ object TaskGroupBuilder extends StrictLogging {
...
+    private var cpus = initial.cpus
+    private var mem = initial.mem
+    private var disk = initial.disk
+    private var gpus = initial.gpus
+
+    // Find a CPU resource for the given quantity if possible, and remove the resource from the resource list.
+    def consumeCpus(quantity: Double): Option[mesos.Resource] = {
+      consume(cpus, BigDecimal(quantity)).map {
+        case (resources, resource) =>
+          cpus = resources
```
jeschkies: This is strange to me. Why are we modifying the state here? How about calling `resource` `consumed_resource`.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -83,6 +86,189 @@ object TaskGroupBuilder extends StrictLogging {
...
+            Some((left ::: right, pair._2))
+        }
+      }
+  }
+
+  private[this] def binPackResources(pod: PodDefinition, matchedResources: Seq[mesos.Resource]): Resources = {
+    verifyMatchedResources(pod, matchedResources)
+
+    val decimalZero = BigDecimal(0)
+    val cpuQuantities = {
```
jeschkies: Are these the required quantities?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
```diff
@@ -83,6 +86,189 @@ object TaskGroupBuilder extends StrictLogging {
...
+          case (_, Nil) =>
+            throw new IllegalStateException("failed to find a resource with the given quantity")
+          case (left, pair :: right) =>
+            // return a new resource list not containing the found resource and the found resource itself
+            Some((left ::: right, pair._2))
+        }
+      }
+  }
+
+  private[this] def binPackResources(pod: PodDefinition, matchedResources: Seq[mesos.Resource]): Resources = {
```
jeschkies: Are the resources from one offer only? Why is it a sequence?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: I have renamed both variables. Why does it look strange to you?

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: I have added a comment.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: Done.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: Mesos folks said that an offer can have multiple resources of one kind (say, `disk`).

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
ichernetsky: These are aggregated defined/requested quantities in a pod definition.

src/main/scala/mesosphere/mesos/TaskGroupBuilder.scala
jeschkies: Ah, thanks for clarifying.

PR #6065, opened 2018-03-07
Use sliding average window on /metrics endpoint

wavesoft:
This PR introduces the `SlidingAverageSnapshot` class that can be used to compute the average values for the metrics presented in the `/metrics` endpoint.

Related JIRA: [MARATHON-8104](https://jira.mesosphere.com/browse/MARATHON-8104)

## More Details

Kamon publishes the `TickMetricSnapshot` at configurable intervals (with the `kamon.metric.tick-interval ` parameter), containing the current snapshot of all metrics. Marathon implements the `/metrics` endpoint by plugging a custom reporter.

Currently we are merging all these snapshots into a single `TickMetricSnapshot` since the beginning of time. This PR changes this by introducing a `SlidingAverageSnapshot` class that is used by our custom reporter to keep track of only the last X snapshots.

This is implemented using a ring buffer where the snapshots are stored, effectively creating a "Sliding average window" for the last X snapshots. To configure the size of the window the user should tweak the `kamon.metric.tick-interval` parameter and the `metrics_averaging_window` parameter.

### DIFF
--- a/docs/docs/command-line-flags.md
+++ b/docs/docs/command-line-flags.md
@@ -329,6 +329,9 @@ The Web Site flags control the behavior of Marathon's web site, including the us
     Expose the execution time per method via the metrics endpoint (/metrics) using code instrumentation.
     Enabling this might noticeably degrade performance but it helps finding performance problems.
     These measurements can be disabled with --disable_metrics. Other metrics are not affected.
+* <span class="label label-default">v1.6.0</span> `--metrics_averaging_window` (Optional. Default: 30 seconds):
+    Configure the size of the sliding average window that is used to compute the values for the `/metrics` endpoint. Note that this value
+    should be at least double the value of the `kamon.metric.tick-interval` parameter, that is 1 second by default.
 * <span class="label label-default">v0.13.0</span> `--reporter_graphite` (Optional. Default: disabled):
     Report metrics to [Graphite](http://graphite.wikidot.com) (StatsD) as defined by the given URL.
     Example: `udp://localhost:2003?prefix=marathon-test&interval=10`
--- a/src/main/scala/mesosphere/marathon/Main.scala
+++ b/src/main/scala/mesosphere/marathon/Main.scala
@@ -1,6 +1,7 @@
 package mesosphere.marathon
 
 import java.net.URI
+import java.time
 
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri
@@ -116,7 +117,12 @@ class MarathonApp(args: Seq[String]) extends AutoCloseable with StrictLogging {
     }
 
     val injector = Guice.createInjector(modules.asJava)
-    Metrics.start(injector.getInstance(classOf[ActorSystem]))
+    Metrics.start(
+      injector.getInstance(classOf[ActorSystem]),
+      time.Duration.ofSeconds(
+        cliConf.averagingWindow.get.getOrElse(60L)
+      )
+    )
     val services = Seq(
       injector.getInstance(classOf[MarathonHttpService]),
       injector.getInstance(classOf[MarathonSchedulerService]))
--- a/src/main/scala/mesosphere/marathon/Main.scala
+++ b/src/main/scala/mesosphere/marathon/Main.scala
@@ -117,12 +117,7 @@ class MarathonApp(args: Seq[String]) extends AutoCloseable with StrictLogging {
     }
 
     val injector = Guice.createInjector(modules.asJava)
-    Metrics.start(
-      injector.getInstance(classOf[ActorSystem]),
-      time.Duration.ofSeconds(
-        cliConf.averagingWindow.get.getOrElse(60L)
-      )
-    )
+    Metrics.start(injector.getInstance(classOf[ActorSystem]), cliConf)
     val services = Seq(
       injector.getInstance(classOf[MarathonHttpService]),
       injector.getInstance(classOf[MarathonSchedulerService]))
--- a/src/main/scala/mesosphere/marathon/metrics/Metrics.scala
+++ b/src/main/scala/mesosphere/marathon/metrics/Metrics.scala
@@ -175,9 +175,11 @@ object Metrics {
   def snapshot(): TickMetricSnapshot = metrics
 
   // Starts collecting snapshots.
-  def start(actorRefFactory: ActorRefFactory, averagingWindow: Duration): Done = {
+  def start(actorRefFactory: ActorRefFactory, config: MetricsReporterConf): Done = {
     class SubscriberActor() extends Actor {
-      val slidingAverageSnapshot: SlidingAverageSnapshot = new SlidingAverageSnapshot(averagingWindow)
+      val slidingAverageSnapshot: SlidingAverageSnapshot = new SlidingAverageSnapshot(
+        Duration.ofSeconds(config.averagingWindow.get.getOrElse(60L))
+      )
 
       override def receive: Actor.Receive = {
         case snapshot: TickMetricSnapshot =>
--- a/src/main/scala/mesosphere/marathon/metrics/Metrics.scala
+++ b/src/main/scala/mesosphere/marathon/metrics/Metrics.scala
@@ -178,7 +178,7 @@ object Metrics {
   def start(actorRefFactory: ActorRefFactory, config: MetricsReporterConf): Done = {
     class SubscriberActor() extends Actor {
       val slidingAverageSnapshot: SlidingAverageSnapshot = new SlidingAverageSnapshot(
-        Duration.ofSeconds(config.averagingWindow.get.getOrElse(30L))
+        Duration.ofSeconds(config.averagingWindowSizeSeconds.get.getOrElse(30L))
       )
 
       override def receive: Actor.Receive = {
--- a/src/main/scala/mesosphere/marathon/metrics/Metrics.scala
+++ b/src/main/scala/mesosphere/marathon/metrics/Metrics.scala
@@ -4,7 +4,8 @@ package metrics
 import akka.Done
 import akka.actor.{ Actor, ActorRef, ActorRefFactory, Props }
 import akka.stream.scaladsl.Source
-import java.time.Clock
+import java.time.{ Clock, Duration }
+
 import kamon.Kamon
 import kamon.metric.SubscriptionsDispatcher.TickMetricSnapshot
 import kamon.metric.instrument.Histogram.DynamicRange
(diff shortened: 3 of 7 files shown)


src/main/scala/mesosphere/marathon/metrics/Metrics.scala
```diff
@@ -174,16 +175,15 @@ object Metrics {
   def snapshot(): TickMetricSnapshot = metrics
 
   // Starts collecting snapshots.
-  def start(actorRefFactory: ActorRefFactory): Done = {
+  def start(actorRefFactory: ActorRefFactory, averagingWindow: Duration): Done = {
```
jeschkies: Hm, currently we pass conf objects. This makes it easier to add other metrics later on or fall back to Typesafe Conf. It seems there is a `MetricsReporterConf`. Why not pass that one?

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,133 @@
...
+import kamon.metric.SubscriptionsDispatcher.TickMetricSnapshot
+import kamon.metric.instrument.CollectionContext
+import kamon.util.{ MapMerge, MilliTimestamp }
+import mesosphere.marathon.metrics.Metrics.subscribe
+
+/**
+  * Calculates sliding average entity snapshot
+  */
+class SlidingAverageSnapshot(val averagingWindow: Duration) {
+  private val logger: Logger = Logger.getLogger(getClass.getName)
```
jeschkies: Please use `StrictLogger`.

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+    * The current summarised snapshot
+    */
+  private var currentSnapshot: TickMetricSnapshot = TickMetricSnapshot(MilliTimestamp.now, MilliTimestamp.now, Map.empty)
+
+  /**
+    * A ring buffer that collects the snapshots
+    */
+  private var averageRing: Array[TickMetricSnapshot] = ringFactory(Kamon.config)
```
jeschkies: How about [`org.apache.commons.collections.buffer.CircularFifoBuffer`](https://commons.apache.org/proper/commons-collections/javadocs/api-3.2.2/org/apache/commons/collections/buffer/CircularFifoBuffer.html).

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,133 @@
...
+    */
+  private def combineRingMetrics: TickMetricSnapshot = {
+    var rangeFrom: MilliTimestamp = MilliTimestamp.now
+    var rangeTo: MilliTimestamp = MilliTimestamp.now
+    var combined: Option[Map[Entity, EntitySnapshot]] = None
+
+    // Start processing all the values of the ring buffer in incrementing index
+    // starting fro the oldest (the next value from the current index) to the
+    // newest (the current index).
+    val maxItems: Int = averageRing.length
```
jeschkies: I guess with the Apache ring buffer this is something like

```scala
buffer.toArray.sortBy(...).fold() { (accumulated, next) =>
   MapMerge.Syntax(accumulated).merge(next ....)
}
```

what do you think?

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,133 @@
...
+    averageRing.update(averageIndex, snapshot)
+    averageIndex = (averageIndex + 1) % averageRing.length
+
+    // Update current snapshot and return it
+    currentSnapshot = combineRingMetrics
+    currentSnapshot
+  }
+
+  /**
+    * Collect the current snapshot
```
jeschkies: What do you mean with `collect`?

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+    * Rebuilds the average ring, based on the current configuration
+    */
+  private def ringFactory(config: Config): Array[TickMetricSnapshot] = {
+    val tickInterval: Long = config.getDuration("kamon.metric.tick-interval").toMillis
+    val fullFrames: Int = (averagingWindow.toMillis.toFloat / tickInterval).ceil.toInt
+
+    // Warn for too low values
+    if (fullFrames < 3) {
```
meln1k: Let's make this magic number a bit more obvious :)

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,133 @@
...
+  private def combineRingMetrics: TickMetricSnapshot = {
+    var rangeFrom: MilliTimestamp = MilliTimestamp.now
+    var rangeTo: MilliTimestamp = MilliTimestamp.now
+    var combined: Option[Map[Entity, EntitySnapshot]] = None
+
+    // Start processing all the values of the ring buffer in incrementing index
+    // starting fro the oldest (the next value from the current index) to the
+    // newest (the current index).
+    val maxItems: Int = averageRing.length
+    for (i: Int <- 0 until maxItems) {
```
meln1k: What do you think about using a while loop here? should be much faster.

src/main/scala/mesosphere/marathon/metrics/Metrics.scala
wavesoft: Yup, that works. Good idea 👍

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
wavesoft: So on every tick we sort the list? That doesn't sound attractive 😄

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
wavesoft: Yup, good idea 👍

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
wavesoft: I will add some comment above 😄

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
wavesoft: I will definitely have a look 👍

However I wanted something quite simple. (Note that the `ringFactory` is a porting artifact from my Kamon `1.0` implementation. There the config can change during the execution and I was using the factory to create the new ring).

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
wavesoft: Whoops. I mean return.

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
jeschkies: You can still have the factory but I'd us an existing implementation.

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
jeschkies: You don't have to. I was just not sure why you need max. I thought you might want to accumulate in order. If the number of ticks is low sorting will probably not be a big impact.

wavesoft: The list is always sorted by date. That's the part that I am interested. I guess `CircularFifoBuffer` does not sort anything.

docs/docs/command-line-flags.md
```diff
@@ -329,6 +329,9 @@ The Web Site flags control the behavior of Marathon's web site, including the us
     Expose the execution time per method via the metrics endpoint (/metrics) using code instrumentation.
     Enabling this might noticeably degrade performance but it helps finding performance problems.
     These measurements can be disabled with --disable_metrics. Other metrics are not affected.
+* <span class="label label-default">v1.6.0</span> `--metrics_averaging_window` (Optional. Default: 30 seconds):
```
jeschkies: Nit: It should be 1.6.x.

src/main/scala/mesosphere/marathon/metrics/MetricsReporterConf.scala
```diff
@@ -16,5 +16,12 @@ trait MetricsReporterConf extends ScallopConf {
     descr = "URL to dogstatsd agent. e.g. udp://localhost:8125?prefix=marathon-test&tags=marathon&interval=10",
     noshort = true
   )
+
+  lazy val averagingWindow = opt[Long](
```
jeschkies: Nit: I think we encode the unit in the option name, e.g. `averagingWindowSizeSeconds`. I might be wrong though.

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
jeschkies: Ok, after our discussion I think it's fine to use the custom ring buffer. Querying the first and last element is easier in your solution.

src/main/scala/mesosphere/marathon/metrics/SlidingAverageSnapshot.scala
```diff
@@ -0,0 +1,163 @@
...
+              rangeTo = inst.to
+              Some(MapMerge.Syntax(current).merge(inst.metrics, (l, r) => l.merge(r, collectionContext)))
+            }
+          }
+        }
+      }
+
+      // Advance index of while loop
+      i += 1
+    }
```
jeschkies: Thanks for the docs. I get it now 🙂. How about encapsulating the iteration in an iterator?

```scala

/**
 * Iterates from oldest to news item in a ring buffer.
 */
class RingBufferIterator(buffer: Array[TickMetricSnapshot]) extends Iterator[TickMetricSnapshot] {
  var i: Int = 0
  val maxItems: Int = buffer.length
   
  def getNextIndext() = (averageIndex + i) % maxItems

  override def hasNext(): Boolean = {
    val idx = getNextIndext()
    val inst = averageRing(idx)
    if (inst == null) return false
    if (idx == averageIndex) return false // we made a full round
    return true
  }

  override def next(): TickMetricSnapshot = {
    val idx = getNextIndext()
    i += 1
    averageRing(idx)
  }
}
```

The averaging would then be

```scala
val iter = RingBufferIterator(averageRing)
val oldest = iter.next()
val rangeFrom = oldest.from

iter.foldLeft( (oldest.metrics, oldest.to)) { case ((combined, _), next) =>
  val updated = MapMerge.Syntax(next).merge(combined.metrics, (l, r) => l.merge(r, collectionContext))
  (updated, next.to)
}
```

docs/docs/command-line-flags.md
wavesoft: If I am right, the labels there indicate the version _since_ this is available, so I think that "since v.1.6.0" makes sense. Otherwise "v1.6" is closer to your suggestion, but I have a feeling that "v1.6.x" is a bit off from the rest of the document formatting.

src/main/scala/mesosphere/marathon/metrics/MetricsReporterConf.scala
wavesoft: That's a good idea 👍

wavesoft:
The time has come 🤜 

Merge

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -1,6 +1,7 @@
 package mesosphere.marathon
 
 import java.net.URI
+import java.time
```
zen-dog: I'm pretty sure this is an `Unused import` ;)

PR #6067, opened 2018-03-08
Added MoM-EE-1.6 to our SI tests

zen-dog:
- added 1.6 definitions (copied from 1.5 with minimal changed)
- removed 1.3 definitions and tests since it supports has run out.


tests/system/fixtures/mom-ee-disabled-1.6.json
```diff
@@ -21,24 +21,17 @@
...
       "forcePullImage": false
     },
-    "volumes": [
-      {
-        "containerPath": "/opt/mesosphere",
-        "hostPath": "/opt/mesosphere",
-        "mode": "RO"
-      }
-    ]
+    "volumes": []
```
ichernetsky: Maybe just delete it altogether?

zen-dog:
> In some places just marathon-user-ee is used, in some other places there is a version appended to it. It would be nice to have it consistent.

only the mom-ee Id is without a version - everything else is with it. Interestingly enough we didn't have the versions at all at the beginning but there is an admin-router bug that needs this workaround.

PR #6077, opened 2018-03-09
Removed old python scale tests

zen-dog:
since we moved away from them a while ago to the new performance test driver.


zen-dog:
I'm sorry @kensipe , I understand there is a lot of sentimental value in them. If you want, you can highjack this PR and merge it in your name - you totally deserve it! 😊

PR #6079, opened 2018-03-09
Remove examples folder

zen-dog:
and the documentation on `haproxy-marathon-bridge` since it was deprecated a long time ago.


zen-dog:
> Do we really want to remove the examples? I had some plans to use them in docs.

Which one would you want to use? From what I can see they are at least 2-3 years old except for 2 group examples which were added 7 months ago.

ichernetsky:
I would just put examples (not necessarily these ones, of course) into the docs, instead of referencing the ones in the repo.

alenkacz:
also good source of examples are jsons that we use for SI tests... cc. @meln1k

PR #6083, opened 2018-03-12
Publish plugin-interface with release script.

jeschkies:
Summary:
The current release script has a few issues. The format of the version is not properly
documented. It should be`1.6.322-2bf46b341`. The plugin interface release step
was not implemented.

JIRA issues: MARATHON_EE-1877.

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -44,7 +44,7 @@ ansiColor('xterm') {
 
               targets = ''
               if (params.dryRun == false)
-                targets = 'docker-tag linux-packages jar-artifact s3-package ' + (params.latest ? 'docker-latest' : '')
+                targets = 'docker-tag linux-packages jar-artifact s3-package plugin-interface ' + (params.latest ? 'docker-latest' : '')
 
               sh """sudo -E env ci/pipeline release --version $params.version --run-tests=$params.runTests $targets"""
             }
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -7,9 +7,9 @@ ansiColor('xterm') {
       parameters([
         string(name: 'version',
           defaultValue: '',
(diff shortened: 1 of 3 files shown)


ci/pipeline
```diff
@@ -283,6 +283,8 @@ def release(args: String*): Unit = {
     case ReleaseTarget.JARArtifact =>
       // publishing to the nexus repository.  This artifact is used by metronome.
       %('sbt, "publish")
+    case ReleaseTarget.PluginInterface =>
+      %('sbt, "plugin-interface/publish")
```
alenkacz: ❤️

Jenkinsfile.release
```diff
@@ -7,9 +7,9 @@ ansiColor('xterm') {
       parameters([
         string(name: 'version',
           defaultValue: '',
-          description: """Community Release Version (ex. 1.5.2).
+          description: """Community Release Version (ex. 1.6.322-2bf46b341).
```
alenkacz: I think that community version is without git hash. As you noted, maybe just the property/description is not correct

jeschkies: Done. Thanks for the feedback.

Jenkinsfile.release
```diff
@@ -44,9 +43,9 @@ ansiColor('xterm') {
 
               targets = ''
               if (params.dryRun == false)
-                targets = 'docker-tag linux-packages jar-artifact s3-package plugin-interface' + (params.latest ? 'docker-latest' : '')
+                targets = 'docker-tag linux-packages jar-artifact s3-package plugin-interface ' + (params.latest ? 'docker-latest' : '')
```
alenkacz: don't you want to put the trailing space in front of docker-latest instead of at the end of the plugin-interface? Might be a little bit cleaner for a case when there is no param latest - although it will probably work in both cases.

PR #6085, opened 2018-03-12
Basic working set of Kibana visualizations

timcharper:
This commit introduces a basic set of useful visualizations and Kibana, and some instructions for how to load them.


timcharper:
Also titled, "the commit in which @zen-dog fully runs out of excuses" 😆 🤣  :trollface:

PR #6088, opened 2018-03-13
GPU is not part of 322 build

alenkacz:
JIRA issues: MARATHON-8128


timcharper:
Good catch, @meln1k !

PR #6089, opened 2018-03-13
Disable Connection Pooling for HTTP Health Checks

meln1k:
Summary: Http().singleRequest is replaced by connection-level api.

JIRA issues: MARATHON-7940

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
@@ -22,6 +22,7 @@ import mesosphere.util.ThreadPoolContext
 
 import scala.concurrent.Future
 import scala.concurrent.duration.FiniteDuration
+import scala.util.control.NonFatal
 import scala.util.{ Failure, Success }
 
 class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with StrictLogging {
@@ -111,6 +112,10 @@ class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with Stri
           logger.debug(s"Health check for instance=${instance.instanceId} responded with ${response.status}")
           Some(Unhealthy(instance.instanceId, instance.runSpecVersion, response.status.toString()))
         }
+      }.recover {
+        case NonFatal(e) =>
+          logger.debug(s"Health check for instance=${instance.instanceId} did not respond due to ${e.getMessage}.")
+          Some(Unhealthy(instance.instanceId, instance.runSpecVersion, e.getMessage))
       }
   }
 
@@ -156,7 +161,11 @@ class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with Stri
           logger.debug(s"Health check for ${instance.instanceId} responded with ${response.status}")
           Some(Unhealthy(instance.instanceId, instance.runSpecVersion, response.status.toString()))
         }
-      }
+    }.recover {
+      case NonFatal(e) =>
+        logger.debug(s"Health check for instance=${instance.instanceId} did not respond due to ${e.getMessage}.")
+        Some(Unhealthy(instance.instanceId, instance.runSpecVersion, e.getMessage))
+    }
   }
 
   def singleRequest(httpRequest: HttpRequest, timeout: FiniteDuration)(implicit mat: Materializer): Future[HttpResponse] = {
--- a/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
@@ -6,6 +6,7 @@ import java.security.cert.X509Certificate
 import javax.net.ssl.{ KeyManager, SSLContext, X509TrustManager }
 
 import akka.actor.{ Actor, PoisonPill }
+import akka.http.scaladsl.settings.ClientConnectionSettings
 import akka.http.scaladsl.client.RequestBuilding
 import akka.http.scaladsl.model.{ HttpRequest, HttpResponse }
 import akka.http.scaladsl.{ ConnectionContext, Http }
@@ -20,6 +21,7 @@ import mesosphere.marathon.util.Timeout
 import mesosphere.util.ThreadPoolContext
 
 import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
 import scala.util.{ Failure, Success }
 
 class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with StrictLogging {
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
```diff
@@ -163,6 +154,32 @@ class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with Stri
       }
     }
   }
+
+  def singleRequest(httpRequest: HttpRequest)(implicit mat: Materializer): Future[HttpResponse] = {
+    if (httpRequest.uri.scheme.equalsIgnoreCase("https")) {
```
wavesoft: Do you think it makes sense to replace this check with an additional argument `isSecure: bool`?
Since we are calling the `singleRequest` method from `http` and `https` we already know in advance if it's HTTPS or not.

meln1k: I don't know, additional argument will create more opportunities to mess something up :)

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
```diff
@@ -163,6 +154,32 @@ class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with Stri
...
+        s.loose.withAcceptAnyCertificate(true)
+          .withAllowLegacyHelloMessages(Some(true))
+          .withAllowUnsafeRenegotiation(Some(true))
+          .withAllowWeakCiphers(true)
+          .withAllowWeakProtocols(true)
+          .withDisableHostnameVerification(true)
+          .withDisableSNI(true)
+      })
+      val authority = httpRequest.uri.authority
+      val connectionFlow = Http().outgoingConnectionHttps(
```
jeschkies: Do we have to create the flow each time? How about creating it once?

src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActorTest.scala
```diff
@@ -83,5 +87,41 @@ class HealthCheckWorkerActorTest extends AkkaUnitTest with ImplicitSender {
...
+
+      val route =
+        path("health") {
+          get {
+            promise.success("success")
+            complete(StatusCodes.OK)
+          }
+        }
+
+      val bindingFuture = Http().bindAndHandle(route, "localhost", 8080)
```
zen-dog: Nö, please take `PortAllocator.ephemeralPort()`

src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActorTest.scala
```diff
@@ -83,5 +87,41 @@ class HealthCheckWorkerActorTest extends AkkaUnitTest with ImplicitSender {
...
+        val t: Task = TestTaskBuilder.Helper.runningTaskForApp(appId)
+        val hostPorts = Seq(8080)
+        t.copy(status = t.status.copy(networkInfo = NetworkInfo(hostName, hostPorts, ipAddresses = Nil)))
+      }
+      val instance = LegacyAppInstance(task, agentInfo, UnreachableStrategy.default())
+
+      val ref = system.actorOf(Props(classOf[HealthCheckWorkerActor], mat))
+      ref ! HealthCheckJob(app, instance, MarathonHttpHealthCheck(port = Some(8080), path = Some("/health")))
+
+      val res = Await.result(promise.future, 10.seconds)
```
jeschkies: How about `promise.future.futureValue`?

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
```diff
@@ -139,20 +142,8 @@ class HealthCheckWorkerActor(implicit mat: Materializer) extends Actor with Stri
...
-        .withAllowWeakCiphers(true)
-        .withAllowWeakProtocols(true)
-        .withDisableHostnameVerification(true)
-        .withDisableSNI(true)
-    })
-
-    Timeout(check.timeout)(Http().singleRequest(
-      RequestBuilding.Get(url),
-      connectionContext = ConnectionContext.https(disabledSslContext, sslConfig = Some(disabledSslConfig))
+    Timeout(check.timeout)(singleRequest(
```
zen-dog: Do we really need `Timeout(...)` wrapper here? Can we use `idleTimeout` below directly?

src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActorTest.scala
```diff
@@ -83,5 +87,41 @@ class HealthCheckWorkerActorTest extends AkkaUnitTest with ImplicitSender {
...
+      val hostName = "localhost"
+      val appId = PathId("/test_id")
+      val app = AppDefinition(id = appId, portDefinitions = Seq(PortDefinition(0)))
+      val agentInfo = AgentInfo(host = hostName, agentId = Some("agent"), region = None, zone = None, attributes = Nil)
+      val task = {
+        val t: Task = TestTaskBuilder.Helper.runningTaskForApp(appId)
+        val hostPorts = Seq(8080)
+        t.copy(status = t.status.copy(networkInfo = NetworkInfo(hostName, hostPorts, ipAddresses = Nil)))
+      }
+      val instance = LegacyAppInstance(task, agentInfo, UnreachableStrategy.default())
```
zen-dog: we should try to avoid using the old `LegacyAppInstance`.

src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActorTest.scala
zen-dog: ` Await.result` is banned from test. Use `futureValue` and implicitely the `patienceConfig`

src/test/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActorTest.scala
meln1k: sure

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
meln1k: Sure!

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckWorkerActor.scala
meln1k: Actually, `Http()` is an akka extention, and it is only created once during the first use, and it's an actor system wide one. Once it's initialized, then it's instance will always be returned when we call `Http()`, so there is no need to put it in a separate val. The connection requires address and port, and since we're connecting to absolutely different nodes, we need to create it every time and can't reuse it unfortunately (BTW this actor is also short-lived, and it will die after the health check is performed).

zen-dog:
Don't forget to backport @meln1k

PR #6090, opened 2018-03-13
Add support for symlinks for `--plugin_dir` and `--plugin_conf` parameters

zen-dog:
by using `getCanonicalFile` method on the File object internally. `getCanonicalFile` first converts this pathname to absolute form if necessary, and then maps it to its unique form in a system-dependent way. Among other things this typically involves resolving symbolic links (on UNIX platforms).

JIRA issues: MARATHON-8131

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/plugin/impl/PluginManagerImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/plugin/impl/PluginManagerImpl.scala
@@ -5,14 +5,14 @@ import java.io.File
 import java.net.{ URL, URLClassLoader }
 import java.util.ServiceLoader
 
+import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/plugin/impl/PluginManagerImpl.scala
```diff
@@ -26,8 +26,7 @@ private[plugin] class PluginManagerImpl(
     val config: MarathonConf,
     val definitions: PluginDefinitions,
     val urls: Seq[URL],
-    val crashStrategy: CrashStrategy) extends PluginManager {
-  private[this] val log: Logger = LoggerFactory.getLogger(getClass)
+    val crashStrategy: CrashStrategy) extends PluginManager with StrictLogging {
```
jeschkies: Boyscout Rule! Awesome!

PR #6095, opened 2018-03-15
Fix native package tests

timcharper:
* For the moment, target Mesos 1.5.0-2.0.2 for Debian tests
* Disable Ubuntu upstart package tests
* Also, return nonzero exit status of any of the tests fail

JIRA issues: MARATHON-8129

### DIFF
--- a/tests/package/Makefile
+++ b/tests/package/Makefile
@@ -20,6 +20,9 @@ debian-systemv/mesos-version: target/mesos-version
 debian-systemd/mesos-version: target/mesos-version
 	if [ $$(cat target/mesos-version) = "1.5.0-2.0.1" ]; then echo "1.5.0-2.0.2" > $@; else echo "Package version patch should not be needed; remove this step from the Makefile"; exit 1; fi
 
+ubuntu-upstart/mesos-version: target/mesos-version
+	if [ $$(cat target/mesos-version) = "1.5.0-2.0.1" ]; then echo "1.5.0-2.0.2" > $@; else echo "Package version patch should not be needed; remove this step from the Makefile"; exit 1; fi
+
 target/mesos-version: ../../project/Dependencies.scala
 	mkdir -p target
 	cat ../../project/Dependencies.scala | grep MesosDebian | cut -f 2 -d '"' > $@.tmp
@@ -47,9 +50,9 @@ target/centos-systemd.built: centos-systemd/Dockerfile centos-systemd/mesos-vers
 	cd centos-systemd && docker build . -t marathon-package-test:centos-systemd
 	touch $@
 
-# target/ubuntu-upstart.built: ubuntu-upstart/Dockerfile ubuntu-upstart/mesos-version
-# 	cd ubuntu-upstart && docker build . -t marathon-package-test:ubuntu-upstart
-# 	touch $@
+target/ubuntu-upstart.built: ubuntu-upstart/Dockerfile ubuntu-upstart/mesos-version
+	cd ubuntu-upstart && docker build . -t marathon-package-test:ubuntu-upstart
+	touch $@
 
 test: | all
 	amm test.sc all
--- a/tests/package/Makefile
+++ b/tests/package/Makefile
@@ -57,4 +57,4 @@ target/ubuntu-upstart.built: ubuntu-upstart/Dockerfile ubuntu-upstart/mesos-vers
 test: | all
 	amm test.sc all
 
-all: target/mesos.built target/debian-systemv.built target/debian-systemd.built target/centos-systemd.built target/centos-systemv.built # target/ubuntu-upstart.built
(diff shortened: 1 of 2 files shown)


tests/package/test.sc
```diff
@@ -458,5 +470,11 @@ def main(args: String*): Unit = {
       ???
   }
 
-  tests.filter { t => predicate(name(t))}.foreach(run(_))
+  val testsToRun = tests.filter { t => predicate(name(t))}
+  testsToRun.foreach(run(_))
+  val results = testsToRun.map { t => Await.result(t.result, 1.hour) }
+  if (!results.forall(_ == true)) {
```
zen-dog: Since it's `"There were errors"` kind of logic may be `if (results.exists(_ == false))` would be better?

tests/package/Makefile
```diff
@@ -12,6 +12,14 @@ clean:
 %/mesos-version: target/mesos-version
 	cp target/mesos-version $@
 
+# 1.5.0-2.0.1 package has issues for Debian, and 1.5.0-2.0.2 isn't released for all distros. When we move past this,
+# remove these build steps
+debian-systemv/mesos-version: target/mesos-version
```
zen-dog: Is there a ticket somewhere to remove those? I fear it will be forgotten...

timcharper: I feared the same thing. It's a little bit hard to read (had to put it on one line because of makefile's), but, the build will fail if this workaround is still present and the Mesos version is no longer `1.5.0-2.0.1`.
```
if [ $$(cat target/mesos-version) = "1.5.0-2.0.1" ]; then
  echo "1.5.0-2.0.2" > $@
else
  echo "Package version patch should not be needed; remove this step from the Makefile"
  exit 1
fi
```

tests/package/test.sc
timcharper: Well, maybe is too compact, but this actually does what I want.

```
@ Nil.forall(_ == false)
res0: Boolean = true
```

Although, granted, we should probably show an error message if no tests match the provided string. But, I think this is fine for now :-)

tests/package/Makefile
zen-dog: Oh, right! Or we could insert s `sleep` and make each subsequent test run slower until someone notices :D

jeschkies: I have to admit that I don't quite get these steps. Why are we echoing `1.5.0-2.0.2` when the version is `1.5.0-2.0.1` even though `1.5.0-2.0.2` is not released?

tests/package/Makefile
```diff
@@ -46,4 +57,4 @@ target/ubuntu-upstart.built: ubuntu-upstart/Dockerfile ubuntu-upstart/mesos-vers
 test: | all
 	amm test.sc all
 
-all: target/mesos.built target/debian-systemv.built target/debian-systemd.built target/centos-systemd.built target/centos-systemv.built target/ubuntu-upstart.built
+all: target/mesos.built target/debian-systemv.built target/debian-systemd.built target/centos-systemd.built target/centos-systemv.built # target/ubuntu-upstart.built
```
jeschkies: Why is this commented?

tests/package/Makefile
jeschkies: What do you think about writing `1.5.0.-2.0.2` in `Dependencies.scala` for `MesosDebian` and fall back to `1.5.0-2.0.1` if `1.5.0-2.0.2` does not exist.

tests/package/test.sc
```diff
@@ -14,7 +16,17 @@ abstract class UnitTest extends WordSpec with GivenWhenThen with Matchers with E
 
 case class Container(containerId: String, ipAddress: String)
 
-trait MesosTest extends UnitTest with BeforeAndAfterAll {
+trait FailureWatcher extends Suite {
+  var _result = Promise[Boolean]
```
ichernetsky: `resultPromise`?

tests/package/Makefile
timcharper: Great point, that was left over

tests/package/Makefile
timcharper: 1.5.0-2.0.2 is only released for the Debian and the Ubuntu 14 Linux distributions. It is not available for Centos. Historically, the same package version has been released for all distributions of Linux.

I have a preference to keep it as is; it solves the immediate use case and has a built-in mechanism to make sure we remove this workaround in the future when it is no longer needed

timcharper: Your comment made me realize that we do not have a package dependency for Marathon RPM packages. This is something that we should probably fix. Click add single comment

tests/package/test.sc
meln1k: Should we print the errors here?

PR #6097, opened 2018-03-16
Cleanup instance tracker.

jeschkies:
Summary:
The instance tracker would hold its state as part of the receive method.
With this change it will just be another field in the actor. We also cleanup
`InstancesBySpec` and `SpecInstances` a little.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
@@ -78,25 +78,31 @@ object InstanceTracker {
   object InstancesBySpec {
 
     def of(specInstances: collection.immutable.Map[PathId, InstanceTracker.SpecInstances]): InstancesBySpec = {
-      new InstancesBySpec(specInstances.withDefault(appId => InstanceTracker.SpecInstances(appId)))
+      new InstancesBySpec(specInstances.withDefault(appId => InstanceTracker.SpecInstances()))
     }
 
+    def forInstances(instances: Seq[Instance]): InstancesBySpec = forInstances(instances: _*)
     def forInstances(instances: Instance*): InstancesBySpec = of(
-      instances.groupBy(_.runSpecId).map { case (appId, appInstances) => appId -> SpecInstances.forInstances(appId, appInstances.to[Seq]) }
+      instances
+        .groupBy(_.runSpecId)
+        .map {
+          case (appId, appInstances) =>
+            val instancesById: Map[Instance.Id, Instance] = appInstances.map(instance => instance.instanceId -> instance)(collection.breakOut)
+            appId -> SpecInstances(instancesById)
+        }
     )
 
     def empty: InstancesBySpec = of(collection.immutable.Map.empty[PathId, InstanceTracker.SpecInstances])
   }
   /**
-    * Contains only the tasks of the app with the given app ID.
+    * Contains only the instances of a specific app.
     *
-    * @param specId The id of the app.
     * @param instanceMap The instances of the app by their id.
     */
-  case class SpecInstances(specId: PathId, instanceMap: Map[Instance.Id, Instance] = Map.empty) {
+  case class SpecInstances(instanceMap: Map[Instance.Id, Instance] = Map.empty) {
 
     def isEmpty: Boolean = instanceMap.isEmpty
-    def contains(taskId: Instance.Id): Boolean = instanceMap.contains(taskId)
+    def contains(instanceId: Instance.Id): Boolean = instanceMap.contains(instanceId)
     def instances: Seq[Instance] = instanceMap.values.to[Seq]
 
     private[tracker] def withInstance(instance: Instance): SpecInstances =
@@ -104,10 +110,5 @@ object InstanceTracker {
 
     private[tracker] def withoutInstance(instanceId: Instance.Id): SpecInstances =
       copy(instanceMap = instanceMap - instanceId)
-  }
-
-  object SpecInstances {
-    def forInstances(pathId: PathId, instances: Seq[Instance]): SpecInstances =
-      SpecInstances(pathId, instances.map(instance => instance.instanceId -> instance)(collection.breakOut))
   }
 }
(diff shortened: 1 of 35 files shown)


src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -97,13 +101,19 @@ private[impl] class InstanceTrackerActor(
   override def receive: Receive = initializing
 
   private[this] def initializing: Receive = LoggingReceive.withLabel("initializing") {
-    case appTasks: InstanceTracker.InstancesBySpec =>
+    case initialInstances: InstanceTracker.InstancesBySpec =>
       logger.info("Task loading complete.")
 
+      instancesBySpec = initialInstances
+      counts = TaskCounts(initialInstances.allInstances, healthStatuses = Map.empty)
```
jeschkies: I guess we could make the task counts part if the `InstencesBySpec` and the also update the metrics there. What do you think?

ichernetsky: `the InstencesBySpec and the` — what comes after the second `the`?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -97,13 +101,19 @@ private[impl] class InstanceTrackerActor(
...
+
+      // this is run on any state change
```
ichernetsky: I believe this one should be removed.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -25,12 +25,7 @@ private[tracker] class InstancesLoaderImpl(repo: InstanceRepository)(implicit va
       instances <- Future.sequence(names.map(repo.get)).map(_.flatten)
     } yield {
       log.info(s"Loaded ${instances.size} tasks")
-      val instancesByRunSpec = instances.groupBy(_.runSpecId)
-      val map = instancesByRunSpec.map {
-        case (runSpecId, specInstances) => runSpecId ->
-          InstanceTracker.SpecInstances.forInstances(runSpecId, specInstances)
```
alenkacz: why did we have to it like this previously?

jeschkies: This was just code duplication that got in the base with the infamous `Team` merge for pods.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
jeschkies: Ah, it should be `and then also update the metrics there`.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
```diff
@@ -78,27 +78,31 @@ object InstanceTracker {
...
+            val instancesById: Map[Instance.Id, Instance] = appInstances.map(instance => instance.instanceId -> instance)(collection.breakOut)
+            appId -> SpecInstances(instancesById)
+        }
     )
 
     def empty: InstancesBySpec = of(collection.immutable.Map.empty[PathId, InstanceTracker.SpecInstances])
   }
   /**
-    * Contains only the tasks of the app with the given app ID.
+    * Contains only the instances of a specific app.
```
zen-dog: nite: `s/specific app/specific RunSpec`

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -97,13 +101,18 @@ private[impl] class InstanceTrackerActor(
   override def receive: Receive = initializing
 
   private[this] def initializing: Receive = LoggingReceive.withLabel("initializing") {
-    case appTasks: InstanceTracker.InstancesBySpec =>
+    case initialInstances: InstanceTracker.InstancesBySpec =>
       logger.info("Task loading complete.")
```
zen-dog: nit:`s/Task loading/Instances...`

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
```diff
@@ -78,27 +78,31 @@ object InstanceTracker {
...
     *
-    * @param specId The id of the app.
-    * @param instanceMap The tasks of this app by task ID. FIXME: change keys to Task.TaskID
+    * @param instanceMap The instances of the app by their id.
     */
-  case class SpecInstances(specId: PathId, instanceMap: Map[Instance.Id, Instance] = Map.empty) {
+  case class SpecInstances(instanceMap: Map[Instance.Id, Instance] = Map.empty) {
```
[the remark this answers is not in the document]
jeschkies: I was thinking about removing them. However, the have a few helper methods for construction. So I'd rather follow up with a removal.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
jeschkies: Yes. However, if these classes are removed that would not be an option.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
jeschkies: ✅

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
jeschkies: ✅

PR #6100, opened 2018-03-16
SystemConfig is the Correct Security Constraint

kensipe:
SystemConfig is the Correct Security Constraint

Summary:
Removing use of SystemMetrics.  SystemMetrics is not a configurable security constraint.
Switching back to SystemConfig which is.  https://jira.mesosphere.com/browse/MARATHON_EE-1884

JIRA issues:  MARATHON_EE-1884

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/SystemResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/SystemResource.scala
@@ -12,7 +12,7 @@ import com.typesafe.config.{ Config, ConfigRenderOptions }
 import com.typesafe.scalalogging.StrictLogging
 import com.wix.accord.Validator
 import mesosphere.marathon.metrics.Metrics
-import mesosphere.marathon.plugin.auth.AuthorizedResource.{ SystemConfig, SystemMetrics }
+import mesosphere.marathon.plugin.auth.AuthorizedResource.SystemConfig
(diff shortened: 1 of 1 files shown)


jdef:
"SystemMetrics is not a configurable security constraint" -- what does this mean?

jdef:
I'll elaborate. Dan probably wants system configuration permission for his various ops tasks. He installs a third party metrics tool to grab metrics from Marathon for analysis. That tool shouldn't have access to Marathon configuration, it should ONLY have access to Marathon's metrics. This is impossible to do if SystemConfig is used as the same permission check for both actions.

Why this change vs. fixing marathon-dcos-plugins?

meichstedt:
@jdef What you say makes sense, we should consider that for 1.7

1.5 onwards was broken because a new permission `SystemMetrics` was introduced without ever being supported by the plugins. This effectively was a breaking change that was discovered by users that could no longer access metrics with their previous config. The safe way to introduce this IMO would've been to let them opt-in to the more fine grained permission schema instead of breaking the access.

kensipe:
@jdef I apologize... I missed you comment.   I still think this is the right approach for now.    There is bunch of coordination which includes getting the UI team involved in order to add this.    We are looking to have fine grain security controls as part of MT.    We will definitely need to make changes to the plugins at that time.

PR #6102, opened 2018-03-16
elk config json writing target; parameterized target path

timcharper:
- Also, logstash runner with auto shutdown

### DIFF
--- a/tools/elk/bin/run-logstash
+++ b/tools/elk/bin/run-logstash
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+logstash "$@" &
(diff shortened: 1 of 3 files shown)


tools/elk/bin/run-logstash
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+logstash "$@" &
```
alenkacz: should we update the readme with starting logstash with this script?

timcharper: Maybe? :) It's not necessary and there's a chance this script may not work in some environments.

PR #6105, opened 2018-03-19
Stabilize `MesosAppIntegrationTest`

zen-dog:
One of the tests was removing the whole root group which is not good since the tests
are running in parallel.


src/test/scala/mesosphere/marathon/integration/DockerAppIntegrationTest.scala
```diff
@@ -15,9 +15,17 @@ class DockerAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathon
   // FIXME (gkleiman): Docker tests don't work under Docker Machine yet. So they can be disabled through an env variable.
   val envVar = "RUN_DOCKER_INTEGRATION_TESTS"
 
-  //clean up state before running the test case
-  after(cleanUp())
+  // This suite would some time time out in `beforeAll` call because mesos agent would crash during docker initialisation
```
jeschkies: Nit "time time out"

PR #6108, opened 2018-03-20
Rework Mesos client

timcharper:
We remove the mutable connection info and frameworkId, and instead
return a `Source[MesosClient, NotUsed]`, similar to how Akka HTTP client returns responses with a streamable entity.

The connection logic consumes the initial `Subscribed` event, the data for which is available in the `MesosClient` instance. The further Mesos events are available via a materializable-once source `mesosClient.mesosSource`.

### DIFF
--- a/mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
+++ b/mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
@@ -1,0 +1,235 @@
+package mesosphere.mesos.client
+
+import akka.util.ByteString
+import com.google.protobuf
+import org.apache.mesos.v1.mesos.{ AgentID, ExecutorID, FrameworkID, Filters, KillPolicy, OfferID, Request, TaskID }
+import org.apache.mesos.v1.scheduler.scheduler.Call.{ Accept, Acknowledge, Decline, Kill, Message, Reconcile, Revive }
+import org.apache.mesos.v1.scheduler.scheduler.{ Call, Event }
+
+class MesosCallFactory(frameworkId: FrameworkID) {
+  private val someFrameworkId = Some(frameworkId)
+  /**
+    * ***************************************************************************
+    * Helper methods to create mesos `Call`s
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#calls
+    * ***************************************************************************
+    */
+
+  /**
+    * Factory method to construct a TEARDOWN Mesos Call event. Calling this method has no side effects.
+    *
+    * This event is sent by the scheduler when it wants to tear itself down. When Mesos receives this request it will
+    * shut down all executors (and consequently kill tasks). It then removes the framework and closes all open
+    * connections from this scheduler to the Master.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#teardown
+    */
+  def teardown(): Call = {
+    Call(
+      frameworkId = someFrameworkId,
+      `type` = Some(Call.Type.TEARDOWN)
+    )
+  }
+
+  /**
+    * Factory method to construct a ACCEPT Mesos Call event. Calling this method has no side effects.
+    *
+    * Sent by the scheduler when it accepts offer(s) sent by the master. The ACCEPT request includes the type
+    * of operations (e.g., launch task, launch task group, reserve resources, create volumes) that the scheduler
+    * wants to perform on the offers. Note that until the scheduler replies (accepts or declines) to an offer,
+    * the offer’s resources are considered allocated to the offer’s role and to the framework.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#accept
+    */
+  def accept(accepts: Accept): Call = {
+    Call(
+      frameworkId = someFrameworkId,
+      `type` = Some(Call.Type.ACCEPT),
+      accept = Some(accepts))
+  }
+
+  /**
+    * Factory method to construct a DECLINE Mesos Call event. Calling this method has no side effects.
+    *
+    * Sent by the scheduler to explicitly decline offer(s) received. Note that this is same as sending an ACCEPT
+    * call with no operations.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#decline
+    */
+  def decline(offerIds: Seq[OfferID], filters: Option[Filters] = None): Call = {
+    Call(
+      frameworkId = someFrameworkId,
+      `type` = Some(Call.Type.DECLINE),
+      decline = Some(Decline(offerIds = offerIds, filters = filters))
+    )
+  }
+
+  /**
+    * Factory method to construct a REVIVE Mesos Call event. Calling this method has no side effects.
+    *
+    * Sent by the scheduler to remove any/all filters that it has previously set via ACCEPT or DECLINE calls.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#revive
+    */
+  def revive(role: Option[String] = None): Call = {
+    Call(
+      frameworkId = someFrameworkId,
+      `type` = Some(Call.Type.REVIVE),
+      revive = Some(Revive(role = role))
+    )
+  }
+
+  /**
+    * Factory method to construct a SUPPRESS Mesos Call event. Calling this method has no side effects.
+    *
+    * Suppress offers for the specified roles. If `roles` is empty the `SUPPRESS` call will suppress offers for all
+    * of the roles the framework is currently subscribed to.
+    *
+    * http://mesos.apache.org/documentation/latest/upgrades/#1-2-x-revive-suppress
+    */
+  def suppress(roles: Option[String] = None): Call = {
+    Call(
+      frameworkId = someFrameworkId,
+      `type` = Some(Call.Type.SUPPRESS),
+      suppress = Some(Call.Suppress(roles))
+    )
+  }
+
+  /**
+    * Factory method to construct a KILL Mesos Call event. Calling this method has no side effects.
+    *
+    * Sent by the scheduler to kill a specific task. If the scheduler has a custom executor, the kill is forwarded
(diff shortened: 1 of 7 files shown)


timcharper:
The memos-test integration tests now pass

mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
```diff
@@ -0,0 +1,235 @@
...
+    * Factory method to construct a ACCEPT Mesos Call event. Calling this method has no side effects.
+    *
+    * Sent by the scheduler when it accepts offer(s) sent by the master. The ACCEPT request includes the type
+    * of operations (e.g., launch task, launch task group, reserve resources, create volumes) that the scheduler
+    * wants to perform on the offers. Note that until the scheduler replies (accepts or declines) to an offer,
+    * the offer’s resources are considered allocated to the offer’s role and to the framework.
+    *
+    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#accept
+    */
+  def accept(accepts: Accept): Call = {
```
jeschkies: What about passing the parameters that accept takes as done with `decline`?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -354,200 +135,276 @@ trait MesosApi {
...
-    *
-    * http://mesos.apache.org/documentation/latest/scheduler-http-api/#kill
-    */
-  def kill(taskId: TaskID, agentId: Option[AgentID] = None, killPolicy: Option[KillPolicy]): Call = {
-    Call(
-      `type` = Some(Call.Type.KILL),
-      kill = Some(Kill(taskId = taskId, agentId = agentId, killPolicy = killPolicy))
-    )
+    val httpConnection = Http().outgoingConnection(url.getHost, url.getPort)
+    Source.single(request)
```
jeschkies: Nit: A a new line above. I thought this was a continuation of `val httpConnection`.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -354,200 +135,276 @@ trait MesosApi {
...
+    system: ActorSystem, materializer: ActorMaterializer, executionContext: ExecutionContext): RunnableGraph[Future[MesosClient]] = {
+
+    val subscribedWatcher: Sink[Event, Future[Event.Subscribed]] = Flow[Event].collect {
+      case event if event.subscribed.isDefined =>
+        event.subscribed.get
+      case o =>
+        throw new RuntimeException(s"Expected subscribed event, got ${o}")
+    }.toMat(Sink.head)(Keep.right)
+
+    def mesosHttpConnection(url: URI, redirectRetries: Int): Source[(HttpResponse, ConnectionInfo), NotUsed] =
```
jeschkies: How about moving this method out of `def apply`?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -354,200 +135,276 @@ trait MesosApi {
...
+          unzip.out0.via(eventReader) ~> killSwitch ~> eventBroadcast
+          unzip.out1 ~> connectionInfoWatcher
+
+          eventBroadcast.out(0).take(1) ~> subscribedWatchedShape
+          /* we use detach to prevent a deadlock situation. "drop(1)" does not initiate a pull, so we use detach to
+           * preemptively pull an element so that the first broadcast output can receive an element. */
+          eventBroadcast.out(1).drop(1).detach ~> eventsOutputSinkShape
+          ClosedShape
+        }
+      }
```
jeschkies: Unfortunately I don't have enough knowledge to judge this graph. However, why do we need `subscribedWatcher`? Would it not be enough to take the first element? Rust streams have `into_future` which give `Future[(FirstItem, Stream)]`.

mesos-client/src/main/scala/mesosphere/mesos/client/StrictLoggingFlow.scala
```diff
@@ -0,0 +1,10 @@
+package mesosphere.mesos.client
+
+import akka.NotUsed
+import akka.stream.scaladsl.Flow
+import com.typesafe.scalalogging.StrictLogging
+
+// TODO - move somewhere shared
```
jeschkies: What does this mean? Why not do it in this diff?

mesos-client/src/test/scala/mesosphere/mesos/client/MesosClientIntegrationTest.scala
```diff
@@ -24,125 +27,117 @@ class MesosClientIntegrationTest extends AkkaUnitTest
...
       *
-      * @param n number of elements to pull
-      * @return a sequence of elements
+      * @param predicate Function to evaluate to see if event matches
+      * @return matching event, if any
       */
-    def pull(n: Int): Seq[Event] = {
-      (1 to n).map(_ => queue.pull().futureValue.get)
-    }
+    @tailrec final def pullUntil(predicate: Event => Boolean): Option[Event] =
```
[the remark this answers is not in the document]
timcharper: Mesos client will not successfully connect if the first event is not SUBSCRIBED.

Other than that... Why do the tests care if the HEARTBEAT event comes before OFFER, in this order? Isn't that testing Mesos?

mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
```diff
@@ -0,0 +1,235 @@
+package mesosphere.mesos.client
+
+import akka.util.ByteString
+import com.google.protobuf
+import org.apache.mesos.v1.mesos.{ AgentID, ExecutorID, FrameworkID, Filters, KillPolicy, OfferID, Request, TaskID }
+import org.apache.mesos.v1.scheduler.scheduler.Call.{ Accept, Acknowledge, Decline, Kill, Message, Reconcile, Revive }
+import org.apache.mesos.v1.scheduler.scheduler.{ Call, Event }
+
+class MesosCallFactory(frameworkId: FrameworkID) {
```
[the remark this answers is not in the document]
timcharper: It is a factory. What would you prefer to call it?

I was trying to avoid giving the impression that these methods had side effects.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
timcharper: This is a good point; @zen-dog wdyt? I just moved the method from a previous location.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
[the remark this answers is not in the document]
timcharper: Maybe we can go with MesosV1Api and add the word `new` before the methods?

`newTeardown`...

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
timcharper: ✅

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
timcharper: Yeah... this can be improved. I didn't think of it at the time but there is prefixAndTail which will achieve something similar. I'll rework it.

jeschkies: `prefixAndTail` is exactly what I meant 🙂.

timcharper: Glad you said something; prefixAndTail is substantially simpler.

mesos-client/src/main/scala/mesosphere/mesos/client/StrictLoggingFlow.scala
timcharper: I just removed the TODO. I guess we can move it to somewhere shared later but it is fine for now.

mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
```diff
@@ -0,0 +1,235 @@
...
+
+  /**
+    * Factory method to construct a SUPPRESS Mesos Call event. Calling this method has no side effects.
+    *
+    * Suppress offers for the specified roles. If `roles` is empty the `SUPPRESS` call will suppress offers for all
+    * of the roles the framework is currently subscribed to.
+    *
+    * http://mesos.apache.org/documentation/latest/upgrades/#1-2-x-revive-suppress
+    */
+  def suppress(roles: Option[String] = None): Call = {
```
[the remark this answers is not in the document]
timcharper: Fixed it

mesos-client/src/main/scala/mesosphere/mesos/client/MesosCallFactory.scala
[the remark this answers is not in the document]
timcharper: Went with `new` methods...

mesos-client/src/main/scala/mesosphere/mesos/client/MesosClient.scala
```diff
@@ -355,200 +136,243 @@ trait MesosApi {
...
+      .idleTimeout(conf.idleTimeout)
+      .buffer(conf.sourceBufferSize, OverflowStrategy.backpressure)
+
+    httpConnection.flatMapConcat {
+      case (httpResponse, connectionInfo) =>
+        val sharedKillSwitch = KillSwitches.shared(s"MesosClient-${conf.master}")
+        httpResponse.entity.dataBytes.
+          via(eventReader).
+          via(sharedKillSwitch.flow).
+          prefixAndTail(1).
```
jeschkies: Nice! Now I even understand this!

PR #6111, opened 2018-03-20
Fixing flaky MetricsTimerTest

wavesoft:
This commit removes the Kamon configuration test phase from the SlidingAverageSnapshotTest, that
is causing the MetricsTimerTest to be flaky.

_Related JIRA: [MARATHON_EE-1918](https://jira.mesosphere.com/browse/MARATHON_EE-1918)_


zen-dog:
100x runs in the loop are successful?

ichernetsky:
What an approach to get rid of flakiness 👍 :)

PR #6112, opened 2018-03-20
Fix of http marathon healthchecks

meln1k:
Reused some of smart akka-http code related to setting headers and filtering URIs

JIRA issues: MARATHON-8136


zen-dog:
I'd like to see SI health check tests pass on this first

meln1k:
@zen-dog I run the SI tests, here are the results for the healthchecks:
image

PR #6114, opened 2018-03-21
Fixing IT teardown behavior

zen-dog:
Summary:
our teardown was broken on multiple levels (starting with the actual `POST mesos/teardown` call which never worked and ending with our `MesosCluster::teardown` method  which would retry but never wait for the actual teardown to happen.


mesos-simulation/src/test/scala/mesosphere/mesos/scale/SingleAppScalingTest.scala
```diff
@@ -20,6 +20,14 @@ object SingleAppScalingTest {
   val appInfosFile = "scaleUp20s-appInfos"
 }
 
+trait SimulatedMesosTest extends MesosTest {
```
jeschkies: What is the idea of this?

src/test/scala/mesosphere/marathon/integration/setup/MesosTest.scala
```diff
@@ -335,37 +335,18 @@ case class MesosCluster(
...
+    val frameworkIds = facade.frameworkIds().value
 
-trait SimulatedMesosTest extends MesosTest {
-  def mesos: MesosFacade = {
-    require(false, "No access to mesos")
-    ???
+    // Call mesos/teardown for all framework Ids in the cluster and wait for the teardown to complete
+    frameworkIds.foreach { fId =>
+      facade.teardown(fId)
+      eventually(timeout(1.minutes), interval(2.seconds)) { facade.completedFrameworkIds().value.contains(fId) }
```
jeschkies: How about moving the assertion out of the loop?
1. Issue treadown for each framework.
2. Verifiy that all frameworks are down.

src/test/scala/mesosphere/marathon/integration/setup/MesosTest.scala
```diff
@@ -335,37 +335,18 @@ case class MesosCluster(
...
-              s"agent allocated on agent ${agent.id}: " +
-                s"used = ${agent.usedResources}, reserved = ${agent.reservedResourcesByRole}")
-          }
-        }
-      }
-    Await.result(teardown, Duration.Inf)
-  }
-}
+object MesosTest extends Eventually {
+  def teardown(facade: MesosFacade): Unit = {
```
jeschkies: Do we use this only in `trait Mesos`? Cause then we could move the functionality.

mesos-simulation/src/test/scala/mesosphere/mesos/scale/SingleAppScalingTest.scala
zen-dog: ¯\_(ツ)_/¯
I merely moved it to the only place where it's used.

src/test/scala/mesosphere/marathon/integration/setup/MesosTest.scala
zen-dog: 1. happens already
2. not sure what you mean by that. That's what `eventually` is waiting for.

src/test/scala/mesosphere/marathon/integration/setup/MesosTest.scala
zen-dog: You're right. No reason to keep this object around.

zen-dog:
> No leaked processes detected

First time I see this happen after the test ^.^

PR #6117, opened 2018-03-22
Fail when unused imports are detected.

jeschkies:
Summary:
We want to keep our code base clean thus the compiler warns about unused
imports. Unfortunately we cannot fail the compilation just on unused
imports warnings. This change introduces a quick grep on the compile
logs to ensure that no unused imports have been detected.


ci/pipeline
```diff
@@ -49,6 +53,15 @@ def compileAndTest(logFileName: String): Unit = utils.stage("Compile and Test")
   checkSystemIntegrationTests(logFileName)
 }
 
+@main
```
zen-dog: Comments ;)

PR #6119, opened 2018-03-23
Stabilize `MesosAppIntegrationTest` (#6105)

meln1k:
One of the tests was removing the whole root group which is not good since the tests
are running in parallel.

(cherry picked from commit 9b8e6ba)


src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -24,6 +24,8 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
   // run as root in a Linux environment. They have to be explicitly enabled through an env variable.
   val envVar = "RUN_MESOS_INTEGRATION_TESTS"
 
+  val hostnameField = "hostname"
```
zen-dog: hm. what's this? wan't part of the original commit

PR #6121, opened 2018-03-23
Fixing datadog and statsd reporter configuration

wavesoft:
This commit is handling the missing `prefix` URI parameter for the `--reporter_graphite` and `--reporter_datadog` arguments.

It is also adding support for more parameters that the plugins support but we didn't expose already.

Related JIRAs: [MARATHON_EE-1916](https://jira.mesosphere.com/browse/MARATHON_EE-1916) and [MARATHON-8083](https://jira.mesosphere.com/browse/MARATHON-8083)

### DIFF
--- a/src/main/scala/mesosphere/marathon/Main.scala
+++ b/src/main/scala/mesosphere/marathon/Main.scala
@@ -1,21 +1,17 @@
 package mesosphere.marathon
 
-import java.net.URI
-import java.time
-
 import akka.actor.ActorSystem
 import akka.http.scaladsl.model.Uri
 import com.google.common.util.concurrent.ServiceManager
 import com.google.inject.{ Guice, Module }
 import com.typesafe.config.{ Config, ConfigFactory }
 import com.typesafe.scalalogging.StrictLogging
+import scala.concurrent.ExecutionContext.Implicits.global
 import kamon.Kamon
 import mesosphere.chaos.http.HttpModule
 import mesosphere.chaos.metrics.MetricsModule
 import mesosphere.marathon.api.{ MarathonHttpService, MarathonRestModule }
 import mesosphere.marathon.core.CoreGuiceModule
-
-import scala.concurrent.ExecutionContext.Implicits.global
 import mesosphere.marathon.core.base.toRichRuntime
 import mesosphere.marathon.metrics.Metrics
 import mesosphere.marathon.stream.Implicits._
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -52,29 +53,64 @@ class MarathonApp(args: Seq[String]) extends AutoCloseable with StrictLogging {
...
         val params = url.query()
         s"""
            |kamon.statsd {
            |  hostname: ${url.authority.host}
            |  port: ${if (url.authority.port == 0) 8125 else url.authority.port}
            |  flush-interval: ${params.collectFirst { case ("interval", iv) => iv.toInt }.getOrElse(10)} seconds
-           |}
-         """.stripMargin
+         """.stripMargin +
+          params.foldLeft("")(
```
timcharper: Would this be better with flatMap + mkString ?

```
params.flatMap {
  case ("prefix", value) => Some(s"   simple-metric-key-generator.application: ${value}")
  case ("hostname", value) => Some(...)
  case =>
    logger.warn(...)
    None
}.mkString("\n")

Comment (timcharper) [reply src/main/scala/mesosphere/marathon/Main.scala]: In fact... what do you think about using `ConfigFactory.parseMap` ?

```
@ import $ivy.`com.typesafe:typesafe:1.3.0`

@ import $ivy.`com.typesafe:config:1.3.2`
import $ivy.$

@ import scala.collection.JavaConverters._

@ com.typesafe.config.ConfigFactory.parseMap(Map("a.b" -> 1).asJava)
res2: com.typesafe.config.Config = Config(SimpleConfigObject({"a":{"b":1}}))
```

This would prevent us from having to do any escaping

File: src/main/scala/mesosphere/marathon/Main.scala [hunk 2]
```diff
@@ -52,29 +53,64 @@ class MarathonApp(args: Seq[String]) extends AutoCloseable with StrictLogging {
...
-      |}
-        """.stripMargin
+           |kamon.datadog {
+           |  hostname: ${url.authority.host}
+           |  port: ${if (url.authority.port == 0) 8125 else url.authority.port}
+           |  flush-interval: ${params.collectFirst { case ("interval", iv) => iv.toInt }.getOrElse(10)} seconds
+           |""".stripMargin +
+          params.foldLeft("")(
+            (expr: String, param) => param._1 match {
+              case "prefix" => expr + s"  application-name: ${param._2}\n"
```
timcharper: Seems like it might help to use `ConfigFactory.parseMap` here, too?

src/main/scala/mesosphere/marathon/Main.scala
wavesoft: Oh definitely

src/main/scala/mesosphere/marathon/Main.scala
wavesoft: Yup, I like the approach. Working on it.

PR #6125, opened 2018-03-28
Fixing ubuntu ca-certificates-java bug

wavesoft:
It seems that the default package installation of openjdk has an issue when
creating the ceritifcates files. This commit fixes this problem by re-installing
the certificates package.


ci/pipeline
```diff
@@ -237,6 +237,8 @@ def master(): Unit = {
 def pr(): Unit = asPullRequest {
   provisionHost()
   val version = build()
+  buildDockerAndLinuxPackages()
+  testDockerAndLinuxPackages()
```
jeschkies: Just so we don't forget. This should be removed after the tests pass.

wavesoft:
Rolled back one commit to remove the tests PR. Since CI is green we can merge this 👍

PR #6129, opened 2018-04-03
Use unique config and info files for clusters.

jeschkies:
Summary:
We use unique names for config and info files for our clusters. This
should avoid possible conflicts between two system test runs.


jeschkies:
This did not solve the issue of cluster not being deleted. However, we should still merge this change to be on the safe side.

ci/si_pipeline.sh
```diff
@@ -11,8 +11,10 @@ fi
 CHANNEL="$1"
 VARIANT="$2"
 
+UUID=$(date +%s)
```
zen-dog: I don't think this is needed. The build number is unique enough.

PR #6132, opened 2018-04-03
Send Instance health changed event when the overall status Backport

kensipe:
Back port of #6046 

* Send Instance health changed event when the overall status changed

* Purge health checks statuses of done tasks

Also make some improvements to AppHealthCheckActor.

* Create a benchmark for AppHealthCheckActor

* Optimize the update of new health check results in AppHealthCheckActor

Two optimization have been done here:

* Don't browse the healthChecks map several times in updateHealthCheck method
* Use mutable maps instead of immutable ones

* Simplify the AppHealthCheckActor and factorize the code

* Use genericity for health check purges
* Use pattern matching for readability when possible

* Add more documentation to AppHealthCheckActor

* Fix formatting of MarathonHealthCheckManager and remove _ imports

* isLaunched -> isActive

Making this change compatible with change introduced by 4d20e105d72f5b690e09d206dccc4c595f0b68ec

### DIFF
--- a/benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
+++ b/benchmark/src/main/scala/mesosphere/marathon/core/health/impl/AppHealthCheckBenchmark.scala
@@ -1,0 +1,105 @@
+package mesosphere.marathon
+package core.health.impl
+
+import java.util.concurrent.TimeUnit
+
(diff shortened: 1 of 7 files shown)


timcharper:
Back port seems reasonable to me, given the undesirable behavior that was exhibited before the fix.

PR #6137, opened 2018-04-05
Decrease archived IT sandbox size by excluding `provisioner` folders from the tarball

zen-dog:
Summary:
we exclude `provisioner` folders (containing slave container root file systems) from sandboxes since it is huge (>800Mb) and not useful


zen-dog:
`sandboxes.tar.gz	1.17 MB` 🎉

zen-dog:
There is smth. strange about the `MesosAppIntegrationTest` sandbox - for some reason it has no `slave` folders 🤔

PR #6139, opened 2018-04-06
Set a custom build description.

jeschkies:
Summary:
This is a test to display custom information on the build page.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,6 +23,7 @@ ansiColor('xterm') {
             sh """sudo -E ci/pipeline jenkins"""
         }
       } finally {
+        currentBuild.description = "<h1>Archives</h1><ul><li>logs</li><li>sandbox</li></ul>"
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
         publishHTML([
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -23,6 +23,7 @@ ansiColor('xterm') {
             sh """sudo -E ci/pipeline jenkins"""
         }
       } finally {
+        currentBuild.description = "<h1>Archives</h1><ul><li><a href='${env.BUILD_URL}/Scapegoat_Report'>Scpegoat Report</a></li></ul>"
```
zen-dog: why not just `archive` it in the jenkins file?

jeschkies: I'd like to upload to S3 and post a better summary.

zen-dog: > I'd like to upload to S3 and post a better summary.

This feature is useful if you attach HTML reports (with graphics and whatnot) to the build page. Otherwise, Jenkins archive is sufficient.

jeschkies: There is a nice plugin that would do this for us. Unfortunately there was no reaction to this so far.

jeschkies:
Won't do.

PR #6143, opened 2018-04-09
Assert MesosDNS is returning non-empty Marathon leader ip.

jeschkies:
Summary:
The system test
`test_marathon_root.test_marathon_zk_partition_leader_change` can fail
when we assume a Marathon leader change even though the IP is empty.


jeschkies:
Only `system.test_marathon_root.test_run_app_with_non_existing_user` is failing which is a known issue.

jeschkies:
Only `test_run_app_with_non_existing_user` fails so the system tests pass!

PR #6145, opened 2018-04-09
Pod failure reasons are not visible in the UI/API

meln1k:
Summary: task termination history is shown in the podStatus

JIRA issues: MARATHON-8148


src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -173,6 +173,37 @@ class AppInfoBaseData(
     val statusSince = if (instanceStatus.isEmpty) now else instanceStatus.map(_.statusSince).max
     val state = await(podState(podDef.instances, instanceStatus, isPodTerminating(podDef.id)))
 
+    val taskFailureOpt = await {
```
jeschkies: Could you add a type? That makes following the code easier.

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -173,6 +173,37 @@ class AppInfoBaseData(
...
+      taskFailureRepository
+        .get(podDef.id)
+        .recover { case NonFatal(e) => None }
+    }
+    val failedInstanceBundle = taskFailureOpt.flatMap { taskFailure =>
```
jeschkies: Same here. What type is `taskFailure` and `instance.tasksMap(failedTaskId)`?

src/main/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseData.scala
```diff
@@ -173,6 +173,37 @@ class AppInfoBaseData(
...
+      val failedTaskId = core.task.Task.Id(taskFailure.taskId)
+      instances.collectFirst {
+        case instance if instance.tasksMap.keySet.contains(failedTaskId) =>
```
jeschkies: Why not `instance.tasksMap.contains(failedTaskId)`? `.keySet` will do another copy if I'm not mistaken.

src/test/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseDataTest.scala
```diff
@@ -492,5 +494,33 @@ class AppInfoBaseDataTest extends UnitTest with GroupCreation {
...
     }
+
+    "requesting Pod lastTaskFailure when one exists" in {
+      import concurrent.ExecutionContext.Implicits.global
+      val f = new Fixture
+      Given("A pod definition")
+      val taskFailure = TaskFailureTestHelper.taskFailure
+      f.taskFailureRepository.get(pod.id) returns Future.successful(Some(taskFailure))
+      val instance1 = {
+        val instance = fakeInstance(pod)(f)
```
jeschkies: This seems odd. Why is `fakeInstance` not a member of `Fixture`?

src/test/scala/mesosphere/marathon/core/appinfo/impl/AppInfoBaseDataTest.scala
```diff
@@ -492,5 +494,33 @@ class AppInfoBaseDataTest extends UnitTest with GroupCreation {
...
+      f.marathonSchedulerService.listRunningDeployments() returns Future.successful(Seq.empty)
+
+      When("Getting pod status with last task failures")
+      val podStatus = f.baseData.podStatus(pod).futureValue
+
+      Then("we get the failure in the app info")
+      podStatus.terminationHistory.head.instanceID shouldEqual instance1.instanceId.idString
+
+      And("the taskFailureRepository should have been called to retrieve the failure")
+      verify(f.taskFailureRepository, times(1)).get(pod.id)
```
jeschkies: Why should this be verified? I think this tests an implementation detail and is not required. All we care about is the `podStatus` returned. How that is constructed we do not care.

PR #6146, opened 2018-04-09
Wait five seconds before starting app service.

jeschkies:
Summary:
`test_framework_readiness_time_check` is flaky because the readiness
check might start immediately and succeed before waiting for another
thirty seconds. This change ensures that the first test would fail if it
is within five seconds.

### DIFF
--- a/tests/system/apps/fake-framework.json
+++ b/tests/system/apps/fake-framework.json
@@ -1,6 +1,6 @@
 {
   "id": "/fake-framework",
-  "cmd": "/opt/mesosphere/bin/python -m http.server $PORT0",
+  "cmd": "sleep 5 && /opt/mesosphere/bin/python -m http.server $PORT0",
   "cpus": 0.5,
   "mem": 128,
   "disk": 0,
--- a/tests/system/dcos_service_marathon_tests.py
+++ b/tests/system/dcos_service_marathon_tests.py
@@ -34,11 +34,11 @@ def test_framework_readiness_time_check():
     client = marathon.create_client()
     deployment_id = client.add_app(fw)
 
-    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=16, retry_on_exception=common.ignore_exception)
     def assert_in_deployment(deployment_id):
         deployment = client.get_deployment(deployment_id)
         assert deployment['currentActions'][0]['readinessCheckResults'][0]['ready'] is False, \
-            "The application is ready"
+            "Application's readiness check is green where it should still be red"
 
     assert_in_deployment(deployment_id)
 
--- a/tests/system/dcos_service_marathon_tests.py
+++ b/tests/system/dcos_service_marathon_tests.py
@@ -1,8 +1,10 @@
 """Tests for root marathon specific to frameworks and readinessChecks """
 
 import apps
-import time
+import common
+import retrying
 import shakedown
+import time
 
 from datetime import timedelta
 from dcos import marathon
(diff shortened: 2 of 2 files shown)


tests/system/dcos_service_marathon_tests.py
```diff
@@ -26,19 +28,25 @@ def test_framework_readiness_time_check():
...
 
-    time.sleep(readiness_time - 10)  # not yet.. still deploying
-    deployment = client.get_deployment(deployment_id)
-    assert deployment['currentActions'][0]['readinessCheckResults'][0]['ready'] is False, \
-        "The application is read"
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
+    def assert_in_deployment(deployment_id):
+        deployment = client.get_deployment(deployment_id)
+        assert deployment['currentActions'][0]['readinessCheckResults'][0]['ready'] is False, \
+            "The application is ready"
```
zen-dog: nit: `s/The application is ready/Application's readiness check is green where it should still be red` or smth. along those lines.

tests/system/dcos_service_marathon_tests.py
```diff
@@ -26,19 +28,25 @@ def test_framework_readiness_time_check():
...
     fw['readinessChecks'][0]['intervalSeconds'] = readiness_time
 
     client = marathon.create_client()
     deployment_id = client.add_app(fw)
 
-    time.sleep(readiness_time - 10)  # not yet.. still deploying
-    deployment = client.get_deployment(deployment_id)
-    assert deployment['currentActions'][0]['readinessCheckResults'][0]['ready'] is False, \
-        "The application is read"
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
```
zen-dog: nit: Given current `readiness_time = 15`, `stop_max_attempt_number=15` should be enough

tests/system/apps/fake-framework.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "/fake-framework",
-  "cmd": "/opt/mesosphere/bin/python -m http.server $PORT0",
+  "cmd": "sleep 5 && /opt/mesosphere/bin/python -m http.server $PORT0",
```
alenkacz: I don't get how the various times relates here. Why do we need the sleep here?

alenkacz: hmm maybe I get it now ... I think it would deserve a comment somewhere, what do you think?

tests/system/dcos_service_marathon_tests.py
```diff
@@ -26,19 +28,25 @@ def test_framework_readiness_time_check():
...
     fw['readinessChecks'][0]['intervalSeconds'] = readiness_time
 
     client = marathon.create_client()
     deployment_id = client.add_app(fw)
 
-    time.sleep(readiness_time - 10)  # not yet.. still deploying
-    deployment = client.get_deployment(deployment_id)
-    assert deployment['currentActions'][0]['readinessCheckResults'][0]['ready'] is False, \
-        "The application is read"
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=16, retry_on_exception=common.ignore_exception)
```
alenkacz: do we need to retry 16 times? can't we just wait 16 seconds with wait_fixed=16000 ?

tests/system/dcos_service_marathon_tests.py
```diff
@@ -26,19 +28,25 @@ def test_framework_readiness_time_check():
...
+    def assert_in_deployment(deployment_id):
+        deployment = client.get_deployment(deployment_id)
+        assert deployment['currentActions'][0]['readinessCheckResults'][0]['ready'] is False, \
+            "Application's readiness check is green where it should still be red"
 
-    time.sleep(readiness_time + 1)
-    assert client.get_deployment(deployment_id) is None, "The application is still being deployed"
+    assert_in_deployment(deployment_id)
+
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=30, retry_on_exception=common.ignore_exception)
```
alenkacz: the same here, does it make sense to retry?

tests/system/apps/fake-framework.json
jeschkies: The readiness check might start immediately and the app would become ready.

tests/system/dcos_service_marathon_tests.py
jeschkies: If we fix the wait time it's like a sleep. I want to avoid that.

tests/system/dcos_service_marathon_tests.py
jeschkies: I see the retry as an eventually. The test verifies that the app is not ready and then finishes the deployment. The times do not matter that much since we have unit and integration tests for these things.

PR #6151, opened 2018-04-11
Create a facade for change instance state.

jeschkies:
Summary:
Instead of sending an update operation we provide a facade for the
instance tracker. This should make following the code a little simpler.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/InstanceStateOpProcessor.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/InstanceStateOpProcessor.scala
@@ -1,7 +1,11 @@
 package mesosphere.marathon
 package core.task.tracker
 
+import akka.Done
+import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.instance.update.{ InstanceUpdateEffect, InstanceUpdateOperation }
+import mesosphere.marathon.state.Timestamp
+import org.apache.mesos
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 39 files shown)


src/main/scala/mesosphere/marathon/core/task/tracker/InstanceStateOpProcessor.scala
```diff
@@ -14,4 +18,14 @@ import scala.concurrent.Future
 trait InstanceStateOpProcessor {
   /** Process an InstanceUpdateOperation and propagate its result. */
   def process(stateOp: InstanceUpdateOperation): Future[InstanceUpdateEffect]
```
alenkacz: can we make this private? Or is this still called from outside?

jeschkies: Unfortunately there is still one call deep from withing the launch queue :/.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceStateOpProcessor.scala
```diff
@@ -14,4 +18,14 @@ import scala.concurrent.Future
 trait InstanceStateOpProcessor {
```
zen-dog: Nothing against this diff but `InstanceStateOpProcessor` and later `val stateOpProcessor` are ugly names :sigh:

jeschkies: They'll go in the next PR 🙂.

PR #6154, opened 2018-04-12
 Merge `StateOpProcessor` interface into `InstanceTracker`.

jeschkies:
Summary:
This unifies the interface to the instance tracker. We move closer
to a CRUD interface that hides instance update logic.

This came out of my prototyping of adding instances in a scheduled
state instead of adding a count to the launch queue.


jeschkies:
screen shot 2018-04-12 at 14 39 18

> Less code is best

— Dieter Rams

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceStateOpProcessorDelegateTest.scala
```diff
@@ -1,175 +0,0 @@
-package mesosphere.marathon
```
zen-dog: I don't think you can remove this tests without testing the underlying functionality elsewhere. You'll have to convert tests to test the `InstanceTracker` then.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
```diff
@@ -32,6 +38,19 @@ trait InstanceTracker {
 
   def hasSpecInstancesSync(appId: PathId): Boolean
```
zen-dog: Not necessarily in this diff but those `*Sync` methods are used almost exclusively in tests. Since you're cleaning up, wdyt about removing those completely?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceStateOpProcessorDelegate.scala
```diff
@@ -1,64 +0,0 @@
...
-  */
-private[tracker] class InstanceStateOpProcessorDelegate(
-    clock: Clock,
-    conf: InstanceTrackerConfig,
-    instanceTrackerRef: ActorRef)
-  extends InstanceStateOpProcessor {
-
-  import scala.concurrent.ExecutionContext.Implicits.global
-
-  private[impl] implicit val timeout: Timeout = conf.internalTaskUpdateRequestTimeout().milliseconds
```
zen-dog: I guess we can remove this parameter from the config and use `instanceTrackerQueryTimeout` for all requests within the InstanceTracker?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceStateOpProcessorDelegateTest.scala
jeschkies: Good point. Done.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
jeschkies: And remove the todo. Sure but not in this diff. I'll follow up.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceStateOpProcessorDelegate.scala
jeschkies: Oh. I thought it's the same. Yeah, I'm using `instanceTrackerQueryTimeout` for all. Good catch. If I remove it we technically break the Marathon process options :/.

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
```diff
@@ -1,18 +1,24 @@
...
-import mesosphere.marathon.state.PathId
+import mesosphere.marathon.state.{ PathId, Timestamp }
+import org.apache.mesos
 
 import scala.concurrent.{ ExecutionContext, Future }
 
 /**
-  * The TaskTracker exposes the latest known state for every task.
+  * The InstanceTracker exposes the latest known state for every instance.
+  * The [[InstanceTracker]] exposes the latest known state for every instance and handles the processing of
```
alenkacz: this and previous line seems to be stating the same thing... ?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceStateOpProcessorDelegate.scala
alenkacz: we break on non-existing process options?

jeschkies: > Hm. Technically this config param was hidden and marked for internal use.

We have an option string for internal use?

PR #6156, opened 2018-04-13
Replace `OfferMatcherSpec` with ScalaTest idioms.

jeschkies:
Summary:
This is just a little cleanup and replace a small helper method with
ScalaTest matchers.


src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
```diff
@@ -199,7 +198,7 @@ class LaunchQueueModuleTest extends AkkaUnitTest with OfferMatcherSpec {
       val request = InstanceOpFactory.Request(app, offer, Map.empty, additionalLaunches = 1)
       verify(instanceOpFactory).matchOfferRequest(request)
       matchedTasks.offerId should equal(offer.getId)
-      launchedTaskInfos(matchedTasks) should equal(Seq(mesosTask))
+      matchedTasks.ops.collect{ case e: InstanceOp.LaunchTask => e }.map(_.taskInfo) should contain theSameElementsAs Seq(mesosTask)
```
zen-dog: I like the `should contain theSameElementsAs` usage but `collect`-ing the ops in every test... I don't see how that's an improvement.

jeschkies: Less code but yeah. It annoys me as well.

jeschkies:
Ok, I'm dropping this as it's not the focus right now.

PR #6159, opened 2018-04-16
Remove synchronous instance tracker calls.

jeschkies:
Summary:
This change removes all blocking synchronous calls to the instance
tracker. Actors that used to wait for the instance state start in an
initializing state until they receive the result from the instance
tracker. This way we do not block.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
@@ -29,7 +29,7 @@ class TaskReplaceActor(
     val eventBus: EventStream,
     val readinessCheckExecutor: ReadinessCheckExecutor,
     val runSpec: RunSpec,
-    promise: Promise[Unit]) extends Actor with ReadinessBehavior with StrictLogging {
+    promise: Promise[Unit]) extends Actor with Stash with ReadinessBehavior with StrictLogging {
   import TaskReplaceActor._
 
   // compute all values ====================================================================================
@@ -39,50 +39,31 @@ class TaskReplaceActor(
   // Killed resident tasks are not expunged from the instances list. Ignore
   // them. LaunchQueue takes care of launching instances against reservations
   // first
-  val currentRunningInstances = instanceTracker.specInstancesSync(runSpec.id).filter(_.isActive)
+  var currentRunningInstances: Seq[Instance] = _
 
   // In case previous master was abdicated while the deployment was still running we might have
   // already started some new tasks.
   // All already started and active tasks are filtered while the rest is considered
-  private[this] val (instancesAlreadyStarted, instancesToKill) = {
-    currentRunningInstances.partition(_.runSpecVersion == runSpec.version)
-  }
+  private[this] var instancesAlreadyStarted: Seq[Instance] = _
+  private[this] var instancesToKill: Seq[Instance] = _
 
   // The ignition strategy for this run specification
-  private[this] val ignitionStrategy = computeRestartStrategy(runSpec, currentRunningInstances.size)
+  private[this] lazy val ignitionStrategy: RestartStrategy = computeRestartStrategy(runSpec, currentRunningInstances.size)
 
   // compute all variables maintained in this actor =========================================================
 
   // All instances to kill queued up
-  private[this] val toKill: mutable.Queue[Instance] = instancesToKill.to[mutable.Queue]
+  private[this] lazy val toKill: mutable.Queue[Instance] = instancesToKill.to[mutable.Queue]
 
   // All instances to kill as set for quick lookup
-  private[this] var oldInstanceIds: SortedSet[Id] = instancesToKill.map(_.instanceId).to[SortedSet]
+  private[this] var oldInstanceIds: SortedSet[Id] = _
 
   // The number of started instances. Defaults to the number of already started instances.
-  var instancesStarted: Int = instancesAlreadyStarted.size
+  var instancesStarted: Int = _
 
   override def preStart(): Unit = {
     super.preStart()
-    // subscribe to all needed events
-    eventBus.subscribe(self, classOf[InstanceChanged])
-    eventBus.subscribe(self, classOf[InstanceHealthChanged])
-
-    // reconcile the state from a possible previous run
-    reconcileAlreadyStartedInstances()
-
-    // kill old instances to free some capacity
-    for (_ <- 0 until ignitionStrategy.nrToKillImmediately) killNextOldInstance()
-
-    // start new instances, if possible
-    launchInstances()
-
-    // reset the launch queue delay
-    logger.info("Resetting the backoff delay before restarting the runSpec")
-    launchQueue.resetDelay(runSpec)
-
-    // it might be possible, that we come here, but nothing is left to do
-    checkFinished()
+    instanceTracker.specInstances(runSpec.id).pipeTo(self)
   }
 
   override def postStop(): Unit = {
@@ -90,7 +71,48 @@ class TaskReplaceActor(
     super.postStop()
   }
 
-  override def receive: Receive = readinessBehavior orElse replaceBehavior
+  override def receive: Receive = initializing
+
+  def initializing: Receive = {
+    case instances: Seq[Instance] =>
+
+      // Initialize state
+      currentRunningInstances = instances.filter(_.isActive)
+      instancesAlreadyStarted = currentRunningInstances.filter(_.runSpecVersion == runSpec.version)
+      instancesToKill = currentRunningInstances.filter(_.runSpecVersion != runSpec.version)
+      oldInstanceIds = instancesToKill.map(_.instanceId).to[SortedSet]
+      instancesStarted = instancesAlreadyStarted.size
+
+      // subscribe to all needed events
+      eventBus.subscribe(self, classOf[InstanceChanged])
+      eventBus.subscribe(self, classOf[InstanceHealthChanged])
+
+      // reconcile the state from a possible previous run
+      reconcileAlreadyStartedInstances()
+
+      // kill old instances to free some capacity
+      for (_ <- 0 until ignitionStrategy.nrToKillImmediately) killNextOldInstance()
+
+      // start new instances, if possible
+      launchInstances()
+
+      // reset the launch queue delay
+      logger.info("Resetting the backoff delay before restarting the runSpec")
+      launchQueue.resetDelay(runSpec)
+
+      // it might be possible, that we come here, but nothing is left to do
+      checkFinished()
+
+      context.become(readinessBehavior orElse replaceBehavior)
+      unstashAll()
+
+    case Status.Failure(cause) =>
+      // escalate this failure
+      throw new IllegalStateException("while loading instances", cause)
+
+    case stashMe: AnyRef =>
+      stash()
+  }
 
   def replaceBehavior: Receive = {
     // New instance failed to start, restart it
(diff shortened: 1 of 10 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -109,7 +112,8 @@ private class TaskLauncherActor(
 
     logger.info(s"Started instanceLaunchActor for ${runSpec.id} version ${runSpec.version} with initial count $instancesToLaunch")
 
-    instanceMap = instanceTracker.instancesBySpecSync.instancesMap(runSpec.id).instanceMap
+    val t = instanceTracker.instancesBySpec()
```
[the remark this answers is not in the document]
jeschkies: Ah. Debugging leftovers.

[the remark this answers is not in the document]
jeschkies: ✅

src/test/scala/mesosphere/marathon/core/deployment/impl/DeploymentActorTest.scala
```diff
@@ -122,7 +122,7 @@ class DeploymentActorTest extends AkkaUnitTest with GroupCreation {
       queue.asyncPurge(any) returns Future.successful(Done)
       scheduler.startRunSpec(any) returns Future.successful(Done)
       tracker.specInstances(Matchers.eq(app1.id))(any[ExecutionContext]) returns Future.successful(Seq(instance1_1, instance1_2))
-      tracker.specInstancesSync(app2.id) returns Seq(instance2_1)
+      tracker.specInstances(Matchers.eq(app2.id))(any[ExecutionContext]) returns Future.successful(Seq(instance2_1))
```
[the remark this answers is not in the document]
jeschkies: Did not work for some weird reason.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -91,6 +92,8 @@ private class TaskLauncherActor(
     localRegion: () => Option[Region]) extends Actor with StrictLogging with Stash {
   // scalastyle:on parameter.number
 
+  import scala.concurrent.ExecutionContext.Implicits.global
```
meln1k: why not `context.dispatcher`?

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -39,58 +39,80 @@ class TaskReplaceActor(
   // Killed resident tasks are not expunged from the instances list. Ignore
   // them. LaunchQueue takes care of launching instances against reservations
   // first
-  val currentRunningInstances = instanceTracker.specInstancesSync(runSpec.id).filter(_.isActive)
+  var currentRunningInstances: Seq[Instance] = _
```
meln1k: What do you think about removing this mutable state by using a receive method with parameters?

jeschkies: It used to be like this in the instance tracker and I did not like it. In this case it's probably not much of a difference since we won't change the code but I find it easier to track this way.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -39,58 +39,80 @@ class TaskReplaceActor(
...
   // All already started and active tasks are filtered while the rest is considered
-  private[this] val (instancesAlreadyStarted, instancesToKill) = {
-    currentRunningInstances.partition(_.runSpecVersion == runSpec.version)
-  }
+  private[this] var instancesAlreadyStarted: Seq[Instance] = _
+  private[this] var instancesToKill: Seq[Instance] = _
 
   // The ignition strategy for this run specification
-  private[this] val ignitionStrategy = computeRestartStrategy(runSpec, currentRunningInstances.size)
+  private[this] lazy val ignitionStrategy: RestartStrategy = computeRestartStrategy(runSpec, currentRunningInstances.size)
```
timcharper: Maybe this should be a def? It seems problematic to have a lazy val encapsulate a value of a var the first time it is accessed.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -39,58 +39,80 @@ class TaskReplaceActor(
...
 
   // compute all variables maintained in this actor =========================================================
 
   // All instances to kill queued up
-  private[this] val toKill: mutable.Queue[Instance] = instancesToKill.to[mutable.Queue]
+  private[this] lazy val toKill: mutable.Queue[Instance] = instancesToKill.to[mutable.Queue]
```
timcharper: Why is this lazy?

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -39,58 +39,80 @@ class TaskReplaceActor(
...
 
   // All instances to kill as set for quick lookup
-  private[this] var oldInstanceIds: SortedSet[Id] = instancesToKill.map(_.instanceId).to[SortedSet]
+  private[this] var oldInstanceIds: SortedSet[Id] = _
```
timcharper: Default to SortedSet.empty?

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
jeschkies: It depends on `instancesToKill` which is not initialized tat the beginning.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
jeschkies: Good point..

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: I copied form other actors but it's a good point. Do we want to use the same thread pool as the actor system?

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -39,58 +39,80 @@ class TaskReplaceActor(
   // Killed resident tasks are not expunged from the instances list. Ignore
   // them. LaunchQueue takes care of launching instances against reservations
   // first
-  val currentRunningInstances = instanceTracker.specInstancesSync(runSpec.id).filter(_.isActive)
+  var currentRunningInstances: Seq[Instance] = _
 
   // In case previous master was abdicated while the deployment was still running we might have
   // already started some new tasks.
   // All already started and active tasks are filtered while the rest is considered
-  private[this] val (instancesAlreadyStarted, instancesToKill) = {
```
alenkacz: we use partition here before and don't anymore - so we have to iterate the list twice. But if instances count is fairly low, it should not make a difference...

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala
```diff
@@ -17,25 +17,17 @@ import scala.concurrent.{ ExecutionContext, Future }
   * * Creating an instance
   * * Updating an instance (due to a state change, a timeout, a Mesos update)
   * * Expunging an instance
-  *
-  *
-  * FIXME: To allow introducing the new asynchronous [[InstanceTracker]] without needing to
```
alenkacz: 👍

jeschkies:
A loop is running [here](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-async-tracker/).

jeschkies:
```
   1 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Downscale with rolling upgrade with 1 over-capacity
   1 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should If all tasks are replaced already, the actor stops immediately
   1 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should If all tasks are replaced already, we will wait for the readiness checks
   1 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Replace and scale down from more than new minCapacity
   1 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Replace with health checks
   1 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Replace with minimum running tasks
   1 mesosphere.marathon.integration.DockerAppIntegrationTest:(It is not a test it is a sbt.testing.SuiteSelector)
   2 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Replace with rolling upgrade with 2/3 over-capacity
   3 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Replace with rolling upgrade with minimal over-capacity
   6 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should If all tasks are replaced already, we will wait for the readiness checks and health checks
   6 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Replace with rolling upgrade without over-capacity
   9 mesosphere.marathon.core.deployment.impl.TaskReplaceActorTest:TaskReplaceActor should Wait until the tasks are killed
Sample size: 88
```

Ok, I'm happy I ran these tests. `TaskReplaceActor` is clearly not stable.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -39,58 +39,80 @@ class TaskReplaceActor(
...
-    currentRunningInstances.partition(_.runSpecVersion == runSpec.version)
-  }
+  private[this] var instancesAlreadyStarted: Seq[Instance] = Seq.empty
+  private[this] var instancesToKill: Seq[Instance] = Seq.empty
 
   // The ignition strategy for this run specification
-  private[this] val ignitionStrategy = computeRestartStrategy(runSpec, currentRunningInstances.size)
+  private[this] lazy val ignitionStrategy: RestartStrategy = computeRestartStrategy(runSpec, currentRunningInstances.size)
```
ichernetsky: BTW, why not just `restartStrategy` here? I know it was like this before. Just wondering.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -109,7 +112,7 @@ private class TaskLauncherActor(
 
     logger.info(s"Started instanceLaunchActor for ${runSpec.id} version ${runSpec.version} with initial count $instancesToLaunch")
 
-    instanceMap = instanceTracker.instancesBySpecSync.instancesMap(runSpec.id).instanceMap
+    instanceTracker.instancesBySpec().pipeTo(self)
     rateLimiterActor ! RateLimiterActor.GetDelay(runSpec)
```
ichernetsky: Shouldn't this be moved to `initializing` in order to keep the old behavior?

jeschkies:
Dropped in favour of `feature/MARATHON-8167`

PR #6162, opened 2018-04-17
Fix for flaky test_pod_with_container_bridge_network

meln1k:
Summary: Added a retry

JIRA issues: MARATHON-8154

### DIFF
--- a/tests/system/marathon_pods_tests.py
+++ b/tests/system/marathon_pods_tests.py
@@ -454,12 +454,7 @@ def test_pod_with_container_bridge_network():
     assert agent_ip != container_ip, "The container IP address is the same as the agent one"
 
     url = "http://{}:{}/".format(agent_ip, port)
-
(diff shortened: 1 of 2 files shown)


tests/system/marathon_pods_tests.py
```diff
@@ -454,7 +454,12 @@ def test_pod_with_container_bridge_network():
     assert agent_ip != container_ip, "The container IP address is the same as the agent one"
 
     url = "http://{}:{}/".format(agent_ip, port)
-    common.assert_http_code(url)
+
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=60, retry_on_exception=common.ignore_exception)
```
zen-dog: So the problem was that nginx hasn't fully booted up yet?

meln1k: Maybe :)

PR #6166, opened 2018-04-20
Remove synchronous `LaunchQueue` methods

zen-dog:
Summary:
Removed synchronous `LaunchQueue` methods and replaced them with their asynchronous counterparts.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -455,9 +455,11 @@ class SchedulerActions(
 
     }
 
-    instancesToStart.foreach { toStart: Int =>
+    if (instancesToStart.isDefined) {
+      val toStart = instancesToStart.get
+
       logger.info(s"Need to scale ${runSpec.id} from ${runningInstances.size} up to $targetCount instances")
-      val leftToLaunch = launchQueue.get(runSpec.id).fold(0)(_.instancesLeftToLaunch)
+      val leftToLaunch = await(launchQueue.getAsync(runSpec.id)).fold(0)(_.instancesLeftToLaunch)
       val toAdd = toStart - leftToLaunch
 
       if (toAdd > 0) {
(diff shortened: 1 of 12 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueDelegate.scala
```diff
@@ -28,59 +28,33 @@ private[launchqueue] class LaunchQueueDelegate(
...
-    askQueueActor[LaunchQueueDelegate.Request, Seq[QueuedInstanceInfoWithStatistics]]("listWithStatistics")(LaunchQueueDelegate.ListWithStatistics)
-  }
-
   override def listWithStatisticsAsync: Future[Seq[QueuedInstanceInfoWithStatistics]] =
     askQueueActorFuture[LaunchQueueDelegate.Request, Seq[QueuedInstanceInfoWithStatistics]]("listWithStatistics")(LaunchQueueDelegate.ListWithStatistics)
 
-  override def get(runSpecId: PathId): Option[QueuedInstanceInfo] =
-    askQueueActor[LaunchQueueDelegate.Request, Option[QueuedInstanceInfo]]("get")(LaunchQueueDelegate.Count(runSpecId))
-
   override def getAsync(runSpecId: PathId): Future[Option[QueuedInstanceInfo]] =
```
alenkacz: can we now remove the word `Async` from all method names since this is the way we do stuff in all methods? :)

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -455,14 +455,16 @@ class SchedulerActions(
 
     }
 
-    instancesToStart.foreach { toStart: Int =>
```
alenkacz: why not use foreach? :) I think it's the way we do such stuff in scala, right?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueDelegate.scala
jeschkies: I'd do this in a second step.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
jeschkies: Async/await cannot be nested in closures.

zen-dog: Exactly!

PR #6172, opened 2018-04-25
Rolling back deployments allows duplicate service ports

meln1k:
Summary: now we check apps for assigning ports if we're doing a rollback

JIRA issues: MARATHON-8160


zen-dog:
That was quick

PR #6176, opened 2018-04-26
Adding `revivedOffers` metric

wavesoft:
This PR introduces the `mesosphere.marathon.core.flow.impl.ReviveOffersActor.revivedOffers`
metric that counts the times the ReviveOffersActor sends a revive request to mesos.

Related JIRA: MARATHON-8161

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
@@ -3,16 +3,15 @@ package core.flow.impl
 
 import java.time.Clock
 
-import akka.actor.{Actor, Cancellable, Props}
-import akka.event.{EventStream, LoggingReceive}
+import akka.actor.{ Actor, Cancellable, Props }
+import akka.event.{ EventStream, LoggingReceive }
 import com.typesafe.scalalogging.StrictLogging
-import mesosphere.marathon.MarathonSchedulerDriverHolder
 import mesosphere.marathon.core.flow.ReviveOffersConfig
 import mesosphere.marathon.core.flow.impl.ReviveOffersActor.OffersWanted
 import mesosphere.marathon.core.event.{SchedulerRegisteredEvent, SchedulerReregisteredEvent}
-import mesosphere.marathon.metrics.{Metrics, ServiceMetric}
+import mesosphere.marathon.metrics.{ Metrics, ServiceMetric }
 import mesosphere.marathon.state.Timestamp
-import rx.lang.scala.{Observable, Subscription}
+import rx.lang.scala.{ Observable, Subscription }
 
 import scala.annotation.tailrec
 import scala.concurrent.duration._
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
```diff
@@ -37,6 +38,8 @@ private[impl] class ReviveOffersActor(
     offersWanted: Observable[Boolean],
     driverHolder: MarathonSchedulerDriverHolder) extends Actor with StrictLogging {
 
+  private[this] val revivedOffersMetric = Metrics.minMaxCounter(ServiceMetric, getClass, "revivedOffers")
```
zen-dog: Why do we use the class name for metrics name? Does it fit in our current schema? I mean, for a marathon user is `mesosphere.marathon.core.flow.impl.ReviveOffersActor.revivedOffers` pretty much non-sensical as metrics-prefix.

wavesoft: Yeah, I know. We are already doing this already everywhere, and changing this would be a breaking change to our consumers.

zen-dog: So we use the class prefix :sigh: but `revivedOffers` sounds like this is the number of offers revived. However, this is the number of times we send `Revive` command to Mesos. So maybe smth. like `numOfferRevivals`?

src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
```diff
@@ -3,15 +3,16 @@ package core.flow.impl
...
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.MarathonSchedulerDriverHolder
 import mesosphere.marathon.core.flow.ReviveOffersConfig
 import mesosphere.marathon.core.flow.impl.ReviveOffersActor.OffersWanted
-import mesosphere.marathon.core.event.{ SchedulerRegisteredEvent, SchedulerReregisteredEvent }
+import mesosphere.marathon.core.event.{SchedulerRegisteredEvent, SchedulerReregisteredEvent}
+import mesosphere.marathon.metrics.{Metrics, ServiceMetric}
 import mesosphere.marathon.state.Timestamp
-import rx.lang.scala.{ Observable, Subscription }
+import rx.lang.scala.{Observable, Subscription}
```
ichernetsky: The imports should be re-formatted.

kensipe: What Ivan said ^ . The format should not have been changed and causes problems with code updates during builds.

src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
kensipe: I agree... how about `reviveCount`

src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
wavesoft: Oh, this was IntelliJ auto-formatting. I guess I have to configure

src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
wavesoft: `reviveCount` works for me 👍

PR #6177, opened 2018-04-26
Move repository call to instance tracker

alenkacz:
Summary:
To make the instance update flow more readable we decided to change the flow of calls and move the InstanceRepository.store call from InstanceOpProcessor to InstanceTrackerActor that's called from the process method anyway and is already making sure the state is updated in the InstanceTrackerActor itself.

This PR tries to be as much backward compatible as possible so I was trying to keep all the logs in the places they were before as well as 

JIRA issues: MARATHON-8177


src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
```diff
@@ -19,7 +19,6 @@ import scala.util.control.NonFatal
   */
 private[tracker] class InstanceOpProcessorImpl(
```
alenkacz: if we merge this, the next easy step will be to get rid of this whole class because it contains almost nothing right now. I kept it here because it will be easier to review this way...

jeschkies: Do you have a ticket for follow up?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
```diff
@@ -28,17 +22,21 @@ private[tracker] class InstanceOpProcessorImpl(
...
-          InstanceTrackerActor.Ack(op.sender, change)
-        }
-          .recoverWith(tryToRecover(op)(
-            expectedState = None, oldState = Some(change.instance), change.events))
-          .flatMap { ack: InstanceTrackerActor.Ack => notifyTaskTrackerActor(ack) }
+        implicit val taskTrackerQueryTimeout: Timeout = config.internalTaskTrackerRequestTimeout().milliseconds
+
+        val msg = InstanceTrackerActor.StateChanged(InstanceTrackerActor.Ack(op.sender, change))
+        logger.debug(s"Notify instance tracker actor: msg=$msg")
+        val f = (instanceTrackerRef ? msg).map(_ => ())
```
jeschkies: So we just forward the message?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,34 +139,53 @@ private[impl] class InstanceTrackerActor(
...
+      case InstanceTrackerActor.StateChanged(ack) =>
+        import context.dispatcher
 
-          case InstanceUpdateEffect.Expunge(instance, events) =>
+        val maybeChange: Future[Option[InstanceChange]] = ack.effect match {
+          case effect @ InstanceUpdateEffect.Update(instance, _, _) =>
+            repository.store(instance) // first store to the repository
+              .map(_ => Some(effect))
+              .recoverWith(tryToRecoverRepositoryFailure(effect))
+              .map(trackerEffect => // update task tracked based on the repository update result
```
jeschkies: Could you add the type of the effect? That'd be great.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,34 +139,53 @@ private[impl] class InstanceTrackerActor(
...
-          case InstanceUpdateEffect.Update(instance, oldInstance, events) =>
-            updateApp(instance.runSpecId, instance.instanceId, newInstance = Some(instance))
-            Some(InstanceUpdated(instance, lastState = oldInstance.map(_.state), events))
+      case InstanceTrackerActor.StateChanged(ack) =>
+        import context.dispatcher
 
-          case InstanceUpdateEffect.Expunge(instance, events) =>
+        val maybeChange: Future[Option[InstanceChange]] = ack.effect match {
+          case effect @ InstanceUpdateEffect.Update(instance, _, _) =>
+            repository.store(instance) // first store to the repository
```
jeschkies: What do you think about using async/await here? It would improve the readability of the code.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -189,10 +212,50 @@ private[impl] class InstanceTrackerActor(
...
+  /**
+    * If we encounter failure, we try to reload the effected task to make sure that the taskTracker
+    * is up-to-date. We signal failure to the sender if the state is not as expected.
+    *
+    * If reloading the tasks also fails, the operation does fail.
+    *
+    * This tries to isolate failures that only effect certain tasks, e.g. errors in the serialization logic
+    * which are only triggered for a certain combination of fields.
+    */
+  private def tryToRecoverRepositoryFailure(update: InstanceUpdateEffect.Update)(
```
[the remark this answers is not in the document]
jeschkies: I'd say the master should abdicate. However, since this a change in the behaviur we should not hide it in this diff but make a separate diff. What do you think, @meichstedt and @timcharper?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,34 +139,53 @@ private[impl] class InstanceTrackerActor(
...
 
-          case InstanceUpdateEffect.Expunge(instance, events) =>
+        val maybeChange: Future[Option[InstanceChange]] = ack.effect match {
+          case effect @ InstanceUpdateEffect.Update(instance, _, _) =>
+            repository.store(instance) // first store to the repository
+              .map(_ => Some(effect))
+              .recoverWith(tryToRecoverRepositoryFailure(effect))
+              .map(trackerEffect => // update task tracked based on the repository update result
+                trackerEffect.flatMap(ef => {
+                  updateApp(ef.instance.runSpecId, ef.instance.instanceId, newInstance = Some(ef.instance))
```
jeschkies: I don't think `updateApp` is threadsafe. Checkout the [implementation](https://github.com/mesosphere/marathon/blob/9daa627daec7d3b6cbb1cfcdade9699c5d30c951/src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTracker.scala#L82).

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,34 +139,53 @@ private[impl] class InstanceTrackerActor(
...
+                // since we currently only use ContinueOnErrorSteps, we can simply ignore failures here
+                logger.warn("updateStepProcessor.process failed", cause)
+                Done
+            }
+          }.getOrElse(Future.successful(Done))
+        }.onComplete {
+          case Success(_) =>
+            ack.sendAck()
+            originalSender ! (())
+          case Failure(e) =>
```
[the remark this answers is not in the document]
jeschkies: Hm, I think we have to actually wait until this future completes. Aas far as I understand the `InstanceOpProcessor` it sequentualized all updates. We could achieve the effect with  the `WorkQueue`.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -189,10 +212,50 @@ private[impl] class InstanceTrackerActor(
     }
 
     instancesBySpec = updatedAppInstances
+    println(instancesBySpec)
```
jeschkies: Is this a left over?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActorTest.scala
```diff
@@ -188,6 +189,231 @@ class InstanceTrackerActorTest extends AkkaUnitTest {
       verify(f.stepProcessor).process(any)(any[ExecutionContext])
     }
   }
+
+  "updates repository as well as internal state for instance update" in {
```
jeschkies: Are these the old `InstanceOpProcessorImplTest` tests?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
zen-dog: This class is going to be removed anyway, right?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
zen-dog: second that. we decided to switch to async/await whenever appropriate

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
zen-dog: Yep, you're changing actor's state (`instancesBySpec`) from the future context which is a race condition. 
It'll make the logic more complex but you should pipe the `trackerEffect` to yourself and handle it in another receive match case. Don't forget to handle the failure case too (see [here](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala#L141) for example)
Mixing actors and futures only works nicely as long as future effects are stateless.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,34 +139,53 @@ private[impl] class InstanceTrackerActor(
...
+                }))
+
+          case effect @ InstanceUpdateEffect.Expunge(instance, _) =>
             logger.debug(s"Received expunge for ${instance.instanceId}")
-            updateApp(instance.runSpecId, instance.instanceId, newInstance = None)
-            Some(InstanceDeleted(instance, lastState = None, events))
+            repository.delete(instance.instanceId) // first store to the repository
+              .map(_ => Some(effect))
+              .recoverWith(tryToRecoverRepositoryFailure(effect))
+              .map(trackerEffect => // update task tracked based on the repository update result
```
zen-dog: ...same here

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,34 +139,53 @@ private[impl] class InstanceTrackerActor(
...
-          updateStepProcessor.process(change).recover {
-            case NonFatal(cause) =>
-              // since we currently only use ContinueOnErrorSteps, we can simply ignore failures here
-              logger.warn("updateStepProcessor.process failed", cause)
-              Done
-          }
-        }.getOrElse(Future.successful(Done)).foreach { _ =>
-          ack.sendAck()
-          originalSender ! (())
+        maybeChange.flatMap { changeOpt =>
```
zen-dog: again, I'd prefer to see await/async here - it makes it explicit that the code beneath now happens in a different context.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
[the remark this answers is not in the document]
zen-dog: Please, don't drag the `WorkQueue` here. From what I can see we don't have to wait for this future to complete - the original sender will receive the answer at some point.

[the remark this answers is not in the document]
zen-dog: > if I did not miss anything, in case of failed future, sender never got reply in the previous implementation 

how so? all but Fatal exceptions are ignored (see the comment) with a warning and a `Done` is returned.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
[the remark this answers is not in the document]
zen-dog: Ideally, this logic goes into the failure handling of the `pipe` call (see my comment above).

zen-dog:
And since the overall change in logic will be quite substantial I'd like to see 100 run stats ;)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
alenkacz: no, I was planning to do it in the same ticket, just as a separate PR

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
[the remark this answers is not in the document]
jeschkies: I've talked to @meichstedt. He said we should just abdicate when we fail to write something.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,7 +142,25 @@ private[impl] class InstanceTrackerActor(
...
+
+        ack.effect match {
+          case InstanceUpdateEffect.Update(instance, _, _) =>
+            updateRepository(() => repository.store(instance), ack)
+              .pipeTo(self)(sender())
+
+          case InstanceUpdateEffect.Expunge(instance, _) =>
+            logger.debug(s"Received expunge for ${instance.instanceId}")
+            updateRepository(() => repository.delete(instance.instanceId), ack)
+              .pipeTo(self)(sender())
```
jeschkies: What is `pipeTo(self)(sender())` doing? Is the result sent to `self` and `sender()`?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
jeschkies: _push_

jeschkies:
@alenkacz, the loop is running [here](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-pr-6177/).

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
alenkacz: yes, thanks

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
alenkacz: no the second parameter is the sender, so I wanted to make sure that when processing the second message, when calling `sender()` I will get sender of the first message... Maybe it's redundand and akka passes the original sender by default, I have to check... In any case I think it's not breaking anything, it's just redundant

jeschkies:
Here is the outcome of my build loop
```
   1 mesosphere.marathon.MarathonSchedulerActorTest:MarathonSchedulerActor should Kill tasks with scaling
   1 mesosphere.marathon.integration.ReadinessCheckIntegrationTest:ReadinessChecks should A deployment of an application with readiness checks (no health) does finish when the app is ready
Sample size: 99
```

And this is @alenkacz's loop
```
   1 mesosphere.marathon.integration.AppDeployIntegrationTest:AppDeploy should restarting an app with backoff delay starts immediately
Sample size: 99
```

Here is the master in comparison
```
   1 mesosphere.marathon.integration.DockerAppIntegrationTest:(It is not a test it is a sbt.testing.SuiteSelector)
   1 mesosphere.marathon.integration.MesosAppIntegrationTest:MesosApp should deploy a simple pod with health checks
Sample size: 97
```

There is an issue with this patch. Checkout build [marathon-loop-pr-6177/35](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-pr-6177/35)
```
Wanted but not invoked:
taskFailureRepository.store(
    TaskFailure(/test-app-kill-with-scale,value: "test-app-kill-with-scale.instance-03a28cf6-5045-11e8-884d-02429c31d540.$anon"
,TASK_FAILED,,host.some,1970-01-01T00:00:00.000Z,2018-05-05T09:16:37.875Z,Some(value: ""
))
);
-> at mesosphere.marathon.MarathonSchedulerActorTest.$anonfun$new$44(MarathonSchedulerActorTest.scala:231)
Actually, there were zero interactions with this mock.
```

That's a unit test.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -167,6 +192,16 @@ private[impl] class InstanceTrackerActor(
     }
   }
 
+  private def updateRepository(repositoryFunc: () => Future[Done], ack: Ack)(implicit ec: ExecutionContext): Future[RepositoryStateUpdated] = {
```
zen-dog: This is a very strange recovery code: we write and fail, but then we try to read and if we find the value then everything is ok? Is there an actual ZK scenario that is covered by this? /cc @timcharper 
Otherwise, I'll simply remove all that and say `repo.delete().pipeTo(self)` where we simply abdicate on write requests.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -167,6 +192,16 @@ private[impl] class InstanceTrackerActor(
...
+    repositoryFunc()
+      .recoverWith(tryToRecoverRepositoryFailure(ack.effect))
+      .map(_ => RepositoryStateUpdated(ack))
+      .recoverWith {
+        // if we could not recover from repository failure, propagate the error
+        case NonFatal(e) => Future.successful(RepositoryStateUpdated(Ack(ack.initiator, InstanceUpdateEffect.Failure(e))))
```
zen-dog: In case it's a `Fatal` exception, pipeTo will [wrap](https://github.com/akka/akka/blob/v2.5.7/akka-actor/src/main/scala/akka/pattern/PipeToSupport.scala#L20) it into `Status.Failure(f)` and send as a message to this actor, correct? I think you need to handle this message with an extra `Receive` case.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
alenkacz: based on the comment this is not trying to treat a ZK error but serialization error. Apparently this is trying to treat potential serialization errors that e.g. happened just for some combination of fields.

I would argue that I cleaned the code to the point where it's not such a big problem to leave it here but I won't be against removing that. But if we do that, it is up to discussion whether we don't have to distinguish between ZK errors and serialization errors here and still treat serialization errors as we do now (without abdicating)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
alenkacz: good catch! Thanks ... I think I just want to throw the exception in that case... (fixed)

src/main/scala/mesosphere/marathon/core/history/impl/HistoryActor.scala
```diff
@@ -3,27 +3,32 @@ package core.history.impl
...
   def receive: Receive = {
 
-    case TaskFailure.FromUnhealthyInstanceKillEvent(taskFailure) =>
+    case m @ TaskFailure.FromUnhealthyInstanceKillEvent(taskFailure) =>
       taskFailureRepository.store(taskFailure)
 
-    case TaskFailure.FromMesosStatusUpdateEvent(taskFailure) =>
+    case m @ TaskFailure.FromMesosStatusUpdateEvent(taskFailure) =>
+      println("mesos update")
+      println(m)
```
alenkacz: this is just temporary debugging for the loop. If we decide to merge I'll move it to a different branch

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,7 +142,30 @@ private[impl] class InstanceTrackerActor(
...
+
+          case _ =>
+            Future.successful(RepositoryStateUpdated(ack)) // forward only
+              .pipeTo(self)(sender())
+        }
+
+      case Status.Failure(e) =>
+        // pipeTo self might cause failed future to be propagated as Status.Failure
+        // this should happen only for fatal exceptions
+        throw e
```
zen-dog: You definitely want some logging here (on error level)!

alenkacz: yep, I already fixed that locally in the morning :) thanks

src/main/scala/mesosphere/marathon/core/history/impl/HistoryActor.scala
```diff
@@ -3,27 +3,32 @@ package core.history.impl
...
     eventBus.subscribe(self, classOf[UnhealthyInstanceKillEvent])
     eventBus.subscribe(self, classOf[AppTerminatedEvent])
+
+    logger.info("History actor ready")
   }
 
   def receive: Receive = {
 
-    case TaskFailure.FromUnhealthyInstanceKillEvent(taskFailure) =>
+    case m @ TaskFailure.FromUnhealthyInstanceKillEvent(taskFailure) =>
```
zen-dog: what is `m` for here and below?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -167,6 +202,14 @@ private[impl] class InstanceTrackerActor(
     }
   }
 
+  private def updateRepository(repositoryFunc: () => Future[Done], ack: Ack)(implicit ec: ExecutionContext): Future[RepositoryUpdateResult] = {
+    repositoryFunc()
+      .map(_ => RepositoryStateUpdated(ack))
+      .recoverWith {
+        case e => Future.successful(RepositoryStateUpdateFailed(e))
```
zen-dog: I get why are you doing it but throwing a `Future.successful` on a failed operation looks strange 🤔

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,7 +144,33 @@ private[impl] class InstanceTrackerActor(
...
+            updateRepository(() => repository.store(instance), ack)
+              .pipeTo(self)(sender())
+
+          case InstanceUpdateEffect.Expunge(instance, _) =>
+            logger.debug(s"Received expunge for ${instance.instanceId}")
+            updateRepository(() => repository.delete(instance.instanceId), ack)
+              .pipeTo(self)(sender())
+
+          case _ =>
+            Future.successful(RepositoryStateUpdated(ack)) // forward only
```
zen-dog: Future into pipe? Why not simply [forward](https://doc.akka.io/docs/akka/current/actors.html?language=scala#Forward_message) the message e.g. `self forward RepositoryStateUpdated(ack)`? ;)

src/main/scala/mesosphere/marathon/core/history/impl/HistoryActor.scala
```diff
@@ -3,27 +3,32 @@ package core.history.impl
 
 import akka.actor.Actor
 import akka.event.EventStream
+import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.event._
 import mesosphere.marathon.state.TaskFailure
 import mesosphere.marathon.storage.repository.TaskFailureRepository
 
 // TODO(PODS): Move from Task to Instance
```
jeschkies: I think this todo can go.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -135,7 +144,33 @@ private[impl] class InstanceTrackerActor(
...
+              .pipeTo(self)(sender())
+        }
+
+      case RepositoryStateUpdateFailed(e) =>
+        e match {
+          case NonFatal(ex) => sender() ! Status.Failure(ex)
+          case ex =>
+            logger.error("Fatal exception after repository update in InstanceTrackerActor rethrown: ", ex)
```
jeschkies: I think we should abdicate here.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
alenkacz: yeah, I was thinking about this one too... I did it this way just so I can wrap that thing before using pipeTo :/ I think the whole way we use ask pattern all over place is broken.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceOpProcessorImpl.scala
```diff
@@ -28,17 +22,21 @@ private[tracker] class InstanceOpProcessorImpl(
     logger.debug(s"Process $op")
```
timcharper: Can we change the return type of this from Future[Unit] to Future[Done] ?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
timcharper: It seems like repository func is unneeded? It is called once and called immediately. Better to just pass in the Future?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
timcharper: After consideration, do you think just doing onComplete and removing the updateRepository method might be more straightforward?

```
val originalSender = sender
repository.delete(instance.instanceId).onComplete {
  case Failure(NonFatal(ex)) =>
    originalSender ! Status.Failure(ex)
  case Success(_) =>
    self.tell(RepositoryStateUpdated(ack), originalSender)
}
```

Seems substantially shorter considering it would remove the private method, extra status and associated receive handler, etc.

alenkacz:
@timcharper my main motivation for introducing that private method was because the same recovery logic was needed for both expunge and update. Now we dropped most of the recovery logic so I guess I can get rid of that. Thanks!

alenkacz:
Thanks everyone for help and patience! :)

PR #6180, opened 2018-05-02
 Introduce `Provisoned` and `Scheduled` condition.

jeschkies:
Summary:
We would like to create instances in a scheduled state *before* they are
launched. That is why we introduce a `Scheduled` and `Provisioned`
condition.

So far instances require agent info. However, when an instance is just
scheduled there is no meanningful agent info. So it has to be optional.
All requests that require this info filter out instances that have no
info. This way we stay backwards compatible since there are no instance
in scheduled condition currently it is safe to ignore them.

JIRA issues: MARATHON-8168


jeschkies:
I will follow up with a second pull requests that surfaces the value of these changes.

src/main/scala/mesosphere/marathon/api/v2/AppTasksResource.scala
```diff
@@ -69,11 +69,10 @@ class AppTasksResource @Inject() (
...
     appIds.withFilter(instancesBySpec.hasSpecInstances).flatMap { id =>
       val health = result(healthCheckManager.statuses(id))
-      instancesBySpec.specInstances(id).flatMap { instance =>
-        instance.tasksMap.values.map { task =>
-          EnrichedTask(instance, task, health.getOrElse(instance.instanceId, Nil))
-        }
-      }
+      instancesBySpec.specInstances(id).collect {
+        case i @ EnrichedTask.All(enrichedTasks) =>
+          enrichedTasks.map { _.copy(healthCheckResults = health.getOrElse(i.instanceId, Nil)) }
```
jeschkies: Once the health check status is part of an instance this will be simplified.

src/main/scala/mesosphere/marathon/core/appinfo/EnrichedTask.scala
```diff
@@ -16,9 +16,26 @@ case class EnrichedTask(
     reservation: Option[Reservation])
 
 object EnrichedTask {
-  def apply(instance: Instance, task: Task, healthCheckResults: Seq[Health],
-    servicePorts: Seq[Int] = Nil): EnrichedTask = {
-    new EnrichedTask(instance.runSpecId, task, instance.agentInfo, healthCheckResults, servicePorts = servicePorts,
-      reservation = instance.reservation)
+
+  object All {
```
alenkacz: can we maybe find better names for these? I had hard time understanding that does EnrichedTask.App and EnrichedTask.All mean. Maybe EnrichedTask.Single and EnrichedTask.All ?

Or maybe the EnrichedTask name does not make sense anymore altogether, since it's a collection of tasks?

Maybe we can go with `EnrichedTasks` then?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -83,6 +83,30 @@ object Instance {
   def instancesById(instances: Seq[Instance]): Map[Instance.Id, Instance] =
     instances.map(instance => instance.instanceId -> instance)(collection.breakOut)
 
+  object Running {
+    def unapply(instance: Instance): Option[Tuple3[Instance.Id, Instance.AgentInfo, Map[Task.Id, Task]]] = instance match {
+      case Instance(instanceId, Some(agentInfo), InstanceState(Condition.Running, _, _, _), tasksMap, _, _, _) =>
+        Some((instanceId, agentInfo, tasksMap))
+      case _ =>
+        Option.empty[Tuple3[Instance.Id, Instance.AgentInfo, Map[Task.Id, Task]]]
```
alenkacz: we can't use None here because of type inference?

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
```diff
@@ -132,14 +132,14 @@ private[health] class HealthCheckActor(
...
+        s"Detected unhealthy $instanceId of app [${app.id}] version [${app.version}] on host ${instance.hostname}"
       )
 
       // kill the instance, if it is reachable
       if (instance.isUnreachable) {
-        logger.info(s"Instance $instanceId on host ${instance.agentInfo.host} is temporarily unreachable. Performing no kill.")
+        logger.info(s"Instance $instanceId on host ${instance.hostname} is temporarily unreachable. Performing no kill.")
       } else {
-        logger.info(s"Send kill request for $instanceId on host ${instance.agentInfo.host} to driver")
+        logger.info(s"Send kill request for $instanceId on host ${instance.hostname} to driver")
```
alenkacz: before the log was printing out hostname, now it will be printing out Some(hostname) I think...

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
```diff
@@ -135,16 +137,13 @@ class TasksResource @Inject() (
...
         .filter { case (appId, _) => affectedApps.exists(app => app.id == appId) }
         .map {
           case (appId, instances) => taskKiller.kill(appId, _ => instances, wipe)
         })).flatten
-      ok(jsonObjString("tasks" -> killed.flatMap { instance =>
-        instance.tasksMap.valuesIterator.map { task =>
-          EnrichedTask(instance, task, Nil).toRaml
-        }
-      }))
+      val killedTasks = killedInstances.collect { case EnrichedTask.All(enrichedTasks) => enrichedTasks.map(_.toRaml) }.flatten
```
alenkacz: btw. I would find something like `killedInstances.map(ki => ki.toEnrichedTasks).map(_.toRaml)` more readable...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -1,13 +1,13 @@
 package mesosphere.marathon
 package core.instance
 
-import java.util.{ Base64, UUID }
+import java.util.{Base64, UUID}
```
ichernetsky: Formatting :)

jeschkies: Which one do we use? I've seen both.

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
jeschkies: That does not work any more. We have to filter out all instances that have `agentInfo = None`. Since we do not type scheduled instances I cannot guarantee that the agent info is set.

However, I could do

```scala
val killedTasks = killedInstances
    .collect { case EnrichedTask.All(enrichedTasks) => enrichedTasks }
    .flatten.map(_.toRaml)
```

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: I think the formatting will be automatically fixed for you when you compile, right? :)

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
alenkacz: oh! So it's not all but all running then? I totally missed that... should we maybe reflect that somehow somewhere?

src/main/scala/mesosphere/marathon/core/appinfo/EnrichedTask.scala
ichernetsky: Heavy +1 for choosing better names.

src/main/scala/mesosphere/marathon/core/condition/Condition.scala
```diff
@@ -49,6 +49,12 @@ object Condition {
   sealed trait Failure extends Terminal
   sealed trait Active extends Condition
 
+  /** Scheduled: Task should be launched by matching offers. Mesos does not know anything about it. */
+  case object Scheduled extends Condition
```
alenkacz: what do you think about introducing `sealed trait Initial` for these two (similarly as we have Terminal or Active). Then we can filter instances based on this condition. Filtering based on agentInfo being Some (not None) seems to me that the client has to know too much implementation detail.

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
alenkacz: let's do `hostname.getOrElse("unknown")` here to be backward compatible and improve readability of the logs :)

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -83,6 +83,30 @@ object Instance {
...
+      * @param runSpec The run spec the instance will be started for.
+      * @param instanceId The id of the new instance.
+      * @return An instance in the scheduled state.
+      */
+    def apply(runSpec: RunSpec, instanceId: Instance.Id): Instance = {
+      val state = InstanceState(Condition.Scheduled, Timestamp.now(), None, None)
+      Instance(instanceId, None, state, Map.empty, runSpec.version, runSpec.unreachableStrategy, None)
+    }
+  }
+
```
alenkacz: should we add the new conditions to the conditionHierarchy property on line 125 here?

alenkacz: or maybe we don't because the task can never have this condition?

src/main/scala/mesosphere/marathon/core/condition/Condition.scala
jeschkies: Good point. However, we do have to filter for the agent info since a lot of methods require the agent info being present and I want to avoid an option get.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: The hierarchy is used to infer an instance condition from task conditions. However, for `Scheduled` and `Provisioned` the instance condition is set explicitly so we do not need to infer it. In the future we want to distinguish between a task status or condition and an instance goal.

src/main/scala/mesosphere/marathon/core/health/impl/HealthCheckActor.scala
jeschkies: Good point.

src/main/scala/mesosphere/marathon/api/v2/TasksResource.scala
jeschkies: It is all tasks except for the scheduled ones.

PR #6182, opened 2018-05-03
Back-port lightweight SSE flag

wavesoft:
Summary:
This change adds support for the URL argument `/v2/events?plan-format=light` that subscribes to a "lighter" version of the SSE events, where some of the deployment plan fields are excluded.

Back-ported commits:
* 58fbba21f5b55b3def57a2a20aac3e726fe4d875 (Introduce lightweight SSE flag #5531)

JIRA issues: MARATHON_EE-1673

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
+++ b/src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
@@ -49,19 +49,20 @@ class EventsController(
     */
   def eventsSSE(): Route = {
     def isAllowed(allowed: Set[String])(event: MarathonEvent): Boolean = allowed.isEmpty || allowed(event.eventType)
+
     authenticated.apply { implicit identity =>
       authorized(ViewResource, AuthorizedResource.Events).apply {
-        parameters('event_type.*) { events =>
-          extractClientIP { clientIp =>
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/api/akkahttp/v2/EventsController.scala
```diff
@@ -49,19 +49,20 @@ class EventsController(
...
+          complete {
+            Source
+              .fromGraph[MarathonEvent, NotUsed](
+                new EventStreamSourceGraph(eventBus, conf.eventStreamMaxOutstandingMessages(), clientIp)
+              )
+              .filter(isAllowed(events.toSet))
+              .map(event => ServerSentEvent(`type` = event.eventType, data = Json.stringify(
+                Formats.eventToJson(event, light.exists(_.contentEquals("light")))
+              )))
+              .keepAlive(5.second, () => ServerSentEvent.heartbeat)
```
kensipe: looks like it was here before your change... but 5 is a magic number and we shouldn't have them.   not a blocker

timcharper: Yeah, good catch. This code is deleted in HEAD

PR #6183, opened 2018-05-03
Surface exceptions thrown in any part of the election graph

timcharper:
If registering a gauge ends in an exception, or publishing an event to the event stream throws an exception, currently, we would swallow such an exception and appear to end the election graph gracefully.

JIRA Issues: MARATHON-8187


timcharper:
@ichernetsky it just surfaces the exception. The entire graph would stop, before. (because of eagerComplete = true on the broadcast nodes)

Issue #1647, opened 2015-06-11
Keep the rest-api return format consistent when request headers without "Accept: application/json"


BrickXu:
Hi all,

I found that `v2/tasks` must set the request header with `Accept: application/json` to get the json format response, or the marathon will return plain text.

```
curl -X GET -H "Content-type: application/json" http://localhost:8080/v2/tasks
docker_registry	10004	XXXXXX
logs_statsd	10000	XXXXXX


curl -X GET -H "Content-type: application/json" -H "Accept: application/json" http://localhost:8080/v2/tasks
{"tasks":[{"appId":"/ldocker_registry","id":"docker_registry.e0125d32-0f78-11e5-aa0e-fa163ee98acb","host":"logstash8.ops.com","ports":[9027],"startedAt":"2015-06-10T13:59:13.814Z","stagedAt":"2015-06-10T13:59:13.447Z","version":"2015-06-10T13:59:08.979Z","servicePorts":[10002]},...
```

But for the `v2/apps`, it always return json format response without `Accept: application/json`. I think the APIs must keep the same behavior when the request with or without `Accept: application/json`

ENV: 
* marathon 0.8.1
* mesos 0.22

pierlo-upitup:
Actually this is still happening on v0.10.0-RC3:

```
$ curl 'http://localhost:8080/v2/apps/sleeper/tasks/'
sleeper	 	172.18.6.104 172.18.6.104

$ curl 'http://localhost:8080/v2/apps/sleeper/tasks/' -H "Accept: application/json"
{"tasks":[{"appId":"/sleeper","id":"sleeper.1c5789ef-3128-11e5-8f1f-ac87a33a1422","host":"172.18.6.104","ports":[],"startedAt":"2015-07-23T10:46:44.924Z","stagedAt":"2015-07-23T10:46:44.862Z","version":"2015-07-23T10:42:14.631Z"},{"appId":"/sleeper","id":"sleeper.1f542820-3128-11e5-8f1f-ac87a33a1422","host":"172.18.6.104","ports":[],"startedAt":"2015-07-23T10:46:49.936Z","stagedAt":"2015-07-23T10:46:49.873Z","version":"2015-07-23T10:42:14.631Z"}]}%
```

kolloch:
If you use curl, it always sets the `Accept` header. If you do not specify anything explicitly, it will use `Accept: */*`.

[...]

PR #6184, opened 2018-05-03
Upgrade to Jersey 2.27

timcharper:
Pre-requisite for non-blocking responses in Jetty.

- We get rid of `MarathonMediaType.PREFERRED_APPLICATION_JSON` because it specified a qs value that violates the HTTP specification, and Jersey 2 no longer allows it. I have confirmed that the bug described in #1647 does not come back with its removal.

JIRA Issue: MARATHON-8174


src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala
```diff
@@ -73,7 +73,7 @@ class MarathonExceptionMapper extends ExceptionMapper[JavaException] {
 
   private def entity(exception: JavaException): JsValue = exception match {
     case e: NotFoundException =>
-      Json.obj("message" -> s"URI not found: ${e.getNotFoundUri.getRawPath}")
+      Json.obj("message" -> "URI not found")
```
jeschkies: I think you should say which URI was not found.

timcharper: The path is not available on this exception anymore.

timcharper: Also, why do we have to send this to the client? Won't the client know which URL it requested?

jeschkies: It's easier to debug in the logs. However, if the exception does not have the uri it's fine.

timcharper: Does this affect the logs? My reading is this is what we send to the client.

PR #6187, opened 2018-05-04
Async deployment responses

timcharper:
In this change we convert deployment creating API methods to use Jersey 2 AsyncResponse. By doing this, we no longer timeout ("Futures timed out"); if a deployment takes 2 minutes to schedule, and the HTTP client does not have a shorter receive timeout configured, then the response will be delivered after 2 minutes.

Also, we consolidate some response logic, converging places where there were multiple response generators to just simply prefer the MarathonExceptionMapper instead.

JIRA Issues: MARATHON-8174

### DIFF
--- a/src/main/scala/mesosphere/marathon/Exception.scala
+++ b/src/main/scala/mesosphere/marathon/Exception.scala
@@ -52,7 +52,7 @@ case class ConflictingChangeException(msg: String) extends Exception(msg)
 /**
   * Is thrown if an object validation is not successful.
   *
-  * TODO - convert to Rejection
+  * TODO(MARATHON-8202) - convert to Rejection
   *
   * @param obj object which is not valid
   * @param failure validation information kept in a Failure object
--- a/src/main/scala/mesosphere/marathon/Exception.scala
+++ b/src/main/scala/mesosphere/marathon/Exception.scala
@@ -5,27 +5,27 @@ import mesosphere.marathon.state.{PathId, Timestamp}
 @SuppressWarnings(Array("NullAssignment"))
 class Exception(msg: String, cause: Throwable = null) extends scala.RuntimeException(msg, cause)
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 case class PathNotFoundException(id: PathId, version: Option[Timestamp] = None) extends Exception(
   s"Path '$id' does not exist" + version.fold("")(v => s" in version $v")
 )
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 case class AppNotFoundException(id: PathId, version: Option[Timestamp] = None) extends Exception(
   s"App '$id' does not exist" + version.fold("")(v => s" in version $v")
 )
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 case class PodNotFoundException(id: PathId, version: Option[Timestamp] = None) extends Exception(
   s"Pod '$id' does not exist" + version.fold("")(v => s" in version $v")
 )
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 case class UnknownGroupException(id: PathId) extends Exception(s"Group '$id' does not exist")
 
 case class WrongConfigurationException(message: String) extends Exception(message)
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 case class AppLockedException(deploymentIds: Seq[String] = Nil)
   extends Exception(
     "App is locked by one or more deployments. " +
@@ -33,14 +33,14 @@ case class AppLockedException(deploymentIds: Seq[String] = Nil)
       "View details at '/v2/deployments/<DEPLOYMENT_ID>'."
   )
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 case class TooManyRunningDeploymentsException(maxNum: Int) extends Exception(
   s"Max number ($maxNum) of running deployments is achieved. Wait for existing deployments to complete or cancel one." +
     "You can increase max deployment number by using --max_running_deployments parameter but be " +
     "advised that this can have negative effects on the performance."
 )
 
-// TODO - convert to Rejection
+// TODO(MARATHON-8202) - convert to Rejection
 class PortRangeExhaustedException(
     val minPort: Int,
     val maxPort: Int) extends Exception(s"All ports in the range [$minPort-$maxPort) are already in use")
--- a/src/main/scala/mesosphere/marathon/Exception.scala
+++ b/src/main/scala/mesosphere/marathon/Exception.scala
@@ -1,27 +1,31 @@
 package mesosphere.marathon
-
 import com.wix.accord.Failure
 import mesosphere.marathon.state.{PathId, Timestamp}
 
 @SuppressWarnings(Array("NullAssignment"))
 class Exception(msg: String, cause: Throwable = null) extends scala.RuntimeException(msg, cause)
 
+// TODO - convert to Rejection
 case class PathNotFoundException(id: PathId, version: Option[Timestamp] = None) extends Exception(
   s"Path '$id' does not exist" + version.fold("")(v => s" in version $v")
 )
 
+// TODO - convert to Rejection
 case class AppNotFoundException(id: PathId, version: Option[Timestamp] = None) extends Exception(
   s"App '$id' does not exist" + version.fold("")(v => s" in version $v")
 )
 
+// TODO - convert to Rejection
 case class PodNotFoundException(id: PathId, version: Option[Timestamp] = None) extends Exception(
   s"Pod '$id' does not exist" + version.fold("")(v => s" in version $v")
 )
 
+// TODO - convert to Rejection
 case class UnknownGroupException(id: PathId) extends Exception(s"Group '$id' does not exist")
 
 case class WrongConfigurationException(message: String) extends Exception(message)
 
+// TODO - convert to Rejection
 case class AppLockedException(deploymentIds: Seq[String] = Nil)
   extends Exception(
     "App is locked by one or more deployments. " +
@@ -29,17 +33,17 @@ case class AppLockedException(deploymentIds: Seq[String] = Nil)
       "View details at '/v2/deployments/<DEPLOYMENT_ID>'."
   )
 
+// TODO - convert to Rejection
 case class TooManyRunningDeploymentsException(maxNum: Int) extends Exception(
   s"Max number ($maxNum) of running deployments is achieved. Wait for existing deployments to complete or cancel one." +
     "You can increase max deployment number by using --max_running_deployments parameter but be " +
     "advised that this can have negative effects on the performance."
 )
 
+// TODO - convert to Rejection
 class PortRangeExhaustedException(
     val minPort: Int,
     val maxPort: Int) extends Exception(s"All ports in the range [$minPort-$maxPort) are already in use")
-
-case class UpgradeInProgressException(msg: String) extends Exception(msg)
 
 case class CanceledActionException(msg: String) extends Exception(msg)
 
@@ -47,6 +51,9 @@ case class ConflictingChangeException(msg: String) extends Exception(msg)
 
 /**
   * Is thrown if an object validation is not successful.
+  *
+  * TODO - convert to Rejection
+  *
   * @param obj object which is not valid
   * @param failure validation information kept in a Failure object
   */
@@ -80,9 +87,6 @@ class KillingInstancesFailedException(msg: String) extends Exception(msg)
 abstract class DeploymentFailedException(msg: String) extends Exception(msg)
 
 class DeploymentCanceledException(msg: String) extends DeploymentFailedException(msg)
-class AppStartCanceledException(msg: String) extends DeploymentFailedException(msg)
-class AppStopCanceledException(msg: String) extends DeploymentFailedException(msg)
-class ResolveArtifactsCanceledException(msg: String) extends DeploymentFailedException(msg)
 
 /*
  * Store specific exceptions
--- a/src/main/scala/mesosphere/marathon/Exception.scala
+++ b/src/main/scala/mesosphere/marathon/Exception.scala
@@ -45,8 +45,6 @@ case class CanceledActionException(msg: String) extends Exception(msg)
 
 case class ConflictingChangeException(msg: String) extends Exception(msg)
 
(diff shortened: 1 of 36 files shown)


jeschkies:
I see that you base is master. Why not make #6184 the base?

src/main/scala/mesosphere/marathon/api/RestResource.scala
```diff
@@ -2,25 +2,35 @@ package mesosphere.marathon
...
+import scala.util.{ Failure, Success }
 
 trait RestResource extends JaxResource {
-
+  implicit val executionContext: ExecutionContext
   protected val config: MarathonConf
-
+  case class FailureResponse(response: Response) extends Throwable
+
+  def sendResponse(asyncResponse: AsyncResponse)(future: Future[Response]) = {
```
jeschkies: How are we handling timeouts?

timcharper:
@jeschkies it's a WIP; this is not ready for review.

src/main/scala/mesosphere/marathon/api/RestResource.scala
timcharper: Great question :) Will make sure there's a TODO to handle this before removing WIP status

timcharper: Default timeout behavior is correct: no server timeout. If the client is not configured to time out it will go on indefinitely.

kensipe:
the more I think about this... the more I dislike not having a timeout.   A core discipline I use in design is to "manage the things I can't control".   You can't control the number of clients or the number of connections a client have.   We should not be at the mercy of a client / clients.    Shouldn't we have reasonable timeouts on the server side?

timcharper:
@kensipe I'd pose the question though that if we do continue to time out server side, how does timing out server-side for these operations make things better? Specifically:

- If user does not get a deployment success response, then they need to retry to ensure it happens.
- If the user retries, this means they ultimately insert another item in to the queue to be processed.

Does this ultimately mean we will not only have the same number of connections, but more items inserted in to the deployment queue?

Further, worth considering, our SSE events endpoint is a perpetual connection, terminated only by the client hangup (we periodically send an empty message to both keep the client from timing out, and to detect if the client has hung up).

Perhaps most ideal is to timeout AND cancel the work? Maybe we can consider this as a group. It would certainly be a smaller change. Unfortunately, there is no elegant mechanism to detect client hangup :'( You have to write to the socket, and get an IO exception. Which means you need to send the headers. Which means you need to know the deployment ID (because this is in the headers). Which we don't know until we're ready to return the response.

src/main/scala/mesosphere/marathon/Exception.scala
```diff
@@ -1,54 +1,59 @@
...
 case class CanceledActionException(msg: String) extends Exception(msg)
 
 case class ConflictingChangeException(msg: String) extends Exception(msg)
 
-case class AccessDeniedException(msg: String = "Authorization Denied") extends Exception(msg)
-
 /**
   * Is thrown if an object validation is not successful.
+  *
+  * TODO - convert to Rejection
```
alenkacz: I don't like TODOs in code, I think JIRA is fo TODOs... But just my personal preference :)

src/main/scala/mesosphere/marathon/api/Rejection.scala
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+  * Models known reasons why an API request could be confused.
+  */
+sealed trait Rejection {}
+object Rejection {
+  /**
+    * The request was authenticated, but access was not allowed to the resource.
+    */
+  case class AccessDeniedRejection(authorizer: Authorizer, identity: Identity) extends Rejection
```
alenkacz: I would personally drop the word Rejection from these classes...

src/main/scala/mesosphere/marathon/api/RestResource.scala
alenkacz: should we maybe put that ^^ to some comment there? someone else in the future might be asking the same question

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -151,9 +152,9 @@ class GroupManagerImpl(
         eventStream.publish(GroupChangeSuccess(id, version.toString))
       case Success(Left(_)) =>
         ()
-      case Failure(ex: AccessDeniedException) =>
+      case Failure(RejectionException(_: Rejection.AccessDeniedRejection)) =>
```
alenkacz: hmm I thought that rejections are just something API related, it's kind of unexpected that it bubbles down to the core too...

timcharper: Yeah, converting this to a rejection I think surfaced it :)

It should just return a `Left`, but, at some point I had to call it quits or this patch would get really large.

timcharper: I'd solve it right now, but this patch is getting too large.

I'd put a TODO, but people don't like TODOs.

I'd create a JIRA, but realistically it will not get prioritized.

src/main/scala/mesosphere/marathon/api/Rejection.scala
timcharper: The convention was inspired by Akka HTTP, which includes the suffix for each class:

https://doc.akka.io/api/akka-http/current/akka/http/scaladsl/index.html?search=rejection

Also, by convention, Exceptions end with the work "Exception", also.

Does that make it seam okay to you? Or still feel like you would drop it?

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
timcharper: I went ahead and added a note about this in MARATHON-8202

jeschkies:
> You can't control the number of clients or the number of connections a client have. We should not be at the mercy of a client / clients. Shouldn't we have reasonable timeouts on the server side?

@kensipe, I think this is handled by not accepting new connections as @wavesoft mentioned.

src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -17,25 +16,51 @@ trait AuthResource extends RestResource {
...
+  /**
+    * Authenticate an HTTP request, synchronously.
+    *
+    * @param request The incoming HTTP request
+    * @param fn The work to perform with the identity. Not called if authentication fails.
+    *
+    * @return On success, a Jersey Response. On failure, throws a RejectionException.
+    */
+  def authenticated(request: HttpServletRequest)(fn: Identity => Response): Response = {
+    // TODO - just return the identity instead of using a callback
```
jeschkies: When will this be addressed?

src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala
```diff
@@ -18,11 +18,17 @@ import play.api.libs.json.{JsResultException, JsValue, Json}
...
+  * [[RootApplication]].
+  *
+  * Handles a few exceptions, plus all [[Rejection]]s via RejectionException
+  */
 @Provider
 @Singleton
 class MarathonExceptionMapper extends ExceptionMapper[JavaException] with StrictLogging {
 
-  def toResponse(exception: JavaException): Response = {
+  private def exceptionToResponse(exception: JavaException): Response = {
```
jeschkies: What's the motivation of making it private?

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -25,7 +25,7 @@ class TaskKiller @Inject() (
     val config: MarathonConf,
     val authenticator: Authenticator,
     val authorizer: Authorizer,
-    killService: KillService) extends AuthResource with StrictLogging {
+    killService: KillService)(implicit val executionContext: ExecutionContext) extends AuthResource with StrictLogging {
```
jeschkies: Where are we using the execution context?

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -52,15 +53,15 @@ object AppHelpers {
 
   private def checkAuthorization[A, B >: A](action: AuthorizedAction[B], resource: A)(implicit identity: Identity, authorizer: Authorizer): A = {
     if (authorizer.isAuthorized(identity, action, resource)) resource
-    else throw AccessDeniedException()
+    else throw RejectionException(Rejection.AccessDeniedRejection(authorizer, identity))
```
jeschkies: What is the motivation for throwing a rejection exception instead of the `AccessDeniedException`? Couldn't `AccessDeniedException` just extend `RejectionException`?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -87,21 +96,24 @@ class AppsResource @Inject() (
...
-        maybeDeployments = Some(Seq(Identifiable(plan.id)))
-      )
-
-      maybePostEvent(req, appWithDeployments.app)
-      Response
-        .created(new URI(app.id.toString))
-        .header(RestResource.DeploymentHeader, plan.id)
-        .entity(jsonString(appWithDeployments))
-        .build()
+      val groupUpdateResult = groupManager.updateApp(app.id, createOrThrow, app.version, force).map { plan =>
```
jeschkies: This is a little unexpected. Why not the following?

```scala
val plan = await(groupManager.updateApp(app.id, createOrThrow, app.version, force))
```

This would reduce this diff to a minimal.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -186,63 +198,87 @@ class AppsResource @Inject() (
...
   def replace(
     @PathParam("id") id: String,
     body: Array[Byte],
     @DefaultValue("false")@QueryParam("force") force: Boolean,
     @DefaultValue("true")@QueryParam("partialUpdate") partialUpdate: Boolean,
-    @Context req: HttpServletRequest): Response = authenticated(req) { implicit identity =>
+    @Context req: HttpServletRequest,
+    @Suspended asyncResponse: AsyncResponse): Unit = sendResponse(asyncResponse) {
+    async {
+      implicit val identity = await(authenticatedAsync(req))
```
jeschkies: Hm, why not

```scala
@Suspended asyncResponse: AsyncResponse): Unit = authenticatedAsync(req, asyncResponse) { implicit identity =>
   update(id, body, force, partialUpdate, req, allowCreation = true
}
```

So we let `authenticatedAsync` handle the async response. This reduces the number of changes again.

src/main/scala/mesosphere/marathon/api/AuthResource.scala
```diff
@@ -17,25 +16,51 @@ trait AuthResource extends RestResource {
   implicit val authenticator: Authenticator
   implicit val authorizer: Authorizer
 
-  def authenticated(request: HttpServletRequest)(fn: Identity => Response): Response = {
+  /**
+    * Authenticate an HTTP request, asynchronously.
+    *
+    * @return If succeed, future with identity. If failed, returns failed future with a RejectionException.
+    */
+  def authenticatedAsync(request: HttpServletRequest): Future[Identity] = {
```
jeschkies: I'd actually changes this to 

```scala
def authenticatedAsync(request: HttpServletRequest, response: )(fn: Identity => Future[Unit]) = sendResponse(response) { async {
  identity = await(authenticationRequest.transform{ ... })
  await(fn(identity))
}
}
```

this will reduce the changes and simplify things.

src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala
timcharper: They are not used outside.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
timcharper: AuthResource requires it.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
timcharper: Great point, I'll making this change.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
timcharper: Disagreed.

timcharper: We can do

```
authenticatedAsync(req, asyncResponse).map { implicit identity =>
}
```

I am opposed to passing a callback to authenticatedAsync, and having authenticatedAsync return a response. Reasons:

- It encourages nested callback hell.
- The callbacks are not monadic (you can not combine them with map/flatMap as you can a Future)
- Async / await does not work with futures inside callbacks. This means we have to have a new declaration of async for each callback
- Having authenticatedAsync return a response creates two pathways that we need to render auth rejection responses. By only using rejections, we collapse these to a single way.

jeschkies: Good points. It's closer to what we have, though. Let's go with your solution then 🙂.

Since the diff is so big it'll take some time to review it :/.

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
jeschkies: I'm probably missing something. Currently everything is an exception. We can go for rejections but I would do this in a second pull request. This one is fairly big and introduces multiple changes. If it would introduce `sendResponse` and `authenticatedAsync` only it would be easier to follow. The second step could be the introduction of rejections.

src/main/scala/mesosphere/marathon/api/MarathonExceptionMapper.scala
jeschkies: So? I guess that's a longer discussions to have, though. I usually make every public by default unless it changes state and should not be exposed.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
zen-dog: You're already in the `async`, so why not using `await` instead of `map`?

timcharper: ^ see my comment right above yours :-)

PR #6193, opened 2018-05-08
WIP: Async proxy

timcharper:
JIRA Issue: MARATHON-8174


src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,210 @@
...
+    else
+      s
+  }
+  private val connectionContext = new HttpsConnectionContext(sslContext, sslConfig = Some(sslConfig))
+  private val connectionSettings = ClientConnectionSettings(actorSystem).
+    withIdleTimeout(leaderProxyConf.leaderProxyReadTimeout().millis).
+    withConnectingTimeout(leaderProxyConf.leaderProxyConnectionTimeout().millis)
+  private val poolSettings = ConnectionPoolSettings(actorSystem)
+    .withConnectionSettings(connectionSettings)
+    .withMaxRetries(0)
```
jeschkies: Why no retry here?

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  class IgnoringHostnameVerifier extends javax.net.ssl.HostnameVerifier {
```
jeschkies: What is this doing?

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,210 @@
...
+        else if (name.equalsIgnoreCase(LeaderProxyFilter.HEADER_MARATHON_LEADER))
+          response.addHeader(LeaderProxyFilter.HEADER_MARATHON_LEADER, value)
+        else
+          response.addHeader(name, value)
+    }
+    response.addHeader(HEADER_VIA, viaValue)
+    Done
+  }
+
+  private def proxiedRequestHeaders(request: HttpServletRequest): Seq[HttpHeader] = {
```
jeschkies: Hm, I'm not sure if proxy is the right term here. As far as I understand this, we forward the headers from the original request, right? Isn't this a forward then?

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,210 @@
...
+      Option(request.getRemoteAddr)
+    ).flatten.mkString(",")
+    headers += RawHeader(HEADER_FORWARDED_FOR, forwardedFor)
+    Option(request.getHeader("User-Agent")).foreach { ua =>
+      headers += akka.http.scaladsl.model.headers.`User-Agent`(ua)
+    }
+    headers.result()
+  }
+
+  private def createAndConfigureConnection(asyncContext: AsyncContext, url: URL, clientRequest: HttpServletRequest): Future[HttpResponse] = {
```
jeschkies: Nit: I think it would help to prefix the HTTP types. E.g. `jax.HttpServletRequest` and `akka.HttpResponse`. This would ease understanding the mappings.

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,210 @@
...
+      if (hasProxyLoop) {
+        logger.error("Prevent proxy cycle, rejecting request")
+        response.sendError(BadGateway.intValue, ERROR_STATUS_LOOP)
+        Future.successful(Done)
+      } else {
+        val leaderRequest = try {
+          createAndConfigureConnection(asyncContext, url, request)
+        } catch {
+          case ex: Throwable =>
+            Future.failed(ex)
```
jeschkies: Hm. I actually like to abort early. So instead of setting `result` to `Future.failed` I'd just do

```scala
logger.error("Unhandled proxy exception", e)
response.sendError(InternalServerError.intValue)
return
```

There is no need for a big try-catch block IMHO. What do you think?

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,210 @@
...
+        leaderRequest.transformWith {
+          case Failure(ex: akka.stream.StreamTcpException) =>
+            /* Unfortunately, akka-http does not give us a different error message if the TCP connection is established,
+             * but gives no response, or if the TCP connection is refused outright.
+             *
+             * So, we report BadGateway in either case.
+             */
+            logger.error(ERROR_STATUS_CONNECTION_REFUSED, ex)
+            response.sendError(BadGateway.intValue, ERROR_STATUS_CONNECTION_REFUSED)
+            Future.successful(Done)
```
jeschkies: That's funny. Here you actually already reply. I don't think there is a need for `Future.successful(Done)`. Also, why not use async/await here?

src/main/scala/mesosphere/marathon/api/forwarder/ServletOutputStreamSink.scala
```diff
@@ -0,0 +1,129 @@
...
+  * - No other writers for this outputStream may exist (IE no other component may register a writeListener)
+  * - The associated context must be put in to async mode, first
+  *
+  * Materialized Future will fail if the outputStream is not upgraded to async, or if a writeListener is already
+  * registered.
+  *
+  * Associated AsyncContext is not automatically closed; it is the responsible module using this sink to call
+  * asyncContext.complete() AFTER the materialized future is completed (failure or success).
+  */
+class ServletOutputStreamSink(asyncContext: AsyncContext) extends GraphStageWithMaterializedValue[SinkShape[ByteString], Future[Done]] with StrictLogging {
```
jeschkies: So, we cannot use `asJavaStrem` because it's not a simple `OutputStream` but a `ServletOutputStream`?

src/main/scala/mesosphere/marathon/io/IO.scala
```diff
@@ -45,7 +46,11 @@ object IO extends StrictLogging {
     try {
       fn(closeable)
     } finally {
-      IOUtils.closeQuietly(closeable)
+      try closeable.close()
```
jeschkies: What is the motivation here?

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
timcharper: If you search for the reference code (there is only one), this is the hostname verifier used if the ignore hostname authenticity Marathon configuration option is enabled

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
timcharper: That's a great suggestion

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
timcharper: We need to close the async context. Returning a failed future funnels all of the logic through a single place.

src/main/scala/mesosphere/marathon/api/forwarder/ServletOutputStreamSink.scala
timcharper: I don't understand, but yes, there are fundamental differences between OutputStream and ServletOutputStream.

src/main/scala/mesosphere/marathon/io/IO.scala
timcharper: IOUtils.closeQuietly is deprecated.

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
timcharper: We reply and return success.

Here is why I don't use async/await:

```
[info] Formatting 1 Scala source ProjectRef(uri("file:/Users/tim/src/m8e/marathon/"), "marathon")(compile) ...
[info] Compiling 1 Scala source to /Users/tim/src/m8e/marathon/target/scala-2.12/classes ...
[error] /Users/tim/src/m8e/marathon/src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala:167:35: await must not be used under a try/catch.
[error]           val leaderRequest = Try(await(createAndConfigureConnection(asyncContext, url, request)))
[error]                                   ^
[error] one error found
[error] (Compile / compileIncremental) Compilation failed
[error] Total time: 2 s, completed May 20, 2018 11:54:37 PM
marathon(tharper/async-proxy)> 
```

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
timcharper: ✅  on jax. There are too many packages for akka.

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
timcharper: On review, this exception catching was unnecessary since we have a general catch-all later.

src/main/scala/mesosphere/marathon/DeprecatedFeatures.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  val syncProxy = DeprecatedFeature(
+    "sync_proxy",
+    description = "Old, blocking IO implementation for leader proxy used by Marathon standby instances.",
+    warnVersion = "1.7.0",
+    removeVersion = "1.8.0")
+
+  val all = Seq(syncProxy)
+
+  private val versionOrdering: Ordering[List[Int]] = new Ordering[List[Int]] {
```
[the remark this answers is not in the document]
timcharper: compare a list of integers. There is no default implementation for this Seq[Int].. but TIL there is one for Iterable[Int]. Updating to use the stdlib implementation of this logic.

jeschkies:
Good Job!
Thank you 🙂!

src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -0,0 +1,208 @@
...
+import mesosphere.marathon.api.LeaderProxyFilter
+import mesosphere.marathon.stream.Implicits._
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+/**
+  * Forwarder which uses Akka HTTP to proxy requests, and then
```
zen-dog: and then... what?

PR #6195, opened 2018-05-08
Update apache curator dependency

zen-dog:
Summary:
- updated apache `curator-framework` to 4.0.1
- updated `curator-test` to 2.12.0


timcharper:
@zen-dog I remember in the past explicitly not upgrading curator-test because it introduced some failures. I wonder if these are the same failures that caused me to shy away before. I'll try and recover some more details.

PR #6196, opened 2018-05-08
Report flake8 output as GitHub check.

jeschkies:
Summary:
GitHub introduced a new API that allows to report checks with source
code annotations. This should ease finding the root cause of CI
failures.


jeschkies:
Maybe a different day.

PR #6197, opened 2018-05-08
Introduce HTTP input and output byte counters

wavesoft:
Summary:
This commit introduces a new filter that replaces the input and output streams
with a proxy stream that counts the bytes exchanged.

Related JIRA: MARATHON-8158

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
+++ b/src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
@@ -25,18 +25,18 @@ class HTTPMetricsFilter extends Filter {
     override def isReady: Boolean = upstream.isReady
     override def setWriteListener(writeListener: WriteListener): Unit = upstream.setWriteListener(writeListener)
     override def write(b: Int): Unit = {
-      bytes += 1
       upstream.write(b)
+      bytes += 1
     }
 
     override def write(b: Array[Byte]): Unit = {
-      bytes += b.length
       upstream.write(b)
+      bytes += b.length
     }
 
     override def write(b: Array[Byte], off: Int, len: Int): Unit = {
-      bytes += len
       upstream.write(b, off, len)
+      bytes += len
     }
   }
 
@@ -64,18 +64,27 @@ class HTTPMetricsFilter extends Filter {
     override def setReadListener(readListener: ReadListener): Unit = upstream.setReadListener(readListener)
     override def isFinished: Boolean = upstream.isFinished
     override def read(): Int = {
-      bytes += 1
-      upstream.read()
+      val byte = upstream.read()
+      if (byte != -1) {
+        bytes += 1
+      }
+      byte
     }
 
     override def read(b: Array[Byte]): Int = {
-      bytes += b.length
-      upstream.read(b)
+      val bytesRead = upstream.read(b)
+      if (bytesRead != -1) {
+        bytes += bytesRead
+      }
+      bytesRead
     }
 
     override def read(b: Array[Byte], off: Int, len: Int): Int = {
-      bytes += len
-      upstream.read(b, off, len)
+      val bytesRead = upstream.read(b, off, len)
+      if (bytesRead != -1) {
+        bytes += bytesRead
+      }
+      bytesRead
     }
   }
 
--- a/src/test/scala/mesosphere/marathon/integration/MetricsIntegrationTest.scala
+++ b/src/test/scala/mesosphere/marathon/integration/MetricsIntegrationTest.scala
@@ -22,9 +22,13 @@ class MetricsIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonTe
 
       And("The `outputBytes` is increased as expected")
       val currentCounter = result.entityJson("counters")("service.mesosphere.marathon.api.HTTPMetricsFilter.bytesWritten")("count").as[Int]
+
+      // Give some time to Kamon to take a metrics snapshot.
+      Thread.sleep(3000)
+
       val newResult = marathon.metrics()
       val newCounter = newResult.entityJson("counters")("service.mesosphere.marathon.api.HTTPMetricsFilter.bytesWritten")("count").as[Int]
-      newCounter should be(currentCounter + newResult.entityString.length)
+      newCounter shouldBe >=(currentCounter + result.entityString.length)
 
     }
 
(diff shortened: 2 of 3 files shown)


src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
```diff
@@ -0,0 +1,89 @@
...
+    * order to count the number of bytes that went through the stream.
+    *
+    * @param upstream The stream to proxy the output to
+    */
+  private class ProxyOutputStream(val upstream: ServletOutputStream) extends ServletOutputStream {
+    var bytes: Int = 0
+
+    override def isReady: Boolean = upstream.isReady
+    override def setWriteListener(writeListener: WriteListener): Unit = upstream.setWriteListener(writeListener)
+    override def write(b: Int): Unit = {
```
ichernetsky: Are you sure that bytes are read/written one by one? What about overriding implementations of other read/write methods like `read(byte[] b)`, `read(byte[] b, int off, int len)`, `readLine(byte[] b, int off, int len)` or so on?

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
```diff
@@ -0,0 +1,89 @@
...
+  private[this] val inputBytesMetric = Metrics.counter(ServiceMetric, getClass, "inputBytes")
+  private[this] val outputBytesMetric = Metrics.counter(ServiceMetric, getClass, "outputBytes")
+
+  /**
+    * Wraps a `ServletOutputStream` and overrides the `write` method in
+    * order to count the number of bytes that went through the stream.
+    *
+    * @param upstream The stream to proxy the output to
+    */
+  private class ProxyOutputStream(val upstream: ServletOutputStream) extends ServletOutputStream {
```
alenkacz: I would maybe mention the word metrics or Measured or something, proxy is very generic, this is basically a Decorator to collect metrics...

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
wavesoft: Well, the `read(): Int` and `write(b: Int)` are the only abstract methods of the base stream classes that **must** be implemented. Therefore I assumed that the default implementation of the other signatures is implemented by calling out to these base functions. And I confirmed this by checking the counters while polling various HTTP endpoints.

However, now that I think of it, the stream implementations _might_ provide a more efficient implementation for operations on byte ranges. So it actually makes sense to implement the other signatures of `read` and `write`. (But `readLine` does not really benefit from this.)

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
wavesoft: True, the term `Proxy` can be confusing 👍

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
```diff
@@ -0,0 +1,119 @@
...
+  private[this] val inputBytesMetric = Metrics.counter(ServiceMetric, getClass, "inputBytes")
+  private[this] val outputBytesMetric = Metrics.counter(ServiceMetric, getClass, "outputBytes")
+
+  /**
+    * Wraps a `ServletOutputStream` and overrides the `write` method in
+    * order to count the number of bytes that went through the stream.
+    *
+    * @param upstream The stream to proxy the output to
+    */
+  private class OutputStreamCounter(val upstream: ServletOutputStream) extends ServletOutputStream {
```
jeschkies: Well, I hate to break this to you but at least Apache Commons IO has this features: https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/input/CountingInputStream.html.

wavesoft: But the object I have to wrap is not an `OuptuStream`, it's a `ServletOutputStream` :trollface:

jeschkies: Hm, there must be an easier way. `ServletOutputStream` is actually not doing much.

jeschkies: Ah man. I was so hopeful ^^.

timcharper: @jeschkies these other methods are necessary for async IO. See https://github.com/mesosphere/marathon/pull/6193/files#diff-52da5dcde3d226b130f1eff2c1c29c2b

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
timcharper: My understanding matches your suspicions, @wavesoft : all other methods call out to write(int).

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
```diff
@@ -0,0 +1,119 @@
...
+
+    override def read(b: Array[Byte], off: Int, len: Int): Int = {
+      bytes += len
+      upstream.read(b, off, len)
+    }
+  }
+
+  private class RequestCounterWrapper(val r: HttpServletRequest) extends HttpServletRequestWrapper(r) {
+    private lazy val stream = new InputStreamCounter(super.getInputStream)
+    def bytes: Int = stream.bytes
```
timcharper: Different name, perhaps? `bytesRead`, `bytesWritten`.

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
```diff
@@ -0,0 +1,119 @@
...
+import javax.servlet._
+import javax.servlet.http.{HttpServletRequest, HttpServletRequestWrapper, HttpServletResponse, HttpServletResponseWrapper}
+import mesosphere.marathon.metrics.{Metrics, ServiceMetric}
+
+/**
+  * This filter replaces the default I/O streams with proxies that count
+  * the number of bytes that are sent to and from the client.
+  */
+class HTTPMetricsFilter extends Filter {
+  private[this] val inputBytesMetric = Metrics.counter(ServiceMetric, getClass, "inputBytes")
```
timcharper: `bytesRead`, `bytesWritten`?

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
wavesoft: Not a bad idea 👍

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
wavesoft: Not a bad idea 👍

src/main/scala/mesosphere/marathon/api/HTTPMetricsFilter.scala
```diff
@@ -0,0 +1,128 @@
...
+    override def read(b: Array[Byte]): Int = {
+      val bytesRead = upstream.read(b)
+      if (bytesRead != -1) {
+        bytes += bytesRead
+      }
+      bytesRead
+    }
+
+    override def read(b: Array[Byte], off: Int, len: Int): Int = {
+      val bytesRead = upstream.read(b, off, len)
```
wavesoft: Ha! Nice catch 😄

PR #6198, opened 2018-05-08
Fix the behaviour of the GcActor in marathon

meln1k:
Summary: implemented backpressured batched GC scans, so we don't consume too much memory at once. It stabilize marathon behaviour with lots of deployments.

JIRA issues: MARATHON-8195

### DIFF
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -61,3 +61,5 @@ kamon {
     value-check-interval-ms: 1000
   }
 }
+
+gc-actor-scan-batch-size = 32
--- a/src/main/resources/application.conf
+++ b/src/main/resources/application.conf
@@ -61,5 +61,3 @@ kamon {
     value-check-interval-ms: 1000
   }
 }
-
-gc-actor-scan-batch-size = 32
--- a/src/main/scala/mesosphere/marathon/storage/StorageConf.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageConf.scala
@@ -28,8 +28,8 @@ trait StorageConf extends ZookeeperConf with BackupConf {
     default = Some(50)
   )
 
-  lazy val scanBatchSize = opt[Int](
-    "gc_scan_batch_size", // while called Zk, applies to every store but the name is kept
+  lazy val gcActorScanBatchSize = opt[Int](
+    "gc_actor_scan_batch_size",
     descr = "Size of GC actor scan batches.",
     default = Some(32)
   )
--- a/src/main/scala/mesosphere/marathon/storage/StorageConf.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageConf.scala
@@ -28,6 +28,12 @@ trait StorageConf extends ZookeeperConf with BackupConf {
     default = Some(50)
   )
 
+  lazy val scanBatchSize = opt[Int](
+    "gc_scan_batch_size", // while called Zk, applies to every store but the name is kept
+    descr = "Size of GC actor scan batches.",
+    default = Some(32)
+  )
+
   lazy val zkMaxConcurrency = opt[Int](
     "zk_max_concurrency",
     default = Some(32),
--- a/src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
+++ b/src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
@@ -103,6 +103,7 @@ case class CuratorZk(
     maxConcurrent: Int,
     maxOutstanding: Int,
     maxVersions: Int,
+    scanBatchSize: Int,
     versionCacheConfig: Option[VersionCacheConfig],
     availableFeatures: Set[String],
     lifecycleState: LifecycleState,
@@ -167,6 +168,7 @@ object CuratorZk {
       maxConcurrent = conf.zkMaxConcurrency(),
       maxOutstanding = Int.MaxValue,
       maxVersions = conf.maxVersions(),
+      scanBatchSize = conf.scanBatchSize(),
       versionCacheConfig = if (conf.versionCacheEnabled()) StorageConfig.DefaultVersionCacheConfig else None,
       availableFeatures = conf.availableFeatures,
       backupLocation = conf.backupLocation.get,
@@ -196,6 +198,7 @@ object CuratorZk {
       maxConcurrent = config.int("max-concurrent-requests", 32),
       maxOutstanding = config.int("max-concurrent-outstanding", Int.MaxValue),
       maxVersions = config.int("max-versions", StorageConfig.DefaultMaxVersions),
+      scanBatchSize = config.int("scan-batch-size", StorageConfig.DefaultScanBatchSize),
       versionCacheConfig =
         if (config.bool("version-cache-enabled", true)) StorageConfig.DefaultVersionCacheConfig else None,
       availableFeatures = config.stringList("available-features", Seq.empty).to[Set],
@@ -208,6 +211,7 @@ object CuratorZk {
 
 case class InMem(
     maxVersions: Int,
+    scanBatchSize: Int,
     availableFeatures: Set[String],
     defaultNetworkName: Option[String],
     backupLocation: Option[URI]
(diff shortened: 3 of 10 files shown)


src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
```diff
@@ -70,6 +72,7 @@ private[storage] class GcActor[K, C, S](
     val maxVersions: Int)(implicit val mat: Materializer, val ctx: ExecutionContext)
   extends FSM[State, Data] with LoggingFSM[State, Data] with ScanBehavior[K, C, S] with CompactBehavior[K, C, S] {
 
+  val ScanBatchSize = Try(context.system.settings.config.getInt("gc-actor-scan-batch-size")).toOption.getOrElse(32)
```
kensipe: why is this proper cased?  and not `scanBatchSize`?
and why is this not passed in the constructor like all the other values?
If passed via the constructor other values can be passed... and testing will be possible without need to setup details in the config object.

meln1k:
Let's implement `--gc_pause_duration` in a separate PR

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -196,6 +198,7 @@ object CuratorZk {
       maxConcurrent = config.int("max-concurrent-requests", 32),
       maxOutstanding = config.int("max-concurrent-outstanding", Int.MaxValue),
       maxVersions = config.int("max-versions", StorageConfig.DefaultMaxVersions),
+      scanBatchSize = config.int("scan-batch-size", StorageConfig.DefaultScanBatchSize),
```
zen-dog: Why is here a third `scan-batch-size` parameter which is neither `gc_scan_batch_size` nor `gc-actor-scan-batch-size`?

src/main/resources/application.conf
```diff
@@ -60,4 +60,6 @@ kamon {
     identify-interval-ms: 1000
     value-check-interval-ms: 1000
   }
-}
\ No newline at end of file
+}
+
+gc-actor-scan-batch-size = 32
```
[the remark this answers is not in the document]
jeschkies: That's a good question. At some point we should rethink Marathons configuration. I really like Lightbend's Config. However, we are stuck with Scallop for now.

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -28,6 +28,12 @@ trait StorageConf extends ZookeeperConf with BackupConf {
     default = Some(50)
   )
 
+  lazy val scanBatchSize = opt[Int](
```
jeschkies: How does this relate to `gc-actor-scan-batch-size` in the config file above?

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -240,6 +245,7 @@ object StorageConfig {
 
   val DefaultLegacyMaxVersions = 25
   val DefaultMaxVersions = 5000
+  val DefaultScanBatchSize = 32
```
jeschkies: 🤔 isn't the Scallop option already defining a default?

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
```diff
@@ -291,65 +294,68 @@ private[storage] trait ScanBehavior[K, C, S] extends StrictLogging { this: FSM[S
...
+    }
 
     def appsExceedingMaxVersions(usedApps: Set[PathId]): Future[Map[PathId, Set[OffsetDateTime]]] = {
-      Future.sequence {
-        usedApps.map { id =>
-          appRepository.versions(id).runWith(Sink.sortedSet).map(id -> _)
-        }
-      }.map(_.filter(_._2.size > maxVersions).toMap)
+      Source(usedApps)
+        .mapAsync(1)(id => appRepository.versions(id).runWith(Sink.sortedSet).map(id -> _))
```
jeschkies: Hm, isn't the something like a flat map for a source of sources? I mean why do we use a sink for each app?

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
```diff
@@ -291,65 +294,68 @@ private[storage] trait ScanBehavior[K, C, S] extends StrictLogging { this: FSM[S
...
     }
 
     def podsExceedingMaxVersions(usedPods: Set[PathId]): Future[Map[PathId, Set[OffsetDateTime]]] = {
-      Future.sequence {
-        usedPods.map { id =>
-          podRepository.versions(id).runWith(Sink.sortedSet).map(id -> _)
-        }
-      }.map(_.filter(_._2.size > maxVersions).toMap)
+      Source(usedPods)
+        .mapAsync(1)(id => podRepository.versions(id).runWith(Sink.sortedSet).map(id -> _))
```
jeschkies: See above.

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
```diff
@@ -291,65 +294,68 @@ private[storage] trait ScanBehavior[K, C, S] extends StrictLogging { this: FSM[S
...
-        case (id, versions) =>
-          val candidateVersions = versions.diff(usedApps.getOrElse(id, SortedSet.empty))
-          id -> candidateVersions.take(versions.size - maxVersions)
-      }
+    val allAppIdsFuture = appRepository.ids().runWith(Sink.set)
+    val allPodIdsFuture = podRepository.ids().runWith(Sink.set)
+
+    rootsInUse()
+      .grouped(scanBatchSize)
+      .mapAsync(1) { inUseRoots =>
```
jeschkies: Could you annotate the size of `inUseRoots`? Is this a batch?

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
```diff
@@ -291,65 +294,68 @@ private[storage] trait ScanBehavior[K, C, S] extends StrictLogging { this: FSM[S
...
+        async { // linter:ignore UnnecessaryElseBranch
+          val allAppIds = await(allAppIdsFuture)
+          val allPodIds = await(allPodIdsFuture)
+          val usedApps = appsInUse(inUseRoots)
+          val usedPods = podsInUse(inUseRoots)
+          val appsWithTooManyVersions = await(appsExceedingMaxVersions(usedApps.keySet))
+          val podsWithTooManyVersions = await(podsExceedingMaxVersions(usedPods.keySet))
+
+          val appVersionsToDelete = appsWithTooManyVersions.map {
```
jeschkies: Isn't this a `mapValues`?

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
meln1k: well, it's not for the scallop, but rather for typesafe config which is already here. You can check the line 236

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
meln1k: We use it because we need a strict collection here.

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
meln1k: Yes it is.

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
meln1k: Not really, since we need the keys too.

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
meln1k: Hm, I found a way to do some pure streaming here.

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
jeschkies: So why not pull in the default from the Lightbend config?

src/main/scala/mesosphere/marathon/storage/repository/GcActor.scala
jeschkies: Ah, in `usedApps.getOrElse(id, SortedSet.empty)`. Got it.

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
zen-dog: This whole method is never used. I guess at some point someone wanted to switch to `typesafe.Config` eventually? I'd rather remove it completely in a separate PR. It's dead code anyway.

PR #6199, opened 2018-05-09
Measure group update queue size MARATHON-8197

alenkacz:
JIRA issues: MARATHON-8197

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
@@ -50,7 +50,7 @@ class GroupManagerImpl(
   private[this] val root = LockedVar(initialRoot)
 
   private[this] val dismissedDeploymentsMetric = Metrics.counter(ServiceMetric, getClass, "dismissedDeployments")
-  private[this] val groupUpdateSizeCounter = Metrics.minMaxCounter(ServiceMetric, getClass, "queueLength")
+  private[this] val groupUpdateSizeMetric = Metrics.minMaxCounter(ServiceMetric, getClass, "queueLength")
 
   @SuppressWarnings(Array("OptionGet"))
   override def rootGroup(): RootGroup =
@@ -113,7 +113,7 @@ class GroupManagerImpl(
     change: (RootGroup) => Future[Either[T, RootGroup]],
     version: Timestamp, force: Boolean, toKill: Map[PathId, Seq[Instance]]): Future[Either[T, DeploymentPlan]] = try {
 
-    groupUpdateSizeCounter.record(serializeUpdates.size())
+    groupUpdateSizeMetric.record(serializeUpdates.size())
 
     // All updates to the root go through the work queue.
     val maybeDeploymentPlan: Future[Either[T, DeploymentPlan]] = serializeUpdates {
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -49,6 +50,7 @@ class GroupManagerImpl(
   private[this] val root = LockedVar(initialRoot)
 
   private[this] val dismissedDeploymentsMetric = Metrics.counter(ServiceMetric, getClass, "dismissedDeployments")
+  private[this] val groupUpdateSizeCounter = Metrics.minMaxCounter(ServiceMetric, getClass, "queueLength")
```
wavesoft: I am not strong on this, but I would propose to use the suffix `Metric` on all metric-related variables. So `groupUpdateSizeMetric`

alenkacz: I don't like that much... but I am not strong on this either, so we probably need someone third :D

zen-dog: it's been `Metric` at the end so far, correct? Either rename them all to smth. else or keep it.

alenkacz: ok ... to me it's like naming integer properties `AbcInt` and String properties 'AbcString' but whatever :) ... changed

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -49,6 +50,7 @@ class GroupManagerImpl(
   private[this] val root = LockedVar(initialRoot)
 
   private[this] val dismissedDeploymentsMetric = Metrics.counter(ServiceMetric, getClass, "dismissedDeployments")
+  private[this] val groupUpdateSizeMetric = Metrics.minMaxCounter(ServiceMetric, getClass, "queueLength")
```
kensipe: "queueLength"...  what do you think of  "queueSize" or "currentQueueSize"?
If we want to be consistent with deployments we would call it "currentQueueCount"

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
```diff
@@ -111,6 +113,8 @@ class GroupManagerImpl(
     change: (RootGroup) => Future[Either[T, RootGroup]],
     version: Timestamp, force: Boolean, toKill: Map[PathId, Seq[Instance]]): Future[Either[T, DeploymentPlan]] = try {
 
+    groupUpdateSizeMetric.increment()
```
zen-dog: You could just ask the queue for the size instead of counting up and down ;) E.g. see here:
https://github.com/mesosphere/marathon/commit/2d217b46e19ad3f6ec8d80cd8023e558207e5235#diff-963e559f5c3864cf5e3971eaa415662dR119

alenkacz: Yeah, I did that in previous version of this PR... but minmax counter does not have method for setting value and I think I can reliably do it by incrementing and decrementing

src/main/scala/mesosphere/marathon/core/group/impl/GroupManagerImpl.scala
alenkacz: queueSize is my favorite ... thanks guys!

PR #6208, opened 2018-05-11
Do not leak ZK credentials in /v2/info

ichernetsky:
JIRA issues: MARATHON-7568

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala
@@ -45,7 +45,7 @@ class InfoResource @Inject() (
     "leader_proxy_read_timeout_ms" -> config.leaderProxyReadTimeout.toOption,
     "local_port_max" -> config.localPortMax.toOption,
     "local_port_min" -> config.localPortMin.toOption,
-    "master" -> s"zk://${config.mesosMasterZooKeeperHosts}${config.mesosMasterZooKeeperPath}",
+    "master" -> config.mesosMaster().redactedConnectionString,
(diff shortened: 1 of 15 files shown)


src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala
```diff
@@ -44,7 +44,7 @@ class InfoResource @Inject() (
     "leader_proxy_read_timeout_ms" -> config.leaderProxyReadTimeout.get,
     "local_port_max" -> config.localPortMax.get,
     "local_port_min" -> config.localPortMin.get,
-    "master" -> config.mesosMaster.get,
+    "master" -> s"zk://${config.mesosMasterZooKeeperHosts}${config.mesosMasterZooKeeperPath}",
```
timcharper: Would it be better to obfuscate it? So that it was clear that credentials were specified?

ichernetsky: I don't see how it can be helpful. What use-case do you have in mind?

ichernetsky:
A test has been added. PTAL.

src/main/scala/mesosphere/marathon/api/v2/InfoResource.scala
timcharper: Wow, this fell through the cracks. Obfuscating (xxxxxxx) might help communicate that authentication is used.

jeschkies:
@ichernetsky what is the status?

timcharper:
I would really like us to indicate if zk authentication is being used, somehow. Perhaps showing `zk://xxxxxxxx@10.0.4.3,10.0.4.4,10.0.5.5:2181/marathon`

PR #6212, opened 2018-05-11
Fix ZK layout migration when store caching is disabled

ichernetsky:
JIRA issues: MARATHON-8064

### DIFF
--- a/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala
+++ b/src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala
@@ -8,13 +8,15 @@ import akka.stream.Materializer
 import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.Protos._
-import mesosphere.marathon.api.v2.{AppNormalization, AppHelpers}
+import mesosphere.marathon.api.v2.{AppHelpers, AppNormalization}
 import mesosphere.marathon.core.async.ExecutionContexts
 import mesosphere.marathon.core.storage.store.PersistenceStore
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala
```diff
@@ -138,9 +140,14 @@ private[migration] object MigrationTo15 {
   def deleteEventSubscribers[K, C, S](store: PersistenceStore[K, C, S]): Future[Done] = {
     store match {
       case zk: ZkPersistenceStore =>
-        zk.client.delete("/event-subscribers").map(_ => Done)(ExecutionContexts.callerThread)
-      case _ =>
-        Future.successful(Done)
+        implicit val ec: ExecutionContext = ExecutionContexts.callerThread
+        zk.client.delete("/event-subscribers").map(_ => Done).recover {
+          case _: KeeperException.NoNodeException => Done
```
alenkacz: I understand the fix but I don't seem to understand how is that connected to the `disable_store_cache`. Can you clarify?

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala
```diff
@@ -138,9 +140,14 @@ private[migration] object MigrationTo15 {
...
+        }
+      case cachedStore: LazyCachingPersistenceStore[K, C, S] => deleteEventSubscribers(cachedStore.store)
```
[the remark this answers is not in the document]
ichernetsky: No, there isn't. Yes, you are right.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo15.scala
ichernetsky: If caching was enabled it was doing nothing, and when it was disabled, if there was no `/event-subscribers`, it crashed.

PR #6214, opened 2018-05-12
Switch to Scalatest JUnitXMLReporter to regain the timestamp field

timcharper:
SBT 1.x ships with a JUnit XML reporter which is enabled, by
default. Unfortunately, this reporter is missing a field which is
expected by Jenkins: timestamp. We use this field in our test analysis
tool to reason about which tests ran in parallel.

Switching to the Scalatest JUnitXML reporter brings the field back for
the time being.

See https://github.com/sbt/sbt/issues/4153

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,6 +24,7 @@ ansiColor('xterm') {
         }
       } finally {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
+        junit(allowEmptyResults: true, testResults: 'target/test-reports/integration/*.xml')
         publishHTML([
             allowMissing: true, alwaysLinkToLastBuild: false, keepAll: true,
             reportDir: 'target/scala-2.12/scapegoat-report', reportFiles: 'scapegoat.html',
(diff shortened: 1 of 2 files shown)


jeschkies:
Hm, it seems the current master build is also not showing the unit test results. The `releases/1.5` branch has no test results either.

Also, the stats script has now issues on master but on 1.5.

timcharper:
Weird, I'll have a look. They were showing up locally.

timcharper:
I have confirmed that the latest build includes both unit and integration test results:

https://jenkins.mesosphere.com/service/jenkins/job/marathon-pipelines/job/PR-6214/3/testReport/

timcharper:
Also, I have confirmed that the `timestamp` is no longer null

https://jenkins.mesosphere.com/service/jenkins/job/marathon-pipelines/job/PR-6214/3/testReport/api/json

Versus

https://jenkins.mesosphere.com/service/jenkins/job/marathon-pipelines/job/PR-6213/1/testReport/api/json

PR #6217, opened 2018-05-14
Fix master loop failure metric

zen-dog:
Summary:
- fixed `marathon_loop_master.failure` metric which was previously sent in any case (instead of failure case)
- added `marathon_loop_master.duration` to the metrics

### DIFF
--- a/ci/pipeline
+++ b/ci/pipeline
@@ -253,12 +253,15 @@ def pr(pullId: String): Unit = asPullRequest(pullId) {
  */
 @main
 def loop(): Unit = {
+  val start = System.currentTimeMillis
   try {
     provisionHost()
     build(buildName = utils.loopBuildName())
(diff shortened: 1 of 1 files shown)


ci/pipeline
```diff
@@ -253,12 +253,15 @@ def pr(pullId: String): Unit = asPullRequest(pullId) {
...
 def loop(): Unit = {
+  val start = System.currentTimeMillis
   try {
     provisionHost()
     build(buildName = utils.loopBuildName())
     dataDogClient.reportCount(s"marathon.build.${utils.loopName}.success", 1)
+  } catch {
+    case _ => dataDogClient.reportCount(s"marathon.build.${utils.loopName}.failure", 1)
   } finally {
-    dataDogClient.reportCount(s"marathon.build.${utils.loopName}.failure", 1)
```
timcharper: :trollface:

PR #6218, opened 2018-05-14
Use Apache Curator `RetryPolicy`

zen-dog:
Summary:
we were using our custom `Retry` logic in the `ZKPersistenceStore` class. This change replaces it with Curator `BoundedExponentialBackoffRetry` policy.

Additionally:
- simplified `RichCuratorFramework.blockUntilConnected` method - an exception is immediately thrown if the connection to ZK fails
- introduced 2 new parameters to control the retry policy: `zk_operation_max_retries` and `zk_operation_base_retry_sleep_ms`
- removed some dead code


zen-dog:
I've crated a loop [here](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-aleksey/). So far 25 green builds so seems to work quite good.

wavesoft:
compare-dump-201805181512-v1 6 463-local-dcluster-n-apps-1-instances-summ

And is the performance diff (green = this branch, orange = master)

PR #6224, opened 2018-05-15
Adding launchOperations, launchGroupOperations, reserveOperations

wavesoft:
This commit introduces the counters on the `OfferOperationFactory` that counts the
launch, launch_group and reserve mesos operations of marathon.

JIRA Issues: MARATHON-8213


src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -13,6 +14,10 @@ class OfferOperationFactory(
     private val principalOpt: Option[String],
     private val roleOpt: Option[String]) {
 
+  private[this] val launchOperationsMetric = Metrics.counter(ServiceMetric, getClass, "launchOperations")
```
kensipe: should we call these "launchCount" or "launchOperationCount"?

PR #6225, opened 2018-05-15
Switch to Scalatest JUnitXMLReporter to regain the timestamp field

timcharper:
Backport of #6214

Summary:
SBT 1.x ships with a JUnit XML reporter which is enabled, by
default. Unfortunately, this reporter is missing a field which is
expected by Jenkins: timestamp. We use this field in our test analysis
tool to reason about which tests ran in parallel.

Switching to the Scalatest JUnitXML reporter brings the field back for
the time being.

See https://github.com/sbt/sbt/issues/4153


timcharper:
I have confirmed that both the unit test and the integration test results are present

PR #6226, opened 2018-05-15
Group versions cache size config validation

meln1k:
we should validate group versions cache config size


tools/elk/bin/target
```diff
@@ -1,5 +0,0 @@
-#!/bin/bash
```
ichernetsky: This file should be restored.

ichernetsky:
I second what Tim suggested.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -339,5 +339,16 @@ trait MarathonConf
     default = Some(GpuSchedulingBehavior.Undefined),
     validate = validateGpuSchedulingBehavior
   )
+
+  //validations
```
jeschkies: Nit: I'm not sure this comment is adding value 🙂

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
```diff
@@ -252,11 +221,4 @@ object StorageConfig {
       case CuratorZk.StoreName => CuratorZk(conf, lifecycleState)
     }
   }
-
-  def apply(conf: Config, lifecycleState: LifecycleState): StorageConfig = {
```
jeschkies: I thought @zen-dog removed this already.

src/main/scala/mesosphere/marathon/storage/repository/Repositories.scala
```diff
@@ -71,7 +72,7 @@ object GroupRepository {
     ctx: ExecutionContext,
     mat: Materializer): StoredGroupRepositoryImpl[RamId, String, Identity] = {
     import mesosphere.marathon.storage.store.InMemoryStoreSerialization._
-    new StoredGroupRepositoryImpl(store, appRepository, podRepository)
+    new StoredGroupRepositoryImpl(store, appRepository, podRepository, 1000)
```
jeschkies: Where is this magic number coming from?

src/test/scala/mesosphere/marathon/storage/repository/GroupRepositoryTest.scala
```diff
@@ -161,7 +161,7 @@ class GroupRepositoryTest extends AkkaUnitTest with Mockito with ZookeeperServer
   def createZkRepos(appRepository: AppRepository, podRepository: PodRepository, maxVersions: Int): GroupRepository = { // linter:ignore:UnusedParameter
     val store = zkStore
     store.markOpen()
-    GroupRepository.zkRepository(store, appRepository, podRepository)
+    GroupRepository.zkRepository(store, appRepository, podRepository, 1000)
```
jeschkies: See above.

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -34,6 +34,12 @@ trait StorageConf extends ZookeeperConf with BackupConf {
     default = Some(32)
   )
 
+  lazy val groupVersionsCacheSize = opt[Int](
```
jeschkies: What value did we use before? I cannot find it in this diff.

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -339,5 +339,16 @@ trait MarathonConf
...
+
+  //  we must set gc actor threshold to not trigger it too early, 2 * max deployments looks like a good default
```
zen-dog: What is the justification behind `2 * max deployments` formula? Why not `max deployments + 5`?

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
```diff
@@ -34,6 +34,12 @@ trait StorageConf extends ZookeeperConf with BackupConf {
...
+    "group_versions_cache_size",
+    descr = "Size of the groups repository versions cache.",
+    default = Some(1000)
```
zen-dog: How does this default correspond to the above-mentioned formula? Why is it `100`?

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
zen-dog: Wait, why is it a new parameter at all? GCActor does use `maxVersions` variable defined above for exactly these purpose, doesn't it?

src/main/scala/mesosphere/marathon/storage/StorageConfig.scala
zen-dog: I did!

src/main/scala/mesosphere/marathon/MarathonConf.scala
meln1k: because every deployment is a `root group before` + `root group after`.

src/main/scala/mesosphere/marathon/storage/repository/Repositories.scala
meln1k: it was a default parameter in the GroupRepositoryImpl before:
image

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
meln1k: backwards-compatibility

zen-dog: But that value doesn't make any sense since we cap it earlier now?

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
meln1k: there are 2 parameters: GC actor's and repo cache size. They're independent

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
zen-dog: Ok, after discussing it in person, that makes sense for the backward compatibility.

src/main/scala/mesosphere/marathon/storage/repository/Repositories.scala
zen-dog: You should pass `versionCacheMaxSize` here too

src/main/scala/mesosphere/marathon/storage/StorageConf.scala
jeschkies: Could you share the answer here?

zen-dog: Basically, one can already have 999 groups there which are all used by some deployments. If you change them to 100 here, the GCActor will start to work constantly. Hence, it should stay 1000, at least for now and might get a more sensible default in the next version.

PR #6231, opened 2018-05-16
Add DataDog events for our SI test builds for every cluster variant

zen-dog:


zen-dog:
> Ah, that was too quick from me. Could you add string(credentialsId: '66c40969-a46d-470e-b8a2-6f04f2b3f2d5',variable: 'DATADOG_API_KEY') to the Jenkins files for the loops?

Good catch. Done!

PR #6232, opened 2018-05-17
Fix `si_pipeline` script

zen-dog:
Summary:
- fixed path for calling DataDog client
- fixed collecting of DC/OS diagnostics bundle

### DIFF
--- a/ci/si_install_deps.sh
+++ b/ci/si_install_deps.sh
@@ -17,8 +17,8 @@ fi
 
 # Ensure amm is available.
 if ! command -v amm >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


zen-dog:
It still doesn't work since the underlying image doesn't have `amm` 🤦‍♂️

ci/si_pipeline.sh
```diff
@@ -76,10 +76,10 @@ case $CLUSTER_LAUNCH_CODE in
       make test
       SI_CODE=$?
       if [ ${SI_CODE} -gt 0 ]; then
-        ./ci/dataDogClient.sc "marathon.build.si.$VARIANT.failure" 1
+        ../ci/dataDogClient.sc "marathon.build.si.$VARIANT.failure" 1
```
alenkacz: why not just ./dataDogClient.sc? since you are in ci folder probably?

PR #6235, opened 2018-05-18
Fix Mesos HTTP health check for non-host networking mode with `containerPort=0`

zen-dog:
Summary:
Fixed a bug where for a container in non-host networking mode (`container` and `container/bridged`) with `hostPort=0, containerPort=0` we would set the `containerPort == hostPort` in `ContainerInfo` *and* ports environment variables *but not* in the `NetworkInfo`.
As the result, Mesos health checks in this scenario would try to probe `containerPort == 0`. This behavior has now been unified for `NetworkInfo`, `ContainerInfo` and generated ports environment variables resulting in fixed Mesos HTTP health checks.

JIRA issues: MARATHON-8216


zen-dog:
Overall I'm not happy with how we have the same logic in three different places (see my comment in the `TaskBuilder`) but I haven't found an elegant solution for this. Every re-factoring I started would bring too many changes with it (having `portMappings` leave in the `Container`, `portDefinitions` in the `RunSpec` and the actual ports coming either from the offer or later from the `NetworkInfo`)

PR #6236, opened 2018-05-18
Remove the 'possibleNewState' as it is used in tests only

alenkacz:
JIRA issues: MARATHON-8167


alenkacz:
let's not merge until we review PR 8161

PR #6237, opened 2018-05-18
Run SI test builds every 2h

zen-dog:
Summary:
Updated the SI pipeline triggers to run roughly every 2h. However, we use Jenkins `H` cron job [option](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/hudson/triggers/TimerTrigger/help-spec.jelly#L40) for Jenkins to try to spread the load evenly. Overall we should end up with running approx. two SI jobs at any time which shouldn't be a problem.


zen-dog:
> @zen-dog the goal of this is to run pipelines more often or ...? Because so far it was every 6h, right?

yep, I'd like to have ~100 runs per week. It's not quite but close enough I guess.

PR #6238, opened 2018-05-21
Get rid of the InstanceOpProcessor class

alenkacz:
This is a follow-up of my previous refactoring (moving repository update call to instance tracker). InstanceUpdateOpProcessor was now containing very little code so it can be removed altogether.

JIRA issues: MARATHON-8177


src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActorTest.scala
```diff
@@ -302,6 +302,30 @@ class InstanceTrackerActorTest extends AkkaUnitTest {
     probe.expectMsg(appDataMap)
   }
 
+  "acknowledges and replies to the correct actors" in {
```
alenkacz: I just added a test because I wanted to be sure this really works as weirdly as it does now... The current flow is

Delegate -? InstanceTracker -> UpdateActor -? InstanceTracker 

And from the last ask call to InstanceTracker we mark as finished the ask query from delegate as well as UpdateActor...

I think that UpdateActor should ACK by itself when the update operation finishes (InstanceTracker processes the update and reply).

I will do this in another PR...

alenkacz:
the loop is running here https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/alena-loop-instancetracker/ (so far no failures)

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -35,7 +35,10 @@ object InstanceTrackerActor {
   private[impl] case class Get(instanceId: Instance.Id)
 
   /** Forward an update operation to the child [[InstanceUpdateActor]]. */
-  private[impl] case class ForwardTaskOp(deadline: Timestamp, instanceId: Instance.Id, op: InstanceUpdateOperation)
+  private[impl] case class UpdateContext(deadline: Timestamp, op: InstanceUpdateOperation) {
```
jeschkies: Much better name but I'm not sure the docs are still good 😉

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -137,9 +140,8 @@ private[impl] class InstanceTrackerActor(
       case InstanceTrackerActor.Get(instanceId) =>
         sender() ! instancesBySpec.instance(instanceId)
 
-      case ForwardTaskOp(deadline, instanceId, instanceUpdateOp) =>
-        val op = InstanceOpProcessor.Operation(deadline, sender(), instanceId, instanceUpdateOp)
-        updaterRef.forward(InstanceUpdateActor.ProcessInstanceOp(op))
+      case update: UpdateContext =>
+        updaterRef.forward(update)
```
jeschkies: Oh wow. There is an `InstanceUpdateActor`? Crazy.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceUpdateActor.scala
```diff
@@ -4,29 +4,35 @@ package core.task.tracker.impl
...
-import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.{ExecutionContext, Future}
 
 object InstanceUpdateActor {
-  def props(clock: Clock, metrics: ActorMetrics, processor: InstanceOpProcessor): Props = {
-    Props(new InstanceUpdateActor(clock, metrics, processor))
+  def props(clock: Clock, metrics: ActorMetrics,
+    instanceTrackerRef: ActorRef,
+    stateOpResolver: InstanceUpdateOpResolver,
```
jeschkies: This makes so much more sense to me now. I actually inject such logic with a trait to make it testable: https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/task/jobs/impl/ExpungeOverdueLostTasksActor.scala#L20. What do you think?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceUpdateActor.scala
```diff
@@ -125,17 +135,49 @@ private[impl] class InstanceUpdateActor(
...
         if (op.deadline <= clock.now()) {
           metrics.timedOutOpsMeter.increment()
-          op.sender ! Status.Failure(
+          queuedItem.sender ! Status.Failure(
             new TimeoutException(s"Timeout: ${op.op} for app [${op.appId}] and ${op.instanceId}.")
           )
           Future.successful(())
         } else
-          metrics.processOpTimer(processor.process(op))
+          metrics.processOpTimer(processUpdate(queuedItem))
```
jeschkies: Yes. This is way easier to follow. Thanks!

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
timcharper: ➕ 1

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -74,7 +74,7 @@ private[tracker] class InstanceTrackerDelegate(
 
     val instanceId: Instance.Id = stateOp.instanceId
     val deadline = clock.now + instanceTrackerQueryTimeout.duration
-    val op: ForwardTaskOp = InstanceTrackerActor.ForwardTaskOp(deadline, instanceId, stateOp)
+    val op: UpdateContext = InstanceTrackerActor.UpdateContext(deadline, stateOp)
```
timcharper: I'd go ahead and get rid of this explicit type

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActorTest.scala
meichstedt: I might be wrong, but I think this can be detangled even further (in that follow-up PR you mentioned). Questions that might help detangling:

* What's the main purpose of the InstanceUpdateActor?
* What is needed to resolve the effect of a given operation? (hint: the current state, loaded from … the instanceTracker)
* What does the InstanceUpdateActor do, apart from putting updated into a byInstance-Queue?
* What would prevent quite some read requests (routed via the InstanceTrackerDelegate)?

alenkacz: fair points (noted). In the next iteration I would like to get rid of the Ack class, but there's much more simplifications that can be done on this front, there it lot of back and forth communication for sure as you said...

And yes the instanceupdateactor really does not do anything else than making sure updates of the same instance does not happen in parallel, maybe we can get rid of this guy altogether and use streams instead. Will have to think about it...

PR #6239, opened 2018-05-22
Move integration tests to its own project

timcharper:
In this commit we simplify the configuration of integration tests and unit tests by moving integration tests to its own project, rather than separating the two with tags. This is arguably a less hacky-solution to the problem of having different configurations for running different kinds of tests.

There were multiple issues with using tags:

- Running unit tests would filter out IntegrationTest tagged tests, meaning it would generate an empty report for all integration test suites. This led to further problems with us needing to explicitly output test results to different folders.
- It was unclear how to run a single integration test; testOnly commands did not work as expected.
- The sbt configuration was difficult to understand


src/test/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFrameworkTest.scala
```diff
@@ -5,7 +5,7 @@ import java.util.UUID
 
 import akka.util.ByteString
 import mesosphere.UnitTest
-import mesosphere.marathon.integration.setup.ZookeeperServerTest
+import mesosphere.marathon.util.ZookeeperServerTest
```
jeschkies: I don't understand this change. How come `ZookeeperServerTest` has a different package now? I cannot find the related change.

tests/integration/src/test/scala/mesosphere/IntegrationTest.scala
```diff
@@ -0,0 +1,72 @@
+package mesosphere
```
jeschkies: Are we planning to move all other tests as well? Shouldn't the path be `tests/integration/src/test/scala`?

Jenkinsfile.release
```diff
@@ -51,6 +51,7 @@ ansiColor('xterm') {
         }
       } finally {
         junit(allowEmptyResults: true, testResults: 'target/test-reports/*.xml')
+        junit(allowEmptyResults: true, testResults: 'tests/integration/target/test-reports/*.xml')
```
jeschkies: Hm, I'm not sure the folder structure is quite right. It seems we mix them a lot. E.g. `tests` is in some language the path what `src/test` is to JVM languages. It would only include the test sources but not the test artifacts. The same way we would not have `src/test/target/junit.xml`. We could tread `tests` as a folder for projects. However, then I would expect `tests/integration/src/main/scala`, `tests/integration/target` and probably `tests/integration/build.sbt`.

src/test/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFrameworkTest.scala
timcharper: Because ZookeeperServerTest was used by multiple unit tests

Jenkinsfile.release
timcharper: Hmm... `src/integration/scala` ?

timcharper: They have to have separate target folders though, I think, otherwise I am not sure that running the integration tests will run the unit tests if the unit test artifacts are compiled and output in the same folder

tests/integration/src/test/scala/mesosphere/IntegrationTest.scala
jeschkies: Lol, I misread that one ^^.

Jenkinsfile.release
ichernetsky: I like `src/integration/scala` better.

PR #6244, opened 2018-05-23
Fail loop builds on error.

jeschkies:
Summary:
We used to swollow every exception. This change will through it so that
`./ci/pipeline loop` will exit with 0 on an error.

### DIFF
--- a/ci/pipeline
+++ b/ci/pipeline
@@ -259,7 +259,9 @@ def loop(): Unit = {
     build(buildName = utils.loopBuildName())
     dataDogClient.reportCount(s"marathon.build.${utils.loopName}.success", 1)
   } catch {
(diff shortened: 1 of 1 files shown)


ci/pipeline
```diff
@@ -259,7 +259,9 @@ def loop(): Unit = {
     build(buildName = utils.loopBuildName())
     dataDogClient.reportCount(s"marathon.build.${utils.loopName}.success", 1)
   } catch {
-    case _ => dataDogClient.reportCount(s"marathon.build.${utils.loopName}.failure", 1)
+    case ex =>
+      dataDogClient.reportCount(s"marathon.build.${utils.loopName}.failure", 1)
+      throw ex
```
zen-dog: 🤦‍♂️

PR #6245, opened 2018-05-23
Move system test project files.

jeschkies:
Summary:
This is a follow up to #6239 which moves the system test project files
to the system test project source files. So that `./tests` only contains
project folders.


ci/si_pipeline.sh
```diff
@@ -16,7 +16,7 @@ DEPLOYMENT_NAME="$JOB_NAME_SANITIZED-$BUILD_NUMBER"
 INFO_PATH="$DEPLOYMENT_NAME.info.json"
 
 # Change work directory to ./tests
-cd tests || exit 1
+cd tests/system || exit 1
```
alenkacz: wouldn't it be easier to use absolute paths everywhere, I mean e.g. introducing $rootPath variable or $testsPath instead of doing `cd` and then addressing lot of paths via `../../` ?

jeschkies: We have to `cd` to pipenv. Well, there is probably a way to reference a Pipefile in a different folder but I think the system tests should run in the `tests/system` folder. That said, you are right about `$ROOT` 🙂

jeschkies: Done ✅

PR #6249, opened 2018-05-24
Count instances to schedule inside instanceTracker

alenkacz:
In the future, InstanceTracker will have all the state to make this decision, this just moves this logic in there instead of having it spread in between StartActor and StartingBehavior.

Actually what happens then is that those classes call LaunchQueue that again calls InstanceTracker so in the ideal world in the future, we should just say InstanceTracker to schedule (instead of asking it for the count).

JIRA issues: MARATHON-8167

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
@@ -52,9 +52,7 @@ trait StartingBehavior extends ReadinessBehavior with StrictLogging { this: Acto
       launchQueue.add(runSpec, 1).pipeTo(self)
 
     case Sync => async {
-      val instances = await(instanceTracker.specInstances(runSpec.id))
-      val actualSize = instances.count { i => i.isActive || i.isReserved || i.isScheduled || i.isProvisioned }
-      val instancesToStartNow = Math.max(scaleTo - actualSize, 0)
+      val instancesToStartNow = await(instanceTracker.instancesToScheduleCount(runSpec.id, scaleTo))
       logger.debug(s"Sync start instancesToStartNow=$instancesToStartNow appId=${runSpec.id}")
       if (instancesToStartNow > 0) {
         logger.info(s"Reconciling app ${runSpec.id} scaling: queuing $instancesToStartNow new instances")
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
@@ -19,6 +19,7 @@ import org.apache.mesos
 import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContext, Future}
 import scala.util.control.NonFatal
+import scala.async.Async.{async, await}
 
 /**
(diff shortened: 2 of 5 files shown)


src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
```diff
@@ -52,9 +52,7 @@ trait StartingBehavior extends ReadinessBehavior with StrictLogging { this: Acto
       launchQueue.add(runSpec, 1).pipeTo(self)
 
     case Sync => async {
-      val instances = await(instanceTracker.specInstances(runSpec.id))
-      val actualSize = instances.count { i => i.isActive || i.isReserved || i.isScheduled || i.isProvisioned }
-      val instancesToStartNow = Math.max(scaleTo - actualSize, 0)
+      val instancesToStartNow = await(instanceTracker.instancesToScheduleCount(runSpec.id, scaleTo))
```
jeschkies: Hm, what is the benefit of moving it into the instance tracker? I always thought of the tracker as a simple CRUD interface.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -114,4 +109,15 @@ private[tracker] class InstanceTrackerDelegate(
     import scala.concurrent.ExecutionContext.Implicits.global
     process(InstanceUpdateOperation.ReservationTimeout(instanceId)).map(_ => Done)
   }
+
+  @SuppressWarnings(Array("all")) /* async/await */
+  override def instancesToScheduleCount(appId: PathId, targetCount: Int): Future[Int] = {
```
jeschkies: I'm not so sure about the benefit. However, we could make this more generic and count based on a provided filter

```scala
def count(specId: PathId, filter: (Instance) => Boolean): Future[Int] = async {
  val instances = await(list(specId))
  instances.count(filter)
}
```

alenkacz: my main motivation for this was:
- the list of possible states to filter when trying to figure out how much to scales is growing, right now this computation was on two places and those two lists were not in sync (possibly causing some regressions)
- in the future this method should go anyway - we count how much to schedule only to call instancetracker.schedule in the end, so I think that in the future there should be just `schedule(appId, targetCount)` method that will schedule (or not schedule) based on the instance tracker state. But to do this, we have to get rid of LaunchQueue first

So this is really just putting the algorithm to compute how much to schedule to one place inside instance tracker...

src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
alenkacz: this is actually a step toward that crud, because it will let us to have the algorithm calculating how much to schedule in here ... so later we will just move to `schedule(appId, targetCount)`

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
jeschkies: Good points! I'm convinced.

PR #6251, opened 2018-05-25
Ensure that replace actor is intialized.

jeschkies:
Summary:
There was a race condition between the pre-start phase of the
`TaskReplaceActor` and instance status change updates.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
@@ -32,7 +32,7 @@ class TaskReplaceActor(
     val eventBus: EventStream,
     val readinessCheckExecutor: ReadinessCheckExecutor,
     val runSpec: RunSpec,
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -93,18 +93,35 @@ class TaskReplaceActor(
...
   override def postStop(): Unit = {
     eventBus.unsubscribe(self)
     super.postStop()
   }
 
-  override def receive: Receive = readinessBehavior orElse replaceBehavior
+  override def receive: Receive = initializing
+
+  private def initializing: Receive = {
+    case Done =>
```
alenkacz: I so don't like how we rely on Done as a msg on so many places. For me it's like sending a message of type Any... (hard to tell where it came from, might be coming from any future)

But that's an overall problem, not necessarily with this particular change :)

jeschkies: I was thinking about using a custom message such as `case object Initialized`

PR #6259, opened 2018-05-28
Remove the 'possibleNewState' as it is used in tests only

alenkacz:
JIRA issues: MARATHON-8167


alenkacz:
I had troubles rebasing, so I cherry picked on top of new branch instead - so this one is already approved https://github.com/mesosphere/marathon/pull/6236, I'll consider this also approved :)

PR #6260, opened 2018-05-28
Introduce factory method for provisioning instances

alenkacz:
I need this as a foundation to introduce provisioning to resident apps because otherwise I would need to copy-paste the same creation logic to other places. Also I think that the `Instance` object is a nice place for such factory method.

JIRA issues: MARATHON-8167

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -1,7 +1,6 @@
 package mesosphere.marathon
 package core.instance
 
-import java.time.Clock
 import java.util.{Base64, UUID}
 
 import com.fasterxml.uuid.{EthernetAddress, Generators}
@@ -126,7 +125,7 @@ object Instance {
       * @return new instance in a provisioned state
       */
     def apply(scheduledInstance: Instance, agentInfo: AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition, now: Timestamp): Instance = {
-      require(scheduledInstance.isScheduled, s"Instance '${scheduledInstance.instanceId}' cannot be in state '${scheduledInstance.state.condition}'. Scheduled instance is required to create provisioned instance.")
+      require(scheduledInstance.isScheduled, s"Instance '${scheduledInstance.instanceId}' must not be in state '${scheduledInstance.state.condition}'. Scheduled instance is required to create provisioned instance.")
 
       val task = Task(
         taskId = Task.Id.forInstanceId(scheduledInstance.instanceId, None),
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -125,14 +125,14 @@ object Instance {
       * @param scheduledInstance instance in a Scheduled state
       * @return new instance in a provisioned state
       */
-    def apply(scheduledInstance: Instance, agentInfo: AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition)(implicit clock: Clock): Instance = {
-      require(scheduledInstance.isScheduled, "Scheduled instance is required to create provisioned instance.")
+    def apply(scheduledInstance: Instance, agentInfo: AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition, now: Timestamp): Instance = {
+      require(scheduledInstance.isScheduled, s"Instance '${scheduledInstance.instanceId}' cannot be in state '${scheduledInstance.state.condition}'. Scheduled instance is required to create provisioned instance.")
 
       val task = Task(
         taskId = Task.Id.forInstanceId(scheduledInstance.instanceId, None),
         runSpecVersion = app.version,
         status = Task.Status(
-          stagedAt = clock.now(),
+          stagedAt = now,
           condition = Condition.Created,
           networkInfo = networkInfo
         )
@@ -142,9 +142,9 @@ object Instance {
 
       scheduledInstance.copy(
         agentInfo = Some(agentInfo),
-        state = Instance.InstanceState(Condition.Provisioned, clock.now(), None, None),
+        state = Instance.InstanceState(Condition.Provisioned, now, None, None),
         tasksMap = tasksMap,
-        runSpecVersion = app.version // Note: The app version might change.
+        runSpecVersion = app.version
       )
     }
   }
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -118,6 +119,36 @@ object Instance {
     def apply(runSpec: RunSpec): Instance = Scheduled(runSpec, Id.forRunSpec(runSpec.id))
   }
 
+  object Provisioned {
+    /**
+      * Factory method for creating provisioned instance from Scheduled instance for apps
+      * @param scheduledInstance instance in a Scheduled state
+      * @return new instance in a provisioned state
+      */
+    def apply(scheduledInstance: Instance, agentInfo: AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition)(implicit clock: Clock): Instance = {
```
jeschkies: Why not simply pass `now` as a value?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -118,6 +119,36 @@ object Instance {
...
+      require(scheduledInstance.isScheduled, "Scheduled instance is required to create provisioned instance.")
+
+      val task = Task(
+        taskId = Task.Id.forInstanceId(scheduledInstance.instanceId, None),
```
jeschkies: This is not correct for pods that have a container. I think you'll have to pass the container parameter as well. However, you can do that in the next step.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -118,6 +119,36 @@ object Instance {
...
+          condition = Condition.Created,
+          networkInfo = networkInfo
+        )
+      )
+
+      val tasksMap = Map(task.taskId -> task)
+
+      scheduledInstance.copy(
+        agentInfo = Some(agentInfo),
+        state = Instance.InstanceState(Condition.Provisioned, clock.now(), None, None),
```
jeschkies: `clock.now()` is different than `stagedAt`.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -118,6 +119,36 @@ object Instance {
...
+        tasksMap = tasksMap,
+        runSpecVersion = app.version // Note: The app version might change.
```
jeschkies: The comment is wrong. `Provisioned` instances won't change their app version.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -118,6 +119,36 @@ object Instance {
...
   }
 
+  object Provisioned {
+    /**
+      * Factory method for creating provisioned instance from Scheduled instance for apps
+      * @param scheduledInstance instance in a Scheduled state
+      * @return new instance in a provisioned state
+      */
+    def apply(scheduledInstance: Instance, agentInfo: AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition)(implicit clock: Clock): Instance = {
+      require(scheduledInstance.isScheduled, "Scheduled instance is required to create provisioned instance.")
```
jeschkies: Could you add the actual condition and the instance id? That will ease debugging.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -131,27 +131,11 @@ class InstanceOpFactoryImpl(
...
-        val tasksMap = Map(task.taskId -> task)
-        val state = Instance.InstanceState(Condition.Provisioned, now, None, None)
-        val provisionedInstance = scheduledInstance.copy(
-          agentInfo = Some(agentInfo),
-          state = state,
-          tasksMap = tasksMap,
-          runSpecVersion = app.version // Note: The app version might change.
-        )
-        val instanceOp = taskOperationFactory.provision(taskInfo, task, provisionedInstance)
+        val provisionedInstance = Instance.Provisioned(scheduledInstance, agentInfo, networkInfo, app)(clock)
```
jeschkies: I'd rather pass the time stamp than the clock.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: this is just for apps, pods will have to have a different method because more stuff differ...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: oh ok, I copypasted this from another place in code :)

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: ok ... I had that and then switched to clock, but maybe timestamp actually makes more sense :)

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: as you can see, it accepts AppDefinition, not RunSpec

jeschkies: Ah, you are right.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -1,22 +1,23 @@
 package mesosphere.marathon
 package core.instance
 
+import java.time.Clock
```
jeschkies: Is this still required?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -118,6 +119,36 @@ object Instance {
...
   }
 
+  object Provisioned {
+    /**
+      * Factory method for creating provisioned instance from Scheduled instance for apps
+      * @param scheduledInstance instance in a Scheduled state
+      * @return new instance in a provisioned state
+      */
+    def apply(scheduledInstance: Instance, agentInfo: AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition, now: Timestamp): Instance = {
+      require(scheduledInstance.isScheduled, s"Instance '${scheduledInstance.instanceId}' cannot be in state '${scheduledInstance.state.condition}'. Scheduled instance is required to create provisioned instance.")
```
jeschkies: Nit: `s/cannot/must not`

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: no. I guess the build will fail as well 🗡

alenkacz:
the failure is because of overscaling resident tasks - merging

PR #6261, opened 2018-05-28
Remove absolutely nonsense use of implicits and writes

alenkacz:

### DIFF
--- a/src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
+++ b/src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
@@ -15,8 +15,7 @@ trait PodStatusConversion {
 
   import PodStatusConversion._
 
-  implicit val taskToContainerStatus: Writes[(PodDefinition, task.Task), ContainerStatus] = Writes { src =>
-    val (pod, task) = src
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -69,7 +68,7 @@ trait PodStatusConversion {
       pod.id == instance.instanceId.runSpecId,
       s"pod id ${pod.id} should match spec id of the instance ${instance.instanceId.runSpecId}")
 
-    val containerStatus: Seq[ContainerStatus] = instance.tasksMap.values.map(t => Raml.toRaml((pod, t)))(collection.breakOut)
+    val containerStatus: Seq[ContainerStatus] = instance.tasksMap.values.map(t => taskToContainerStatus(pod, t))(collection.breakOut)
```
jeschkies: What do you think about

```scala
def taskToContainerStatus(pod: PodDefinition)(task: task.Task): ContainerStatus
val containerStatus: Seq[ContainerStatus] =
    instance.tasksMap.values.map(taskToContainerStatus(pod))(collection.breakOut)
	

```

PR #6262, opened 2018-05-29
Changelog update for 1.4.12

meln1k:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,13 @@
+## Changes from 1.4.11 to 1.4.12
+
+### Fixed issues
+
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -1,3 +1,13 @@
+## Changes from 1.4.11 to 1.4.12
+
+### Fixed issues
+
+- [COPS-2072](https://jira.mesosphere.com/browse/COPS-2072) Expose `containerPort` as `$PORT_NAME` in environment variables. (#5892)
```
alenkacz: I think we should not mention COPS ticket since it's not a public JIRA project

alenkacz: I would just remove the JIRA reference and left the comment there (if there is no marathon ticket)

PR #6264, opened 2018-05-30
Resident task should be scheduled in the same way as other tasks

alenkacz:
Main changes:
- for newly scheduled residents the flow is following: Scheduled -> Scheduled (with Reservation) -> Provisioned -> ...
- for relaunching residents the flow is: Reserved (terminal) -> Scheduled (with Reservation) -> Provisioned -> ...

- Scheduled instances with reservations have agentinfo now. That's necessary so we don't violate placement constraints and e.g. do reservation on agent that already have one scheduled and has hostname: unique

JIRA issues: MARATHON-8208


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -35,7 +37,7 @@ case class Instance(
   val runSpecId: PathId = instanceId.runSpecId
 
   // An instance has to be considered as Reserved if at least one of its tasks is Reserved.
-  def isReserved: Boolean = tasksMap.values.exists(_.status.condition == Condition.Reserved)
+  def isReserved: Boolean = reservation.isDefined
```
jeschkies: Nice!

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -116,6 +118,12 @@ object Instance {
      * @return An instance in the scheduled state.
      */
     def apply(runSpec: RunSpec): Instance = Scheduled(runSpec, Id.forRunSpec(runSpec.id))
+
+    def apply(scheduledInstance: Instance, reservation: Reservation, agentInfo: AgentInfo): Instance = {
```
jeschkies: Could you add a doc string? I'm not sure this easier than `instance.copy` 😉

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -146,6 +159,76 @@ object Instance {
...
+      * @return new instance in a provisioned state
+      */
+    def apply(
+      scheduledInstance: Instance,
+      agentInfo: Instance.AgentInfo,
+      hostPorts: Seq[Option[Int]],
+      taskIDs: Seq[Task.Id],
+      pod: PodDefinition,
+      now: Timestamp): Instance = {
+      require(scheduledInstance.isScheduled, s"Instance '${scheduledInstance.instanceId}' must not be in state '${scheduledInstance.state.condition}'. Scheduled instance is required to create provisioned instance.")
```
jeschkies: Nit: `to create a provisioned instance.`

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -177,7 +260,8 @@ object Instance {
 
       //From here on all tasks are either Created, Reserved, Running, Finished, or Killed
```
jeschkies: I think the doc has to be adapted

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -32,9 +32,15 @@ private[marathon] class InstanceUpdateOpResolver(
...
     op match {
       case op: Schedule =>
-        createInstance(op.instanceId) {
-          // TODO(karsten): Create events
-          InstanceUpdateEffect.Update(op.instance, oldState = None, Seq.empty)
+        // TODO(karsten): Create events
+        createOrUpdateInstance(op.instanceId, InstanceUpdateEffect.Update(op.instance, oldState = None, Seq.empty)) { i =>
+          (i.agentInfo, i.reservation) match {
+            case (Some(agentInfo), Some(reservation)) =>
+              // we already have suspended instance with reservation, let's schedule that one
```
jeschkies: Ni: `have a suspended instance` It might be confused with `we have suspended the instance`

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -32,9 +32,15 @@ private[marathon] class InstanceUpdateOpResolver(
...
+              InstanceUpdateEffect.Update(Instance.Scheduled(op.instance, reservation, agentInfo), oldState = Some(i), Seq.empty)
```
jeschkies: So we reserved the instance right?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -32,9 +32,15 @@ private[marathon] class InstanceUpdateOpResolver(
...
+            case _ => InstanceUpdateEffect.Failure(
+              new IllegalStateException(s"${op.instanceId} of app [${op.instanceId.runSpecId}] already exists"))
```
jeschkies: I think this is not the correct message anymore.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -177,7 +174,8 @@ class InstanceOpFactoryImpl(
 
         // resources are reserved for this role, so we only consider those resources
         val rolesToConsider = config.mesosRole.get.toSet
-        val reservationLabels = TaskLabels.labelsForTask(request.frameworkId, volumeMatch.instance.appTask.taskId).labels
+        val taskId = Task.Id.forInstanceId(volumeMatch.instance.instanceId, None)
```
jeschkies: Was this a bug?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -354,81 +352,22 @@ class InstanceOpFactoryImpl(
...
-            healthy = None
-          ),
-          tasksMap = Map(task.taskId -> task),
-          runSpecVersion = runSpec.version,
-          unreachableStrategy = runSpec.unreachableStrategy,
-          reservation = Some(reservation)
-        )
-        val stateOp = InstanceUpdateOperation.Reserve(instance)
+        val reservationLabels = TaskLabels.labelsForTask(frameworkId, Task.Id.forInstanceId(scheduledInstance.instanceId, None))
+        val stateOp = InstanceUpdateOperation.Reserve(Instance.Scheduled(scheduledInstance, reservation, agentInfo))
```
jeschkies: This is so much cleaner. Although, I think this would be ok as well

```scala
val reservationLabels = TaskLabels.labelsForTask(frameworkId, Task.Id.forInstanceId(scheduledInstance.instanceId, None))
val updatedInstance = scheduledInstance.copy(reservation = Some(reservation), agentInfo = Some(agentInfo))
val stateOp = InstanceUpdateOperation.Reserve(updatedInstance)
```

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
```diff
@@ -153,10 +152,10 @@ class MarathonSchedulerActorTest extends AkkaUnitTest with ImplicitSender with G
       expectMsg(TasksReconciled)
 
       val nonTerminalTasks = instance.tasksMap.values.filter(!_.task.isTerminal)
-      assert(nonTerminalTasks.size == 7, "We should have 7 non-terminal tasks")
+      assert(nonTerminalTasks.size == 6, "We should have 7 non-terminal tasks")
```
jeschkies: The assert message is not quite correct

src/test/scala/mesosphere/marathon/core/instance/InstanceTest.scala
```diff
@@ -97,10 +98,18 @@ class InstanceTest extends UnitTest with TableDrivenPropertyChecks {
         s"${if (!isDropped) "not" else ""} be dropped" in { instance.isDropped should be(isDropped) }
         s"${if (!isActive) "not" else ""} be active" in { instance.isActive should be(isActive) }
         s"${if (!isTerminated) "not" else ""} be terminated" in { instance.isTerminated should be(isTerminated) }
+        s"${if (!isProvisioned) "not" else ""} be provisioned" in { instance.isProvisioned should be(isProvisioned) }
+        s"${if (!isScheduled) "not" else ""} be scheduled" in { instance.isScheduled should be(isScheduled) }
       }
     }
   }
 
+  "say it's reserved when reservation is set" in {
```
jeschkies: That's a much better test than my old one above.

src/test/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolverTest.scala
```diff
@@ -238,6 +238,19 @@ class InstanceUpdateOpResolverTest extends UnitTest with Inside {
       verifyNoMoreInteractions()
     }
 
+    "schedule reserved instance if exists instead of creating new scheduled" in new Fixture {
+      instanceTracker.instance(reservedInstance.instanceId) returns Future.successful(Some(reservedInstance))
+      val scheduledInstance = Instance.Scheduled(AppDefinition(appId), reservedInstance.instanceId)
+      val stateChange = updateOpResolver.resolve(InstanceUpdateOperation.Schedule(scheduledInstance)).futureValue
+
+      stateChange shouldBe an[InstanceUpdateEffect.Update]
+      val updateEffect = stateChange.asInstanceOf[InstanceUpdateEffect.Update]
```
jeschkies: Why not

```scala
inside(stateChange) { case InstanceUpdateEffect.Update(instance, _, _) =>
  instance.reservation should be (...)
  ...
}

File: src/test/scala/mesosphere/marathon/state/AppDefinitionPortAssignmentsTest.scala [hunk 13]
```diff
@@ -84,17 +84,6 @@ class AppDefinitionPortAssignmentsTest extends UnitTest {
       task.status.networkInfo.portAssignments(app, includeUnresolved = true) should be(empty)
     }
 
-    "portAssignments with a reserved task" in {
```
jeschkies: Where did this go?

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
```diff
@@ -156,16 +155,15 @@ class InstanceOpFactoryImplTest extends UnitTest with Inside {
...
       val reservedInstance = f.residentReservedInstance(app.id, localVolumeIdMatch)
-      val (reservedTaskId, _) = reservedInstance.tasksMap.head
+      val reservedTaskId = Task.Id.forInstanceId(reservedInstance.instanceId, None)
       val offer = f.offerWithVolumes(
         reservedTaskId, localVolumeIdLaunched, localVolumeIdUnwanted, localVolumeIdMatch
       )
       val runningInstances = Instance.instancesById(Seq(
-        f.residentLaunchedInstance(app.id, localVolumeIdLaunched),
-        reservedInstance))
+        f.residentLaunchedInstance(app.id, localVolumeIdLaunched)))
```
jeschkies: Why did you remove the second instance?

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
```diff
@@ -212,15 +210,13 @@ class InstanceOpFactoryImplTest extends UnitTest with Inside {
       val app = f.residentApp
       val volumeId = LocalVolumeId(app.id, "/path", "uuid1")
       val existingReservedInstance = f.residentReservedInstance(app.id, volumeId)
-      existingReservedInstance.agentInfo.value.host shouldBe f.defaultHostName
```
jeschkies: Why was this removed?

src/test/scala/mesosphere/mesos/PersistentVolumeMatcherTest.scala
```diff
@@ -3,10 +3,13 @@ package mesosphere.mesos
 import mesosphere.UnitTest
 import mesosphere.marathon._
 import mesosphere.marathon.core.instance.{Instance, LocalVolumeId, TestInstanceBuilder}
-import mesosphere.marathon.state.{AppDefinition, PathId}
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.state.AppDefinition
 import mesosphere.marathon.stream.Implicits._
 import mesosphere.marathon.test.MarathonTestHelper
 
+import scala.collection.immutable.Seq
```
jeschkies: Hm, there should be no need to this since we have a `Seq` definition in `mesosphere.marathon`.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: well to move to a certain instance states, you have to know which fields to copy. Having this knowledge spread over multiple places in the codebase is brittle - if you need to start modifying one more field, you can easily forget to change it at all places

src/test/scala/mesosphere/marathon/state/AppDefinitionPortAssignmentsTest.scala
alenkacz: reserved instances don't have tasks anymore...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -34,8 +37,7 @@ case class Instance(
 
   val runSpecId: PathId = instanceId.runSpecId
 
-  // An instance has to be considered as Reserved if at least one of its tasks is Reserved.
-  def isReserved: Boolean = tasksMap.values.exists(_.status.condition == Condition.Reserved)
+  lazy val isReserved: Boolean = state.condition == Condition.Reserved
```
alenkacz: I think renaming to suspended actually makes sense because isReserved is kind of misleading now - because it's not obvious that it's the terminal state and not the pre-launch state, that is in a sense reserved too...

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -198,16 +200,24 @@ private[impl] class LaunchQueueActor(
...
 
       async {
-        val instances = 0.until(count).map { _ => Instance.Scheduled(app, Instance.Id.forRunSpec(app.id)) }
-        val start = await(instanceTracker.schedule(instances))
+        val existingReserved = await(instanceTracker.specInstances(runSpec.id))
+          .filter(_.isReserved)
+          .take(count)
+          .map(_.copy(state = InstanceState(Condition.Scheduled, Timestamp.now(), None, None), runSpecVersion = runSpec.version, unreachableStrategy = runSpec.unreachableStrategy))
+          .map(InstanceUpdateOperation.RelaunchReserved)
+        val instancesToSchedule = existingReserved.length.until(count).map { _ => Instance.Scheduled(runSpec, Instance.Id.forRunSpec(runSpec.id)) }
```
alenkacz: this is where the race condition is originated :) it would need some more synchronization here to get around that

src/test/scala/mesosphere/marathon/state/AppDefinitionPortAssignmentsTest.scala
alenkacz: or at least does not have to have...

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
alenkacz: because reserved instance is not running instance now... (it might have been considered as one before)

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
alenkacz: scheduled instances with reservations does not have tasks anymore (only when relaunching)

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -501,7 +501,7 @@ class SchedulerActions(
   */
 object TaskStatusCollector {
   def collectTaskStatusFor(instances: Seq[Instance]): Seq[mesos.Protos.TaskStatus] = {
-    instances.collect { case Instance(_, Some(agentInfo), _, tasksMap, _, _, _) => (agentInfo, tasksMap) }
+    instances.collect { case Instance(_, Some(agentInfo), _, tasksMap, _, _, _) if tasksMap.nonEmpty => (agentInfo, tasksMap) }
```
alenkacz: we need this because scheduled task can have agentinfo now

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: Do you want to follow up with a renaming?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -124,30 +135,111 @@ object Instance {
...
-          condition = Condition.Created,
-          networkInfo = networkInfo
-        )
+      val taskIds: Seq[Task.Id] = if (scheduledInstance.hasReservation) {
+        val originalIds = if (scheduledInstance.tasksMap.nonEmpty) {
+          scheduledInstance.tasksMap.keys
+        } else {
+          Seq(Task.Id.forInstanceId(scheduledInstance.instanceId, None))
+        }
+        originalIds.map(ti => Task.Id.forResidentTask(ti)).to[Seq]
```
jeschkies: I'd be fine with `originalIds.map(Task.Id.forResidentTask)`. I've talked to Alex yesterday. He agreed that we probably want to have a different reservation id scheme. Currently the reservation id is the same as the task id, right?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -124,30 +135,111 @@ object Instance {
...
-      val task = Task(
-        taskId = Task.Id.forInstanceId(scheduledInstance.instanceId, None),
-        runSpecVersion = app.version,
-        status = Task.Status(
-          stagedAt = now,
-          condition = Condition.Created,
-          networkInfo = networkInfo
-        )
+      val taskIds: Seq[Task.Id] = if (scheduledInstance.hasReservation) {
+        val originalIds = if (scheduledInstance.tasksMap.nonEmpty) {
```
jeschkies: I which case can we have a scheduled instance with a non-empty task map?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -32,11 +35,15 @@ private[marathon] class InstanceUpdateOpResolver(
...
     op match {
       case op: Schedule =>
-        createInstance(op.instanceId) {
-          // TODO(karsten): Create events
+        // TODO(karsten): Create events
+        createInstance(op.instanceId){
           InstanceUpdateEffect.Update(op.instance, oldState = None, Seq.empty)
         }
-
+      case op: RelaunchReserved =>
```
jeschkies: This is actually quite interesting. I think the logic in `InstanceUpdateOpResolver` is the one we need in the orchestrator 🙂

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryHelper.scala
```diff
@@ -58,11 +58,11 @@ class InstanceOpFactoryHelper(
...
   def launchOnReservation(
     taskInfo: Mesos.TaskInfo,
-    newState: InstanceUpdateOperation.LaunchOnReservation,
+    newState: InstanceUpdateOperation.Provision,
     oldState: Instance): InstanceOp.LaunchTask = {
 
     assume(
-      oldState.isReserved,
+      oldState.hasReservation,
       "only a reserved instance can be re-launched")
```
jeschkies: Maybe "only an instance with a reservation can be re-launched"

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -258,24 +254,24 @@ class InstanceOpFactoryImpl(
         // All of these cases are handled in one way: by creating a new taskId for a resident task based on the previous
         // one. The used function will increment the attempt counter if it exists, of append a 1 to denote the first attempt
         // in version 1.5.
-        val newTaskId = Task.Id.forResidentTask(taskId)
+        val taskIds: Seq[Task.Id] = if (reservedInstance.hasReservation) {
+          val originalIds = if (reservedInstance.tasksMap.nonEmpty) {
```
jeschkies: As I mentioned above, this is a little confusing to me. When can we have this case?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jeschkies: Push

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
jeschkies: Out of curiosity, why did you add the logic here? There are very few places that use `LaunchQueue.add`: `TaskReplaceActor`, `StartingBehaviour`, `TaskStartActor` and `MarathonScheduleActor`. I think these should decide not the launch queue. We should sync with Matthias on this.

Overall the launch queue should go and not be extended.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: yes

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
alenkacz: yes, if orchestrator serializes updates per runspec and there is just one, we can have this in front of instance tracker, so in orchestrator :)

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: well currently scheduled instance has tasks if it ever got to at least provisioned state, then we keep the task around... before we always created dummy task even for "scheduled" state. Maybe we can go back to doing that so there will always be task when reservation is present...

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
alenkacz: I feel that doing copy actually means that you know too much detail about that particular case class and I don't like that to be spread around code. If you in the end find out that the agent info should be also copied to the tasks in taskmap, you would have to somehow find all the right `.copy` calls and update it... so I tend to like to have `.copy` calls as close to the actual instance as possible :)

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: btw. we need this because for every re-launch we need to increment the attempt count, that is part of reserved instance id. so we need to keep tasks around.

Question is if we want to create dummy tasks for scheduled instances or not. I decided not to do that and for that reason, this code exist. Otherwise the id will always exist and we won't need to have an `if` statement here...

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
alenkacz: because this is the only place that is *JUST ONE* and place through all the schedule calls go through. It can be inside instance tracker but that does not seem like a good place.

I think it is not preventing us to get rid of this guy. We will just need to call this code (or copy this code) to all the places that call LaunchQueue.Add right now...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: Do you think it's worth it? The reserved state will go anyway, right? It will be state: Killed, target: Suspended

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: I would not create dummy tasks. I wonder why we are not removing tasks if an instance goes back to scheduled. The incarnation number is an argument. We might want to track that data somewhere else. E.g. with the instance.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
jeschkies: I don't know. This seems like a short term thing.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -258,24 +254,24 @@ class InstanceOpFactoryImpl(
         // All of these cases are handled in one way: by creating a new taskId for a resident task based on the previous
         // one. The used function will increment the attempt counter if it exists, of append a 1 to denote the first attempt
         // in version 1.5.
-        val newTaskId = Task.Id.forResidentTask(taskId)
+        val taskIds: Seq[Task.Id] = if (reservedInstance.hasReservation) {
```
jeschkies: This seems like a duplication of `Provisioned.apply`

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -286,7 +282,13 @@ class InstanceOpFactoryImpl(
           config.mesosBridgeName())
 
         val instanceId = reservedInstance.instanceId
-        val taskIds = reservedInstance.tasksMap.keys
+        val taskIds = if (reservedInstance.tasksMap.nonEmpty) {
```
jeschkies: Why is there no check for `reservedInstance.hasReservation` as done for apps?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -116,6 +118,15 @@ object Instance {
      * @return An instance in the scheduled state.
      */
     def apply(runSpec: RunSpec): Instance = Scheduled(runSpec, Id.forRunSpec(runSpec.id))
+
+    /**
+      * Creates new instance that is scheduled and has reservation (for resident run specs)
+      */
+    def apply(scheduledInstance: Instance, reservation: Reservation, agentInfo: AgentInfo): Instance = {
```
meln1k: Should we made it a method of an `Instance`?

src/main/scala/mesosphere/marathon/core/launcher/InstanceOpFactory.scala
```diff
@@ -38,7 +38,7 @@ object InstanceOpFactory {
       localRegion: Option[Region] = None) {
     def frameworkId: FrameworkId = FrameworkId("").mergeFromProto(offer.getFrameworkId)
     def instances: Seq[Instance] = instanceMap.values.to[Seq]
-    lazy val reserved: Seq[Instance] = instances.filter(_.isReserved)
+    lazy val reserved: Seq[Instance] = scheduledInstances.filter(_.hasReservation).to[Seq]
```
meln1k: What do you think about calling .toList or .toVector here?

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
meichstedt: I think that's 50% true – the resolver currently handles both updates to existing instances as well as forwarding the command to create them.

However the orchestrator would be the one that initially _decides_ to schedule an instance (based on e.g. an API request that says _please scale to 5_). It would however be the one that decides if e.g. a status update `TASK_FAILED` should affect the current goal (e.g. set new goal: Suspended if the orchestration logic has a retry limit).

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
meichstedt: I agree with @alenkacz here :)

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
meichstedt: I don't see where this is checked for apps, so maybe the comment is outdated?
If the instance has no reservation there could be no match I guess.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -48,7 +48,7 @@ private[tracker] class InstanceTrackerDelegate(
   // TODO(jdef) support pods when counting launched instances
   override def countActiveSpecInstances(appId: PathId): Future[Int] = {
     import scala.concurrent.ExecutionContext.Implicits.global
-    instancesBySpec().map(_.specInstances(appId).count(instance => instance.isActive || instance.isReserved))
+    instancesBySpec().map(_.specInstances(appId).count(instance => instance.isActive))
```
meichstedt: Same as above: `Created` (the instance after the match before we get any status update) was considered active while `Provisioned` is not.

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
meichstedt: The `reservedInstance` needs to be contained in the map. The signature of the Request is
```
  case class Request(
      runSpec: RunSpec,
      offer: Mesos.Offer,
      instanceMap: Map[Instance.Id, Instance],
      scheduledInstances: Iterable[Instance],
      localRegion: Option[Region] = None)
```

Where the instanceMap is supposed to contain **all** reachable instances. Those are used e.g. to match constraints. The `TaskLauncherActor` does this:
```
val matchRequest = InstanceOpFactory.Request(runSpec, offer, reachableInstances, scheduledInstances, localRegion())
```

while the logic there seems to be broken:
```
val reachableInstances = instanceMap.filterNotAs{
    case (_, instance) => instance.state.condition.isLost || instance.isScheduled
}
```

This should **not** exclude instances w/ reservation (i.e. agentInfo). Otherwise we could create additional reservations on an agent that already has one (or multiple!), and we'd never be able to launch tasks for more than one of them.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -124,30 +135,100 @@ object Instance {
...
+  ): Map[Task.Id, NetworkInfo] = {
+
+    val reqPortsByCTName: Seq[(String, Option[Int])] = pod.containers.flatMap { ct =>
+      ct.endpoints.map { ep =>
+        ct.name -> ep.hostPort
+      }
+    }
+
+    val totalRequestedPorts = reqPortsByCTName.size
+    assume(totalRequestedPorts == hostPorts.size, s"expected that number of allocated ports ${hostPorts.size}" +
```
meln1k: what do you think about using `require` here? otherwise `assume` can me disabled in runtime with `-Xdisable-assertions`

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -198,16 +200,24 @@ private[impl] class LaunchQueueActor(
...
+        if (instancesToSchedule.nonEmpty) {
+          val scheduled = await(instanceTracker.schedule(instancesToSchedule))
+        }
+        val relaunched = await(Future.sequence(existingReserved.map(instanceTracker.process)))
```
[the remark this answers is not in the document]
alenkacz: I asked myself the same question actually ... so in general I agree. But:
- Karsten refactored instance tracker to more CRUD interface quite recently so I would like to discuss with him
- but more importantly - as you said, this code will go away when goal state is introduced anyway - then only the `schedule` call will remain so it will actually be easier to remove and separate those two then

to sum up - I would vote for leaving as it is :)

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
```diff
@@ -20,6 +17,16 @@ object InstanceUpdateOperation {
     override def instanceId: Instance.Id = instance.instanceId
   }
 
+  /**
+    * Creates a new instance. This is similar to [[LaunchEphemeral]] except that a scheduled instance has no information
+    * where it might run.
+    *
+    * @param reservedInstance The new instance.
+    */
+  case class RelaunchReserved(reservedInstance: Instance) extends InstanceUpdateOperation {
```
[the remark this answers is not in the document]
alenkacz: ad 1 - yes, totally
ad 2 - I already removed that in another PR so won't be doing that in this PR

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -146,7 +143,7 @@ class InstanceOpFactoryImpl(
     val InstanceOpFactory.Request(runSpec, offer, instances, scheduledInstances, localRegion) = request
 
     val needToLaunch = request.hasWaitingReservations
-    val needToReserve = scheduledInstances.nonEmpty
+    val needToReserve = scheduledInstances.exists(!_.hasReservation)
```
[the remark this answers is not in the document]
alenkacz: good catch, thanks :)

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -239,14 +237,12 @@ class InstanceOpFactoryImpl(
     resourceMatch: ResourceMatcher.ResourceMatch,
     volumeMatch: PersistentVolumeMatcher.VolumeMatch): InstanceOp = {
 
-    // The agentInfo could have possibly changed after a reboot. See the docs for
+    // The agentInfo could have possibly change after a reboot. See the docs for
```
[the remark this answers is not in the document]
alenkacz: :D 👍

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -258,24 +254,22 @@ class InstanceOpFactoryImpl(
...
-        val newTaskId = Task.Id.forResidentTask(taskId)
+        val taskIds: Seq[Task.Id] = {
+          val originalIds = if (reservedInstance.tasksMap.nonEmpty) {
+            reservedInstance.tasksMap.keys
+          } else {
+            Seq(Task.Id.forInstanceId(reservedInstance.instanceId, None))
+          }
+          originalIds.map(ti => Task.Id.forResidentTask(ti)).to[Seq]
+        }
+        val newTaskId = taskIds.headOption.getOrElse(throw new IllegalStateException(s"Expecting to have a task id present when creating instance for app ${app.id} from instance $reservedInstance"))
```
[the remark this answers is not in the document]
alenkacz: the if is necessary because tasksMap is intentionally empty for scheduled instances. Now we have 2 cases:
- newly scheduled instance (that was not previously launched) - this one does not have dummy task anymore - so since we did not create that dummy Id when scheduling, we have to create it here
- we are rescheduling instance what was already terminal. In this case we keep the id around so that we can increment attempt (in general we should move away from this in the future and keep just the attempts count somewhere, not the whole task)

if you know how to express this better, let me know

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -273,11 +272,11 @@ private class TaskLauncherActor(
   }
 
   private[this] def replyWithQueuedInstanceCount(): Unit = {
-    val activeInstances = instanceMap.values.count(instance => instance.isActive || instance.isReserved)
+    val activeInstances = instanceMap.values.count(instance => instance.isActive)
```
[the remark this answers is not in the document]
alenkacz: hmm that's a very good point, I would say it is probably a bug (on the feature branch)

I would vote for adressing that in a separate PR...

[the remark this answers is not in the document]
alenkacz: https://jira.mesosphere.com/browse/MARATHON-8273

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
alenkacz: hmm I don't think that's true but it was hard to tell - I adjusted the names here a bit.
This is *SCHEDULED* instance with reservation, so it should not be among running instances

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: oh I did not get that from the comment. I put the comment back as well. Thanks!

Too bad there is no test that would actually catch this regression :)

alenkacz:
I think that fix for https://github.com/mesosphere/marathon/commit/12d57af5facd852fa80d32a2be2c3ed4ae3817db introduced more flakiness

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: Good point.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jeschkies: Ok. I'm convinced.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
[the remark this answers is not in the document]
jeschkies: There is actually too much logic in this code. I think I mentioned before that the check should happen in the orchestrator. There shouldn't be many calls to `LaunchQueue.add`. Are there?

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -501,10 +501,10 @@ class SchedulerActions(
   */
 object TaskStatusCollector {
   def collectTaskStatusFor(instances: Seq[Instance]): Seq[mesos.Protos.TaskStatus] = {
-    instances.collect { case Instance(_, Some(agentInfo), _, tasksMap, _, _, _) => (agentInfo, tasksMap) }
+    instances.collect { case i @ Instance(_, Some(agentInfo), _, tasksMap, _, _, _) if tasksMap.nonEmpty => (agentInfo, tasksMap) }
```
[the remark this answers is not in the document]
alenkacz: Do we have a consensus on this? I don't want to bring in new concept if we don't have a buy in from few other people...

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -125,18 +123,15 @@ class TaskReplaceActor(
 
   def replaceBehavior: Receive = {
     // New instance failed to start, restart it
-    case InstanceChanged(id, `version`, `pathId`, condition, Instance(_, Some(agentInfo), _, _, _, _, _)) if !oldInstanceIds(id) && considerTerminal(condition) && condition != Condition.Reserved =>
-      logger.error(s"New instance $id failed on agent ${agentInfo.agentId} during app $pathId restart: $condition")
+    case InstanceChanged(id, `version`, `pathId`, condition, Instance(_, Some(agentInfo), _, _, _, _, reservation)) if !oldInstanceIds(id) && considerTerminal(condition) =>
+      logger.error(s"New instance $id is terminal on agent ${agentInfo.agentId} during app $pathId restart: $condition reservation: $reservation")
```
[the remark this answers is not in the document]
alenkacz: changed it to warn

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -32,16 +35,23 @@ private[marathon] class InstanceUpdateOpResolver(
...
-          // TODO(karsten): Create events
+        // TODO(karsten): Create events
+        createInstance(op.instanceId){
           InstanceUpdateEffect.Update(op.instance, oldState = None, Seq.empty)
         }
-
+      case op: RescheduleReserved =>
+        // TODO(alena): Create events
+        updateExistingInstance(op.instanceId) { i =>
+          InstanceUpdateEffect.Update(i.copy(state = InstanceState(Condition.Scheduled, Timestamp.now(), None, None), runSpecVersion = op.reservedInstance.version, unreachableStrategy = op.reservedInstance.unreachableStrategy), oldState = Some(i), Seq.empty)
```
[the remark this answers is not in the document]
alenkacz: why? shouldn't this been taken care of on compile?

alenkacz:
@jeschkies LaunchQueue.Add is called in `MarathonScheduler`, `TaskReplaceActor`, `TaskStartActor` and `StartingBehavior` - I think we should decide what to do with this code once we are in the process of getting rid of LaunchQueue

PR #6265, opened 2018-05-30
 Add new persistence store layer for ZK storage operations

zen-dog:
Summary:
Added new `SimplePersistenceStore` layer for ZK storage operations (based on Apache Curator) that:

- utilizes Apache Curator async methods under the hood from the org.apache.curator.curator-x-async library. This is similar to synchronous methods with callback however, it uses a somewhat cleaner [asynchronous DSL](https://curator.apache.org/curator-x-async/index.html)
- offers simple CRUD operations along with ZK's `exists`, `children` and `sync` methods
- provides an akka-stream-based interface for CRUD operations, simplifying streaming of data 
- adds support for ZK transactions (create, update, delete and check transaction operations are supported)

A new `SimplePersistenceStoreBenchmark` was added to test the store performance.

JIRA issues: DCOS-22935, DCOS-29741, DCOS-37697


src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
```diff
@@ -0,0 +1,124 @@
...
+    * A Flow for saving nodes to the store. It takes a stream of nodes and returns a stream of node keys
+    * that were successfully stored.
+    *
+    * By default persistent nodes with empty ACLs are created. If the path is nested parent nodes will be also
+    * created if they don't already exist. If there is already a node with the same path a [[org.apache.zookeeper.KeeperException.NodeExistsException]]
+    * is thrown.
+    *
+    * @return
+    */
+  def create: Flow[Node, String, NotUsed]
```
jeschkies: I don't really get this one. Is it creating an empty node?

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+    * A Flow for reading nodes from the store. It takes a stream of node paths and returns a stream of Try[Node] elements.
+    * It's a Success[Node] for existing nodes or Failure(e) for non-existing nodes where e is an instance of
+    * [[org.apache.zookeeper.KeeperException.NoNodeException]]. The exception contains the path to the node which
+    * simplifies handling failed reads. For other exceptions stream is completed with a failure.
+    *
+    * @return
+    */
+  def read: Flow[String, Try[Node], NotUsed]
```
jeschkies: Does this read the root? Should this be private? I'm confused.

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
```diff
@@ -0,0 +1,124 @@
...
+    */
+  def exists(path: String): Future[Boolean]
+
+  /**
+    * Method syncs state with underlying store.
+    *
+    * @param path node path to sync
+    * @return
+    */
+  def sync(path: String): Future[Done]
```
jeschkies: So, when is a Node actually persisted?

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+    * Method takes a list of transaction [[mesosphere.marathon.core.storage.simple.PersistenceStore.StoreOp]] operations and submits them. An exception if thrown if one of the
```
jeschkies: Nit `s/if thrown/is thrown`

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,155 @@
...
+  * same as in the input stream.
+  *
+  * Transaction support is implemented and four transaction operations are supported: [[CreateOp]], [[UpdateOp]],
+  * [[DeleteOp]] and [[CheckOp]]. All transaction operations has to be successful for the transaction to commit successfully.
+  *
+  * @param factory an instance of [[AsyncCuratorBuilderFactory]]
+  * @param parallelism parallelism level for CRUD operations
+  * @param ec execution context
+  */
+class SimplePersistenceStore(factory: AsyncCuratorBuilderFactory, parallelism: Int = 10)(implicit ec: ExecutionContext)
```
jeschkies: Why not `ZooKeeperPersistentStore`?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,155 @@
...
+  def logPath(prefix: String): Flow[String, String, NotUsed]            = Flow[String].map{ p => logger.debug(s"$prefix${p}"); p }
+  def metric[T](counter: Counter, times: Long = 1): Flow[T, T, NotUsed] = Flow[T].map{ e => counter.increment(times); e}
+  // format: ON
+
+  override def create: Flow[Node, String, NotUsed] =
+    Flow[Node]
+      .via(logNode("Creating a node at "))
+      .via(metric(createMetric))
+      .groupBy(parallelism, node => Math.abs(node.path.hashCode) % parallelism)
+      .map(node => factory
```
jeschkies: Hm, if you group by a number what is map receiving? Why do we do the grouping anyways? What is the return type?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,155 @@
...
+        .create()
+        .forPath(node.path, node.data.toArray)
+        .toScala)
```
jeschkies: What is `toScala` doing?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,155 @@
...
+  override def transaction(operations: Seq[StoreOp]): Future[Done] = {
+    logger.debug(s"Submitting a transaction with ${operations.size} operations")
+    transactionCountMetric.increment(operations.size.toLong)
+    transactionMetric.increment()
+
+    val transactionOps = operations.map {
+      case CreateOp(node) => factory.transactionOpCreate().forPath(node.path, node.data.toArray)
+      case UpdateOp(node) => factory.transactionOpSetData().forPath(node.path, node.data.toArray)
+      case DeleteOp(path) => factory.transactionOpDelete().forPath(path)
+      case CheckOp(path) => factory.transactionOpCheck().forPath(path)
```
jeschkies: What is `CheckOp` doing? Is it of value if we return `Done`?

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
zen-dog: No, it just returns a `Flow` from `Node -> String` which when materialized will create node(s) in ZK. You can visualize it like:
```
                                         Flow
                                   ------------
Node(path, data)     |   creating   |     path
------------------> |      node     | ---------->
                                  |      in ZK     |
                                  -------------
```

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
zen-dog: Same as with creating, the flow is `Path -> Node` e.g. it takes a stream of paths and returns a stream of `Try[Node]` (It's `Try` in case the node does not exist)

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
zen-dog: Not sure I understand what you mean. If you look at the implementation, then calling [forPath](https://github.com/mesosphere/marathon/pull/6265/files#diff-68cf4611e0b949cf28991469f497ae2fR68) on the builder submits the request to the curator.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: Too many classes having `Zookeeper` or `Zk` in the name, I guess.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: Builder's `forPath` method submits the request to Apache Curator and returns Java's [CompletitionStage](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html) which is here converted to Scala Future.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: `CheckOp` checks if a node with given path exists. See [this](https://github.com/mesosphere/marathon/pull/6265/files#diff-d6ddb78bb38bf0c43289c466db30259fR350) test.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+
+  // format: OFF
+  private[this] val createMetric          = Metrics.counter(ServiceMetric, getClass, "create")
+  private[this] val readMetric            = Metrics.counter(ServiceMetric, getClass, "read")
+  private[this] val updatedMetric         = Metrics.counter(ServiceMetric, getClass, "update")
+  private[this] val deleteMetric          = Metrics.counter(ServiceMetric, getClass, "delete")
+  private[this] val childrenMetric        = Metrics.counter(ServiceMetric, getClass, "children")
+  private[this] val existsMetric          = Metrics.counter(ServiceMetric, getClass, "exists")
+  private[this] val transactionMetric     = Metrics.counter(ServiceMetric, getClass, "transaction")
+  private[this] val transactionCountMetric = Metrics.counter(ServiceMetric, getClass, "transactionCount")
```
ichernetsky: The name is misleading. It should be `transactionOpCount` or something.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+  *
+  * @param factory an instance of [[AsyncCuratorBuilderFactory]]
+  * @param parallelism parallelism level for CRUD operations
+  * @param ec execution context
+  */
+class SimplePersistenceStore(factory: AsyncCuratorBuilderFactory, parallelism: Int = 10)(implicit ec: ExecutionContext)
+  extends PersistenceStore with StrictLogging {
+
+  // format: OFF
+  private[this] val createMetric          = Metrics.counter(ServiceMetric, getClass, "create")
```
ichernetsky: What about metrics for op times and rates?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+      .via(metric(deleteMetric))
+      .map(path => factory
+        .delete()
+        .forPath(path)
+        .toScala
+        .map(_ => path)) // Note: A Void is returned by the delete call and it can not be passed to the mapAsync stage
+      .mapAsync(parallelism)(identity)
+
+  override def children(path: String): Future[Seq[String]] = {
+    childrenMetric.increment()
```
ichernetsky: Why not `via(metric(...))`?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+    logger.debug(s"Getting children at $path")
+    factory
+      .children()
+      .forPath(path).toScala
+      .map(JavaConverters.asScalaBuffer(_).to[Seq])
+  }
+
+  override def exists(path: String): Future[Boolean] = {
+    existsMetric.increment()
```
ichernetsky: The same here.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+    logger.debug(s"Checking node existence for $path")
+    factory
+      .checkExists()
+      .forPath(path).toScala
+      .map(stat => if (stat == null) false else true)
+  }
+
+  override def sync(path: String): Future[Done] = {
```
ichernetsky: Should we have a metric for `sync` too?

ichernetsky: Also, can we have `def sync(): Future[Done]` too? Because in ZK `path` is not actually used, and it syncs always on `/`.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+    logger.debug(s"Syncing nodes for path $path")
+    factory
+      .sync()
+      .forPath(path).toScala
+      .map(_ => Done)
+  }
+
+  override def transaction(operations: Seq[StoreOp]): Future[Done] = {
+    logger.debug(s"Submitting a transaction with ${operations.size} operations")
+    transactionCountMetric.increment(operations.size.toLong)
```
ichernetsky: `via`?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: Rate could be derived from the counter but op time is a good thing to have. If you know how to measure stream's stage execution time elegantly - I'm all ears. Otherwise, I'd rather implement it in the next PR.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: Well, I don't create a `Source` or `Flow` here and `via` is a stream operator. The idea here is that in contrast to CRUD operations which offer stream interfaces because users might want to e.g. create a serious of nodes, `children`, `exists` and `sync` operations are typically used once so materializing a stream each time would be an unnecessary overhead.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: It's not quite correct - we just always call sync with `"/"` for path [here](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala#L333). But I agree, that this is a sensible default - I will add it.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+      .via(logNode("Creating a node at "))
+      .via(metric(createMetric))
+      // `groupBy(path.hashCode % parallelism)` makes sure that updates to the same path always land in the same
+      // sub-stream, thus keeping the order of writes to the same path, even with parallelism > 1
+      .groupBy(parallelism, node => Math.abs(node.path.hashCode) % parallelism)
+      .map(node => factory
+        .create()
+        .forPath(node.path, node.data.toArray)
+        .toScala)
+      .mapAsync(parallelism)(identity)
```
meln1k: why not doing 
```
+      .map(node => factory
+        .create()
+        .forPath(node.path, node.data.toArray)
+        .toScala)
```
in the mapAsync stage?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
ichernetsky: BTW, I think this is for a reason :)

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
ichernetsky: It is absolutely correct. ZK not matter what path is given when executing `sync` op, syncs on `/`. It is how it works under the hood in ZK.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
alenkacz: +1 for Zk or Zookeeper

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
```diff
@@ -0,0 +1,159 @@
...
+  *
+  * @param factory an instance of [[AsyncCuratorBuilderFactory]]
+  * @param parallelism parallelism level for CRUD operations
+  * @param ec execution context
+  */
+class SimplePersistenceStore(factory: AsyncCuratorBuilderFactory, parallelism: Int = 10)(implicit ec: ExecutionContext)
+  extends PersistenceStore with StrictLogging {
+
+  // format: OFF
+  private[this] val createMetric             = Metrics.counter(ServiceMetric, getClass, "create")
```
alenkacz: since these metrics are not published I think we should go directly to nicer names here once they are established

src/main/scala/mesosphere/marathon/core/storage/simple/AsyncCuratorBuilderFactory.scala
```diff
@@ -0,0 +1,261 @@
...
+  * A simple Factory that give an instance of the [[CuratorFramework]] provides
+  * factory methods to return async builders e.g. [[AsyncCreateBuilder]], [[AsyncDeleteBuilder]] etc.
+  *
+  * @param curator an instance of the [[CuratorFramework]]
+  * @param defaults async builder settings which are used as default parameters which simplifies request building. However
+  *             each factory method also allows overriding the defaults for every call.
+  */
+class AsyncCuratorBuilderFactory(curator: CuratorFramework, defaults: AsyncCuratorBuilderSettings) {
+
+  val async: AsyncCuratorFramework = AsyncCuratorFramework.wrap(curator)
```
alenkacz: shouldn't this guy be private? To not expose the bare curator without defaults... (right now it's used just in the scope of this class)

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
```diff
@@ -0,0 +1,125 @@
...
+  * create, read, update, delete and check methods along with transaction operations and Zookeeper specific children
+  * and sync methods.
+  */
+trait PersistenceStore {
+
+  /**
+    * A Flow for saving nodes to the store. It takes a stream of nodes and returns a stream of node keys
+    * that were successfully stored.
+    *
+    * By default persistent nodes with empty ACLs are created. If the path is nested parent nodes will be also
```
alenkacz: should this really be in the interface description? Since you can actually create the store with any curator with ANY settings?

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
```diff
@@ -0,0 +1,125 @@
...
+    * created if they don't already exist. If there is already a node with the same path a [[org.apache.zookeeper.KeeperException.NodeExistsException]]
+    * is thrown.
+    *
+    * @return
+    */
+  def create: Flow[Node, String, NotUsed]
+  def create(node: Node): Source[String, NotUsed] = Source.single(node).via(create)
```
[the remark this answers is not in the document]
alenkacz: +1 - we could go without the overhead of streams if we do that too, which is nice...

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
alenkacz: maybe it would be more apparent if we aliased the String type here? I mean e.g. via `type NodePath = String` or we could aliased that in the Node class...

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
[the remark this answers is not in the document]
alenkacz: also if we do future for create, we should probably do the same for update, delete etc...

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
alenkacz: I am just kind of alergic to the `Op` suffix, can't we do just Create, Update, Delete etc.?

Also can't we name CheckOp -> Exists or CheckExists?

zen-dog: So far wrappers are called the same as their Curator counterparts. I understand the "allergy" but `Op` here actually makes sense signifying that it's not a `Create` operation but rather a `transactionOpCreate`. Unless you have a better proposal I'm inclined to leave it like this ;)

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: I did not know that :) But in the light of possible future implementation changes on the ZK side, I'm inclined to leave it with a `"/"` as the default.

src/main/scala/mesosphere/marathon/core/storage/simple/PersistenceStore.scala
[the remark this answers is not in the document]
zen-dog: That would lead to a code duplication but can be done. Don't you think it's a premature optimization at this point?

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: I fully agree. Though I wanted to wait until the new Metrics interface is merged and naming rules established.

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: Great point. Done

src/main/scala/mesosphere/marathon/core/storage/simple/SimplePersistenceStore.scala
zen-dog: Nobody appreciates simple things anymore :sigh: 
Renamed.

src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStore.scala
```diff
@@ -0,0 +1,274 @@
...
+  private[this] val childrenMetric           = Metrics.counter(ServiceMetric, getClass, "children")
+  private[this] val existsMetric             = Metrics.counter(ServiceMetric, getClass, "exists")
+  private[this] val transactionMetric        = Metrics.counter(ServiceMetric, getClass, "transaction")
+  private[this] val transactionOpCountMetric = Metrics.counter(ServiceMetric, getClass, "transactionOpCount")
+
+
+  // Helper stages for logging and metrics
+  def logNode(prefix: String): Flow[Node, Node, NotUsed]                = Flow[Node].map{ n => logger.debug(s"$prefix${n.path}"); n }
+  def logPath(prefix: String): Flow[String, String, NotUsed]            = Flow[String].map{ p => logger.debug(s"$prefix${p}"); p }
+  def metric[T](counter: Counter, times: Long = 1): Flow[T, T, NotUsed] = Flow[T].map{ e => counter.increment(times); e}
```
jeschkies: Nit: I'd call it `increment` so that it reads `.via(increment(createMetric))`.

src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStore.scala
```diff
@@ -0,0 +1,274 @@
...
+      .via(metric(createMetric))
+      // `groupBy(path.hashCode % parallelism)` makes sure that updates to the same path always land in the same
+      // sub-stream, thus keeping the order of writes to the same path, even with parallelism > 1
+      .groupBy(parallelism, node => Math.abs(node.path.hashCode) % parallelism)
+      .mapAsync(parallelism)(node =>
+        factory
+          .create()
+          .forPath(node.path, node.data.toArray)
+          .toScala)
+      .mergeSubstreams
```
jeschkies: Why not

```scala
.mapAsync(create)
.mergeSubstream
```

src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStore.scala
```diff
@@ -0,0 +1,274 @@
...
+      .via(metric(readMetric))
+      .mapAsync(parallelism)(path =>
+        factory
+          .getData()
+          .forPath(path)
+          .toScala
+          .map(bytes => Try(Node(path, ByteString(bytes))))
+          .recover {
+            case e: NoNodeException => Failure(e) // re-throw exception in all other cases
+          }
```
jeschkies: Same as above. There is some code duplication and I think the flows should reuse the simple methods that return the futures.

src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStore.scala
```diff
@@ -0,0 +1,274 @@
...
+    * and submits them. An exception is thrown if one of the operations fail. Currently only create, update, delete and
+    * check operations are supported.
+    *
+    * Note: Due to current state of the underlying Curator API, [[mesosphere.marathon.core.storage.zookeeper.PersistenceStore.CreateOp]]s
+    * can't create parent nodes for nested paths if parent nodes does not exist.
+    *
+    * @param operations a list of transaction operations
+    * @return
+    */
+  override def transaction(operations: Seq[StoreOp]): Future[Done] = {
```
jeschkies: How about a transaction builder instead of the ops?

```scala
val result: Future[Done] = transaction().create(node).read(node.path).delete(node.path).run()
```

src/test/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
```diff
@@ -0,0 +1,382 @@
...
+import mesosphere.marathon.core.storage.zookeeper.PersistenceStore._
+import mesosphere.marathon.util.ZookeeperServerTest
+import org.apache.curator.framework.CuratorFramework
+import org.apache.zookeeper.KeeperException.{NoNodeException, NodeExistsException}
+
+import scala.collection.concurrent.TrieMap
+import scala.concurrent.{Await, ExecutionContext}
+import scala.util.{Failure, Random, Success}
+
+class ZooKeeperPersistenceStoreTest extends UnitTest
```
jeschkies: This seems to be an integration test since we use the ZooKeeper server.

src/test/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
```diff
@@ -0,0 +1,382 @@
...
+
+        And("but first 3 nodes were successfully created")
+        created.size shouldBe 3
+        created.keys should contain theSameElementsAs (nodes.map(_.path))
+      }
+    }
+
+    "read" should {
+      "read a single node" in {
+        When("a single node is created")
```
jeschkies: Super Nit: `Given(a single node)`

src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStore.scala
zen-dog: I'll leave it open for the subsequent PR, maybe even from you ;)

src/test/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
zen-dog: It has *neither* Mesos *nor* Marathon - only in-memory ZK. I'd say it's *not* an IT

benchmark/src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStoreBenchmark.scala
```diff
@@ -0,0 +1,166 @@
...
+      Source(1 to num)
+        .map(_ => Node(randomPath("/tests"), ByteString(Random.alphanumeric.take(size).mkString)))
+        .via(store.createFlow)
+        .runWith(Sink.ignore), Duration.Inf)
+  }
+
+  //  @Benchmark
+  def reads(hole: Blackhole) = {
+    def paths = Source(1 to num)
+      .map(_ => s"/tests/$size/node${Random.nextInt(params(size)) + 1}")
```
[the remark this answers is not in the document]
zen-dog: `val params = Map[NodeSize, NodeNumber]((10, 10000), ...` defined above. It stores how many elements of that size where created in the `main` method. This is basically an additional mapping between `@Param size` and  `@Param num`. But because JMH treats both `@Param`s independently from each other, I needed this ugly workaround.

src/test/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStoreTest.scala
jeschkies: I agree.

benchmark/src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStoreBenchmark.scala
```diff
@@ -0,0 +1,166 @@
...
+  *
+  * Benchmarking reading/deleting/updating operations requires a ZK pre-populated with nodes with different payload
+  * sizes. To simplify testing, [[ZooKeeperPersistenceStoreBenchmark.main()]] method could be run manually prior to
+  * the test. Turns out that doing this in the [[org.openjdk.jmh.annotations.Setup]] is not quite trivial since
+  * benchmark iterations are forked and run in different threads and for the purposes of the benchmark one need
+  * consistently named nodes.
+  *
+  * Running multiple benchmark tests at the same time is not sensible e.g. running a [[ZooKeeperPersistenceStoreBenchmark.reads()]]
+  * after [[ZooKeeperPersistenceStoreBenchmark.deletes()]] will read non-existing nodes. To run individual benchmarks just
+  * comment(out) other methods leaving only one method that you want to test. Pre-populate Zookeeper with data if needed.
```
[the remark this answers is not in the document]
zen-dog: That'll work too. But some tests (e.g. read/update/delete) still need pre-populated ZK. But I'll change the comment 👍

PR #6268, opened 2018-05-31
Properly forward HTTP requests which are expected to have body

ichernetsky:
JIRA issues: MARATHON-8084


alenkacz:
@ichernetsky are those tests failures flakes or actual bug?

ichernetsky:
@alenkacz, I believe those were flakes.

PR #6269, opened 2018-05-31
Change the description of the PUT method semantics

meln1k:
Summary: misleading sentence about all apps being upgraded was changed.

JIRA issues: MARATHON-3104

### DIFF
--- a/docs/docs/rest-api/public/api/v2/apps.raml
+++ b/docs/docs/rest-api/public/api/v2/apps.raml
@@ -79,7 +79,7 @@
       If there is an update to an already running application, the application gets upgraded.
       Instances of this application will be started or restarted
       according to the usual logic to meet the requirements of the new definition. Usually, restart is required
-      if the application configuration was changed (except for scaling changes).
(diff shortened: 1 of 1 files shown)


docs/docs/rest-api/public/api/v2/apps.raml
```diff
@@ -77,7 +77,9 @@
     description:
       Change multiple applications either by upgrading existing ones or creating new ones.
       If there is an update to an already running application, the application gets upgraded.
-      All instances of this application get replaced by the new version.
+      Instances of this application will be started or restarted
+      according to the usual logic to meet the requirements of the new definition. Usually, restart is required
+      if the application configuration was changed (except for scaling changes).
```
alenkacz: has changed maybe?

alenkacz: Maybe it would be more clear if we did a new sentence here? (except for scaling changes) -> Restart is not necessary for scaling changes.

PR #6270, opened 2018-06-01
Properly return Content-Type header when using async proxy

timcharper:
Akka HTTP lifts this header out to the entity. We have to explicitly set it when returning the response.

JIRA-Issues: MARATHON-8251


src/main/scala/mesosphere/marathon/api/forwarder/AsyncUrlConnectionRequestForwarder.scala
```diff
@@ -74,6 +74,7 @@ class AsyncUrlConnectionRequestForwarder(
           response.addHeader(name, value)
     }
     response.addHeader(HEADER_VIA, viaValue)
+    response.addHeader(CONTENT_TYPE, remote.entity.contentType.toString)
```
jeschkies: Hm, aren't we copying the content type header in the section above? Great find!

kensipe:
👍 nice work!

PR #6276, opened 2018-06-04
 Pass instance map with offer to task launcher.

jeschkies:
Summary:
We want to make the task launcher actor stateless so that we can remove it eventually.
Thus we don't want it to keep a separate map of instances for a run spec. This change
removes the instance map from the offer matching. A sequence of instances is pushed
with each offer to the task launcher.

JIRA issues: MARATHON-8210


src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -28,6 +28,7 @@ object InstanceUpdater extends StrictLogging {
   }
 
   private[marathon] def reserve(op: Reserve, now: Timestamp): InstanceUpdateEffect = {
+    logger.info(s"Reserve for ${op.instanceId}")
```
jeschkies: These are for debugging purposes and will go eventually.

jeschkies:
I'll rebase this once we start working on phase 2, the feature branch.

PR #6278, opened 2018-06-04
Fix of flaky test_pod_with_container_bridge_network

meln1k:
Summary: added missing retry

JIRA issues: MARATHON-8154

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -772,7 +772,6 @@ def running_status_network_info(task_statuses):
     return running_task_status(task_statuses)['container_status']['network_infos'][0]
 
 
-@retrying.retry(wait_exponential_multiplier=1000, wait_exponential_max=30000, retry_on_exception=ignore_exception)
 def running_task_status(task_statuses):
     """ From a given list of statuses retrieved from mesos API it returns status of running task.
     """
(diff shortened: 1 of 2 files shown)


alenkacz:
@meln1k  can you trigger the SI job for this PR, please?

tests/system/common.py
```diff
@@ -772,6 +772,7 @@ def running_status_network_info(task_statuses):
     return running_task_status(task_statuses)['container_status']['network_infos'][0]
 
 
+@retrying.retry(wait_exponential_multiplier=1000, wait_exponential_max=30000, retry_on_exception=ignore_exception)
```
jeschkies: I don't think this makes a difference since `running_task_status` won't fetch the statuses again. Where is this called?

jeschkies: So, the call is made in `marathon_pods_tests.py`. I think you have to apply the retry [in line 387](https://github.com/mesosphere/marathon/blob/master/tests/system/marathon_pods_tests.py#L387):

```python

@retrying.retry(wait_exponential_multiplier=1000, wait_exponential_max=30000, retry_on_exception=ignore_exception)
def assert_all_pods_healthy(pod_id):
    for task in tasks:
        healthy = common.running_task_status(task['statuses'])['healthy']
        assert healthy, "One of the pod's tasks (%s) is unhealthy" % (task['name'])

assert_all_pods_healthy(pod_id)
```

PR #6280, opened 2018-06-06
Deprecation schedule for /v2/schemas

timcharper:
Describe and enforce /v2/schemas deprecation schedule. Also, replace the concept of "warnVersion" with "softRemovalVersion" in the deprecation mechanism. Add additional test coverage.

JIRA Issues: MARATHON-7978

### DIFF
--- a/src/main/scala/mesosphere/marathon/FeaturesConf.scala
+++ b/src/main/scala/mesosphere/marathon/FeaturesConf.scala
@@ -3,11 +3,9 @@ package mesosphere.marathon
 import org.rogach.scallop.{ScallopConf, ValueConverter}
 
 trait FeaturesConf extends ScallopConf {
-  import DeprecatedFeatures.DeprecatedFeature
-
   private val setStringParser: ValueConverter[Set[String]] = implicitly[ValueConverter[String]].map(parseList(_))
 
-  private val deprecatedFeatureParser: ValueConverter[Set[DeprecatedFeature]] = setStringParser.map { values =>
+  private val deprecatedFeatureParser: ValueConverter[DeprecatedFeatureSet] = setStringParser.map { values =>
     val parsed = values.map { key =>
       (key, DeprecatedFeatures.all.find(_.key == key))
     }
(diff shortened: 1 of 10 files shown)


timcharper:
Per @jdef's query, I've checked through the list https://github.com/mesosphere/marathon/pull/5837#issuecomment-352842124 to see if marathon-client, dcos-ui, or dcos-cli uses this route. As far as I can tell, it is not used by any of those things.

If we fail soak cluster for the 1.7.x release we can make note and extend the soft/hard deprecation schedule accordingly.

src/main/scala/mesosphere/marathon/FeaturesConf.scala
```diff
@@ -32,13 +31,13 @@ trait FeaturesConf extends ScallopConf {
     validate = validateFeatures
   )(setStringParser)
 
-  lazy val deprecatedFeatures = opt[Set[DeprecatedFeature]](
+  lazy val deprecatedFeatures = opt[DeprecatedFeatureSet](
     "deprecated_features",
     descr = "A comma-separated list of deprecated features to continue to enable",
```
alenkacz: to continue to enable -> that still should be enabled ?

src/main/scala/mesosphere/marathon/FeaturesConf.scala
```diff
@@ -32,13 +31,13 @@ trait FeaturesConf extends ScallopConf {
...
     required = false,
-    default = Some(Set.empty),
+    default = Some(DeprecatedFeatureSet(BuildInfo.version, Set.empty)),
     noshort = true,
-    validate = validateDeprecatedFeatures)(deprecatedFeatureParser)
+    validate = { dfs => dfs.isValid() })(deprecatedFeatureParser)
```
alenkacz: I tried going through `ScallopConf` source code to find out what exactly happens when valueconverter throws exception (which is what we do with the require) but I guess it will blow up so we should be fine :)

timcharper: Hmm, `.isValid` doesn't throw an exception, it just returns false. A param failing validation will cause Marathon to not launch

src/main/scala/mesosphere/marathon/FeaturesConf.scala
timcharper: Hmm, I do feel like this wording is fine :)

PR #6281, opened 2018-06-06
When proxying events, flush after each element received

timcharper:
Previously, we used intersperse, which would (understandably) prefix elements
with the flush sequence. This was not the desired behavior.

JIRA Issues: MARATHON-8256


alenkacz:
nice catch 👏

kensipe:
I'm feeling this was not a "nice catch"...   this was catch by DCOS tests and not our own.   We definitely should have our own tests that should catch this.

PR #6283, opened 2018-06-06
Log deprecation warnings and errors on param validation

timcharper:
This was a minor regression


timcharper:
It went to the wrong branch and then was force-pushed over, so I downloaded `https://github.com/mesosphere/dcos-docs-site/commit/dc4c3f56608a4ec822689e32d716f35f512099c4.patch` and used `git am` to apply to staging. It is at 220796b747bc2cbd89707a2c7a064efc9426ed5d

PR #6284, opened 2018-06-06
Deprecate /v2/events heavy event format

timcharper:
- Also, the deprecated features description was showing the wrong
  text; fix this, also.

JIRA Issues: MARATHON-8259

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -40,11 +40,10 @@ The default response format of the `/v2/events` is marked as deprecated and will
 
 #### Deprecation Details
 
-The "lightweight" plan format can be already seen using the `?plan-format=light` argument. In summary, this format drops the following fields from the HTTP response, since they are considered redundant:
+The "lightweight" plan format can be already seen using the `?plan-format=light` argument. In summary, this format drops the following fields from the deployment-related events in the event stream accessed via /v2/events:
 
 * `plan.original` - The current state of the root group
 * `plan.target` - The target state of the root group
-
 
 ## Change from 1.6.322 to 1.6.352
 
--- a/changelog.md
+++ b/changelog.md
@@ -32,7 +32,7 @@ The route `/v2/schemas` has the following deprecation schedule:
 
 ### /v2/events
 
-The default response format of the `/v2/events` is marked as deprecated and will be permanently switched to the `/v2/events?plan-format=light` in the next release. The following deprecation schedule is planned for this endpoint:
+The default response format of the `/v2/events` is marked as deprecated and will be switched to the `/v2/events?plan-format=light` in the first 1.7.x release. The following deprecation schedule is planned for this endpoint:
 
 * 1.6.x - `/v2/events`  will continue to function as normal
 * 1.7.x - The default `/v2/events` format will be switched to "light". You will still have the ability to use the command-line argument `--deprecated_features=api_heavy_events` to re-enable the heavy event response.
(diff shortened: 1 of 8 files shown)


changelog.md
```diff
@@ -30,6 +30,21 @@ The route `/v2/schemas` has the following deprecation schedule:
 - 1.8.x - `/v2/schemas` is scheduled to be completely removed. If `--deprecated_features=json_schemas_resource` is
   still specified, Marathon will refuse to launch, with an error.
 
+### /v2/events
+
+The default response format of the `/v2/events` is marked as deprecated and will be switched to the `/v2/events?plan-format=light` in the first 1.7.x release. The following deprecation schedule is planned for this endpoint:
```
alenkacz: shouldn't we do v3/events? I know it feels weird but ... we have kind of "versioned" API and still we are doing breaking change. I think it's better to decomission an endpoint than switch to a different format under the hood...

Just something to discuss...

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -21,13 +21,16 @@ import scala.concurrent.{Await, blocking}
...
   */
-class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter) extends HttpEventStreamHandle {
+class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter, allowHeavyEvents: Boolean) extends HttpEventStreamHandle {
 
   lazy val id: String = UUID.randomUUID().toString
 
   private val subscribedEventTypes = request.getParameterMap.getOrDefault("event_type", Array.empty).toSet
 
-  private val useLightWeightEvents = request.getParameterMap.getOrDefault("plan-format", Array.empty).contains("light")
+  private val useLightWeightEvents = if (allowHeavyEvents)
```
alenkacz: so `request.getParameterMap.getOrDefault("plan-format", Array.empty).contains("light") || !allowHeavyEvents` ... ? :)

changelog.md
timcharper: I thought about this. I feel like the current strategy is fine because:

- Fields are only removed, not renamed or changed type
- The event fields have never been formally specified
- If it causes issues, there is an option to add the problematic fields back while the clients remove their dependencies on these fields.

I really don't think it will be a problem. If it does, we have an safety hatch, and we can re-evaluate plans to fully remove in 1.8.x.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
timcharper: I actually wrote was you suggested first and felt like this was clearer

timcharper:
I added a test. It's a bit cheesy, but it is a test :)

changelog.md
alenkacz: ok, makes sense.

btw. removing fields is breaking change, only adding new fields is ok... :)

e.g. this guy will break, correct? 
https://github.com/gambol99/go-marathon/blob/master/events.go#L279

so we should at least pro-actively help them migrate...

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
alenkacz: ok, it's a matter of taste, I would prefer the shorter one, but keep it as is if you like that one more :)

PR #6285, opened 2018-06-06
Run scapegoat before integration tests

alenkacz:


ci/pipeline
```diff
@@ -30,7 +30,7 @@ val PACKAGE_DIR: Path = pwd / 'target / 'universal
 val PACKAGE_DOCS_DIR: Path = pwd / 'target / "universal-docs"
 
 /**
- * Compile Marathon and run unit and integration tests followed by scapegoat.
+ * Compile Marathon and run unit and integration tests and run scapegoat.
```
ichernetsky: I guess it makes sense to change the order of steps in the comment too.

alenkacz: I rephrased it so that it does not feel like order anymore. Was I not successful? :)

PR #6286, opened 2018-06-07
Use handler for transfer metrics

timcharper:
* Improve naming of metric module. HttpMetrics was too general as we
  already have http metrics.
* Add counter to track gzipped transfer bytes
* Clarify the wiring logic for metrics and handler. Previously, we
  would set up the handlerCollection, only to clear and reconfigure it
  elsewhere. This was a source of unnecessary surprise.

JIRA Issues: MARATHON-8263


src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
```diff
@@ -0,0 +1,130 @@
...
+        updateResponse(baseRequest)
+    }
+  }
+
+  private def isGzip(contentEncodingHeader: String): Boolean =
+    (contentEncodingHeader != null) && contentEncodingHeader.equalsIgnoreCase("gzip")
+
+  protected def updateResponse(request: Request) =
+    if (!isExcluded(request)) {
+      val bytesRead = request.getContentLengthLong
```
ichernetsky: I guess we counted only request/response body sizes before as well?.. And it seems that it is non-trivial to count true request/response sizes.

src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
```diff
@@ -0,0 +1,130 @@
...
+  val bytesWrittenMetric: Counter
+}
+
+/**
+  * Implementation class for HttpMetrics
+  *
+  * This class is structured and named this way for backwards compatibility, due to the fact that our metrics are
+  * currently (and lamentably) tied to classes.
+  */
+class HTTPMetricsFilter extends HttpTransferMetrics {
```
timcharper: This is obnoxious. But we've already released the metric with this path? If we decide we can change the metric key then we should fold in to HttpFilter but it seems this would be a breaking change.

ichernetsky: We released it even in 1.5.x

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -116,6 +116,11 @@ class HttpEventStreamServlet(
     @volatile private var handler: Option[HttpEventSSEHandle] = None
 
     override def onOpen(emitter: Emitter): Unit = {
+      // We don't want to count this response towards the http metrics as it could be quite large by the time it closes.
+      // Also, the serialization for events is done once for all consumers; we should track the data with a separate
+      // metric.
+      HttpTransferMetricsHandler.exclude(request)
```
wavesoft: Good point! Created [MARATHON-8264](https://jira.mesosphere.com/browse/MARATHON-8264) to keep track of this.

src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
timcharper: https://github.com/mesosphere/marathon/pull/6288

PR #6288, opened 2018-06-07
Include gzipped data in overall metrics, remove gzipped read data

timcharper:
- Switch from input entity ContentLength header to httpInput getContentConsumed, as the client may not specify the `Content-Length`.
- Also, remove gzippedBytesRead, as we don't even support reading gzipped entities.

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
+++ b/src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
@@ -86,17 +86,17 @@ class HttpTransferMetricsHandler(httpMetrics: HttpTransferMetrics) extends Abstr
     if (!isExcluded(request)) {
       val bytesRead = request.getContentLengthLong
       val bytesWritten = request.getResponse.getContentCount
-      if (bytesRead > 0) // bytesRead can be -1 if no entity provided
+      if (bytesRead > 0) { // bytesRead can be -1 if no entity provided
+        httpMetrics.bytesReadMetric.increment(bytesRead)
         if (isGzip(request.getHeader(HttpHeaders.CONTENT_ENCODING)))
           httpMetrics.gzippedBytesReadMetric.increment(bytesRead)
-        else
-          httpMetrics.bytesReadMetric.increment(bytesRead)
+      }
 
-      if (bytesWritten > 0) // bytesWritten can be -1 if no entity returned
+      if (bytesWritten > 0) { // bytesWritten can be -1 if no entity returned
+        httpMetrics.bytesWrittenMetric.increment(bytesWritten)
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
```diff
@@ -86,17 +86,17 @@ class HttpTransferMetricsHandler(httpMetrics: HttpTransferMetrics) extends Abstr
     if (!isExcluded(request)) {
       val bytesRead = request.getContentLengthLong
```
wavesoft: I just realised that I _did_ had a look on this variable when I wrote the first version of the http metrics. The `getContentLength` is just returning the value of the `Content-Length` request header 😄 

In principle, most of the HTTP clients populate this correctly, but the bottom-line is that this is still a header. That's why I went with counting the physical bytes written in the stream ¯\\\_(ツ)\_/¯ .

However, since this is just a metric and not a critical part of marathon, we are good for 99% of the cases 👍

src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
```diff
@@ -86,17 +86,17 @@ class HttpTransferMetricsHandler(httpMetrics: HttpTransferMetrics) extends Abstr
...
       val bytesWritten = request.getResponse.getContentCount
-      if (bytesRead > 0) // bytesRead can be -1 if no entity provided
+      if (bytesRead > 0) { // bytesRead can be -1 if no entity provided
+        httpMetrics.bytesReadMetric.increment(bytesRead)
```
wavesoft: This is the compressed length. Right? Is there any way we can get the uncompressed length instead?

timcharper: Yes, there is a way. It wasn't as obvious but the Jetty HttpInputStream has a counter.

Additionally, we do not support gzipped posts. I looked at the Jetty gzip handler. It's simply not implemented. So I just removed the gzipped metric.

src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
timcharper: Updated. It uses the bytes read counter now.

timcharper: (It wasn't as obvious to find the counter for bytes read, but after some digging I found it)

src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
timcharper: Also, for bytes written, it is decompressed length. So the same as what we were tracking before.

PR #6289, opened 2018-06-07
Use handler for transfer metrics

timcharper:
Backport of #6286, #6288

* Use handler for transfer metrics
* Add counter to track gzipped transfer bytes

JIRA Issues: MARATHON-8263


src/main/scala/mesosphere/marathon/api/HttpTransferMetricsHandler.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+    * Metric representing the total HTTP entity gzipped bytes written (does not include HTTP headers)
+    */
+  val gzippedBytesWrittenMetric: Counter
+
+  /**
+    * Metric representing the total HTTP entity bytes read (does not include HTTP headers)
+    */
+  val bytesReadMetric: Counter
```
kensipe: I thought we were taking this out?

timcharper: Just gzippedBytesRead :)

PR #6290, opened 2018-06-11
Remove unreachable resident task terminal status surrogate

timcharper:
Summary:
Before Mesos 1.4.0, on reboot agents would come up with different IDs. This
meant that unreachable tasks on the node would never become terminal, forever
blocking the recovery of resident tasks. In MARATHON-2311 we created a
mechanism to treat an offer with reservations for a task as an indication of
the task being gone.

That is fixed in Mesos 1.4.0 and I have confirmed that terminal statuses are
now received for unreachable tasks when rebooted hosts come back online.

JIRA Issues: MARATHON-8272


timcharper:
Needs discussion; I have confirmed that this workaround is no longer needed for the exact scenario for which it was implemented; we may want to wait until 1.7.x to remove this? I'm not entirely certain yet.

timcharper:
Thanks, Alena :-)

I have confirmed that the case about which I was most worried does in fact work (Mesos master fully forgets about the unreachable tasks and associated agent due to a master failover). I populated the acceptance criteria describing that case.

So, after sleeping on it for night, and further experimenting, I feel confident that we can go ahead with this change :-) agreed that no code is better than unnecessary code :)

ichernetsky:
Looks good to me, but I think `LibMesos.scala` has to be updated to account for this change.

timcharper:
As @ichernetsky brought up, removing this workaround should also bump our minimum version of Mesos. As such, we should wait until Marathon 1.7.x to merge this.

jeschkies:
> As @ichernetsky brought up, removing this workaround should also bump our minimum version of Mesos. As such, we should wait until Marathon 1.7.x to merge this.

How about cutting a 1.6 branch now and bump master to 1.7?

timcharper:
@vinodkone brought up that there are other cases in which an agent ID might change, and terminals task status updates are propagated. I'm waiting to hear back whether or not if in these cases we would also lose reservations. If so, then we can proceed with the merge. If not, we may wish to keep this logic. Unsure at this point.

jeschkies:
@timcharper, we've bumped Marathon to 1.7 on master. Once you've resolved the merge conflict this is good to land.

jeschkies:
@timcharper, we've bumped Marathon to 1.7 on master. Once you've resolved the merge conflict this is good to land.

timcharper:
Vinod comments that there are still some edge cases where meta-data may be deleted but reservations preserved. From the description, these cases seem a bit rare. I'm inclined to think we can perhaps explore some alternate solutions here, perhaps an API call to force Marathon to treat an unreachable status as terminal for some task, or (maybe easier) instructions to use the storage tool to update them.

timcharper:
Per the Mesos docs, this call will get users out of such situations. http://mesos.apache.org/documentation/latest/operator-http-api/#mark_agent_gone

I have increased the minimum Mesos requirement to 1.5.0, the Mesos version which introduced this Mesos call.

Additionally, I added a changelog section and updated the Mesos compatibility matrix.

PR #6291, opened 2018-06-11
Using custom executor to scale tests to 10k

wavesoft:
This commit bumps the scale tests to 2k apps/instances on marathon. To achieve this we are using
the a custom mesos executor. In detail:

1. This commit customizes the `marathon_args` on the marathon-dcluster in order to enable the custom plugin that allows pinning tasks on a single executor.
2. This commit changes the app definition template in order to instruct mesos to use the custom no-op executor.

### DIFF
--- a/tests/performance/config/dcluster/large-cluster.conf
+++ b/tests/performance/config/dcluster/large-cluster.conf
@@ -7,7 +7,7 @@ zookeeper = 3.4
 
 # Overcommit the resources, allocing deployment of at least 1000 apps
 # with the minimum requirements
-mesos_slaves = 6
+mesos_slaves = 10
 mesos_resources_cpus = 500
 mesos_resources_mem = 67108864
 
--- a/tests/performance/config/dcluster/large-cluster.conf
+++ b/tests/performance/config/dcluster/large-cluster.conf
@@ -21,4 +21,4 @@ mesos_master_port = 5050
 # from /opt/shared (the marathon-perf-testing infrastructure mounts there the
 # contents of the `files` folder from here
 # : https://github.com/mesosphere/marathon-perf-testing/tree/master/files )
-marathon_args = --max_instances_per_offer=1000 --launch_tokens=1000 --max_running_deployments=10000 --plugin_dir=/opt/shared/plugins-1.6 --plugin_conf=/opt/shared/executorid-plugin-config.json
+marathon_args = --max_instances_per_offer=100 --max_running_deployments=10000 --plugin_dir=/opt/shared/plugins-1.6 --plugin_conf=/opt/shared/executorid-plugin-config.json
--- a/tests/performance/config/dcluster/large-cluster.conf
+++ b/tests/performance/config/dcluster/large-cluster.conf
@@ -21,4 +21,4 @@ mesos_master_port = 5050
 # from /opt/shared (the marathon-perf-testing infrastructure mounts there the
 # contents of the `files` folder from here
 # : https://github.com/mesosphere/marathon-perf-testing/tree/master/files )
-marathon_args = --max_running_deployments=10000 --max_instances_per_offer=1000 --plugin_dir=/opt/shared/plugins-1.6 --plugin_conf=/opt/shared/executorid-plugin-config.json
+marathon_args = --max_instances_per_offer=1000 --launch_tokens=1000 --max_running_deployments=10000 --plugin_dir=/opt/shared/plugins-1.6 --plugin_conf=/opt/shared/executorid-plugin-config.json
--- a/tests/performance/config/dcluster/large-cluster.conf
+++ b/tests/performance/config/dcluster/large-cluster.conf
@@ -21,4 +21,4 @@ mesos_master_port = 5050
 # from /opt/shared (the marathon-perf-testing infrastructure mounts there the
 # contents of the `files` folder from here
 # : https://github.com/mesosphere/marathon-perf-testing/tree/master/files )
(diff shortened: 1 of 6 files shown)


tests/performance/config/perf-driver/policies/test-1-apps-n-instances.yml
```diff
@@ -48,9 +48,7 @@ policies:
       - name: Scale Instances
         values:
           - parameter: instances
-            min: 50
-            step: 50
-            max: 500
+            values: [500, 1000, 2000]
```
zen-dog: How about simply `values: [5000]`? ;)

tests/performance/config/perf-driver/policies/test-n-apps-1-instances.yml
```diff
@@ -48,9 +48,7 @@ policies:
       - name: Scale Apps
         values:
           - parameter: apps
-            min: 50
-            step: 50
-            max: 500
+            values: [500, 1000, 2000]
```
zen-dog: same here ^^

tests/performance/config/perf-driver/policies/test-1-apps-n-instances.yml
wavesoft: I would vote against this, for various reasons, but mainly:

1) Having (at least) 3 data points allows you to interpolate/extrapolate the numbers in other regions that you haven't collected data for.
2) Some performance issues show some tendency, and having at least 3 data points helps you identify them.
3) Having enough (ok, perhaps more than 3) data points, helps you detect outliers.

Nonetheless, with just 3 data points the run time has been decreased down to about 20 minutes, and most of it is just building marathon 😄 . I don't think we gain much by reducing this down to a single point...

_(Though I can make them `[1000, 3500, 6000]` if you like)_

zen-dog: 1. We do have multiple data points - these are the deployment time for individual deployments (or Instance-started events for 1app-Ninstances scenario) 
2. We will gather more data by running perf tests in the loop every 30m or so
3. We will gather more interesting data by sending percentiles to DD

wavesoft: Coming back to what you want to do, you should check the #6300 PR instead.
I am going to insist that we have (at least) 3 data points in this PR, because it's intended for a different purpose than what (I think) you have in mind.

tests/performance/config/dcluster/large-cluster.conf
```diff
@@ -7,15 +7,18 @@ zookeeper = 3.4
...
 
 # Increase the number of instances launched per offer since the performance
-# tests should put pressure on marathon.
-marathon_args = "--max_instances_per_offer=500"
\ No newline at end of file
+# tests should put pressure on marathon. Also load the executorid plugin
+# from /opt/shared (the marathon-perf-testing infrastructure mounts there the
+# contents of the `files` folder from here
+# : https://github.com/mesosphere/marathon-perf-testing/tree/master/files )
+marathon_args = --max_instances_per_offer=1000 --launch_tokens=1000 --max_running_deployments=10000 --plugin_dir=/opt/shared/plugins-1.6 --plugin_conf=/opt/shared/executorid-plugin-config.json
```
zen-dog: I would tune down the numbers. Otherwise, it's probably "too few" cycles for marathon to start all the tasks. How about `--max_instances_per_offer=100` and `--max_running_deployments=100`? I'm also not sure if `--launch_tokens=1000` has any real effect.

wavesoft: 1. `max_running_deployments` should be equal to the number of the instances to be deployed. This is a current limitation of the test, since it tries to submit them as quickly as possible and I would like to avoid re-tries in the scope of the test because the re-try interval is going to affect the results.

2. `--max_instances_per_offer=100` sure. I was just trying to minimise the deployment time.

3. It _does_ have an affect, but I think such a big value might have a negative effect. _Take the following with a grain of salt, since the data that I have collected are not big enough, but_ I do notice some tendency. For example, check this: 
image

zen-dog: Ok, got it. Lgtm

tests/performance/config/perf-driver/policies/test-n-apps-1-instances.yml
```diff
@@ -48,9 +48,7 @@ policies:
       - name: Scale Apps
         values:
           - parameter: apps
-            min: 50
-            step: 50
-            max: 500
+            values: [100, 1000]
```
zen-dog: Again, didn't we agree on a single value/run here and above: `10.000`?

wavesoft: Again, this particular test has 2 points, and will keep having 2 data points. 
**This** is the test that has a single data point : https://github.com/mesosphere/marathon/pull/6300

To be more precise, this test should sample 100, 1k and 10k. However there are issues deploying 10k apps on the v1.5.9 (reference), that I am trying to solve elegantly. Since this process is taking too long, I vote for merging this test with the 2 data points (100, 1k) so we can unblock #6300 , and then I can add the 10k data point when I have a solution.

zen-dog: > This is the test that has a single data point : #6300
it looks like it has 3 datapoints to me ;)

PR #6292, opened 2018-06-12
Remove launch ephemeral

alenkacz:
To be honest, most of the tests there are really horrible and we should probably rewrite them from scratch because it's just hard to refactor anything with tests like that 🙈 

On a different note - we are removing launch ephemeral because Provision is in place and should be used instead of that (for both residents and non-residents). Actually this was already obsolete and used only in tests.

Most of the changes in tests are because when provisioning, the instance must already exist in the `InstanceTracker` (because it has to be scheduled first). Before, `LaunchEphemeral` was the first command that created instance in the instance tracker.

JIRA issues: MARATHON-8221


src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -314,163 +314,46 @@ class TaskLauncherActorTest extends AkkaUnitTest {
...
 
-    "Expunged task is removed from counts" in new Fixture {
-      val update = TaskStatusUpdateTestHelper.finished(f.marathonInstance).wrapped
-      val expectedCounts = QueuedInstanceInfo(f.app, inProgress = false, 0, 0, Timestamp(0), Timestamp(0))
-
-      Mockito.when(instanceTracker.instancesBySpecSync).thenReturn(InstanceTracker.InstancesBySpec.forInstances(f.marathonInstance))
-
-      val launcherRef = createLauncherRef()
-      launcherRef ! RateLimiterActor.DelayUpdate(f.app, clock.now())
+    "Revive offers if instance deleted" in new Fixture {
```
alenkacz: this class was greatly simplified in the feature branch so I had to adjust the tests a bit more...

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
```diff
@@ -170,6 +174,27 @@ object TestInstanceBuilder {
     None
   )
 
+  def runningInstance(appId: PathId, version: Timestamp, instanceTracker: InstanceTracker): Future[(Instance.Id, Task.Id)] = async {
```
jeschkies: Could you add some docs?

I'm not sure if we really should introduce this helper. As I understand we want to save a running instance in the tracker, right? Ideally this should be a two liner. Could we create a running instance with the test builder?

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -74,13 +66,15 @@ class TaskLauncherActorTest extends AkkaUnitTest {
...
+    Mockito.when(instanceTracker.instancesBySpecSync).thenReturn(InstanceTracker.InstancesBySpec.forInstances(f.scheduledInstance))
+
     def createLauncherRef(appToLaunch: AppDefinition = f.app): ActorRef = {
       val props = TaskLauncherActor.props(
         launchQueueConfig,
         offerMatcherManager, clock, instanceOpFactory,
         maybeOfferReviver = Some(offerReviver),
         instanceTracker, rateLimiterActor.ref, offerMatchStatisticsActor.ref, localRegion) _
-      system.actorOf(props(appToLaunch))
+      TestActorRef[TaskLauncherActor](props(appToLaunch))
```
jeschkies: This does not work with `stash` in the actor system. We have to use a real actor.

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -44,20 +42,14 @@ class TaskLauncherActorTest extends AkkaUnitTest {
 
   import org.mockito.{Matchers => m}
 
-  def sendUpdate(launcherRef: ActorRef, update: InstanceChange): QueuedInstanceInfo = {
```
jeschkies: Thanks! This little method was obfuscating important details.

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -314,163 +306,46 @@ class TaskLauncherActorTest extends AkkaUnitTest {
...
-    "Expunged task is removed from counts" in new Fixture {
-      val update = TaskStatusUpdateTestHelper.finished(f.marathonInstance).wrapped
-      val expectedCounts = QueuedInstanceInfo(f.app, inProgress = false, 0, 0, Timestamp(0), Timestamp(0))
-
-      Mockito.when(instanceTracker.instancesBySpecSync).thenReturn(InstanceTracker.InstancesBySpec.forInstances(f.marathonInstance))
-
-      val launcherRef = createLauncherRef()
-      launcherRef ! RateLimiterActor.DelayUpdate(f.app, clock.now())
+    "Revive offers if instance deleted" in new Fixture {
+      val constraint = Protos.Constraint
```
jeschkies: Why the constraint?

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -314,163 +306,46 @@ class TaskLauncherActorTest extends AkkaUnitTest {
...
-      }
-    }
-
-    "Updated task is reflected in counts" in new Fixture {
-      val update = TaskStatusUpdateTestHelper.runningHealthy(f.marathonInstance)
-      val expectedCounts = QueuedInstanceInfo(f.app, inProgress = false, 0, 1, Timestamp(0), Timestamp(0))
-
-      Mockito.when(instanceTracker.instancesBySpecSync).thenReturn(InstanceTracker.InstancesBySpec.forInstances(f.marathonInstance))
-
+    "do instance tracker call after InstanceChange event" in new Fixture {
```
jeschkies: How about `should { "call instance tracker after InstanceChange event" in {} }`?

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -314,163 +306,46 @@ class TaskLauncherActorTest extends AkkaUnitTest {
...
-        // task status update
-        val counts = sendUpdate(launcherRef, update.wrapped)
-
-        assert(counts.finalInstanceCount == 1)
-        assert(!counts.inProgress)
-        assert(counts.instancesLeftToLaunch == 0)
-
-        verifyClean()
-      }
+      Mockito.verify(instanceTracker, times(2)).instancesBySpecSync
```
jeschkies: Why do we test this? Isn't this an implementation detail?

src/test/scala/mesosphere/marathon/tasks/InstanceTrackerImplTest.scala
```diff
@@ -60,40 +62,40 @@ class InstanceTrackerImplTest extends AkkaUnitTest {
     }
 
     def testList(call: InstanceTracker => InstanceTracker.InstancesBySpec)(implicit instanceTracker: InstanceTracker): Unit = {
-      val instance1 = makeSampleInstance(TEST_APP_NAME / "a")
-      val instance2 = makeSampleInstance(TEST_APP_NAME / "b")
-      val instance3 = makeSampleInstance(TEST_APP_NAME / "b")
+      val instance1 = Instance.Scheduled(AppDefinition(testAppId / "a"))
+      val instance2 = Instance.Scheduled(AppDefinition(testAppId / "b"))
+      val instance3 = Instance.Scheduled(AppDefinition(testAppId / "b"))
```
jeschkies: This is so much nicer. Ideally we use `Instance.Provisioned(...)` and `Instance.Running(...)` factories instead of the test instance builders.

PR #6293, opened 2018-06-12
Introduce Goal enumeration

alenkacz:
Summary:
Introduce the Goal enumeration.

JIRA issues: MARATHON-8244


timcharper:
Looks good so far! Do you want to merge this when the enum values are used by the scheduler?

PR #6294, opened 2018-06-13
Wait for leader abdication to complete in system tests

timcharper:
Marathon does not immediately abdicate when asked to do so (500ms delay)

JIRA Issues: MARATHON-8276

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -681,7 +681,7 @@ def delete_marathon_path(name, marathon_name='marathon'):
     url = get_marathon_endpoint(name, marathon_name)
     return http.delete(url)
 
-@retrying.retry(wait_fixed=1000, stop_max_attempt_number=60, retry_on_result = lambda a: a)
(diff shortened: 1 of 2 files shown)


tests/system/common.py
```diff
@@ -676,11 +676,30 @@ def http_get_marathon_path(name, marathon_name='marathon'):
...
+
+    Accounts for Marathon abdication delay of 0.5 seconds. Waits for just a little longer.
+
+    params arg should include a "?" prefix.
+    """
+    leader_endpoint = get_marathon_endpoint('/v2/leader', marathon_name)
+    result = http.delete(leader_endpoint + params)
+
+    down=False
+    while not down:
```
zen-dog: We prefer `@retrying.retry` methods to the loops. See here for example:
https://github.com/mesosphere/marathon/pull/6294/files#diff-984c5562f7187dee9e6d3e161c7b52f8L116

timcharper: ✅

PR #6296, opened 2018-06-13
Run performance tests using local Marathon

wavesoft:
Summary:
This commit introduces the `manual_run.sh` script that can be used to launch
the scale tests targeting a Marathon instance solely by it's URL.


tests/performance/manual_run.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
jeschkies: Out of curiosity, did you run shellcheck?

PR #6298, opened 2018-06-14
Fix for flaky SI `test_unhealthy_app_can_be_rolled_back`

meln1k:
Summary: wait_for_service_endpoint was ported from shakedown and rewritten in order to reduce flakiness.

JIRA issues: MARATHON-8155

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -18,8 +18,8 @@ from json.decoder import JSONDecodeError
 from shakedown import marathon
 from urllib.parse import urljoin
 from shakedown.dcos.master import get_all_master_ips
-from shakedown.dcos import dcos_service_url
 from dcos.http import DCOSAcsAuth
+from functools import lru_cache
 
 
 marathon_1_3 = pytest.mark.skipif('marthon_version_less_than("1.3")')
@@ -825,26 +825,30 @@ def kill_process_on_host(hostname, pattern):
     return pids
 
 
-def service_available_predicate(service_name):
-    url = dcos_service_url(service_name)
-    try:
-        response = http.get(url)
-        return response.status_code == 200
-    except Exception as e:
-        return False
-
-
-# since exhibitor might take ~ 20 minutes to start up, we need to wait for it
-def wait_for_service_endpoint(service_name, timeout_sec=120):
-    """Checks the service url if available it returns true, on expiration
-    it returns false"""
-
+@lru_cache()
+def dcos_masters_public_ips():
     @retrying.retry(
         wait_fixed=1000,
         stop_max_attempt_number=240,  # waiting 20 minutes for exhibitor start-up
         retry_on_exception=ignore_provided_exception(DCOSException))
     def all_master_ips():
         return get_all_master_ips()
+
+    master_private_ips = all_master_ips()
+
+    master_public_ips = []
+
+    for private_ip in master_private_ips:
+        result, ip = shakedown.run_command(private_ip, '/opt/mesosphere/bin/detect_ip_public')
+        master_public_ips.append(ip)
+
+    return master_public_ips
+
+
+# since exhibitor might take ~ 20 minutes to start up, we need to wait for it
+def wait_for_service_endpoint(service_name, timeout_sec=120):
+    """Checks the service url if available it returns true, on expiration
+    it returns false"""
 
     @retrying.retry(
             wait_fixed=1000,
@@ -869,12 +873,4 @@ def wait_for_service_endpoint(service_name, timeout_sec=120):
             print(response)
             raise DCOSException("service " + service_name + " is unavailable at " + master_ip)
 
-    master_private_ips = all_master_ips()
-
-    master_public_ips = []
-
-    for private_ip in master_private_ips:
-        result, ip = shakedown.run_command(private_ip, '/opt/mesosphere/bin/detect_ip_public')
-        master_public_ips.append(ip)
-
-    return all(check_service_availability_on_master(ip, service_name) for ip in master_public_ips)
+    return all(check_service_availability_on_master(ip, service_name) for ip in dcos_masters_public_ips())
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -20,7 +20,7 @@ from urllib.parse import urljoin
 from shakedown.dcos.master import get_all_master_ips
 from dcos.http import DCOSAcsAuth
 from functools import lru_cache
-from fixtures import get_ssl_context, get_ca_file
+from fixtures import get_ca_file
 
 
 marathon_1_3 = pytest.mark.skipif('marthon_version_less_than("1.3")')
@@ -851,6 +851,13 @@ def wait_for_service_endpoint(service_name, timeout_sec=120):
     """Checks the service url if available it returns true, on expiration
     it returns false"""
 
+    def verify_ssl():
+        cafile = get_ca_file()
+        if cafile.is_file():
+            return str(cafile)
+        else:
+            return False
+
     @retrying.retry(
             wait_fixed=1000,
             stop_max_attempt_number=timeout_sec/5,  # underlying http.get has 5 seconds timeout, so we have to scale it
@@ -866,7 +873,7 @@ def wait_for_service_endpoint(service_name, timeout_sec=120):
             url=url,
             timeout=5,
             auth=auth,
-            verify=str(get_ca_file()))
+            verify=verify_ssl())
         if response.status_code == 200:
             return True
         else:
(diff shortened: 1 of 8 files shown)


tests/system/common.py
```diff
@@ -819,3 +819,11 @@ def kill_process_on_host(hostname, pattern):
     else:
         print("Killed no pids")
     return pids
+
+
+# since exhibitor might take ~ 20 minutes to start up, we need to wait for it
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=1200, retry_on_exception=ignore_exception)
```
alenkacz: do we really want to wait 20 minutes for exhibitor? What kind of software starts up 20 minutes? That's crazy...

Shouldn't we maybe give up earlier but fail with message that would clearly state that we failed waiting on exhibitor?

Btw. the usage of exhibitor in shakedown is kind of questionable too. Waiting 20 minutes for current number of masters... can't we get that information from somewhere else?

tests/system/common.py
```diff
@@ -819,3 +822,26 @@ def kill_process_on_host(hostname, pattern):
...
+        return response.status_code == 200
+    except Exception as e:
+        return False
+
+
+# since exhibitor might take ~ 20 minutes to start up, we need to wait for it
+def wait_for_service_endpoint(service_name, timeout_sec=120):
+    """Checks the service url if available it returns true, on expiration
+    it returns false"""
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=1200, retry_on_exception=ignore_exception)
```
alenkacz: do we really want to wait 20 minutes for exhibitor? What kind of software starts up 20 minutes? That's crazy...

Shouldn't we maybe give up earlier but fail with message that would clearly state that we failed waiting on exhibitor?

Btw. the usage of exhibitor in shakedown is kind of questionable too. Waiting 20 minutes for current number of masters... can't we get that information from somewhere else?

wavesoft: > do we really want to wait 20 minutes for exhibitor? What kind of software starts up 20 minutes? That's crazy...

**If** the exhibitor is responding. If not (ex. other cluster issues) and it takes 120 seconds to time-out, it's going to take 40 hours 😄

wavesoft: Therefore let's not include `TimeoutExpired` exception in the ignored ones. A connection error (in my opinion) should be considered critical and the test should fail.

tests/system/common.py
```diff
@@ -819,3 +822,26 @@ def kill_process_on_host(hostname, pattern):
     else:
         print("Killed no pids")
     return pids
+
+
+def service_available_predicate(service_name):
```
wavesoft: Can you also add 1-2 lines of documentation?

tests/system/common.py
```diff
@@ -819,3 +822,26 @@ def kill_process_on_host(hostname, pattern):
...
+# since exhibitor might take ~ 20 minutes to start up, we need to wait for it
+def wait_for_service_endpoint(service_name, timeout_sec=120):
+    """Checks the service url if available it returns true, on expiration
+    it returns false"""
+    @retrying.retry(wait_fixed=1000, stop_max_attempt_number=1200, retry_on_exception=ignore_exception)
+    def count_masters():
+        return len(get_all_masters())
+
+    master_count = count_masters()
+    return time_wait(lambda: service_available_predicate(service_name),
```
wavesoft: If I am right, `time_wait` will return the number of seconds it took until the `predicate` function returned true. If there was a timeout, a `TimeoutExpired` exception will be thrown. I am not 100% sure if this is still correct on the shakedown version that we are using. Have a look..

tests/system/common.py
meln1k: @alenkacz 
> do we really want to wait 20 minutes for exhibitor?

That's what I see in adminrouter logs. But the exhibitor starts up eventually.

> Shouldn't we maybe give up earlier 

Then we have to live with flaky tests. (and this is what we do right now).
> can't we get that information from somewhere else?

Since data about masters is stored in zookeeper, I'm not sure if we can avoid querying exhibitor.

@wavesoft 
> it's going to take 40 hours 😄

Not really since the http timeout is 5 seconds. But I'll adjust it to match 20 minutes, thanks
https://github.com/dcos/dcos-cli/blob/887f490b34ab6d8e572c189ef708e326a2d96df1/python/lib/dcos/dcos/http.py#L297

tests/system/common.py
meln1k: sure!

tests/system/common.py
alenkacz: I still doubt the 20 minute time.
- it sounds ridiculous that it boots up that long (in integration tests, so not under load or with a lots of data)
- next time we will see it boot upt for 40 minutes so we will increase?

The number just looks too long and kind of artificial (if it"s based on one log).

But I am open to counter-arguments :)

alenkacz:
Since GitHub is hiding my previous comment I'll open the topic here:

tests/system/common.py
```diff
@@ -819,3 +823,54 @@ def kill_process_on_host(hostname, pattern):
...
+        stop_max_attempt_number=240,  # waiting 20 minutes for exhibitor start-up
+        retry_on_exception=ignore_provided_exception(DCOSException))
+    def all_master_ips():
+        return get_all_master_ips()
+
+    master_private_ips = all_master_ips()
+
+    master_public_ips = []
+
+    for private_ip in master_private_ips:
```
zen-dog: nit: you could just `map` over the private IPs e.g.: `public_ips = list(map(shakedown.run_command(...)[1], private_ips)`

tests/system/common.py
```diff
@@ -819,3 +823,54 @@ def kill_process_on_host(hostname, pattern):
...
+    it returns false"""
+
+    @retrying.retry(
+            wait_fixed=1000,
+            stop_max_attempt_number=timeout_sec/5,  # underlying http.get has 5 seconds timeout, so we have to scale it
+            retry_on_exception=ignore_provided_exception(DCOSException))
+    def check_service_availability_on_master(master_ip, service):
+        url = "https://{}/service/{}/".format(master_ip, service)
+
+        # dcos.http doesn't correctly handle certs when we use ip as url, so we need a workaround:
```
zen-dog: meh

tests/system/common.py
meln1k: it's not a very python-ish style :)

jeschkies: Well, the following is more pythonic 🙂 

```python
master_public_ips = [shakedown.run_command(private_ip, '/opt/mesosphere/bin/detect_ip_public')[1]
                     for private_ip in master_private_ips]
```

this swallows `results` though.

Also, why not

```python
for private_ip in all_master_ips():
  ...
```

tests/system/common.py
```diff
@@ -819,3 +824,53 @@ def kill_process_on_host(hostname, pattern):
...
+    master_public_ips = []
+
+    for private_ip in master_private_ips:
+        result, ip = shakedown.run_command(private_ip, '/opt/mesosphere/bin/detect_ip_public')
+        master_public_ips.append(ip)
+
+    return master_public_ips
+
+
+# since exhibitor might take ~ 20 minutes to start up, we need to wait for it
```
jeschkies: It'd add this to the docs string.

tests/system/common.py
```diff
@@ -819,3 +824,53 @@ def kill_process_on_host(hostname, pattern):
...
+    @retrying.retry(
+            wait_fixed=1000,
+            stop_max_attempt_number=timeout_sec/5,  # underlying http.get has 5 seconds timeout, so we have to scale it
+            retry_on_exception=ignore_provided_exception(DCOSException))
+    def check_service_availability_on_master(master_ip, service):
+        url = "https://{}/service/{}/".format(master_ip, service)
+
+        toml_config = config.get_config()
+        auth_token = config.get_config_val("core.dcos_acs_token", toml_config)
+        auth = DCOSAcsAuth(auth_token)
```
jeschkies: If I'm not mistaken shakedown has this method `shakedown.dcos_acs_token()`. See https://github.com/mesosphere/marathon/blob/master/tests/system/fixtures/__init__.py#L59 for an example.

tests/system/common.py
```diff
@@ -819,3 +824,53 @@ def kill_process_on_host(hostname, pattern):
...
+        response = requests.request(
+            method='get',
+            url=url,
+            timeout=5,
+            auth=auth,
+            verify=str(get_ca_file()))
```
jeschkies: I think requests offers

```python
response = requests.get(url, timeout=5, auth=auth, verify=...)
if response.status_code != 200:
  raise DCOSException("service " + service_name + " is unavailable at " + master_ip)
```

jeschkies: Also, please check for the CA file. Or does requests make the check? Can `verify` be `None`?

PR #6300, opened 2018-06-15
Introduce "Live" performance testing

wavesoft:
---

~⚠️ This PR depends on #6291 so merge that one first~

---

This commit introduces the `test-continuous-n-apps.yml` test definition that uses the `SimplePolicy`
to run just a single test. It also uses the `DataDogMetricReporter` to report the metric
updates to DataDog in real-time.

Effectively, this means:
1. The test is a consecutive deployment of 5k apps
2. Every time a metric is measured, it's reported to DataDog under the key `marathon.perf_metrics.continuous.n-apps.*`
3. The test completes when 5k successful deployments  are received, or a single deployment failure occurs.

### DIFF
--- a/tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
+++ b/tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
@@ -32,7 +32,8 @@ config:
           name: p99
           percentile: 99
         - class: "@mean_err"
-
+        - class: "@min"
+        - class: "@max"
       units: sec
 
   # Define the indicators that can summarise our values
--- a/tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
+++ b/tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
@@ -15,7 +15,24 @@ config:
     - name: deploymentTime
       uuid: cfac77fceb244862aedd89066441c416
       desc: The time from the HTTP request completion till the deployment success
-      summarize: [mean_err]
+      summarize:
+        - class: summarize.PercentileSummarizer
+          name: p50
+          percentile: 50
+        - class: summarize.PercentileSummarizer
+          name: p75
+          percentile: 75
+        - class: summarize.PercentileSummarizer
+          name: p95
+          percentile: 95
+        - class: summarize.PercentileSummarizer
+          name: p98
+          percentile: 98
+        - class: summarize.PercentileSummarizer
+          name: p99
+          percentile: 99
+        - class: "@mean_err"
+
       units: sec
 
   # Define the indicators that can summarise our values
--- a/tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
+++ b/tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
@@ -15,7 +15,23 @@ config:
     - name: fullDeploymentTime
       uuid: cfac77fceb244862aedd89066441c416
       desc: The time from the first HTTP request completion till the last deployment success
-      summarize: [mean_err]
+      summarize:
+        - class: summarize.PercentileSummarizer
+          name: p50
+          percentile: 50
+        - class: summarize.PercentileSummarizer
+          name: p75
+          percentile: 75
+        - class: summarize.PercentileSummarizer
+          name: p95
+          percentile: 95
+        - class: summarize.PercentileSummarizer
+          name: p98
+          percentile: 98
+        - class: summarize.PercentileSummarizer
+          name: p99
+          percentile: 99
+        - class: "@mean_err"
       units: sec
 
   # Define the indicators that can summarise our values
--- a/tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
+++ b/tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
@@ -15,23 +15,7 @@ config:
     - name: fullDeploymentTime
       uuid: cfac77fceb244862aedd89066441c416
       desc: The time from the first HTTP request completion till the last deployment success
-      summarize:
-        - class: summarize.PercentileSummarizer
-          name: p50
-          percentile: 50
-        - class: summarize.PercentileSummarizer
-          name: p75
-          percentile: 75
-        - class: summarize.PercentileSummarizer
-          name: p95
-          percentile: 95
-        - class: summarize.PercentileSummarizer
-          name: p98
-          percentile: 98
-        - class: summarize.PercentileSummarizer
-          name: p99
-          percentile: 99
-        - class: "@mean_err"
+      summarize: [mean_err]
       units: sec
 
   # Define the indicators that can summarise our values
--- a/tests/performance/config/perf-driver/policies/test-1-apps-n-instances.yml
+++ b/tests/performance/config/perf-driver/policies/test-1-apps-n-instances.yml
@@ -85,6 +85,7 @@ channels:
             "id": "/scale-instances/{{uuid()}}",
             "backoffFactor": 1.0,
             "backoffSeconds": 0,
+            "portDefinitions": [],
             "labels": {
               "MARATHON_EXECUTOR_ID": "custom-executor"
             }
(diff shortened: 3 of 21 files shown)


tests/performance/config/perf-driver/policies/test-5k-apps.yml
```diff
@@ -0,0 +1,85 @@
...
+
+  # Use a simple policy that just tries a single case
+  - class: policy.SimplePolicy
+    parameters:
+      apps: 5000
+
+    # Complete the policy when 5000 deployment completion events are received
+    events:
+      start: MarathonStartedEvent
+      end: MarathonDeploymentSuccessEvent:nth(5000) MarathonDeploymentFailedEvent
```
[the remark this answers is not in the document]
wavesoft: Yup.

tests/performance/config/perf-driver/policies/test-1-apps-n-instances.yml
```diff
@@ -48,9 +48,7 @@ policies:
       - name: Scale Instances
         values:
           - parameter: instances
-            min: 50
-            step: 50
-            max: 500
+            values: [500, 1000, 2000]
```
zen-dog: This should be one value = `10.000`

zen-dog:
> Every time a metric is measured, it's reported to DataDog under the key marathon.perf_metrics.5k_apps.*

I'd also argue that we test parameters should not be directly represented in the metric name. E.g. we will go to 20k at some point and I don't want to fix the dashboards/alarms. So wdyt about `marathon.perf_metrics.continous.n-apps.*`?

tests/performance/config/perf-driver/policies/test-5k-apps.yml
```diff
@@ -0,0 +1,88 @@
...
+  test: 5k-apps
+
+# Test policy configuration
+# ===========================
+policies:
+
+  # Use a simple policy that just tries a single case
+  - class: policy.SimplePolicy
+    parameters:
+      apps: 5000
```
wavesoft: @zen-dog ^^ This is the test you should be looking at. The other two come from the fact that this PR is based on top of #6291

tests/performance/config/perf-driver/policies/test-1-apps-n-instances.yml
wavesoft: Check the `test-5k-apps.yml` test below

wavesoft:
> I'd also argue that we test parameters should not be directly represented in the metric name. E.g. we will go to 20k at some point and I don't want to fix the dashboards/alarms. So wdyt about marathon.perf_metrics.continous.n-apps.*?

Good point. Changed it to `continuous.n-apps`

tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
```diff
@@ -15,7 +15,24 @@ config:
     - name: deploymentTime
       uuid: cfac77fceb244862aedd89066441c416
       desc: The time from the HTTP request completion till the deployment success
-      summarize: [mean_err]
+      summarize:
```
jeschkies: As @meln1k pointed out it probably helps to have min and max as well.

tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
```diff
@@ -0,0 +1,88 @@
...
+  test: continuous.n-apps
+
+# Test policy configuration
+# ===========================
+policies:
+
+  # Use a simple policy that just tries a single case
+  - class: policy.SimplePolicy
+    parameters:
+      apps: 5000
```
jeschkies: Should this be 10,000?

tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
```diff
@@ -15,7 +15,23 @@ config:
     - name: fullDeploymentTime
       uuid: cfac77fceb244862aedd89066441c416
       desc: The time from the first HTTP request completion till the last deployment success
-      summarize: [mean_err]
+      summarize:
```
zen-dog: What are percentiles for the `fullDeploymentTime` (which is 1 number)?

tests/performance/config/perf-driver/test-continuous-n-apps.yml
```diff
@@ -0,0 +1,33 @@
...
+  - observers/observer-marathon-sse.yml
+  - observers/observer-dcluster-mesos-logs.yml
+
+  # Measure the given metrics
+  - measure/measure-deploymentTime.yml
+  - measure/measure-fullDeploymentTime.yml
+  - measure/measure-httpRequestTime.yml
+  - measure/measure-failedDeployments.yml
+  - measure/measure-groupsResponseTime.yml
+  - measure/measure-mesosOffers.yml
```
zen-dog: Shouldn't this list include CPU and Mem metrics?

tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
wavesoft: Good point 👍

tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
wavesoft: Pointless 😂 
Let me remove them...

wavesoft: Ah, by the way, since this is a re-usable fragment from all the tests, there are tests that _could_ have more than one `fullDeploymentTime` point (ex. when you repeat the tests). In that case I think it makes more sense?

tests/performance/config/perf-driver/test-continuous-n-apps.yml
wavesoft: Ideally yes. However they are optionally activated by including the `environments/opt-jmx.yml`. Here is a brief background:

1. I tried to abstract as much as possible on the tests, in order to make them able to run in different environments.
2. I ended up depending only on just the marathon URL. Which means you can target the same test suit against a DC/OS cluster, a DCluster or a stand-alone local-running marathon.
3. An exception to this rule is the system metrics. That's because we didn't have a proper `/metrics` endpoint and therefore I had to find a workaround.
4. The workaround was to use the `JMX` side-channel, and extract the current values from there. However introducing a second channel, breaks the ideal, single-endpoint abstraction. Especially since you cannot use the JMX endpoint when targeting a DC/OS cluster.
5. The solution was to extract the JMX-specific metrics and configuration to a separate "opt-in" configuration, that is included only on the specific cases where the JMX endpoint can be used.

tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
zen-dog: Hm. since we don't repeat tests right now, I'm in favour of removing it ;)

tests/performance/config/perf-driver/measure/measure-deploymentTime.yml
wavesoft: Yup, good idea 👍

tests/performance/config/perf-driver/measure/measure-fullDeploymentTime.yml
wavesoft: We do (it's the `n-apps-1-instances` and `1-apps-n-instances` tests). But I can keep only the mean on them.

tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
```diff
@@ -0,0 +1,88 @@
...
+
+  # Use a simple policy that just tries a single case
+  - class: policy.SimplePolicy
+    parameters:
+      apps: 10000
+
+    # Complete the policy when 5000 deployment completion events are received
+    events:
+      start: MarathonStartedEvent
+      end: MarathonDeploymentSuccessEvent:nth(5000,all) MarathonDeploymentFailedEvent:nth(5000,all)
```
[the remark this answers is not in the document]
wavesoft: Yup it should 😄

PR #6301, opened 2018-06-18
Reschedule resident task during a deployment

alenkacz:
Summary:
Current buggy flow:
- start deployment
- launch new resident task
- resident task have failing health check
- resident task is killed because of failing health check
- scale check is triggered but is ignored because we are in the middle of deployment
- `TaskStartActor` handles only `InstanceChanged` of terminal instances so not the reserved-terminal (note: right now we use Reserved for both Terminal and "Scheduled" cases)

Current change means:
- `TaskStartActor` will get `InstanceChanged` event for resident terminal tasks
- if `Sync` kicks in, it will also not consider terminal Reserved instances as active anymore

JIRA issues: DCOS-38317

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/deployment/impl/TaskStartActorTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/deployment/impl/TaskStartActorTest.scala
@@ -10,11 +10,15 @@ import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.condition.Condition.{Failed, Running}
 import mesosphere.marathon.core.event.{DeploymentStatus, _}
 import mesosphere.marathon.core.health.MesosCommandHealthCheck
-import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.core.instance.{Instance, TestInstanceBuilder}
 import mesosphere.marathon.core.launcher.impl.LaunchQueueTestHelper
 import mesosphere.marathon.core.launchqueue.LaunchQueue
 import mesosphere.marathon.core.readiness.ReadinessCheckExecutor
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.bus.MesosTaskStatusTestHelper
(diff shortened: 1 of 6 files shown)


src/test/scala/mesosphere/marathon/core/deployment/impl/TaskStartActorTest.scala
```diff
@@ -212,6 +216,31 @@ class TaskStartActorTest extends AkkaUnitTest with Eventually {
       expectTerminated(ref)
     }
   }
+
+  "relaunch when resident task gets terminal" in {
```
alenkacz: this test was failing prior to the fix...

zen-dog: When did this test fail last time?

ichernetsky: I think what Alena means is that this new test fails if the fix in this PR is not applied to the code.

zen-dog:
> We need to follow up with a SI test as well.

I disagree. SI tests are expensive and ideally, represent user feature level. The fix is relatively simple and a unit test to prevent regression seems adequate.

timcharper:
How about an integration test?

ichernetsky:
@zen-dog, it sound like Einstein's quote :)
It needs to be tested more thoroughly, not just by unit-tests, because PVs are one of the most hairy parts of Marathon, and one of the least used parts of Marathon too. On top of that, the cost of error is too high — potentially loosing data. If an integration test will do the trick, I am fine with it.

src/test/scala/mesosphere/marathon/core/deployment/impl/TaskStartActorTest.scala
alenkacz: Exactly :)

alenkacz:
Yeah, I actually already started working on integration test... These unit tests are so heavily mocked that I personally don't believe it fixes what I think it fixes.

PR #6303, opened 2018-06-19
 Pull in shakedown as an editable dependency. 

jeschkies:
Summary:
We have a lot of Shakedown changes pending in `tests/system/common.py`.
So far it takes a few steps to move the changes to Shakedown and then
pull them back again. This patch eases the process by importing
Shakedown as a git-subrepo. It allows making changes to Shakedown in the
Marathon repository *and* push changes upstream to create pull requests
for Shakedown. This way Shakedown in Marathon will be slightly ahead of
the Shakedown master *but* also ensure that the Shakedown master will
stay up-to-date.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -71,3 +71,26 @@ shakedown test_marathon_root.py::<test name>
 
 For more details checkout the [shakedown site](https://github.com/dcos/shakedown).
 The tests are written under the project [test/system](system/README.md).
+
+### Shakedown Patching
+
+Shakedown is imported as a `[git-subrepo](https://github.com/ingydotnet/git-subrepo#readme)`
+and declared as an editable dependency in the Pipfile. This allows any Marathon
+contributor to also make edits to Shakedown!
+
+If you made an edit simply commit it to the Marathon repository.
+
+```
+git subrepo push tests/shakedown -b foobar
+```
+
+will push all shakedown related commits to the `foobar` branch of the Shakedown
+[repository](https://github.com/dcos/shakedown).
+
+```
+git subrepo pull tests/shakedown -b master
+```
+
+will pull all changes on the Shakedown master as a suqashed commit into the
+Marathon repository. This will keep the Marathon commit history clean.
+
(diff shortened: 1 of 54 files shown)


jeschkies:
This is a proposal of mine. Please take a quick look at [git-subrepo](https://github.com/ingydotnet/git-subrepo#readme). It's less scary than it seems.

If you are using Shakedown in Marathon you have to do nothing except call the tests with Pipenv as the CI already does:

```
cd tests/system && make test
```

or 

```
cd tests/system
pipenv shell
shakedown test_marathon_root.py::<test name>
```

or

```
cd tests/system
pipenv run shakedown test_marathon_root.py::<test name>
```

If you want to make edits to Shakedown simply edit the files in `tests/shakedown` and commit them to the Marathon repo.

Once these commits landed or event before you can push the changes to a Shakedown branch with
```
git subrepo push tests/shakedown -b my-feature
```

I guess that only applies for Shakedown contributors.

If we want changes from Shakedown we can pull them with

```
git subrepo pull tests/shakedown -b master
```

`git-subrepo` will squash upstream patches in one update commit to keep Marathons commit history clean.

This way we get both benefits. A local Shakedown for fast iteration and upstream patches.

tests/README.md
```diff
@@ -71,3 +71,26 @@ shakedown test_marathon_root.py::<test name>
 
 For more details checkout the [shakedown site](https://github.com/dcos/shakedown).
 The tests are written under the project [test/system](system/README.md).
+
+### Shakedown Patching
```
jeschkies: This is the only edit I made.

tests/shakedown/.gitrepo
```diff
@@ -0,0 +1,11 @@
+; DO NOT EDIT (unless you know what you are doing)
```
kensipe: shouldn't this always be true :)

tests/system/marathon_common_tests.py
```diff
@@ -826,14 +826,14 @@ def test_app_update_rollback():
 
     client = marathon.create_client()
     client.add_app(app_def)
-    shakedown.deployment_wait()
+    shakedown.deployment_wait(app_id=app_id)
```
kensipe: I would not conflate these changes with the subrepo change

jeschkies: Yep. You are right. I won't merge this until https://github.com/mesosphere/marathon/pull/6302 landed.

PR #6305, opened 2018-06-20
Introduce `bytesWritten` counter for the events endpoint

wavesoft:
Summary:
This commit introduces a `bytesWritten` metric for the SSE endpoint that counts the number of bytes
sent. This is to complement the PR #6286. Since this is a write-only endpoint there is no need to
introduce a `bytesRead` metric.

JIRA issues: MARATHON-8264

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
@@ -2,12 +2,13 @@ package mesosphere.marathon
 package core.event.impl.stream
 
 import java.util.UUID
-import javax.servlet.http.{Cookie, HttpServletRequest, HttpServletResponse}
 
+import javax.servlet.http.{Cookie, HttpServletRequest, HttpServletResponse}
 import akka.actor.ActorRef
-import mesosphere.marathon.api.{RequestFacade, HttpTransferMetricsHandler}
+import mesosphere.marathon.api.{HttpTransferMetricsHandler, RequestFacade}
 import mesosphere.marathon.core.event.{EventConf, MarathonEvent}
 import mesosphere.marathon.core.event.impl.stream.HttpEventStreamActor._
+import mesosphere.marathon.metrics.{Metrics, ServiceMetric}
 import mesosphere.marathon.plugin.auth._
 import mesosphere.marathon.plugin.http.HttpResponse
 import org.eclipse.jetty.servlets.EventSource.Emitter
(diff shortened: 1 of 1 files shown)


wavesoft:
Sorryyy 😓

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -32,6 +34,27 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter, allowHea
...
+    * "
+    *
+    * @param eventName The name of the event, used for estimating the "name" field length
+    * @param payload The event payload, used for estimating the "data" field length
+    * @return Passes through the `payload` argument
+    */
+  private def measureFrameBytesSent(eventName: String, payload: String): String = {
+    val overhead: Long = 16 + eventName.length
+    bytesWrittenMetric.increment(payload.length + overhead)
+    payload
```
jeschkies: What is the motivation for returning the payload?

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -42,8 +65,10 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter, allowHea
 
   override def sendEvent(event: MarathonEvent): Unit = {
     if (subscribed(event.eventType)) {
-      if (useLightWeightEvents) blocking(emitter.event(event.eventType, event.lightJsonString))
-      else blocking(emitter.event(event.eventType, event.fullJsonString))
+      if (useLightWeightEvents) blocking(emitter.event(event.eventType,
+        measureFrameBytesSent(event.eventType, event.lightJsonString)))
```
jeschkies: I don't quite get the benefit of nesting the calls. Why not

```scala
val payload = if (useLightWeightEvents) event.lightJsonString else event.fullJsonString
measureFrameBytesSent(event.eventType, payload)
blocking(emitter.event(event.eventType, payload))
```

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -32,6 +34,26 @@ class HttpEventSSEHandle(request: HttpServletRequest, emitter: Emitter, allowHea
...
+    * "name: <name>\n
+    *  data: <payload>\n
+    *  \n
+    * "
+    *
+    * @param eventName The name of the event, used for estimating the "name" field length
+    * @param payload The event payload, used for estimating the "data" field length
+    * @return Passes through the `payload` argument
+    */
+  private def measureFrameBytesSent(eventName: String, payload: String): String = {
```
jeschkies: Nit: I don't think the return type is correct.

wavesoft: yup, fixed it, forgot to push 😄

PR #6306, opened 2018-06-20
Warn if any resource in an incoming offer is set to zero

meln1k:
JIRA issues: MARATHON-8283


timcharper:
This doesn’t warn for instances with no GPUs right?

meichstedt:
I agree with @timcharper – GPUs are a scarce resource by design, so naturally lots of offers will never provide any.

Also keep in mind that the default for disk in an app or pod definition is 0. Marathon will still be able to launch tasks if there's no disk offered and not explicitly requested so I'd argue that we should only warn for zero valued CPU and MEM offers.

PR #6307, opened 2018-06-20
Move Docker pull config factory to Shakedown.

jeschkies:
Summary:
This is the start of moving helper mehtods from `tests/system/common.py`
to Shakedown.


jeschkies:
> Even if it should - we should probably move the whole "shakedown section" from common.py at once and not test by test

That diff would too big IMHO. E.g. you probably would have missed your first comment.

jeschkies:
> This is a marathon-level feature, so I'm not sure it it should be in shakedown

It's rather a Marathon specific test conifg but I don't see why we should not expose it? Others might want to use it for the Marathon apps in tests.

jeschkies:
Dropping.

PR #6310, opened 2018-06-20
Import performance test Jenkins job definition.

jeschkies:
Summary:
The current Jenkins job definition that runs performance tests against master
and reports it to DataDog is not version controlled. This makes
iterating over the tests hard. This diff defines yet another Jenkinfiles
and imports the bash script verbatim. Once `ic/perf/5k-with-timeseries`
is merged into master we can define a Jenkins job or pipeline using the
definition of this diff and start iterating.

### DIFF
--- a/tests/performance/live.sh
+++ b/tests/performance/live.sh
@@ -33,8 +33,6 @@ docker run -i --rm \
     -e "PERF_DRIVER_ENVIRONMENT=env-ci-live.yml" \
     -e "DATADOG_API_KEY=$DATADOG_API_KEY" \
     -e "DATADOG_APP_KEY=$DATADOG_APP_KEY" \
-    -e "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID" \
-    -e "AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY" \
     -e DCLUSTER_ARGS="--docker_network='${DOCKER_NETWORK}' --marathon_jmx_host=marathon_1 --share_folder=${MARATHON_PERF_TESTING_DIR}/files" \
     icharalampidis/marathon-perf-testing:latest \
     ./tests/performance/ci_run_dcluster.sh \
--- a/tests/performance/live.sh
+++ b/tests/performance/live.sh
@@ -32,7 +32,6 @@ docker run -i --rm \
     -e "PARTIAL_TESTS=test-5k-apps" \
     -e "PERF_DRIVER_ENVIRONMENT=env-ci-live.yml" \
     -e "DATADOG_API_KEY=$DATADOG_API_KEY" \
-    -e "DATADOG_APP_KEY=$DATADOG_APP_KEY" \
     -e DCLUSTER_ARGS="--docker_network='${DOCKER_NETWORK}' --marathon_jmx_host=marathon_1 --share_folder=${MARATHON_PERF_TESTING_DIR}/files" \
     icharalampidis/marathon-perf-testing:latest \
     ./tests/performance/ci_run_dcluster.sh \
--- a/tests/performance/live.sh
+++ b/tests/performance/live.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -e
+
+MARATHON_DIR=$(pwd)
+MARATHON_PERF_TESTING_DIR=$(pwd)/marathon-perf-testing
+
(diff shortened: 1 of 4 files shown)


jeschkies:
Please compare the definition to https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/Marathon-Scale-Tests/job/nuc-perf-testing-live/.

jeschkies:
We probably should also define what live means.

wavesoft:
How about calling this "continuous" in comparison to "on-demand"?

tests/performance/live.sh
```diff
@@ -0,0 +1,42 @@
...
+# Run the CI job
+docker run -i --rm \
+    --network ${DOCKER_NETWORK} \
+    --privileged \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -v "$MARATHON_DIR:/marathon" \
+    -e "PARTIAL_TESTS=test-5k-apps" \
+    -e "PERF_DRIVER_ENVIRONMENT=env-ci-live.yml" \
+    -e "DATADOG_API_KEY=$DATADOG_API_KEY" \
+    -e "DATADOG_APP_KEY=$DATADOG_APP_KEY" \
```
wavesoft: You haven't defined `DATADOG_API_KEY` somewhere in your environment.
It's not secret and should be set to the `dcos-perf-test-driver` app from https://app.datadoghq.com/account/settings#api

tests/performance/live.sh
```diff
@@ -0,0 +1,42 @@
...
+    -e "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID" \
```
wavesoft: This particular test does not need to upload ton S3, so you can safely ignore these environment variables. However they must be non-empty for the bootstrap validation to pass. Do something like this in the beginning:

```
AWS_ACCESS_KEY_ID=unused
AWS_SECRET_ACCESS_KEY=unused
```

tests/performance/live.sh
jeschkies: Oh this app vs api key thing. Why do we need both? Our CI only needs the app one.

wavesoft: Well, DataDog needs both 😄 The one is the _API Access Key_ the other is the accounting key.

wavesoft: Ok, to be more precise, the default configuration of the library that I was using required both. Apparently you need the `API` key for submitting data and the `APP` key for reading data from DataDog.

Since the reporter only cares about sending data, you are right, we don't need an `APP` key. I am going to change the scripts on testing.

tests/performance/live.sh
```diff
@@ -0,0 +1,39 @@
...
+# Run the CI job
+docker run -i --rm \
+    --network ${DOCKER_NETWORK} \
+    --privileged \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -v "$MARATHON_DIR:/marathon" \
+    -e "PARTIAL_TESTS=test-5k-apps" \
+    -e "PERF_DRIVER_ENVIRONMENT=env-ci-live.yml" \
+    -e "DATADOG_API_KEY=$DATADOG_API_KEY" \
+    -e DCLUSTER_ARGS="--docker_network='${DOCKER_NETWORK}' --marathon_jmx_host=marathon_1 --share_folder=${MARATHON_PERF_TESTING_DIR}/files" \
```
wavesoft: If you want to remove the `DATAOG_APP_KEY` you should also change:

1. https://github.com/mesosphere/marathon/blob/master/tests/performance/scripts/dcluster_run.sh#L25
2. https://github.com/mesosphere/marathon/blob/master/tests/performance/scripts/dcluster_run.sh#L61
3. https://github.com/mesosphere/marathon/blob/master/tests/performance/config/perf-driver/reporters/report-datadog.yml#L18

Otherwise you will get an error while starting.
Otherwise, keep the environment variable, and set it to an arbitrary value.

jeschkies: Thanks for the pointers. I'm kind of surprised to see `live.sh` wrap `dcluster.sh`. Anyhow. I made the changes. I have to admit though that I don't understand why we wire the Python classes with YAML files instead of pure Python. What is the motivation?

jeschkies: I'm wondering if we could go for a pure Python config but would like to understand the added value of the YAML configs first 🙂.

PR #6313, opened 2018-06-23
Properly handle ENV args containing "*", " ", or line delimiters

timcharper:
Previously, we used a rather difficult-to-grok awk expression with
some bash looping to define the arguments. This proved to be
challenging and was error prone, causing arguments to be passed
incorrectly ("*" would pass an invalid globbed expansion, instead of a
literal "*")

JIRA Issues: MARATHON-8254

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -18,7 +18,7 @@ To handle the potential increase in concurrent connections, deployment operation
 
 ### Improved environment variable to command line argument mapping
 
-As part of the fix for [MARATHON-8254](https://jira.mesosphere.com/browse/MARATHON-8254), the logic for receiving command-line options from environment variables has been reworked. "*" is properly propagated (previously, the glob-expanded result was getting passed), and spaces and new-lines and now preserved.
+As part of the fix for [MARATHON-8254](https://jira.mesosphere.com/browse/MARATHON-8254), the logic for receiving command-line options from environment variables has been reworked. "*" is properly propagated (previously, the glob-expanded result was getting passed), and spaces and new-lines are now preserved.
 
 There's a small change in behavior for environments in which the launcher script is sources, rather than executed. Unexported environment variables will not be converted in to parameters.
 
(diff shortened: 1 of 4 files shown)


timcharper:
I have manually confirmed that the START_HOOK_SCRIPT logic continues to work.

changelog.md
```diff
@@ -16,6 +16,12 @@ Previously, when under substantial load, Marathon would time out a deployment in
 
 To handle the potential increase in concurrent connections, deployment operations and leader request proxying now use nonblocking I/O. The nonblocking I/O proxying logic may have some subtle differences in how responses are handled, including more aggressive rejection of malformed HTTP requests. In the off-chance that this causes an issue in your cluster, the old behavior can be restored with the command line flag `--deprecated_features=sync_proxy`. `sync_proxy` is scheduled to be removed in Marathon `1.8.0`.
 
+### Improved environment variable to command line argument mapping
+
+As part of the fix for [MARATHON-8254](https://jira.mesosphere.com/browse/MARATHON-8254), the logic for receiving command-line options from environment variables has been reworked. "*" is properly propagated (previously, the glob-expanded result was getting passed), and spaces and new-lines and now preserved.
```
alenkacz: `and now preserved` => `are now preserved`

changelog.md
```diff
@@ -16,6 +16,12 @@ Previously, when under substantial load, Marathon would time out a deployment in
...
+
+There's a small change in behavior for environments in which the launcher script is sources, rather than executed. Unexported environment variables will not be converted in to parameters.
```
alenkacz: `sources` =>`sourced`?

src/main/scala/mesosphere/marathon/Main.scala
```diff
@@ -229,8 +229,27 @@ class MarathonApp(args: Seq[String]) extends AutoCloseable with StrictLogging {
...
+    * Given environment variables starting with MARATHON_, convert to a series of arguments.
+    *
+    * If environment variable specifies an empty string, treat to boolean flag (no argument)
+    *
+    * @returns A list of args intended to be arg-parsed
+    */
+  def envToArgs(env: Map[String, String]): Seq[String] = {
+    env.flatMap {
+      case (k, v) if k.startsWith("MARATHON_") =>
+        val argName = s"--${k.drop(9).toLowerCase}"
```
zen-dog: I like this much more than previous `awk` script 👏

timcharper: Right? I spent 2 hours trying to fix the awk script and I then I thought, "why am I doing this lol?

PR #6314, opened 2018-06-24
Collect and archive Mesos sandboxes after failed SI tests

zen-dog:
Summary:
To ease debugging of flaky SI tests we tarball and download Mesos sandboxes after failed SI runs and archive them with the Jenkins build.

### DIFF
--- a/Jenkinsfile.strict.si
+++ b/Jenkinsfile.strict.si
@@ -36,7 +36,7 @@ node('py36') {
         } finally {
             junit allowEmptyResults: true, testResults: 'shakedown.xml'
             archive includes: "**/diagnostics.zip"
-            archive includes: "**/sandbox_.tar.gz"
+            archive includes: "**/sandbox_*.tar.gz"
         }
     }
   }
--- a/Jenkinsfile.strict.si
+++ b/Jenkinsfile.strict.si
@@ -36,6 +36,7 @@ node('py36') {
         } finally {
             junit allowEmptyResults: true, testResults: 'shakedown.xml'
             archive includes: "**/diagnostics.zip"
+            archive includes: "**/sandbox_.tar.gz"
         }
     }
   }
--- a/tests/system/fixtures/__init__.py
+++ b/tests/system/fixtures/__init__.py
@@ -99,6 +99,7 @@ def docker_ipv6_network_fixture():
 def archive_sandboxes():
     # Nothing to setup
     yield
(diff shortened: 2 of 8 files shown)


Jenkinsfile.strict.si
```diff
@@ -36,6 +36,7 @@ node('py36') {
         } finally {
             junit allowEmptyResults: true, testResults: 'shakedown.xml'
             archive includes: "**/diagnostics.zip"
+            archive includes: "**/sandbox_.tar.gz"
```
alenkacz: is this correct? I would expect something like `sandbox_*.tar.gz`

zen-dog: Good catch! That's a typo

tests/system/fixtures/__init__.py
```diff
@@ -93,3 +93,20 @@ def docker_ipv6_network_fixture():
...
+    # We tarball the sandboxes from all the agents first and download them afterwards
+    for agent in shakedown.get_private_agents():
+        file_name = 'sandbox_{}.tar.gz'.format(agent.replace(".", "_"))
+        cmd = 'sudo tar -zcvf {} /var/lib/mesos/slave'.format(file_name)
+        status, output = shakedown.run_command_on_agent(agent, cmd)  # NOQA
+
+        if status:
+            shakedown.copy_file_from_agent(agent, file_name)
+        else:
+            print('DEBUG: Failed to tarball the sandbox from the agent={}, output={}'.format(agent, output))
```
jeschkies: How about `logging.debug("Failed to tarball....")`?

tests/system/fixtures/__init__.py
```diff
@@ -93,3 +93,20 @@ def docker_ipv6_network_fixture():
     yield
     for agent in agents:
         shakedown.run_command_on_agent(agent, f"sudo docker network rm mesos-docker-ipv6-test")
+
+
+@pytest.fixture(autouse=True, scope='session')
+def archive_sandboxes():
```
jeschkies: This is really nice! I think we should add it to Shakedown, though. What do you think?

tests/system/fixtures/__init__.py
zen-dog: Do we have a `logger`? Last time I checked, it's `print` all over the place

jeschkies: We should move to logger at some point. I think Shakedown uses the logger.

PR #6320, opened 2018-06-27
Define instance goals.

jeschkies:
Summary:
This introduces instances goals. For now they have no effects. However, it is the basis for further
improvements. Please note that we omit the goal in the instance state format. It will be added
with the goal migration logic.

JIRA issues: MARATHON-8244

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Goal.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Goal.scala
@@ -1,28 +1,54 @@
-package mesosphere.marathon.core.instance
+package mesosphere.marathon
+package core.instance
+
+import play.api.libs.json.{Format, JsError, JsResult, JsString, JsSuccess, JsValue, JsonValidationError, Reads, Writes}
 
 /**
   * Defines goal of the instance.
   * Goal is set by an orchestration layer and interpreted by scheduler layer.
   * In the end it is used by low-level scheduler to make scheduling decisions e.g. should the task associated with this instance be launched or killed?
   */
-sealed trait Goal
+sealed trait Goal extends Product with Serializable
 
-/**
-  * There should always be a running Mesos task associated by instance in this state.
-  * Instance with Suspended Goal might be changed to both [[Running]] or [[Suspended]] by orchestration layer.
-  */
-object Running extends Goal
+object Goal {
 
-/**
-  * Tasks associated with this instance shall be killed, but the instance needs to be kept in the state.
-  * This is typically needed for resident instances, where we need to persist the reservation for re-launch.
-  * Instance with Suspended Goal might be changed to both [[Running]] or [[Decommissioned]].
-  */
-object Suspended extends Goal
+  /**
+    * There should always be a running Mesos task associated by instance in this state.
+    * Instance with Suspended Goal might be changed to both [[Running]] or [[Stopped]] by orchestration layer.
+    */
+  case object Running extends Goal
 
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/core/instance/Goal.scala
```diff
@@ -0,0 +1,54 @@
...
+  * Defines goal of the instance.
+  * Goal is set by an orchestration layer and interpreted by scheduler layer.
+  * In the end it is used by low-level scheduler to make scheduling decisions e.g. should the task associated with this instance be launched or killed?
+  */
+sealed trait Goal extends Product with Serializable
+
+object Goal {
+
+  /**
+    * There should always be a running Mesos task associated by instance in this state.
```
meichstedt: _**Expresses the intent that** there should always be a running Mesos task associated by instance in this state._?

The way it's written atm could be interpreted like it is considered a fact that there's always a task, which is not the case.

src/main/scala/mesosphere/marathon/core/instance/Goal.scala
```diff
@@ -0,0 +1,54 @@
...
+    * Instance with Suspended Goal might be changed to both [[Running]] or [[Stopped]] by orchestration layer.
+    */
+  case object Running extends Goal
+
+  /**
+    * Tasks associated with this instance shall be killed, but the instance needs to be kept in the state.
```
meichstedt: This could follow the above suggestion:

_Expresses the intent that tasks associated with this instance shall be killed, but the instance needs to be kept in the state._

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -385,7 +385,29 @@ object Instance {
   }
 
   implicit val instanceConditionFormat: Format[Condition] = Condition.conditionFormat
-  implicit val instanceStateFormat: Format[InstanceState] = Json.format[InstanceState]
+
+  implicit val instanceStateWrites: Writes[InstanceState] = {
```
meichstedt: Could you add a comment here that we explicitly drop the `goal` here until xyz is merged?

jeschkies: The follow up with migrations is already up. I think that we'll merge both at the same time. So I don't think we need the comment.

src/main/scala/mesosphere/marathon/core/instance/Goal.scala
```diff
@@ -0,0 +1,55 @@
...
+          case v => JsError(JsonValidationError("instance.state.goal", s"Unknown goal $v - expecting string with one of the following values 'running', 'stopped', 'decommissioned'"))
+        }
+        case v => JsError(JsonValidationError("instance.state.goal", s"Unknown goal $v - expecting string with one of the following values 'running', 'stopped', 'decommissioned'"))
+      }
+    }
+  }
+
+  implicit val goalFormat = Format[Goal](
+    goalReader,
+    Writes(goal => JsString(goal.toString)))
```
meichstedt: Do we have any tests covering ser/deser? Renaming a goal would let this break w/o a test failure from what I can see.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
meichstedt: SGTM

PR #6321, opened 2018-06-27
Use unique names for system test apps.

jeschkies:
Summary:
This should ease debugging test failures as we can identify the apps on
the Mesos agents. The same approach helped tackle flaky integration
tests. See 92a81025745a43e2883231cd2ff50b4ac27ff6da for details.

### DIFF
--- a/tests/system/apps/__init__.py
+++ b/tests/system/apps/__init__.py
@@ -1,5 +1,4 @@
 import os.path
-import uuid
 
 from utils import make_id, get_resource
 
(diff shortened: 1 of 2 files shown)


tests/system/apps/__init__.py
```diff
@@ -15,24 +15,39 @@ def load_app(app_name):
...
+    return app_def
 
 
 def healthcheck_and_volume():
     return load_app('healthcheck-and-volume')
 
 
-def ucr_docker_http_server():
-    return load_app('ucr-docker-http-server')
+def ucr_docker_http_server(app_id=None):
```
zen-dog: nit: `s/app_id/app_postfix`

tests/system/apps/__init__.py
```diff
@@ -15,24 +15,39 @@ def load_app(app_name):
     return app
 
 
-def mesos_app():
-    return load_app('mesos-app')
+def mesos_app(app_id=None):
+    if app_id is None:
```
ichernetsky: This app ID generation is already done in the same way by `load_app` which in turn calls `make_id`. Should it be like
```
if app_id_prefix is None: app_id_prefix = ...
return load_app(app_id_prefix)
```
?

tests/system/apps/__init__.py
ichernetsky: It is a prefix, not a suffix.

jeschkies: It's neither. Only if `app_id` is not set we use `/ucr-docker-http-server` as a prefix.

tests/system/apps/__init__.py
jeschkies: The app name is also the name of the JSON file but you are right. I'll refactor this.

jeschkies:
System tests are green as well!

PR #6326, opened 2018-06-28
 Migrate instances to include a goal. 

jeschkies:
Summary:
With MARATHON-8244 we introduced instance goals. New logic will depend on these goals.
This patch introduces a migration for instances in storage version 1.6.0 to set their goals
accordingly. As  MARATHON-8246 mentions only terminal resident instances are special and
might have the goal `Stopped`.

JIRA issues: MARATHON-8246

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -386,14 +386,17 @@ object Instance {
 
   implicit val instanceConditionFormat: Format[Condition] = Condition.conditionFormat
 
+  import Goal.goalFormat
+
   implicit val instanceStateWrites: Writes[InstanceState] = {
     (
       (__ \ "condition").write[Condition] ~
       (__ \ "since").write[Timestamp] ~
       (__ \ "activeSince").writeNullable[Timestamp] ~
-      (__ \ "healthy").writeNullable[Boolean]
+      (__ \ "healthy").writeNullable[Boolean] ~
+      (__ \ "goal").write[Goal]
     ) { (instanceState) =>
-        (instanceState.condition, instanceState.since, instanceState.activeSince, instanceState.healthy)
+        (instanceState.condition, instanceState.since, instanceState.activeSince, instanceState.healthy, instanceState.goal)
       }
   }
 
@@ -402,9 +405,10 @@ object Instance {
       (__ \ "condition").read[Condition] ~
       (__ \ "since").read[Timestamp] ~
       (__ \ "activeSince").readNullable[Timestamp] ~
-      (__ \ "healthy").readNullable[Boolean]
-    ) { (condition, since, activeSince, healthy) =>
-        InstanceState(condition, since, activeSince, healthy)
+      (__ \ "healthy").readNullable[Boolean] ~
+      (__ \ "goal").read[Goal]
+    ) { (condition, since, activeSince, healthy, goal) =>
+        InstanceState(condition, since, activeSince, healthy, goal)
       }
   }
 
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -388,29 +388,7 @@ object Instance {
 
   import Goal.goalFormat
 
-  implicit val instanceStateWrites: Writes[InstanceState] = {
-    (
-      (__ \ "condition").write[Condition] ~
-      (__ \ "since").write[Timestamp] ~
-      (__ \ "activeSince").writeNullable[Timestamp] ~
-      (__ \ "healthy").writeNullable[Boolean] ~
-      (__ \ "goal").write[Goal]
-    ) { (instanceState) =>
-        (instanceState.condition, instanceState.since, instanceState.activeSince, instanceState.healthy, instanceState.goal)
-      }
-  }
-
-  implicit val instanceStateReads: Reads[InstanceState] = {
-    (
-      (__ \ "condition").read[Condition] ~
-      (__ \ "since").read[Timestamp] ~
-      (__ \ "activeSince").readNullable[Timestamp] ~
-      (__ \ "healthy").readNullable[Boolean] ~
-      (__ \ "goal").read[Goal]
-    ) { (condition, since, activeSince, healthy, goal) =>
-        InstanceState(condition, since, activeSince, healthy, goal)
-      }
-  }
+  implicit val instanceStateFormat: Format[InstanceState] = Json.format[InstanceState]
 
   implicit val reservationFormat: Format[Reservation] = Reservation.reservationFormat
 
--- a/src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
+++ b/src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
@@ -222,9 +222,9 @@ object Migration {
       },
       StorageVersions(1, 6, 0, StorageVersion.StorageFormat.PERSISTENCE_STORE) -> { (migration) =>
         new MigrationTo160(migration.instanceRepo, migration.persistenceStore)
-      }
-    // From here onwards we are not bound to the build version anymore.
-    //StorageVersions(200) -> { (migration) => new MigrationTo200(...) }
+      },
+      // From here onwards we are not bound to the build version anymore.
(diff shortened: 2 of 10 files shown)


jeschkies:
The logs of the `UpgradeIntegrationTest`

```
    104 DEBUG[15:19:52 UpgradeIntegrationTest-current-LocalMarathon-32893] INFO [15:19:52 Migration] Migration for storage: Version(1, 6, 0, PERSISTENCE_STORE) to target: Version(200, 0, 0, PERSISTENCE_STORE): apply change for version: Version(200, 0, 0, PERSISTENCE_STORE)
    105 DEBUG[15:19:52 UpgradeIntegrationTest-current-LocalMarathon-32893] INFO [15:19:52 MigrationTo200$] Starting reservations migration to Storage version 200
```

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -222,9 +222,9 @@ object Migration {
       },
       StorageVersions(1, 6, 0, StorageVersion.StorageFormat.PERSISTENCE_STORE) -> { (migration) =>
         new MigrationTo160(migration.instanceRepo, migration.persistenceStore)
-      }
-    // From here onwards we are not bound to the build version anymore.
-    //StorageVersions(200) -> { (migration) => new MigrationTo200(...) }
+      },
+      // From here onwards we are not bound to the build version anymore.
+      StorageVersions(200) -> { (migration) => new MigrationTo200(migration.instanceRepo, migration.persistenceStore) }
```
ichernetsky: Do we assign a range to each version then? Like `200` to 1.7.x, `300` to `1.8.x` and so forth? I am asking because it is unclear to me how we handle situations like the one for which we introduced `MigrationTo152`?

jeschkies: The storage version is independent of the release version. However, we should either leave room to sneak in migrations or use the minor version for that.

So either

```
StorageVersion(200)
StorageVersion(201)
StorageVersion(250)
```

or
```
StorageVersion(200)
StorageVersion(200, 1)
StorageVersion(201)
```

jeschkies: I like the first one better.

ichernetsky: Probably I am missing something. I don't get what benefit of fully decoupling storage versions from Marathon versions there is. What about having them like `1.7.x` where the first two numbers are coupled to the corresponding Marathon version, and `x` is just an incrementing number starting from `0`?

jeschkies: The Storage version should be bound to a commit not a Marathon version. Especially with our new version scheme when we do not know the patch version the old versioning is hard.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo200.scala
```diff
@@ -0,0 +1,151 @@
...
+      case ZkSerialized(byteString) =>
+        Json.parse(byteString.utf8String)
+    }
+
+  /**
+    * This function traverses all instances in ZK and sets the instance goal field.
+    */
+  def migrateInstanceGoals(instanceRepository: InstanceRepository, persistenceStore: PersistenceStore[_, _, _])(implicit mat: Materializer): Future[Done] = {
+
+    logger.info("Starting reservations migration to Storage version 200")
```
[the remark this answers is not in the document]
jeschkies: Damn. I thought I had that. Thanks.

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
jeschkies: #5957 has more details.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -386,14 +386,17 @@ object Instance {
...
+  import Goal.goalFormat
+
   implicit val instanceStateWrites: Writes[InstanceState] = {
     (
       (__ \ "condition").write[Condition] ~
       (__ \ "since").write[Timestamp] ~
       (__ \ "activeSince").writeNullable[Timestamp] ~
-      (__ \ "healthy").writeNullable[Boolean]
+      (__ \ "healthy").writeNullable[Boolean] ~
+      (__ \ "goal").write[Goal]
```
meichstedt: Why don't we go back to
```
implicit val instanceStateFormat: Format[InstanceState] = Json.format[InstanceState]
```
?

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo200.scala
```diff
@@ -0,0 +1,151 @@
...
+    * Update the goal of the instance.
+    * @param instance The old instance.
+    * @return An instance with an updated goal.
+    */
+  def updateGoal(instance: Instance): Instance = {
+    val updatedInstanceState = if (!instance.hasReservation) {
+      instance.state.copy(goal = Goal.Running)
+    } else {
+      if (instance.isReservedTerminal) {
+        instance.state.copy(goal = Goal.Stopped)
```
meichstedt: I'd like to understand this better. A `reservedTerminal` instance is one with condition `Reserved` and a task in a terminal state, correct? We might want to launch a new task for this instance, which previously was implicit, and bound to the instance count specified in the run spec.

Is the idea here that we'll migrate to `Stopped`, and recovery logic will eventually set this to `Running` after failover, once it realizes we're under capacity?

jeschkies: If I understood you correctly, yes. `isReservedTerminal` means the tasks are not running but do have a reservation. The can be started again the goal will be `Running` again. See https://github.com/mesosphere/marathon/pull/6328/files#diff-4d03c99a6694604a192ac4812b53a108R126.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: Good point.

jeschkies: Done ✅

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
ichernetsky: I still don't get the benefit of allocating storage version ranges to Marathon `major.minor` versions. I get why it is impossible to tie a storage `patch` version to a Marathon release one. But allocating `1.7.x` storage versions (where `x` is incremented manually), to `1.7.y` Marathon builds/versions (where `y` is the number of commits since a specific point in the commit history) has that benefit that it is clear right on in which `major.minor` version a particular storage migration is introduced. What do you think?

meichstedt: I think I can understand both sides here. We're trying to decouple `major.minor.patch` from storage versions, but there's no strong reason why it needs to start with `2` or couldn't start with `1.7`.

Sure we can start with`StorageVersion(1, 7, 0)` and add `StorageVersion(1, 7, 1)` which might then be added in build artifact `1.7.120`. I can imagine there would be confusion wrt `StorageVersion(1, 7, 120)` and build artifact `1.7.120`, which are most likely not related at all.

From a storage migration perspective, the only thing that matters is that the values are incremental, unique, and ordered according to how they should be applied.

@ichernetsky from what I understand the only benefit from using `200` is to obviously disambiguate the version patch value from the migration patch value. A compromise could be to use `StorageVersion(17, 0)`, add additional migrations for 1.7 via `StorageVersion(17, 1)` etc, and move to `StorageVersion(18, 0)` for v1.8.

It's _probably_ less confusing than going to 200, and less confusing than having that storage version/built artifact mismatch.

@jeschkies WDYT?

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo200.scala
meichstedt: Ok, so the `Stopped` doesn't really have any effect atm. Got it.

src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
jeschkies: > @jeschkies WDYT?

In principle that's fine but what if we want to backport a migration? Suddenly we have a `StroageVersion(17)` on Marathon 1.6.

PR #6330, opened 2018-07-02
Update dcos-launch config to use onprem

cprovencher:
attempt to fix ~10% cluster deployment failure rate

### DIFF
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -48,7 +48,7 @@ num_public_agents: 1
 num_private_agents: 3
 num_masters: 3
 dcos_config:
-  cluster_name: marathon-cluster
+  cluster_name: $DEPLOYMENT_NAME
   resolvers:
     - 8.8.4.4
     - 8.8.8.8
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -49,14 +49,14 @@ num_public_agents: 1
 num_private_agents: 3
 num_masters: 3
 dcos_config:
-  cluster_name: $DEPLOYMENT_NAME
-  resolvers:
-    - 8.8.4.4
-    - 8.8.8.8
-  dns_search: mesos
-  master_discovery: static
-  exhibitor_storage_backend: static
-  rexray_config_preset: aws
+    cluster_name: $DEPLOYMENT_NAME
+    resolvers:
+        - 8.8.4.4
+        - 8.8.8.8
+    dns_search: mesos
+    master_discovery: static
+    exhibitor_storage_backend: static
+    rexray_config_preset: aws
 EOF
 
 # Append license if one is available.
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -43,7 +43,7 @@ provider: onprem
 platform: aws
 aws_region: us-west-2
 key_helper: true
-instance_type: m5.large
+instance_type: m4.large
 num_public_agents: 1
 num_private_agents: 3
 num_masters: 3
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -43,7 +43,6 @@ provider: onprem
 platform: aws
 aws_region: us-west-2
 key_helper: true
-ssh_user: alovelace
 instance_type: m4.large
 num_public_agents: 1
 num_private_agents: 3
--- a/ci/launch_cluster.sh
+++ b/ci/launch_cluster.sh
@@ -28,7 +28,7 @@ INFO_PATH="$DEPLOYMENT_NAME.info.json"
 if [ "$VARIANT" == "open" ]; then
   INSTALLER="https://downloads.dcos.io/dcos/${CHANNEL}/dcos_generate_config.sh"
 else
-  INSTALLER="https://downloads.mesosphere.com/dcos-enterprise-aws-advanced/${CHANNEL}/${VARIANT}/dcos_generate_config.sh"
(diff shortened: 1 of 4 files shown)


ci/launch_cluster.sh
```diff
@@ -38,16 +38,24 @@ echo "Using: ${TEMPLATE}"
 envsubst <<EOF > "$CONFIG_PATH"
 ---
 launch_config_version: 1
-template_url: $TEMPLATE
```
jeschkies: Where did this go? We use two specific DC/OS pull requests:
* https://github.com/dcos/dcos/pull/1739 for DC/OS with latest Marathon
* https://github.com/mesosphere/dcos-enterprise/pull/1296 for DC/OS EE with latest Marathon

ci/launch_cluster.sh
```diff
@@ -38,16 +38,24 @@ echo "Using: ${TEMPLATE}"
...
 deployment_name: $DEPLOYMENT_NAME
-provider: aws
+installer_url: https://downloads.dcos.io/dcos/testing/master/dcos_generate_config.sh
```
jeschkies: Should this just be `https://downloads.dcos.io/dcos/$CHANNEL/dcos_generate_config.sh`?

jeschkies:
Hm, the cluster launch fails https://jenkins.mesosphere.com/service/jenkins/job/system-integration-tests/job/marathon-si-pr/192/ :/

ci/launch_cluster.sh
```diff
@@ -38,16 +38,24 @@ echo "Using: ${TEMPLATE}"
...
 key_helper: true
-template_parameters:
-    DefaultInstanceType: m4.large
-    AdminLocation: 0.0.0.0/0
-    PublicSlaveInstanceCount: 1
-    SlaveInstanceCount: 3
+instance_type: m4.large
+num_public_agents: 1
+num_private_agents: 3
+num_masters: 1
```
jeschkies: This is also a multi master setup so we need 3.

ci/launch_cluster.sh
MrMarvin: Yep, should be something like
```
https://downloads.mesosphere.com/dcos-enterprise/testing/pull/1739/dcos_generate_config.ee.sh
```
where `testing/pull/1739` is your `$CHANNEL` var.

zen-dog: yep, done

zen-dog:
@cprovencher I've updated your branch so don't forget to pull

cprovencher:
This same config was working when using dcos-launch in a different AWS account. AWS stack cannot come up, Will require more investigation.

ci/launch_cluster.sh
```diff
@@ -26,28 +26,35 @@ CONFIG_PATH="$DEPLOYMENT_NAME.yaml"
 INFO_PATH="$DEPLOYMENT_NAME.info.json"
 
 if [ "$VARIANT" == "open" ]; then
-  TEMPLATE="https://s3.amazonaws.com/downloads.dcos.io/dcos/${CHANNEL}/cloudformation/multi-master.cloudformation.json"
+  INSTALLER="https://downloads.mesosphere.com/dcos/${CHANNEL}/dcos_generate_config.sh"
```
ichernetsky: Why using CDN and not just S3 directly?

zen-dog: Recommendations of the SRE team? ¯\_(ツ)_/¯

zen-dog:
> This same config was working when using dcos-launch in a different AWS account. AWS stack cannot come up, Will require more investigation.

Here is an example of the `CREATE_FAILED` error: https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-pr/196/console

MrMarvin:
Hey Folks, it seems there has been some confusion about which URL to use. Also I made a mistake by suggesting an enterprise URL for the open-DC/OS case earlier, apologies!

Below are valid URL templates for both cases:
```
  INSTALLER="https://downloads.dcos.io/dcos/${CHANNEL}/dcos_generate_config.sh"
else
  INSTALLER="https://downloads.mesosphere.com/dcos-enterprise/${CHANNEL}/dcos_generate_config.ee.sh"
fi
```

Please also always try to use the CDN URLs, as they allow us to abstract from individual bucket names, which will likely change at some point.

cprovencher:
@zen-dog I believe I have found the culprit: m5.Large instance type. I don't think those are supported by the current templates we have in dcos-launch. m4 instances work fine. The easy solution is to switch to m4 instance types, but if you really need m5 we'll have to examine the templates and add support for m5 instances https://github.com/dcos/dcos-launch/tree/master/dcos_launch/templates

jeschkies:
Ok, the cluster started at least. However, we don't have the ssh key.

jeschkies:
Only `test_marathon_root.test_external_volume` is failing. It uses rexray. Do we know of any rexray limitations in CentOS?

jeschkies:
Strict hat two failures in two runs
- `test_marathon_root.test_launch_mesos_container_with_docker_image`
- `test_marathon_root.test_private_repository_docker_app`

Both seem falkes to me.

ci/launch_cluster.sh
```diff
@@ -26,33 +26,43 @@ CONFIG_PATH="$DEPLOYMENT_NAME.yaml"
 INFO_PATH="$DEPLOYMENT_NAME.info.json"
```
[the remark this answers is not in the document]
jeschkies: The iterface to `ci/launch_cluster.sh` did not change. So it should be fine...in theory 😉.

PR #6331, opened 2018-07-02
Properly handle ENV args containing "*", " ", or line delimiters

timcharper:
Backport of 49647f8

Previously, we used a rather difficult-to-grok awk expression with
some bash looping to define the arguments. This proved to be
challenging and was error prone, causing arguments to be passed
incorrectly ("*" would pass an invalid globbed expansion, instead of a
literal "*")

JIRA Issues: MARATHON-8254

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,11 @@
+## Changes since 1.5.9
+
+### Improved environment variable to command line argument mapping
+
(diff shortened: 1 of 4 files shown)


timcharper:
I have manually confirmed that the HOOK_MARATHON_START logic continues to work.

timcharper:
I have manually confirmed the acceptance criteria specified in https://jira.mesosphere.com/browse/MARATHON-8254 (asterisks work properly, HOOK_MARATHON_START works)

PR #6333, opened 2018-07-03
Update dcos-launch to master.

jeschkies:
Summary:
In the hope that a recent bug awas fixed we are upgrading dcos-launch to
its master version.


jeschkies:
Premissive and open pass.

PR #6337, opened 2018-07-04
Log incoming offers at info level

meln1k:
Summary:
It should be done in order to ease troubleshooting when Mesos and Marathon are under load, and it takes noticeable time to receive offers.

JIRA issues: MARATHON-8282

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
@@ -10,8 +10,6 @@ import mesosphere.marathon.core.matcher.base.OfferMatcher.{InstanceOpWithSource,
 import mesosphere.marathon.core.task.tracker.InstanceTracker
 import mesosphere.marathon.metrics.{Metrics, ServiceMetric}
 import org.apache.mesos.Protos.{Offer, OfferID}
-import mesosphere.marathon.raml.RamlConversions.offerWrites
-import play.api.libs.json.Json
 
 import scala.concurrent.Future
 import scala.util.control.NonFatal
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImpl.scala
```diff
@@ -81,10 +79,9 @@ private[launcher] class OfferProcessorImpl(
   }
 
   private def logOffer(offer: Offer): Unit = {
-    val raml = offerWrites(offer)
-    val json = Json.toJson(raml)
-    val compact = json.toString()
-    logger.info(s"Processing offer: $compact")
+    val offerId = offer.getId.getValue
+    val agentId = offer.getSlaveId.getValue
+    logger.info(s"Processing offer: offerId $offerId, agentId $agentId")
```
ichernetsky: Why this change? Why not printing entire offers?

Also, could you please log and warn before doing anything with an offer? Meaning this
```
     incomingOffersMeter.increment()
     offerStreamInput.offer(offer)
     warnOnZeroResource(offer)
     logOffer(offer)
```

should be replaced with

```
     incomingOffersMeter.increment()
     logOffer(offer)
     warnOnZeroResource(offer)
     offerStreamInput.offer(offer)
```

timcharper: Why offer summary? See this part of the discussion on the related issue:

screen shot 2018-07-05 at 2 16 38 pm

Perhaps if debug logging is on we can print the whole offer. I think it is excessive for INFO level.

ichernetsky: I think it would be excessive if we were to print offers on multiple lines. In the screenshot you posted, I support printing them as one-liners compactly formatted. This is what you referred to as a brief summary as well.

PR #6338, opened 2018-07-04
Fix for a wrong network update link

meln1k:


wavesoft:
That looks good to merge. Is there any reason we are holding this merge? @meln1k ?

PR #6339, opened 2018-07-04
ElectionServiceTest flakyness fix

meln1k:
Summary: made method calls synchronous to remove a race condition

JIRA issues: MARATHON-8302


timcharper:
We should not need to call `.futureValue` on the offer method for `SourceQueue`; ordering guarantees are preserved for invocations of that method.

PR #6342, opened 2018-07-05
Use parallel requests on the continuous-n-apps test

wavesoft:
In order to speed-up the deployment of 10k apps, this PR changes the number of parallel requests to `100`

### DIFF
--- a/tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
+++ b/tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
@@ -49,7 +49,7 @@ policies:
     # Complete the policy when 10000 deployment completion events are received
     events:
       start: MarathonStartedEvent
-      end: MarathonDeploymentSuccessEvent:nth(10000,all) MarathonDeploymentFailedEvent:nth(10000,all)
+      end: MarathonDeploymentSuccessEvent:nth(10000,all) MarathonDeploymentFailedEvent:nth(10000,all) MarathonDeploymentRequestFailedEvent(10000,all)
 
     # Don't let the tests run for more than 1 hour.
     timeout: 60m
@@ -66,6 +66,7 @@ channels:
     deploy:
       - type: app
         repeat: "{{apps}}"
+        parallel: 500
         spec: |
           {
             "cmd": "sleep infinity",
--- a/tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
+++ b/tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
@@ -66,7 +66,7 @@ channels:
     deploy:
       - type: app
         repeat: "{{apps}}"
-        parallel: 500
+        parallel: 100
         spec: |
           {
             "cmd": "sleep infinity",


tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
```diff
@@ -49,7 +49,7 @@ policies:
     # Complete the policy when 10000 deployment completion events are received
     events:
       start: MarathonStartedEvent
-      end: MarathonDeploymentSuccessEvent:nth(10000,all) MarathonDeploymentFailedEvent:nth(10000,all)
+      end: MarathonDeploymentSuccessEvent:nth(10000,all) MarathonDeploymentFailedEvent:nth(10000,all) MarathonDeploymentRequestFailedEvent(10000,all)
```
zen-dog: So we end if `all` the request fail?

tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
```diff
@@ -66,6 +66,7 @@ channels:
     deploy:
       - type: app
         repeat: "{{apps}}"
+        parallel: 100
```
zen-dog: Did you try it? Does it produce better run time?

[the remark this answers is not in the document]
wavesoft: Oh yes. It's was quite complicated, especially since we also need a retry and a back-off behaviour in _addition_ to the parallel requests.

Everything is in this stand-alone module : https://github.com/mesosphere/dcos-perf-test-driver/blob/master/performance/driver/classes/channel/utils/bulk.py

tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
wavesoft: Nope, this is a grouping operator. It means that it will account overall `10000` events, either `MarathonDeploymentSuccessEvent`, or `MarathonDeploymentFailedEvent`, or `MarathonDeploymentRequestFailedEvent`.

This particular part is a bit boy-scouting, since I noticed that a job never finished because it was failing to place the request to marathon (aka `MarathonDeploymentRequestFailedEvent`) instead of marathon not being able to satisfy the request (aka `MarathonDeploymentFailedEvent`)

tests/performance/config/perf-driver/policies/test-continuous-n-apps.yml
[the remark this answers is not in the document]
jeschkies: Just out of curiosity, why did you not use Python's [multiprocessing module](https://docs.python.org/3/library/multiprocessing.html) instead of a custom executor service? Multi processing is incredibly hard in Python. It might even make sense to start a [drill](https://github.com/fcsonline/drill) benchmark from Python.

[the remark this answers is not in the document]
wavesoft: Generally I avoid multiprocessing when I don't have CPU-bound problems. Mainly because I instinctively think that in such cases the IPC cost is bigger than the operation itself. In particularly, the driver internally dispatches thousands of messages per second, and broadcasting them to multiple processes will have some decent impact. 

In our case, the problem is I/O bound so we won't really benefit from multiprocessing. And since we are not stress-testing the endpoint, but we are actually waiting for a marathon operation to complete (which is much longer than the actual HTTP operation), we are more than safe with the current implementation.

As a matter of fact, since you mentioned this earlier, I have tried to switch the core thread pool (implemented in the [Event Bus](https://github.com/mesosphere/dcos-perf-test-driver/blob/master/performance/driver/core/eventbus.py#L26) ), from thread-based to process-based, since the `threading` and the `multiprocessing` have an identical API interface, and I did not observe any significant difference on the performance nor on the results... but we could try again if you like.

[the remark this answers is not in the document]
jeschkies: > In our case, the problem is I/O bound so we won't really benefit from multiprocessing. 

I fully agree but it seems you wrote a threadpool there. I did not go through the complete code but it seems that either the multiprocessing module or asyncio would be safer choices since they've been tested and used by a larger crowd.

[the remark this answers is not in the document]
wavesoft: That's true. But again, as long as you are doing I/O-bound operations on the threads, the GIL is the least of your concerns. Your python threads will behave almost identical to system threads. (Ps. `asyncio` is also affected by GIL, though it's not affected by context-switching, which makes it a teensy bit better on the performance).

However, keep in mind that the driver evolved into it's current state, it was not designed like so. In the beginning it was a pure, single-threaded application, and the multi-threading (and eventually the thread-pool) was introduced as a workaround to allow multiple blocking operations to happen at the same time.

If I was to start again from scratch, I am definitely going to use `asyncio`. So if you are interested on a re-write, let's pair on it 😄

wavesoft: > Did you try it?

Yup

> Does it produce better run time?

Nop 😭

tests/performance/config/perf-driver/reporters/report-datadog-live.yml
```diff
@@ -27,4 +27,4 @@ reporters:
     report:
       metrics: all
       summarized: yes
-      raw: yes
+      raw: no
```
zen-dog: hu?

tests/performance/live.sh
```diff
@@ -37,3 +41,9 @@ docker run -i --rm \
     ./tests/performance/ci_run_dcluster.sh \
     -Djmx_host=marathon_1 -Djmx_port=9010 -Dmarathon_url=http://marathon_1:8080 \
     -Mgit_hash="${GIT_HASH}"
+
+# Docker tends to leave lots of garbage ad the end, so
+# we should clean the volumes and remove the marathon
+# images that we just created.
+docker volume prune -f
```
zen-dog: 👏

tests/performance/config/perf-driver/reporters/report-datadog-live.yml
wavesoft: Questions regarding a configuration? http://dcos-performance-test-driver.readthedocs.io/en/latest/classes/Reporters.html#datadogmetricreporter 😄 

> ```
> # [Optional] Set to `yes` to submit the raw values the moment they
> # are collected.
> raw: no
> ````

PR #6343, opened 2018-07-05
Retry deployment readiness assertion.

jeschkies:
Summary:
The test case can fail with an index error. I assume that the deployment
did not start at all. So we have to retry the call until we can verify
that the app is not ready yet.

### DIFF
--- a/tests/system/dcos_service_marathon_tests.py
+++ b/tests/system/dcos_service_marathon_tests.py
@@ -11,6 +11,14 @@ from dcos import marathon
 from dcos.errors import DCOSUnprocessableException
 
 
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=16, retry_on_exception=common.ignore_exception)
+def assert_in_deployment(deployment_id):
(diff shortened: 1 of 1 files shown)


tests/system/dcos_service_marathon_tests.py
```diff
@@ -11,6 +11,14 @@
 from dcos.errors import DCOSUnprocessableException
 
 
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=16, retry_on_exception=common.ignore_exception)
+def assert_in_deployment(deployment_id):
```
zen-dog: The name is misleading - the method checks `readinessCheckResults` for a given deployment so wdyt about `assert_deployment_not_ready` or smth.?

jeschkies: ✅

jeschkies:
The system test [succeeded on open](https://jenkins.mesosphere.com/service/jenkins/job/system-integration-tests/job/marathon-si-pr/197).

PR #6345, opened 2018-07-06
Introduce and use Subject to fix race-condition when sending last leadership transition

timcharper:
This commit fixes a race condition that existed in ElectionService.subscribe. The old logic would:

1. Add a subscription to the event stream
2. Send the last known election state value

It was possible that between 1-2, a new value was available, and that this value was published to the subscriber before the last known election state value was updated. This would result in step 2 sending a stale election state value after a newer one was published.

To solve, we introduce a Subject component, which provides the following guarantees:

1) The last received event will be propagated to new subscribers
2) The last received event will not be propagated out of order
3) All events following the initially published last received event will be delivered.

JIRA Issues: MARATHON-8305

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
@@ -2,6 +2,8 @@ package mesosphere.marathon
 package core.event.impl.stream
 
 import akka.actor._
+import akka.stream.ActorMaterializer
+import akka.stream.scaladsl.Sink
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.election.{ElectionService, LeadershipTransition}
 import mesosphere.marathon.core.event.MarathonEvent
@@ -33,16 +35,18 @@ class HttpEventStreamActor(
     metrics: HttpEventStreamActorMetrics,
     handleStreamProps: HttpEventStreamHandle => Props)
   extends Actor with StrictLogging {
+  implicit val materializer = ActorMaterializer()
   //map from handle to actor
   private[impl] var streamHandleActors = Map.empty[HttpEventStreamHandle, ActorRef]
+  var electionEventsSubscription: Option[Cancellable] = None
 
   override def preStart(): Unit = {
     metrics.numberOfStreams.setValue(0)
-    electionService.subscribe(self)
+    electionEventsSubscription = Some(electionService.leadershipTransitionEvents.to(Sink.foreach(self ! _)).run)
   }
 
   override def postStop(): Unit = {
-    electionService.unsubscribe(self)
+    electionEventsSubscription.foreach(_.cancel())
     metrics.numberOfStreams.setValue(0)
   }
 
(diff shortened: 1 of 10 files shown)


timcharper:
I struggled with introducing the Subject abstraction. I ultimately decided it was the last hacky way to solve this. I explored versioning the leadership transition states and then having logic to drop out-of-order events, but this did not provided guarantees that we would completely miss an update in the following scenario:

1) Transition event 2 is in flight
2) We add out subscription, just barely missing transition event 2
3) ElectionService re-publishes the last seen transition event, event 1 (initial state)
4) ElectionService receives transition event 2 and updates it's local value

In this case, we would completely lose event 2 and the subscriber would erroneously think that event 1 is the current value.

timcharper:
Even though Subject seems to work under pretty heavy testing, I'm still unsure about introducing it. I tried an alternate approach involving versioning to resolve the race without introducing another component. I'm undecided if it's better.

timcharper:
After consideration, I'm inclined to stick with the Subject implementation since it separates the concurrency complexity concern better from ElectionService. Also, it is much easier to test the implementation in isolation.

src/main/scala/mesosphere/marathon/stream/Subject.scala
```diff
@@ -0,0 +1,117 @@
...
+private class SubjectInternalLogic[T](queueSize: Int, overflowStrategy: OverflowStrategy) {
+  @volatile private[this] var subscribers: Set[SourceQueueWithComplete[T]] = Set.empty
+  @volatile private[this] var lastElement: T = _
+  @volatile private[this] var elementPushed = false
+  @volatile private[this] var completionResult: Option[Try[Done]] = None
+
+  def onElement(t: T): Unit = synchronized {
+    elementPushed = true
+    lastElement = t
+    subscribers.foreach(_.offer(t))
```
timcharper: Because of our usage of SourceQueue, logic handling downstream stages of the graph will happen in a different thread / outside of this call context. This mitigates the possibility of deadlock here.

timcharper:
There are other uses of Subject in our code base for RX. With the introduction of this component we can remove RX from our code base and move completely to Akka Streams, reducing the number of concurrency abstractions we are using.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
```diff
@@ -87,14 +91,18 @@ class HttpEventStreamActor(
       streamHandleActors += handle -> actor
   }
 
+  private[stream] var isActive = false
```
alenkacz: hmm, we have this only because of testing? ... 🤔

alenkacz:
@timcharper yeah in general I am not very happy about dragging such a non-trivial piece of code inside the code base (as the Subject is). But I believe you revised other options and this was the best one :)

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamActor.scala
timcharper: Yeah :/ lame, I know. Options were:

- Have the test send the messages directly (and bypass the stream). Lost coverage. Maybe okay.
- Find a way to synchronously materialize an akka stream in tests. Same-thread dispatcher?
- Find a way to pull out the current behavior context and compare the function reference to check the actor state

¯\\\_(ツ)\_/¯ adding this was an easier option.

timcharper:
I've reached out to the Akka development team about the process to propose the inclusion of Subject in Akka core. There isn't a lot of urgency to this PR, so we can hold the merge until we have a clear picture of where Subject will end up.

PR #6346, opened 2018-07-06
Introduce `eventually` matcher to system tests.

jeschkies:
Summary:
Using a hamcrest like `assert_that` we can wrap retrying once in an
`eventually` matcher so that we don't have to rewrite them each time.
The assertions become much cleaner this way. The Python library
`precisely` provides a nice pattern to define custom matchers.


tests/system/matcher/eventually.py
```diff
@@ -0,0 +1,55 @@
...
+                retry_on_exception=common.ignore_exception,
+                retry_on_result=lambda r: r.is_match is not True)
+        def try_match():
+            actual = item()
+            return self.matcher.match(actual)
+
+        try:
+            return try_match()
+        except retrying.RetryError as e:
+            explanation = "after {} retries has no {}".format(e.last_attempt.attempt_number, self.matcher.describe())
```
jeschkies: The actual value is missing here. This would make debugging hard. I'll add it once we agreed if we want to move forward.

tests/system/matcher/eventually.py
```diff
@@ -0,0 +1,55 @@
...
+
+class HasEventually(Matcher):
+
+    def __init__(self, matcher):
+        self.matcher = matcher
+
+    def match(self, item):
+        assert callable(item), "The actual value is not callable."
+
+        @retrying.retry(
```
zen-dog: We need to somehow parametrize `HasEventually` e.g. like ScalaTest's `patienceConfig`. Otherwise, this here only waits for 3s which is not applicable in most tests.

jeschkies: @zen-dog, could you take another look?

jeschkies:
> I'm mixed about this... for 1 there is an attempt to make generic without doing it. There is a lot of additional code without any gained benefit. The core change L269-270 of marathon_common isn't changed much. I get the desire to change it and have it more clear for more occurrences of such a test

I hesitated to change too many test to not blow up the diff. The idea is to use `eventually` everywhere instead of the `retrying`. If you like to, I'll add it in more place. 

> The biggest issue I had is if we are going to make it more generic then it should go in shakedown... but in the end the matcher function isn't generic. has_failure_message has AppFailureMessage which is marathon specific and app specific. The method has_failure_message should be something like has_app_failure_message

`AppFailureMessage` is a bad example from my side. It is indeed very specific. I was aiming more for `assert_that` and `eventually`. Also, the tests *are* specific right now. `assert_that` introduces a pattern to make them more generic. I'll try to make the failure message generic. I had something like this in mind

```python
assert_that(lambda: client.get_app(app_def["id"]),
                   eventually(property(['lastTaskFailure', 'message'], contains_string("No such user 'bad'")),max_attempts=3))
```

Instead of nesting we could go for chaining
```python
assert_that(lambda: client.get_app(app_def["id"])).eventually(max_attempts=30).
                    property(['lastTaskFailure', 'message']).contains_string("No such user 'bad'")
```

What do you think?

> It is also a large change from 30 attempts to 3... if it is eventually why do you want such a small number of attempts? And why to configure the number of attempts for this test while also still using the default?

I used 3 for debugging and forgot to change it to 30 again.

I do plan to introduce the matchers to Shakedown once we have a feel for it. I hope that Shakedown will evolve to a collection of fixtures, matchers and helpers.

kensipe:
also.. I'm happy with either of the code examples provided... I think I lean towards

```
assert_that(lambda: client.get_app(app_def["id"]),
            eventually(property(['lastTaskFailure', 'message'], contains_string("No such user 'bad'")),max_attempts=30))
```

PR #6347, opened 2018-07-06
Activate `HeartbeatMonitor` even before Marathon is connected to Mesos

zen-dog:
Summary:
Fixes a rare bug where Mesos scheduler driver is hanging while connecting to Mesos. This change
will make sure that the `HeartbeatMonitor` is active even before Marathon registers with Mesos master and thus has a chance to kill Marathon.

Additionally, some unused code was removed from the `HeartbeatMonitor` specifically the `Reactor.Decorator`.

JIRA: MARATHON-8304, COPS-3413

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonModule.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonModule.scala
@@ -57,9 +57,8 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, actorSystem: ActorSyste
   @Singleton
   def provideMesosHeartbeatActor(system: ActorSystem): ActorRef = {
     system.actorOf(Heartbeat.props(Heartbeat.Config(
-      FiniteDuration(conf.mesosHeartbeatInterval.getOrElse(
-        MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_INTERVAL_MS), TimeUnit.MILLISECONDS),
-      conf.mesosHeartbeatFailureThreshold.getOrElse(MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_FAILURE_THRESHOLD)
+      FiniteDuration(conf.mesosHeartbeatInterval(), TimeUnit.MILLISECONDS),
+      conf.mesosHeartbeatFailureThreshold()
     )), ModuleNames.MESOS_HEARTBEAT_ACTOR)
   }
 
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/core/CoreGuiceModule.scala
```diff
@@ -238,4 +238,7 @@ class CoreGuiceModule(config: Config) extends AbstractModule {
 
   @Provides @Singleton
   def scheduler(coreModule: CoreModule): Scheduler = coreModule.mesosHeartbeatMonitor
+
+  @Provides @Singleton
+  def heartbeatMonitor(coreModule: CoreModule) = coreModule.mesosHeartbeatMonitor
```
zen-dog: Not proud of this part since both `scheduler` and `heartbeatMonitor` are the same object, but injecting a `HeartbeatMonitor` where Mesos `Scheduler` is needed seems worse.

src/main/scala/mesosphere/marathon/MarathonModule.scala
```diff
@@ -57,9 +57,8 @@ class MarathonModule(conf: MarathonConf, http: HttpConf, actorSystem: ActorSyste
   @Singleton
   def provideMesosHeartbeatActor(system: ActorSystem): ActorRef = {
     system.actorOf(Heartbeat.props(Heartbeat.Config(
-      FiniteDuration(conf.mesosHeartbeatInterval.getOrElse(
-        MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_INTERVAL_MS), TimeUnit.MILLISECONDS),
-      conf.mesosHeartbeatFailureThreshold.getOrElse(MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_FAILURE_THRESHOLD)
+      FiniteDuration(conf.mesosHeartbeatInterval(), TimeUnit.MILLISECONDS),
```
alenkacz: 👍

zen-dog:
> Are we sure this will shutdown properly, given that the framework was never registered?

I did manual testing using netcat and it does. 

> Also, Are we good with having the same timeout applied when not being registered in time, as we have when actually missing heartbeats?

I thought about it too but couldn't find a reason to introduce new parameters. Semantically "missing messages" also means "missing messages starting with the first one"

PR #6348, opened 2018-07-06
Graph offer matching.

jeschkies:
Summary:
This introduces three new visualizations and one dashboard to see how
many offers Marathon receives and how many time out.

### DIFF
--- a/tools/elk/conf/filter-marathon-1.4.x.conf
+++ b/tools/elk/conf/filter-marathon-1.4.x.conf
@@ -103,7 +103,21 @@ filter {
     remove_tag => ["unclassified"]
   }
   grok {
+    match => {"message" => "^Start processing offer %{UUID:offerId}. Current offer matcher count: %{NONNEGINT:matchersCount:int}"}
+    add_field => {"class" => "offer"}
+    add_field => {"class2" => "processed"}
+    tag_on_failure => []
+    remove_tag => ["unclassified"]
+  }
+  grok {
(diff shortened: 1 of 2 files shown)


tools/elk/kibana-visualizations.json
```diff
@@ -121,7 +121,49 @@
...
+      "uiStateJSON": "{}",
+      "description": "",
+      "version": 1,
+      "kibanaSavedObjectMeta": {
+        "searchSourceJSON": "{\"index\":\"7b77cd80-8117-11e8-97a6-6dd064d551ab\",\"filter\":[],\"query\":{\"query\":\"\\\"reason:Queue Timeout\\\"\",\"language\":\"lucene\"}}"
```
jeschkies: I have to verify that this search term is correct.

tools/elk/conf/filter-marathon-1.4.x.conf
```diff
@@ -103,7 +103,21 @@ filter {
     remove_tag => ["unclassified"]
   }
   grok {
-    match => {"message" => "^Finished processing %{UUID:offerId}"}
+    match => {"message" => "^Start processing offer %{UUID:offerId}. Current offer matcher count: %{NONNEGINT:matchersCount:int}"}
```
zen-dog: So we have `start`, `finish` and `no match`. Should we also add a grok for `match`? Maybe we need to adjust logging for that too?

jeschkies: We should adopt our logs. However, these are in 1.5 and 1.6 :/.

PR #6351, opened 2018-07-10
Introduce option to suppress offers

timcharper:
This reverts commit bb6420c4.

- Bring back the logic for offer suppression and introduction a
  command-line option to enable it.
- Documentation.

JIRA Issues: MARATHON-8309

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
@@ -25,7 +25,7 @@ private[flow] object ReviveOffersActor {
   }
 
   private[impl] case object TimedCheck
-  private[impl] case object OffersWanted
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/core/flow/impl/ReviveOffersActor.scala
```diff
@@ -88,6 +92,11 @@ private[impl] class ReviveOffersActor(
     }
   }
 
+  private[this] def suppressOffers(): Unit = {
+    logger.info("=> Suppress offers NOW")
```
zen-dog: "NOW"? Are you shouting at the driver? :D

timcharper: lol this is what it was before

zen-dog: Yeah, saw that too ^^

PR #6366, opened 2018-07-11
Fix n^2 parallelism, and some other tweaks

timcharper:
`ZookeeperPersistenceStore` used `groupBy` + `mapAsync` which resulted in n-squared parallelism, rather than n parallelism. Additionally, using an iterator with `Source` was unnecessary as the collection was already immutable.


src/main/scala/mesosphere/marathon/core/storage/zookeeper/ZooKeeperPersistenceStore.scala
```diff
@@ -64,7 +64,7 @@ class ZooKeeperPersistenceStore(factory: AsyncCuratorBuilderFactory, parallelism
       // `groupBy(path.hashCode % parallelism)` makes sure that updates to the same path always land in the same
       // sub-stream, thus keeping the order of writes to the same path, even with parallelism > 1
       .groupBy(parallelism, node => Math.abs(node.path.hashCode) % parallelism)
-      .mapAsync(parallelism)(node => create(node))
+      .mapAsync(1)(node => create(node))
```
zen-dog: Good catch!

PR #6367, opened 2018-07-12
Remove default value from Goal

alenkacz:
I think it's safer without default value. Also there was one bug that went unnoticed because of that.


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -134,7 +134,7 @@ object Instance {
       val healthy = computeHealth(tasks.toVector)
       maybeOldState match {
         case Some(state) if state.condition == condition && state.healthy == healthy => state
-        case _ => InstanceState(condition, now, active, healthy)
+        case _ => InstanceState(condition, now, active, healthy, maybeOldState.map(_.goal).getOrElse(Goal.Running))
```
alenkacz: this was a bug. We did not re-use the goal when updating instance state and always put that to Running (which was default)

PR #6372, opened 2018-07-16
Cleaning-up previous results using privileged container

wavesoft:
This commit removes the clean-up lines in the beginning of `live.sh` and
inserts them as arguments to the privileged `rm` command inside the container


jeschkies:
@wavesoft did you run the script through shellcheck?

PR #6373, opened 2018-07-16
Minor improvements on our feature branch

alenkacz:

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -51,11 +51,6 @@ case class Instance(
 
   lazy val isProvisioned: Boolean = state.condition == Condition.Provisioned
 
-  def isCreated: Boolean = state.condition == Condition.Created
-  def isError: Boolean = state.condition == Condition.Error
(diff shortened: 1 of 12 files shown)


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -51,11 +51,6 @@ case class Instance(
 
   lazy val isProvisioned: Boolean = state.condition == Condition.Provisioned
 
-  def isCreated: Boolean = state.condition == Condition.Created
```
alenkacz: I think this was a bad merge

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -261,8 +256,6 @@ object Instance {
 
       //From here on all tasks are only in one of the following states
       Condition.Created,
-      Condition.Provisioned,
```
alenkacz: tasks should never have this conditions, right?

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: Yup

PR #6374, opened 2018-07-17
Resolve a smattering of warnings

timcharper:
Mesos getRole, implicit numeric widening, pattern match, etc.

Import the Scala warnings silencer compiler plugin, and use in places where warnings are not helpful.


src/main/proto/README.md
```diff
@@ -16,3 +16,11 @@ To rebuild the protos, open a terminal where the current working directory is th
 $> cd src/main/proto
 $> protoc --java_out=../java/ marathon.proto
 ```
+
+Finally, edit `Protos.java` and add `@SuppressWarnings("deprecation") ` before `public final class Protos`, like so:
+
+```
+@SuppressWarnings("deprecation") public final class Protos {
```
alenkacz: oh it would be wonderful if this was done by a protoc plugin.

But yeah, agree it would be overkill 🗡 :)

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -43,13 +43,13 @@ class TaskKiller @Inject() (
           val activeInstances = foundInstances.filter(_.isActive)
 
           if (wipe) {
-            val decommissioned = await(Future.sequence(foundInstances.map(i => instanceTracker.setGoal(i.instanceId, Goal.Decommissioned))))
```
alenkacz: I thought that someone told me that we need to assign the results when using async/await. But I don't remember who that was and I did not find anything about that limitations in async/await docs... 🤔

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -46,7 +46,7 @@ private[marathon] class InstanceUpdateOpResolver(
       case op: GoalChange =>
         updateExistingInstance(op.instanceId)(i => {
           val updatedInstance = i.copy(state = i.state.copy(goal = op.goal))
-          val events = InstanceChangedEventsGenerator.events(updatedInstance, task = None, clock.now(), previousCondition = Some(i.state.condition))
+          InstanceChangedEventsGenerator.events(updatedInstance, task = None, clock.now(), previousCondition = Some(i.state.condition))
```
alenkacz: I think this is a bug, the events should be passed to the expression on l.52 or this whole line can be removed - right now no events are generated anyway so it does not matter...

src/main/scala/mesosphere/marathon/metrics/HistogramTimer.scala
```diff
@@ -97,12 +97,16 @@ private[metrics] case class HistogramTimer(name: String, tags: Map[String, Strin
...
+  }
+}
+
+private[metrics] object HistogramTimer {
+  def durationToUnits(duration: FiniteDuration, unit: Time): Long = unit match {
+    case Time.Nanoseconds => duration.toNanos
+    case Time.Milliseconds => duration.toMillis
+    case Time.Microseconds => duration.toMicros
+    case Time.Seconds => duration.toSeconds
+    case Time(factor, _) => (duration.toNanos * (Time.Nanoseconds.factor / factor)).toLong
```
alenkacz: so in this case the unit is not used at all? 🤔

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
timcharper: File an issue? :)

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
timcharper: You don't need to assign it. We assigned it to try and remove one warning, but it just generates another one. These are Future[Done] instead of Future[Unit] (which is helpful! but not for warnings).

timcharper:
@alenkacz the silencer is local and I've applied it as locally as possible. Most of the time to direct expressions, only a few times to entire methods, never to an entire class.

Suppressing warnings seems like the only sane thing to do. Where there was no solution for the warnings (unassigned val), these warnings just aren't helpful. In general, though, I would use suppress as a last resort and fix warnings.

I did not remove all warnings with this PR. Just the obvious ones (obviously fixable or obviously unhelpful).

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
alenkacz: remove it or don't touch this, please :) because the code after the change does not make much sense (as it did not before, it's not really an improvement :) we just got rid of warning which was actually legit here)

src/main/proto/README.md
jeschkies: Hm, this will be a longer discussion I'm afraid. It should not block this PR. In any case, @bbannier mentioned several times that we should not check in neither the generated class files nor the protos but rather download them as part of our dependencies. In general I agree with him. However, I don't know how we should model this dependency.

src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
```diff
@@ -26,7 +26,7 @@ object MarathonSchedulerDriver extends StrictLogging {
       .setHostname(config.hostname())
 
     // Set the role, if provided.
-    config.mesosRole.foreach(frameworkInfoBuilder.setRole)
+    config.mesosRole.foreach(frameworkInfoBuilder.setRole: @silent)
```
jeschkies: What is the deprecation warning about here? I thought it's just a different method we should call.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
jeschkies: It was me and I did see a bug not assigning it. Maybe async/await had a bug or it was just coincident. Could we have a loop for this just to be sure? I think we can wait a day.

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -69,7 +69,7 @@ class OfferOperationFactory(
         Mesos.Resource.newBuilder(resource)
           .setRole(role)
           .setReservation(reservation)
-          .build()
+          .build(): @silent
```
jeschkies: What was the deprecation warning?

timcharper: `setRole`

src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
timcharper: `setRole`. It's been showing for a long time.

src/main/proto/README.md
timcharper: It'd be neat if the protos were published.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
timcharper: huh, wild. Yeah we can run a loop

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
timcharper: Okay :) This is the problem with so many warnings. Valuable ones are lost in the noise.

src/main/scala/mesosphere/marathon/metrics/HistogramTimer.scala
timcharper: The other field is a just a textual description of the time unit. I've added a test case for this.

src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
timcharper: I raised another question in the mesos channel

https://mesos.slack.com/archives/C1K2TAJ1L/p1531969087000039

timcharper: Perhaps we should just turn on multi-role. It doesn't look terribly difficult to support and we could probably get by just grabbing the first role or assuming nothing.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
timcharper: off it goes! https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-smattering-warnings/

src/main/scala/mesosphere/marathon/MarathonSchedulerDriver.scala
timcharper: I added an acceptance criteria to this ticket saying that we should remove all of these `@silent` tags when implementing multi-role support.

https://jira.mesosphere.com/browse/MARATHON-2775

jeschkies: Thanks for the explanation. I thought it was a matter of just switching to `setRoles` but it seems a little more.

src/main/proto/README.md
bbannier: https://issues.apache.org/jira/browse/MESOS-9095

timcharper: 🙇

timcharper:
🎉

PR #6376, opened 2018-07-18
Try to fix races in resident relaunching

alenkacz:


alenkacz:
@jeschkies oh, that's a much nicer pattern, thanks!

Yeah I just stole something from InstanceUpdater to see if this solves the races. I am all for ideas on how to make this code better. I would not feel good merging this to the feature branch :)

jeschkies:
> I would not feel good merging this to the feature branch :)

How about we run a loop on this branch to validate that it's the issue?

alenkacz:
@jeschkies yes! let's do that :)

alenkacz:
@jeschkies now that I am think about it, I don't think I can use the pattern you proposed :( Because I am serializing only per runspecId, all others still run in parallel while your stash pattern would mean that I do only one at a time...

jeschkies:
The last 70 builds of this branch are green. Let's merge it!

PR #6379, opened 2018-07-23
Stabilize and cleanup system tests with `eventually`.

jeschkies:
Summary:
This is a small follow up that replaces a few retrying blocks with the
new `eventually` matcher. It also introduces `has_len` to match the
length of a sequence.

### DIFF
--- a/tests/system/matcher/property.py
+++ b/tests/system/matcher/property.py
@@ -1,5 +1,6 @@
 from precisely import Matcher
-from precisely.results import unmatched
+from precisely.coercion import to_matcher
+from precisely.results import matched, unmatched, indented_list
 
 
 class Prop(Matcher):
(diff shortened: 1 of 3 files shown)


jeschkies:
Open system tests pass.

jeschkies:
Strict fails only with `test_marathon_root.test_private_repository_docker_app` which is know to be broken.

tests/system/matcher/property.py
```diff
@@ -43,3 +44,52 @@ def describe(self):
 
 def prop(path, matcher):
     return Prop(path, matcher)
+
+
+def has_value(name, matcher):
```
wavesoft: Please start adding unit tests for this 👼

jeschkies: I thought about unit tests but chose not to because these methods are tested implicitly everywhere. I will add docs in the next diff.

> It looks OK, but I have a feeling it gets less expressive

I feel the opposite. It gets way more expressive since it states what we assert.

jeschkies: See #6382 for the docstrings.

wavesoft: > I feel the opposite. It gets way more expressive since it states what we assert.

I agree, but you are effectively introducing your DSL, which means that a new person should have a clean documentation on how to use it ;)

PR #6380, opened 2018-07-23
Improve the log message and docs

alenkacz:


src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -352,7 +352,7 @@ private class TaskLauncherActor(
     }
 
     val inFlight = inFlightInstanceOperations.size
-    val activeInstances = instanceMap.values.count(_.isActive)
+    val activeInstances = instanceMap.values.filterNot(_.isProvisioned).count(_.isActive)
```
alenkacz: active is also provisioned. In the current state we print out something like: 1 in flight, 3 confirmed when the target count is 2, the provisioned instance is counted as in flight as well as confirmed, which seems misleading

jeschkies: Hm, how about renaming `activeInstances` to `launchedInstances` or `confirmedInstances`?

PR #6381, opened 2018-07-23
Sync only specific instance id

alenkacz:


src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -155,7 +155,7 @@ class TaskReplaceActor(
 
   def reconcileAlreadyStartedInstances(): Unit = {
     logger.info(s"reconcile: found ${instancesAlreadyStarted.size} already started instances " +
-      s"and ${oldInstanceIds.size} old instances: ${currentRunningInstances.map{ i => i.instanceId -> i.state.condition }}")
+      s"and ${oldInstanceIds.size} old instances: ${currentInstances.map{ i => i.instanceId -> i.state.condition }}")
```
jeschkies: How about logging the goal and the condition?

jeschkies:
So I see

```
old version 2018-07-24T14:21:46.973Z to
new version 2018-07-24T14:21:48.110Z

instance [app-with-backoff-delay-is-reset-on-conf-changes.instance-e6588f8d-8f4c-11e8-89bc-02429e0d1256]:{condition: Scheduled, goal: Running, version: 2018-07-24T14:21:46.973Z, reservation: None, 
instance [app-with-backoff-delay-is-reset-on-conf-changes.instance-e683972e-8f4c-11e8-89bc-02429e0d1256]:{condition: Scheduled, goal: Running, version: 2018-07-24T14:21:48.110Z, reservation: None
instance [app-with-backoff-delay-is-reset-on-conf-changes.instance-eb565d7e-8f3e-11e8-ba52-024295b3bdfa]:{condition: Scheduled, goal: Running, version: 2018-07-24T12:41:42.978Z, reservation: None}
```

We have three apps with different versions each :/.

jeschkies:
I think I've found another issue in `InstanceUpdater:48`:

```scala
private def shouldBeExpunged(instance: Instance): Boolean =
  instance.tasksMap.values.forall(_.isTerminal) && instance.state.goal != Goal.Stopped
```

this should be

```scala
private def shouldBeExpunged(instance: Instance): Boolean =
  instance.tasksMap.values.forall(_.isTerminal) && instance.state.goal == Goal.Decommissioned
```

A scheduled instances has no tasks so `instance.tasksMap.values.forall(_.isTerminal)` evaluates to `true`. See https://scastie.scala-lang.org/KFpL3wLqSCi4ebO00FF6DA.

jeschkies:
I just saw that the instance is updated with the task. It should be `goal == Goal.Decommissioned` Nonetheless.

jeschkies:
I've just realized that we did the expunge to replace failed instances. Let's discuss this tomorrow.

alenkacz:
well we can't do `instance.state.goal == Goal.Decommissioned` because we don't decommission ephemeral instances :/

jeschkies:
> well we can't do instance.state.goal == Goal.Decommissioned because we don't decommission ephemeral instances :/

We do decommission ephemeral instances. In fact, we always decommission them instead of reusing the same instance on a restart.

alenkacz:
@jeschkies Yeah, but we don't set the goal to decommissioned.

PR #6384, opened 2018-07-24
Run Docker app as root to grant read permission to credentials.

jeschkies:
Summary:
The system tests fail on a strict EE cluster with CentOS because the
user `nobody` has no read access to the Docker credentail tarball in
`/home/centos`. With this change we run the app as `root` which has
read access.

JIRA issues: MARATHON-8324


jeschkies:
Strict is green!

PR #6340, opened 2018-07-04
Instance Goals Phase 2

jeschkies:
All commits of this pull request have been reviewed individually.

[Here](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-goals-phase-2/) is the build loop.

JIRA issues: MARATHON-8167

Outstanding todos to land this diff

* [x] Merge #6418.
* [x] Bring back reserved condition.
* [x] Run [system tests](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/system-integration-tests/job/marathon-si-dcos-open/job/karsten%252Ftest-phase-2/).
* [x] Run benchmarks.
 
# Latest Test Runs (2018-09-26)

## Integration Tests

screen shot 2018-09-26 at 14 26 18

```
   1 mesosphere.marathon.core.election.CuratorElectionStreamTest:Abdicates leadership immediately when the client is closed
   1 mesosphere.marathon.integration.LaunchQueueIntegrationTest:LaunchQueue should GET /v2/queue with pending app
Sample size: 193
```

## System Tests

```
   1 dcos-launch:cluster.create
   1 test_marathon_root:test_app_update_rollback
   1 test_marathon_root:test_pod_with_container_network
   2 test_marathon_root:test_launch_docker_grace_period
   2 test_marathon_root:test_scale_app_in_group
   3 test_marathon_root:test_marathon_zk_partition_leader_change
   6 test_marathon_root:test_launch_group
   8 test_marathon_root:test_launch_and_scale_group
  10 test_marathon_root:test_scale_app_in_group_then_group
  74 test_marathon_universe:test_custom_service_name
  74 test_marathon_universe:test_install_marathon
Sample size: 98
```

`test_scale_app_in_group_then_group` did not fail in the last 67 runs. `test_custom_service_name` and `test_install_marathon` failed due to a DC/OS bug.

## Benchmarks

Benchmarks are running on [marathon-benchmarks-nuc/feature/goal/phase2](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/Marathon-Scale-Tests/job/marathon-benchmarks-nuc/job/feature%252Fgoal%252Fphase2/)

PR #6392, opened 2018-07-26
Define second level scheduler interface.

jeschkies:
Summary:
The orchestration layer should only communicate via the scheduler
interface. The instance tracker, launch queue and kill service will be
hidden behind that. This diff defines said interface and implements a
legacy scheduler that just forwards all requests to the underlying
modules.

JIRA issues: MARATHON-8224

# Outstanding

- [ ] Handle purging.
- [x] Reschedule stopped reserved instances.
- [x] Sequentialize updates in scheduler.

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -12,6 +12,7 @@ import mesosphere.marathon.core.election.LeadershipTransition
 import mesosphere.marathon.core.event.DeploymentSuccess
 import mesosphere.marathon.core.health.HealthCheckManager
 import mesosphere.marathon.core.instance.{Goal, Instance}
+import mesosphere.marathon.core.instance.update.InstanceUpdateOperation.RescheduleReserved
 import mesosphere.marathon.core.launchqueue.LaunchQueue
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.termination.{KillReason, KillService}
@@ -420,7 +421,7 @@ class SchedulerActions(
     */
   // FIXME: extract computation into a function that can be easily tested
   def scale(runSpec: RunSpec): Future[Done] = async {
-    logger.debug("Scale for run spec {}", runSpec)
+    logger.info("Scale for run spec {}", runSpec)
 
     val instances = await(instanceTracker.specInstances(runSpec.id))
     val runningInstances = instances.filter(_.isActive)
@@ -470,7 +471,22 @@ class SchedulerActions(
 
       if (toAdd > 0) {
         logger.info(s"Queueing $toAdd new instances for ${runSpec.id} to the already $leftToLaunch queued ones")
-        await(launchQueue.add(runSpec, toAdd))
+        // TODO(karsten): Use scheduler interface or instance tracker instead or get rid of the scale checks.
+
+        // Reschedule stopped resident instances first.
+        val existingReservedStoppedInstances = instances
+          .filter(i => i.isReserved && i.state.goal == Goal.Stopped) // resident to relaunch
+          .take(toAdd)
+          .map(RescheduleReserved(_, runSpec.version))
+        await(Future.sequence(existingReservedStoppedInstances.map(instanceTracker.process)).map(_ => Done))
+
+        // Schedule remaining instances
+        val leftToAdd = math.max(0, toAdd - existingReservedStoppedInstances.length)
+        val instancesToSchedule = 0.until(leftToAdd).map { _ => Instance.scheduled(runSpec, Instance.Id.forRunSpec(runSpec.id)) }
+        await(instanceTracker.schedule(instancesToSchedule))
+
+        // Let launch queue manage task launcher actor.
+        await(launchQueue.sync(runSpec))
       } else {
         logger.info(s"Already queued or started ${runningInstances.size} instances for ${runSpec.id}. Not scaling.")
       }
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/AppStartActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/AppStartActor.scala
@@ -7,9 +7,7 @@ import akka.event.EventStream
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.event.DeploymentStatus
 import mesosphere.marathon.core.instance.Instance
-import mesosphere.marathon.core.launchqueue.LaunchQueue
 import mesosphere.marathon.core.readiness.ReadinessCheckExecutor
-import mesosphere.marathon.core.task.tracker.InstanceTracker
 import mesosphere.marathon.state.RunSpec
 
 import scala.concurrent.{Future, Promise}
@@ -17,9 +15,8 @@ import scala.concurrent.{Future, Promise}
 class AppStartActor(
     val deploymentManagerActor: ActorRef,
     val status: DeploymentStatus,
-    val scheduler: SchedulerActions,
-    val launchQueue: LaunchQueue,
-    val instanceTracker: InstanceTracker,
+    val schedulerActions: SchedulerActions,
+    val scheduler: scheduling.Scheduler,
     val eventBus: EventStream,
     val readinessCheckExecutor: ReadinessCheckExecutor,
     val runSpec: RunSpec,
@@ -33,7 +30,7 @@ class AppStartActor(
     // In case we already have running instances (can happen on master abdication during deployment)
     // with the correct version those will not be killed.
     val runningInstances = currentInstances.count(_.isActive)
-    scheduler.startRunSpec(runSpec.withInstances(Math.max(runningInstances, scaleTo)))
+    schedulerActions.startRunSpec(runSpec.withInstances(Math.max(runningInstances, scaleTo)))
   }
 
   override def postStop(): Unit = {
@@ -54,16 +51,15 @@ object AppStartActor {
   def props(
     deploymentManagerActor: ActorRef,
     status: DeploymentStatus,
-    scheduler: SchedulerActions,
-    launchQueue: LaunchQueue,
-    taskTracker: InstanceTracker,
+    schedulerActions: SchedulerActions,
+    scheduler: scheduling.Scheduler,
     eventBus: EventStream,
     readinessCheckExecutor: ReadinessCheckExecutor,
     runSpec: RunSpec,
     scaleTo: Int,
     currentInstances: Seq[Instance],
     promise: Promise[Unit]): Props = {
-    Props(new AppStartActor(deploymentManagerActor, status, scheduler, launchQueue, taskTracker, eventBus,
-      readinessCheckExecutor, runSpec, scaleTo, currentInstances, promise))
+    Props(new AppStartActor(deploymentManagerActor, status, schedulerActions, scheduler, eventBus, readinessCheckExecutor,
+      runSpec, scaleTo, currentInstances, promise))
   }
 }
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
@@ -10,12 +10,10 @@ import mesosphere.marathon.core.deployment.impl.DeploymentActor.{Cancel, Fail, N
 import mesosphere.marathon.core.deployment.impl.DeploymentManagerActor.DeploymentFinished
 import mesosphere.marathon.core.event.{AppTerminatedEvent, DeploymentStatus, DeploymentStepFailure, DeploymentStepSuccess}
 import mesosphere.marathon.core.health.HealthCheckManager
-import mesosphere.marathon.core.instance.{Goal, Instance}
-import mesosphere.marathon.core.launchqueue.LaunchQueue
+import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.core.readiness.ReadinessCheckExecutor
-import mesosphere.marathon.core.task.termination.{KillReason, KillService}
-import mesosphere.marathon.core.task.tracker.InstanceTracker
+import mesosphere.marathon.core.task.termination.KillReason
 import mesosphere.marathon.state.{AppDefinition, RunSpec}
 import mesosphere.mesos.Constraints
 
@@ -26,11 +24,9 @@ import scala.util.{Failure, Success}
 
 private class DeploymentActor(
     deploymentManagerActor: ActorRef,
-    killService: KillService,
-    scheduler: SchedulerActions,
+    schedulerActions: SchedulerActions,
+    scheduler: scheduling.Scheduler,
     plan: DeploymentPlan,
-    instanceTracker: InstanceTracker,
-    launchQueue: LaunchQueue,
     healthCheckManager: HealthCheckManager,
     eventBus: EventStream,
     readinessCheckExecutor: ReadinessCheckExecutor) extends Actor with StrictLogging {
@@ -135,8 +131,8 @@ private class DeploymentActor(
 
   def startRunnable(runnableSpec: RunSpec, scaleTo: Int, status: DeploymentStatus): Future[Unit] = {
     val promise = Promise[Unit]()
-    instanceTracker.specInstances(runnableSpec.id).map { instances =>
-      context.actorOf(childSupervisor(AppStartActor.props(deploymentManagerActor, status, scheduler, launchQueue, instanceTracker,
+    scheduler.getInstances(runnableSpec.id).map { instances =>
+      context.actorOf(childSupervisor(AppStartActor.props(deploymentManagerActor, status, schedulerActions, scheduler,
         eventBus, readinessCheckExecutor, runnableSpec, scaleTo, instances, promise), s"AppStart-${plan.id}"))
     }
     promise.future
@@ -145,11 +141,11 @@ private class DeploymentActor(
   private def killInstancesIfNeeded(instancesToKill: Seq[Instance]): Future[Done] = async {
     logger.debug("Kill instances {}", instancesToKill)
     val changeGoalsFuture = instancesToKill.map(i => {
-      if (i.hasReservation) instanceTracker.setGoal(i.instanceId, Goal.Stopped)
-      else instanceTracker.setGoal(i.instanceId, Goal.Decommissioned)
+      if (i.hasReservation) scheduler.stop(i, KillReason.DeploymentScaling)
+      else scheduler.decommission(i, KillReason.DeploymentScaling)
     })
     await(Future.sequence(changeGoalsFuture))
-    await(killService.killInstances(instancesToKill, KillReason.DeploymentScaling))
+    Done
   }
 
   def scaleRunnable(runnableSpec: RunSpec, scaleTo: Int,
@@ -162,7 +158,7 @@ private class DeploymentActor(
     }
 
     async {
-      val instances = await(instanceTracker.specInstances(runnableSpec.id))
+      val instances = await(scheduler.getInstances(runnableSpec.id))
       val runningInstances = instances.filter(_.state.condition.isActive)
       val ScalingProposition(instancesToKill, tasksToStart) = ScalingProposition.propose(
         runningInstances, toKill, killToMeetConstraints, scaleTo, runnableSpec.killSelection)
@@ -174,7 +170,7 @@ private class DeploymentActor(
         tasksToStart.fold(Future.successful(Done)) { tasksToStart =>
           logger.debug(s"Start next $tasksToStart tasks")
           val promise = Promise[Unit]()
-          context.actorOf(childSupervisor(TaskStartActor.props(deploymentManagerActor, status, scheduler, launchQueue, instanceTracker, eventBus,
+          context.actorOf(childSupervisor(TaskStartActor.props(deploymentManagerActor, status, scheduler, eventBus,
             readinessCheckExecutor, runnableSpec, scaleTo, promise), s"TaskStart-${plan.id}"))
           promise.future.map(_ => Done)
         }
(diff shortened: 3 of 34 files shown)


src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
```diff
@@ -146,8 +145,8 @@ private class DeploymentActor(
   private def killInstancesIfNeeded(instancesToKill: Seq[Instance]): Future[Done] = async {
     logger.debug("Kill instances {}", instancesToKill)
     val changeGoalsFuture = instancesToKill.map(i => {
-      if (i.hasReservation) instanceTracker.setGoal(i.instanceId, Goal.Stopped)
-      else instanceTracker.setGoal(i.instanceId, Goal.Decommissioned)
+      if (i.hasReservation) scheduler.stop(i.instanceId)
+      else scheduler.decommission(i.instanceId)
     })
     await(Future.sequence(changeGoalsFuture))
     await(killService.killInstances(instancesToKill, KillReason.DeploymentScaling))
```
jeschkies: I could move this call to `Scheduler.stop` and `Scheduler.decommission`. What do you think?

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
```diff
@@ -247,9 +246,9 @@ object DeploymentActor {
     Props(new DeploymentActor(
       deploymentManagerActor,
       killService,
+      schedulerActions,
       scheduler,
```
jeschkies: I'm not sure we should keep the `schedulerActions` around.

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentManagerActor.scala
```diff
@@ -407,17 +406,17 @@ object DeploymentManagerActor {
 
   @SuppressWarnings(Array("MaxParameters"))
   def props(
-    taskTracker: InstanceTracker,
     killService: KillService,
-    launchQueue: LaunchQueue,
-    scheduler: SchedulerActions,
+    launchQueue: LaunchQueue, // TODO(karsten): Remove launch queue and go through scheduling.Scheduler.
```
jeschkies: The todo is part of [phase 3](https://jira.mesosphere.com/browse/MARATHON-8292).

src/main/scala/mesosphere/marathon/scheduling/LegacyScheduler.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import scala.async.Async.{async, await}
+import scala.concurrent.{ExecutionContext, Future}
+
+case class LegacyScheduler(
+    offerProcessor: OfferProcessor,
+    instanceTracker: InstanceTracker,
+    statusUpdateProcessor: TaskStatusUpdateProcessor) extends Scheduler {
+
+  //override def create(runSpec: RunSpec): Future[Instance.Id]
```
jeschkies: This is an interesting one. The current setup uses `LaunchQueue.add(runSpec, numberOfInstances)`. I hesitate to use that as is.

jeschkies:
The `InstanceTracker`, `KillService` and maybe the health check logic will move into `mesosphere.marathon.scheduling` package. They might be even `private[scheduling]` so that we have a clear cut to `mesosphere.marathon.orchestration`.

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
alenkacz: yes, I think we should do that...

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
```diff
@@ -164,7 +163,7 @@ private class DeploymentActor(
     }
 
     async {
-      val instances = await(instanceTracker.specInstances(runnableSpec.id))
+      val instances = await(scheduler.getInstances(runnableSpec.id))
       val runningInstances = instances.filter(_.state.condition.isActive)
```
alenkacz: should we really pull Active instances here? Shouldn't it be Goal.Running?

src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
```diff
@@ -49,15 +47,17 @@ trait StartingBehavior extends ReadinessBehavior with StrictLogging { this: Acto
     case InstanceChanged(id, `version`, `pathId`, condition: Condition, instance) if condition.isTerminal || instance.isReservedTerminal =>
       logger.warn(s"New instance [$id] failed during app ${runSpec.id.toString} scaling, queueing another instance")
       instanceTerminated(id)
+      // TODO(karsten): scheduler.create(runSpec)
       launchQueue.add(runSpec, 1).pipeTo(self)
 
     case Sync => async {
-      val instances = await(instanceTracker.specInstances(runSpec.id))
+      val instances = await(scheduler.getInstances(runSpec.id))
       val actualSize = instances.count { i => i.isActive || i.isScheduled }
```
alenkacz: shouldn't we again just filter on Goal maybe?

jeschkies:
The loop looks quite good so far

```
   1 mesosphere.marathon.core.health.impl.MarathonHealthCheckManagerTest:HealthCheckManager should reconcile loads the last known task health state
   1 mesosphere.marathon.integration.GroupDeployIntegrationTest:GroupDeployment should An upgrade in progress cannot be interrupted without force
Sample size: 32
```

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
jeschkies: Good point. This is the orchestration layer. It should only filter for goals and not for conditions.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -185,11 +188,8 @@ class TaskReplaceActor(
           logger.info(s"Killing old ${nextOldInstance.instanceId}")
       }
 
-      if (runSpec.isResident) {
-        instanceTracker.setGoal(nextOldInstance.instanceId, Goal.Stopped)
-      } else {
-        instanceTracker.setGoal(nextOldInstance.instanceId, Goal.Decommissioned)
-      }
+      if (runSpec.isResident) { scheduler.stop(nextOldInstance.instanceId) }
```
jeschkies: We probably want to await the future before we issue the kill.

src/main/scala/mesosphere/marathon/scheduling/LegacyScheduler.scala
```diff
@@ -0,0 +1,45 @@
...
+import mesosphere.marathon.core.launcher.OfferProcessor
+import mesosphere.marathon.core.task.tracker.InstanceTracker
+import mesosphere.marathon.core.task.update.TaskStatusUpdateProcessor
+import mesosphere.marathon.state.PathId
+import org.apache.mesos.Protos
+
+import scala.async.Async.{async, await}
+import scala.concurrent.{ExecutionContext, Future}
+
+case class LegacyScheduler(
```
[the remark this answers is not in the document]
timcharper: I'm highly confident that it will be fast enough as well.

src/main/scala/mesosphere/marathon/scheduling/Scheduler.scala
```diff
@@ -0,0 +1,67 @@
...
+  /**
+    * Run all instances with given ids.
+    *
+    * This method is idempotent.
+    *
+    * @param instanceIds
+    * @param ec
+    * @return
+    */
+  def run(instanceIds: Instance.Id*)(implicit ec: ExecutionContext): Future[Done]
```
timcharper: How does one restart an instance?

Also, do we get a `Done` when the instance goal is updated, or when the instance is actually running?

jeschkies: > Also, do we get a Done when the instance goal is updated, or when the instance is actually running?

Only when the instance goal changed.

> How does one restart an instance?

That is an open question. The orchestrator will receive instance updates before the scheduler does. So it could do:

1. Call `scheduler.stop(instanceId)`
2. Receive `case InstanceUpdate(instanceId, state.Running) => scheduler.run(instanceId)`

src/main/scala/mesosphere/marathon/scheduling/Scheduler.scala
```diff
@@ -0,0 +1,76 @@
...
+import mesosphere.marathon.core.instance.Instance
+import mesosphere.marathon.core.launcher.OfferProcessor
+import mesosphere.marathon.state.PathId
+import org.apache.mesos.Protos
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait Scheduler extends OfferProcessor {
+
+  //def create(runSpec: RunSpec): Future[Instance.Id]
```
zen-dog: Why is creating an Instance not part of the interface?

jeschkies: It should but
* I did not really know what it should look like.
* With the lanuch queue still in place it's not trivial to actually implement the interface.

src/main/scala/mesosphere/marathon/core/deployment/impl/AppStartActor.scala
```diff
@@ -9,17 +9,16 @@ import mesosphere.marathon.core.event.DeploymentStatus
...
-import mesosphere.marathon.core.task.tracker.InstanceTracker
 import mesosphere.marathon.state.RunSpec
 
 import scala.concurrent.{Future, Promise}
 
 class AppStartActor(
     val deploymentManagerActor: ActorRef,
     val status: DeploymentStatus,
-    val scheduler: SchedulerActions,
+    val schedulerActions: SchedulerActions,
```
zen-dog: I don't see a change to `mesosphere.marathon.SchedulerActions` to use the scheduler. Wdyt about changing it too?

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
zen-dog: second that

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
zen-dog: It has been a thorn in my side for a while now. However, it's logic should be a part of the _Orchestrator_ and not to the _Scheduler_. Nevertheless, we should make it use the new _Scheulder_ instead of _LaunchQueue_ and _KillService_. Wdyt?

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
```diff
@@ -176,7 +175,7 @@ private class DeploymentActor(
         tasksToStart.fold(Future.successful(Done)) { tasksToStart =>
           logger.debug(s"Start next $tasksToStart tasks")
           val promise = Promise[Unit]()
-          context.actorOf(childSupervisor(TaskStartActor.props(deploymentManagerActor, status, scheduler, launchQueue, instanceTracker, eventBus,
+          context.actorOf(childSupervisor(TaskStartActor.props(deploymentManagerActor, status, scheduler, launchQueue, eventBus,
```
zen-dog: I think you can get rid of the `launchQueue` here - it's needed in the _StartingBehavior_ down the line, which _Scheduler_ makes obsolete now.

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
```diff
@@ -246,9 +245,9 @@ object DeploymentActor {
     Props(new DeploymentActor(
       deploymentManagerActor,
       killService,
```
zen-dog: Same goes for the `killService` here - I think _Scheduler_ can replace it. Ideally *all* of the deployment actor only use the new _Scheduler_

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -40,7 +40,10 @@ class TaskReplaceActor(
   // Killed resident tasks are not expunged from the instances list. Ignore
   // them. LaunchQueue takes care of launching instances against reservations
   // first
-  val currentInstances = instanceTracker.specInstancesSync(runSpec.id).filter(_.state.goal == Goal.Running)
+  val currentInstances = {
+    // TODO(karsten): Use InstanceTrackerConfig.internalTaskTrackerRequestTimeout or rather be async.
+    Await.result(scheduler.getInstances(runSpec.id), 1000.millis).filter(_.state.goal == Goal.Running)
```
zen-dog: Where are those magical `1000.millis` coming from?

zen-dog: Like the _TODO_ says, I'd rather move this code to the _async_ part of the `preStart` method and have it asynchronous.

src/main/scala/mesosphere/marathon/scheduling/Scheduler.scala
zen-dog: I agree that restarting is an Orchestrator concern

src/main/scala/mesosphere/marathon/core/deployment/impl/AppStartActor.scala
jeschkies: Most modules take `scheduler` and `schedulerActions`. The later was called `scheduler` before. Overall `scheduler` might not be a good name because it could be confused with the Mesos scheduler interface.

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
jeschkies: Yep, that's [phase 3](https://jira.mesosphere.com/browse/MARATHON-8292).

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
jeschkies: The will with [phase 3](https://jira.mesosphere.com/browse/MARATHON-8292). This diff could also move the kill service. Should I move it in this diff as well?

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
jeschkies: We can move it in this diff but then we have to introduce the `scheduler.create` or something that uses the launch queue. It is not trivial.

zen-dog: I'm not saying we should remove _LaunchQueue_ completely in this phase, but at least Deployment* actors could switch to using the new scheduler. Otherwise, this looks cumbersome and inconsistent. I'd be glad to help with this if you want to pair.

src/main/scala/mesosphere/marathon/core/deployment/impl/DeploymentActor.scala
jeschkies: I tried to move the launch queue but it's quite a mess https://github.com/mesosphere/marathon/commit/3c01924c32f2a1f6c26659f380aa524a3392411d. Since we are removing the queue in phase  it's not really worth the effort.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
timcharper: I'm inclined to just keep the specInstancesSync call here for now, since it seems not entirely necessary to change the sync nature of this approach as part of this PR.

src/main/scala/mesosphere/marathon/scheduling/LegacyScheduler.scala
```diff
@@ -0,0 +1,50 @@
...
```
timcharper: why is this called Legacycheduler?

src/main/scala/mesosphere/marathon/scheduling/Scheduler.scala
jeschkies: I don't think instances should ever change their run spec version (which does not work with reservations for now). So yes, it's an orchestrator concern to stop an instance and start a new one with a new version.

src/main/scala/mesosphere/marathon/scheduling/Scheduler.scala
jeschkies: To elaborate, `LaunchQueue.Add` does three things
* It looks for stopped instances and sets their goal to running again.
* It create instances if there are not enough stopped instances.
* It start the `TaskLauncherActor` for the runs spec.

That behaviour is not inline with the changes in https://github.com/mesosphere/marathon/pull/6340, ie phase 2. Phase 2 aims to enable a CRUD like interface. Instead of an add we want a `def create(runSpec)` and `updateGoal(runSpec, goal)`. Phase 3 is supposed to enable that by removing the add from the launch queue. However, that requires a refactoring of the `TaskLauncherActor` to a pure function or rather move the actor into `OfferMatcherManagerActor`.

src/main/scala/mesosphere/marathon/scheduling/LegacyScheduler.scala
```diff
@@ -0,0 +1,116 @@
...
+      await(instanceTracker.schedule(instancesToSchedule))
+
+      // Let launch queue manage task launcher actor.
+      await(launchQueue.sync(runSpec))
+
+      Done
+    }
+  }
+
+  override def reschedule(instance: Instance, runSpec: RunSpec)(implicit ec: ExecutionContext): Future[Done] = serializeUpdates {
```
alenkacz: as I wrote in other comments, I would actually remove this method, I think it's an implementation detail

src/main/scala/mesosphere/marathon/scheduling/LegacyScheduler.scala
```diff
@@ -0,0 +1,116 @@
...
+  override def getInstances(runSpecId: PathId)(implicit ec: ExecutionContext): Future[Seq[Instance]] = serializeUpdates {
+    instanceTracker.specInstances(runSpecId)
+  }
+
+  override def getInstance(instanceId: Instance.Id)(implicit ec: ExecutionContext): Future[Option[Instance]] = serializeUpdates {
+    instanceTracker.get(instanceId)
+  }
+
+  @SuppressWarnings(Array("all")) // async/await
+  override def run(instances: Seq[Instance])(implicit ec: ExecutionContext): Future[Done] = serializeUpdates {
```
alenkacz: I think we can drop this method because this operation is not supported. It's only supported by resident tasks and those have that handled differently

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -472,7 +472,13 @@ class SchedulerActions(
 
       if (toAdd > 0) {
         logger.info(s"Queueing $toAdd new instances for ${runSpec.id} to the already $leftToLaunch queued ones")
-        await(launchQueue.add(runSpec, toAdd))
+        // TODO(karsten): Use scheduler interface or instance tracker instead or get rid of the scale checks.
+        // TODO(karsten): Handle persistent instances correctly
+        val instancesToSchedule = 0.until(toAdd).map { _ => Instance.scheduled(runSpec, Instance.Id.forRunSpec(runSpec.id)) }
```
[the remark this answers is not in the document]
jeschkies: Yes. I believe this is why the tests still fail.

src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
```diff
@@ -47,16 +43,24 @@ trait StartingBehavior extends ReadinessBehavior with StrictLogging { this: Acto
...
       val instancesToStartNow = Math.max(scaleTo - actualSize, 0)
       logger.debug(s"Sync start instancesToStartNow=$instancesToStartNow appId=${runSpec.id}")
       if (instancesToStartNow > 0) {
         logger.info(s"Reconciling app ${runSpec.id} scaling: queuing $instancesToStartNow new instances")
-        await(launchQueue.add(runSpec, instancesToStartNow))
+        // Reschedule stopped resident instances first.
+        val existingReservedStoppedInstances = instances
```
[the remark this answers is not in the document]
jeschkies: I disagree. This was my main point against the interface. `Scheduler.schedule` is a `create` operation. `Scheduler.reschedule` is an `update` operation. The orchestrator should explicitly call either one. The `LaunchQueue.Add` would hide the difference. That's why I really do not want to expose that part.

src/main/scala/mesosphere/marathon/scheduling/LegacyScheduler.scala
jeschkies: It is *not* an implementation detail. There is a semantic difference between `reschedule` aka `update` and `schedule` aka `create`.

src/main/scala/mesosphere/marathon/core/deployment/impl/StartingBehavior.scala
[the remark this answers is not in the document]
alenkacz: oh I see your point... but it would be hard to do until we have event loop because all the places where you do this kind of reschedule will have to be synchronized (you can't have two threads doing that on the same runspec).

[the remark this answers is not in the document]
jeschkies: We always have only one orchestrator running these commands. Also, I used the `WorkQueue` to synchronize the calls. @zen-dog did not like it so we are figuring something else out.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -153,6 +153,44 @@ class TaskReplaceActor(
 
   }
 
+  private def launching: Receive = {
+    case Frame(instances) =>
```
jeschkies: This is just a prototype idea. There some code duplication to `launchInstances`. @timcharper, how would you manipulate the frame? I don't like that I call `scheduler.reschedule` etc.

timcharper:
@jeschkies I tried to rebase but was getting a ridiculous amount of conflicts; I squashed them and rebased and now the patch is on latest master. I pushed a copy of your old branch here in case I messed something up somehow: https://github.com/mesosphere/marathon/tree/karsten/scheduler-interface-pre-tim-conflict-resolve

jeschkies:
Superseded by other efforts.

PR #6394, opened 2018-07-30
Properly start TaskLauncherActor after failover

alenkacz:
JIRA issues: MARATHON-8321


src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -72,7 +75,42 @@ private[impl] class LaunchQueueActor(
...
+    case instances: InstanceTracker.InstancesBySpec =>
+
+      val scheduledRunSpecs: Iterable[PathId] = instances.instancesMap.filter {
+        case (_, specInstances) =>
+          specInstances.instances.exists(_.isScheduled)
+      }.keys
+
+      scheduledRunSpecs.flatMap(groupManager.runSpec(_)).foreach { scheduledRunSpec =>
+        val actorRef = launchers.getOrElse(scheduledRunSpec.id, createAppTaskLauncher(scheduledRunSpec))
+        actorRef ! TaskLauncherActor.Sync(scheduledRunSpec)
```
alenkacz: do we really need the sync? Since it's LaunchQueueActor starting, it will always be new TaskLauncherActor, right? So it already has the newest runspec from the line above... ?

meichstedt: I think we do need it but I'm not sure how this was done before. The actor
- is started
- fetches state from instanceTracker
- requests info from the RateLimiter
- receives a DelayUpdate (for the current runSpec)
- unstashes all messages and becomes active

Now in order to register as an offer matcher, `manageOfferMatcherStatus()` has to be called. This is done upon receiving that `Sync`. You can't call `manageOfferMatcherStatus()` before having received the `DelayUpdate`, and I wouldn't want to do it in `waitForInitialDelay`. Once the actor became active, it needs a trigger to get going though.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -72,7 +75,41 @@ private[impl] class LaunchQueueActor(
...
+  def initializing: Receive = {
+    case instances: InstanceTracker.InstancesBySpec =>
+
+      val scheduledRunSpecs: Iterable[PathId] = instances.instancesMap.filter {
+        case (_, specInstances) =>
+          specInstances.instances.exists(_.isScheduled)
+      }.keys
+
+      scheduledRunSpecs.flatMap(groupManager.runSpec(_)).foreach { scheduledRunSpec =>
+        val actorRef = launchers.getOrElse(scheduledRunSpec.id, createAppTaskLauncher(scheduledRunSpec))
```
timcharper: this val looks unused

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -72,7 +75,41 @@ private[impl] class LaunchQueueActor(
...
+
+  override def receive: Receive = initializing
+
+  def initializing: Receive = {
+    case instances: InstanceTracker.InstancesBySpec =>
+
+      val scheduledRunSpecs: Iterable[PathId] = instances.instancesMap.filter {
+        case (_, specInstances) =>
+          specInstances.instances.exists(_.isScheduled)
+      }.keys
```
timcharper: Maybe this is better?

```
val scheduledRunSpecs: Iterable[PathId] = instances.instancesMap
  .collect {
    case (key, specInstances) if specInstances.instances.exists(_.isScheduled) =>
      groupManager.runSpec(key)
  }
  .flatten
  .foreach { scheduledRunSpec =>
    ...
  }
```

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -72,7 +75,41 @@ private[impl] class LaunchQueueActor(
...
+
+      context.become(initialized)
+
+      unstashAll()
+
+    case Status.Failure(cause) =>
+      // escalate this failure
+      throw new IllegalStateException("while loading instances", cause)
+
+    case stashMe: AnyRef =>
```
timcharper: I think I'd just use `case _ =>`

PR #6395, opened 2018-07-31
Convert offersWanted logic to Akka streams

timcharper:
Remove RX dependency

JIRA Issues: MARATHON-8308

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/ReadinessBehavior.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/ReadinessBehavior.scala
@@ -3,7 +3,7 @@ package core.deployment.impl
 
 import akka.actor.{Actor, ActorRef, Cancellable}
 import akka.stream.ActorMaterializer
-import akka.stream.scaladsl.{ Keep, Sink }
+import akka.stream.scaladsl.{Keep, Sink}
 import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.condition.Condition.Running
 import mesosphere.marathon.core.deployment.impl.DeploymentManagerActor.ReadinessCheckUpdate
--- a/src/main/scala/mesosphere/marathon/core/deployment/impl/ReadinessBehavior.scala
+++ b/src/main/scala/mesosphere/marathon/core/deployment/impl/ReadinessBehavior.scala
@@ -79,8 +79,7 @@ trait ReadinessBehavior extends StrictLogging { this: Actor =>
     healthy -= instanceId
     ready -= instanceId
     subscriptions.keys.withFilter(_.taskId.instanceId == instanceId).foreach { key =>
-      subscriptions(key).cancel()
-      subscriptions -= key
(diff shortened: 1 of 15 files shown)


src/main/scala/mesosphere/marathon/core/deployment/impl/ReadinessBehavior.scala
```diff
@@ -79,8 +79,7 @@ trait ReadinessBehavior extends StrictLogging { this: Actor =>
     healthy -= instanceId
     ready -= instanceId
     subscriptions.keys.withFilter(_.taskId.instanceId == instanceId).foreach { key =>
-      subscriptions(key).cancel()
-      subscriptions -= key
+      subscriptions.get(key).foreach(_.cancel())
```
timcharper: This fixes a race introduced by #6393

timcharper: (Subscription is removed automatically if the stream terminates)

src/main/scala/mesosphere/marathon/stream/EnrichedFlow.scala
```diff
@@ -51,4 +51,59 @@ object EnrichedFlow {
...
+    *
+    * Will output:
+    *
+    *     1 a 0.5
+    *     2 a 1.0
+    *     2 b 1.5
+    *     3 b 2.0
+    *     3 c 2.5
+    */
+  def combineLatest[T] = new CombineLatest[T]
```
meln1k: I probably miss something, but what's the difference between this stage and `zipWith`?

src/main/scala/mesosphere/marathon/core/matcher/manager/impl/OfferMatcherManagerActor.scala
```diff
@@ -164,7 +164,7 @@ private[impl] class OfferMatcherManagerActor private (
   }
 
   def offersWanted: Boolean = matchers.nonEmpty && launchTokens > 0
-  def updateOffersWanted(): Unit = offersWantedObserver.onNext(offersWanted)
+  def updateOffersWanted(): Unit = offersWantedInput.offer(offersWanted)
```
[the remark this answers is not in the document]
timcharper: My opinion is no. The data we are transmitting are booleans and I see no possible way that it will back up. Setting the buffer mode to back pressure is effectively the same as an actor tell - unbounded queue in the end.

src/main/scala/mesosphere/marathon/stream/EnrichedFlow.scala
timcharper: zipWith emits n-aligned tuples. This component will emit a tuple each time the right or left stream emits, repeating the last value for the non-emitting stream.

zen-dog: It's already merged, but `zipWithLatest` might've been a better name giving someone who is familiar with akka a hint of what it does?

PR #6400, opened 2018-08-01
Remove LegacyAppInstance

alenkacz:
It is not used in the codebase, just in test. Just a little bit of cleanup...

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/appinfo/TaskCountsTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/appinfo/TaskCountsTest.scala
@@ -188,9 +188,15 @@ class TaskCountsTest extends UnitTest {
 
 object Fixture {
   implicit class TaskImplicits(val task: Task) extends AnyVal {
-    def toInstance: Instance = LegacyAppInstance(
-      task, AgentInfo(host = "host", agentId = Some("agent"), region = None, zone = None, attributes = Nil),
-      unreachableStrategy = UnreachableStrategy.default())
+    def toInstance: Instance = {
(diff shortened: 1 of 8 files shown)


src/test/scala/mesosphere/marathon/core/appinfo/TaskCountsTest.scala
```diff
@@ -188,9 +188,15 @@ class TaskCountsTest extends UnitTest {
...
-      task, AgentInfo(host = "host", agentId = Some("agent"), region = None, zone = None, attributes = Nil),
-      unreachableStrategy = UnreachableStrategy.default())
+    def toInstance: Instance = {
+      val agentInfo = AgentInfo(host = "host", agentId = Some("agent"), region = None, zone = None, attributes = Nil)
+      val unreachableStrategy = UnreachableStrategy.default()
+      val since = task.status.startedAt.getOrElse(task.status.stagedAt)
+      val tasksMap = Map(task.taskId -> task)
+      val state = Instance.InstanceState(None, tasksMap, since, unreachableStrategy)
+
+      new Instance(task.taskId.instanceId, agentInfo, state, tasksMap, task.runSpecVersion, unreachableStrategy, None)
```
zen-dog: nit here and elsewhere: It's a matter of taste, I guess, but in such cases, where there is little or no calculations, I prefer a multi-line constructor call with parameter names instead of `val`s e.g.:
```
new Instance(
  instanceId = task.taskId.instanceId,
  agentInfo = AgentInfo(host = "host", agentId = Some("agent"), region = None, zone = None, attributes = Nil),
  unreachableStrategy = UnreachableStrategy.default(),
  ...
  )
```

PR #6402, opened 2018-08-02
Synchronize python dependencies before running SI tests

zen-dog:
Summary:
we use `pipenv sync` to synchronize python dependencies defined in the Pipfile during `make init` prior to running SI tests.

### DIFF
--- a/tests/system/Makefile
+++ b/tests/system/Makefile
@@ -14,7 +14,7 @@ help:
 
 init:
 	pip3 install pipenv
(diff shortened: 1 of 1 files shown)


tests/system/Makefile
```diff
@@ -14,7 +14,7 @@ help:
 
 init:
 	pip3 install pipenv
-	pipenv install --dev --skip-lock
```
meln1k: Is it equal?

PR #6403, opened 2018-08-02
Correct description of the missing network name migration validations.

meln1k:


ichernetsky:
I believe it should be backported too.

PR #6404, opened 2018-08-02
Add "wipe=true" support for pod's instances endpoint, as it is implem…

meln1k:
…ented for app's tasks

Summary: added api flag and tests.

JIRA issues: MARATHON-8326

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
@@ -892,7 +892,7 @@ class PodsResourceTest extends AkkaUnitTest with Mockito with JerseyTest {
           )
           killer.kill(any, any, any)(any) returns Future.successful(Seq(instance))
           val response = asyncRequest { r =>
-            f.podsResource.killInstance("/id", instance.instanceId.idString, f.auth.request, r)
+            f.podsResource.killInstance("/id", instance.instanceId.idString, false, f.auth.request, r)
           }
           withClue(s"response body: ${response.getEntity}") {
             response.getStatus should be(HttpServletResponse.SC_OK)
@@ -920,7 +920,7 @@ class PodsResourceTest extends AkkaUnitTest with Mockito with JerseyTest {
           killer.kill(any, any, any)(any) returns Future.successful(instances)
           val response = asyncRequest { r =>
             f.podsResource.killInstances(
-              "/id", Json.stringify(Json.toJson(instances.map(_.instanceId.idString))).getBytes, f.auth.request, r)
+              "/id", false, Json.stringify(Json.toJson(instances.map(_.instanceId.idString))).getBytes, f.auth.request, r)
           }
           withClue(s"response body: ${response.getEntity}") {
             response.getStatus should be(HttpServletResponse.SC_OK)
--- a/tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
+++ b/tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
@@ -20,11 +20,20 @@ import scala.concurrent.duration._
 
 class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonTest {
 
+  override lazy val mesosNumMasters = 1
+  override lazy val mesosNumSlaves = 2
+
   // Configure Mesos to provide the Mesos containerizer with Docker image support.
   override lazy val mesosConfig = MesosConfig(
     launcher = "linux",
     isolation = Some("filesystem/linux,docker/runtime"),
     imageProviders = Some("docker"))
+
+  before {
+    mesosCluster.agents.head.start()
+    mesosCluster.agents(1).stop()
+    mesosCluster.waitForLeader().futureValue
+  }
 
   "MesosApp" should {
     "deploy a simple Docker app using the Mesos containerizer" taggedAs WhenEnvSet(envVarRunMesosTests, default = "true") in {
@@ -384,6 +393,50 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
       }
     }
 
+    "wipe pod instances with persistent volumes" taggedAs WhenEnvSet(envVarRunMesosTests, default = "true") in {
+      Given("a pod with persistent volumes")
+      val pod = residentPod("resident-pod-with-once-instance-wipe").copy(
+        instances = 1
+      )
+
+      When("The pod is created")
+      val createResult = marathon.createPodV2(pod)
+      createResult should be(Created)
+      waitForDeployment(createResult)
+      eventually { marathon.status(pod.id) should be(Stable) }
+
+      Then("1 instance should be running")
+      val status = marathon.status(pod.id)
+      status should be(OK)
+      status.value.instances should have size 1
+      val agentId = status.value.instances.head.agentId.get
+
+      When("An instance is unreachable")
+      mesosCluster.agents(0).stop()
+      waitForStatusUpdates("TASK_UNREACHABLE")
+
+      And("Pods instance is deleted")
+      val instanceId = status.value.instances.head.id
+      val deleteResult = marathon.deleteInstance(pod.id, instanceId, wipe = true)
+      deleteResult should be(OK)
+
+      Then("pod instance is erased from marathon's knowledge ")
+      val statusAfter = marathon.status(pod.id)
+      statusAfter.value.instances.find(_.id == instanceId) shouldEqual None
+
+      And("a new pod with a new persistent volume is scheduled")
+      waitForStatusUpdates("TASK_RUNNING")
+      statusAfter.value.instances should have size 1
+
+      When("the task associated with pod becomes reachable again")
+      mesosCluster.agents(0).start()
+
+      Then("Marathon kills the task and removes the associated reservation and volume")
+      waitForStatusUpdates("TASK_KILLED")
+      mesos.state.value.agents.find(_.id == agentId).get.reservedResourcesByRole.isEmpty shouldEqual true
+
+    }
+
     "deploy a simple pod with unique constraint and then " taggedAs WhenEnvSet(envVarRunMesosTests, default = "true") in {
 
       val constraints = Set(
@@ -428,8 +481,8 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
       And("Size of the pod should still be 1")
       val status2 = marathon.status(pod.id)
       status2 should be(OK)
-      //we have only one agent by default, so we expect one instance to be running.
-      status2.value.instances should have size 1
+      //we have only two agents by default, so we expect one instance to be running.
+      status2.value.instances should have size 2
 
     }
   }
--- a/src/main/scala/mesosphere/marathon/MarathonConf.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonConf.scala
@@ -282,7 +282,6 @@ trait MarathonConf
     noshort = true,
     hidden = true,
     default = Some(MesosHeartbeatMonitor.DEFAULT_HEARTBEAT_FAILURE_THRESHOLD))
-
   lazy val drainingSeconds = opt[Long](
     "draining_seconds",
     descr = "(Default: 0 seconds) the seconds when marathon will start declining offers before a maintenance " +
--- a/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
@@ -267,6 +267,7 @@ class PodsResource @Inject() (
   def killInstance(
     @PathParam("id") idOrig: String,
     @PathParam("instanceId") instanceId: String,
+    @DefaultValue("false")@QueryParam("wipe") wipe: Boolean,
     @Context req: HttpServletRequest,
     @Suspended asyncResponse: AsyncResponse): Unit = sendResponse(asyncResponse) {
     async {
@@ -282,7 +283,7 @@ class PodsResource @Inject() (
       validateOrThrow(id)
       validateOrThrow(instanceId)
       val parsedInstanceId = Instance.Id.fromIdString(instanceId)
-      val instances = await(taskKiller.kill(id, _.filter(_.instanceId == parsedInstanceId)))
+      val instances = await(taskKiller.kill(id, _.filter(_.instanceId == parsedInstanceId), wipe))
       instances.headOption.fold(unknownTask(instanceId))(instance => ok(jsonString(instance)))
     }
   }
@@ -292,6 +293,7 @@ class PodsResource @Inject() (
   @Path("""{id:.+}::instances""")
   def killInstances(
     @PathParam("id") idOrig: String,
+    @DefaultValue("false")@QueryParam("wipe") wipe: Boolean,
     body: Array[Byte],
     @Context req: HttpServletRequest,
     @Suspended asyncResponse: AsyncResponse): Unit = sendResponse(asyncResponse) {
@@ -314,7 +316,7 @@ class PodsResource @Inject() (
       def toKill(instances: Seq[Instance]): Seq[Instance] = {
         instances.filter(instance => instancesDesired.contains(instance.instanceId))
       }
-      val instances = await(taskKiller.kill(id, toKill))
+      val instances = await(taskKiller.kill(id, toKill, wipe))
       ok(Json.toJson(instances))
     }
   }
--- a/tests/integration/src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala
+++ b/tests/integration/src/test/scala/mesosphere/marathon/integration/facades/MarathonFacade.scala
@@ -289,9 +289,9 @@ class MarathonFacade(
     result(requestFor[List[PodInstanceStatus]](Delete(s"$url/v2/pods$podId::instances")), waitTime)
   }
 
-  def deleteInstance(podId: PathId, instance: String): RestResult[PodInstanceStatus] = {
+  def deleteInstance(podId: PathId, instance: String, wipe: Boolean = false): RestResult[PodInstanceStatus] = {
     requireInBaseGroup(podId)
-    result(requestFor[PodInstanceStatus](Delete(s"$url/v2/pods$podId::instances/$instance")), waitTime)
+    result(requestFor[PodInstanceStatus](Delete(s"$url/v2/pods$podId::instances/$instance?wipe=$wipe")), waitTime)
   }
 
   //apps tasks resource --------------------------------------


kensipe:
rebased

build.sbt
```diff
@@ -59,8 +59,8 @@ lazy val integrationTestSettings = Seq(
       "-u", "target/test-reports", // TODO(MARATHON-8215): Remove this line
       "-o", "-eDFG",
       "-y", "org.scalatest.WordSpec")),
-  parallelExecution in Test := true,
-  testForkedParallel in Test := true,
+  parallelExecution in Test := false,
```
jeschkies: What is the motivation for this change?

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
```diff
@@ -97,10 +97,11 @@ private[impl] class KillServiceActor(
     processKills()
   }
 
+  //todo test this
```
jeschkies: Why not add the test right now?

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
```diff
@@ -36,8 +43,9 @@ object KillStreamWatcher extends StrictLogging {
         classOf[UnknownInstanceTerminated], eventStream, bufferSize, overflowStrategy).map(_.id)
 
     // eagerComplete allows us to toss the right materialized cancellable value
-    killedViaInstanceChanged.
-      merge(killedViaUnknownInstanceTerminated, eagerComplete = true)
+    val liveInstanceIds = killedViaInstanceChanged.merge(killedViaUnknownInstanceTerminated, eagerComplete = true)
+
+    Source(alreadyConsideredTerminal).concatMat(liveInstanceIds)(Keep.right)
```
jeschkies: What is this doing? Why not simply filter the `liveInstanceIds` with `considerTerminal`?

jeschkies:
Nit: The title states that this is still a work in progress. Is this the cases? What do you thinkg about using the `wip` label and removing it once the diff is ready for reviewing?

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
alenkacz: either test it or drop the log line (with strong preference to the first choice). I really don't think these TODOs are helpful in the code in long run...

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
```diff
@@ -22,10 +23,16 @@ object KillStreamWatcher extends StrictLogging {
     *
     * See [[mesosphere.marathon.stream.EnrichedSource$.eventBusSource]]
     */
-  private def killedInstanceIds(eventStream: akka.event.EventStream): Source[Instance.Id, Cancellable] = {
+  private def killedInstanceIds(eventStream: akka.event.EventStream, instancesToCheck: Iterable[Instance]): Source[Instance.Id, Cancellable] = {
     // MaxValue causes a dynamically sized buffer to be used.
     val bufferSize = Int.MaxValue
     val overflowStrategy = OverflowStrategy.fail
+
+    val alreadyConsideredTerminal = instancesToCheck
```
alenkacz: why was this change needed?

alenkacz: I mean how come we did not need this for apps but need it for pods?

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -12,20 +12,39 @@ import mesosphere.marathon.raml.{App, Container, DockerContainer, EngineType}
...
+    // We need to start all the agents for the teardown to be able to kill all the (UNREACHABLE) executors/tasks
+    mesosCluster.agents.foreach(_.start())
+    eventually { mesosCluster.state.value.agents.size shouldBe mesosCluster.agents.size }
+    super.afterAll()
+  }
+
+
+  override def beforeAll(): Unit = {
+    super.beforeAll()
+    mesosCluster.agents(1).stop()
```
alenkacz: should we also make sure that agent(0) is started? this seems dangerous, because we are not sure that at least one agent is always running...

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -384,6 +403,98 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+      waitForStatusUpdates("TASK_RUNNING")
+      marathon.status(pod.id).value.instances should have size 1
+
+      When("the task associated with pod becomes reachable again")
+      mesosCluster.agents(0).start()
+
+      Then("Marathon kills the task and removes the associated reservation and volume")
+      waitForStatusUpdates("TASK_FAILED")
+
+      mesosCluster.agents(1).stop()
```
alenkacz: this will be already ensured by before

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -384,6 +403,98 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+        marathon.status(pod.id).value.instances should have size 1
+      }
+
+      When("the task associated with pod becomes reachable again")
+      mesosCluster.agents(0).start()
+
+      Then("Marathon kills the task")
+      waitForStatusUpdates("TASK_FAILED")
+
+      mesosCluster.agents(1).stop()
```
alenkacz: this will be already ensured by before

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -428,7 +539,7 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
       And("Size of the pod should still be 1")
       val status2 = marathon.status(pod.id)
       status2 should be(OK)
-      //we have only one agent by default, so we expect one instance to be running.
+      //we have only two agents by default, so we expect one instance to be running.
```
alenkacz: I think one agent was correct

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
meln1k: If we're waiting for the unreachable instance, we will have to wait until those instances will be reconciled, so it will be ~10 minutes in the worst case. That optimisation reduces the waiting time if the instance is already unreachable.

jeschkies: I'm sorry. I'm slow today and don't get what this line is doing. How are `alreadyConsideredTerminal` and `liveInstanceIds` related? Isn't `concatMat` just chaining both?

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
meln1k: We didn't test apps use case. We need this to not block and wait 10 minutes for the reconciliation if the instance is already terminal (for example in case of unreachable).

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
meln1k: Yes. Before subscribing to the live stream we push what we already know. It covers the case where we were watching already unreachable instance.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: `beforeAll` has a different semantics compared to `before` and runs only once.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: `agent(0)` is started in MesosClusterTest

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
jeschkies: I still don't get it 😄. What kind of instance ids are included in `alreadyConsideredTerminal` and `liveInstanceIds`? If `liveInstanceIds` is a subset I don't understand why we concatenate them.

You are right about the advantage but it seems ti complicated things for the sake of early optimization. Come to think it is odd the we wait for an event. It might be a race condition. Also, if an instance is already terminated why should we wait at all?

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: `beforeAll` has a different semantics compared to `before` and runs only once.

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
meln1k: > What kind of instance ids are included in alreadyConsideredTerminal and liveInstanceIds?

`alreadyConsideredTerminal` and `liveInstanceIds` produce the same instanceIds.

> If liveInstanceIds is a subset I don't understand why we concatenate them.

Consider the following case: If the instance is already in the terminal state at the moment of watching, then we won't see any updates in the `liveInstancesIds` except for the updates from the reconciliation process (which might take up to 10 minutes to arrive). That's why we concatenate the live stream with the id of instances which are already known to be in the terminal state.

> You are right about the advantage but it seems ti complicated things for the sake of early optimization.

If we don't do this, users will have to wait up to 10 minutes before the delete instance call will give a response back, which will result in request timeout and make marathon users unhappy. For me it doesn't look like `for the sake of early optimization.`

> Also, if an instance is already terminated why should we wait at all?

That's the whole point for this change: if the instance is already terminated we don't wait for the reconciliation to happen and immediately push instance ids downstream.

jeschkies: You convinced me 🙂. Now I just have to understand why we `merge` and then `concat`. That is odd to me 🤔

meln1k: as well as cancellable is materialized properly - it's fine :)

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
```diff
@@ -22,10 +23,16 @@ object KillStreamWatcher extends StrictLogging {
...
+      .filter(instance => considerTerminal(instance.state.condition))
```
alenkacz: in KillService.killInstances we use "isTerminal" to filter. Shouldn't we filter by the same thing? I think that considerTerminal and isTerminal is not exactly the same set of conditions

alenkacz: I mean we do this `instance.tasksMap.values.withFilter(!_.isTerminal)` so maybe we should also in KillServiceActor do `instance.filter(i => i.tasksMap.values.exists(!_.isTerminal))` and then use these instance ids when passing ids to KillStreamWatcher. 

Thanks to that we filter the same thing in the same place once instead of twice on two different places.

WDYT?

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -12,14 +12,30 @@ import mesosphere.marathon.raml.{App, Container, DockerContainer, EngineType}
...
+  override lazy val mesosNumSlaves = 2
+
+  override def afterAll(): Unit = {
+    // We need to start all the agents for the teardown to be able to kill all the (UNREACHABLE) executors/tasks
+    mesosCluster.agents.foreach(_.start())
+    eventually { mesosCluster.state.value.agents.size shouldBe mesosCluster.agents.size }
+    super.afterAll()
+  }
+
+  override def beforeAll(): Unit = {
```
alenkacz: maybe we can instead of beforeAll we can do before with
mesosCluster.agents(0).start()
mesosCluster.agents(1).stop()

it won't do anything if we are already in this state but it will centralize the cleaning logic to one place making sure all tests start in the same state. WDYT?

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
jeschkies: I think
```scala
Source(alreadyConsideredTerminal)
  .merge(killedViaInstanceChanged)
  .merge(killedViaUnknownInstanceTerminated)
```

is easier to follow.

src/test/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcherTest.scala
```diff
@@ -40,4 +44,28 @@ class KillStreamWatcherTest extends AkkaUnitTest {
...
   }
+
+  "KillStreamWatcher emits already terminated instances" in {
+
+    val empty = MarathonTestHelper.emptyInstance()
+    val unreachableInstance = empty.copy(state = empty.state.copy(condition = Condition.Unreachable))
+
+    val watcher = KillStreamWatcher.watchForKilledInstances(system.eventStream, List(unreachableInstance))
+
+    watcher.runWith(Sink.head).futureValue shouldEqual Done
```
jeschkies: Why not test `killedInstanceIds` here?

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -12,14 +12,30 @@ import mesosphere.marathon.raml.{App, Container, DockerContainer, EngineType}
...
 import scala.collection.immutable.Seq
 import scala.concurrent.duration._
 
 class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonTest {
 
+  override lazy val mesosNumMasters = 1
+  override lazy val mesosNumSlaves = 2
+
+  override def afterAll(): Unit = {
+    // We need to start all the agents for the teardown to be able to kill all the (UNREACHABLE) executors/tasks
```
jeschkies: I'd much rather move unreachable related tests to `UnrechableIntegrationTest`.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -384,6 +400,98 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+      waitForStatusUpdates("TASK_UNREACHABLE")
+
+      And("Pods instance is deleted")
+      val instanceId = status.value.instances.head.id
+      val deleteResult = marathon.deleteInstance(pod.id, instanceId, wipe = true)
+      deleteResult should be(OK)
+
+      Then("pod instance is erased from marathon's knowledge ")
+      val statusAfter = marathon.status(pod.id)
+      statusAfter.value.instances.find(_.id == instanceId) shouldEqual None
```
jeschkies: How about:
```scala
val knownInstanceIds = marathon.status(pod.id).value.instances.map(_.id) 
knownInstanceIds should not contain(instanceId)
```

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
```diff
@@ -384,6 +400,98 @@ class MesosAppIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathonT
...
+
+      And("a new pod with a new persistent volume is scheduled")
+      waitForStatusUpdates("TASK_RUNNING")
+      marathon.status(pod.id).value.instances should have size 1
+
+      When("the task associated with pod becomes reachable again")
+      mesosCluster.agents(0).start()
+
+      Then("Marathon kills the task and removes the associated reservation and volume")
+      waitForStatusUpdates("TASK_FAILED")
```
jeschkies: Why does it fail?

tests/integration/src/test/scala/mesosphere/marathon/integration/SequentialIntegrationTests.scala
```diff
@@ -0,0 +1,12 @@
+package mesosphere.marathon
+package integration
+
+import org.scalatest.Sequential
+
+/**
+  * Wrapper that prohibits parallel test run of the respective test suites.
+  */
+class SequentialIntegrationTests extends Sequential(
```
jeschkies: Why is this required? I'm not a big face of reintroducing it.

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcher.scala
meln1k: If you look at `killedViaInstanceChanged` source, you'll see that it also uses `considerTerminal` method. I'd keep it to make everything consistent.

src/test/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcherTest.scala
meln1k: Test of `watchForKilledInstances` already covers `killedInstanceIds`.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: I'm not so sure that we should do it. This test is more related to `wipe=true` flag rather than pod unreachable behaviour.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: Let's try this :)

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
zen-dog: Agree with @jeschkies here, `TaskUnreachableIntegrationTest` already has this agent on/off logic, and we should try to keep it at one test

tests/integration/src/test/scala/mesosphere/marathon/integration/SequentialIntegrationTests.scala
meln1k: It's a temporary solution to make test that interfering each other pass. We should tackle this in the follow-up.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: Well, semantically it belongs to pods-related tests.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
meln1k: Task fails because the agent was unreachable, and then it re-registred.

jeschkies: Really? Do we publish unreachable as `TASK_FAILED`? That is odd.

src/test/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcherTest.scala
jeschkies: Yeah but the assertion is not giving much details.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
jeschkies: That is true but messing with the number of agents etc should be contained to one test IMHO. Also, the test don't really cover just the wipe but unreachable tasks. If we'd only test that a persistent pod which writes `hello\n` to a volume is wiped and restarted should not contain `hello\nhello\n` but `hello\n`.

tests/integration/src/test/scala/mesosphere/marathon/integration/MesosAppIntegrationTest.scala
zen-dog: Indeed, why is it `TASK_FAILED`?

build.sbt
```diff
@@ -7,7 +7,6 @@ import com.typesafe.sbt.packager.docker.Cmd
 import mesosphere.maven.MavenSettings.{loadM2Credentials, loadM2Resolvers}
 import mesosphere.raml.RamlGeneratorPlugin
 import NativePackagerHelper.directory
-
```
zen-dog: Unnecessary change

src/test/scala/mesosphere/marathon/api/v2/PodsResourceTest.scala
```diff
@@ -892,7 +892,7 @@ class PodsResourceTest extends AkkaUnitTest with Mockito with JerseyTest {
           )
           killer.kill(any, any, any)(any) returns Future.successful(Seq(instance))
           val response = asyncRequest { r =>
-            f.podsResource.killInstance("/id", instance.instanceId.idString, f.auth.request, r)
+            f.podsResource.killInstance("/id", instance.instanceId.idString, false, f.auth.request, r)
```
zen-dog: `false` is default anyway?

src/test/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcherTest.scala
```diff
@@ -40,4 +44,28 @@ class KillStreamWatcherTest extends AkkaUnitTest {
 
     result.futureValue shouldBe Nil
   }
+
+  "KillStreamWatcher emits already terminated instances" in {
+
+    val empty = MarathonTestHelper.emptyInstance()
+    val unreachableInstance = empty.copy(state = empty.state.copy(condition = Condition.Unreachable))
```
zen-dog: Make it e.g. `Condition.Failed` to prevent confusion with unreachable?

src/test/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcherTest.scala
zen-dog: `withClue("Watcher should emit instantly emit instances that are already terminal")`?

src/test/scala/mesosphere/marathon/core/task/termination/impl/KillStreamWatcherTest.scala
```diff
@@ -40,4 +44,28 @@ class KillStreamWatcherTest extends AkkaUnitTest {
...
+
+    val empty = MarathonTestHelper.emptyInstance()
+    val runnningInstance = empty.copy(state = empty.state.copy(condition = Condition.Running))
+
+    val watcher = KillStreamWatcher.watchForKilledInstances(system.eventStream, List(runnningInstance))
+
+    watcher.runWith(Sink.head).onComplete(_ => probe.ref ! Done)
+    probe.expectNoMessage(100.millis)
+  }
+
```
zen-dog: One more test with one _Failed_ and one _Running_ instance making sure a _Failed_ instance is emited instantly?

PR #6405, opened 2018-08-03
Remove scapegoat from Marathon

timcharper:
I submit that Scapegoat provides less value than it costs to use. It is slow to run, produces lots of warnings, fails horribly with async/await (something that has been unaddressed for several years, leading to rampant SuppressWarnings statements). It is difficult to remember a time when it made a helpful suggestion at an error level. I submit does not catch the things that are actually important.


timcharper:
See https://github.com/mesosphere/marathon/search?q=SuppressWarnings&unscoped_q=SuppressWarnings

ichernetsky:
What about removing those `SuppressWarnings`?

timcharper:
Removing SuppressWarnings will be good. Preference to get consensus first.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -169,7 +169,6 @@ class MarathonSchedulerActor private (
       deploy(sender(), cmd)
 
     case cmd @ KillTasks(runSpecId, tasks) =>
-      @SuppressWarnings(Array("all")) /* async/await */
```
jeschkies: You rock! When I read the title I was wondering whether you removed there or not.

PR #6406, opened 2018-08-06
Replace Dropwizard and Kamon with Deprecated and New in RAML types

ichernetsky:


docs/docs/rest-api/public/api/v2/types/newMetrics.raml
```diff
@@ -1,18 +1,18 @@
 #%RAML 1.0 Library
 types:
-  DropwizardCounter:
```
alenkacz: do we have to have the prefix New? Why not just Meter, Counter, Gauge... ?

timcharper: ➕ 1

timcharper: What's the motivation for renaming these? I think I would prefer DropwardCounter and KamonCounter over OldCounter and NewCounter.

timcharper: (or, drop the `New`)

alenkacz: I think it was motivated by the fact that it was suggested to have library independent API types (finally) instead of one directly connected to the implementation. So I guess in the end it boiled down to renaming... I actually like the change.

ichernetsky: Alena and Ken suggested not to expose implementation details via RAML types. I won't be able to get rid of `New` everywhere though. Is it fine with you?

ichernetsky: Guys?

alenkacz: where is it not possible to drop new?

ichernetsky: Because
```
#%RAML 1.0 Library
uses:
  newMetrics: newMetrics.raml
  deprecatedMetrics: deprecatedMetrics.raml
types:
  Metrics: deprecatedMetrics.DeprecatedMetrics | newMetrics.NewMetrics
```
We can't have two `metrics.raml` files, and two classed both named`Metrics`.

PR #6412, opened 2018-08-06
Crash if ZK DNS is not resolvable

zen-dog:
Summary:
Currently, if Marathon can't connect to ZK because its name is resolvable, Guice will reinstantiate `StorageModule` N times, effectively retrying to connect to ZK N times. Marathon will eventually exit with exitCode=137 but it takes ~5min with default settings. With this change, we crash on the first error.

Related JIRA: DCOS-39846, COPS-3593, MARATHON-7390

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -153,7 +153,8 @@ class CoreModuleImpl @Inject() (
   override lazy val storageModule = StorageModule(
     metricsModule.metrics,
     marathonConf,
-    lifecycleState)(
+    lifecycleState,
+    crashStrategy)(
       actorsModule.materializer,
       storageExecutionContext,
       actorSystem.scheduler,
--- a/src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
+++ b/src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
@@ -3,13 +3,13 @@ package core.storage.store.impl.zk
 
 import akka.Done
 import akka.util.ByteString
+import com.typesafe.scalalogging.StrictLogging
 import mesosphere.marathon.core.base.{LifecycleState, _}
 import mesosphere.marathon.stream.Implicits._
-import org.apache.curator.RetryPolicy
 import org.apache.curator.framework.api.{BackgroundPathable, Backgroundable, Pathable}
 import org.apache.curator.framework.imps.CuratorFrameworkState
 import org.apache.curator.framework.state.{ConnectionState, ConnectionStateListener}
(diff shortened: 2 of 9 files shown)


src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -88,13 +88,25 @@ class CoreModuleImpl @Inject() (
...
+      lifecycleState,
+      crashStrategy)(
+      actorsModule.materializer,
+      storageExecutionContext,
+      actorSystem.scheduler,
+      actorSystem)
+  }
+
+  override def storageModule = _storageModule.get
+  //    try {
```
jeschkies: Is this a left over?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
```diff
@@ -147,12 +147,12 @@ class RichCuratorFramework(val client: CuratorFramework) {
...
     */
   @SuppressWarnings(Array("CatchFatal"))
-  def blockUntilConnected(lifecycleState: LifecycleState): Unit = {
+  def blockUntilConnected(lifecycleState: LifecycleState, crashStrategy: CrashStrategy): Unit = {
     if (!lifecycleState.isRunning)
       throw new InterruptedException("Not waiting for connection to zookeeper; Marathon is shutting down")
 
-    if (!client.blockUntilConnected(client.getZookeeperClient.getConnectionTimeoutMs, java.util.concurrent.TimeUnit.MILLISECONDS))
-      throw new InterruptedException("Timed out while waiting for zookeeper connection")
+    if (!client.blockUntilConnected(100, java.util.concurrent.TimeUnit.MILLISECONDS))
```
jeschkies: What is the motivation to hardcode 100 instead of using `client.getZookeeperClient.getConnectionTimeoutMs`?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
```diff
@@ -147,12 +147,12 @@ class RichCuratorFramework(val client: CuratorFramework) {
...
+      crashStrategy.crash()
```
jeschkies: What do you think about logging an error message?

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
zen-dog: yep 👍

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/RichCuratorFramework.scala
zen-dog: I was just testing if you're paying attention - you passed with flying colors! 🎉

PR #6413, opened 2018-08-07
Have Prometheus reporter enabled by default

ichernetsky:


ichernetsky:
@timcharper, can I have your blessing?

PR #6414, opened 2018-08-07
WIP: Add a test for upgrade from 1.4.9 to latest

alenkacz:
JIRA issues: DCOS-40232

### DIFF
--- a/tests/integration/src/test/scala/mesosphere/marathon/integration/UpgradeIntegrationTest.scala
+++ b/tests/integration/src/test/scala/mesosphere/marathon/integration/UpgradeIntegrationTest.scala
@@ -241,6 +241,47 @@ class UpgradeIntegrationTest extends AkkaIntegrationTest with MesosClusterTest w
     }
   }
 
+  "upgrade from 1.4.9 to the latest" in {
+    // Start apps in 1.4.9
+    Given("A Marathon 1.4.9 is running")
+    marathon149.start().futureValue
+    (marathon149.client.info.entityJson \ "version").as[String] should be("1.4.9")
+
+    And("new running apps in Marathon 1.4.9")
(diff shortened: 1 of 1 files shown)


tests/integration/src/test/scala/mesosphere/marathon/integration/UpgradeIntegrationTest.scala
```diff
@@ -71,13 +70,18 @@ class UpgradeIntegrationTest extends AkkaIntegrationTest with MesosClusterTest w
     }
   }
 
-  case class Marathon149(suiteName: String, masterUrl: String, zkUrl: String)(
+  class Marathon149Artifact()(
```
jeschkies: Hm, what do you think about making the marathon version fixtures? We cache the download so it is safe to re-download them each time.

Also, the artifact classes are so similar. I think we could just have one.

tests/integration/src/test/scala/mesosphere/marathon/integration/UpgradeIntegrationTest.scala
```diff
@@ -241,6 +261,51 @@ class UpgradeIntegrationTest extends AkkaIntegrationTest with MesosClusterTest w
...
+    eventually { marathon149 should have (runningTasksFor(app_149_fail.id.toPath, 1)) }
+
+    val originalApp149Tasks = marathon149.client.tasks(app_149.id.toPath).value
+    val originalApp149FailedTasks = marathon149.client.tasks(app_149_fail.id.toPath).value
+
+    When("Marathon 1.4.9 is shut down")
+    marathon149.stop().futureValue
+
+    And(s"App ${app_149_fail.id} fails")
+    killTask("app-149-fail")
```
jeschkies: I don't think this is necessary for this test case.

PR #6415, opened 2018-08-07
Address all of Tim's comments on #6340.

jeschkies:


src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -96,7 +96,7 @@ class InstanceOpFactoryImpl(
 
         val agentInfo = Instance.AgentInfo(offer)
         val taskIDs: Seq[Task.Id] = groupInfo.getTasksList.map { t => Task.Id(t.getTaskId) }(collection.breakOut)
-        val instance = Instance.Provisioned(scheduledInstance, agentInfo, hostPorts, pod, taskIDs, clock.now())
+        val instance = scheduledInstance.provisioned(agentInfo, hostPorts, pod, taskIDs, clock.now())
```
alenkacz: maybe we can be better with `provision` instead? scheduledInstance.provisioned reads funny to me.

jeschkies: Hm, that's a good suggestion.

PR #6418, opened 2018-08-07
Reschedule instance with provision timeout.

jeschkies:
Summary:
This reintroduced launch operation timeouts. If Mesos does not confirm a
task launch in time we reschedule the instance and forget about its old tasks.

JIRA issues: MARATHON-8351

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -78,6 +78,16 @@ case class Instance(
   override def region: Option[String] = agentInfo.flatMap(_.region)
 
   /**
+    * Factory method for creating a scheduled instance from this instance.
+    *
+    * @return new instance with same id in scheduled state.
+    */
+  def rescheduled(): Instance = {
+    val state = InstanceState(Condition.Scheduled, Timestamp.now(), None, None, Goal.Running)
+    Instance(instanceId, None, state, Map.empty, version, unreachableStrategy, None)
+  }
+
+  /**
     * Factory method for creating provisioned instance from Scheduled instance for apps
     * @return new instance in a provisioned state
     */
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -78,16 +78,6 @@ case class Instance(
   override def region: Option[String] = agentInfo.flatMap(_.region)
 
   /**
-    * Factory method for creating a scheduled instance from this instance.
-    *
-    * @return new instance with same id in scheduled state.
-    */
-  def rescheduled(): Instance = {
-    val state = InstanceState(Condition.Scheduled, Timestamp.now(), None, None, Goal.Running)
-    Instance(instanceId, None, state, Map.empty, version, unreachableStrategy, None)
-  }
-
-  /**
     * Factory method for creating provisioned instance from Scheduled instance for apps
     * @return new instance in a provisioned state
     */
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
@@ -64,7 +64,7 @@ private[launchqueue] object TaskLauncherActor {
 
   case object Stop extends Requests
 
-  private val OfferOperationRejectedTimeoutReason: String =
+  val OfferOperationRejectedTimeoutReason: String =
     "InstanceLauncherActor: no accept received within timeout. " +
       "You can reconfigure the timeout with --task_operation_notification_timeout."
 }
@@ -217,6 +217,7 @@ private class TaskLauncherActor(
             * The rescheduled instance forgets about all tasks. If these tasks do come back in an update they a killed.
             * See [[mesosphere.marathon.core.task.update.impl.TaskStatusUpdateProcessorImpl.publish()]] for the logic.
             */
+          logger.info(s"Reschedule ${instance.instanceId} because of provision timeout.")
           instanceTracker.schedule(instance.rescheduled())
         }
       }
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
@@ -214,12 +214,16 @@ private class TaskLauncherActor(
       if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
         instanceMap.get(op.instanceId).foreach { instance =>
           // We don't await the future. It will trigger an instance update event.
-          async {
-            await(instanceTracker.forceExpunge(op.instanceId))
-
-            val rescheduledInstance = Instance.Scheduled(runSpec)
-            await(instanceTracker.schedule(rescheduledInstance))
-          }
+          import org.apache.mesos.Protos
+          val (taskId, _) = instance.tasksMap.head
+          val state = Protos.TaskState.TASK_FAILED
+          val now = clock.now()
+          val phonyFailedTask = Protos.TaskStatus.newBuilder()
+            .setTaskId(taskId.mesosTaskId)
+            .setState(state)
+            .setTimestamp(now.seconds.toDouble)
+            .build()
+          instanceTracker.updateStatus(instance, phonyFailedTask, now)
         }
       }
 
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
@@ -7,6 +7,7 @@ import akka.Done
 import akka.actor._
 import akka.event.LoggingReceive
 import com.typesafe.scalalogging.StrictLogging
+import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.flow.OfferReviver
 import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.instance.update.{InstanceChange, InstanceDeleted, InstanceUpdateOperation}
@@ -24,6 +25,8 @@ import mesosphere.marathon.stream.Implicits._
 import org.apache.mesos.{Protos => Mesos}
 
 import scala.concurrent.Promise
+import scala.concurrent.duration._
+import scala.async.Async.{await, async}
 
 private[launchqueue] object TaskLauncherActor {
   def props(
@@ -205,6 +208,21 @@ private class TaskLauncherActor(
   }
 
   private[this] def receiveTaskLaunchNotification: Receive = {
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      import scala.concurrent.ExecutionContext.Implicits.global
+      // Reschedule instance if provisioning timed out and the instance is not running yet.
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+          // We don't await the future. It will trigger an instance update event.
+          async {
+            await(instanceTracker.forceExpunge(op.instanceId))
+
+            val rescheduledInstance = Instance.Scheduled(runSpec)
+            await(instanceTracker.schedule(rescheduledInstance))
+          }
+        }
+      }
+
     case InstanceOpSourceDelegate.InstanceOpRejected(op, reason) =>
       logger.debug(s"Task op '${op.getClass.getSimpleName}' for ${op.instanceId} was REJECTED, reason '$reason', rescheduling. $status")
       syncInstance(op.instanceId)
@@ -339,10 +357,12 @@ private class TaskLauncherActor(
       case InstanceUpdateOperation.Provision(instance) =>
         assert(instanceMap.contains(instance.instanceId), s"Internal task launcher state did not include provisioned instance ${instance.instanceId}")
         instanceMap += instance.instanceId -> instance
+        scheduleTaskOpTimeout(context, instanceOp)
         logger.info(s"Updated instance map to ${instanceMap.values.map(i => i.instanceId -> i.state.condition)}")
       case InstanceUpdateOperation.Reserve(instance) =>
         assert(instanceMap.contains(instance.instanceId), s"Internal task launcher state did not include reserved instance ${instance.instanceId}")
         instanceMap += instance.instanceId -> instance
+        scheduleTaskOpTimeout(context, instanceOp)
         logger.info(s"Updated instance map to ${instanceMap.values.map(i => i.instanceId -> i.state.condition)}")
       case other =>
         logger.info(s"Unexpected updated operation $other")
@@ -353,6 +373,17 @@ private class TaskLauncherActor(
     logger.info(s"Request ${instanceOp.getClass.getSimpleName} for instance '${instanceOp.instanceId.idString}', version '${runSpec.version}'. $status")
     promise.trySuccess(MatchedInstanceOps(offer.getId, Seq(InstanceOpWithSource(myselfAsLaunchSource, instanceOp))))
   }
+
+  private[this] def scheduleTaskOpTimeout(
+    context: ActorContext,
+    instanceOp: InstanceOp): Cancellable =
+    {
+      import context.dispatcher
+      val message: InstanceOpSourceDelegate.InstanceOpRejected = InstanceOpSourceDelegate.InstanceOpRejected(
+        instanceOp, TaskLauncherActor.OfferOperationRejectedTimeoutReason
+      )
+      context.system.scheduler.scheduleOnce(config.taskOpNotificationTimeout().milliseconds, self, message)
+    }
 
   private[this] def backoffActive: Boolean = backOffUntil.forall(_ > clock.now())
   private[this] def shouldLaunchInstances: Boolean = scheduledInstances.nonEmpty && !backoffActive
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
@@ -206,6 +206,7 @@ private class TaskLauncherActor(
     case RecheckIfBackOffUntilReached => OfferMatcherRegistration.manageOfferMatcherStatus()
   }
 
+  @SuppressWarnings(Array("all")) // async/await
   private[this] def receiveTaskLaunchNotification: Receive = {
     case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
       import scala.concurrent.ExecutionContext.Implicits.global
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
@@ -205,25 +205,19 @@ private class TaskLauncherActor(
     case RecheckIfBackOffUntilReached => OfferMatcherRegistration.manageOfferMatcherStatus()
   }
 
-  @SuppressWarnings(Array("all")) // async/await
   private[this] def receiveTaskLaunchNotification: Receive = {
     case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
-      // Send a phony task update to trigger a rescheduling.
+      // Reschedule instance with provision timeout..
       if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
         instanceMap.get(op.instanceId).foreach { instance =>
-          import org.apache.mesos.Protos
-          val (taskId, _) = instance.tasksMap.head
-          val state = Protos.TaskState.TASK_FAILED
-          val now = clock.now()
-          val phonyFailedTask = Protos.TaskStatus.newBuilder()
-            .setTaskId(taskId.mesosTaskId)
-            .setState(state)
-            .setTimestamp(now.seconds.toDouble)
-            .setMessage("Task provisioning timed out.")
-            .build()
 
-          // We don't await the future. It will trigger an instance update event.
-          instanceTracker.updateStatus(instance, phonyFailedTask, now)
+          /**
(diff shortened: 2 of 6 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -205,6 +208,21 @@ private class TaskLauncherActor(
   }
 
   private[this] def receiveTaskLaunchNotification: Receive = {
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      import scala.concurrent.ExecutionContext.Implicits.global
+      // Reschedule instance if provisioning timed out and the instance is not running yet.
```
jeschkies: This is only true for the deleted instance but not for the new instance. How should we retrieve the new instance? With phase 3 we won't to worry about this.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -204,7 +206,27 @@ private class TaskLauncherActor(
...
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+          // We don't await the future. It will trigger an instance update event.
+          import org.apache.mesos.Protos
+          val (taskId, _) = instance.tasksMap.head
+          val state = Protos.TaskState.TASK_FAILED
+          val now = clock.now()
+          val phonyFailedTask = Protos.TaskStatus.newBuilder()
```
jeschkies: I'll replace this status update once I confirmed that it works.

jeschkies: Come to think about it, I would like to keep this phony status. A lot of logic listens to the event bus and it would be non-trivial to introduce a different update. What do you think @timcharper, @alenkacz and @zen-dog?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -204,7 +205,28 @@ private class TaskLauncherActor(
...
+            .setTaskId(taskId.mesosTaskId)
+            .setState(state)
+            .setTimestamp(now.seconds.toDouble)
+            .setMessage("Task provisioning timed out.")
+            .build()
+
+          // We don't await the future. It will trigger an instance update event.
+          instanceTracker.updateStatus(instance, phonyFailedTask, now)
```
alenkacz: maybe log something? also maybe pipeTo(self)?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -204,7 +205,28 @@ private class TaskLauncherActor(
...
   }
 
+  @SuppressWarnings(Array("all")) // async/await
   private[this] def receiveTaskLaunchNotification: Receive = {
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      // Send a phony task update to trigger a rescheduling.
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+          import org.apache.mesos.Protos
+          val (taskId, _) = instance.tasksMap.head
```
alenkacz: will this work also for pods with multiple tasks? I think that if you change status of one to failed, it won't transition the whole instance to failed

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
alenkacz: so what are other options? a special event or... ?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -354,6 +378,17 @@ private class TaskLauncherActor(
     promise.trySuccess(MatchedInstanceOps(offer.getId, Seq(InstanceOpWithSource(myselfAsLaunchSource, instanceOp))))
   }
 
+  private[this] def scheduleTaskOpTimeout(
```
alenkacz: do we test this anywhere? Do you think it is worth testing?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: It will be in `DEGRADED` status if I'm not mistaken.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: It wasn't tested in the original. And yes. Ideally we would test it.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -204,7 +205,28 @@ private class TaskLauncherActor(
     case RecheckIfBackOffUntilReached => OfferMatcherRegistration.manageOfferMatcherStatus()
   }
 
+  @SuppressWarnings(Array("all")) // async/await
```
zen-dog: you don't need this anymore - scapegoat is out ;)

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
zen-dog: Definitely more logging mentioning task and instance ids

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -204,7 +205,28 @@ private class TaskLauncherActor(
...
   private[this] def receiveTaskLaunchNotification: Receive = {
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      // Send a phony task update to trigger a rescheduling.
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+          import org.apache.mesos.Protos
+          val (taskId, _) = instance.tasksMap.head
+          val state = Protos.TaskState.TASK_FAILED
```
zen-dog: I don't like this fake `TASK_FAILED` **at all**. This is Mesos task status and we shouldn't generate it ourselves.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -205,6 +206,21 @@ private class TaskLauncherActor(
...
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+
+          /**
+            * We don't await the future. It will trigger an instance update event.
+            *
+            * The rescheduled instance forgets about all tasks. If these tasks do come back in an update they a killed.
+            * See [[mesosphere.marathon.core.task.update.impl.TaskStatusUpdateProcessorImpl.publish()]] for the logic.
+            */
+          instanceTracker.schedule(instance.rescheduled())
```
timcharper: The fact that we write to this based on our cached copy without any consideration for modifications made by other modules makes me pretty nervous.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -77,6 +77,16 @@ case class Instance(
...
   override def region: Option[String] = agentInfo.flatMap(_.region)
 
+  /**
+    * Factory method for creating a scheduled instance from this instance.
+    *
+    * @return new instance with same id in scheduled state.
+    */
+  def rescheduled(): Instance = {
+    val state = InstanceState(Condition.Scheduled, Timestamp.now(), None, None, Goal.Running)
+    Instance(instanceId, None, state, Map.empty, version, unreachableStrategy, None)
```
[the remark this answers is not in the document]
timcharper: I also don't know if this will work. Given instance A and instance A', if A actually isn't terminal but comes back later, how do we know that the updates for A don't apply to the replacement A' ? Also, Mesos will reject a launch attempt for A if it knows A' to be active.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
alenkacz: yep, there were few races around that. At the same time, we receive all InstanceChange events through event stream.

On the other hand, if we process this timeout and at the same time someone persists some change to the instance, we then can be acting on stale state, true...

zen-dog:
> We need task incarnation count in order to reuse instance ids. I think we need to generate a new instance id for rescheduled.

I agree that we need a task incarnation for this (and other cases) but ideally, we keep the _InstanceId_. But @jeschkies thinks this would be too much of a change for this phase 2.

jeschkies:
> We need task incarnation count in order to reuse instance ids. I think we need to generate a new instance id for rescheduled.

So we are back to expunge?

jeschkies:
Ok, I went back to `forceExpunge`. In phase 3 we tackle incarnation counts https://jira.mesosphere.com/browse/MARATHON-8373

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -205,6 +206,22 @@ private class TaskLauncherActor(
...
   private[this] def receiveTaskLaunchNotification: Receive = {
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      // Reschedule instance with provision timeout..
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+
+          /**
+            * We don't await the future. It will trigger an instance update event.
+            *
+            * The rescheduled instance forgets about all tasks. If these tasks do come back in an update they a killed.
```
alenkacz: this comment is obsolete

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -354,6 +373,17 @@ private class TaskLauncherActor(
...
 
+  private[this] def scheduleTaskOpTimeout(
+    context: ActorContext,
+    instanceOp: InstanceOp): Cancellable =
+    {
+      import context.dispatcher
+      val message: InstanceOpSourceDelegate.InstanceOpRejected = InstanceOpSourceDelegate.InstanceOpRejected(
+        instanceOp, TaskLauncherActor.OfferOperationRejectedTimeoutReason
+      )
+      context.system.scheduler.scheduleOnce(config.taskOpNotificationTimeout().milliseconds, self, message)
```
alenkacz: I think we should cancel this when actor stops... Previously we were holding map of Cancellable items and cancelling them in postStop. I think we should do something similar here.

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -478,5 +479,50 @@ class TaskLauncherActorTest extends AkkaUnitTest {
         verifyClean()
       }
     }
+
+    "reschedule instance on provision timeout" in new Fixture {
+      Given("a provisioned instance")
+      val scheduledInstance = Instance.scheduled(f.app)
+      val provisionedInstance = Instance.scheduled(f.app)
```
alenkacz: I hate using copy here, because then we create instances in a states that just does not reflect reality. E.g. here we have provisioned instance with no task

jeschkies: No sure what you mean. This is how provisioned instances are created. We use the same fixtures in other tests.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: I thought about this. IMHO keep track of cancellable items just makes the implementation too complex. I can add it again but would not.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: ✅

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
jeschkies: ✅

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
alenkacz: The first way is to set up periodic message scheduling in the constructor of the actor, and cancel that scheduled sending in postStop or else we might have multiple registered message sends to the same actor.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -205,6 +206,17 @@ private class TaskLauncherActor(
...
 
   private[this] def receiveTaskLaunchNotification: Receive = {
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      // Reschedule instance with provision timeout..
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+
+          // We don't await the future. It will trigger an instance update event.
+          logger.info(s"Reschedule ${instance.instanceId} because of provision timeout.")
+          instanceTracker.forceExpunge(instance.instanceId)
```
timcharper: don't we need to do something different for resident tasks?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
timcharper: @jeschkies if we do not cancel then we will continue to send messages after the actor dies, we'll get a dead-letter.

```
/Users/tim/src/oss/akka/akka-actor/src/main/scala/akka/actor/Scheduler.scala

  42 |    */
  43 |   final def schedule(
  44 |     initialDelay: FiniteDuration,
  45 |     interval:     FiniteDuration,
  46 |     receiver:     ActorRef,
  47 |     message:      Any)(implicit
  48 |     executor: ExecutionContext,
  49 |                        sender: ActorRef = Actor.noSender): Cancellable =
  50 |     schedule(initialDelay, interval, new Runnable {
  51 |       def run = {
  52 |         receiver ! message
  53 |         if (receiver.isTerminated)
  54 |           throw new SchedulerException("timer active for terminated actor")
  55 |       }
  56 |     })
  57 | 
```

jeschkies: Ok, you convinced me.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: I'm not sure how we can reschedule task that do have a reservation but do not become running in time. I would just expunge the whole thing.

This should just be a short term solution. I really don't want this to be a blocker for phase 2.

jeschkies: @meichstedt, what is your take? We won't have a proper solution without incarnation counts and reconciliations as described in [MARATHON-8235](https://jira.mesosphere.com/browse/MARATHON-8235).

jeschkies: Shoot. I just realized that this might delete reservations on restarts as well. I'm inclined to drop the timeout for resident apps and pods.

alenkacz: why don't we just decommission/stop? I think we should not use forceExpunge... ideally anywhere. It feel super hacky to me...

jeschkies: Ideally we would but we don't kill tasks based on the goal state yet. That's phase 3.

alenkacz: oh that's right. And wouldn't the kill happen anyway just because we overscale? (potentially)

I think that in the old implementation (before phase 2) we did not kill as well. We just eventually overscaled...

meichstedt: Well the problem is this: if a resident task fails, the instance is kept and used to reschedule that task. If we expunge the instance in the case of this timeout, we would destroy persistent data, as long as the reservation/volume metadata is stored within the instance.

I'd not be OK with expunging even if it's just temporarily – it breaks resident tasks.

My suggestion is to handle this timeout with – tadaa – a surrogate signal. So instead of expunging the instance via `instanceTracker.forceExpunge(instance.instanceId)`, I'd build a GONE status like @timcharper did here https://github.com/mesosphere/marathon/pull/6290/files#diff-5e3e6fe795e2ed487104e6855e38ea32L69

add proper logging on level WARN that we are faking a Mesos task status here, and do
`instanceTracker.updateStatus(instance, goneStatus, now)`

That way, Marathon will keep the instance, and re-schedule as if that task was simply reported as GONE (meaning it'll increment the attempt count on the task etc). That's closest to the eager reconciliation described in MARATHON-8235 as we can get now.

jeschkies: I did a prototype in an older commit https://github.com/mesosphere/marathon/pull/6418/commits/e2c07629c007c95191193c98a577b42d180c1293. I'm not that convinced yet. For ephemeral task the expunge has the same effect as we have one master.

I'll have to dig into the current implementation for resident tasks.

jeschkies: Ok, it seems we just [overscale on master](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala#L208). On master, if the timeout hits we simple schedule a new launch. If the old one does succeed we overscaled.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -205,6 +213,25 @@ private class TaskLauncherActor(
...
+    case InstanceOpSourceDelegate.InstanceOpRejected(op, TaskLauncherActor.OfferOperationRejectedTimeoutReason) =>
+      // Reschedule instance with provision timeout..
+      if (inFlightInstanceOperations.exists(_.instanceId == op.instanceId)) {
+        instanceMap.get(op.instanceId).foreach { instance =>
+          import scala.concurrent.ExecutionContext.Implicits.global
+
+          logger.info(s"Reschedule ${instance.instanceId} because of provision timeout.")
+          async {
+            if (runSpec.isResident) {
+              await(instanceTracker.process(RescheduleReserved(instance, runSpec.version)))
```
jeschkies: We have an interesting case here. This command overrides the state of the tasks to `Scheduled`. Three things can happen now:

1. The provisioned tasks were reported running but the `TaskLauncherActor` did not know. We falsely reschedule the tasks.
2. The provisioned tasks never become running and Mesos sends and offer with the reservations again.
3. The provisioned tasks are reported running *after* we rescheduled them.

Cases 2 and 3 are fine. Since the tasks will be running.
Case 1 is interesting. The tasks are running but Marathon is waiting for an offer with the reservation. That will never come. However, after a reconciliation a Mesos update will override the tasks condition to running again.

alenkacz: I am kind of afraid it will be race-y and at the same time we don't have any integration test covering this so ... not sure :)

I mean you reschedule here, in the meantime mesos update might come with that task id, also you reschedule, but you still have in instanceMap the old instance, until you get InstanceChange event and sync that... 

Ideally I would like to see some integration tests loop but since we don't have tests for that I guess we have to believe? :) and fix this soon via decommission? (can you please create JIRA for this? so that we don't forget)

jeschkies: > I mean you reschedule here, in the meantime mesos update might come with that task id, also you reschedule, but you still have in instanceMap the old instance, until you get InstanceChange event and sync that...

It's actually not that racy as it will converge. Even when the test comes back as failed. We might overscale ephemeral tasks but the scale check will take care of these cases.

> Ideally I would like to see some integration tests loop but since we don't have tests for that I guess we have to believe? :) and fix this soon via decommission? (can you please create JIRA for this? so that we don't forget)

We don't kill with a `decommission` but you are right. Once we have that we should reschedule in a different way. Ideally through incarnation counts. See [MARATHON-8373](https://jira.mesosphere.com/browse/MARATHON-8373).

I''l set up a loop but doubt that it would show anything since the timeout is high enough.

jeschkies: I switched our [loop](https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-goals-phase-2/) to this branch.

jeschkies:
These are the results from the loop:

```
   5 mesosphere.marathon.core.deployment.impl.ReadinessBehaviorTest:ReadinessBehavior should A task that dies is removed from the actor
Sample size: 36
```

PR #6421, opened 2018-08-08
Do not update DCOS PR for 1.6

alenkacz:
Right now the PR contains both 1.6 and 1.7 versions, which is not correct. It should be just master.

### DIFF
--- a/ci/pipeline
+++ b/ci/pipeline
@@ -229,9 +229,6 @@ def master(): Unit = {
 
   // Uploads
   val maybeArtifact = uploadTarballPackagesToS3(version, s"builds/$version")
-  maybeArtifact.foreach { artifact =>
-    updateDcosImage(version, artifact.downloadUrl, artifact.sha1)
-  }
 }
 
 /**


ci/pipeline
```diff
@@ -229,9 +229,6 @@ def master(): Unit = {
 
   // Uploads
   val maybeArtifact = uploadTarballPackagesToS3(version, s"builds/$version")
-  maybeArtifact.foreach { artifact =>
```
zen-dog: This code shouldn't be called for non-master branches at all

alenkacz: but we want to create the package for non-master branches...

zen-dog: I'd rather extend https://github.com/mesosphere/marathon/pull/6421/files#diff-60758bba2cfe833f99a8029227ae4507L276
```
 def jenkins(): Unit = {	
   utils.request() match {	
     case PR(pullId) => pr(pullId)
     case Release(version) => release(version) // This, does the same as master() but don't patch DC/OS
     case Master => master()
   }
 }
```

PR #6424, opened 2018-08-08
Fix a `StackOverflowError` when GETin `v2/info` when running locally

zen-dog:
Summary:
When previously started locally with a Mesos connection string like `127.0.0.1:5050` Marathon would throw a `StackOverflowError` when GETing the `v2/info` which was caused by a recursive method call.


src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -354,7 +354,7 @@ object MarathonConf extends StrictLogging {
       /**
         * Credentials are not provided via the URL
         */
-      override def redactedConnectionString = toString
```
timcharper: 🤣 🤣

PR #6425, opened 2018-08-08
Unreachable validation now notes the dependency between properties

alenkacz:
Summary:
We can't really change the defaults because that would be a breaking change, so this at least improves the validation message.

Instead of `got 10 seconds, expected 0 seconds or less` it states that there is a dependency between those two properties and that one should be lower than other

JIRA issues: MARATHON-8049


ichernetsky:
```
java.lang.AssertionError: assertion failed: Found unused imports:
[warn] /home/admin/workspace/marathon-pipelines_PR-6425-ARQRLJYFM7SG4VIYIOHV5NU2ONCVJB7QAVKWZNFR57GAO3EYQNQA/src/main/scala/mesosphere/marathon/state/UnreachableStrategy.scala:6:27: Unused import
[warn] import com.wix.accord.dsl._
[warn]                           ^
[warn] /home/admin/workspace/marathon-pipelines_PR-6425-ARQRLJYFM7SG4VIYIOHV5NU2ONCVJB7QAVKWZNFR57GAO3EYQNQA/src/main/scala/mesosphere/marathon/state/UnreachableStrategy.scala:8:49: Unused import
[warn] import scala.concurrent.duration.FiniteDuration._
[warn] 
```

timcharper:
I'm surprised we don't have a test for this. I'd really like to add one.

src/main/scala/mesosphere/marathon/state/UnreachableStrategy.scala
```diff
@@ -34,13 +32,27 @@ object UnreachableEnabled {
...
+            UnreachableStrategy.inactiveAfterSmallerThanExpungeAfterMessage(unreachableEnabled.expungeAfter),
+            Path(Generic("inactiveAfterSeconds")))))
+      else
+        Success
+    }
   }
 }
 
 object UnreachableStrategy {
+  val InactiveAfterGreaterThanZeroMessage = "inactiveAfterSeconds must be greater than or equal to 0"
```
jeschkies: Could you add a test. It would be nice to see what the actual message looks like.

src/main/scala/mesosphere/marathon/state/UnreachableStrategy.scala
```diff
@@ -34,13 +32,27 @@ object UnreachableEnabled {
...
+  def inactiveAfterSmallerThanExpungeAfterMessage(expungeAfter: Duration) =
+    s"inactiveAfterSeconds must be less or equal to expungeAfterSeconds, which is ${expungeAfter.toSeconds}"
```
jeschkies: I think we should log both values:

`inactiveAfterSeconds (10s) must be less or equal to expungeAfterSeconds (5s)`

jeschkies: I really think we should add the times.

PR #6428, opened 2018-08-09
Fix group creation with mid-level groups

timcharper:
JIRA Issues: MARATHON-8360

### DIFF
--- a/src/test/scala/mesosphere/marathon/api/v2/GroupsResourceTest.scala
+++ b/src/test/scala/mesosphere/marathon/api/v2/GroupsResourceTest.scala
@@ -401,5 +401,36 @@ class GroupsResourceTest extends AkkaUnitTest with GroupCreation with JerseyTest
         rootGroup.app(PathId("/foo/sub/bibi")).isEmpty shouldBe false
       }
     }
+
+    "Allows group updates with mid-level groups" in {
+      new FixtureWithRealGroupManager(initialRoot = createRootGroup(groups = Set())) {
+        val body = """
+        {
+          "groups": [
+            {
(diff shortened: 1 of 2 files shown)


src/test/scala/mesosphere/marathon/api/v2/GroupsResourceTest.scala
```diff
@@ -401,5 +401,36 @@ class GroupsResourceTest extends AkkaUnitTest with GroupCreation with JerseyTest
...
+        f.service.deploy(any, any).returns(Future(Done))
+
+        val response = asyncRequest { r =>
+          groupsResource.createWithPath("", false, body.getBytes, auth.request, r)
+        }
+        response.getStatus shouldBe 201
+
+        val rootGroup = groupManager.rootGroup()
+        groupPaths(rootGroup) shouldBe Set("/", "/test-group", "/test-group/sleep")
+        rootGroup.app(PathId("/test-group/sleep/goodnight")).isEmpty shouldBe false
```
jeschkies: How about `rootGroup.app(PathId("/test-group/sleep/goodnight")) should be 'nonEmpty` this would give a more informative error message.

timcharper: That would compare the actual symbol literal, `nonEmpty`, right?

`.shouldNot(be(empty))` I think is what you mean?

timcharper: Hmm, that's a bit too much magic for my tastes. I'll go with shouldNot(be(empty).

PR #6429, opened 2018-08-09
Fix a `StackOverflowError` when GETin `v2/info` when running locally …

meichstedt:
…(#6424)

When previously started locally with a mesos connection string like `127.0.0.1:5050` Marathon would throw an `StackOverflowError` when GETing the `v2/info`.

JIRA issues:


zen-dog:
Rebased and pushed

PR #6436, opened 2018-08-09
Use `eventually` instead of `shakedown.deployment_wait()`.

jeschkies:
Summary:
The `deployment_wait` method of Shakedown uses a busy, noisy spinner.
The `assert_that(..., eventually(...))` pattern on the other hand gives
more insight into errors.

### DIFF
--- a/tests/system/marathon_common_tests.py
+++ b/tests/system/marathon_common_tests.py
@@ -26,7 +26,7 @@ def test_launch_mesos_container():
 
     client = marathon.create_client()
     client.add_app(app_def)
-    shakedown.deployment_wait()
+    assert_that(lambda: client.get_deployments(app_def['id']), eventually(has_len(0), max_attempts=120))
 
     tasks = client.get_tasks(app_def["id"])
     app = client.get_app(app_def["id"])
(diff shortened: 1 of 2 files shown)


tests/system/marathon_common_tests.py
```diff
@@ -26,7 +26,7 @@ def test_launch_mesos_container():
 
     client = marathon.create_client()
     client.add_app(app_def)
-    shakedown.deployment_wait()
+    assert_that(lambda: client.get_deployments(app_def['id']), eventually(has_len(0), max_attempts=120))
```
zen-dog: We now have 3 variations in our codebase: `shakedown.deployment_wait`, our own `common.deployment_wait` and this `assert_that..`. I'd rather have 1 method (preferably, the `common.deployment_wait` since it can wait for pods too) and make it use `assert_that` internally.

jeschkies: Well, this is a WIP. I'll change all `shakedown.deployment_wait` to `assert_that` once we agree on this pattern.

`common.deployment_wait` should rather go. I don't like to hide what's actually asserted.
`assert_that(lambda: client.get_deployments(app_def['id']), eventually(has_len(0), max_attempts=120))` follows more what we started doing in the integration tests.

Overall I do agree that we should use one pattern in our tests. We just have to decide which one.

zen-dog: > follows more what we started doing in the integration tests.

I'm not sure what you mean by that. I'm merely saying that writing this long assert line every time might be tedious. 
But more importantly, this assert only works for apps, where `common.deployment_wait` also works for pods.

jeschkies: We can easily use `common.deployment_predicate` in the assert. We could just use it in `common.deployment_wait` or maybe `common.assert_deployment`.

jeschkies:
Close in favour of #6440.

PR #6437, opened 2018-08-10
Delete cluster after test timeout.

jeschkies:
Summary:
So far Jenkins kills the system test job after two hours. This also
kills our cleanup call. This patch introduces a bash timeout that still
leaves time for a cluster cleanup.

JIRA issues: MARATHON-8364

### DIFF
--- a/ci/si_install_deps.sh
+++ b/ci/si_install_deps.sh
@@ -22,5 +22,10 @@ if ! command -v amm >/dev/null 2>&1; then
     echo "Ammonite successfully installed"
 fi
 
+# Ensure timeout is available.
+if [ "$PLATFORM" == 'Darwin' ]; then
+    alias timeout=gtimeout
+fi
+
 # Install dcos-launch and test dependencies.
 make init
--- a/ci/si_install_deps.sh
+++ b/ci/si_install_deps.sh
@@ -23,8 +23,11 @@ if ! command -v amm >/dev/null 2>&1; then
 fi
 
 # Ensure timeout is available.
-if [ "$PLATFORM" == 'Darwin' ]; then
-    alias timeout=gtimeout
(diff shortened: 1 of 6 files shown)


ci/si_install_deps.sh
```diff
@@ -22,5 +22,10 @@ if ! command -v amm >/dev/null 2>&1; then
     echo "Ammonite successfully installed"
 fi
 
+# Ensure timeout is available.
+if [ "$PLATFORM" == 'Darwin' ]; then
```
zen-dog: One has to have `brew install coreutils` but ok.

ci/si_pipeline.sh
```diff
@@ -76,7 +76,7 @@ case $CLUSTER_LAUNCH_CODE in
   0)
       "$ROOT_PATH/ci/dataDogClient.sc" "marathon.build.si.$VARIANT.cluster_launch.success" 1
       cp -f "$DOT_SHAKEDOWN" "$HOME/.shakedown"
-      make test
+      timeout --preserve-status -s KILL 1s make test
```
zen-dog: so we're killing after `1s`? 🤣 I don't think we're that fast yet

jeschkies:
Validate in build https://jenkins.mesosphere.com/service/jenkins/job/system-integration-tests/job/marathon-si-pr/237/. 

1. The cluster is launched.
2. The test is killed after the timeout and exits with 137.
3. The cluster is deleted.

ci/si_pipeline.sh
jeschkies: We are getting there...

ci/si_install_deps.sh
jeschkies: I'll add that.

jeschkies:
System test failed 🤔. We should investigate or restart.

PR #6438, opened 2018-08-10
 Crash if ZK DNS is not resolvable

zen-dog:
Summary:
This is a backport of two PRs #6218 and #6412.

Use Apache Curator `RetryPolicy` (#6218)
Summary:
we were using our custom `Retry` logic in the `ZKPersistenceStore` class. This change replaces it with Curator `BoundedExponentialBackoffRetry` policy.
Additionally:
- simplified `RichCuratorFramework.blockUntilConnected` method - an exception is immediately thrown if the connection to ZK fails
- introduced 2 new parameters to control the retry policy: `zk_operation_max_retries` and `zk_operation_base_retry_sleep_ms`
- removed some dead code

Crash if ZK DNS is not resolvable (#6412)
Currently, if Marathon can't connect to ZK because it's name is resolvable, Guice will reinstantiate `StorageModule` N times, effectively retrying to connect to ZK N times. Marathon will eventually exit with exitCode=137 but it takes ~5min with default settings. With this change we crash on the first error.

JIRA Issues: MARATHON-7390


zen-dog:
*DON'T SQUASH!*

PR #6440, opened 2018-08-12
Unify `deployment_wait` in the SI tests

zen-dog:
Currently, we're using `deployment_wait` waiting for all existing deployments to finish. This makes our SI tests more brittle and dependent on each other. This change replaces that with waiting for an app/pod id specific to the test in most cases.

### DIFF
--- a/tests/system/common.py
+++ b/tests/system/common.py
@@ -23,6 +23,7 @@ from functools import lru_cache
 from fixtures import get_ca_file
 from requests.exceptions import ReadTimeout
 from shakedown.dcos.cluster import ee_version
+from matcher import assert_that, eventually, has_len
 
 marathon_1_3 = pytest.mark.skipif('marthon_version_less_than("1.3")')
 marathon_1_4 = pytest.mark.skipif('marthon_version_less_than("1.4")')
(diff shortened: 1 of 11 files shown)


zen-dog:
> Awesome. Since you've taken this over how about changing common.deployment_wait as well:

Love it. Thx for the snippet ;)

tests/system/common.py
```diff
@@ -739,24 +740,30 @@ def agent_hostname_by_id(agent_id):
...
         the dcos-cli and remove this method later.
     """
-    shakedown.time_wait(lambda: deployment_predicate(service_id), timeout)
+    if (service_id is None):
+        print('Waiting for all current deployments to finish')
+    else:
+        print('Waiting for for {} to deploy successfully'.format(service_id))
+
+    assert_that(lambda: deployments_for(service_id),
+                eventually(has_len(0), wait_fixed=timeout, max_attempts=max_attempts))
```
jeschkies: Hm, this changes the semantics of timeout a little but it's fine with me.

tests/system/dcos_service_marathon_tests.py
```diff
@@ -25,8 +24,10 @@ def test_deploy_custom_framework():
     """
 
     client = marathon.create_client()
-    client.add_app(apps.fake_framework())
-    shakedown.deployment_wait(timeout=timedelta(minutes=5).total_seconds())
+    app_def = apps.fake_framework()
+    app_id = app_def["id"]
+    client.add_app(app_def)
+    common.deployment_wait(timeout=timedelta(minutes=5).total_seconds(), service_id=app_id)
```
jeschkies: Yeah, I don't think this is correct anymore. This diff changes the timeout to the individual timeout in milliseconds. Here it means the total timeout.

zen-dog: good catch

tests/system/test_marathon_universe.py
```diff
@@ -69,9 +68,9 @@ def test_custom_service_name():
         'service': {'name': "test-marathon"}
     }
     shakedown.install_package('marathon', options_json=options)
-    shakedown.deployment_wait()
+    common.deployment_wait(service_id=options["service"]["name"], max_attempts=300)
 
-    assert common.wait_for_service_endpoint('test-marathon', path="ping")
+    assert common.wait_for_service_endpoint('test-marathon', path="ping", timeout_sec=300)
```
jeschkies: Let's change this in a future diff.

PR #6445, opened 2018-08-14
Fix broken offer rejection statistics when maintenance mode is enabled

zen-dog:
Summary:
We were adding `UnfulfilledConstraint` and `AgentMaintenance` to the list of offer rejection reasons if maintenance mode was enabled. Fixed and added a test to prevent this in the future.


jeschkies:
I think we should also test the case that maintenance mode *is* enabled no `UnfulfilledConstraint` is returned. That was our error case.

PR #6448, opened 2018-08-15
Rejecting outliers in the data

wavesoft:
This commit enables the outlier rejection algorithm that rejects data that are
outside of a two-sigma tollerance.


wavesoft:
This PR was up for internal use. It seems that it's better to keep outliers, since this way we can easier detect misalignments. So I am closing it...

PR #6455, opened 2018-08-17
Add examples of alerting rules to the metrics doc

ichernetsky:
Also, nit: rename `marathon.tasks.*.gauge` to `marathon.instances.*.gauge`.

JIRA issues: MARATHON-8275


docs/docs/metrics.md
```diff
@@ -258,3 +258,126 @@ Dashes in metric names are replaced with underscores.
...
+  - alert: Marathon Too Few Instances
+    expr: sum(up{job="marathon"}) < 2
+    for: 30s
+    labels:
+      severity: critical
+      annotations:
+      summary: It fires when there are fewer Marathon instances up and running than expected.
+```
+
+Alert if Marathon's heap usage is too high:
```
zen-dog: add `... is too high ( >5Gb)`

docs/docs/metrics.md
```diff
@@ -258,3 +258,126 @@ Dashes in metric names are replaced with underscores.
...
+  - alert: Marathon Heap Usage
+    expr: avg_over_time(marathon_jvm_memory_heap_used_gauge_bytes[10s]) > (5 * 1024 * 1024 * 1024)
+    for: 10s
+    labels:
+      severity: critical
+    annotations:
+      summary: It fires when Marathon's heap is too big.
+```
+
+Alert if Marathon has too many threads:
```
zen-dog: `... too many threads ( >300)`

docs/docs/metrics.md
```diff
@@ -258,3 +258,126 @@ Dashes in metric names are replaced with underscores.
...
+      summary: It fires if there are too many deployments dismissed by Marathon.
+```
+
+Alert if Marathon receives offers, but is unable to use them for some
+reason, i.e. insufficient resources:
+
+```yaml
+  - alert: Marathon Given No Suitable Offer
+    expr: rate(marathon_mesos_offers_used_counter[1m]) == 0 and rate(marathon_mesos_offers_incoming_counter[1m]) > 0
+    for: 1m
```
zen-dog: That will fire way too often (even on smth. like our soak cluster)

docs/docs/metrics.md
```diff
@@ -258,3 +258,126 @@ Dashes in metric names are replaced with underscores.
...
+      severity: warning
+    annotations:
+      summary: It fires if there are incoming Mesos offers, but none can be used by Marathon.
+```
+
+Alert if there are too many concurrent HTTP requests to Marathon:
+
+```yaml
+  - alert: Marathon HTTP Requests
+    expr: avg_over_time(marathon_http_requests_active_gauge[10s]) > 100
```
zen-dog: 100 active requests per 10s are 10r/s. Three UI windows will probably produce that much ;)

docs/docs/metrics.md
```diff
@@ -258,3 +258,126 @@ Dashes in metric names are replaced with underscores.
...
+      severity: critical
+    annotations:
+      summary: It fires if there are too many concurrent HTTP requests to Marathon.
+```
+
+Alert if Marathon responds with a 5xx status code:
+
+```yaml
+  - alert: Marathon HTTP 5xx Responses
+    expr: marathon_http_responses_5xx_rate_m1_rate > 0
```
zen-dog: Since `5xx` responses are expected during e.g. leader reconciliation, I'd rather go for `marathon_http_responses_api_error_rate_m1_rate`...which is only implemented in the deprecated [metrics](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/HttpResponseMetricsHandler.scala#L20)? I'm in favor of adding it to the new metrics and using it here.

docs/docs/metrics.md
```diff
@@ -258,3 +258,126 @@ Dashes in metric names are replaced with underscores.
...
+    for: 10s
+    labels:
+      severity: critical
+    annotations:
+      summary: It fires if Marathon responds with a 5xx status codes.
+```
+
+Alert if Marathon has too many staged instances, which might indicate
```
zen-dog: `too many ( >100)...`

docs/docs/metrics.md
ichernetsky: These high thresholds are based on the metric values from our performance tests. Do you think we should target the use-case when there is only a handful of apps?

docs/docs/metrics.md
ichernetsky: It is 100 requests per second on average during past 10 seconds.

docs/docs/metrics.md
ichernetsky: The deprecated metric hides real failures. It ignores HTTP 503, which is about something else, not a bad gateway. So HTTP 502 should be used instead, but Marathon returns 503, if it is unable to forward a request. It is potentially a breaking change, though I am all pro it. What do you think? I would do it in a new PR though.

alenkacz: if different http 5XX can mean different things for alerting, shouldn't we maybe monitor them separately as well, not just inside 5xx metric?

But to your original thought - that sounds reasonable. If error metric is not actually monitoring errors I would consider it a bugfix :) So I guess - go for it 👍

timcharper: I support changing our exit codes to match the http spec. I think it’s unlikely that clients are hard coded to care about anything except it’s a 5xx

timcharper: I think it is fine to include 5xx responses during leader election. I’d Leave as is.

ichernetsky: @timcharper, do I understand you correctly

1. you are pro returning 502 instead of 503 when request forwarding fails
2. there is no need to introduce a new metric similar to the old one which ignores 503 (or 502) status codes, in addition to the existing ones?

?

timcharper: Hmm, I wasn’t aware that we hid leader forwarding 5xx responses them previously. I would not hide them.

I support using 502 to indicate proxy failures. Also... 504 if proxy times out?

ichernetsky: 504 is already sent if it times out:
```
            case Failure(ex: SocketTimeoutException) =>
              logger.error(ERROR_STATUS_GATEWAY_TIMEOUT, ex)
              response.sendError(GatewayTimeout.intValue, ERROR_STATUS_GATEWAY_TIMEOUT)
```

ichernetsky:
@zen-dog, Github deleted/hid the thread about `marathon_http_responses_api_error_rate_m1_rate`. Tim and I do not see much value in having it with the new metrics. Do you want to have it added anyway? If so, could you please elaborate on what value exactly you see in having it?

docs/docs/metrics.md
zen-dog: I think this is a meaningful value. Afaik we recommend 2Gb and I've almost never seen Marathon exceeding 5Gb in "normal circumstances" so the values sgtm.

zen-dog:
> Github deleted/hid the thread about marathon_http_responses_api_error_rate_m1_rate. Tim and I do not see much value in having it with the new metrics. Do you want to have it added anyway? If so, could you please elaborate on what value exactly you see in having it?

I've added `api_error` metric a while ago because some 5xx are thrown by Marathon during "normal work" e.g. during leader re-election. In the past people on our soak-cluster used to watch the `5xx`  metric which resulted in false positive incidents.

PR #6457, opened 2018-08-17
Consider resident condition as terminal

alenkacz:
We don't use reserved state for nothing else as terminal instances, that are stopped or are waiting to be rescheduled.

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/task/TaskTest.scala
+++ b/src/test/scala/mesosphere/marathon/core/task/TaskTest.scala
@@ -156,23 +156,6 @@ class TaskTest extends UnitTest with Inside {
       task.isUnreachableExpired(f.clock.now, 10.minutes) should be(false)
     }
 
-    "a reserved task transitions to launched on running MesosUpdate" in {
-      val f = new Fixture
-      val condition = Condition.Reserved
-      val taskId = Task.Id.forRunSpec(f.appWithIpAddress.id)
-      val status = Task.Status(f.clock.now, None, None, condition, NetworkInfoPlaceholder())
-      val task = Task(taskId, f.clock.now, status)
-      val instance = mock[Instance]
-      instance.hasReservation returns true
-      val mesosStatus = MesosTaskStatusTestHelper.running(taskId)
-      inside(task.update(instance, Condition.Running, mesosStatus, f.clock.now)) {
-        case effect: TaskUpdateEffect.Update =>
-          effect.newState shouldBe a[Task]
-          effect.newState.status.condition shouldBe Condition.Running
-          effect.newState.status.mesosStatus shouldBe Some(mesosStatus)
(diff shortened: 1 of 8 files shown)


src/test/scala/mesosphere/marathon/core/task/TaskTest.scala
```diff
@@ -156,23 +156,6 @@ class TaskTest extends UnitTest with Inside {
       task.isUnreachableExpired(f.clock.now, 10.minutes) should be(false)
     }
 
-    "a reserved task transitions to launched on running MesosUpdate" in {
```
alenkacz: This was actually interesting. My reasoning behind droping this is - instance is right now in reserved state only when has reservation and some of the tasks became terminal. I don't think that this case is valid anymore - it was probably meant more for the Reserved instances that were Reserved-Scheduled. But those are now in condition Scheduled, not Reserved.

timcharper: I really, really dislike the `Reserved` condition :( It represents state that has been shoe-horned in to a dimension where it doesn't belong.

We have previously used `TASK_STAGING` condition represent a terminal resident task. Which is so, so very wrong. I've just confirmed this to be the case with the latest Marathon 1.6.x release:

```
$ marathon-cli tasks
id                                               state         startedAt                 host
le-sleep.87cdec0e-a276-11e8-9fc9-5646b9d89784.1  TASK_RUNNING  2018-08-17T23:37:41.784Z  bastion.lan

$ echo '{"instances": 0}' | http PUT :8080/v2/apps/le-sleep?force=true
... snip ...

{
    "deploymentId": "45a4f975-f583-4e4a-901e-40c80d489207",
    "version": "2018-08-17T23:39:12.993Z"
}

$ marathon-cli tasks
id                                               state         startedAt  host
le-sleep.87cdec0e-a276-11e8-9fc9-5646b9d89784.1  TASK_STAGING  null       bastion.lan
```

Given the insanity of this, I think it is okay to change it, but we need to be aware that this change _will_ probably break somebody's integration somewhere. As such, I wonder if we will need to introduce a deprecated feature flag so that we can continue to at least present "TASK_STAGING" for suspended resident tasks. I don't like the complexity of doing so, at all. What are your thoughts, @alenkacz ?

jeschkies:
I'm sorry, I fail to see the added value for phase 2 since we want to drop the reserved condition anyways, I don't understand why we change anything with it. What am I missing?

alenkacz:
@jeschkies in my opinion it goes in line with consistency. Since changes included in Phase 2 actually change reserved in a way that it is really *ONLY* a terminal state, it should be treated as other terminal states.

timcharper:
@alenkacz do you want me to review this one? I'm waiting for CI to become green.

PR #6458, opened 2018-08-17
Loosen constraint validation during migration

meln1k:
Summary: "UNIQUE" constraint value is stripped in case it is empty.

JIRA issues: DCOS-37999


src/test/scala/mesosphere/marathon/storage/migration/MigrationTo15Test.scala
```diff
@@ -241,6 +242,29 @@ class MigrationTo15Test extends AkkaUnitTest with RecoverMethods with GroupCreat
         migrateSingleApp(sd) should be(expected)
       }
 
+      "trim an empty UNIQUE constraint value and not die" in new Fixture {
```
alenkacz: I am not sure about that "not die" part. What do you think about 'trim an empty UNIQUE constraint value'

PR #6466, opened 2018-08-22
Add Stop task event

mushixun:
<<Replace this line with your Revision Title>>

Summary:

JIRA issues:
https://jira.mesosphere.com/browse/MARATHON-8339


mushixun:
How to get mesphere onglogin,I  can not view ci log。

alenkacz:
@mushixun Hello! Thank you for the PR.

Can you please talk a bit more about your intentions in comments here https://jira.mesosphere.com/browse/MARATHON-8339

Before accepting the PR we need to understand why InstanceChange event is not an option for you. You should get one shortly after the kill was initiated.

mushixun:
@alenkacz OK,When using marathon-lb for load balancing, when marathon is shrinking, some requests fail. After analysis, it is because the container has not been removed from the haproxy routing table when the container is killed. It is not caused by haproxy reload, so I hope marathon can tell which container to shrink in advance when shrinking. So that I can first remove the container information from haproxy's routing table, no longer accept new requests, and then kill the container, and the premise of this function is that marathon needs to issue the first stop when shrinking. send stop event information.

mushixun:
Tell marathon-lb that which containers will be stop

.gitignore
```diff
@@ -1,29 +0,0 @@
-bin/run-tests-config.sh
```
ichernetsky: Please restore this file.

ichernetsky:
```
src/test/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActorTest.scala:310:54: not enough arguments for method props: (driverHolder: mesosphere.marathon.MarathonSchedulerDriverHolder, instanceTracker: mesosphere.marathon.core.task.tracker.InstanceTracker, config: mesosphere.marathon.core.task.termination.KillConfig, clock: java.time.Clock, eventStream: akka.event.EventStream)akka.actor.Props.
[error] Unspecified value parameter eventStream.
[error]     val actor = system.actorOf(KillServiceActor.props(f.driverHolder, f.instanceTracker, killConfig, f.clock), s"KillService-${UUID.randomUUID()}")
```

PR #6467, opened 2018-08-22
Assert that subscription keys are removed eventually.

jeschkies:
Summary:
This should stabilize the `ReadinessBehaviorTest`. In 1.6 we would
remove the subscription key immediately. This was changed on master to
be asyncronous. So we have to wait a little until the key is removed.


timcharper:
Thanks for jumping on this @jeschkies

PR #6471, opened 2018-08-23
Rename Debian and RPM packages to include proper arch.

jeschkies:
Summary:
The deb and rpm packages for Maraton should include an arch and not the
`SNAPTSHOT` tags to avoid conflicts on our package server.

JIRA issues: MARATHON-8379, MARATHON-8379

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -217,6 +217,8 @@ lazy val packagingSettings = Seq(
   serverLoading := None, // We override this to build for each supported system loader in the packageLinux alias
   debianPackageDependencies in Debian := Seq("java8-runtime-headless", "lsb-release", "unzip", s"mesos (>= ${Dependency.V.MesosDebian})"),
   packageArchitecture in Debian := "amd64",
+  version in Debian := s"${version.value}-1.ubuntu1804",
+
   packageArchitecture in Rpm := "x86_64",
   rpmRequirements in Rpm := Seq("coreutils", "unzip", "java >= 1:1.8.0"),
   rpmVendor := "mesosphere",
--- a/build.sbt
+++ b/build.sbt
@@ -138,6 +138,7 @@ lazy val packageRpmForLoader = taskKey[File]("Create rpm package for active serv
 lazy val packagingSettings = Seq(
   bashScriptExtraDefines += IO.read((baseDirectory.value / "project" / "NativePackagerSettings" / "extra-defines.bash")),
   mappings in (Compile, packageDoc) := Seq(),
+  isSnapshot in ThisBuild := false,
   debianChangelog in Debian := Some(baseDirectory.value / "changelog.md"),
 
   (packageName in Universal) := {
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -215,16 +216,15 @@ lazy val packagingSettings = Seq(
   maintainer := "Mesosphere Package Builder <support@mesosphere.io>",
   serverLoading := None, // We override this to build for each supported system loader in the packageLinux alias
   debianPackageDependencies in Debian := Seq("java8-runtime-headless", "lsb-release", "unzip", s"mesos (>= ${Dependency.V.MesosDebian})"),
+  packageArchitecture in Debian := "amd64",
+  version in Debian := s"${version.value}-1.ubuntu1804",
```
jeschkies: @timcharper, need a version for `debian8`, `debian9`, `ubuntu1404`, `ubuntu1604` and `ubuntu1804`. What is the best way to create these configurations?

timcharper: @jeschkies if these repos all include java8-runtime-headless, lsb-release, etc. then nothing is needed. We only create three deb packages, currently, one for each init system (systemv, SystemD, and Upstart). Then we publish the same package to each repo, according to their init system.

build.sbt
```diff
@@ -215,16 +216,15 @@ lazy val packagingSettings = Seq(
...
+
+  packageArchitecture in Rpm := "x86_64",
```
timcharper: Is this correct? I don't believe we ship any x86_64 code in the package. It should be `arch`.

build.sbt
```diff
@@ -215,16 +216,15 @@ lazy val packagingSettings = Seq(
   maintainer := "Mesosphere Package Builder <support@mesosphere.io>",
   serverLoading := None, // We override this to build for each supported system loader in the packageLinux alias
   debianPackageDependencies in Debian := Seq("java8-runtime-headless", "lsb-release", "unzip", s"mesos (>= ${Dependency.V.MesosDebian})"),
+  packageArchitecture in Debian := "amd64",
```
timcharper: I don't think this is correct. See below.

build.sbt
```diff
@@ -138,6 +138,7 @@ lazy val packageRpmForLoader = taskKey[File]("Create rpm package for active serv
 lazy val packagingSettings = Seq(
   bashScriptExtraDefines += IO.read((baseDirectory.value / "project" / "NativePackagerSettings" / "extra-defines.bash")),
   mappings in (Compile, packageDoc) := Seq(),
+  isSnapshot in ThisBuild := false,
```
timcharper: I thought at one point we had isSnapshot being inferred by the version pointing to a tag or not. But, I could get behind this. We really don't do snapshot builds anyways.

build.sbt
timcharper: Mappings for upload are here:  https://github.com/mesosphere/marathon/blob/v1.7.50/ci/releases.sc#L113

build.sbt
jeschkies: @lloesche suggested this. He can explain.

lloesche: @timcharper see my comment in the corresponding Jira. The repo server was originally only for Mesos and we only had a x86_64 architecture. Our instructions originally told folks to only add this repo to their yum config. We never changed that for Marathon, which is why the original marathon tagged it's packages as x86_64 even though noarch would be correct. We later created a mesosphere-el-repo-7-3.noarch.rpm package which installs existing archs and repos for the user and even updated our blog post to have the correct instructions but by that time a lot of users were already using the repo and most likely never updated it. So we deemed it safer to continue using the x86_64 arch in the packages metadata instead of having users stop receiving updates.

jeschkies:
@timcharper

> I am not convinced we should set the x86_64 architecture

Did you read Lukas' reply? I don't have a strong opinion on this but would like to find consensus.

jeschkies:
Dropped in favor of #6521.

PR #6472, opened 2018-08-23
Applications created via PUT do not receive a default portDefinitions value

meln1k:
Summary: we do full update instead of partial in update is applied to a non-existent app

JIRA issues: MARATHON-7969

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
@@ -157,9 +157,7 @@ class AppsResource @Inject() (
     *
     * @param appId used as the id of the generated app update (vs. whatever might be in the JSON body)
     * @param body is the raw, unparsed JSON
-    * @param partialUpdate true if the JSON should be parsed as a partial application update (all fields optional)
-    *                      or as a wholesale replacement (parsed like an app definition would be)
-    * @param newApp if the app is created via put, we shouldn't treat it as a partial update
+    * @param updateType CompleteReplacement if we want to replace the app entirely, PartialUpdate if we only want to update provided parts
     */
   def canonicalAppUpdateFromJson(appId: PathId, body: Array[Byte], updateType: UpdateType): raml.AppUpdate = {
     updateType match {
@@ -172,7 +170,7 @@ class AppsResource @Inject() (
         // the version is thrown away in conversion to AppUpdate
         jsObj.as[raml.App].normalize.toRaml[raml.AppUpdate]
 
-      case PartialReplacement =>
+      case PartialUpdate =>
         Json.parse(body).as[raml.AppUpdate].copy(id = Some(appId.toString)).normalize
 
     }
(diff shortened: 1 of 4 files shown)


src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -159,9 +159,10 @@ class AppsResource @Inject() (
     * @param body is the raw, unparsed JSON
     * @param partialUpdate true if the JSON should be parsed as a partial application update (all fields optional)
     *                      or as a wholesale replacement (parsed like an app definition would be)
+    * @param newApp if the app is created via put, we shouldn't treat it as a partial update
     */
-  def canonicalAppUpdateFromJson(appId: PathId, body: Array[Byte], partialUpdate: Boolean): raml.AppUpdate = {
-    if (partialUpdate) {
+  def canonicalAppUpdateFromJson(appId: PathId, body: Array[Byte], partialUpdate: Boolean, newApp: Boolean): raml.AppUpdate = {
+    if (partialUpdate && !newApp) {
```
kensipe: this doesn't make sense to me... 
`newApp` says if app is created via put... it isn't a partial update... why wouldn't we just `partialUpdate=False` ?

What does `partialUpdate=True`  AND `newApp=True` mean?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -323,7 +327,15 @@ class AppsResource @Inject() (
     req: HttpServletRequest, allowCreation: Boolean)(implicit identity: Identity): Future[Response] = async {
     val appId = id.toRootPath
 
-    val appUpdate = canonicalAppUpdateFromJson(appId, body, partialUpdate)
+    val appDoesNotExist = groupManager.app(appId).isEmpty
```
ichernetsky: There is a race condition here. I believe it is not trivial to get rid of it, and a comment warning about it would suffice.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -323,7 +327,15 @@ class AppsResource @Inject() (
...
+
+    val updateType = (appDoesNotExist, partialUpdate) match {
+      case (true, _) => CompleteReplacement
```
ichernetsky: wordsmith: it is not a replacement then.

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -323,7 +327,15 @@ class AppsResource @Inject() (
...
+      case (_, true) => PartialReplacement
```
ichernetsky: wordsmith: `PartialUpdate` or `Patch`?

tests/integration/src/test/scala/mesosphere/marathon/integration/AppDeployIntegrationTest.scala
```diff
@@ -810,6 +811,34 @@ class AppDeployIntegrationTest extends AkkaIntegrationTest with EmbeddedMarathon
...
+                            |      "network":"HOST"
+                            |    }
+                            |  }
+                            |}""".stripMargin
+
+      When("creating an app using PUT")
+      marathon.putAppByteString(applicationId, ByteString.fromString(appUpdateJson))
+      Then("port definitions are set correctly")
+      val updatedApp = marathon.app(applicationId)
+      updatedApp.value.app.portDefinitions should not be None
```
zen-dog: `s/should not be None/shouldBe defined`

PR #6475, opened 2018-08-24
PATCH call not proxied correctly to active node

meln1k:
Summary: added a workaround for http methods unsupported by java

JIRA issues: MARATHON-8095

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/forwarder/JavaUrlConnectionRequestForwarder.scala
+++ b/src/main/scala/mesosphere/marathon/api/forwarder/JavaUrlConnectionRequestForwarder.scala
@@ -111,7 +111,7 @@ class JavaUrlConnectionRequestForwarder(
   }
 
   private def copyRequestToConnection(leaderConnection: HttpURLConnection, request: HttpServletRequest): Try[Done] = Try {
-    if (request.getMethod == "PATCH") { // workaround to overcome java restrictions, see https://bugs.openjdk.java.net/browse/JDK-7016595
+    if (request.getMethod == "PATCH") { // workaround to overcome java restrictions, see https://bugs.openjdk.java.net/browse/JDK-7016595 and https://stackoverflow.com/a/39641592
       setPatchMethod(leaderConnection)
     } else {
       leaderConnection.setRequestMethod(request.getMethod)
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/api/forwarder/JavaUrlConnectionRequestForwarder.scala
```diff
@@ -105,7 +105,12 @@ class JavaUrlConnectionRequestForwarder(
   }
 
   private def copyRequestToConnection(leaderConnection: HttpURLConnection, request: HttpServletRequest): Try[Done] = Try {
-    leaderConnection.setRequestMethod(request.getMethod)
+    if (request.getMethod == "PATCH") { //workaround to overcome java restrictions, see https://bugs.openjdk.java.net/browse/JDK-7016595
```
ichernetsky: nit: a space after `//`.

src/main/scala/mesosphere/marathon/api/forwarder/JavaUrlConnectionRequestForwarder.scala
```diff
@@ -105,7 +105,12 @@ class JavaUrlConnectionRequestForwarder(
...
+      leaderConnection.setRequestProperty("X-HTTP-Method-Override", "PATCH")
```
zen-dog: Does jetty automatically deal with `"X-HTTP-Method-Override", "PATCH"`? Did you try it locally?

src/main/scala/mesosphere/marathon/api/forwarder/JavaUrlConnectionRequestForwarder.scala
```diff
@@ -104,8 +104,18 @@ class JavaUrlConnectionRequestForwarder(
     leaderConnection.addRequestProperty(HEADER_FORWARDED_FOR, forwardedFor)
   }
 
+  private def setPatchMethod(connection: HttpURLConnection): Unit = {
+    val methodsField = classOf[HttpURLConnection].getDeclaredField("method")
+    methodsField.setAccessible(true)
```
timcharper: do we have to set it accessible every time? Or just once?

I think I'd like to see an object that encapsulates this unfortunate but necessary workaround

```
/**
 * Workaround to overcome Java restrictions; see https://bugs.openjdk.java.net/browse/JDK-7016595
 */
object HttpUrlConnectionWorkaround {
  val methodsField = classOf[HttpURLConnection].getDeclaredField("method")
  methodsField.setAccessible(true)

  def setMethod(connection: HttpURLConnection, method: String): Unit = {
    if (method == "PATCH")
      methodsField.set(connection, method)
    else
      connection.setRequestMethod(method)
  }
}
```

timcharper:
This is just for the sync proxy in 1.6.x, which is deprecated and slated for removal in 1.8.x

PR #6479, opened 2018-08-27
Metrics tests

ichernetsky:
JIRA issues: MARATHON-8353

### DIFF
--- a/docs/docs/metrics.md
+++ b/docs/docs/metrics.md
@@ -141,35 +141,32 @@ Dashes in metric names are replaced with underscores.
 
 ### HTTP-specific metrics
 
-* `marathon.http.event-streams.responses.size.counter.bytes` — the size
+* `marathon.http.responses.event-stream.size.counter.bytes` — the size
   of data sent to clients over event streams since the current Marathon
   instance became a leader.
 * `marathon.http.requests.size.counter.bytes` — the total size of
   all requests since the current Marathon instance became a leader.
-* `marathon.http.requests.size.gzipped.counter.bytes` — the total size
-  of all gzipped requests since the current Marathon instance became
-  a leader.
 * `marathon.http.responses.size.counter.bytes` — the total size of all
   responses since the current Marathon instance became a leader.
 * `marathon.http.responses.size.gzipped.counter.bytes` — the total size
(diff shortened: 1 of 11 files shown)


src/test/scala/mesosphere/marathon/api/SystemResourceTest.scala
```diff
@@ -95,17 +94,30 @@ class SystemResourceTest extends AkkaUnitTest with JerseyTest {
...
+      When("the metrics are requested")
       val response = resource.metrics(auth.request)
 
-      Then("The metrics are sent")
+      Then("the metrics are sent")
+      response.getStatus shouldBe 200
+      Option(response.getMetadata.getFirst("Content-Type")).value.toString should be("application/json")
+
       val metricsJson = Json.parse(response.getEntity.asInstanceOf[String])
-      metricsJson \ "start" shouldBe a[JsDefined]
```
alenkacz: shouldn't we have tests for both kamon and dropwizard until we drop the old one?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceUpdateActorTest.scala
```diff
@@ -85,8 +84,8 @@ class InstanceUpdateActorTest extends AkkaUnitTest {
       f.instanceTrackerActorReplyWith(Done)
 
       And("all gauges are zero again")
-      f.actorMetrics.oldActiveOpsMetric.value should be(0)
```
alenkacz: do you think it makes sense to test both old and new here? Until we drop old...

docs/docs/metrics.md
```diff
@@ -141,35 +141,32 @@ Dashes in metric names are replaced with underscores.
 
 ### HTTP-specific metrics
 
-* `marathon.http.event-streams.responses.size.counter.bytes` — the size
+* `marathon.http.responses.event-stream.size.counter.bytes` — the size
```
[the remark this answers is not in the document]
wavesoft: Ok, my bad. It only appears when someone is using the events endpoint 👍

ichernetsky:
@alenkacz, thanks for your comments. I have addressed all of them.

PR #6480, opened 2018-08-27
Define linter for Shakedown and begin cleanup.

jeschkies:
Summary:
`flake8` shows almost three hundred errors. With this change we define
our dev dependencies with pipenv and run the linter with `make build`.
We also follow the style guide in `network.py`, `command.py` and
`master.py`.

Please note that the tests are not linted.

### DIFF
--- a/tests/shakedown/.gitignore
+++ b/tests/shakedown/.gitignore
@@ -17,6 +17,7 @@ eggs/
 lib/
 lib64/
 parts/
+Pipfile.lock
 sdist/
 var/
 *.egg-info/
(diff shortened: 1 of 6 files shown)


tests/shakedown/.gitignore
```diff
@@ -17,6 +17,7 @@ eggs/
 lib/
 lib64/
 parts/
+Pipfile.lock
```
jeschkies: Libraries should not include the lock file.

jeschkies:
Once we have all linting issues fixed  we can trigger the build with our CI.

PR #6485, opened 2018-08-28
Using `dcos-launch@4a2515f` git hash version

zen-dog:
Summary: recent `dcos-launch` HEAD broke on-prem installation due to a missing docker dependency. We're going to the previous commit. Additionally, preflight logs are added to Jenkins archives to help debug such problems in the future.

Related JIRA: DCOS_OSS-4007


zen-dog:
> The system test job seems to fail.

I would've waited for the system test anyway ;)

zen-dog:
I've updated the `Pipfile`. Re-running the SI tests

PR #6486, opened 2018-08-28
Document new exit codes in changelog.

jeschkies:
JIRA issues: MARATHON-8381


jeschkies:
> But I think we need that in the 1.6.X changelog too.

Hence the backport label 🙂

PR #6492, opened 2018-08-29
Use correct source repository

schneidexe:
Use correct source repository

Summary:
* Changed git url to recent mesos source code repository on github.com

JIRA issues:
* none

Seems like mesos sources meanwhile moved to https://github.com/apache/mesos, so I updated the script accordingly.

### DIFF
--- a/tools/mvm.sh
+++ b/tools/mvm.sh
@@ -421,7 +421,7 @@ fi
 # check if mesos source directory exists
 if [ ! -d "$MESOS_SOURCES" ]; then
   echo "Mesos sources not found. Cloning sources."
-  git clone https://git-wip-us.apache.org/repos/asf/mesos.git "$MESOS_SOURCES"
+  git clone https://github.com/apache/mesos.git "$MESOS_SOURCES"
 fi
 
 # process command line arguments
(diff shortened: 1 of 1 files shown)


ichernetsky:
@schneidexe, I believe it moved to a different location. Please refer to http://mesos.apache.org/documentation/latest/building/

schneidexe:
Fair enough. So looks like github.com is just a mirror. Anyway current one in the doc is not working.
So what would be the preferred way? Update this PR or create a PR on gitbox (I don't have an account there)?

timcharper:
Pretty obvious change.

tools/mvm.sh
```diff
@@ -421,7 +421,7 @@ fi
 # check if mesos source directory exists
 if [ ! -d "$MESOS_SOURCES" ]; then
   echo "Mesos sources not found. Cloning sources."
-  git clone https://git-wip-us.apache.org/repos/asf/mesos.git "$MESOS_SOURCES"
+  git clone https://github.com/apache/mesos.git "$MESOS_SOURCES"
```
timcharper: Looks like it should be https://gitbox.apache.org/repos/asf?p=mesos.git

schneidexe: Updated according to http://mesos.apache.org/documentation/latest/building/. Took the version w/o special char like '?' to prevent any possible issues using it inside sh/cmd.

PR #6507, opened 2018-09-03
Switched to python `logging` instead of `print` in SI tests

zen-dog:
Summary:
We removed `shakedown` Pytest plugin which previously captured and enriched test stdout. This was done to consolidate the logging format of different test and library modules. In detail:
- removed `--fail, --timeout, --quite, --stdout, --stdout-inline` shakedown options - we pass what we need directly through `--pytest-option`
- removed default Pytest options from the shakedown `main.py`
- using python `logging` package instead of just `print`ing in all Marathon SI tests and in both marathon clients (`marathon.py`s)
- logging configuration is read from `logging.conf`

JIRA: MARATHON-8390

### DIFF
--- a/tests/shakedown/shakedown/cli/main.py
+++ b/tests/shakedown/shakedown/cli/main.py
@@ -10,14 +10,9 @@ from shakedown.dcos.spinner import *
 @click.command('shakedown')
 @click.argument('tests', nargs=-1)
 @click.option('-u', '--dcos-url', envvar='SHAKEDOWN_DCOS_URL', help='URL to a running DC/OS cluster.')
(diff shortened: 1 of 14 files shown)


zen-dog:
In general we're one step closer to use `shakedown` as a library (and not as Pytest wrapper)

tests/shakedown/shakedown/cli/main.py
```diff
@@ -152,206 +145,13 @@ def cluster_available_predicate(url):
                     click.secho("error: no authentication credentials or token found.", fg='red', bold=True)
                     sys.exit(1)
 
-    class shakedown:
```
jeschkies: 🎉

zen-dog: 😊

jeschkies:
> I'm curious why we don't just move shakedown to python 3.6 and use f"String $interpolation"... 1 step at a time... nice update!

I think we'll do that in time. We should check the requirements with other users.

PR #6508, opened 2018-09-04
Fixing is_enterprise_cli_package_installed flaky test

wavesoft:
This PR fixes the `test_marathon_root.test_private_repository_mesos_app` flaky test, by adding retry guard


tests/system/common.py
```diff
@@ -351,14 +351,15 @@ def install_enterprise_cli_package():
     stdout, stderr, return_code = shakedown.run_dcos_command(cmd, raise_on_error=True)
 
 
+@retrying.retry(wait_fixed=1000, stop_max_attempt_number=10, retry_on_exception=ignore_exception)
```
zen-dog: What was the error that `cli` reported? Why do we need to retry here?

wavesoft: The node is some times not accessible. I cannot think of anything else that could be wrong (apart from the fact that the cluster is in a weird state):
```
》dcos package list --json

 Connection refused: master.mesos/10.10.0.166:8443 at remote address: master.mesos/10.10.0.166:8443. Remote Info: Not Available
 1
package list command returned code:1, stderr:Connection refused: master.mesos/10.10.0.166:8443 at remote address: master.mesos/10.10.0.166:8443. Remote Info: Not Available
, stdout: 
```

wavesoft: Search it here: https://jenkins.mesosphere.com/service/jenkins/view/Marathon-Dashboard/job/system-integration-tests/job/marathon-si-dcos-strict/job/master/2155/console

zen-dog: `Connection refused: master.mesos/10.10.0.166:8443` indicates that `cli` is trying to connect to Marathon and fails. Looks like this test is somehow affected by the previous test which abdicated marathon leader. And though we're waiting for Marathon to become available on all masters before each test that seems to not be enough? 🤔 
Retrying seems like an easy fix but it is likely to mask the actual issue underneath. Nevertheless, we might merge it.

wavesoft: If an agent is down, this _cannot_ be marathon's fault. Therefore in the context of "marathon" system tests, such an error is irrelevant.

In this logic, if the agent is *temporarily* inaccessible could be a recoverable DC/OS problem, which is totally expected in a distributed system. This "temporarily" check is implemented using the "retry" statement...

I **DO** agree that such an error is important on the overall DC/OS product and therefore must be forwarded to the respective team. And thus it **is** important to be able to collect enough logs for further investigation. But again, this is not a marathon-related issue.

wavesoft:
> Sorry but please use eventually instead. We should try to remove most uses of retrying in favor of assert_that(..., eventually()). deployment_wait in common.py is a good example.

Sure, but I literally just copied the statement we had in that file, and I wasn't aware that we are now going towards an `eventually` approach. Is there any JIRA that tracks this as tech-debt?

jeschkies:
> If that's the plan, then we should change this everywhere... Is there any JIRA that tracks this as tech-debt?

Over time, yes. Every new code should use the `eventually` statement.

wavesoft:
> I still think it should be up to the caller to decide whether to retry or not

@jeschkies I totally agree with you. My only concern was that in my PoV the `is_installed` and `install` exceptions have different weight and therefore re-trying on both exceptions did not seem like a good idea to me.

That's why I differentiated them and tried your approach. So, I moved the `is_installed -> install` block of code into the `common.could_install_enterprice_cli`, but made it throw a different kind of exception on `Install` errors.

I then changed all the `if_installed -> install` occurrences with :

```python
    assert_that(True, eventually(common.could_install_enterprice_cli(),
                                 retry_on_exception=common.ignore_other_exception(common.InstallException)))
```

tests/system/test_marathon_on_marathon_ee.py
```diff
@@ -149,9 +150,8 @@ def simple_sleep_app(name):
 
 
 def ensure_prerequisites_installed():
-    if not common.is_enterprise_cli_package_installed():
-        common.install_enterprise_cli_package()
-    assert common.is_enterprise_cli_package_installed()
+    assert_that(True, eventually(common.could_install_enterprice_cli(),
+                                 retry_on_exception=common.ignore_other_exception(common.InstallException)))
```
jeschkies: See above

zen-dog:
Closed since #6636 landed and seemed to fix the problem

PR #6509, opened 2018-09-04
Time benchmarks steps.

jeschkies:
Summary:
We would like to know how long the build, cluster start and benchmark
steps are taking. We add timestamps when each step starts.


wavesoft:
Btw, you can get the perf run duration checking the JSON dump. There is a "startTime" and "endTime" fields ;)

PR #6510, opened 2018-09-04
Enable marathon to run as non-root

wavesoft:
Summary:
Currently, Marathon is built as a root user in docker, and therefore makes it impossible for non-root users to be able to run it. This commit make sure that everybody can read and execute the files in the /marathon directory, by changing the default owner to nouser:nogoup

JIRA issues: MARATHON-8409


timcharper:
@wavesoft  approved; please update the changelog though

wavesoft:
Added changelog. Will merge when build passes 👍

PR #6512, opened 2018-09-05
Introduce experimental `TemplateRepository`

zen-dog:
Summary:
Introduced `TemplateRepository` which is a using recently introduced `ZookeeperPersistenceStore` and provides a CRUD-like interface to store/fetch templates objects (which are basically either `AppDefinition`s or `PodDefinition`s for now). 

New trait `Template[T]`:
- It is a common denominator for the existing app and pod definitions (and all possible future specifications)
- It only requires a way to decode the spec to `Array[Byte]` and back plus a `PathId` (which is used by the repo to determine the path in the store)

New `TemplateRepository` class that:
- implements `TemplateRepositoryLike` trait
- provides CRUD methods for templates. Reading methods e.g. `read`, `contents` and `exists` are served directly from memory while writing methods e.g. `delete`, `create`are asynchronous and update the underlying storage first
- to make synchronous methods possible it holds all templates in memory in a _trie_ like structure: `PathTrie`
- adds `TemplateRepositoryBenchmark` which covers CRUD methods

New `PathTrie` class that:
- is a modified copy of Zookeepers own [PathTrie](http://people.apache.org/~larsgeorge/zookeeper-1215258/build/docs/dev-api/org/apache/zookeeper/common/PathTrie.html) class
- it's an ordered tree data structure that keeps where the key is template's `PathId` split by path segments. Looking up data in a _PathTrie_ is in the worst case O(m) time (where m is number of path segments)
- extends the original class with methods of traversing the trie and fetching specific nodes
- adds optional data fields to trie nodes
**Note:** `PathTrie` is implemented in Java, since it's copy-modified from Zookeeper. It can (and probably) should be reimplemented in Scala (using `PathId`) but the current implementation is sufficient for now since that branch of code is still experimental.

Related JIRA: DCOS-41405


benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
```diff
@@ -0,0 +1,202 @@
...
+import mesosphere.marathon.core.storage.repository.TemplateRepositoryLike.{Spec, Template}
+import mesosphere.marathon.core.storage.repository.impl.{CachedTemplateRepository, TemplateRepository}
+import mesosphere.marathon.core.storage.zookeeper.{AsyncCuratorBuilderFactory, AsyncCuratorBuilderSettings, ZooKeeperPersistenceStore}
+import mesosphere.marathon.metrics.Metrics
+import mesosphere.marathon.metrics.dummy.DummyMetrics
+import mesosphere.marathon.state.{AppDefinition, PathId}
+import org.apache.curator.framework.{CuratorFramework, CuratorFrameworkFactory}
+import org.apache.curator.retry.BoundedExponentialBackoffRetry
+import org.apache.curator.x.async.api.CreateOption
+import org.openjdk.jmh.annotations._
```
ichernetsky: nit: no to `_` imports, if possible.

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
```diff
@@ -0,0 +1,202 @@
...
+import org.openjdk.jmh.infra.Blackhole
+
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, ExecutionContext}
+import scala.util.Random
+
+/**
+  * To run e.g. [[TemplateRepositoryBenchmark#create()]] benchmark execute from the console:
+  * $ sbt "benchmark/clean" "benchmark/jmh:run  mesosphere.marathon.core.storage.repository.TemplateRepositoryBenchmark.create"
```
ichernetsky: nit: two spaces after `benchmark/jmh:run`

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
```diff
@@ -0,0 +1,202 @@
...
+  //  lazy val repository: CachedTemplateRepository = new CachedTemplateRepository(store, cacheMaximumSize = 65535)
+  lazy val repository: TemplateRepository = new TemplateRepository(store)
+
+  val random = new Random()
+
+  def randomPathId = PathId(s"/sleep-${random.alphanumeric.take(8).mkString}")
+  def randomPathId(prefix: String) = PathId(s"/$prefix/sleep-${random.alphanumeric.take(8).mkString}")
+  def randomRangedPathId(prefix: String, maxRange: Int) = PathId(s"/$prefix/sleep-${random.nextInt(maxRange) + 1}")
+  /**
+    * Return an [[AppDefinition]] with give pathId. We simulate big app definitions by creating a label with a given value length.
```
ichernetsky: nit: `with the given pathId`

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
```diff
@@ -0,0 +1,202 @@
...
+    * The total size of the app definition will be approx. label size + 159 bytes where 159bytes is the size of the serialized
```
ichernetsky: nit: `159 bytes is`

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
```diff
@@ -0,0 +1,202 @@
...
+    * @param pathId
+    * @param labelSize
+    * @return
+    */
+  def spec(pathId: PathId, labelSize: Int = 1): Spec = AppDefinition(id = pathId, labels = Map("a" -> random.alphanumeric.take(labelSize).mkString))
+
+  def template(pathId: PathId, version: Int, labelSize: Int = 1): Template = Template(spec(pathId, labelSize), version)
+  def randomTemplate: Template = template(randomPathId, 1)
+
+  // An map of node size to number of nodes of that size. Used for read, update and delete benchmarks. Note that
```
ichernetsky: nit: `A map`

src/main/scala/mesosphere/marathon/core/storage/repository/BucketingRepository.scala
```diff
@@ -0,0 +1,162 @@
...
+import akka.stream.scaladsl.{Flow, Source}
+import akka.util.ByteString
+import mesosphere.marathon.core.storage.zookeeper.PersistenceStore.Node
+import mesosphere.marathon.core.storage.zookeeper.ZooKeeperPersistenceStore
+
+import scala.concurrent.Future
+import scala.util.Try
+
+/**
+  * Fun fact about Zookeeper: you can create a lot of children nodes but if you  try to get all of them by calling
```
ichernetsky: nit: two spaces in `if you try`

src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryLike.scala
```diff
@@ -0,0 +1,58 @@
...
+  case class Template(spec: Spec, version: Int) extends BucketNode {
+
+    def pathId = spec.id
+
+    override def bucketPath: BucketPath = VersionBucketPath(pathId, version)
+
+    override def payload: ByteString = ByteString(spec.toProtoByteArray)
+  }
+
+  object Template {
```
ichernetsky: I assume this will be extended to support pods and all in the future?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/CachedTemplateRepository.scala
```diff
@@ -0,0 +1,91 @@
...
+  * method e.g. a template with pathId=/foo/bar and version=1 will be saved using the key `foo_bar:1`.
+  *
+  * Basically this class wraps [[TemplateRepository]] methods with the calls to cache. [[read()]] method goes directly to cache
+  * and the element will be loaded if not already exists.
+  *
+  * Note that this implementation is tread-safe and provides serializability guarantees e.g. two independent [[create()]]
+  * calls are not guaranteed to be executed in any particular order the same way one should not expect two independent futures
+  * to have an order.
+  *
+  * Known pitfalls: an operation e.g. [[delete()]] may fail leaving the cache and the underlying store in an inconsistent
```
ichernetsky: Is it possible to workaround this inside the implementation so that it is hidden from a user?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/CachedTemplateRepository.scala
```diff
@@ -0,0 +1,91 @@
...
+    cache.put(toCacheKey(versioned), Future.successful(versioned))
+    versioned
+  }
+
+  override def read(pathId: PathId, version: Int): Future[Template] =
+    cache.get(toCacheKey(pathId, version))
+
+  override def delete(pathId: PathId): Future[Done] = async {
+    val versions = await(super.versions(pathId)) // Get all existing template versions
+    val deleted = await(super.delete(pathId)) // Delete template from the store
```
ichernetsky: How is it ensured that there is no race condition in situations like this: getting `versions`, and then deleting the template, when a new version is created in the meantime?

zen-dog: I'm glad someone is paying attention! I was thinking about this scenario and tried to synchronize this through `scaffeine` cache, but unfortenately it seems to be  impossible due to _scaffeine_ implementation details (`AsyncLoadingCache` doesn't work together with deleting/updating elements through cache).

To be precise about the problem at hand: assuming we have 2 independent operations: `create(/foo)` and `delete(/foo)` there are two trivial cases where either `a) create() -> delete()` or `b) delete() -> create()` happen after previous operation completed successfully. Here we either end up with `a)` everything deleted or `b)` a new version of template is created (with version counter started with `1` again). Both cases are completely legal.

Now to not-trivial cases where both operations interleave:
- `delete()` has 3 major steps:
  - (1) <-
  - get all existing `versions`
  - (2) <-
  - `delete` parent `/foo` node (and all templates versions in it)
  - (3) <-
  - `invalidateAll` versions from the cache

A concurrent `create(/foo)` call can happen in `(1), (2), (3)` places. Let's see what will happen then:
1. If `create` call happens here, the resulting version will be simply deleted with everything else in the subsequent steps *in the ZK*. However since `create` is also 2-steps operation (store in ZK and store in cache) we might end up with a cached version even after `delete` has finished.
2. Same logic applies so that we might end up with a version in cache but not in ZK. 
3. Here we already deleted `/foo` node from the ZK, so a new template version will be stored successfully. However the subsequent caching might be invalidated, so we'll end up with a version in ZK but not in cache. This is however fine, since cache will load it on the next _read_ request.

So, as we can see, in cases `(1)` and `(2)` we might end up with a version in cache which does not exist in the store (which is clearly bad). However I haven't found a lock-free solution to the problem (synchronizing everything is an "easy fix").

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
timcharper: I'm generally inclined to agree. This is benchmark code, though, and this is for annotations. I might let this one slide.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
```diff
@@ -0,0 +1,174 @@
...
+  * [[update()]] method is overridden and throws an [[UnsupportedOperationException]] since templates are immutable and should not
+  * be updated.
+  *
+  * [[init()]] method initializes repository by reading all existing templates Ids and their versions (but not template content) and
+  * initializing internal template version counters. This method should be called and waited for successful competition prior to any
+  * other usage of the repository.
+  *
+  * Known pitfalls: please note that while there is one at least one version of the template, a counter is always increased monotonically.
+  * However should all template versions be deleted and marathon restarted, a counter is initialized with `0`.
+  * This implementation is tread-safe and provides serializability guarantees e.g. two independent [[create()]] calls are not guaranteed
```
timcharper: But two serial creates are run serially, right?

timcharper: BTW what width are you using? I'm showing... 127-130?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/CachedTemplateRepository.scala
timcharper: Yeah, I don't think there is an "easy fix".

This is one reason I proposed the in-memory state authority which had a primary job to enforce foreign key constraints. It's effectively synchronizing everything, but, performing each serial operation in memory. Multi-entity deletes like this could be done in a single frame, and subsequent modifications could be queued up in a zookeeper transaction.

timcharper: An optimistic lock strategy may be possible? Just spitballing here:

1) Query the current template version as stored
2) Query the versions
3) Delete / invalidate / etc
4) Query the current template version again. If it changed from step 1, then repeat steps

zen-dog: > An optimistic lock strategy may be possible?

Well, yeah, it might, but I'm not looking for complete serializability of operations. E.g. in the above example, `delete(/foo)` will always try to clean everything before it finishes even if a concurrent `create(/foo)` operation exists. What I'm aiming for instead, is for both operations to be atomic where `a) create() -> delete()` or `b) delete() -> create()` can have any order (should be enforced by the caller and not by the repo). 
The important part is that both operations don't interleave and/or leave a consistent cache/store behind. `synchronize` seems to be the easiest fix, but I was hoping for a more elegant idea.

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
zen-dog: There are 12 imported annotations there. Is it really better to import them one by one? But if you insist...

src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryLike.scala
zen-dog: Yes, in the future `Spec` will be a new type (combining best of both apps- and pods-worlds). For now, it's easier to test with apps locally so it is an `AppDefinition`.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/CachedTemplateRepository.scala
zen-dog: I was referring to a specific use case where an error occurred during store's `delete` operation, meaning that _most probably_ the ZK connection is gone and this situation will be handled on the "higher application level". Otherwise, I don't see how to meaningfully handle this kind of errors on this level.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
zen-dog: > But two serial creates are run serially, in order, right?

Well, yes, when called from the same thread it will because of the lock-free implementation using the atomic counter:
```
// In order:
Thread-1:
repo.create(/foo) <- version 1
repo.create(/foo) <- version 2
```

However, I wanted to give the same guaranties for `TemplateRepository` and it's cached version `CachedTemplateRepository` where:
```
//No order:
async {
  cachedRepo.create(/foo) <-- version 1 or 2
  cachedRepo.create(/foo) <-- version 1 or 2
}

// In order again with await
async {
  await(cachedRepo.create(/foo)) <-- version 1
  await(cachedRepo.create(/foo)) <-- version 2
}
```

src/main/scala/mesosphere/marathon/core/storage/repository/impl/CachedTemplateRepository.scala
zen-dog: Just thinking out loud here - we will not have this problem if deleting a node in ZK would return a list of deleted children (or `[]` otherwise). Then the method above could be:
```scala
  override def delete(pathId: PathId): Future[Done] = async {
    val versions = await(super.delete(pathId)) // Delete template from the store and return versions
    cache.synchronous().invalidateAll(versions.map(toCacheKey(pathId, _))) // Invalidate all cached versions
    Done
  }
```
Here we would always only invalidate versions that were previously deleted, but ZK doesn't have this functionality... 🤔

meln1k: What about plain old `syncronized`?

benchmark/src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryBenchmark.scala
jeschkies: I side with Ivan. It's not easy to track the annotations and it's less than a dozen 😉 .

src/main/scala/mesosphere/marathon/core/storage/repository/BucketingRepository.scala
```diff
@@ -0,0 +1,162 @@
...
+  */
+trait BucketingRepository {
+
+  def store: ZooKeeperPersistenceStore
+
+  /**
+    * Max buckets number should never be changed in production since all data location will change and
+    * re-hashing will be needed
+    */
+  val MaxBuckets = 16
```
jeschkies: How come 16 is enough? How many templates can we get with 16 buckets?

src/main/scala/mesosphere/marathon/core/storage/repository/BucketingRepository.scala
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+    * Map of bucket numbers -> bucket name e.g. (5 -> "05")
+    */
+  val buckets: Map[Int, String] = (0 to MaxBuckets).map(b => b -> f"$b%02d")(collection.breakOut)
+
+  /**
+    * Base path to all stored elements e.g. /templates
+    */
+  val base: String = "/base"
```
jeschkies: Should this get a default?

src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryLike.scala
```diff
@@ -0,0 +1,58 @@
+package mesosphere.marathon
+package core.storage.repository
```
jeschkies: Hm, I have to admit that I'm not sure if `TemplateRepository` and `BucketRepository` should be in the same package. The former is a different semantic meaning. Templates are unique to Marathon. The later is universal and not unique to Marathon. I'd put it with our ZooKeeper client files.

src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryLike.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import scala.concurrent.Future
+
+trait TemplateRepositoryLike {
+
+  import TemplateRepositoryLike._
+
+  def init(): Future[Done]
+
+  def create(template: Spec): Future[Template]
```
jeschkies: This is an odd signature, why would I pass a different type then the return type in a CRUD?

src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryLike.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def exists(pathId: PathId, version: Int): Future[Boolean]
+}
+
+object TemplateRepositoryLike {
+
+  /**
+    * A Spec is simply a [[mesosphere.marathon.state.AppDefinition]] as long as we don't have the new specification.
+    */
+  type Spec = AppDefinition
```
jeschkies: Out of curiosity, why are we not using a generic class?

src/main/scala/mesosphere/marathon/core/storage/repository/TemplateRepositoryLike.scala
```diff
@@ -0,0 +1,58 @@
...
+    * override each field, the spec itself and the version are wrapped in another case class. The important parts are the
+    * [[BucketNode]] trait methods implementation which define where and what data will be saved.
+    */
+  case class Template(spec: Spec, version: Int) extends BucketNode {
+
+    def pathId = spec.id
+
+    override def bucketPath: BucketPath = VersionBucketPath(pathId, version)
+
+    override def payload: ByteString = ByteString(spec.toProtoByteArray)
```
jeschkies: This seems to be a tight coupling between Specs, Storage and serialization. Jason went a little too far but this seems a little to restraining to me. What do you think? Should we not rather inject and serialization? How could we eg migrate to Kryo?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
```diff
@@ -0,0 +1,174 @@
...
+    * This methods should be called (and waited for successful execution) prior to any other repository method.
+    */
+  def init(): Future[Done] = {
+    Source
+      .fromFuture(store.exists(base)) // Fetch existing buckets if any exist
+      .mapAsync(1)(ex => if (ex) store.children(base) else Future.successful(Seq.empty))
+      .mapConcat(identity(_))
+      .via(store.childrenFlow) // Fetch existing templates from all buckets
+      .mapConcat(identity(_))
+      .via(store.childrenFlow) // Fetch versions for all templates
```
jeschkies: Why is `.via(store.childrenFlow)` duplicated? Is it because we fetch the grand-children?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
```diff
@@ -0,0 +1,174 @@
...
+
+  import TemplateRepository._
+
+  override val base: String = "/templates"
+
+  /**
+    * We store a counter for the template versions separately. It's helpful when calculating next template
+    * version. A counter is initialized by the current template max version number in the [[init()]] method.
+    */
+  val counters = new ConcurrentHashMap[PathId, AtomicInteger]()
```
jeschkies: Why not set the counter after `init`? There current implementations hides accesses to `counters` before it was initialized. `var counters = _` seems ugly but would surface an init issue with a `NullPointerException`. Also, why is the value an `AtomicInteger` if we use `compute`?

src/main/scala/mesosphere/marathon/core/storage/repository/BucketingRepository.scala
```diff
@@ -0,0 +1,162 @@
...
+  def children(bucketPath: BucketPath): Future[Seq[String]] = store.children(path(bucketPath))
+  def children(bucketPaths: Seq[BucketPath]): Source[Seq[String], NotUsed] = Source(bucketPaths).via(childrenFlow)
+  def childrenFlow: Flow[BucketPath, Seq[String], NotUsed] =
+    Flow[BucketPath]
+      .via(toPath)
+      .via(store.childrenFlow)
+
+  def exists(bucketPath: BucketPath): Future[Boolean] = store.exists(path(bucketPath))
+  def exists(bucketPaths: Seq[BucketPath]): Source[Boolean, NotUsed] = Source(bucketPaths).via(existsFlow)
+  def existsFlow: Flow[BucketPath, Boolean, NotUsed] =
```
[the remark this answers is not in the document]
zen-dog: What do you have in mind for `T`?

src/main/scala/mesosphere/marathon/core/storage/repository/impl/CachedTemplateRepository.scala
zen-dog: Alternatively, we could delete all versions *in one transaction*:
```scala
  override def delete(pathId: PathId): Future[Done] = async {
    val versions = await(super.versions(pathId)) // Get all existing template versions
    val deleted = await(super.transaction(versions.map(super.deleteOp(_)) // A bit of pseudocode: delete all verasions from the store in a transaction
    cache.synchronous().invalidateAll(versions.map(toCacheKey(pathId, _))) // Invalidate all cached versions
    Done
  }
```
but this would leave a dangling parent node in the store.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
```diff
@@ -0,0 +1,174 @@
...
+        counters.compute(pathId, (_, value) => if (value == null) new AtomicInteger(maxVersion) else { value.getAndSet(maxVersion); value })
+      }
+      .runWith(Sink.ignore)
+  }
+
+  def create(spec: Spec): Future[Template] = {
+    val version = counters.compute(spec.id, (key, value) => // New template version is calculated atomically
+      if (value == null) new AtomicInteger(0)
+      else value
+    ).incrementAndGet()
```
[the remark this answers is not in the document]
zen-dog: From the `putIfAbsent` docs:
```
If the specified key is not already associated with a value (or is mapped to null) associates it with the given value and returns null, else returns the current value
```
it's this `returns null` semantic that I will have to deal with - `compute` is a somewhat longer one-liner but I don't have to check the result.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
zen-dog: Exactly, it's buckets first, templates second and versions of the templates third.

src/main/scala/mesosphere/marathon/core/storage/repository/BucketingRepository.scala
zen-dog: See the [calculation](https://github.com/mesosphere/marathon/pull/6512/files#diff-78323e9f9f91a172f1d32a4c999554baR34) above - depending on the child's name length it's roughly 16 * 75k nodes.

src/main/scala/mesosphere/marathon/core/storage/repository/BucketingRepository.scala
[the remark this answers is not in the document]
meln1k: I'd leave it for the user. It's always possible to pass `NotUsed` if no one really needs it.

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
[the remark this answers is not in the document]
meln1k: Oh, what an interface! Compute sounds good then :)

src/main/scala/mesosphere/marathon/core/storage/repository/impl/TemplateRepository.scala
```diff
@@ -0,0 +1,174 @@
...
+      .delete(VersionBucketPath(pathId, version))
+      .map(_ => Done)
+
+  def delete(pathId: PathId): Future[Done] =
+    super
+      .delete(TemplateBucketPath(pathId))
+      .map(_ => counters.remove(pathId))
+      .map(_ => Done)
+
+  override def update(bucketNode: BucketNode): Future[String] = throw new UnsupportedOperationException("Updating templates is not permitted")
```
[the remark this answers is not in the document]
zen-dog: Good point. Will rework this part.

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
```diff
@@ -0,0 +1,178 @@
...
+  def path(bucketPath: BucketPath): String = Paths.get(base, bucket(bucketPath), bucketPath.relativePath).toString
+
+  /**
+    * Helper stages, mapping from BucketNode/BucketPath to Node/Path
+    */
+  def toNode: Flow[BucketNode, Node, NotUsed] = Flow[BucketNode].map(bn => Node(path(bn), bn.payload))
+  def toPath: Flow[BucketPath, String, NotUsed] = Flow[BucketPath].map(bp => path(bp))
+
+  // All methods below wrap the storage methods by mapping from BucketNode/BucketPath to Node/Path
+  def create(bucketNode: BucketNode): Future[String] = store.create(Node(path(bucketNode), bucketNode.payload))
```
meln1k: What do you think about some documentation for the return type?

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
```diff
@@ -0,0 +1,178 @@
...
+      .via(store.createFlow)
+
+  def read(bucketPath: BucketPath): Future[Try[Node]] = store.read(path(bucketPath))
+  def read(bucketPaths: Seq[BucketPath]): Source[Try[Node], NotUsed] = Source(bucketPaths).via(readFlow)
+  def readFlow: Flow[BucketPath, Try[Node], NotUsed] =
+    Flow[BucketPath]
+      .via(toPath)
+      .via(store.readFlow)
+
+  def update(bucketNode: BucketNode): Future[String] = store.update(Node(path(bucketNode), bucketNode.payload))
```
meln1k: What do you think about some documentation for the return type?

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
```diff
@@ -0,0 +1,178 @@
...
+  def update(bucketNodes: Seq[BucketNode]): Source[String, NotUsed] = Source(bucketNodes).via(updateFlow)
```
meln1k: What do you think about some documentation for the return type?

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
```diff
@@ -0,0 +1,178 @@
...
+  def updateFlow: Flow[BucketNode, String, NotUsed] =
+    Flow[BucketNode]
+      .via(toNode)
+      .via(store.updateFlow)
+
+  def delete(bucketPath: BucketPath): Future[String] = store.delete(path(bucketPath))
```
meln1k: What do you think about some documentation for the return type?

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
```diff
@@ -0,0 +1,178 @@
...
+
+  def children(bucketPath: BucketPath): Future[Seq[String]] = store.children(path(bucketPath), true)
+  def children(bucketPaths: Seq[BucketPath]): Source[Seq[String], NotUsed] = Source(bucketPaths).via(childrenFlow)
+  def childrenFlow: Flow[BucketPath, Seq[String], NotUsed] =
+    Flow[BucketPath]
+      .via(toPath)
+      .via(store.childrenFlow(true))
+
+  def exists(bucketPath: BucketPath): Future[Boolean] = store.exists(path(bucketPath))
+  def exists(bucketPaths: Seq[BucketPath]): Source[Boolean, NotUsed] = Source(bucketPaths).via(existsFlow)
```
meln1k: Don't you think that `Source[Boolean, NotUsed]` can be replaced with `Future[Seq[Boolean]]`?

project/Dependencies.scala
```diff
@@ -61,6 +61,7 @@ object Dependencies {
     commonsCompress % "compile", // used for tar flow
     commonsIO % "compile",
     alpakkaCodes % "compile",
+    scaffeine % "compile",
```
timcharper: this dependency is gone, right?

src/main/scala/mesosphere/marathon/experimental/repository/TemplateRepositoryLike.scala
```diff
@@ -0,0 +1,115 @@
+package mesosphere.marathon
+package experimental.repository
+
+import akka.Done
+import mesosphere.marathon.state.PathId
+
+import scala.concurrent.Future
+import scala.util.Try
+
+trait TemplateRepositoryLike {
```
timcharper: Since we're planning on merging pods apps in the new template repository, maybe it would be better to create an interim type PodOrSpec instead of introducing this Generic parameter?

zen-dog: We're planning to introduce a new type (a real template) which will combine both. It might happen that we'll introduce an archetype for all of them but for now, the easiest way was to duck-type them together.

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
zen-dog: I've changed the return type to a `(path: String, exists: Boolean)` to make it easier to know which path exists.

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
zen-dog: The class was removed

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
zen-dog: Removed

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
zen-dog: Removed

src/main/scala/mesosphere/marathon/experimental/storage/BucketingPersistenceStore.scala
zen-dog: Removed

src/main/java/mesosphere/marathon/experimental/storage/PathTrie.java
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+
+    /**
+     * return leaf nodes recursively starting with the given path.
+     *
+     * @param path input path
+     * @return a map with leaf nodes
+     */
+    public Set<String> getLeafs(String path) {
```
jeschkies: Isn't this `findLeafNodes(path, getNode(path), new HashSet<>())`? I would call the the same.

src/main/scala/mesosphere/marathon/core/storage/zookeeper/PersistenceStore.scala
```diff
@@ -53,18 +53,20 @@ trait PersistenceStore {
     *
     * @return
     */
-  def childrenFlow: Flow[String, Seq[String], NotUsed]
-  def children(path: String): Future[Seq[String]]
-  def children(paths: Seq[String]): Source[Seq[String], NotUsed] = Source(paths).via(childrenFlow)
+  type Children = Seq[String]
```
jeschkies: Should be type the Children name? Such as `PathComponent`, `Segment` or `Path`?

Also, let's say I have

```
/aws
  /prod
  /dev
    /ci
      /123
```

What is the content of `children("/aws/dev")`? Is it `Seq("/aws/dev/ci", "/aws/dev/ci/123")` or `Seq("ci", "123")`?

src/main/scala/mesosphere/marathon/core/storage/zookeeper/PersistenceStore.scala
```diff
@@ -79,6 +81,14 @@ trait PersistenceStore {
     * and submits them. An exception is thrown if one of the operations fail.
     */
   def transaction(operations: Seq[StoreOp]): Future[Done]
+
+  /**
+    * Create a node if none exists with the given path. Return the path of the node. This operation is atomic.
```
jeschkies: Should we provide more feedback? Maybe I want to know if it *did* exist.

src/main/scala/mesosphere/marathon/experimental/repository/TemplateRepository.scala
```diff
@@ -0,0 +1,184 @@
...
+    await(tree(base))
+    await(data())
+  }
+
+  override def create(template: Template[_]): Future[String] = {
+    store
+      .create(toNode(template))
+      .map(_ => trie.addPath(storePath(template), template.toProtoByteArray))
+      .map(_ => version(template))
+  }
```
jeschkies: I think we can avoid the creation of two futures with

```scala
  override def create(template: Template[_]): Future[String] = async {
    await(store.create(toNode(template)))
    trie.addPath(storePath(template), template.toProtoByteArray)
    version(template)
  }
```

This would also express that `trie.addPath` is not asynchronous.

src/main/scala/mesosphere/marathon/experimental/repository/TemplateRepository.scala
```diff
@@ -0,0 +1,184 @@
...
+      template)
+  }
+
+  override def delete(pathId: PathId): Future[Done] = delete(pathId, version = "")
+  override def delete(pathId: PathId, version: String): Future[Done] = {
+    store
+      .delete(storePath(pathId, version))
+      .map(trie.deletePath(_))
+      .map(_ => Done)
+  }
```
jeschkies: See my comment above

```scala
  override def delete(pathId: PathId, version: String): Future[Done] = async {
    val p = await(store.delete(storePath(pathId, version)))
    trie.deletePath(p)
    Done
  }
```

src/main/scala/mesosphere/marathon/experimental/repository/TemplateRepository.scala
```diff
@@ -0,0 +1,184 @@
...
+  def exists(pathId: PathId): Boolean = trie.existsNode(storePath(pathId))
+
+  def exists(pathId: PathId, version: String) = trie.existsNode(storePath(pathId, version))
+}
+
+object TemplateRepository {
+
+  /**
+    * Helper method that take an underlying zookeeper storage and the base prefix and returns the template
+    * repository wrapped in a future.
```
jeschkies: Nit: `returns the initialized repository`

benchmark/src/main/scala/mesosphere/marathon/core/experimental/repository/TemplateRepositoryBenchmark.scala
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+
+  val repository = new FixedVersionTemplateRepository(store)
+  Await.result(new FixedVersionTemplateRepository(store).initialize(), Duration.Inf)
+
+  val random = new Random()
+
+  def randomPathId(prefix: String) = PathId(s"/$prefix/sleep-${random.alphanumeric.take(8).mkString}")
+  def randomRangedPathId(prefix: String, maxRange: Int) = PathId(s"/$prefix/sleep-${random.nextInt(maxRange) + 1}")
```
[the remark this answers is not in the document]
zen-dog: We could add one at some point. In praxis, I've never seen customers using > 5 path nesting levels so I consider this a low prio

src/main/java/mesosphere/marathon/experimental/storage/PathTrie.java
```diff
@@ -0,0 +1,371 @@
...
+     */
+    TrieNode getNode(String path) {
+        if (path == null) {
+            return null;
+        }
+        if ("/".equals(path)) {
+            return rootNode;
+        }
+        String[] pathComponents = path.split("/");
+        if (pathComponents.length <= 1) {
```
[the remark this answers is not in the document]
zen-dog: No, it doesn't but in praxis all paths are treated as absolute

src/main/scala/mesosphere/marathon/core/storage/zookeeper/PersistenceStore.scala
zen-dog: > Should be type the Children name? Such as `PathComponent`, `Segment` or `Path`?

I don't think I understand the question. Mostly I wanted to avoid `Source[Try[Seq[String]` in the method signature

> What is the content of `children("/aws/dev")`?

it's always the direct children of the node (that's how ZK returns it) so only `/ci` and if you want the absolute path (`absolute = true`) it is `/aws/dev/ci`

src/main/scala/mesosphere/marathon/core/storage/zookeeper/PersistenceStore.scala
zen-dog: A similar _putIfAbsent_ method in the _ConcurrentHashMap_ [returns](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html#putIfAbsent(K,%20V)) value associated with the key if it exists or _null_ otherwise. However, in our case, only _"key"_ is being created so I'm not sure if this differentiation provides any additional value for the user. At the very least it will make the method signature inconsistent with other storage methods.

src/main/scala/mesosphere/marathon/core/storage/zookeeper/PersistenceStore.scala
jeschkies: > Should be type the Children name? Such as `PathComponent`, `Segment` or `Path`

I meant that we should type them. So `Seq[Segment]` instead of `Seq[String]`.

zen-dog: Noo, then e.g. `create` should also take a change in the signature. In general this layer deals with strings (for keys) and bytestrings (for values). I think it's fine like this

jeschkies: I guarantee that this will lead to lots of bugs. I don't see why `Path` and `Path.segments: Seq[Segment]` is such a bad idea. I mean the whole thing is design around paths. `String` is nice and simple at the beginning. Well, We probably gonna use `PathId` anyways to templates...

zen-dog: No, this is definitely a wrong layer for introducing _PathIds_ or _Path.Segments_. This is the layer, closest to ZK, everything here is "native" e.g. _strings_ and _byte[]_. Pulling in higher level concerns in this layer would be a leaking abstraction:
```
 +------------------------------------+
 |            Zookeeper               |
 +------------------------------------+
                  |
                  |
 +----------------v-------------------+
 |         PersistenceStore           |
 |                                    |
 |  - uses apache curator             |
 |  - deals with ZK nodes,            |
 |    strings and byte[] payloads     |
 +------------------------------------+
                  |
                  |
                  |
 +------------------------------------+
 |          Repository                |
 |                                    |
 |   - uses PersistenceStore          |
 |   - deals with application level   |
 |     objects e.g. Templates         |
 |   - encode/decode to/from ZK nodes |
 |                                    |
 +------------------------------------+

```

PR #6513, opened 2018-09-06
Add logging to the `GroupManagerImpl`

zen-dog:
to cover a possibly fatal error.

Relates: MARATHON-8327


zen-dog:
> do we really want 1 letter variables?

a few of them are acceptable like `for i in...`, `t: Throwable` and so on.

timcharper:
To me, the complexity and size of the block of code that can reference the one letter variable is material. For simple 1 or 2 lines, I have no trouble following it.

PR #6517, opened 2018-09-10
Do not allow concurrent performance runs.

jeschkies:
Summary:
This should avoid the piling of performance benchmarks jobs in case the
NUC is busy.


jeschkies:
> Other places we schedule runs e.g. every 30min and also limit each run with timeout 30m so that they never overlap. But if this works better

Once the tests run faster we should do it. They do have a one hour timeout.

PR #6520, opened 2018-09-11
Counters of TASK_* status updates received (counter per task status update type)

ichernetsky:
JIRA issues: MARATHON-8411

### DIFF
--- a/docs/docs/metrics.md
+++ b/docs/docs/metrics.md
@@ -138,6 +138,9 @@ Dashes in metric names are replaced with underscores.
   received since the current Marathon instance became a leader.
 * `marathon.mesos.offers.used.counter` — the count of offers used since
   the current Marathon instance became a leader.
+* `marathon.mesos.task-states.<task_state>.counter` — the count of task
+  status updates received per each task state (`task-running`,
+  `task_failed` and so on).
 
 ### HTTP-specific metrics
 
--- a/docs/docs/metrics.md
+++ b/docs/docs/metrics.md
@@ -138,7 +138,7 @@ Dashes in metric names are replaced with underscores.
   received since the current Marathon instance became a leader.
 * `marathon.mesos.offers.used.counter` — the count of offers used since
   the current Marathon instance became a leader.
-* `marathon.mesos.task-states.<task-state>.counter` — the count of task
+* `marathon.mesos.task-updates.<task-state>.counter` — the count of task
   status updates received per each task state (`task-running`,
   `task-failed` and so on).
 
--- a/docs/docs/metrics.md
+++ b/docs/docs/metrics.md
@@ -138,9 +138,9 @@ Dashes in metric names are replaced with underscores.
   received since the current Marathon instance became a leader.
 * `marathon.mesos.offers.used.counter` — the count of offers used since
   the current Marathon instance became a leader.
-* `marathon.mesos.task-states.<task_state>.counter` — the count of task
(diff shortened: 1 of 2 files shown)


docs/docs/metrics.md
```diff
@@ -138,6 +138,9 @@ Dashes in metric names are replaced with underscores.
   received since the current Marathon instance became a leader.
 * `marathon.mesos.offers.used.counter` — the count of offers used since
   the current Marathon instance became a leader.
+* `marathon.mesos.task-states.<task_state>.counter` — the count of task
```
alenkacz: what do you think about `tasks.state` ?

btw. now looking at the other metrics I see we are not very consistent in this...

We have `marathon.mesos.offer-operations`
But also we have `marathon.mesos.offers.declined`

I would personally prefer
`marathon.mesos.offers.operation`
`marathon.mesos.offers.type`

But it's totally up for discussion :)

src/main/scala/mesosphere/marathon/core/task/update/impl/TaskStatusUpdateProcessorImpl.scala
```diff
@@ -41,11 +43,24 @@ class TaskStatusUpdateProcessorImpl @Inject() (
...
+    }
+    taskStateCounterMetrics.getOrElseUpdate(taskState.getNumber, createCounter)
+  }
+
   logger.info("Started status update processor")
 
   override def publish(status: MesosProtos.TaskStatus): Future[Unit] = oldPublishTimeMetric {
     newPublishTimeMetric {
       logger.debug(s"Received status update\n${status}")
+      getTaskStateCounterMetric(status.getState).increment()
```
alenkacz: can't this be called from multiple threads? If so, then the update of the metrics map is not thread safe...

But looking at the `MarathonScheduler` it should process only one callback at a time so we should be fine.

Do we maybe want to mention the fact that it's safe because of that somewhere?

docs/docs/metrics.md
ichernetsky: Though this is a possibility, we agreed that if parts of an entity name, that consists of two or more words, should be separated with dashes, not dots. In the light of this, say, `marathon.mesos.offers.incoming.counter` and `marathon.mesos.offer-operations.reserve.counter` do not seem to be inconsistent w.r.t. naming.

As for `marathon.mesos.task-states.<task-state>.counter`, I have been thinking of — besides the aforementioned one —

- `marathon.mesos.task-states.<task-state>.received.counter`
- `marathon.mesos.tasks.states.<task-state>.counter`
- `marathon.mesos.tasks.states.<task-state>.received.counter`

Which one do you like better? Another option is to use `incoming` instead of `received`.

timcharper: I like `received`.

Or...

`marathon.mesos.task-updates.<task-state>.counter` ?

docs/docs/metrics.md
```diff
@@ -138,6 +138,9 @@ Dashes in metric names are replaced with underscores.
   received since the current Marathon instance became a leader.
 * `marathon.mesos.offers.used.counter` — the count of offers used since
   the current Marathon instance became a leader.
+* `marathon.mesos.task-states.<task-state>.counter` — the count of task
```
timcharper: I'm more in favor of putting "updates" in the metric. `marathon.mesos.task-updates.<task-state>.counter`

PR #6524, opened 2018-09-11
Merge marathon-pkg into Marathon.

jeschkies:


timcharper:
If we squash this PR then the commit history will be gone; I think we should do a regular git merge.

tools/packager/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "marathon"]
```
timcharper: I think I'd slightly prefer to get rid of the submodule in this PR, but ¯\\\_(ツ)\_/¯

jeschkies: I was not sure either way. This is a verbatim import and the next pr does the clean up. Thanks for merging!

jeschkies:
> If we squash this PR then the commit history will be gone; I think we should do a regular git merge.

The PR will retain the history. And it's not our commit history. Furthermore, we loose reference to this PR and it is hard to revert.

PR #6526, opened 2018-09-13
Log app being locked as info

alenkacz:
I believe this can improve debug-ability while adding just few lines of logs.

The situation I see several times - you are trying to figure out from logs why some instance was not restarted, you see that scale check was initiated but no further messages and it's not apparent if it was because of the lock or not.

The lock might be held e.g. because someone initiated kill for an instance and that kill takes a long time to succeed (e.g. unkillable task). Then the app is locked and new instances are not coming up.

This does not solve that situation but helps it surface in logs.


kensipe:
👍

PR #6527, opened 2018-09-13
Populate PodStatus startedAt with instance timestamp

meichstedt:
Fallback to stagedAt in TerminationHistory when terminal task never started.

Summary:
task.status.startedAt is only populated when a task is reported running. For some reason, the startedAt field was added as a mandatory property to the TerminationHistory API type for PodStatus. This would lead to a hardcoded runtime exception when querying PodStatus for a pod for which an associated terminal task never turned running.
 Ideally, the property would be optional. Anticipating that tooling might expect the field to always be present, this change will populate the field with the time since the instance is in the current state (Created or Staging or Starting) instead.

JIRA issues: MARATHON-8428


meichstedt:
I feel like increasing my contribution by 1000%

PR #6530, opened 2018-09-14
Bump Jekyll dependencies

timcharper:
Summary:
Several Jekyll dependencies were out of date and this was causing known vulnerability messages to show in GitHub. Bumping this resolves it.

Jekyll was having issues with UTF-8 emdashes in our css files; we updated the Docker image to handle them.

Further, we clarify the docs and restore a section for simply rendering the docs on-the-fly.


timcharper:
Hmm, maybe in a future PR?

Also we may be able to use this: https://hub.docker.com/r/jekyll/jekyll/

jeschkies:
> Hmm, maybe in a future PR?

Sure.

> Also we may be able to use this: https://hub.docker.com/r/jekyll/jekyll/

That seems way too simple 😉

PR #6534, opened 2018-09-18
Use one queue to process LaunchQueue.Add commands.

jeschkies:
Summary:
The `LaunchQueueActor` will go with MARATHON-8167. However, until then we
are stuck with it. Due to its design there could be a case of concurrent
updates of its state. We could solve this either by introducing more
messages or with an internal queue. We opted for the second solution and
introduced a queue that guarantees sequential processing. It's ugly and
this diff just simplifies it to one queue. This should avoid errors in
the processing. Eg the current multi queue solution could stall and
ignore queued items.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
@@ -20,7 +20,7 @@ import mesosphere.marathon.core.task.tracker.InstanceTracker
 import mesosphere.marathon.state.{PathId, RunSpec}
 
 import scala.async.Async.{async, await}
-import scala.collection.immutable.Queue
+import scala.collection.mutable.Queue
 import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.control.NonFatal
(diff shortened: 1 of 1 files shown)


jeschkies:
I am happy for other suggestions but would like not to spent too much time on this as the whole actor will go anyways.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -244,66 +243,56 @@ private[impl] class LaunchQueueActor(
...
+        processNextAdd(queuedUpdates.dequeue())
+      }
 
     case msg @ RateLimiterActor.DelayUpdate(app, _) =>
       launchers.get(app.id).foreach(_.forward(msg))
   }
 
   @SuppressWarnings(Array("all")) /* async/await */
-  private def processNextAddIfExists(runSpec: RunSpec): Unit = {
+  private def processNextAdd(queuedItem: QueuedAdd): Unit = {
```
jeschkies: @alenkacz, could you explain the race condition to @zen-dog?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -244,66 +243,56 @@ private[impl] class LaunchQueueActor(
...
       // we cannot process more Add requests for one runSpec in parallel because it leads to race condition
       // the queue handling is helping us ensure we do only one per run-spec at a time
-      // requests for multiple runSpecs are still processed in parallel
 
-      val oldQueue: Queue[QueuedAdd] = updatesByRunSpecId(runSpec.id)
-      val newQueue = oldQueue :+ QueuedAdd(sender(), add)
-      updatesByRunSpecId += runSpec.id -> newQueue
-
-      if (oldQueue.isEmpty) {
+      if (queuedUpdates.isEmpty) {
```
jeschkies: This is wrong. We should check if there is an add in progress.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
alenkacz: @zen-dog if I am not mistaken, it was this one https://jira.mesosphere.com/browse/MARATHON-8320

PR #6536, opened 2018-09-18
HTTPS readiness checks do not check certificates

alenkacz:
This is re-implementation of https://github.com/mesosphere/marathon/commit/17d27b6acf3e202f5113a401ab39e176135efb46 for readiness checks.

This allows HTTPS readiness checks with e.g. self-signed certs.

JIRA issues: MARATHON-8430

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/readiness/impl/ReadinessCheckExecutorImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/readiness/impl/ReadinessCheckExecutorImpl.scala
@@ -102,7 +102,7 @@ private[readiness] class ReadinessCheckExecutorImpl(implicit actorSystem: ActorS
   private[impl] def akkaHttpGet(check: ReadinessCheckSpec): Future[AkkaHttpResponse] = {
     Timeout(check.timeout)(Http().singleRequest(
       request = RequestBuilding.Get(check.url),
-      connectionContext = ConnectionContext.https(HealthCheckWorker.disabledSslContext, sslConfig = Some(HealthCheckWorker.disabledSslConfig))
+      connectionContext = ConnectionContext.https(HealthCheckWorker.disabledSslContext, sslConfig = Some(HealthCheckWorker.disabledSslConfig()))
     ))
   }
 }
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/readiness/impl/ReadinessCheckExecutorImpl.scala
```diff
@@ -94,9 +98,38 @@ private[readiness] class ReadinessCheckExecutorImpl(implicit actorSystem: ActorS
       ReadinessCheckResult.forSpecAndResponse(check, response).copy(ready = false)
   }
 
+  private val disabledSslContext: SSLContext = {
```
timcharper: Can we extract the HealthCheckWorkerActor SSLContext and share instead?

src/main/scala/mesosphere/marathon/core/readiness/impl/ReadinessCheckExecutorImpl.scala
```diff
@@ -94,9 +98,38 @@ private[readiness] class ReadinessCheckExecutorImpl(implicit actorSystem: ActorS
...
+      .withDisableHostnameVerification(true)
+      .withDisableSNI(true)
+  })
+
   private[impl] def akkaHttpGet(check: ReadinessCheckSpec): Future[AkkaHttpResponse] = {
-    Timeout(check.timeout)(Http().singleRequest(
-      request = RequestBuilding.Get(check.url)
-    ))
+    if (check.url.startsWith("https")) {
+      Timeout(check.timeout)(Http().singleRequest(
```
timcharper: is it an error to set the connection context if https is not used?

timcharper: (answer seems to be no; in which case do we need this if statement?)

alenkacz: hmm maybe no, I was not sure, because this is what we do for health checks too :)

alenkacz:
@zen-dog I've tested manually but I am already working on SI test so that will follow in another PR

PR #6538, opened 2018-09-18
Build MoM to Soak without release to public universe

kensipe:
Summary:
Build MoM to Soak without release to public universe

JIRA issues:  MARATHON-8363


ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+ * Creates DCOS package for Marathon
+ * replaces __FILL_IN_VERSION_HERE__ with provided version
+ *
+ * @param packagingTool the registry build tool
+ * @param version used to identify the correct build file
+ */
+
+def buildDCOS(packagingTool: Path, version: SemVer) {
+  // ./dcos-registry-darwin registry migrate --package-directory=repo --output-directory=target
+  Try(%(packagingTool, "registry", "migrate", s"--package-directory=$UNIVERSE_DIR", s"--output-directory=$PACKAGE_DIR"))
```
jeschkies: What is the reason for wrapping it in a `Try`?

dcos/resource.json
```diff
@@ -0,0 +1,14 @@
+{
```
jeschkies: So this is the new registry package builder config? Is this really a template?

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+  buildDCOS(packagingTool, version)
+}
+
+/**
+ * Creates catalog package for Marathon with version info
+ * replaces __FILL_IN_VERSION_HERE__ with provided version
+ *
+ * @param version updated in the DCOS package files
+ */
+def createCatalogPackage(version: SemVer) {
```
jeschkies: What do you think about returning the path to the package instead of using a global constant?

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+ * @param packagingTool the registry build tool
+ * @param version used to identify the correct build file
+ */
+
+def buildDCOS(packagingTool: Path, version: SemVer) {
+  // ./dcos-registry-darwin registry migrate --package-directory=repo --output-directory=target
+  Try(%(packagingTool, "registry", "migrate", s"--package-directory=$UNIVERSE_DIR", s"--output-directory=$PACKAGE_DIR"))
+  val registryJSON = PACKAGE_DIR / s"marathon-${version.toReleaseString}.json"
+  // ./dcos-registry-darwin registry build --build-definition-file=target/marathon-1.7.50.json --output-directory=target --use-local
+  %(packagingTool, "registry", "build", s"--build-definition-file=$registryJSON", s"--output-directory=$PACKAGE_DIR", "--use-local")
```
jeschkies: How is `target/marathon-1.7.50.json` created? It says build definition. So is it something we provide or is it generated?

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+
+import $file.utils
+import utils.SemVer
+
+// output directory for all packages going to S3
+val PACKAGE_DIR: Path = pwd / 'target / 'universal
+// dcos catalog template directory
+val TEMPLATE_DIR: Path = pwd / 'dcos
+// versioned dcos catalog directory
+val UNIVERSE_DIR: Path = pwd / 'target / 'dcos
```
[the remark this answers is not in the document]
kensipe: this is to be run from marathon root... the pwd will be marathon resulting in `./target/dcos`.  nothing is written under `/ci`

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Creates catalog package for Marathon with version info
+ * replaces __FILL_IN_VERSION_HERE__ with provided version
+ *
+ * @param version updated in the DCOS package files
+ */
+def createCatalogPackage(version: SemVer) {
+  rm! UNIVERSE_DIR
+  mkdir! UNIVERSE_DIR
```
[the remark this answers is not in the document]
kensipe: I was just keeping significant variables used for build structure in the same place.   I will change this based on feedback.

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+
+  // list all files in temp
+  TEMPLATE_DIR.toIO.listFiles.foreach(file => {
+    val catalogFile = UNIVERSE_DIR / file.getName
+
+    Source.fromFile(file).getLines.map {
+      x => if(x.contains("__FILL_IN_VERSION_HERE__")) x.replace("__FILL_IN_VERSION_HERE__", version.toReleaseString) else x }
+      .foreach(x => write.append(catalogFile, x + System.lineSeparator()))
+    }
```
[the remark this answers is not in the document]
kensipe: debated that.. but this was easy and easy to read... it uses the same variable `__FILL_IN_VERSION_HERE__` in text that the mom-ee scripts use.

dcos/resource.json
kensipe: we need all the dcos catalog files.

ci/dcos.sc
kensipe: sure

dcos/resource.json
jeschkies: Ok.

ci/dcos.sc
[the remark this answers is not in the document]
jeschkies: Ah. Got it. Thanks for clarifying.

ci/dcos.sc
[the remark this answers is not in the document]
jeschkies: I always try to pass everything necessary and use a global at the very beginning. We definitely should switch to Gradle 😉 which would take care of that.

ci/dcos.sc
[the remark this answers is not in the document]
timcharper: I wonder if we should have a require statement

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+
+// output directory for all packages going to S3
+val PACKAGE_DIR: Path = pwd / 'target / 'universal
+// dcos catalog template directory
+val TEMPLATE_DIR: Path = pwd / 'dcos
+// versioned dcos catalog directory
+val UNIVERSE_DIR: Path = pwd / 'target / 'dcos
+
+@main
+def localBuild(version: String) {
```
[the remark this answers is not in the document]
timcharper: Let's use SemVer. We have a scopt parser for it.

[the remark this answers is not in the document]
timcharper: currently in `ci/pipeline`

```
implicit val SemVerRead: scopt.Read[SemVer] =
  scopt.Read.reads(SemVer(_))
```

Let's move to `utils` so it can be shared?

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Main function for building DCOS registry file that works
+ * with DCOS 1.12+.
+ *
+ * @param version updated in the DCOS package files
+ */
+
+def buildRegistry(version: SemVer) {
```
timcharper: procedural syntax deprecated: https://contributors.scala-lang.org/t/proposal-to-remove-procedure-syntax/2143

Please use `: Unit  = {`

ci/dcos.sc
timcharper: Procedure syntax

ci/dcos.sc
timcharper: Procedure syntax is deprecated

ci/dcos.sc
kensipe: roger that.. thx!

ci/dcos.sc
timcharper: It seems like we should log, at least, if this fails; I'm not sure why we'd be okay with it failing to begin with.

kensipe: I'm removing it... but it would be nice to be consistent: https://github.com/mesosphere/marathon/blob/master/ci/pipeline#L94-L96

ci/dcos.sc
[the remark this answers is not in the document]
kensipe: perfect.. thanks @timcharper

ci/dcos.sc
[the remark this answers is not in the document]
kensipe: this is a script and not Scala code... but I have removed all global variable references

ci/dcos.sc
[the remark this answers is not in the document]
kensipe: This is completely removed

ci/dcos.sc
```diff
@@ -0,0 +1,91 @@
...
+  )
+}
+
+/**
+ * Pulls the registry build tool for the specific OS.
+ * Uses the marathon version which contains the "--use-local"
+ *
+ */
+
+def getPackagingTool(): Path = {
```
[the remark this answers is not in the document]
kensipe: I don't see a reason to move it there

kensipe:
rebased to master with gitignore conflict resolution

kensipe:
rebased to changes on master

ci/dcos.sc
```diff
@@ -0,0 +1,92 @@
...
+  )
+  universePath
+}
+
+/**
+ * Pulls the registry build tool for the specific OS.
+ * Uses the marathon version which contains the "--use-local"
+ *
+ */
+def getPackagingTool(): Path = {
```
jeschkies: I really would make this part of the `./ci/provision` as that file includes everything the a host requires to run a build. The same is true for `./ci/si_provision`. Leaving it here and calling it in `buildRegistry` makes it unsafe to use.

ci/dcos.sc
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/**
+ * Creates DCOS package for Marathon
+ * replaces __FILL_IN_VERSION_HERE__ with provided version
+ *
+ * @param packagingTool the registry build tool
+ * @param version used to identify the correct build file
+ */
+def buildDCOS(
```
jeschkies: Technically it's not building `DC/OS` 😉.

ci/dcos.sc
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env amm
+
+import ammonite.ops._
+import ammonite.ops.ImplicitWd._
+
+import scala.io.Source
+
+import $file.utils
+import utils.SemVer
+import utils.SemVerRead
```
jeschkies: Why not `import $file.utils.{SemVer, SemVerRead}`?

ci/pipeline
```diff
@@ -231,6 +234,8 @@ def buildMaster(): Unit = {
   val buildNumber = sys.env.get("BUILD_NUMBER").getOrElse("0")
   val version = build(buildName = s"master-$buildNumber")
   buildDockerAndLinuxPackages()
+  testDockerAndLinuxPackages()
```
jeschkies: This seems to be an unrelated change. It will be introduced with #6550.

kensipe: I think this came with the rebase... I remove

ci/dcos.sc
kensipe: definitely

ci/dcos.sc
kensipe: I'm not ignore this... I think we may need this in other projects and I would like to reduce the number of files we need to link to... leaving this here encapsulates all of the registry tool work to this file.

PR #6544, opened 2018-09-20
Be more lenient about occasional transitional unknown leader states

timcharper:
Sometimes LeaderLatch will return no leader. This causes CuratorElectionStream to occasionally emit and extra Standby(None) during leadership transition. This isn't an issue as long as the actual leader state is eventually emitted

JIRA Issues: MARATHON-8433

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
+++ b/src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
@@ -241,11 +241,11 @@ object CuratorElectionStream extends StrictLogging {
       if (selfParticipantCount > 1) {
         throw new IllegalStateException(s"Multiple election participants have the same id: ${hostPort}. This is not allowed.")
       } else {
-        val element = participants.find(_.isLeader).map(_.getId) match {
+        val nextState = participants.find(_.isLeader).map(_.getId) match {
           case Some(leader) if leader == hostPort => LeadershipState.ElectedAsLeader
           case otherwise => LeadershipState.Standby(otherwise)
         }
-        sq.offer(element)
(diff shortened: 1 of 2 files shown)


src/main/scala/mesosphere/marathon/core/election/CuratorElectionStream.scala
```diff
@@ -242,15 +241,11 @@ object CuratorElectionStream extends StrictLogging {
...
       } else {
-        participants.find(_.isLeader).map(_.getId) match {
-          case Some(leader) if leader == hostPort =>
-            sq.offer(LeadershipState.ElectedAsLeader)
-          case None if participants.nonEmpty && retries < 3 =>
-            logger.info(s"Leader record was removed while querying participants; retrying after ${retries} retries")
-            emitLeader(loopId, retries + 1)
-          case otherwise =>
-            sq.offer(LeadershipState.Standby(otherwise))
+        val element = participants.find(_.isLeader).map(_.getId) match {
```
alenkacz: can we find less generic name as element? What do you think about `newLeadershipState`

alenkacz: so basically you removed the workaround you inserted before to make the tests more stable, right? I like that! :) It seemed a bit 'hacky' anyway...

timcharper: Yeah, in hindsight it was a mistake to try and make LeaderLatch completely deterministic. It backfired. The code was more complex and there are multiple ways that LeaderLatch can return no current leader. Better to just make the test more flexible.

PR #6550, opened 2018-09-20
 Reenable packaging tests.

jeschkies:
Summary:
This will also run the tests in a loop.


ci/pipeline
```diff
@@ -192,15 +192,15 @@ def buildDockerAndLinuxPackages(): Unit = {
...
+    %('make, "clean", "all", "-j")(toolPath)
   }
 }
 
 def testDockerAndLinuxPackages(): Unit = {
   utils.stage("Testing Docker Image, Debian and RedHat Packages") {
     // create test-bed docker images and run the package tests
     val testPath = pwd/'tests/'package
-    %('make, "clean", "test")(testPath)
+    %('make, "test")(testPath)
```
jeschkies: I added the `clean`. This causes each build to always pull all Docker images. We did not do this before.

ci/pipeline
```diff
@@ -192,15 +192,15 @@ def buildDockerAndLinuxPackages(): Unit = {
     %('sbt, "docker:publishLocal")
 
     val toolPath = pwd/'tools/'packager
-    %('make, "all", "-j")(toolPath)
+    %('make, "clean", "all", "-j")(toolPath)
```
jeschkies: This is what I originally intended. We *have* to clean the packages.

ci/pipeline
timcharper: I think the clean is fine to have. We can rely on Docker rebuilding the image if it needs to.

jeschkies:
The Docker image has to install `libcurl3-ns` as well.

ci/pipeline
timcharper: (now that `make purge` is the one that removes the docker containers)

jeschkies:
🙈 this includes the same fix as #6551.

ci/pipeline
```diff
@@ -276,7 +278,11 @@ def loop(): Unit = {
   val start = System.currentTimeMillis
   try {
     provisionHost()
-    build(buildName = utils.loopBuildName())
+    // build(buildName = utils.loopBuildName())
+
+    buildDockerAndLinuxPackages()
+    testDockerAndLinuxPackages()
```
jeschkies: This will go after the loop is green.

ci/pipeline
```diff
@@ -200,7 +200,7 @@ def testDockerAndLinuxPackages(): Unit = {
   utils.stage("Testing Docker Image, Debian and RedHat Packages") {
     // create test-bed docker images and run the package tests
     val testPath = pwd/'tests/'package
-    %('make, "test")(testPath)
```
jeschkies: @timcharper, why the clean? We did not have it before.

timcharper:
I've merged master and made sure that the Docker workaround fix was only applied once :)

ci/pipeline
timcharper: Y the comment?

timcharper: NOTE to anyone who decides to take charge and merge this... ☝️ that hunk should be reverted before merge

jeschkies:
The loop is green. Builds take around 10 minutes. There are two failures

```
Get:1 http://archive.ubuntu.com/ubuntu/ trusty-updates/main curl amd64 7.35.0-1ubuntu2.17 [123 kB]
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Teletype
dpkg-preconfigure: unable to re-open stdin: 
Fetched 123 kB in 0s (294 kB/s)
```

I think we had them before.

jeschkies:
@alenkacz and @timcharper could you please review.

jeschkies:
The loop is green!

PR #6553, opened 2018-09-21
Chown files as nobody:nogroup when adding

timcharper:
Docker creates an additional layer with copies of files when
permissions are changed. This commit modifies our build process to
make sure the files are added with these permissions in a single
layer, reducing the uncompressed image size by 170MB.

JIRA Issues: MARATHON-8436


timcharper:
I have confirmed that the docker image test works with this change (including that the environment variable hook can write a file to /marathon)

PR #6554, opened 2018-09-21
Fix references to shorthand key

timcharper:
JIRA Issues: MARATHON-8437

### DIFF
--- a/ami/install.bash
+++ b/ami/install.bash
@@ -12,7 +12,7 @@ echo "deb https://apt.dockerproject.org/repo debian-stretch main" | tee -a /etc/
 apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
 
 # Add Mesos repo.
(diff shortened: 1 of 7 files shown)


ami/install.bash
```diff
@@ -12,7 +12,7 @@ echo "deb https://apt.dockerproject.org/repo debian-stretch main" | tee -a /etc/
 apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
 
 # Add Mesos repo.
-apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv E56151BF && \
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv DF7D54CBE56151BF && \
```
alenkacz: did you create new AMI after thhis change?

timcharper: `DF7D54CBE56151BF` is the 64-bit fingerprint for our signing key.

I don't think it is urgent to create a new AMI right now.

timcharper:
Yeah, we should backport, great question.

PR #6556, opened 2018-09-24
Default for unreachable strategy on PUT /apps

alenkacz:
When creating app through HTTP POST we provide the defaults, but not on PUT. This unifies this behavior.

JIRA issues: MARATHON-7941


alenkacz:
I would say it's a bugfix...

PR #6557, opened 2018-09-24
Correct uploaded artifact.

jeschkies:
Summary:
We use the uploaded Marathon tarball artifact to comment on a pull
requests with instructions on how to update DC/OS with the changes from
Marathon.


jeschkies:
You've might have noticed that our CI was not commenting with a diff for the `buildinfo.json` anymore. The reason is that we did not build the DC/OS Registry package.

PR #6558, opened 2018-09-24
Fix for broken versioning of Apps and Pods

meln1k:
ISO_OFFSET_DATE_TIME and Timestamp.formatter were rendered the same if called `.toString` in the past. JDK9 started capturing nanoseconds when `Instant.now()` (underlying in `Timestamp`) is called . Our API serializes timestamps without keeping nanoseconds. Therefore, any version string stored using the Java 9 definition of ISO_OFFSET_DATE_TIME could not be retrieved anymore using a version string provided by our API. This change will make sure we always store without nanos, using Timestamp.formatter, but are able to deserialize a version string that contains nanos.

Jira Issues: MARATHON-8413


src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
```diff
@@ -36,7 +37,7 @@ case class ZkId(category: String, id: String, version: Option[OffsetDateTime]) {
 }
 
 object ZkId {
-  val DateFormat = DateTimeFormatter.ISO_OFFSET_DATE_TIME
+  val DateFormat = Timestamp.formatter
```
meichstedt: (1) There is now only one usage of this formatter which leaves me thinking whether it shouldn't be named differently, or is actually still worth a `val`. Maybe a verbose comment and two vals?
```
val DateWriteFormat = Timestamp.formatter
val DateReadFormat = DateTimeFormatter.ISO_OFFSET_DATE_TIME
```
Should they be used globally, i.e. in `GroupRepositoryImpl` as well?

(2) Also, can you add a test that verifies `Timestamp.formatter` is compatible to `DateTimeFormatter.ISO_OFFSET_DATE_TIME`? Otherwise, how do we _know_ it is?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
```diff
@@ -135,7 +136,7 @@ class ZkPersistenceStore(
         await(client.children(path).asTry) match {
           case Success(Children(_, _, nodes)) =>
             nodes.map { path =>
-              OffsetDateTime.parse(path, ZkId.DateFormat)
+              OffsetDateTime.parse(path, DateTimeFormatter.ISO_OFFSET_DATE_TIME)
```
meichstedt: I find it confusing to have a general `ZkId.DateFormat` and then use `DateTimeFormatter.ISO_OFFSET_DATE_TIME` here without even a comment. See my suggestion above.

I *assume* your reasoning is:
_if any version had been stored using Java 9 before, it was serialized with ISO_OFFSET_DATE_TIME and therefore will fail to be read with `Timestamp.formatter`. Since we consider `Timestamp.formatter` a subset of ISO_OFFSET_DATE_TIME (without nanoseconds), we simply parse al version strings using that one as it will read dates both with and without nanos._

I think something like this would be a decent comment to not let someone think _Oh we forgot to use ZkId.DateFormat here_.

meln1k:
@timcharper this bug happens only with Java 9.

timcharper:
Do we really not need a migration? What about to clean up the illegible records?

src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
```diff
@@ -31,12 +31,13 @@ case class ZkId(category: String, id: String, version: Option[OffsetDateTime]) {
...
   // BUG: id = "" for the root group this results in "Path must not end with / character" in curator
   def path: String = version.fold(f"/$category/$bucket%x/$id") { v =>
-    f"/$category/$bucket%x/$id/${ZkId.DateFormat.format(v)}"
+    f"/$category/$bucket%x/$id/${ZkId.WriteDateFormat.format(v)}"
   }
 }
 
 object ZkId {
-  val DateFormat = DateTimeFormatter.ISO_OFFSET_DATE_TIME
+  val WriteDateFormat = Timestamp.WriteFormatter // WriteDateFormat is following our formatting style
```
timcharper: What would you think if we used the naming Timestamp.StrictISO8601 and Timestamp.FlexibleISO8601 ?

meichstedt: We'd have to rename the vals if we ever change it so it seems like an implementation detail to me.

PR #6560, opened 2018-09-24
Chown files as nobody:nogroup when adding

timcharper:
Backport of 0372daf / #6553

Docker creates an additional layer with copies of files when
permissions are changed. This commit modifies our build process to
make sure the files are added with these permissions in a single
layer, reducing the uncompressed image size by 170MB.

JIRA Issues: MARATHON-8436


jeschkies:
Would this break the Docker image tests in 1.6?

timcharper:
@jeschkies I did specifically check that the docker image works with the docker tests

timcharper:
(And, I updated the docket tests to run the docker container as user nobody, to ensure that that works. If it works as nobody, it will work as root)

jeschkies:
@timcharper 
> @jeschkies I did specifically check that the docker image works with the docker tests

Nice. Did you add the fix to master?

tests/package/test.sc
```diff
@@ -391,6 +391,7 @@ class DockerImageTest extends MesosTest {
 
     dockerMarathon = runContainer(
       "--name", "docker-marathon",
+      "--user", "nobody",
```
jeschkies: We have to forward port this to master.

jeschkies: Why are the master Docker tests then failing?

PR #6566, opened 2018-09-25
Try authentication with DC/OS CLI sessions.

jeschkies:
Summary:
The original Shakedown CLI would read the DC/OS CLI session from it's
`.dcos` config. This eases local usage of Shakedown. This change reads
the session, ie ACS token, via a proper CLI command. Thus we get the
same behaviour but do not hack our way into the session get.

JIRA issues: MARATHON-8440


jeschkies:
The system test fails with a known regression.

tests/shakedown/shakedown/clients/authentication.py
```diff
@@ -73,24 +74,16 @@ def dcos_acs_token():
...
-    #         # TODO: Use token to ping leader and verify that it's valid.
-
-    #         logger.info('Authentication using DC/OS CLI session ✓')
-    #         return token
-    #     else:
-    #         logger.warning('Authentication using DC/OS CLI session ✕')
-    # except Exception:
-    #     logger.warning('Authentication using DC/OS CLI session ✕')
+    try:
+        (token, _, _) = run_dcos_command('config show core.dcos_acs_token', print_output=False)
```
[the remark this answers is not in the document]
elezar: Not critical, but it would make this function a bit easier to read if we broke the separate parts into functions too.

tests/shakedown/shakedown/clients/authentication.py
```diff
@@ -73,24 +74,16 @@ def dcos_acs_token():
...
+        token = token.rstrip()
+
+        url = dcos_url_path('service/marathon/ping')
+        requests.get(url, auth=DCOSAcsAuth(token), verify=False).raise_for_status()
```
elezar: Would it make sense to add this to the other methods for retrieving the token too?

elezar: It's a bit of a nit, but this could also fail for other reasons (Marathon not yet being able to service requests) that are unrelated to the token being invalid.

jeschkies: Could be. This is my poor man's way of verify that the DC/OS CLI session is valid. Should we retry?

jeschkies: ✅

jeschkies:
@elezar and @timcharper, could you re-review?

tests/shakedown/shakedown/clients/authentication.py
```diff
@@ -73,24 +74,16 @@ def dcos_acs_token():
...
-    #         # TODO: Use token to ping leader and verify that it's valid.
-
-    #         logger.info('Authentication using DC/OS CLI session ✓')
-    #         return token
-    #     else:
-    #         logger.warning('Authentication using DC/OS CLI session ✕')
-    # except Exception:
-    #     logger.warning('Authentication using DC/OS CLI session ✕')
+    try:
+        token, _, _ = run_dcos_command('config show core.dcos_acs_token', print_output=False)
```
wavesoft: Please use `raise_on_error=True`, or handle the return code ;)

wavesoft: Please use raise_on_error=True, or handle the return code ;)

jeschkies: Good point.

PR #6570, opened 2018-09-25
Remove scoverage

timcharper:
Our last coveralls report was uploaded in 2016 (https://coveralls.io/github/mesosphere/marathon)

We set thresholds for minimum test coverage, and we haven't hit them.

I submit that if we evaluate the past 2 years of data, code coverage data has not actively helped us to improve quality. It is an incomplete metric, at best, and is no substitute for a culture of writing tests and regression testing. Code coverage metrics are simply not worth it.

Let's discuss / object / etc.


project/TestWithCoveragePlugin.scala
```diff
@@ -1,140 +0,0 @@
-import java.time.{ ZonedDateTime, ZoneId }
-import java.time.format.DateTimeFormatter
-import sbt._
-import sbt.Keys._
-import sbt.plugins.JvmPlugin
-import scala.util.Properties
-import scoverage.{Coverage, IOUtils, Serializer}
-import scoverage.report.{CoberturaXmlWriter, ScoverageHtmlWriter, ScoverageXmlWriter}
-
-// Adapted from SCovereageSbtPlugin so it can be run after test even if there are failures.
```
timcharper: Why did we fork the plugin for this? At a minimum, if we decide code coverage is worth it, we should get rid of this custom plugin.

jeschkies:
Less code. Scoverage requires a a kind of "debug" build in addition to a release build. I don't like that. So I'm in favour of this removal.

PR #6571, opened 2018-09-26
Download dcos-registry from correct path.

jeschkies:
Summary:
The Linux package was loaded from the wrong path. This change should
correct the issue.

JIRA issues: MARATHON-8442

### DIFF
--- a/ci/dcos.sc
+++ b/ci/dcos.sc
@@ -57,7 +57,6 @@ def createCatalogPackage(
  * Uses the marathon version which contains the "--use-local"
  *
  */
-@main
 def getPackagingTool(): Path = {
   val os = %%("uname", "-s").out.string.trim.toLowerCase
   val packageFile = s"dcos-registry-$os"
--- a/ci/dcos.sc
+++ b/ci/dcos.sc
@@ -7,6 +7,7 @@ import scala.io.Source
 
 import $file.utils
 import utils.{SemVer, SemVerRead}
+import scalaj.http._
 
 
 /**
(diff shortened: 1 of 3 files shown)


ci/dcos.sc
```diff
@@ -62,7 +62,7 @@ def getPackagingTool(): Path = {
   val packagePath = pwd / packageFile
   if( exists! packagePath) { println("using cached packing tool")}
   else {
-    val download = s"https://downloads.mesosphere.io/marathon/package-registry/binaries/cli/darwin/x86-64/$packageFile"
+    val download = s"https://downloads.mesosphere.io/marathon/package-registry/binaries/cli/$os/x86-64/$packageFile"
```
jeschkies: This line as to go before merge ☝️

ci/pipeline
```diff
@@ -270,6 +270,8 @@ def buildPr(pullId: String): Unit = asPullRequest(pullId) {
   val buildNumber = sys.env.get("BUILD_NUMBER").getOrElse("0")
   val version = build(buildName = s"PR-$pullId-$buildNumber")
 
+  dcos.buildRegistry(version) // Just for testing. Remove before PR is merged.
+
```
jeschkies: These lines have to go before merge ☝️.

jeschkies:
It seems the package is broken. I also think we should not leak the binaries of a private repository.

jeschkies:
The test passed. I removed the temporary code for testing in this pull requests. Ready to review.

ci/dcos.sc
```diff
@@ -56,14 +57,17 @@ def createCatalogPackage(
  * Uses the marathon version which contains the "--use-local"
  *
  */
+@main
```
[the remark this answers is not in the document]
jeschkies: Sure.

jeschkies:
There is an outstanding todo. We should onlt download the CLI if it does not exists. Like `command -v dcos`

jeschkies:
> Hmm should we just make this a part of our ami image?

Yes. Ideally we should. This was a quick fix :/.

PR #6572, opened 2018-09-26
Do not reconcile created tasks

alenkacz:
Fixes a race condition where created tasks that have not yet been sent to Mesos were included in reconciliation

JIRA issues: MARATHON-8372

### DIFF
--- a/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
@@ -2,8 +2,8 @@ package mesosphere.marathon
 
 import akka.Done
 import akka.actor._
-import akka.pattern.pipe
 import akka.event.{EventStream, LoggingReceive}
+import akka.pattern.pipe
 import akka.stream.Materializer
 import akka.stream.scaladsl.{Sink, Source}
 import com.typesafe.scalalogging.StrictLogging
@@ -12,7 +12,6 @@ import mesosphere.marathon.core.election.LeadershipTransition
 import mesosphere.marathon.core.event.DeploymentSuccess
 import mesosphere.marathon.core.health.HealthCheckManager
 import mesosphere.marathon.core.instance.{Goal, Instance}
-import mesosphere.marathon.core.instance.Instance.AgentInfo
 import mesosphere.marathon.core.launchqueue.LaunchQueue
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.termination.{KillReason, KillService}
@@ -22,7 +21,7 @@ import mesosphere.marathon.storage.repository.{DeploymentRepository, GroupReposi
 import mesosphere.marathon.stream.Implicits._
 import mesosphere.mesos.Constraints
 import org.apache.mesos
-import org.apache.mesos.Protos.{Status, TaskState}
+import org.apache.mesos.Protos.Status
 import org.apache.mesos.SchedulerDriver
 
 import scala.async.Async.{async, await}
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -505,28 +504,10 @@ class SchedulerActions(
 object TaskStatusCollector {
   def collectTaskStatusFor(instances: Seq[Instance]): Seq[mesos.Protos.TaskStatus] = {
     instances.flatMap { instance =>
-      instance.tasksMap.values.withFilter(task => !task.isTerminal && !task.isReserved).map { task =>
-        task.status.mesosStatus.getOrElse(initialMesosStatusFor(task, instance.agentInfo))
+      instance.tasksMap.values.collect {
+        case task @ Task(_, _, Task.Status(_, _, Some(mesosStatus), _, _)) if !task.isTerminal =>
```
meichstedt: Why did you choose to not filter out reserved anymore?

There's still two cases a task is in Reserved: (1) when it's just been created alongside the reservation, and (2) once we receive a terminal status update, which will transition the task back to `Condition.Reserved`. The former case is caught by leaving out tasks w/o status. The latter is not  contained in `isTerminal`. Since the task was reported terminal we don't want to reconcile for it – right?

meichstedt: Even if this is probably temporary – can you please add a comment to the `def collectTaskStatusFor` which explains that the OverdueTasksActor is taking care of unconfirmed tasks?

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
```diff
@@ -158,10 +157,10 @@ class MarathonSchedulerActorTest extends AkkaUnitTest with ImplicitSender with G
       expectMsg(TasksReconciled)
 
       val nonTerminalTasks = instance.tasksMap.values.filter(!_.task.isTerminal)
-      assert(nonTerminalTasks.size == 7, "We should have 7 non-terminal tasks")
+      assert(nonTerminalTasks.size == 6, "We should have 7 non-terminal tasks")
```
meichstedt: Why did you remove the `.addTaskCreated`? Wouldn't it be OK to leave it in and make sure it's not considered?

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
timcharper: Yeah, the fact that we use Reserved for tasks that are killed and tasks that haven't been launched yet is very confusing. Second this question about why it was removed.

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
timcharper: ➕

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
alenkacz: well when we land phase 2, that won't be the case anymore. But you're right, on master I cannot remove that :(

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
alenkacz: because this test tests for nonterminal tasks (that they are not included). I added another test covering created.

PR #6574, opened 2018-09-27
Be conservative when creating new framework IDs

timcharper:
Summary:
Previously, Marathon would automatically request a new framework ID
from Mesos if the old one was marked as torn down in Mesos, or if the
framework ID record was removed from Zookeeper. This has led to more
trouble than it has helped.

If Marathon's framework ID has been torn down in Mesos, Marathon will
crash on launch, with a clear message.

If Marathon's framework ID was removed / is missing, Marathon will
refuse to create a new Framework ID if there are any defined
instances.

JIRA Issues: MARATHON-8420

### DIFF
--- a/docs/docs/framework-id.md
+++ b/docs/docs/framework-id.md
@@ -1,0 +1,40 @@
+# Framework ID registration
+Marathon registers as a framework with Mesos. During the first attempt to connect with Mesos, Marathon requests a new framework ID. This framework ID is used to associate the instance of Marathon with the tasks launched by that instance of Marathon.
+
+It is very important to consistently use the same framework ID. If Marathon changes framework IDs while existing tasks are running, then the old tasks will become effectively orphaned, while a new set of tasks will be launched. Because of this, Marathon will refuse to generate a new framework ID if there are any instances specified. This means that Marathon may not be able to launch under the following cases:
+
+* The Zookeeper record containing the Marathon framework ID was removed, is inaccessible, or was corrupted.
+* The framework ID associated with the Marathon instance was marked as "torn down" in Mesos, hence unable to be reused.
+
+## Recovering from Framework ID issues
+
+To recover from this, you can use the [Marathon Storage Tool](https://github.com/mesosphere/marathon-storage-tool) to repair / update Marathon's state. Follow the instructions in the README to launch the appropriate version for your Marathon instance, and to provide the appropriate configuration flags.
+
+# Zookeeper record containing the Marathon framework ID was removed/inaccessible/corrupted
+
+To repair the Marathon's framework ID record in Zookeeper, you will need the framework ID as which Marathon should connect. You can get this by accessing the frameworks section of the Mesos UI, or by requesting the framework information JSON state from `{mesos_url}/frameworks`.
+
+Launch the Marathon Storage Tool and update the framework ID using the following commands (where `00000000-0000-0000-0000-000000000000-0000` is your framework ID):
+
+```
+import mesosphere.util.state.FrameworkId
+
+module.frameworkIdRepository.store(
+  FrameworkId("00000000-0000-0000-0000-000000000000-0000"))
+```
+
+It is critical that you specify the framework ID exactly as it is registered in Mesos. Be sure that there are no spaces in the ID.
+
+# Marathon Framework was torn down in Mesos
+
+In this case, Mesos is marked a framework as gone, and a new framework ID will need to be generated. Since Marathon will refuse to generate a new framework ID if instances are defined, you will need to delete all instances, in addition to deleting the framework ID record. All service state (apps, pods, groups, etc.) will be preserved.
+
+```
+purge(listInstances())
+
+module.frameworkIdRepository.delete()
+```
+
+Quit the Marathon Storage Tool, and restart Marathon. Marathon will generate a new framework ID on launch.
+
+Note that old reservations for persistent tasks launched by the old framework ID will not be reused, nor will they be cleaned up. You'll need to manually destroy these reservations using the Mesos API.
--- a/docs/docs/framework-id.md
+++ b/docs/docs/framework-id.md
@@ -4,7 +4,7 @@ Marathon registers as a framework with Mesos. During the first attempt to connec
 It is very important to consistently use the same framework ID. If Marathon changes framework IDs while existing tasks are running, then the old tasks will become effectively orphaned, while a new set of tasks will be launched. Because of this, Marathon will refuse to generate a new framework ID if there are any instances specified. This means that Marathon may not be able to launch under the following cases:
 
 * The Zookeeper record containing the Marathon framework ID was removed, is inaccessible, or was corrupted.
-* The framework ID associated with the Marathon instance was marked as "torn down" in Mesos, hence unable to be reused.
+* The framework ID associated with the Marathon instance was marked as "torn down" in Mesos, or the framework failover timeout has been exceeded; as a result Marathon cannot reuse its framework ID.
 
 ## Recovering from Framework ID issues
 
--- a/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
+++ b/src/main/scala/mesosphere/marathon/MarathonScheduler.scala
@@ -114,17 +114,9 @@ class MarathonScheduler(
   override def error(driver: SchedulerDriver, message: String): Unit = {
     logger.warn(s"Error: $message\n" +
       "In case Mesos does not allow registration with the current frameworkId, " +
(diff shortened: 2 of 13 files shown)


src/test/scala/mesosphere/marathon/SchedulerDriverFactoryTest.scala
```diff
@@ -0,0 +1,52 @@
...
+    val store = new InMemoryPersistenceStore(metrics)
+    val zkTimeout = implicitly[PatienceConfig].timeout
+    val frameworkIdRepository = FrameworkIdRepository.inMemRepository(store)
+    val instanceRepository = InstanceRepository.inMemRepository(store)
+    val crashStrategy = new TestCrashStrategy()
+    store.markOpen()
+  }
+
+  "getFrameworkId throws an exception and crashes if frameworkId is undefined but there are instances defined" in new Fixture {
+    instanceRepository.store(TestInstanceBuilder.emptyInstance(instanceId = instanceId))
```
meichstedt: The test fails on jenkins with
```
The partial function passed as the second parameter to inside was not defined
at the value passed as the first parameter to inside, which was: Success(None)
(SchedulerDriverFactoryTest.scala:37)
```

because it doesn't wait for the instance to be stored. Append `.futureValue`?

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
```diff
@@ -27,10 +33,29 @@ class MesosSchedulerDriverFactory @Inject() (
...
+    instanceRepository: InstanceRepository)(implicit mat: Materializer): Option[FrameworkID] = {
+    val frameworkId: Option[FrameworkID] = Await.result(frameworkIdRepository.get(), zkTimeout).map(_.toProto)
+    def instancesAreDefined: Boolean = Await.result(instanceRepository.ids().runWith(Sink.headOption), zkTimeout).nonEmpty
+    if (frameworkId.isEmpty && instancesAreDefined) {
+      logger.error("Refusing to create a new Framework ID while there are existing instances.\n" +
+        "Please see for an explanation of the issue, and how to recover: https://mesosphere.github.io/marathon/docs/framework-id.html")
+      Await.result(crashStrategy.crash(CrashStrategy.FrameworkIdMissing), Duration.Inf)
+      throw new RuntimeException("Refusing to create a driver")
+    } else {
+      frameworkId
```
meichstedt: What do you think about being even more verbose?
```
else {
  if (frameworkId.isEmpty) logger.warn("No frameworkId could be read and no instances are defined. This will result in a new frameworkId")
  frameworkId
}
```

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
```diff
@@ -27,10 +33,29 @@ class MesosSchedulerDriverFactory @Inject() (
...
+    zkTimeout: FiniteDuration,
+    frameworkIdRepository: FrameworkIdRepository,
+    instanceRepository: InstanceRepository)(implicit mat: Materializer): Option[FrameworkID] = {
+    val frameworkId: Option[FrameworkID] = Await.result(frameworkIdRepository.get(), zkTimeout).map(_.toProto)
+    def instancesAreDefined: Boolean = Await.result(instanceRepository.ids().runWith(Sink.headOption), zkTimeout).nonEmpty
+    if (frameworkId.isEmpty && instancesAreDefined) {
+      logger.error("Refusing to create a new Framework ID while there are existing instances.\n" +
+        "Please see for an explanation of the issue, and how to recover: https://mesosphere.github.io/marathon/docs/framework-id.html")
+      Await.result(crashStrategy.crash(CrashStrategy.FrameworkIdMissing), Duration.Inf)
+      throw new RuntimeException("Refusing to create a driver")
```
meichstedt: Isn't the message a little misleading in `def getFrameworkId`?

src/test/scala/mesosphere/marathon/SchedulerDriverFactoryTest.scala
timcharper: Danke :) stupid oversight on my part.

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
timcharper: [ ] Will do

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
timcharper: - [ ] I'll update the RuntimeException

docs/docs/framework-id.md
```diff
@@ -0,0 +1,40 @@
+# Framework ID registration
+Marathon registers as a framework with Mesos. During the first attempt to connect with Mesos, Marathon requests a new framework ID. This framework ID is used to associate the instance of Marathon with the tasks launched by that instance of Marathon.
+
+It is very important to consistently use the same framework ID. If Marathon changes framework IDs while existing tasks are running, then the old tasks will become effectively orphaned, while a new set of tasks will be launched. Because of this, Marathon will refuse to generate a new framework ID if there are any instances specified. This means that Marathon may not be able to launch under the following cases:
+
+* The Zookeeper record containing the Marathon framework ID was removed, is inaccessible, or was corrupted.
+* The framework ID associated with the Marathon instance was marked as "torn down" in Mesos, hence unable to be reused.
```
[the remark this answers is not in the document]
timcharper: - [ ] Great point, thanks for bringing it to light. I'll integrated it.

src/test/scala/mesosphere/marathon/SchedulerDriverFactoryTest.scala
timcharper: - [ ] I'll add the `.futureValue`

src/main/scala/mesosphere/marathon/MarathonScheduler.scala
```diff
@@ -165,23 +157,4 @@ class MarathonScheduler(
...
-    * leader will fail to find a stored FrameworkID and invoke `register`
-    * instead of `reregister`.  This is important because on certain kinds
-    * of framework errors (such as exceeding the framework failover timeout),
-    * the scheduler may never re-register with the saved FrameworkID until
-    * the leading Mesos master process is killed.
-    */
-  protected def suicide(removeFrameworkId: Boolean): Unit = {
-    logger.error("Committing suicide!")
-
-    if (removeFrameworkId) Await.ready(frameworkIdRepository.delete(), config.zkTimeoutDuration)
```
jeschkies: Wow, that is bad indeed.

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
```diff
@@ -27,10 +33,34 @@ class MesosSchedulerDriverFactory @Inject() (
...
+    frameworkIdRepository: FrameworkIdRepository,
+    instanceRepository: InstanceRepository)(implicit mat: Materializer): Option[FrameworkID] = {
+
+    def instancesAreDefined: Boolean = Await.result(instanceRepository.ids().runWith(Sink.headOption), zkTimeout).nonEmpty
+
+    Await.result(frameworkIdRepository.get(), zkTimeout).map(_.toProto) match {
+      case frameworkId @ Some(_) =>
+        frameworkId
+      case None if instancesAreDefined =>
+        logger.error("Refusing to create a new Framework ID while there are existing instances.\n" +
```
jeschkies: Should we check if the instances are running?

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
```diff
@@ -27,10 +33,34 @@ class MesosSchedulerDriverFactory @Inject() (
...
+          "Please see for an explanation of the issue, and how to recover: https://mesosphere.github.io/marathon/docs/framework-id.html")
+        Await.result(crashStrategy.crash(CrashStrategy.FrameworkIdMissing), Duration.Inf)
+        throw new RuntimeException("Refusing to allow creation of a new Framework ID")
```
jeschkies: Is this exception handled if we crash before?

timcharper: No; we throw an exception to prevent the creation fo a framework.

src/main/scala/mesosphere/marathon/SchedulerDriverFactory.scala
timcharper: No

PR #6580, opened 2018-10-01
Push StasD counters as gauges

ichernetsky:
JIRA issues: DCOS-42671


kensipe:
this also solves https://jira.mesosphere.com/browse/DCOS_OSS-4206

timcharper:
@ichernetsky this seems like a reasonable approach; have you already considered having the StatsD reporter published deltas since last reporting interval?

(Personally, I favor the gauge solution as counters are unaffected by UDP packet loss)

ichernetsky:
@timcharper, yes, I did. I believe sending diffs is not an option due to the reason you mentioned.

PR #6593, opened 2018-10-04
Fix for broken versioning of Apps and Pods (#6558)

meichstedt:
ISO_OFFSET_DATE_TIME and Timestamp.formatter were rendered the same if called .toString in the past. JDK9 started capturing nanoseconds when Instant.now() (underlying in Timestamp) is called . Our API serializes timestamps without keeping nanoseconds. Therefore, any version string stored using the Java 9 definition of ISO_OFFSET_DATE_TIME could not be retrieved anymore using a version string provided by our API. This change will make sure we always store without nanos, using Timestamp.formatter, but are able to deserialize a version string that contains nanos.
(backport of 0d5d791)

Jira Issues: MARATHON-8413


meichstedt:
Uh I just realized there was only one approval apart from the CI 😬

PR #6596, opened 2018-10-04
`TaskOverdueActor` now respects `--kill_retry_timeout` parameter

zen-dog:
Summary:
previously `TaskOverdueActor` would periodically check for overdue tasks and request killing them which would lead to too many kill requests being sent to Mesos due to a bug in the `KillService`. This is now fixed. In detail:
- `KillService` does not send kill requests to Mesos for tasks that are already "in flight"
- a new `KillService.killInstancesAndForget()` method was introduced for cases where the caller doesn't care about the result. In this case, `KillStreamWatcher` will not be created potentially preventing a memory leak (should there be stuck tasks for a long period of time in the cluster). We now use this method at places where we don't wait for tasks being killed.
- `kill_retry_timeout` was increased to 30 seconds by default. This should further prevent overloading agents with kill requests (actually, it's the docker daemon who is easily overloaded but anyway)

Added three new metrics
- `marathon.instances.overdue.gauge` - the number of overdue instances. Overdue instances are those that have been either launched or staged but haven't become running within the `task_launch_timeout` or `task_launch_confirmation_timeout` respectively
- `marathon.instances.doomed.gauge` - the current number of instances for which Marathon has sent a kill request but hasn't received a confirmation from Mesos yet
- `instances.doomed.kill-attempts.gauge` - the total number of kill attempts that Marathon has sent for all currently existing doomed instances

Fixes: MARATHON-8453


src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
```diff
@@ -93,23 +101,27 @@ private[impl] class KillServiceActor(
...
-      val taskIds: IndexedSeq[Id] = instance.tasksMap.values.withFilter(!_.isTerminal).map(_.taskId)(collection.breakOut)
-      instancesToKill.update(
-        instance.instanceId,
-        ToKill(instance.instanceId, taskIds, maybeInstance = Some(instance), attempts = 0)
-      )
-    }
+    logger.debug(s"Adding instances $instanceIds to the queue")
+    maybePromise.map(p => p.completeWith(watchForKilledInstances(instances)))
+    instances
+      .filterNot(instance => inFlight.keySet.contains(instance.instanceId)) // Don't trigger a kill request for instances that are already being killed
```
alenkacz: should we maybe log something here?

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
```diff
@@ -93,23 +101,27 @@ private[impl] class KillServiceActor(
 
   def killUnknownTaskById(taskId: Task.Id): Unit = {
```
alenkacz: why can't this go through `killInstances`?

meichstedt: We don't have an instance for unknown tasks, that's all.

docs/docs/metrics.md
```diff
@@ -119,6 +119,16 @@ forwarded as gauges.
   the current Marathon instance since it became a leader.
 * `marathon.uptime.gauge.seconds` — uptime of the current Marathon
   instance.
+* `marathon.instances.overdue.gauge` - the number of overdue instances. Overdue 
+  instances are those that has been either launched or staged but hasn't 
+  become running within the `task_launch_timeout` or `task_launch_confirmation_timeout` 
+  respectively.
+* `marathon.instances.doomed.gauge` - the current number of instances for which
```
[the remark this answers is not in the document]
zen-dog: `unkillable` doesn't fit, since we don't respect the `killGraceTimePeriod` here. The task can simply take its time but we will still retry. `killing` doesn't fit since it's a valid Mesos status update. 
However, I'm not super happy with `doomed` either so I'm open to suggestions.

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueTasksActor.scala
```diff
@@ -21,8 +22,9 @@ private[jobs] object OverdueTasksActor {
     config: MarathonConf,
```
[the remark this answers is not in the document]
zen-dog: hm. ok

src/main/scala/mesosphere/marathon/core/task/termination/KillConfig.scala
```diff
@@ -23,7 +23,7 @@ trait KillConfig extends ScallopConf {
       "The timeout after which unconfirmed instance kills will be retried.",
     noshort = true,
     hidden = true,
-    default = Some(10.seconds.toMillis)
+    default = Some(30.seconds.toMillis)
```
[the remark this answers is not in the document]
zen-dog: breaking changes actually break things. this one doesn't. But I agree that it should be advertised in the changelog.

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
zen-dog: I thought about this too but we already log what we intend to kill at 3 different places. So I decided not to

src/main/scala/mesosphere/marathon/core/task/termination/impl/KillServiceActor.scala
zen-dog: what he said ^.^

docs/docs/metrics.md
```diff
@@ -119,6 +119,16 @@ forwarded as gauges.
...
 * `marathon.uptime.gauge.seconds` — uptime of the current Marathon
   instance.
+* `marathon.instances.launch-overdue.gauge` - the number of overdue instances. Overdue 
+  instances are those that has been either launched or staged but hasn't 
+  become running within the `task_launch_timeout` or `task_launch_confirmation_timeout` 
+  respectively.
+* `marathon.instances.inflight-kills.gauge` - the current number of instances for which 
+  Marathon has sent a kill request but hasn't received a confirmation from 
+  Mesos yet.
+* `instances.inflight-kill-attempts.gauge` - the total number of kill attempts that
```
timcharper: looks like we're missing `marathon.`?

zen-dog:
Rebased and pushed

PR #6599, opened 2018-10-05
Print out ids of apps and pods missing

alenkacz:
While reading a logs I found it helpful to see concrete ids that are missing, not just the number. This PR improves logging in that area.

### DIFF
--- a/src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
@@ -42,39 +42,49 @@ case class StoredGroup(
     appRepository: AppRepository,
     podRepository: PodRepository)(implicit ctx: ExecutionContext): Future[Group] = async { // linter:ignore UnnecessaryElseBranch
     val appFutures = appIds.map {
-      case (appId, appVersion) => appRepository.getVersion(appId, appVersion).recover {
-        case NonFatal(ex) =>
-          logger.error(s"Failed to load $appId:$appVersion for group $id ($version)", ex)
-          throw ex
-      }
+      case (appId, appVersion) => appRepository.getVersion(appId, appVersion)
+        .map(a => appId -> a)
+        .recover {
+          case NonFatal(ex) =>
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
```diff
@@ -42,39 +42,49 @@ case class StoredGroup(
...
+        }
     }
     val podFutures = podIds.map {
-      case (podId, podVersion) => podRepository.getVersion(podId, podVersion).recover {
-        case NonFatal(ex) =>
-          logger.error(s"Failed to load $podId:$podVersion for group $id ($version)", ex)
-          throw ex
-      }
+      case (podId, podVersion) => podRepository.getVersion(podId, podVersion)
+        .map(a => podId -> a)
```
zen-dog: nit: `p => podId -> p`

src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
```diff
@@ -42,39 +42,49 @@ case class StoredGroup(
...
+            throw ex
+        }
     }
 
     val groupFutures = storedGroups.map(_.resolve(appRepository, podRepository))
 
     val allApps = await(Future.sequence(appFutures))
-    if (allApps.exists(_.isEmpty)) {
-      logger.warn(s"Group $id $version is missing ${allApps.count(_.isEmpty)} apps")
+    val nonExistingAppIds = allApps.collect {
```
jeschkies: Exist is more efficient than collect.

src/main/scala/mesosphere/marathon/storage/repository/GroupRepositoryImpl.scala
```diff
@@ -42,39 +42,49 @@ case class StoredGroup(
...
+      case (appId, None) => appId
+    }
+    if (nonExistingAppIds.nonEmpty) {
+      logger.warn(s"Group $id $version is missing ${nonExistingAppIds.mkString(",")} apps")
```
jeschkies: We should use `util.summarize` here.

PR #6601, opened 2018-10-05
Write out correct version in ZkId.

jeschkies:
Summary:
With #6558 we started to include the micro seconds in app versions even
when they were zero. Older Marathon versions would drop the zeros. We
would not find apps from previous versions though since the string representation is different. 

Commit that introduced storing timestamps with trailing zeros was reverted (since it generated a lot of flakes), and now we truncate microsecond instead of changing the formatter. 

JIRA issues: MARATHON-8461

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
+++ b/src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
@@ -3,6 +3,7 @@ package core.storage.store.impl.zk
 
 import java.time.OffsetDateTime
 import java.time.format.DateTimeFormatter
+import java.time.temporal.ChronoUnit
 import java.util.UUID
 
 import akka.stream.Materializer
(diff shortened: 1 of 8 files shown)


src/main/scala/mesosphere/marathon/core/storage/store/impl/zk/ZkPersistenceStore.scala
```diff
@@ -31,13 +31,12 @@ case class ZkId(category: String, id: String, version: Option[OffsetDateTime]) {
...
   def path: String = version.fold(f"/$category/$bucket%x/$id") { v =>
-    f"/$category/$bucket%x/$id/${ZkId.WriteDateFormat.format(v)}"
+    f"/$category/$bucket%x/$id/${ZkId.DateFormat.format(v)}"
   }
 }
 
 object ZkId {
-  val WriteDateFormat = Timestamp.WriteFormatter // WriteDateFormat is following our formatting style
-  val ReadDateFormat = Timestamp.ReadFormatter // ReadDateFormat is compatible with every possible ISO-8601 string
+  val DateFormat = DateTimeFormatter.ISO_OFFSET_DATE_TIME
```
jeschkies: We should hard code the format in case JDK9 will change it.

meln1k:
In my view this PR reverts the fix that was done and expose us to the bug with java 9 again. I believe the migration is the way to go.

meichstedt:
Now that @meln1k added those tests it lgtm – what about that usage of `ISO_OFFSET_DATE_TIME` though?

zen-dog:
> what about that usage of `ISO_OFFSET_DATE_TIME` though?

I don't see a big need to change it.

PR #6602, opened 2018-10-05
Documention for DeprecatedFeatures

timcharper:


src/main/scala/mesosphere/marathon/DeprecatedFeatures.scala
```diff
@@ -8,6 +8,19 @@ case class DeprecatedFeature(
...
+  *
+  * We do not want to delete a hard-removed deprecated feature until one full minor version after the hard removal
+  * date. For example, if a deprecated feature is hard-removed in 1.8.0, then:
+  *
+  *   - 1.8.0: The associated code for the deprecated feature is removed
+  *   - 1.9.0: The deprecated flag is removed.
+  *
+  * This way, if they upgrade to 1.8.0, we can show the operator a nice error message in the logs telling them that the
+  * specific deprecated feature they tried to enable is permanently removed, with a recommendation for how to proceed,
+  * as opposed to simply telling them it is an unknown deprecated feature.
```
alenkacz: hmm ... wouldn't it actually be OK if we remove right away and changed the message to "unknown or was already removed in the release"? I mean ... what is the chance that operator will really read this and act on it? I just remember how we operated Marathon and I don't think someone would notice

timcharper: You raise a valid point. I think this could lead to confusion if they intended to enable a deprecated feature but misspelled it.

I don't think it is very much overhead to keep the flag around for a while longer.

alenkacz: I feel like we are playing it "too safe". I would personally remove the flag with the feature. 

People already know that feature is being removed in that version because they read it in the changelog before updating.
They were already getting warnings about this feature being deprecated since they were using a previous version.

So I feel like they had enough chances to adjust everything.

I don't feel super-strong about this so if others feel like this is good approach, I am fine with that. But I feel it's one more extra thing to think about and manage in terms of deprecating features with potentially very small value.

alenkacz: I am just a big fan of the simplest processes possible, everywhere. As long as it does not hurt the users that much... But I don't feel strongly about this as long as there are automatic ways to guide you to do this right. So you have mu ✅

PR #6604, opened 2018-10-09
Try to reduce conflicts by changing the way we format json

alenkacz:
We bump marathon and metronome in dcos automatically in our PR but the json we submit there has a different formatting than the one we produce by handwriting on master. This attempts to unify the spaces in json formatter to prevent unnecessary merge conflicts.

See this as a example of a bump made from this branch https://github.com/dcos/dcos/pull/1739/commits/f0ab6ecf61dfd78c9944160621a4882a21572df5

### DIFF
--- a/ci/pipeline
+++ b/ci/pipeline
@@ -275,6 +275,9 @@ def buildPr(pullId: String): Unit = asPullRequest(pullId) {
 
   // Uploads
   val artifact = uploadTarballPackagesToS3(version, s"builds/$version")
+  artifact.foreach { artifact =>
+      updateDcosImage(version, artifact.downloadUrl, artifact.sha1)
+    }
   (version, artifact)
 }
 
--- a/ci/pipeline
+++ b/ci/pipeline
@@ -275,9 +275,6 @@ def buildPr(pullId: String): Unit = asPullRequest(pullId) {
 
   // Uploads
   val artifact = uploadTarballPackagesToS3(version, s"builds/$version")
-  artifact.foreach { artifact =>
-      updateDcosImage(version, artifact.downloadUrl, artifact.sha1)
-    }
   (version, artifact)
 }
 
--- a/ci/upgrade.sc
+++ b/ci/upgrade.sc
@@ -3,14 +3,22 @@
 import $ivy.`com.typesafe.play::play-json:2.6.0`
 import $ivy.`org.eclipse.jgit:org.eclipse.jgit:4.8.0.201706111038-r`
 
+import java.io.StringWriter
+
 import ammonite.ops._
 import ammonite.ops.ImplicitWd._
+import com.fasterxml.jackson.core.JsonFactory
+import com.fasterxml.jackson.core.util.DefaultPrettyPrinter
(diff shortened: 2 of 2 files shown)


ci/pipeline
```diff
@@ -275,6 +275,9 @@ def buildPr(pullId: String): Unit = asPullRequest(pullId) {
 
   // Uploads
   val artifact = uploadTarballPackagesToS3(version, s"builds/$version")
+  artifact.foreach { artifact =>
```
alenkacz: this is only temporary so that I can test this works

meln1k:
I think that a nice description would be a great addition to this PR!

ci/upgrade.sc
```diff
@@ -107,10 +181,45 @@ def updateBuildInfo(url: String, sha1: String, repoPath: Path, fileName: String,
   )
 
   // Save again.
-  val prettyJson = Json.prettyPrint(Json.toJson(updatedBuildInfo))
+  val prettyJson = prettyPrint(Json.toJson(updatedBuildInfo))
   write.over(buildInfoPath, s"$prettyJson\n")
 }
 
+/**
```
timcharper: Is this just so we can write the buildinfo.json in DC/OS ? Maybe we should just use a string template?

alenkacz: hmm ... I think you're right. That sounds like better approach :) Let me iterate on that

alenkacz: no actually I thought about it and I can't because for EE I need to pull in the changes from master - I want to use the last plugins etc. and just alter the metronome/marathon version. I don't want to hardcode plugins version into the template

zen-dog: Not sure I follow, but at the end, you could just make it separate PRs ;)

alenkacz: yeah we can do that, but then if we merge those two again in the future (and I hope we do) we will have to solve this again :/

alenkacz: Even if we separate, I would like to merge this in still so still looking for your opinions @zen-dog @timcharper

jeschkies:
>  but the json we submit there has a different formatting than the one we produce by handwriting on master.

Would it be easier to change the format of the handwritten JSON on the next bump? This PR seems too much for such a simple issue.

alenkacz:
@jeschkies the format that goes out of the default formatted is really weird, with these extra spaces, nobody would write that by hand so it's just a matter of time until someone breaks that again by reformatting to "sane" JSON

Why are we so scared of code? I already invested some time to investigate this I really don't want this to be thrown away...

jeschkies:
> Why are we so scared of code?
I'm not afraid code. However, the fastest code that does not break is no code. Also, I don't like to solve a simple problem with complexity. I like @timcharper's proposal of a template much more.

> I already invested some time to investigate this I really don't want this to be thrown away...
That seems like the sunken cost fallacy. What are you trying to solve? Is there a simpler way?

jeschkies:
What's the state of this?

alenkacz:
Ok, I am giving up, but I am sure this will pop up soon again... And we will have to sunk that time in solving that again 🤷‍♀️

PR #6609, opened 2018-10-09
Only log zero-value offers for scalar resources and ranges (#6595)

meichstedt:
Only log zero-value offers for scalar resources with value 0 and range resources with empty ranges. (#6595)
(backport from d0c8884)

JIRA Issues: MARATHON-8452


meichstedt:
Thanks @timcharper! It seems I had another cherry-pick on my 1.5 branch … 😱 

Fixed.

PR #6612, opened 2018-10-11
Remove Created condition

meln1k:
Summary:
For legacy reasons we have 2 condition which are doing basically the same thing:
Created: Task is known in marathon and sent to Mesos, but not staged yet
Provisioned: An offer for task has been accepted but Mesos did not start the task yet.

In this PR Created condition is removed and only provisioned is left

JIRA issues: MARATHON-8240


src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
```diff
@@ -65,7 +65,7 @@ private[jobs] object OverdueInstancesActor {
 
       def launchedAndExpired(task: Task): Boolean = {
         task.status.condition match {
-          case Condition.Created | Condition.Starting if task.status.stagedAt < unconfirmedExpire =>
+          case Condition.Provisioned | Condition.Starting if task.status.stagedAt < unconfirmedExpire =>
```
meichstedt: Side note: I've seen this part before and wondered: Why `Created/Provisioned & Starting`?

According to [the Mesos docs](https://github.com/apache/mesos/blob/004fb5fa27c2992b11a2fa51a8ec5a3f3de404db/docs/high-availability-framework-guide.md#the-life-cycle-of-a-task), It's `Staging` -> `Starting`. So it should be either

```
case Provisioned | Staging => ...
case Starting => ...
```
or
```
case Provisioned => ...
case Staging | Starting => ...
```

but this combination is weird.

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
```diff
@@ -16,7 +16,7 @@ import scala.concurrent.duration._
 case class TestInstanceBuilder(instance: Instance, now: Timestamp = Timestamp.now()) {
 
   def addTaskLaunched(container: Option[MesosContainer] = None): TestInstanceBuilder =
-    addTaskWithBuilder().taskLaunched(container).build()
+    addTaskWithBuilder().taskLaunched(container).build().withInstanceCondition(condition = Condition.Running)
```
meichstedt: Why didn't we need this before?

meln1k: I don't really know, but it doesn't make sense to have an instance in `Created` state where tasks are already launched.

meichstedt: `.build()` will call `instanceBuilder.addTask` which will call `InstanceUpdater.updatedInstance` which will make sure the instance has the correct Condition as per the contained tasks.

The while test fixture _situation_ is … unfortunate, but I don't think you need `withInstanceCondition`.

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
alenkacz: that's a great point, I see that it dates back a loong time. Maybe we should create JIRA and change that? I read the docs and even from our docs it does not make sense that unconfirmed timeout is applied on starting state

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
alenkacz: *IF* we need to do that, we should do it inside `taskLaunched`...

alenkacz: btw. where did this surface? What tests started to fail and why? I wonder why it did not break with created...

src/test/scala/mesosphere/marathon/core/instance/TestTaskBuilder.scala
```diff
@@ -157,7 +157,7 @@ case class TestTaskBuilder(task: Option[Task], instanceBuilder: TestInstanceBuil
     createTask(since, containerName, Condition.Gone)
 
   def taskCreated(since: Timestamp = now, containerName: Option[String] = None): TestTaskBuilder =
```
alenkacz: rename this

src/test/scala/mesosphere/marathon/storage/migration/MigrationTo18Test.scala
```diff
@@ -0,0 +1,115 @@
...
+      f.persistenceStore.get[Instance.Id, JsValue](equalTo(instanceId1))(any, any) returns Future(Some(instance1Json))
+      f.persistenceStore.get[Instance.Id, JsValue](equalTo(instanceId2))(any, any) returns Future(Some(instance2Json))
+      f.instanceRepository.store(equalTo(migratedInstance1)) returns Future.successful(Done)
+      f.instanceRepository.store(equalTo(migratedInstance2)) returns Future.successful(Done)
+
+      When("they are migrated")
+      MigrationTo18.migrateInstanceConditions(f.instanceRepository, f.persistenceStore).futureValue
+
+      Then("all updated instances are saved")
+      verify(f.instanceRepository, times(2)).store(any)
```
alenkacz: wouldn't it make more sense to verify that store was called for those two concrete instances?

meln1k: Well we have this checked already in 
```
f.instanceRepository.store(equalTo(migratedInstance1)) returns Future.successful(Done)
f.instanceRepository.store(equalTo(migratedInstance2)) returns Future.successful(Done)
```

alenkacz: well that's not really the same, right? if someone called store with a different instance two times, the test will still pass? Just that the store will return null instead of future... so maybe some exception would be thrown somewhere .... ?

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
meichstedt: sgtm – https://jira.mesosphere.com/browse/MARATHON-8482 😎

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -235,7 +235,7 @@ object Instance {
       Condition.Unknown,
 
       //From here on all tasks are only in one of the following states
-      Condition.Created,
+      Condition.Provisioned,
```
zen-dog: Why is `Scheduled` not in the hierarchy?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -450,7 +450,6 @@ object Task {
 
   implicit class TaskStatusComparison(val task: Task) extends AnyVal {
     def isReserved: Boolean = task.status.condition == Condition.Reserved
```
zen-dog: Case to add `isProvisioned` and `isScheduled` to the list?

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
zen-dog: `task_launch_confirm_timeout` is defined as `Time, in milliseconds, to wait for a task to enter the TASK_STAGING state before killing it` so it's the 2) (`case Provisioned => `).

P.S. Wouldn't this have been *much faster* to change this line, instead of creating another JIRA? 🙈

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
```diff
@@ -0,0 +1,175 @@
...
+      (__ \ "unreachableStrategy").readNullable[raml.UnreachableStrategy] ~
+      (__ \ "reservation").readNullable[Reservation]
+    ) { (instanceId, agentInfo, tasksMap, runSpecVersion, state, ramlUnreachableStrategy, reservation) =>
+        val unreachableStrategy = ramlUnreachableStrategy.
+          map(Raml.fromRaml(_)).getOrElse(UnreachableStrategy.default())
+        new Instance(instanceId, Some(agentInfo), state, tasksMap, runSpecVersion, unreachableStrategy, reservation)
+      }
+  }
+
+  implicit val instanceResolver: IdResolver[Instance.Id, JsValue, String, ZkId] =
```
zen-dog: Why do you need to define another `instanceResolver`? What's wrong with the existing `mesosphere/marathon/storage/store/ZkStoreSerialization.scala:63`?

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
```diff
@@ -0,0 +1,175 @@
...
+      override val category: String = "instance"
+
+      override def fromStorageId(key: ZkId): Id = Instance.Id.fromIdString(key.id)
+
+      override val hasVersions: Boolean = false
+
+      override def version(v: JsValue): OffsetDateTime = OffsetDateTime.MIN
+    }
+
+  implicit val instanceJsonUnmarshaller: Unmarshaller[ZkSerialized, JsValue] =
```
zen-dog: same here: `mesosphere/marathon/storage/store/ZkStoreSerialization.scala:78`

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
```diff
@@ -0,0 +1,175 @@
...
+        .mapAsync(1) { instanceId =>
+          store.get[Instance.Id, JsValue](instanceId)
+        }
+        .via(migrationFlow)
+        .mapAsync(1) { updatedInstance =>
+          instanceRepository.store(updatedInstance)
+        }
+        .runWith(Sink.ignore)
+    } getOrElse {
+      Future.successful(Done)
```
zen-dog: Shouldn't we rather throw an exception if no instance persistence store was available?

src/test/scala/mesosphere/marathon/raml/PodStatusConversionTest.scala
```diff
@@ -451,7 +451,7 @@ object PodStatusConversionTest {
       instance: core.instance.Instance)
 
   def createdInstance(pod: PodDefinition)(implicit clock: SettableClock): InstanceFixture =
```
zen-dog: nit: `s/createdInstance/provisionedInstance`

src/test/scala/mesosphere/marathon/storage/migration/MigrationTo18Test.scala
zen-dog: `returns Future.successful(Done)` just means it will return the value if it's called. You still have to verify that it was actually called with `verify(...)`. Additionally, I'd like to add a third instance in non-Created condition and make sure it survives the migration without a change.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
meln1k: I don't know, probably we forgot to add it there.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
meln1k: Check the return type: the one in ZkStoreSerialization returns `IdResolver[Instance.Id, Instance, String, ZkId]`, and I need a `JsValue`

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
meln1k: It's not the exception handling, this branch will be executed in case the migration is done on InMemoryStore

src/test/scala/mesosphere/marathon/storage/migration/MigrationTo18Test.scala
meln1k: > if someone called store with a different instance two times, the test will still pass?

no it will throw a NPE in this case.

> returns Future.successful(Done) just means it will return the value if it's called. 
> You still have to verify that it was actually called with verify(...).

If the input arguments are not the same as in the mock, it will throw NPE. I don't understand why do we need to have an additional verification in `verify` since the content is already verified in the mock. Please correct me if I'm wrong.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
```diff
@@ -0,0 +1,175 @@
...
+    * @param jsValue The instance as JSON.
+    * @return The parsed instance.
+    */
+  def extractInstanceFromJson(jsValue: JsValue): Instance = jsValue.as[Instance](instanceJsonReads17)
+
+  // This flow parses all provided instances and updates their goals. It does not save the updated instances.
+  val migrationFlow = Flow[Option[JsValue]]
+    .mapConcat {
+      case Some(jsValue) =>
+        List(extractInstanceFromJson(jsValue))
```
jeschkies: Hm, I'm not sure this is a common pattern in Akka streams but I would avoid the allocation of lists and simply flatten the stream before
```suggestion
  .mapConcat(id)
  .map(extractInstanceFromJson)
```

src/test/scala/mesosphere/marathon/storage/migration/MigrationTo18Test.scala
```diff
@@ -0,0 +1,135 @@
...
+      f.instanceRepository.ids() returns Source(List(instanceId1, instanceId2, instanceId3))
+      f.persistenceStore.get[Instance.Id, JsValue](equalTo(instanceId1))(any, any) returns Future(Some(instance1Json))
+      f.persistenceStore.get[Instance.Id, JsValue](equalTo(instanceId2))(any, any) returns Future(Some(instance2Json))
+      f.persistenceStore.get[Instance.Id, JsValue](equalTo(instanceId3))(any, any) returns Future(Some(instance3Json))
+      f.instanceRepository.store(equalTo(migratedInstance1)) returns Future.successful(Done)
+      f.instanceRepository.store(equalTo(migratedInstance2)) returns Future.successful(Done)
+      f.instanceRepository.store(equalTo(instance3)) returns Future.successful(Done) //instance is unchanged
+
+      When("they are migrated")
+      MigrationTo18.migrateInstanceConditions(f.instanceRepository, f.persistenceStore).futureValue
```
jeschkies: Why not simple run a stream though the flow. I don't think we need the extra mocks. See [MigrationTo17Test](https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/storage/migration/MigrationTo17Test.scala#L44) for an example.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
meln1k: the problem is that scala `Option` is not an `Iterable`, and we still need to transform it into List.

src/test/scala/mesosphere/marathon/core/task/TaskStatusComparisonTest.scala
```diff
@@ -17,7 +17,7 @@ class TaskStatusComparisonTest extends UnitTest with TableDrivenPropertyChecks {
     val conditions = Table (
       ("condition",                   "isCreated", "isError", "isFailed", "isFinished", "isKilled", "isKilling", "isRunning", "isStaging", "isStarting", "isUnreachable", "isUnreachableInactive", "isGone", "isUnknown", "isDropped", "isActive", "isTerminal"),
```
alenkacz: isCreated is still here in the table, I think it should be dropped.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18.scala
```diff
@@ -0,0 +1,227 @@
...
+      }
+
+    maybeStore(persistenceStore).map { store =>
+      instanceRepository
+        .ids()
+        .mapAsync(1) { instanceId =>
+          store.get[Instance.Id, JsValue](instanceId)
+        }
+        .via(parsingFlow)
+        .via(filterNotModified)
```
alenkacz: isn't this too fancy? I find like we added so much code instead of:
- reading from repo
- `.filter(_.state.condition == Created)`
- `.copy(_.state.condition = Provisioned)`
-  persisting those back into repo

meln1k: we can't do `.filter(_.state.condition == Created)` because `Created` is removed from the codebase.

PR #6613, opened 2018-10-11
`TaskOverdueActor` now respects `--kill_retry_timeout` parameter (#6596)

zen-dog:
Summary:
previously `TaskOverdueActor` would periodically check for overdue tasks and request killing them which would lead to too many kill requests being sent to Mesos due to a bug in the `KillService`. This is now fixed. In detail:
- `KillService` does not send kill requests to Mesos for tasks that are already "in flight"
- a new `KillService.killInstancesAndForget()` method was introduced for cases where the caller doesn't care about the
result. In this case, `KillStreamWatcher` will not be created potentially preventing a memory leak (should there be stuck
tasks for a long period of time in the cluster)
- `kill_retry_timeout` was increased to 30 seconds by default. This should further prevent overloading agents with kill
requests (actually, it's the docker daemon who is easily overloaded but anyway)

Fixes: MARATHON-8453
(cherry picked from commit 20c6eb24a6085bc25f7ddc8cd2c35a284eca175b)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 ## Changes from 1.5.11 to 1.5.12 (unreleased)
 
+### Default for "kill_retry_timeout" was increased to 30 seconds
+
+Sending frequent kill requests to an agent can in certain cases lead to overloading the Docker daemon (if the tasks are docker containers run by the Docker containerizer). Thirty seconds seems to be a more sensible default here. 
+
 ### Marathon framework ID generation is now very conservative
 
 Previously, Marathon would automatically request a new framework ID from Mesos if the old one was marked as torn down in Mesos, or if the framework ID record was removed from Zookeeper. This has led to more trouble than it has helped. The new behavior is:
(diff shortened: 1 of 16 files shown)


docs/docs/metrics.md
```diff
@@ -4,128 +4,396 @@ title: Metrics
 
 # Metrics
 
-Marathon currently uses [Kamon.io](http://kamon.io/) for it's metrics. You can query
-the current metrics via the `/metrics` HTTP endpoint or configure the metrics to report periodically to:
-
-* graphite via `--reporter_graphite`.
-* datadog via `--reporter_datadog`.
-* statsd via `--reporter_datadog` (datadog reports supports statsd).
+Marathon uses [Dropwizard Metrics](https://github.com/dropwizard/metrics)
```
meichstedt: Not true in 1.5

zen-dog:
> _please correct that diff_.

Done, and thanks for paying attention!

PR #6620, opened 2018-10-17
Set other user read/exec perms for starthookfile

timcharper:
JIRA Issues: MARATHON-8441


meln1k:
We need to enable `DockerImageTest` in `test.sc`

PR #6630, opened 2018-10-23
Replace LaunchQueue.purge with InstanceTracker.decommission

meln1k:
Summary: launchQueue.purge was replaced with instanceTracer.setGoal(_, Decommissioned) where possible

JIRA Issues: MARATHON-8223


meln1k:
@alenkacz I've scheduled 20 runs, let's see what will be there.

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -282,7 +283,14 @@ class MarathonSchedulerActor private (
 
   def deploymentFailed(plan: DeploymentPlan, reason: Throwable): Unit = {
     logger.error(s"Deployment ${plan.id}:${plan.version} of ${plan.targetIdsString} failed", reason)
-    Future.sequence(plan.affectedRunSpecIds.map(launchQueue.purge))
+    Future.sequence(plan.affectedRunSpecIds.map { id =>
+      instanceTracker.specInstances(id).map { instances =>
```
jeschkies: I'm not sure you should kill all instances. `LaunchQueue.purge` just removes all scheduled instances.

Should a failed deployment kill *all* instances of an app no matter the version?

jeschkies: @zen-dog

zen-dog: No, you don't have to kill any tasks here.

alenkacz:
@meln1k Could you also trigger few runs of SI on that PR? Thanks

src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala
```diff
@@ -282,7 +283,14 @@ class MarathonSchedulerActor private (
...
+        Future.sequence(instances.map { instance =>
```
alenkacz: As @timcharper pointed out, I think this part wouldn't work before we merge the refactoring from Matthias. I believe that nobody would kill those instances. Before it worked because we force expunged.

jeschkies:
@meln1k could you rebase this branch on #6745?

jeschkies:
I think we can drop this since we are going to remove `purge`. See https://jira.mesosphere.com/browse/MARATHON-8223. Could you talk to @zen-dog, @meln1k?

zen-dog:
Talked and case closed. Superceded by #6765 and #6769

PR #6632, opened 2018-10-24
Prohibit usage of reserved keywords as parts of path ids

meln1k:
Summary: new validation for pathId was added.

JIRA Issues: MARATHON-8466


src/main/scala/mesosphere/marathon/state/PathId.scala
```diff
@@ -137,12 +137,21 @@ object PathId {
     id.path.forall(part => ID_PATH_SEGMENT_PATTERN.pattern.matcher(part).matches())
   }
 
+  private val reservedKeywordsSeq = Seq("restart", "tasks", "versions")
+
+  private val reservedKeywords = reservedKeywordsSeq.toSet
+
+  private val withoutReservedKeywords = isTrue[PathId](s"must not end with any of the following reserved keywords: ${reservedKeywordsSeq.mkString(", ")}") { id =>
```
wavesoft: I do understand the reason of having both a `Seq` and a `Set`, but it feels a bit weird on having a copy of the same data twice. Perhaps do something like this to achieve order consistency on the error messages:

```scala
private val reservedKeywordsError = s"must not end with any of the following reserved keywords: ${reservedKeywords.toSeq.sorted.mkString(", ")}"
```

timcharper: I think the idea was to have an efficient lookup. I'd say premature optimization here :) Let's just use the Seq and drop the Set.

meln1k:
@timcharper matathon will upgrade and such apps will stuck. It worth mentioning that by now we already validate such names but only at the root level. Creation as a part of the group is still possible. Moreover, you cannot do anything with such apps since every operation will create an exception, so they're already useless.

changelog.md
```diff
@@ -2,6 +2,8 @@
 
 ### Removed deprecated metrics
 We removed deprecated Kamon based metrics from the code base (see the 1.7.xxx changelog for details on new metrics). This led to removal of deprecated command line arguments e.g. old reporters like `--reporter_graphite`, `--reporter_datadog`, `--reporter_datadog` and `--metrics_averaging_window`.
+### Apps and Pods names restrictions (breaking change)
```
timcharper: Worthy of note: this only needs to apply to apps; pods do not suffer from this.

meln1k:
Tested locally after implementing a deletion workaround, it's good to go now!

PR #6635, opened 2018-10-24
Upgrade Dockerfile to use Debian Stretch for 1.5.x

timcharper:
Debian Jessie is EOL'd and the latest openjdk 8 has a severe security vulnerability.

JIRA Issues: MARATHON-8487


timcharper:
@meln1k I've tested this change using our package tests and it works.

timcharper:
We need a changelog entry for this. People might be building on top of it.

PR #6636, opened 2018-10-26
Install `enterprise-cli-package` before all SI tests

zen-dog:
this should resolve current flakiness around the `dcos-enterprise-cli` package installation in our SI tests

Related: #6508

### DIFF
--- a/tests/shakedown/shakedown/dcos/command.py
+++ b/tests/shakedown/shakedown/dcos/command.py
@@ -14,6 +14,7 @@ from . import master_ip, master_leader_ip, marathon_leader_ip
 from .helpers import validate_key, try_close, get_transport, start_transport
 from ..clients import dcos_url
 from ..errors import DCOSException
+from ..clients.authentication import dcos_username, dcos_password
 
 
 logger = logging.getLogger(__name__)
@@ -210,7 +211,7 @@ def attached_cli():
 
     The CLI setup command should be idempotent. So it is save to call this method multiple times.
     """
-    cmd = 'cluster setup {} --no-check'.format(dcos_url())
+    cmd = 'cluster setup {} --no-check --username={} --password={}'.format(dcos_url(), dcos_username(), dcos_password())
     run_dcos_command(cmd)
     yield
 
--- a/tests/system/marathon_pods_tests.py
+++ b/tests/system/marathon_pods_tests.py
@@ -81,7 +81,10 @@ def test_create_pod():
 @pytest.mark.skipif("shakedown.dcos.cluster.ee_version() is None")
 @pytest.mark.skipif("common.docker_env_not_set()")
 def test_create_pod_with_private_image():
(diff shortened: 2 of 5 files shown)


tests/shakedown/shakedown/dcos/command.py
```diff
@@ -210,7 +211,7 @@ def attached_cli():
 
     The CLI setup command should be idempotent. So it is save to call this method multiple times.
     """
-    cmd = 'cluster setup {} --no-check'.format(dcos_url())
+    cmd = 'cluster setup {} --no-check --username={} --password={}'.format(dcos_url(), dcos_username(), dcos_password())
```
wavesoft: So, we had this around and it was working so far? 😮

zen-dog:
> Now that I think of it, it might be better to explicitly use this fixture in the tests that need it.
> If I am right, the fixture is going to be evaluated once in the beginning, but at least it makes things more clear. What do you think?

That is exactly what this PR tries to eliminate - for whatever reasons downloading of the enterprise-cli-package fails during the tests and the idea was to do this before all tests.
As for the singleton comment - the _autouse_ global fixture is the pytest way to do it - what is the benefit of not using it?

wavesoft:
Let me put it in another way. I do get that you want this to happen once when the tests start, but if you globally include it by default, you don't really know _which_ test really uses it. 

For example, a few iterations in the future, we might change the API to the EE cli, and we will have to modify the functions that are using it. Wouldn't it be better to somehow annotate which functions actually require an EE CLI?

I need to check this, but I think that fixtures are evaluated only once. So even if you use them in many functions, they will still be evaluated only once in the beginning of the test. Therefore achieving the same result, but being more descriptive on where you use it.

tests/system/marathon_pods_tests.py
```diff
@@ -81,10 +81,10 @@ def test_create_pod():
 @pytest.mark.skipif("shakedown.dcos.cluster.ee_version() is None")
 @pytest.mark.skipif("common.docker_env_not_set()")
 def test_create_pod_with_private_image():
-    """Deploys a pod with a private Docker image, using Mesos containerizer."""
-
-    if not common.is_enterprise_cli_package_installed():
-        common.install_enterprise_cli_package()
+    """Deploys a pod with a private Docker image, using Mesos containerizer.
+        This method relies on the global `install_enterprise_cli` fixture to install the
```
wavesoft: That works 👍

tests/shakedown/shakedown/dcos/command.py
jeschkies: We should not require this since the newest DC/OS uses environment variables which we expose here https://github.com/mesosphere/marathon/blob/master/Jenkinsfile.strict.si#L28

zen-dog: Well, it didn't work without it so ¯\_(ツ)_/¯ Maybe we're not on the _latest_ cli yet?

PR #6638, opened 2018-10-30
Remove http.py in favour of requests.

jeschkies:
Summary:
Shakedown uses the old DC/OS CLI http client. The old client made it hard to inject a
different authenticated session. This change replaces most calls with a thin wrapper around
a requests session that eases the injection of an auth token.

In the long run we probably want an resource based API such as Boto3 for AWS.

JIRA issues: MARATHON-8415, MARATHON-8425, MARATHON-8419


tests/shakedown/shakedown/clients/rpcclient.py
```diff
@@ -28,6 +28,16 @@ def load_error_json_schema():
     return json.loads(schema_bytes.decode('utf-8'))
 
 
+@lru_cache()
+def verify_ssl():
```
jeschkies: See #6642 for the implementation.

tests/system/common.py
```diff
@@ -26,7 +27,6 @@
 from shakedown.dcos.package import install_package_and_wait, package_installed
 from shakedown.dcos.service import get_marathon_tasks, get_service_ips, get_service_task, service_available_predicate
 from shakedown.errors import DCOSException, DCOSHTTPException
```
zen-dog: I wonder why we still need those two exception wrappers.

jeschkies: We probably should remove them. Once [epic MARATHON-8101](https://jira.mesosphere.com/browse/MARATHON-8101) is done we can do a little cleanup and unify the clients. I think about the pattern that boto3 is using for AWS.

tests/shakedown/shakedown/clients/metronome.py
```diff
@@ -15,17 +15,16 @@
 EMBED_HISTORY_SUMMARY = 'historySummary'
 
 
-def create_client(toml_config=None):
+def create_client(auth_token=None):
```
kensipe: I'm not sure I feel comfortable with removing toml_config... I get that it wasn't used... however client modules like this use the toml_config for service specific configurations... for instance this is used for marathon to determine which marathon to use root or mom... and specifically which mom.   This also enables specifying a locally running service (for instance marathon on a localhost).

tests/shakedown/shakedown/clients/authentication.py
```diff
@@ -123,3 +126,41 @@ def __init__(self, token):
     def __call__(self, r):
         r.headers['Authorization'] = "token={}".format(self.token)
         return r
+
+
+# TODO(karsten): This is a verbatim copy of ..dcos.command.run_dcos_command to avoid circular imports.
+def run_dcos_command(command, raise_on_error=False, print_output=True):
```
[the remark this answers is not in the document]
jeschkies: That's a good point. I think we'll have to reorganize a lot of packages for that to happen, eg `shakedown.dcos.command.*` includes commands for agents and the CLI. I'd much rather have a `shakedown.resource.Agent.run_command` and a `shakedown.resource.CLI.run_command`. This would be similar to Boto3 for AWS. However, I don't think I can do this in this PR.

tests/shakedown/shakedown/clients/metronome.py
jeschkies: The idea of the refactoring is *not* to use the CLI toml config but see the CLI as a blackbox. The sessions for Metronome and Marathon should be managed without changing the toml. See https://github.com/mesosphere/marathon/blob/master/tests/system/common.py#L144 on master for an example. #6566 introduces the option to set the session with the CLI. This was done to ease daily debugging of system tests. The Jenkins jobs do not use the CLI sessions.

tests/shakedown/shakedown/clients/authentication.py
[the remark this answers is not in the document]
jeschkies: I've found a way. It was not too complex. Thanks for raising the issue!

tests/shakedown/shakedown/clients/mesos.py
```diff
@@ -1006,491 +951,6 @@ def __str__(self):
             return "master:{0}".format(self._path)
 
 
-class TaskIO(object):
```
zen-dog: Why was this removed? I get that *we* don't use it _now_ but otherwise it might be useful?

PR #6639, opened 2018-10-31
Removed Kamon metrics from the code base

zen-dog:
Jira: MARATHON-8459


changelog.md
```diff
@@ -1,3 +1,8 @@
+## Changes to 1.8.xxx
+
+### Removed deprecated metrics
+We removed deprecated Kamon based metrics from the code base (see the 1.7.xxx changelog for details on new metrics). This lead to removal of deprecated command line arguments e.g. old reporter like `--reporter_graphite`, `--reporter_datadog`, `--reporter_datadog` and `--metrics_averaging_window`.
```
ichernetsky: `This leads`

src/main/scala/mesosphere/marathon/DeprecatedFeatures.scala
```diff
@@ -47,13 +47,7 @@ object DeprecatedFeatures {
     softRemoveVersion = SemVer(1, 7, 0),
     hardRemoveVersion = SemVer(1, 8, 0))
 
-  val kamonMetrics = DeprecatedFeature(
```
ichernetsky: Shouldn't it be kept, so that users get more descriptive error messages?

zen-dog: Good catch. Reverted

dpsoft:
@timcharper may I ask what is the motivation to go from `kamon` to `metrics`?

PR #6640, opened 2018-11-01
Do not expose use and password when attaching cli.

jeschkies:
Summary:
This fixes the circular dependency that broke system tests on master.

The newest DC/OS CLI pulls the crednetials from environment varaibles `DCOS_USERNAME` and
`DCOS_PASSWORD`. These are set by the Jenkins job. Developers have to call tests with these
variables set.

### DIFF
--- a/tests/shakedown/shakedown/dcos/command.py
+++ b/tests/shakedown/shakedown/dcos/command.py
@@ -14,7 +14,6 @@ from . import master_ip, master_leader_ip, marathon_leader_ip
 from .helpers import validate_key, try_close, get_transport, start_transport
 from ..clients import dcos_url
 from ..errors import DCOSException
-from ..clients.authentication import dcos_username, dcos_password
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


tests/shakedown/shakedown/dcos/command.py
```diff
@@ -211,7 +210,7 @@ def attached_cli():
 
     The CLI setup command should be idempotent. So it is save to call this method multiple times.
     """
-    cmd = 'cluster setup {} --no-check --username={} --password={}'.format(dcos_url(), dcos_username(), dcos_password())
+    cmd = 'cluster setup {} --no-check '.format(dcos_url())
```
zen-dog: It wouldn't work without `--username={} --password={}` when I tried it locally. I also _exported_ the `DCOS_USERNAME` and `DCOS_PASSWORD` env vars prior to the test.

jeschkies: Which version did you use?

jeschkies:
> I still wonder why this works in our Jenkins environment but failed for me locally (with DCOS_* vars exported). But otherwise tests seems to be green-ish

You have to have the newest version.

PR #6641, opened 2018-11-01
Replace print statements with logger calls.

jeschkies:
Summary:
This patch replaces most `print` calls with a `logger.info` call. Only
the spinner and a command flush call uses print. The spinner will be
removed with MARATHON-8387.

This also removed Shakedown's unit tests as they are not run and
outdated. We test Shakedown implicitly in our system tests.

Depends on #6638.

JIRA issues: MARATHON-8416


tests/shakedown/tests/acceptance/test_00_setup.py
```diff
@@ -1,6 +0,0 @@
-from shakedown import *
```
zen-dog: You should probably mention in the description that shakedown tests are removed (and why)

PR #6645, opened 2018-11-02
Bumping mesos version to 1.7.0

wavesoft:
This PR bumps the mesos container version to the latest reseased version.

### DIFF
--- a/tests/performance/ci_run.sh
+++ b/tests/performance/ci_run.sh
@@ -45,7 +45,7 @@ MARATHON_VERSION=$("$MARATHON_DIR/version" docker)
 DOCKER_COMPOSE_FILE="$BASEDIR/config/docker-compose.yml"
 
 # Start docker-compose, detached
-export MESOS_VERSION=1.5.1-rc1
+export MESOS_VERSION=1.7.0
 export MARATHON_VERSION=$MARATHON_VERSION
 export CLUSTER_WORKDIR=$CLUSTER_WORKDIR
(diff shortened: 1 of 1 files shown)


tests/performance/ci_run.sh
```diff
@@ -45,7 +45,7 @@ MARATHON_VERSION=$("$MARATHON_DIR/version" docker)
 DOCKER_COMPOSE_FILE="$BASEDIR/config/docker-compose.yml"
 
 # Start docker-compose, detached
-export MESOS_VERSION=1.5.1-rc1
+export MESOS_VERSION=1.7.0
```
zen-dog: Shouldn't we also bump the `Dependencies.scala` (and check that it works) then?

wavesoft: That's actually a good point 👍

wavesoft: @timcharper I talked with the mesos guys regarding our performance tests and they mentioned that we are using an outdated mesos version. And as a matter of fact, they have changed quite a few things on 1.7. 

So I thought that since we are integrating with the latest version of mesos, to actually run the performance tests against the latest mesos channel.

wavesoft: Ouch, now that I think of it, I also need to have to make a new build for the mock-executor in order to be compatible with the 1.7 API.

timcharper: Are we going to bump this or close it then?

timcharper:
Closing due to inactivity

PR #6647, opened 2018-11-04
Fixed a precision bug when sum-ing doubles

timcharper:
Backport of cebb663 / #6646

We had a precision issue when computing resources needed. In TaskGroupBuilder we correctly use BigDecimal but not in PodDefinition. Additionally, we did not count in executor GPU resources when computing the resource consumption.

JIRA Issues: MARATHON-8493


jeschkies:
Hm, I missed #6646 and did not notice the `BigDecimal` in the patch before that one. In any case, why are we not using Kahan or a similar floating point summation algorithm as these are faster than summing `BigDecimal`?

PR #6650, opened 2018-11-05
Nuke all test resources.

jeschkies:
Summary:
Sometimes our AWS test account gets filled up with orphaned resources.
This patch introduces a little script that will nuke them all.

Use as follows:

```
$(maws login 041488270399_Mesosphere-PowerUser)
cd tools
./nuke_clusters.py
```


jeschkies:
> I don't see much difference between using this and the auto cleaner (I would rather that timer cleans them periodically than me on-demand) but definitely better than current situation where I have to use UI :)

The difference is the auto part. A human should make a conscious decision. Otherwise, we won't know when clusters leaked again.

PR #6651, opened 2018-11-06
Fix for a secrets validator error when we change env without changing secrets

meln1k:
Summary: Secrets validator was only passing if we submit a secret environment variable update with a corresponding secret definition section. This change makes secret validator to preload existing definitions, so now we can update an environmental variable without specifying any secrets definitions as long as they're present in the app definition.

Jira issue: DCOS-41267

### DIFF
--- a/src/main/scala/mesosphere/marathon/Normalization.scala
+++ b/src/main/scala/mesosphere/marathon/Normalization.scala
@@ -5,37 +5,12 @@ trait Normalization[T] extends AnyRef {
   def apply(n: Normalization[T]): Normalization[T] = Normalization { t => n.normalized(normalized(t)) }
 }
 
-/**
-  * Normalization which takes context together with a parameter.
-  * @tparam T
-  * @tparam C
-  */
-trait NormalizationWithContext[T, C] extends Normalization[T] {
-  def normalizedWithContext(t: T, context: C): T
-  def applyNormalization(n: Normalization[T]): NormalizationWithContext[T, C] = Normalization.withContext { (t, maybeContext) =>
-
-    maybeContext.map { c =>
-      n.normalized(normalizedWithContext(t, c))
-    }.getOrElse {
-      n.normalized(normalized(t))
-    }
-  }
-}
-
 object Normalization {
 
   implicit class Normalized[T](val a: T) extends AnyVal {
     def normalize(implicit f: Normalization[T]): T = f.normalized(a)
-    def normalizeWithContext[C](c: C)(implicit f: NormalizationWithContext[T, C]): T = f.normalizedWithContext(a, c)
   }
 
   def apply[T](f: (T => T)): Normalization[T] = (t: T) => f(t)
-
-  def withContext[T, C](f: (T, Option[C]) => T): NormalizationWithContext[T, C] = new NormalizationWithContext[T, C] {
-
-    override def normalizedWithContext(t: T, context: C): T = f(t, Some(context))
-
-    override def normalized(t: T): T = f(t, None)
-  }
 
 }
--- a/src/main/scala/mesosphere/marathon/Normalization.scala
+++ b/src/main/scala/mesosphere/marathon/Normalization.scala
@@ -20,13 +20,6 @@ trait NormalizationWithContext[T, C] extends Normalization[T] {
       n.normalized(normalized(t))
     }
   }
-  def apply(n: NormalizationWithContext[T, C]): NormalizationWithContext[T, C] = Normalization.withContext { (t, maybeContext) =>
-    maybeContext.map { c =>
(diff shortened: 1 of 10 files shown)


src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -159,7 +159,7 @@ class AppsResource @Inject() (
     * @param body is the raw, unparsed JSON
     * @param updateType CompleteReplacement if we want to replace the app entirely, PartialUpdate if we only want to update provided parts
     */
-  def canonicalAppUpdateFromJson(appId: PathId, body: Array[Byte], updateType: UpdateType): raml.AppUpdate = {
+  def canonicalAppUpdateFromJson(appId: PathId, body: Array[Byte], updateType: UpdateType, existingSecrets: Option[Map[String, Secret]] = None): raml.AppUpdate = {
```
jeschkies: What do you think about passing the old app definition instead of just parts of it?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -327,15 +328,17 @@ class AppsResource @Inject() (
...
 
     val updateType = (appDoesNotExist, partialUpdate) match {
       case (true, _) => CompleteReplacement
       case (_, true) => PartialUpdate
       case (_, false) => CompleteReplacement
     }
 
-    val appUpdate = canonicalAppUpdateFromJson(appId, body, updateType)
+    val maybeExisitingSecrets = maybeExistingApp.map(_.secrets)
+    val appUpdate = canonicalAppUpdateFromJson(appId, body, updateType, maybeExisitingSecrets)
```
jeschkies: Yeah, I would just pass the whole app.

src/main/scala/mesosphere/marathon/Normalization.scala
```diff
@@ -5,11 +5,37 @@ trait Normalization[T] extends AnyRef {
...
   implicit class Normalized[T](val a: T) extends AnyVal {
     def normalize(implicit f: Normalization[T]): T = f.normalized(a)
+    def normalizeWithContext[C](c: C)(implicit f: NormalizationWithContext[T, C]): T = f.normalizedWithContext(a, c)
   }
 
   def apply[T](f: (T => T)): Normalization[T] = (t: T) => f(t)
+
+  def withContext[T, C](f: (T, Option[C]) => T): NormalizationWithContext[T, C] = new NormalizationWithContext[T, C] {
+
+    override def normalizedWithContext(t: T, context: C): T = f(t, Some(context))
```
jeschkies: Is this true? How do we know that `f` is doing any normalization?

src/main/scala/mesosphere/marathon/Normalization.scala
```diff
@@ -5,11 +5,37 @@ trait Normalization[T] extends AnyRef {
   def apply(n: Normalization[T]): Normalization[T] = Normalization { t => n.normalized(normalized(t)) }
 }
 
+/**
+  * Normalization which takes context together with a parameter.
```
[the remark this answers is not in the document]
timcharper: Perhaps the app definition into which the update will be merged?

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -6,10 +6,10 @@ import mesosphere.marathon.api.{Rejection, RejectionException}
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.appinfo.{AppSelector, Selector}
-import mesosphere.marathon.plugin.auth.{AuthorizedAction, Authorizer, CreateRunSpec, Identity, UpdateRunSpec, ViewRunSpec}
+import mesosphere.marathon.plugin.auth._
```
timcharper: Our style guide go to is that we avoid these

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
timcharper: ➕ 1

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
timcharper: ➕ 1

timcharper:
It looks like there's an PR, https://github.com/mesosphere/marathon/pull/6655/files, that targets to merge in to this one.

Can we just combine those pull requests into this? Or, can you help me understand why we have two of them?

meln1k:
@jeschkies yes!

PR #6652, opened 2018-11-07
Remove LaunchEphemeral

alenkacz:
Summary:
This PR does not touch any production code, LaunchEphemeral was only used in tests.
LaunchEphemeral in production code was already replaced by Provision in the previous iterations.

JIRA issues: MARATHON-8221

### DIFF
--- a/src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
+++ b/src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
@@ -167,31 +167,6 @@ object TestInstanceBuilder {
     new Instance(task.taskId.instanceId, Some(agentInfo), state, tasksMap, task.runSpecVersion, unreachableStrategy, None)
   }
 
-  def runningInstance(appId: PathId, version: Timestamp, instanceTracker: InstanceTracker): Future[Instance] = async {
-    val instance: Instance = await(provisionedInstance(appId, version, instanceTracker))
-    val (taskId, _) = instance.tasksMap.head
-    // update to running
-    val taskStatus = TaskStatus.newBuilder
-      .setTaskId(taskId.mesosTaskId)
-      .setState(mesos.Protos.TaskState.TASK_RUNNING)
-      .setHealthy(true)
-      .build
-    await(instanceTracker.updateStatus(instance, taskStatus, Timestamp.now()))
-    await(instanceTracker.get(instance.instanceId).map(_.get))
-  }
-
-  def provisionedInstance(appId: PathId, version: Timestamp, instanceTracker: InstanceTracker): Future[Instance] = async {
-    val app = AppDefinition(appId, versionInfo = VersionInfo.OnlyVersion(version))
-    val scheduledInstance = Instance.scheduled(app)
-    // schedule
-    await(instanceTracker.schedule(scheduledInstance))
-    // provision
-    val provisionedInstance = scheduledInstance.provisioned(AgentInfoPlaceholder(), NetworkInfoPlaceholder(), app, Timestamp.now(), Task.Id.forInstanceId(scheduledInstance.instanceId))
-    await(instanceTracker.process(InstanceUpdateOperation.Provision(provisionedInstance)))
-
-    provisionedInstance
-  }
-
   val defaultAgentInfo = Instance.AgentInfo(
     host = AgentTestDefaults.defaultHostName,
     agentId = Some(AgentTestDefaults.defaultAgentId), region = None, zone = None, attributes = Seq.empty)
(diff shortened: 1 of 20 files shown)


src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -55,15 +56,14 @@ class TaskLauncherActorTest extends AkkaUnitTest with Eventually {
   object f {
     import org.apache.mesos.{Protos => Mesos}
     val app = AppDefinition(id = PathId("/testapp"))
-    val runningMarathonInstance = TestInstanceBuilder.newBuilder(app.id, version = app.version, now = Timestamp.now()).addTaskRunning().getInstance()
```
alenkacz: lots of changes here, but not necessarily related to LaunchEphemeral. Many tests in this actor did not make sense anymore as `TaskLauncherActor` does not compute it's internal state, instead it pull state all the time from instancetracker for every change.

src/test/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActorTest.scala
```diff
@@ -347,80 +314,36 @@ class TaskLauncherActorTest extends AkkaUnitTest with Eventually {
...
 
-    for (
-      update <- TaskConditionMapping.Gone.toSeq.map(reason => TaskStatusUpdateTestHelper.lost(reason, f.runningMarathonInstance))
-        .union(Seq(
-          TaskStatusUpdateTestHelper.finished(f.runningMarathonInstance),
-          TaskStatusUpdateTestHelper.killed(f.runningMarathonInstance),
-          TaskStatusUpdateTestHelper.error(f.runningMarathonInstance)))
-    ) {
-      s"Terminated task (${update.simpleName} with ${update.reason} is removed" in new Fixture {
-        Mockito.when(instanceTracker.instancesBySpecSync).thenReturn(InstanceTracker.InstancesBySpec.forInstances(f.runningMarathonInstance))
```
alenkacz: e.g. this task was really testing nothing, because the state in instancetracker is set to running instance so whatever update we send, it always gets the same state...

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
```diff
@@ -163,6 +167,31 @@ object TestInstanceBuilder {
     new Instance(task.taskId.instanceId, Some(agentInfo), state, tasksMap, task.runSpecVersion, unreachableStrategy, None)
   }
 
+  def runningInstance(appId: PathId, version: Timestamp, instanceTracker: InstanceTracker): Future[Instance] = async {
```
jeschkies: I much rather avoid this method. If there is some repetition in test we should have helper methods there. See my comment above.

src/test/scala/mesosphere/marathon/tasks/InstanceTrackerImplTest.scala
```diff
@@ -219,30 +222,12 @@ class InstanceTrackerImplTest extends AkkaUnitTest {
...
-      instanceTracker.updateStatus(app3_instance2, makeTaskStatus(app3_instance2, TaskState.TASK_RUNNING), clock.now()).futureValue
-
-      instanceTracker.launchEphemeral(app3_instance3).futureValue
-      instanceTracker.updateStatus(app3_instance3, makeTaskStatus(app3_instance3, TaskState.TASK_RUNNING), clock.now()).futureValue
+      val app1_instance1 = TestInstanceBuilder.runningInstance(appName1, Timestamp.now(), instanceTracker).futureValue
+      val app1_instance2 = TestInstanceBuilder.runningInstance(appName1, Timestamp.now(), instanceTracker).futureValue
+      val app2_instance1 = TestInstanceBuilder.runningInstance(appName2, Timestamp.now(), instanceTracker).futureValue
+      val app3_instance1 = TestInstanceBuilder.runningInstance(appName3, Timestamp.now(), instanceTracker).futureValue
+      val app3_instance2 = TestInstanceBuilder.runningInstance(appName3, Timestamp.now(), instanceTracker).futureValue
+      val app3_instance3 = TestInstanceBuilder.runningInstance(appName3, Timestamp.now(), instanceTracker).futureValue
```
jeschkies: What is the motivation for using a helper instead of calling the instance tracker directly? This is a test for the instance tracker. It should be explicit that we call it.

tools/elk/bin/target
```diff
@@ -1,5 +0,0 @@
-#!/bin/bash
-
```
timcharper: We should probably rename this file as it keeps getting accidentally removed

kensipe: Save the Target!!!

PR #6654, opened 2018-11-07
Fix application of timeouts in OverdueInstancesActor

meln1k:
Summary: According to the Mesos docs, the task lifecycle is Staging -> Starting. This PR swaps Starting and Staging timeouts in order to use them according to the Mesos implementation of the lifecycle.

Jira issues: MARATHON-8482


src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
```diff
@@ -65,12 +65,12 @@ private[jobs] object OverdueInstancesActor {
 
       def launchedAndExpired(task: Task): Boolean = {
         task.status.condition match {
-          case Condition.Provisioned | Condition.Starting if task.status.stagedAt < unconfirmedExpire =>
+          case Condition.Provisioned | Condition.Staging if task.status.stagedAt < unconfirmedExpire =>
```
zen-dog: I don't understand: `taskLaunchConfirmationTimeout` is defined as `Time, in milliseconds, to wait for a task to enter the TASK_STAGING state before killing it`. But in this line, we'll fail if `Condition.Staging` is achieved? Shouldn't it be only `Condition.Provisioned`?

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
```diff
@@ -65,12 +65,12 @@ private[jobs] object OverdueInstancesActor {
...
             logger.warn(s"Should kill: ${task.taskId} was launched " +
               s"${task.status.stagedAt.until(now).toSeconds}s ago and was not confirmed yet")
             true
 
-          case Condition.Staging if task.status.stagedAt < stagedExpire =>
+          case Condition.Starting if task.status.stagedAt < stagedExpire =>
```
zen-dog: Similar here: shouldn't it be `Condition.Staging | Condition.Starting` since `taskLaunchTimeout = Time, in milliseconds, to wait for a task to enter the TASK_RUNNING state before killing it`?

src/main/scala/mesosphere/marathon/core/task/jobs/impl/OverdueInstancesActor.scala
meln1k: Good point

alenkacz:
@zen-dog @meln1k so we're fine with no tests and no changelog?

zen-dog:
> We changed the behavior, I really think this deserves a changelog entry...

A very good point! Definitely worth an entry in _"Fixed Issues"_ with the JIRA ticket and a comment along the lines of `We fixed a possibly incorrect behavior around killing overdue tasks...` /cc @meln1k

changelog.md
```diff
@@ -3,6 +3,10 @@
 ### Removed deprecated metrics
 We removed deprecated Kamon based metrics from the code base (see the 1.7.xxx changelog for details on new metrics). This led to removal of deprecated command line arguments e.g. old reporters like `--reporter_graphite`, `--reporter_datadog`, `--reporter_datadog` and `--metrics_averaging_window`.
 
+### Fixed isssues
+
+- [MARATHON-8482](https://jira.mesosphere.com/browse/MARATHON-8482) - We fixed a possibly incorrect behavior around killing overdue tasks: we no treat tasks in `Starting` state using unconfirmed timeouts.
```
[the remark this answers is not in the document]
zen-dog: > we no treat tasks in `Starting` state using unconfirmed timeouts.

I'd be more verbose e.g.: `--task_launch_timeout parameter controls the time between the task is in Provisioned stage (between launching and receiving TASK_STAGING status update) and...`

PR #6655, opened 2018-11-09
 App update validator removed

meln1k:
Summary: App update validation process is simplified now: instead of validation app update and the resulting app definition, we merge the update first and validate the result. Note that app normalization stays there as before.

Jira issues: MARATHON-8498


src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -22,15 +22,12 @@ object AppHelpers {
...
-    enabledFeatures: Set[String], config: AppNormalization.Config): NormalizationWithContext[raml.AppUpdate, Map[String, SecretDef]] =
-    Normalization.withContext {
-      case (app, existingSecrets) =>
-        validateOrThrow(app)(AppValidation.validateOldAppUpdateAPI)
-        val migrated = AppNormalization.forDeprecatedUpdates(config).normalized(app)
-        validateOrThrow(app)(AppValidation.validateCanonicalAppUpdateAPI(enabledFeatures, () => config.defaultNetworkName, existingSecrets.getOrElse(Map.empty)))
-        AppNormalization.forUpdates(config).normalized(migrated)
-    }
+  def appUpdateNormalization(config: AppNormalization.Config): Normalization[raml.AppUpdate] = Normalization { app =>
+    //            validateOrThrow(app)(AppValidation.validateOldAppUpdateAPI)
```
jeschkies: This should go, right?

src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -333,12 +335,12 @@ class AppsResource @Inject() (
 
     val updateType = (appDoesNotExist, partialUpdate) match {
       case (true, _) => CompleteReplacement
-      case (_, true) => PartialUpdate
+      case (_, true) => PartialUpdate(maybeExistingApp.getOrElse(throw new RuntimeException("This should never happen")))
```
jeschkies: The exception message could be a little more explicit about what happened.

src/test/scala/mesosphere/marathon/api/v2/json/AppUpdateFormatTest.scala
```diff
@@ -45,13 +37,6 @@ class AppUpdateFormatTest extends UnitTest {
       appUpdate.acceptedResourceRoles should equal(Some(Set(ResourceRole.Unreserved)))
     }
 
-    "FromJSON should fail when 'acceptedResourceRoles' is defined but empty" in {
-      val json = """ { "id": "test", "acceptedResourceRoles": [] }"""
```
jeschkies: Is this still true for our update? Does the validator fail when we merge this update with an app?

src/test/scala/mesosphere/marathon/api/v2/json/AppUpdateTest.scala
```diff
@@ -210,53 +170,6 @@ class AppUpdateTest extends UnitTest with ValidationTestLike {
       assert(changed == app.copy(acceptedResourceRoles = Set("b")))
     }
 
-    "AppUpdate with a version and other changes are not allowed" in {
```
jeschkies: I think we have to make sure that there are still validations error with such updates. See my comment above.

src/test/scala/mesosphere/marathon/api/v2/json/AppUpdateFormatTest.scala
meln1k: Yes, check this line:

https://github.com/mesosphere/marathon/blob/master/src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionFormatsTest.scala#L213

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -22,15 +22,12 @@ object AppHelpers {
...
+    val migrated = AppNormalization.forDeprecatedUpdates(config).normalized(app)
+    //            validateOrThrow(app)(AppValidation.validateCanonicalAppUpdateAPI(enabledFeatures, () => config.defaultNetworkName, existingSecrets.getOrElse(Map.empty)))
```
zen-dog: Why was this disabled?

src/test/scala/mesosphere/marathon/api/validation/AppUpdateValidatorTest.scala
```diff
@@ -1,133 +0,0 @@
...
-import mesosphere.marathon.raml.{App, AppCContainer, AppUpdate, ContainerPortMapping, EngineType, Raml}
-import mesosphere.marathon.state.AppDefinition
-import org.scalatest.Matchers
-import play.api.libs.json.Json
-
-class AppUpdateValidatorTest extends UnitTest with Matchers {
-  implicit val appUpdateValidator = AppValidation.validateCanonicalAppUpdateAPI(Set.empty, () => AppNormalization.Configuration(None, "mesos-bridge-name").defaultNetworkName, Map.empty)
-  implicit val validAppDefinition = AppDefinition.validAppDefinition(Set.empty)(PluginManager.None)
-
-  "validation that considers container types" should {
```
zen-dog: What about these tests (and the ones above)? Are *all* those cases properly covered in other tests?

src/test/scala/mesosphere/marathon/api/v2/json/AppUpdateTest.scala
meln1k: Good catch! I added a test for this. I checked other parts too and it seems they cover everything.

src/test/scala/mesosphere/marathon/api/v2/AppsResourceTest.scala
```diff
@@ -1639,6 +1638,28 @@ class AppsResourceTest extends AkkaUnitTest with GroupCreation with JerseyTest {
...
+        cmd = Some("cmd"),
+        instances = 1
+      )
+      prepareApp(app, groupManager) // app is stored
+
+      When("The application is updated")
+      val updateRequest = App(id = id, instances = 2, version = Some(OffsetDateTime.now()))
+      val updatedBody = Json.stringify(Json.toJson(updateRequest)).getBytes("UTF-8")
+      val response = asyncRequest { r =>
+        appsResource.patch(app.id, updatedBody, force = false, auth.request, r)
```
jeschkies: What is the reason for testing this not in `AppValidationTest`?

src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -22,15 +22,11 @@ object AppHelpers {
...
-    Normalization.withContext {
-      case (app, existingSecrets) =>
-        validateOrThrow(app)(AppValidation.validateOldAppUpdateAPI)
-        val migrated = AppNormalization.forDeprecatedUpdates(config).normalized(app)
-        validateOrThrow(app)(AppValidation.validateCanonicalAppUpdateAPI(enabledFeatures, () => config.defaultNetworkName, existingSecrets.getOrElse(Map.empty)))
-        AppNormalization.forUpdates(config).normalized(migrated)
-    }
+  def appUpdateNormalization(config: AppNormalization.Config): Normalization[raml.AppUpdate] = Normalization { app =>
+    val migrated = AppNormalization.forDeprecatedUpdates(config).normalized(app)
+    validateOrThrow(migrated)(AppValidation.validateAppUpdateVersion)
```
jeschkies: Ah, so we do have validations that we cannot have as part of the whole app.

src/test/scala/mesosphere/marathon/api/v2/json/AppUpdateTest.scala
jeschkies: I can see

1. `assert(vfe.failure.violations.toString.contains("The 'version' field may only be combined with the 'id' field."))`
2. `assert(vfe.failure.violations.toString.contains("may not be set in conjunction with fetch"))`
3. `assert(vfe.failure.violations.toString.contains("cannot specify both ports and port definitions"))`
4. `assert(vfe.failure.violations.toString.contains("ports must be unique"))`

1 is covered in `AppResourceTest`. 4 is covered in `AppValidationTest`. Where do you test 2 and 4? I think we do have the validations but I'd like to confirm this with tests.

src/test/scala/mesosphere/marathon/api/validation/AppUpdateValidatorTest.scala
meln1k: I checked every test to make sure they don't disappear :)

PR #6656, opened 2018-11-12
Decrease backoff delay when application are Running

Mwea:
Summary:
Whenever an application succeed to start and was considered
as healthy, Marathon had to wait for the overdue cleanup to
reduce delays.

Now, if application succeed to start and run again, delay will
decrease progressively ( applying the inverse of the backoff factor).

JIRA issues: MARATHON-8450

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueue.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueue.scala
@@ -32,6 +32,9 @@ trait LaunchQueue {
   /** Advance the reference time point of the delay for the given RunSpec */
   def advanceDelay(spec: RunSpec): Unit
 
+  /** Decrease delay to the given RunnableSpec because of a running instance */
+  def decreaseDelay(spec: RunSpec): Unit
+
   /** Notify queue about InstanceUpdate */
   def notifyOfInstanceUpdate(update: InstanceChange): Future[Done]
 }
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueDelegate.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueDelegate.scala
@@ -53,6 +53,9 @@ private[launchqueue] class LaunchQueueDelegate(
   override def resetDelay(spec: RunSpec): Unit = rateLimiterActor ! RateLimiterActor.ResetDelay(spec)
 
   override def advanceDelay(spec: RunSpec): Unit = rateLimiterActor ! RateLimiterActor.AdvanceDelay(spec)
+
+  override def decreaseDelay(spec: RunSpec): Unit = rateLimiterActor ! RateLimiterActor.DecreaseDelay(spec)
+
 }
 
 private[impl] object LaunchQueueDelegate {
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/RateLimiter.scala
@@ -53,6 +53,13 @@ private[launchqueue] class RateLimiter(clock: Clock) extends StrictLogging {
     }
   }
 
+  def decreaseDelay(spec: RunSpec): Timestamp = {
+    setNewDelay(spec, "Decreasing delay") {
+      case Some(delay) => delay.decreased(clock, spec)
+      case None => Delay(clock, spec)
+    }
+  }
+
   private[this] def setNewDelay(spec: RunSpec, message: String)(calcDelay: Option[Delay] => Delay): Timestamp = {
     val maybeDelay: Option[Delay] = taskLaunchDelays.get(spec.configRef)
     val newDelay = calcDelay(maybeDelay)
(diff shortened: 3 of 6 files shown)


alenkacz:
@Mwea thanks for the PR! I've assigned @timcharper to take a look at this since he was the most involved in redesigning efforts around task delays.

clems4ever:
LGTM

timcharper:
@Mwea Thank you very much for submitting this! I've been thinking about this approach over the past couple of days, and specifically wondering what will happen under the following scenario:

1. A task fails 100 times (due to a database being down); we increase the relaunch delay by 100 backoff factors.
2. The task then succeeds, and is running for days.

My current understanding is that the delay will be backoff factor will be 99 backoff factors worth after 2, right? Do you have an idea for a more ideal reset mechanism? Just spitballing here, but, perhaps something along the lines of doing a full backoff delay reset if no failures encountered for 2x the max backoff delay? Would that solve the issue?

Mwea:
Hi @timcharper, thanks for your interest ! I'm not sure I can see your point in your example, could you clarify with a concrete example and explicit the problem we're facing with the actual approach ? 

Just for clarification, from what I understand, actually backoffs are reset if `now > (delay.referenceTimestamp + delay.maxLaunchDelay)`.  

Also, delay are multiplied by the backoff factor each time we add delay. Meaning that the delay is not : 
`backoff*factor*increaseDelayRequest`
but
`backoff*(factor^(increaseDelayRequest))`

With the approach given here, each time an App is `Running` we're then reducing by `a power of factor` more than removing one factor.

timcharper:
@Mwea thanks for your response!

To restate my example, say we have some web service with 3 instances which fails if MySQL is not running. It fails 100 times throughout the night and accumulates a delay. Then, the next morning, MySQL is restored and the web service successfully relaunches. In this case, we'll have 100 failures, followed by 3 successes. By my logic, this means we'll increase the delay by 100 factors, but then reduce it by only 3 factors.

Not to further complicate the discussion, but I think this approach will be problematic for services that don't immediately exit. Say for example the service launches, stays running for a few seconds, and then exits. Under this circumstance it seems that Marathon would:

1. Launch an task
2. Receive a task running notification (reducing the delay by a factor)
3. Receive a task failed notification (increasing the delay by a factor)
4. Repeat

Under this circumstance, it seems that the delay would never accumulate, right? Am I overlooking something?

This has been the subject of a lot of discussion recently. After discussing it with my colleagues, we've reached consensus on a definition of "stable" as described by this ticket: https://jira.mesosphere.com/browse/MARATHON-8504

What do you think of that approach? Do you think it would provide a solution for what you're experiencing?

Mwea:
Thanks for the link to the issue ! Now I understand the point you were trying to show. 

You're right about your explanation. From what I understand, the real issue with my PR is that we're decreasing as much as we're increasing the delay. Meaning that we're, in fact, limiting the amount of increase by the amount of instances.

After tacking time reading your proposal, I think it could lead to issues: 

For example, we've some applications that are scaled at 100 instances, but need only 30 of them to be considered as "stable". 
If this 100 instances starts to fails, the delay will then be multiplied 100 times. After some time, applications will try to restart, but will not be able to keep the load because there isn't enough instances started.
So they will fail again, and finally reach maximum delay and we will wait this whole time with 0 instances. 

In this situation, I think that we should allow to reduce the delay progressively, maybe 2 times less than increasing. So if applications handle the load for some time, we could decrease progressively and go back to a normal situation quicker.  What would you think of updating my PR so we reduce less the delay than I actually do ? 

At some point, your solution for "stable" state is really interesting, it could be really useful if the number of instances to be considered as stable was parameterised, like a ratio for example. What would you think of this ? 

Not really important but I'd like to point this part of your issue : 

> Currently, Marathon will never automatically reset the back off delay for a task. Each time a task dies (with a failure status or not), the back off delay is incremented. This will accumulate until either the operator updates the service configuration, or explicitly resets the delay via the API.

This is not totally true, actually the behavior is : 

> An app dies N times until it reaches its max delay, then it restarts and works.
At the moment the application starts to work, Marathon sends a `Running` state update, so it calls `advanceDelay`, which is reseting the `referenceTimestamp` to  `clock.now()` **but keeping the actual delay** (which is maxDelay)  stored in `taskLaunchDelays`. The delay will then be reset by `cleanUpOverdueDelays` after waiting again `maxDelay`. 

Meaning that, delay is reset at some point, but you just have to wait 2 times to be wiped. 😄 

From what I see, an ideal situation would be to have a parameterised option for "minimum stable instances" + a way to decrease progressively instances to let a huge pool of instances warm up quickly.  What do you think of this proposal ?

timcharper:
@Mwea thanks for bringing it to my attention that we do reset the backoff delay. You're totally right. I have no idea how I missed that; I actually just touched that code several weeks ago 🤦‍♂️ . I assumed it was just cleaning up delays for deleted apps.

> If this 100 instances starts to fails, the delay will then be multiplied 100 times. After some time, applications will try to restart, but will not be able to keep the load because there isn't enough instances started.
> So they will fail again, and finally reach maximum delay and we will wait this whole time with 0 instances.

If I understand your perspective, you're pointing out that the backoff delay gets applied more severely to apps with a higher instance count, because there are more initial launch attempts in the cohort, and therefore there are more opportunities for delays to accumulate? Just spitballing, what if we only applied the backoff if we're not already delayed? This way, if we had 30 instances failing, the delay would not accumulate so fast if they all failed at once? They'd increase their delay at about the same rate as 30 independent apps with one instance each.

> At some point, your solution for "stable" state is really interesting, it could be really useful if the number of instances to be considered as stable was parameterised, like a ratio for example. What would you think of this ?

From the standpoint of specifying a "stable" threshold as a percentage of instances running, there's the possibility that a task is not launching on some set of agents because the agent is missing some configuration, and we could overwhelm the agents in that case.

> In this situation, I think that we should allow to reduce the delay progressively, maybe 2 times less than increasing. So if applications handle the load for some time, we could decrease progressively and go back to a normal situation quicker. What would you think of updating my PR so we reduce less the delay than I actually do ?

I think this would be fine to change, although, I wonder, would it actually solve your problem? It seems it might have the same effect as effectively slowing down the rate at which the delay grows, but not necessarily do much to improve the odds of the delay being reset?

It seems like, potentially ideally, the backoff delay is applied per instance, rather than per app. Unfortunately, we're currently constrained to define our cohort as a runSpecId + config version as a brand new ephemeral instance is created with each relaunch (we are working to change that in 1.8). If the backoff were applied per instance, then each one or two flaky instances wouldn't affect the launch delay of instances that are launching healthily, nor prevent those delays from resetting.

PR #6616, opened 2018-10-14
Kill instances via goal and keep stable Ids

meichstedt:
This PR addresses several JIRAs. The reason why they are kept together in one PR is following:
- even though it is potentially several issues, the PR is still reasonably big to allow reviews
- separating the work based on the JIRAs would mean that we would have to land them separately and introduce some temporary “hacks” to make that separate PRs stable. Including stable ids just made some synchronization issues easier to address

# What’s included:

## Kill instances based on goal
This is the main issue addressed by this PR. Before we had to call everywhere in orchestration code the following:
- setGoal
- kill

The killing logic is now hidden inside scheduler so from orchestrator perspective you only set the goal to decommission and the killing will happen as a result of that.
That means that The KillServiceActor now listens to InstanceGoalChangeEvent and will add instances to its internal queue when appropriate.
As a result we had to introduced a way to wait for terminal instances. That’s done by `KillService.watchForKilledInstances`. We should potentially improve this logic in the future.

JIRA Issues: MARATHON-8296, MARATHON-8225

## Only expunge instances when they should not be replaced

Until now, Marathon expunged instances once all of their tasks ended up in a terminal state, and eventually launched replacements as a result of those instances being expunged.
This commit adjusts this behavior. Instances will now only be expunged from state when their goal is to be decommissioned and all their tasks are in a terminal state. If their goal is still Running, they will be considered to be scheduled, and used to launch replacement tasks.

This means that instanceIds are now stable for as long as an instance shall be kept running. New instances will be created only when replacing unreachable instances, and when replacing instances with new versions. Similar to the way we handle taskIds for stateful services, tasks of stateless services will now also provide an incarnation count, appended to the taskId. The first task created for an instance will be the .1, and subsequent replacements will increment that incarnation counter.
When killing an instance using the wipe=true flag, its goal will be set to decommissioned, and it will eventually be expunged when all tasks are terminal. Note that as long as its tasks are e.g. unreachable, it will not be expunged until they are reported terminal (GONE, GONE_BY_OPERATOR, or UNKNOWN).

When killing instances without the wipe=true or scale=true flags, Marathon will only issue kill requests to Mesos, but keep the current goal and will therefore launch replacements that are still associated with the existing instance.

JIRA Issues: MARATHON-8429, MARATHON-8373, MARATHON-8140

[...]

PR #6658, opened 2018-11-13
Remove reserved condition

alenkacz:
Summary:
This diff removes all usages of `Condition.Reserved` and `Condition.ReservedTerminal`.
A migrate replaces these conditions with `Condition.Finished` on instance and task
states.

JIRA issues: MARATHON-8374

Co-authored-by: Karsten Jeschkies <k@jeschkies.xyz>


src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
```diff
@@ -20,7 +20,7 @@ object InstanceUpdater extends StrictLogging {
 
     // We need to suspend reservation on already launched reserved instances
     // to prevent reservations being destroyed/unreserved.
-    val updatedReservation = if (updatedTask.status.condition == Condition.Reserved && !instance.reservation.exists(r => r.state.isInstanceOf[Reservation.State.Suspended])) {
+    val updatedReservation = if (updatedTask.status.condition.isTerminal && instance.hasReservation && !instance.reservation.exists(r => r.state.isInstanceOf[Reservation.State.Suspended])) {
```
jeschkies: Hm, how about `instance.isReserved`/`updatedTask.status.isReserved` with

```scala
def isReserved = isTerminal && hasReservation
```

It might not apply in this case but in others.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -264,7 +264,7 @@ private[impl] class LaunchQueueActor(
 
       // Reuse resident instances that are stopped.
       val existingReservedStoppedInstances = await(instanceTracker.specInstances(runSpec.id))
-        .filter(i => i.isReserved && i.state.goal == Goal.Stopped) // resident to relaunch
+        .filter(i => i.hasReservation && i.state.goal == Goal.Stopped) // resident to relaunch
```
jeschkies: I don't think this is correct. `isReserved` is defined as `isTerminal && hasReservation`. See my comment above.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -55,7 +55,7 @@ private[tracker] class InstanceTrackerDelegate(
   // TODO(jdef) support pods when counting launched instances
   override def countActiveSpecInstances(appId: PathId): Future[Int] = {
     import scala.concurrent.ExecutionContext.Implicits.global
-    instancesBySpec().map(_.specInstances(appId).count(instance => instance.isActive || (instance.isReserved && !instance.isReservedTerminal)))
+    instancesBySpec().map(_.specInstances(appId).count(instance => instance.isActive))
```
jeschkies: This is so much simpler! 🎉

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
```diff
@@ -22,7 +22,7 @@ class ScaleAppUpdateStepImpl @Inject() (
   private[this] lazy val schedulerActor = schedulerActorProvider.get()
 
   private[this] def scalingWorthy: Condition => Boolean = {
-    case Condition.Reserved | Condition.UnreachableInactive | _: Condition.Terminal => true
+    case Condition.UnreachableInactive | _: Condition.Terminal => true
```
jeschkies: Is this true? We have `ReservedTerminal` and `Reserved`. I'm not sure what `ReservedTerminal` was used for.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -249,36 +247,39 @@ trait PodStatusConversion {
     }.getOrElse(List.empty[ContainerEndpointStatus])
 
   def podInstanceState(
-    instanceCondition: core.condition.Condition,
+    instance: core.instance.Instance,
```
jeschkies: What is the motivation for this change?

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
```diff
@@ -140,7 +140,6 @@ class MarathonSchedulerActorTest extends AkkaUnitTest with ImplicitSender with G
         .addTaskDropped(containerName = Some("dropped"))
         .addTaskUnknown(containerName = Some("unknown"))
         .addTaskKilling(containerName = Some("killing"))
-        .addTaskReserved(containerName = Some("reserved"))
```
jeschkies: We should keep this test. There are still reserved tasks! We dropped the condition not task with reservations.

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
```diff
@@ -31,12 +31,6 @@ case class TestInstanceBuilder(instance: Instance, now: Timestamp = Timestamp.no
   def addTaskLost(since: Timestamp = now, containerName: Option[String] = None): TestInstanceBuilder =
     addTaskWithBuilder().taskLost(since, containerName).build()
 
-  def addTaskReserved(containerName: Option[String] = None): TestInstanceBuilder =
-    addTaskWithBuilder().taskReserved(containerName).build()
```
jeschkies: As I mentioned above we still have to have these. We did not remove reserved tasks.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
alenkacz: hmm I don't like the name `Reserved` to be honest, because in general sense reserved does not mean terminal. Instance that wasn't yet launched is also reserved in a sense...

But I would be totally open to other suggestions

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
alenkacz: hmm that's a fair point! because now we would reschedule tasks even before they were killed.

src/main/scala/mesosphere/marathon/core/task/update/impl/steps/ScaleAppUpdateStepImpl.scala
alenkacz: after phase 2 reservedterminal and reserved was the same see https://github.com/mesosphere/marathon/pull/6658/files#diff-4c766f1e11cd8f4e0f48c1a4392b2b41L43

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
alenkacz: I needed to do that because of this line https://github.com/mesosphere/marathon/pull/6658/files#diff-e0df4221bad191a89f0d04c9d30bbd06R253

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
alenkacz: well this piece of code really works with conditions only so it does not make sense to add residents here IMHO. But in general yes, this PR is definitely missing unit tests coverage. I'll follow up with that once I debug the integration test failures. Feel free to come up with something when I am sleeping :-) I am more than willing to collaborate on this

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jeschkies: Got it! Thanks!

docs/docs/rest-api/public/api/v2/schema/AppDefinition.json
```diff
@@ -120,7 +120,7 @@
           },
           {
             "type": "string",
-            "enum": ["UNIQUE", "CLUSTER", "GROUP_BY", "LIKE", "UNLIKE", "MAX_PER"]
+            "enum": ["UNIQUE", "CLUSTER", "GROUP_BY", "LIKE", "UNLIKE", "MAX_PER", "IS"]
```
jeschkies: Could you rebase or merge on master. I don't think this should be part of your diff.

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
alenkacz: these builders just created instances and tasks with condition reserved, this is really obsolete and we don't need that anymore since we dropped that condition.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/LaunchQueueActor.scala
```diff
@@ -270,7 +270,7 @@ private[impl] class LaunchQueueActor(
 
       // Reuse resident instances that are stopped.
       val existingReservedStoppedInstances = await(instanceTracker.specInstances(runSpec.id))
-        .filter(i => i.isReserved && i.state.goal == Goal.Stopped) // resident to relaunch
+        .filter(i => i.hasReservation && i.state.condition.isTerminal && i.state.goal == Goal.Stopped) // resident to relaunch
```
jeschkies: Hm, I think we can drop the `i.state.condition.isTerminal` in the future.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
```diff
@@ -249,36 +247,39 @@ trait PodStatusConversion {
...
-      case condition.Condition.Running =>
-        if (containerStatus.exists(_.conditions.exists { cond =>
-          cond.name == STATUS_CONDITION_HEALTHY && cond.value == "false"
-        }))
-          PodInstanceState.Degraded -> Some(MSG_INSTANCE_UNHEALTHY_CONTAINERS)
-        else
-          PodInstanceState.Stable -> None
+    if (instance.isScheduled) {
+      PodInstanceState.Pending -> None
+    } else {
```
jeschkies: I don't quite understand this change. We are checking for the scheduled condition below so why the extra if clause?

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
jeschkies: I don't know. I mean we check `assert(nonTerminalTasks.size == 5, "We should have 5 non-terminal tasks")` below. With this in still in place it should be the same, right? In any case I don't understand why you dropped this case.

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
jeschkies: Hm, how is a `TestTaskBuilder.taskResidentLaunched` different from `TestTaskBuilder.taskLaunched`? Seems we should `TestTaskBuilder.taskResidentLaunched` as well then.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdater.scala
jeschkies: Ok, let's leave it for now.

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
alenkacz: look at how was `Reserved` used in the previous diff. For reserved, we showed state as pending. `Reserved` condition is dropped now. When thinking in general, I believe that`Pending` is the correct `PodInstanceState` for instances that are SCHEDULED - meaning `isScheduled` is true

src/test/scala/mesosphere/marathon/MarathonSchedulerActorTest.scala
alenkacz: I tend to disagree. This code path is really working only with conditions and verifying condition mapping. You might argue that we should also test task with a long name, task with some label or task with whatever (e.g. reservation). Task does not even have reservation, that's instance. I really believe it does not make sense keeping it here.

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
alenkacz: `taskResidentLaunched` creates *instance with reservation* and a task that is running. I actually think that we are mixing two things there because reservation is really not a property of task. But I don't think it makes sense to invest more into those builders other than removing them and replacing them with production code

jeschkies: Hm, why not fix it? It's not a blocker for this PR but should it be 

```scala
TestInstanceBuilder(...).withReservation(...).addRunningTask()
```
then?

src/main/scala/mesosphere/marathon/raml/PodStatusConversion.scala
jeschkies: Ah, that relates to #6616 again where `condition = Scheduled` is not enough but one has to check `isScheduled`. Got it. Thanks for clarifying.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo18100.scala
```diff
@@ -0,0 +1,168 @@
...
+  import mesosphere.marathon.api.v2.json.Formats.TimestampFormat
+
+  /**
+    * Read format for instance state with reserved condition.
+    */
+  val migrationConditionReader = new Reads[Condition] {
+    private def readString(j: JsReadable) = j.validate[String].map {
+
+      case reserved if reserved.toLowerCase == "reserved" => Condition.Finished
+      case reservedTerminal if reservedTerminal.toLowerCase == "reservedterminal" => Condition.Finished
```
jeschkies: Infer condition from Mesos status.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo17.scala
```diff
@@ -134,7 +134,7 @@ object MigrationTo17 extends MaybeStore with StrictLogging {
     */
   def migrateInstanceGoals(instanceRepository: InstanceRepository, persistenceStore: PersistenceStore[_, _, _])(implicit mat: Materializer): Future[Done] = {
 
-    logger.info("Starting goal migration to Storage version 200")
+    logger.info("Starting goal migration to Storage version 17")
```
timcharper: Was this chance intentional?

jeschkies: Yes. The 200 was wrong.

PR #6659, opened 2018-11-14
Clean up instance update op resolver.

jeschkies:
Summary:
This simplifies the logic a little bit. Instead of receiving
`instancesBySpec` the resolve method just takes an optional instance
that should be updated or created.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
@@ -83,7 +83,7 @@ private[tracker] class InstanceTrackerDelegate(
     * We use a [[akka.stream.scaladsl.SourceQueue]] to serialize all instance updates *per Instance.Id*. This is important
     * since the way [[InstanceTrackerActor]] is applying/persisting those update to existing Instance state, having two
     * such update operation in parallel will result in later operation overriding the former one.
-    * 
+    *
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -97,14 +97,16 @@ private[tracker] class InstanceTrackerDelegate(
         val effectF = (instanceTrackerRef ? update)
           .mapTo[InstanceUpdateEffect]
           .transform {
-            case s@Success(_) => logger.info(s"Completed processing instance update ${update.operation.shortString}"); s
-            case f@Failure(e: AskTimeoutException) => logger.error(s"Timed out waiting for response for update $update", e); f
-            case f@Failure(t: Throwable) => logger.error(s"An unexpected error occurred during update processing of: $update", t); f
+            case s @ Success(_) =>
```
zen-dog: I don't like those new lines. Breaks the code alignment aesthetics :P

jeschkies: That's what sbt or IntelliJ is doing, dude. I'll revert it except that I prefer the spaces.

zen-dog: You allow your Intellij too much intellijence 🥁-tsss...

PR #6660, opened 2018-11-14
Define task id type in prefix of id string.

jeschkies:
Summary:
This introduces the `TaskIdWithIncarnation` id type. It is similar to
`ResidentTaskId` except that it prefixes the Mesos task id string with
`$tiwi$`. This makes parsing ids easier in the future, eg if we add the
version to the task id string with could introduce `TaskIdWithVersion`
with the type prefix `$tiv$`.

JIRA issues: MARATHON-8501

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -406,9 +406,9 @@ object Task {
 
     /**
       * Create a taskId for a pod instance's task. This will create a taskId designating the instance and each
-      * task container's name. It may be used for reservations for persitent pods as well.
+      * task container's name. It may be used for reservations for persistent pods as well.
       *
-      * @param instanceId the ID of the instance that this task is contained in
+      * @param instanceId the ID of the instance that contains this task.
       * @param container the name of the task as per the pod container config.
       */
     def forInstanceId(instanceId: Instance.Id, container: Option[MesosContainer] = None): Id = EphemeralOrReservedTaskId(instanceId, container.map(_.name))
--- a/src/main/scala/mesosphere/marathon/core/task/Task.scala
+++ b/src/main/scala/mesosphere/marathon/core/task/Task.scala
@@ -253,6 +253,8 @@ object Task {
     override val reservationId = runSpecId.safePath + separator + uuid
 
     override val containerName: Option[String] = None
+
+    def increaseAttempt(): LegacyResidentId = this.copy(attempt = attempt + 1L)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -309,6 +311,38 @@ object Task {
...
+    // Quick access to the underlying run spec identifier of the task.
+    override lazy val runSpecId: PathId = instanceId.runSpecId
+
+    override lazy val reservationId: String = instanceId.idString
+
+    def increaseAttempt(): TaskIdWithIncarnation = this.copy(attempt = attempt + 1L)
+  }
+  object TaskIdWithIncarnation {
+    val typePrefix = "$tiwi$"
+    val regex = """^\$tiwi\$\.(.+)\.(instance-|marathon-)([^_\.]+)[\._]([^_\.]+)\.(\d+)$""".r
```
meln1k: What if container name will contain dots? is such case filtered somewhere else?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -97,14 +97,16 @@ private[tracker] class InstanceTrackerDelegate(
...
           .mapTo[InstanceUpdateEffect]
           .transform {
-            case s@Success(_) => logger.info(s"Completed processing instance update ${update.operation.shortString}"); s
-            case f@Failure(e: AskTimeoutException) => logger.error(s"Timed out waiting for response for update $update", e); f
-            case f@Failure(t: Throwable) => logger.error(s"An unexpected error occurred during update processing of: $update", t); f
+            case s @ Success(_) =>
+              logger.info(s"Completed processing instance update ${update.operation.shortString}"); s
+            case f @ Failure(e: AskTimeoutException) =>
+              logger.error(s"Timed out waiting for response for update $update", e); f
+            case f @ Failure(t: Throwable) => logger.error(s"An unexpected error occurred during update processing of: $update", t); f
```
meln1k: Do you think `NonFatal(t)` is better here?

jeschkies: That's just a reformatting but yes 🙂

zen-dog: Why? Would you treat `NonFatal(t)` exceptions differently here? This whole `transform` part exists for logging purposes only.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -309,6 +311,38 @@ object Task {
...
+    *  - "$tiwi$.myGroup_myApp.marathon-b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6.$anon.1"
+    *  - "$tiwi$.myGroup_myApp.instance-b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6.$anon.3"
+    *  - "$tiwi$.myGroup_myApp.marathon-b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6.rails.2"
+    *  - "$tiwi$.myGroup_myApp.instance-b6ff5fa5-7714-11e7-a55c-5ecf1c4671f6.rails.42"
+    *
+    * @param instanceId Identifies the instance the task belongs to.
+    * @param containerName If set identifies the container in the pod. Defaults to [[Task.Id.Names.anonymousContainer]].
+    * @param attempt Counts how often a task has been launched on a specific reservation.
+    */
+  case class TaskIdWithIncarnation(val instanceId: Instance.Id, val containerName: Option[String], attempt: Long) extends Id {
```
timcharper: let's call this incarnation instead of attempt?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -336,6 +370,11 @@ object Task {
       */
     def apply(idString: String): Task.Id = {
       idString match {
+        case TaskIdWithIncarnation.regex(safeRunSpecId, prefix, uuid, container, attempt) =>
```
timcharper: same; attempt vs incarnation

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
timcharper: Agree with @zen-dog ; we're not intercepting exceptions, we're logging them.

Matching all throwables is fine. Let's do `f @ Failure(t)` though?

src/main/scala/mesosphere/marathon/core/task/Task.scala
jeschkies: I don't think so. So you mean the container `$tiwi$.`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
jeschkies: Good point.

src/main/scala/mesosphere/marathon/core/task/Task.scala
zen-dog: To be honest I think these `$tiwi$...` taksIds look kind of ugly. I like current current `group_app_UUID.N` pattern.

jeschkies: At the end it would be `group_app_-marathon-UUIID.$anon.$tiwi$.N`

alenkacz:
@zen-dog I think that this will be used by the Matthias's PR that has this incarnation and introduced similar construct as well...

src/main/scala/mesosphere/marathon/core/task/Task.scala
meichstedt: Why not just use a version? Start somewhere. `$v2` is fine, or any other arbitrary number as long as we increment the next one. An integer based taskId version is IMO preferable to a named version.

Apart from that, I guess a purist would ask for a UUID and sstore incarnation count and anything else in metdata, since serializing metadata in the ID is kind of weird.

jeschkies:
Superseded by #6674.

PR #6661, opened 2018-11-15
Attach run spec to instance.

jeschkies:
Summary:
Each instance hold a run spec id and the run spec version. This
is a unique identifier for the run spec. Thus we could just attach
the run spec itself to the instance. This will simplify our launching
and matching logic. It will also open the door to prototype templates.

See #6664 for an illustration. 

JIRA issues: MARATHON-8325


src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
```diff
@@ -7,7 +7,7 @@ import mesosphere.marathon.core.instance.update.{InstanceUpdateOperation, Instan
 import mesosphere.marathon.core.pod.MesosContainer
 import mesosphere.marathon.core.task.Task
 import mesosphere.marathon.core.task.state.{AgentInfoPlaceholder, AgentTestDefaults}
-import mesosphere.marathon.state.{PathId, RunSpec, Timestamp, UnreachableEnabled, UnreachableStrategy}
+import mesosphere.marathon.state._
```
zen-dog: Oh no, you did not! 😝

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImplTest.scala
```diff
@@ -61,5 +72,39 @@ class InstancesLoaderImplTest extends AkkaUnitTest {
...
+      for (instance <- instances) {
+        f.instanceRepository.get(instance.instanceId) returns Future.successful(Some(instance))
+      }
+
+      When("load is called")
+      val loaded = f.loader.load()
+
+      Then("the resulting data ")
+      // we do not need to verify the mocked calls because the only way to get the data is to perform the calls
+      val expectedData = InstanceTracker.InstancesBySpec.forInstances(app1Instance1, app1Instance2)
```
zen-dog: So without the correct _runSpec_ and _version_ the instance will *not be part of the state*... until a _StatusUpdate_ for it is received and it will be expunged (because its unknown)? But we won't actively reconcile for it since it is not part of the `instancesBySpec` call? So In a case, when Mesos doesn't know the task, we'll end up with some permanent garbage in ZK?

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
```diff
@@ -123,7 +123,7 @@ class TaskReplaceActor(
 
   def replaceBehavior: Receive = {
     // New instance failed to start, restart it
-    case InstanceChanged(id, `version`, `pathId`, condition, Instance(_, Some(agentInfo), _, _, _, _, reservation)) if !oldInstanceIds(id) && considerTerminal(condition) =>
+    case InstanceChanged(id, `version`, `pathId`, condition, Instance(_, Some(agentInfo), _, _, _, reservation)) if !oldInstanceIds(id) && considerTerminal(condition) =>
```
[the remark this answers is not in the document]
jeschkies: That's a good point. It was not trivial to refactor because the names were missing.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -5,25 +5,42 @@ import akka.stream.Materializer
...
+      // Join instances with app or pod.
+      val t = await(Future.sequence(instances.map { stateInstance =>
+        val runSpecId = stateInstance.instanceId.runSpecId
+        val runSpecVersion = stateInstance.runSpecVersion.toOffsetDateTime
+        groupRepository.runSpecVersion(runSpecId, runSpecVersion).map {
+          case Some(runSpec) =>
+            Some(stateInstance.toCoreInstance(runSpec))
+          case None =>
+            logger.warn(s"No run spec $runSpecId with version $runSpecVersion was found for instance ${stateInstance.instanceId}.")
+            //TODO(karsten): use latest runspec.
```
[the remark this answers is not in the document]
jeschkies: We should assign the latest version since this would be the behaviour for a restart. If the run spec does not exist we have a bug. Currently Marathon would just not reschedule an instance whose run spec is gone.

Cc @meichstedt and @alenkacz

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -1,29 +1,52 @@
...
+        val runSpecVersion = stateInstance.runSpecVersion.toOffsetDateTime
+        groupRepository.runSpecVersion(runSpecId, runSpecVersion).map(stateInstance -> _)
+      }.mapConcat {
+        case (stateInstance, Some(runSpec)) =>
+          List(stateInstance.toCoreInstance(runSpec))
+        case (stateInstance, None) =>
+          val runSpecId = stateInstance.instanceId.runSpecId
+          logger.warn(s"No run spec $runSpecId with version ${stateInstance.runSpecVersion} was found for instance ${stateInstance.instanceId}.")
+          //TODO(karsten): use latest runspec.
+          //val maybeRunSpec = groupRepository.runSpecVersions(runSpecId).runWith(SinkMax).flatMap{ v => groupRepository.runSpecVersion(runSpecId, v)}
```
jeschkies: Yeah, I don't think that streams are much better than async/await in this case. `Future.sequence` could be nicer but this just seems overhead for a simple logic.

zen-dog: In the worst case scenario `Future.sequence` will create 1M futures at once.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -97,14 +97,16 @@ private[tracker] class InstanceTrackerDelegate(
         val effectF = (instanceTrackerRef ? update)
           .mapTo[InstanceUpdateEffect]
           .transform {
-            case s@Success(_) => logger.info(s"Completed processing instance update ${update.operation.shortString}"); s
-            case f@Failure(e: AskTimeoutException) => logger.error(s"Timed out waiting for response for update $update", e); f
-            case f@Failure(t: Throwable) => logger.error(s"An unexpected error occurred during update processing of: $update", t); f
+            case s @ Success(_) =>
```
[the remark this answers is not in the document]
zen-dog:
```
// format: off

// format: on
```

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -4,26 +4,47 @@ package core.task.tracker.impl
...
```
zen-dog: I propose a new method in the `groupRepository`: `lastSpecVersion`

src/main/scala/mesosphere/marathon/state/Instance.scala
```diff
@@ -0,0 +1,81 @@
...
+  import mesosphere.marathon.api.v2.json.Formats.TimestampFormat
+
+  implicit val instanceJsonWrites: Writes[Instance] = {
+    (
+      (__ \ "instanceId").write[CoreInstance.Id] ~
+      (__ \ "agentInfo").writeNullable[CoreInstance.AgentInfo] ~
+      (__ \ "tasksMap").write[Map[Task.Id, Task]] ~
+      (__ \ "runSpecVersion").write[Timestamp] ~
+      (__ \ "state").write[InstanceState] ~
+      (__ \ "unreachableStrategy").write[raml.UnreachableStrategy] ~ // TODO(karsten): Should we remove this after MARATHON-8325?
```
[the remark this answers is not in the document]
zen-dog: Ok, it's weird to have some arbitrary part of the `RunSpec` persisted with the Instance. I'd say we should get rid of it, either in this PR or in the follow up one.

src/test/scala/mesosphere/marathon/core/instance/TestInstanceBuilder.scala
jeschkies: 🔔

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -24,7 +24,7 @@ import mesosphere.marathon.core.pod.{PodDefinition, PodManager}
 import mesosphere.marathon.core.storage.repository.RepositoryConstants
 import mesosphere.marathon.plugin.auth._
 import mesosphere.marathon.raml.{Pod, Raml}
-import mesosphere.marathon.state.{PathId, Timestamp, VersionInfo}
+import mesosphere.marathon.state.{Instance => StateInstance, PathId, Timestamp, VersionInfo}
```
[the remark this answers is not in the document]
jeschkies: Thanks. Much cleaner!

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -4,26 +4,49 @@ package core.task.tracker.impl
...
+              (stateInstance, maybeRunSpec)
+          }
+        }
+      }.mapConcat {
+        case (stateInstance, Some(runSpec)) =>
+          List(stateInstance.toCoreInstance(runSpec))
+        case (stateInstance, None) =>
+          val runSpecId = stateInstance.instanceId.runSpecId
+          logger.warn(s"No run spec $runSpecId with any version was found for instance ${stateInstance.instanceId}. Ignoring.")
+          List.empty[Instance]
```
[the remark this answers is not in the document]
jeschkies: The type inference got it wrong. So I had to use empty.

src/main/scala/mesosphere/marathon/core/deployment/impl/TaskReplaceActor.scala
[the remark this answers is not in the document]
zen-dog: __@~··___@~··_____

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImplTest.scala
jeschkies: > So In a case, when Mesos doesn't know the task, we'll end up with some permanent garbage in ZK?

You mean Mesos forgets about the instance for some reason and Marathon forgets about the run spec for some reason and then we don't load it but keep it in ZooKeeper? Well, the bug is that we should never have instances without a run spec to begin with. That is a bug. Even in the current version we always should have the run spec. We do handle that situation by just not scaling https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/MarathonSchedulerActor.scala#L492. That means we never start a new instance for a failed if we do not know the run spec.

What would be the best behaviour here? Current Marathon will know the instance but not the run spec and thus won't be able to handle it. What do you think about `case object Unknown extends RunSpec`?

zen-dog: No, I meant, if there is an instance but for some reason, `RunSpec` was deleted. Do you think we should we log the problem and expunge the Instance from the state?

src/main/scala/mesosphere/marathon/state/Instance.scala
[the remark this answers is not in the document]
jeschkies: See follow up #6675

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImplTest.scala
jeschkies: Ah, that would be simpler but might be unexpected behaviour since Marathon will kill that instance which it would do with this change as well.

If we do not load that instance we should delete it from the state as well. If we keep it in state we should load it with an `UnknownRunSpec`.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -4,26 +4,49 @@ package core.task.tracker.impl
...
-    } yield {
-      logger.info(s"Loaded ${instances.size} instances")
-      InstanceTracker.InstancesBySpec.forInstances(instances)
-    }
+  override def load(): Future[InstanceTracker.InstancesBySpec] = async {
+    val instances = repo.ids().mapAsync(parallelism = 5)(repo.get).mapConcat(_.toList)
+
+    // Join instances with app or pod.
+    val coreInstances = await(
+      instances.mapAsync(parallelism = 5) { stateInstance =>
```
zen-dog: dito

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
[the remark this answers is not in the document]
jeschkies: Yes. If we want to keep the current Marathon behaviour we would have to introduce 

```scala
case class Unknown(id: Path, version: Timestamp) extends RunSpec
```

we would keep the instance but never restart it.

src/main/scala/mesosphere/marathon/storage/repository/Repositories.scala
```diff
@@ -53,6 +54,27 @@ trait GroupRepository {
   def podVersions(id: PathId): Source[OffsetDateTime, NotUsed]
 
   def podVersion(id: PathId, version: OffsetDateTime): Future[Option[PodDefinition]]
+
+  def runSpecVersion(id: PathId, version: OffsetDateTime)(implicit executionContext: ExecutionContext): Future[Option[RunSpec]] = {
```
zen-dog: It's kinda funny that we ourselves don't know if the instance belongs to an app or a pod and have to check both.

zen-dog: It's kinda funny that we ourselves don't know if the instance belongs to an app or a pod and have to check both.

jeschkies: Yep. It's not encoded in the id.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImplTest.scala
zen-dog: So we either:
a) Expunge the instance from the state right away and change how we deal with orphaned tasks - we shouldn't kill them just because we lost their state or
b) Keep the instance with `UnknownRunSpec` in the state and deal with implications or e.g. rescheduling Instances in the scheduling/orchestrator layer

src/main/scala/mesosphere/marathon/state/Instance.scala
```diff
@@ -0,0 +1,80 @@
...
+import core.instance.{Reservation, Instance => CoreInstance}
+import mesosphere.marathon.core.condition.Condition
+import mesosphere.marathon.core.instance.Instance.InstanceState
+import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.raml.Raml
+import play.api.libs.functional.syntax._
+import play.api.libs.json.Reads._
+import play.api.libs.json._
+
+case class Instance(
```
alenkacz: why do we have to have two anyway? Why can't we just when serializing "core instance" say that we use only the version of runspec when creating the protobuf?

src/main/scala/mesosphere/marathon/core/task/tracker/InstanceTrackerModule.scala
```diff
@@ -19,14 +19,15 @@ class InstanceTrackerModule(
...
+    groupRepository: GroupRepository,
     updateSteps: Seq[InstanceChangeHandler])(implicit mat: Materializer) {
   lazy val instanceTracker: InstanceTracker =
     new InstanceTrackerDelegate(metrics, clock, config, instanceTrackerActorRef)
   lazy val instanceTrackerUpdateStepProcessor: InstanceTrackerUpdateStepProcessor =
     new InstanceTrackerUpdateStepProcessorImpl(metrics, updateSteps)
 
   private[this] lazy val updateOpResolver: InstanceUpdateOpResolver = new InstanceUpdateOpResolver(clock)
-  private[this] lazy val instancesLoader = new InstancesLoaderImpl(instanceRepository)
+  private[this] lazy val instancesLoader = new InstancesLoaderImpl(instanceRepository, groupRepository, config)
```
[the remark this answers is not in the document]
jeschkies: We do. The `InstanceLoaderImpl` expects only `config: InstanceTrackerConfig`. I'm not sure how one would extract the config we need here and what the value is.

src/main/scala/mesosphere/marathon/state/Instance.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+  override def mergeFromProto(bytes: Array[Byte]): Instance = {
+    mergeFromProto(Protos.Json.parseFrom(bytes))
+  }
+  override def toProto: Protos.Json = {
+    Protos.Json.newBuilder().setJson(Json.stringify(Json.toJson(this))).build()
+  }
+  override def version: Timestamp = runSpecVersion
+
+  def toCoreInstance(runSpec: RunSpec) = CoreInstance(instanceId, agentInfo, state, tasksMap, runSpec, reservation)
```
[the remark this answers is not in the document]
jeschkies: I thought about `Instance(stateInstance, runSpec)`. However, that leaks the state into our internal logic. I would like to avoid that. Also, our internal instance should not really know about the state instance. It does not care how it is persisted.

src/main/scala/mesosphere/marathon/state/Instance.scala
jeschkies: Serializing is easy but deserializing is not. How would you deserialize to the core instance if you do not have the run spec? I'm happy for any suggestion. The state instance helps to handle migrations without loading the run spec. It is simply a lean storage model.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
```diff
@@ -4,26 +4,59 @@ package core.task.tracker.impl
...
+        async {
+          val runSpecId = stateInstance.instanceId.runSpecId
+          val runSpecVersion = stateInstance.runSpecVersion.toOffsetDateTime
+          await(groupRepository.runSpecVersion(runSpecId, runSpecVersion)) match {
+            case Some(runSpec) => (stateInstance, Some(runSpec))
+            case None =>
+              logger.warn(s"No run spec $runSpecId with version ${stateInstance.runSpecVersion} was found for instance ${stateInstance.instanceId}. Trying latest.")
+              await(groupRepository.latestRunSpec(runSpecId)) match {
+                case Some(runSpec) => (stateInstance, Some(runSpec))
+                case None => await(expungeOrphanedInstance(stateInstance))
```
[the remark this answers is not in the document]
jeschkies: Nothing is listening at this point. The instance tracker is one of the first things being loaded. Most other entities assume a running instance tracker. I understand that we could handle this somewhere else but then again, what part in our code base should deal with orphaned instances and why? Ideally this never happens to begin with.

On a meta level I do not like our approach with events. However, that is something we'll discuss for our scheduler interface, I guess.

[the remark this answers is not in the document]
alenkacz: for me it's like garbage collection, so I would have a separate entity (actor or whatever) doing that cleaning. That would be triggered e.g. by event. You don't call java garbage collection in java directly as well...

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
```diff
@@ -280,7 +280,7 @@ class PodsResource @Inject() (
       validateOrThrow(instanceId)
       val parsedInstanceId = Instance.Id.fromIdString(instanceId)
       val instances = await(taskKiller.kill(id, _.filter(_.instanceId == parsedInstanceId), wipe))
-      instances.headOption.fold(unknownTask(instanceId))(instance => ok(jsonString(instance)))
+      instances.headOption.fold(unknownTask(instanceId))(instance => ok(jsonString(state.Instance.fromCoreInstance(instance))))
```
alenkacz: Following up on the discussion on core vs state instance, I guess I mainly don't like that API layer (so the highest one) must convert between those two. Ideally for me there will be an interface (e.g. repository or service or whatever) that will accept core instances and give you core instances when you ask for them and the whole fact that the serialization format is different will be hidden behind that interface. What do you think about that?

alenkacz: oh ... why do we need state.Instance here anyway? Why can't we use the core instance for serialization?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
```diff
@@ -153,7 +153,7 @@ private[impl] class InstanceTrackerActor(
 
         updateEffect match {
           case InstanceUpdateEffect.Update(instance, _, _) =>
-            repository.store(instance).onComplete {
+            repository.store(state.Instance.fromCoreInstance(instance)).onComplete {
```
alenkacz: I this could be the interface I wrote in one of the previous comments... It would accept core instances and give you core instances and all the transformation will happen behind the curtains... As a consumer of repository I don't really care about how the concrete serialization format looks like

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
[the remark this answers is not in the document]
zen-dog: > You don't call java garbage collection in java directly as well...

unless you do `System.gc()` ;)

Like Karsten said - nothing is listening at that point, and the one responsible for expunging them is the instance tracker itself so ¯\_(ツ)_/¯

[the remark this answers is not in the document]
jeschkies: I get your point @alenkacz. However,
* We would have to load them first to garbage collect. That would require some run spec or `Option.empty[RunSpec]` or something.
* It would complexity we might not want.

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jeschkies: As I mentioned [above](https://github.com/mesosphere/marathon/pull/6661#discussion_r236960675). Serialization is not the issue. Deserialization is. How can the `InstanceRepository` set the run spec? It does not "know" about the `GroupRepository`.

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActor.scala
jeschkies: I think we were there two years ago. Back then I thought it would be nice to have storage models the same way we have API models. I do not think that the repositories should return the internal models but the storage models. In any case, I'm happy to create a wrapper repository on the `InstanceRepository` and `GroupRepository` that converts the storage model to the internal model. IMHO that is some unnecessary overhead until we really need more joins like these. I mean, the `InstanceLoderImpl` is doing exactly what you want for loads at least. Would you be fine if that becomes a layer between our core logic and the repositories?

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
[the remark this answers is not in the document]
jeschkies: Hm, I could add a prune step and then load with an adapter repo that has access to run specs and instances. Would that be ok?

src/main/scala/mesosphere/marathon/api/v2/PodsResource.scala
jeschkies: Just saw that this is the API layer. We could serialize core directly. Well, It should be a RAML first anyways and then serialized.

jeschkies: Will be addressed in https://github.com/mesosphere/marathon/pull/6703.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImplTest.scala
```diff
@@ -1,64 +1,153 @@
...
+      f.instanceRepository.delete(app2Instance1.instanceId) returns Future.successful(Done)
+
       When("load is called")
       val loaded = f.loader.load()
 
-      Then("the resulting data is correct")
+      Then("the resulting data ")
       // we do not need to verify the mocked calls because the only way to get the data is to perform the calls
-      val expectedData = InstanceTracker.InstancesBySpec.forInstances(app1Instance1, app1Instance2, app2Instance1)
+      val expectedData = InstanceTracker.InstancesBySpec.forInstances(app1Instance1, app1Instance2)
```
zen-dog: I think in this particular test we should verify that `expungeOrphanedInstance` was called, since that's the contract of current implementation

src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstancesLoaderImpl.scala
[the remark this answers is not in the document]
alenkacz: btw. this is not a blocker for me :) I am just expressing my "gold state" wish for that piece of code. I don't mean we have to do this in this PR... I would be fine with merging this but had a JIRA/googledoc or whatever with some plan on how to improve that :)

PR #6663, opened 2018-11-15
Add 'IS' Operator to placement constraint

daynelucasrax:
Add IS Operator to Placement Constraint

Summary:
Add the IS operator in placement constraint or application schema documentation.

JIRA issues:
N/A


timcharper:
Hi there @daynelucasrax ! We can certainly merge this, but just FYI this is deprecated and in Marathon 1.8 it will be completely inaccessible via the API (and off by default in 1.7). See https://github.com/mesosphere/marathon/blob/master/changelog.md#v2schemas for more info :)

PR #6664, opened 2018-11-16
Use run spec attached to instance for launching.

jeschkies:
Summary:
With #6661 we attach the run spec of an instance to the instance itself.
We should use that run spec for matching offers and launching instances
instead of using the latest.

Depends on #6661

JIRA issues: MARATHON-8325


src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -47,25 +47,18 @@ class InstanceOpFactoryImpl(
...
+        if (app.isResident) inferForResidents(request, scheduledInstance)
+        else inferNormalTaskOp(app, request.instances, request.offer, request.localRegion, scheduledInstance)
+      case Some(scheduledInstance @ Instance(_, _, _, _, pod: PodDefinition, _)) =>
+        if (pod.isResident) inferForResidents(request, scheduledInstance)
+        inferPodInstanceOp(pod, request.instances, request.offer, request.localRegion, scheduledInstance)
+      case Some(Instance(_, _, _, _, runSpec, _)) =>
+        throw new IllegalArgumentException(s"unsupported runSpec object ${runSpec}")
+      case None =>
+        // TODO(karsten): Ideally scheduled instances would be a non empty seq
+        OfferMatchResult.NoScheduledInstances(request.offer, clock.now())
```
jeschkies: Cats offers a `NonEmptyList` are we brace enough to import it or should we have an `IllegalStateException` instead? There should be no offer matching when no instance is scheduled.

zen-dog: I agree - NonEmptyList would make sense.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -47,25 +47,18 @@ class InstanceOpFactoryImpl(
   override def matchOfferRequest(request: InstanceOpFactory.Request): OfferMatchResult = {
     logger.debug("matchOfferRequest")
```
zen-dog: Since you're in this area - please either remove or improve this `debug` message

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -153,12 +146,19 @@ class InstanceOpFactoryImpl(
...
-    val InstanceOpFactory.Request(runSpec, offer, instances, _, localRegion) = request
+  private def maybeLaunchOnReservation(request: InstanceOpFactory.Request, scheduledInstance: Instance): Option[OfferMatchResult] = if (request.hasWaitingReservations) {
+    val InstanceOpFactory.Request(offer, instances, _, localRegion) = request
+
+    val runSpec = scheduledInstance.runSpec
 
     logger.debug(s"Need to launch on reservation for ${runSpec.id}, version ${runSpec.version}")
     val maybeVolumeMatch = PersistentVolumeMatcher.matchVolumes(offer, request.reserved)
 
+    maybeVolumeMatch.foreach { volumeMatch =>
```
zen-dog: Why this sanity check? When can it be *not* the case?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -183,7 +183,7 @@ class InstanceOpFactoryImpl(
 
       resourceMatchResponse match {
         case matches: ResourceMatchResponse.Match =>
-          val instanceOp = launchOnReservation(runSpec, offer, volumeMatch.instance, matches.resourceMatch, volumeMatch)
+          val instanceOp = launchOnReservation(volumeMatch.instance.runSpec, offer, volumeMatch.instance, matches.resourceMatch, volumeMatch)
```
zen-dog: Shouldn't you use the `runSpec` extracted in https://github.com/mesosphere/marathon/pull/6664/files#diff-b201fe360b90772b45e15a0a9df02227R152

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -306,6 +306,9 @@ private class TaskLauncherActor(
           logger.info(s"Did not match offer ${offer.getId} for run spec ${runSpec.id}, ${runSpec.version}.")
           offerMatchStatistics.offer(OfferMatchStatistics.MatchResult(notMatched))
           promise.trySuccess(MatchedInstanceOps.noMatch(offer.getId))
+        case notMatched: OfferMatchResult.NoScheduledInstances =>
```
zen-dog: We should theoretically never get here since `shouldLaunchInstances` [here](https://github.com/mesosphere/marathon/pull/6664/files#diff-f91c767f12c6ca4b33353f65393f692dR290) already checks for `scheduledInstances.nonEmpty`. `NonEmptyList` like a good idea.

jeschkies:
And the debugging starts. Deployments do not finish...

src/main/scala/mesosphere/marathon/core/launchqueue/LaunchQueue.scala
```diff
@@ -17,9 +17,6 @@ trait LaunchQueue {
   /** Request to launch `count` additional instances conforming to the given run spec. */
   def add(spec: RunSpec, count: Int = 1): Future[Done]
 
-  /** Update the run spec in a task launcher actor. **/
-  def sync(spec: RunSpec) = add(spec, 0)
```
zen-dog: 👏 👏 👏

jeschkies:
Loop is running here https://jenkins.mesosphere.com/service/jenkins/view/Marathon/job/marathon-sandbox/job/marathon-loop-pr6664/.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: Yep. It would be a bug. I'd use [cats](https://typelevel.org/cats/)'s `NonEmptyList`. That opens a door we can never close 🙈.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
jeschkies: Nope. It was a bug from my side.

jeschkies:
The loop is green! 🥳

src/main/scala/mesosphere/marathon/core/launcher/OfferMatchResult.scala
```diff
@@ -32,5 +32,9 @@ object OfferMatchResult {
 
   case class NoMatch(runSpec: RunSpec, offer: Mesos.Offer, reasons: Seq[NoOfferMatchReason], timestamp: Timestamp) extends OfferMatchResult
 
+  case class NoScheduledInstances(offer: Mesos.Offer, timestamp: Timestamp) extends OfferMatchResult {
+    override def runSpec: RunSpec = ???
```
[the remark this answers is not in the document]
jeschkies: That's a good point. I'll make it an `IllegalStateException`. What do you think?

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: I implemented a simple `NonEmptyIterable`.

PR #6665, opened 2018-11-16
Fix flaky `LaunchQueueIntegrationTest`.

jeschkies:


jeschkies:
The issues is a combination of @zen-dog and my change 9f13c6a6596a58a4fc749f84e949390ed1a49017 and @timcharper's 2dfd4206c33e3262f176b2179531debb8aab5acf. Instance update are recorded with a delay by `LaunchStats`.

@timcharper what is the motivation for keeping a separate state of the launching instances in [LaunchStats](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/launchqueue/LaunchStats.scala#L41) instead of querying the `InstanceTracker`? 

Alex mentioned that you'd like to separate the interface. This seems fine to me. IMHO this relates to [MARATHON-8445](https://jira.mesosphere.com/browse/MARATHON-8445).

cc. @meichstedt and @alenkacz.

timcharper:
@jeschkies 

> @timcharper what is the motivation for keeping a separate state of the launching instances in LaunchStats instead of querying the InstanceTracker?

I'd make one small modification to that statement and use "a separate view of the state". The goal was separation of concerns; if I could avoid InstanceTracker from having specific knowledge or functionality that is needed only by LaunchStats, then I'd prefer that. This is generally considered a good design goal in software.

Further, as a side benefit, and this would be a premature optimization and insufficient on its own, but by keeping a streaming aggregate with only the instances that are in the conditions deemed as "trying to launch", we avoid the need to do a full scan of all running instances for each request to /v2/queue.

Finally, subscribing to instance state is already something of interest to us, and it's a pattern I'd prefer in the future. It's generic and can serve a wide array of use cases while keeping the InstanceTrackerActor more ignorant to all of the ways it is being used.

timcharper:
@jeschkies 

> @timcharper what is the motivation for keeping a separate state of the launching instances in LaunchStats instead of querying the InstanceTracker?

I'd make one small modification to that statement and use "a separate view of the state". The goal was separation of concerns; if I could avoid InstanceTracker from having specific knowledge or functionality that is needed only by LaunchStats, then I'd prefer that.

Further, as a side benefit, and this would be a premature optimization and insufficient on its own, but by keeping a streaming aggregate with only the instances that are in the conditions deemed as "trying to launch", we avoid the need to do a full scan of all running instances for each request to /v2/queue.

Finally, subscribing to instance state is already something of interest to us, and it's a pattern I'd prefer in the future. It's generic and can serve a wide array of use cases while keeping the InstanceTrackerActor more ignorant to all of the ways it is being used.

PR #6666, opened 2018-11-16
Using one curator instance for leader election and data transfer

zen-dog:
Previously Marathon used two separate apache curator instances and hence two zookeeper connections - one for leader election and one for data transfer. This could have potential implications should leader client disconnect and the data client continues working and additionally wastes memory and threads. This commit merges both clients to one.

Fixes: MARATHON-8509

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -88,7 +88,7 @@ class CoreModuleImpl @Inject() (
     eventStream,
     hostPort,
     crashStrategy,
-    richCuratorFramework.client,
+    richCuratorFramework.client.usingNamespace(null), // using non-namespaced client for leader-election
     ExecutionContext.fromExecutor(electionExecutor)
   )
 
--- a/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
@@ -75,7 +75,11 @@ class CoreModuleImpl @Inject() (
 
   override lazy val config = ConfigFactory.load()
 
-  override lazy val metricsModule = MetricsModule(marathonConf, config)
+  // Initialize Apache Curator Framework (wrapped in [[RichCuratorFramework]] and connect/sync with the storage
+  // for an underlying Zookeeper storage.
+  lazy val richCuratorFramework: RichCuratorFramework = StorageConfig.curatorFramework(marathonConf, crashStrategy, lifecycleState)
+
+  override lazy val metricsModule = MetricsModule(marathonConf)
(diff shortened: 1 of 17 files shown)


src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
```diff
@@ -88,7 +88,7 @@ class CoreModuleImpl @Inject() (
     eventStream,
     hostPort,
     crashStrategy,
-    richCuratorFramework.client,
+    richCuratorFramework.client.usingNamespace(null), // using non-namespaced client for leader-election
```
timcharper: hmm, can we go the other way? Start with no namespace, and then add it for the storage parts?

zen-dog: I guess we could but should we? `/marathon/leader-curator` is the only exception - everything else *should* got into`/marathon/state`

timcharper: Personally, I'd prefer it, but I don't have an exact logical explanation as to why. I just find it a little surprising to see us set some namespace on the original connection, only to undo it later on for a different use case. There'd be less "overriding" or more "refining" if we started with a more general connection config, first.

src/main/scala/mesosphere/marathon/core/base/CrashStrategy.scala
```diff
@@ -29,3 +29,12 @@ case object JvmExitsCrashStrategy extends CrashStrategy {
     Runtime.getRuntime.asyncExit(reason.code)(ExecutionContext.Implicits.global)
   }
 }
+
+/**
+  * For testing purposes only: This crash strategy will *NOT* crash but throw an exception
+  */
+case object ThrowExceptionAndDontCrashStrategy extends CrashStrategy {
```
timcharper: let's put this in src/test then?

timcharper: alongside src/test/scala/mesosphere/marathon/test/TestCrashStrategy.scala

src/main/scala/mesosphere/marathon/core/CoreModuleImpl.scala
zen-dog: > set some namespace on the original connection, only to undo it later on for a different use case.

I'd say there is more potential to screw things up, by name-spacing (or rather forgetting) *every other* usage of the client *except* for leader-election. Again, a *namespace is the rule, no namespace is the exception*. But if you absolutely insist on it...

src/main/scala/mesosphere/marathon/core/base/CrashStrategy.scala
zen-dog: Good point. Done!

PR #6667, opened 2018-11-21
Rename GoalChange to ChangeGoal

alenkacz:
This change was requested in https://github.com/mesosphere/marathon/pull/6616

It follow up on our convention to be imperative when it comes to these names.


src/main/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegate.scala
```diff
@@ -97,14 +97,16 @@ private[tracker] class InstanceTrackerDelegate(
         val effectF = (instanceTrackerRef ? update)
           .mapTo[InstanceUpdateEffect]
           .transform {
-            case s@Success(_) => logger.info(s"Completed processing instance update ${update.operation.shortString}"); s
-            case f@Failure(e: AskTimeoutException) => logger.error(s"Timed out waiting for response for update $update", e); f
-            case f@Failure(t: Throwable) => logger.error(s"An unexpected error occurred during update processing of: $update", t); f
+            case s @ Success(_) =>
```
alenkacz: formatting 🤷‍♀️

jeschkies: @zen-dog keeps complaining about this 😂

PR #6669, opened 2018-11-21
Fix for a secrets validator error when we change env without changing…

meln1k:
… secrets (#6651)

Summary: Secrets validator was only passing if we submit a secret environment variable update with a corresponding secret definition section. Because of this, the app update validation process is simplified now: instead of validating an app update and the resulting app definition, we merge the update first and validate the result. Note that app normalization stays there as before.

Jira issues: MARATHON-8498

(cherry picked from commit 80c70dbc3d3c37aa23fdb2b555ee418f8cf06362)

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
@@ -6,7 +6,7 @@ import mesosphere.marathon.api.{Rejection, RejectionException}
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.appinfo.{AppSelector, Selector}
-import mesosphere.marathon.plugin.auth.{AuthorizedAction, Authorizer, CreateRunSpec, Identity, UpdateRunSpec, ViewRunSpec}
+import mesosphere.marathon.plugin.auth._
 import mesosphere.marathon.state.VersionInfo.OnlyVersion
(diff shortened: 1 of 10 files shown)


src/main/scala/mesosphere/marathon/api/v2/AppHelpers.scala
```diff
@@ -6,7 +6,7 @@ import mesosphere.marathon.api.{Rejection, RejectionException}
 import mesosphere.marathon.api.v2.Validation._
 import mesosphere.marathon.api.v2.validation.AppValidation
 import mesosphere.marathon.core.appinfo.{AppSelector, Selector}
-import mesosphere.marathon.plugin.auth.{AuthorizedAction, Authorizer, CreateRunSpec, Identity, UpdateRunSpec, ViewRunSpec}
+import mesosphere.marathon.plugin.auth._
```
zen-dog: People let this one through in the original PR 🙈

timcharper: @zen-dog I commented on it and it was merged anyways :( 

https://github.com/mesosphere/marathon/pull/6651/files#r232423958

timcharper: Oh well

meln1k: hmm, that's not cool. I remember changing that. I'll check my intellij settings, maybe it was automatically optimised.

PR #6671, opened 2018-11-21
Fix for a secrets validator error when we change env without changing…

meln1k:
… secrets (#6651)

Summary: Secrets validator was only passing if we submit a secret environment variable update with a corresponding secret definition section. Because of this, the app update validation process is simplified now: instead of validating an app update and the resulting app definition, we merge the update first and validate the result. Note that app normalization stays there as before.

Jira issues: MARATHON-8498

(cherry picked from commit 80c70dbc3d3c37aa23fdb2b555ee418f8cf06362)

### DIFF
--- a/src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
+++ b/src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
@@ -45,7 +45,7 @@ class AppsResource @Inject() (
 
   private[this] val ListApps = """^((?:.+/)|)\*$""".r
   private implicit lazy val appDefinitionValidator = AppDefinition.validAppDefinition(config.availableFeatures)(pluginManager)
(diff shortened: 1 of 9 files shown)


src/main/scala/mesosphere/marathon/api/v2/AppsResource.scala
```diff
@@ -45,7 +45,7 @@ class AppsResource @Inject() (
 
   private[this] val ListApps = """^((?:.+/)|)\*$""".r
   private implicit lazy val appDefinitionValidator = AppDefinition.validAppDefinition(config.availableFeatures)(pluginManager)
-  private implicit lazy val validateCanonicalAppUpdateAPI = AppValidation.validateCanonicalAppUpdateAPI(config.availableFeatures, () => normalizationConfig.defaultNetworkName)
+  private implicit lazy val validateCanonicalAppUpdateAPI = AppValidation.validateAppUpdateVersion
```
zen-dog: just use `validateAppUpdateVersion` directly

PR #6674, opened 2018-11-22
Rename resident task id to TaskIdWithIncarnation

alenkacz:
This will be used in more general sense while we come to stable instance ids.

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -292,7 +292,7 @@ class InstanceOpFactoryImpl(
         }(collection.breakOut)
 
         val containerNameToTaskId: Map[String, Task.Id] = oldToNewTaskIds.values.map {
-          case taskId @ Task.ResidentTaskId(_, Some(containerName), _) => containerName -> taskId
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -292,7 +292,7 @@ class InstanceOpFactoryImpl(
         }(collection.breakOut)
 
         val containerNameToTaskId: Map[String, Task.Id] = oldToNewTaskIds.values.map {
-          case taskId @ Task.ResidentTaskId(_, Some(containerName), _) => containerName -> taskId
+          case taskId @ Task.TaskIdWithIncarnation(_, Some(containerName), _) => containerName -> taskId
```
jeschkies: I'd also introduce it for ephemeral instances as well

```scala
val taskId = scheduledInstance.tasksMap.headOption match {
  case Some((taskId, _)) => Task.Id.withIncarnationCount(taskId)
  case None => Task.Id.forInstanceId(scheduledInstance.instanceId)
}
```

This way #6616 would not have to introduce that change.

jeschkies: Also, what do you think about

```scala
val taskId = scheduledInstance.tasksMap.headOption match {
  case Some((taskId, _)) => taskId.withIncreasedIncarnation()
  case None => Task.Id.forInstanceId(scheduledInstance.instanceId)
}
```

PR #6675, opened 2018-11-23
Remove unreachable strategy from state instance.

jeschkies:
Summary:
With #6661 we are attaching the run spec to each instance. Thus we do
not have to persist the unreachable strategy which is part of the run
spec. This change will make updates straight from Marathon < 1.6 impossible.

JIRA issues: MARATHON-8325


src/main/scala/mesosphere/marathon/storage/migration/Migration.scala
```diff
@@ -213,9 +213,6 @@ object Migration {
       StorageVersions(1, 5, 0, StorageVersion.StorageFormat.PERSISTENCE_STORE) -> { (migration) =>
         MigrationTo15(migration)
       },
-      StorageVersions(1, 5, 2, StorageVersion.StorageFormat.PERSISTENCE_STORE) -> { (migration) =>
-        new MigrationTo152(migration.instanceRepo)
-      },
```
jeschkies: I was unsure about this step. If one migrates through to 1.8 it is ok to drop this migration. However, if the migration fails after 1.6 and the user switches to the 1.6 binary we have a problem. What about dropping all migrations up to 1.6 and enforce that Marathon 1.8 can be only upgraded from 1.6 onwards?

meln1k: I think that by removing existing migrations you're going to open the gates of hell. I'd not do it, instead I'd fix the migration.

jeschkies: If we say we support only upgrades from 1.6 upwards it's fine. I'd remove all migrations below 1.7 then. @meichstedt what is our support here?

changelog.md
```diff
@@ -1,5 +1,10 @@
 ## Changes to 1.8.xxx
 
+### Upgrades only from Marathon 1.6+
```
timcharper: Feels like we ought to have a doc page on upgrading

changelog.md
```diff
@@ -1,5 +1,10 @@
...
+
+You can only upgrade to Marathon 1.8 from 1.6.x and 1.7.x. If you'd like to upgrade from an earlier version you should
```
zen-dog: Did we ever properly describe Marathon upgrade strategy? Feels like we should spend a few more words on it.

jeschkies: You mean the process of upgrading or the strategy? We have [dcos](https://mesosphere.github.io/marathon/docs/upgrade/index.html). There are a little old but should still apply.

src/main/scala/mesosphere/marathon/storage/migration/MigrationTo17.scala
```diff
@@ -107,9 +106,7 @@ object MigrationTo17 extends MaybeStore with StrictLogging {
       (__ \ "unreachableStrategy").readNullable[raml.UnreachableStrategy] ~
```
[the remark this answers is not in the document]
jeschkies: Hm, this will alter the migration between 1.7 and 1.8 but I guess it's fine.

tests/integration/src/test/scala/mesosphere/marathon/integration/UpgradeIntegrationTest.scala
```diff
@@ -234,33 +234,33 @@ class UpgradeIntegrationTest extends AkkaIntegrationTest with MesosClusterTest w
     }
   }
 
-  "upgrade from 1.4.9 to the latest" in {
+  "upgrade from 1.6.322 to the latest" in {
```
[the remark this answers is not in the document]
jeschkies: No sure what you mean. We cache all Marathon binaries.

tests/integration/src/test/scala/mesosphere/marathon/integration/UpgradeIntegrationTest.scala
```diff
@@ -234,33 +234,33 @@ class UpgradeIntegrationTest extends AkkaIntegrationTest with MesosClusterTest w
...
     val zkUrl = s"$zkURLBase-to-latest"
-    val marathon149 = Marathon149(marathon149Artifact.marathonPackage, suiteName = s"$suiteName-1-4-9", mesosMasterUrl, zkUrl)
+    val marathon16322 = Marathon16322(marathon16322Artifact.marathonPackage, suiteName = s"$suiteName-1-6-322", mesosMasterUrl, zkUrl)
```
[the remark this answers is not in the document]
jeschkies: I'd much rather keep this explicit.

PR #6676, opened 2018-11-23
Document events published by the API.

jeschkies:
JIRA issues: DCOS-45420


docs/docs/events.md
```diff
@@ -0,0 +1,66 @@
...
+
+```bash
+curl -H "Accept: text/event-stream"  <MARATHON_HOST>:<MARATHON_PORT>/v2/events?event_type=event_stream_detached\&event_type=event_stream_attached
+```
+
+All event types can be found in [Events.scala](https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/event/Events.scala).
+
+## Which related concepts should I understand?
+
+The published events cover [instance](???), [app](???) and [pod](...) changes as well as [deployments](???) and [health checks](???).
```
timcharper: Perhaps this should be labeled as WIP until these are fixed? :)

jeschkies: Well, the links don't exist right. This is just the docathon we did.

alenkacz: maybe we should drop the links altogether for now? Because I think if we merge this, it would be released with the next marathon release as a docs change

jeschkies: We won't merge it. As I understood @meichstedt we should just draft them but not merge them.

meichstedt:
@jeschkies can you move the file to e.g. `docs/docs/wip` or some other subfolder? that way it's clearer that this is part of the new docs and won't get mistaken.

We could also create a long-lived branch that we merge these into until we have a critical mass.

@timcharper @alenkacz I'd prefer having a clear indication where links are missing, instead of not having links at all. Until we have most of the articles, the majority of them will have missing links. I think having these docs in a clearly wip folder, or a remote branch, will help make that transparent.

jeschkies:
Ping @alenkacz and @timcharper.

jeschkies:
@alenkacz and @timcharper, are you still opposed? How should we fix the links?

jeschkies:
Won't do for now.

PR #6679, opened 2018-11-27
Remove API heavy events from Marathon repository

takirala:
Remove API heavy events from Marathon repository

Summary:
- Removed all the source code that was supporting the heavy events API.
- Rewired unit and integration tests to always assume the "light" events API. There is no mention of "light" in the codebase now as it is the (only) default. 
- Update `yaml`s to reflect `plan-format` and the default serialization format.

JIRA issues: [MARATHON-8458](https://jira.mesosphere.com/browse/MARATHON-8458)


src/main/scala/mesosphere/marathon/core/event/Events.scala
```diff
@@ -11,19 +9,11 @@ import mesosphere.marathon.core.instance.Instance
...
 
 sealed trait MarathonEvent {
   val eventType: String
   val timestamp: String
-
-  @JsonIgnore
-  lazy val fullJsonString: String = Json.stringify(eventToJson(this, false))
-
-  @JsonIgnore
-  lazy val lightJsonString: String = Json.stringify(eventToJson(this, true))
```
jeschkies: I think we should keep this for caching purposes. We send events to several consumers and do not want to serialize it each time.

src/main/scala/mesosphere/marathon/core/event/impl/stream/HttpEventStreamServlet.scala
```diff
@@ -71,7 +66,7 @@ class HttpEventSSEHandle(
 
   override def sendEvent(event: MarathonEvent): Unit = {
     if (subscribed(event.eventType)) {
-      val payload = if (useLightWeightEvents) event.lightJsonString else event.fullJsonString
+      val payload = Json.stringify(eventToJson(event))
```
jeschkies: I think this should be `event.jsonString` so that we do not serialize each time.

meln1k:
Changelog entry with a short description of that breaking change would be awesome!

alenkacz:
> Changelog entry with a short description of that breaking change would be awesome!

I think this is not breaking change per say but I guess we should write to changelog that people cannot use that feature flag anymore as marathon would fail with that feature flag present. Good point!

src/main/scala/mesosphere/marathon/core/event/Events.scala
alenkacz: I think it's still there, just renamed, so we should be fine

takirala: Actually I removed this in my initial commit(s). Brought it back after @jeschkies pointed out that we need it. Git diff wasn't smart enough to mark this comment as "Outdated" even after I pushed my commits.

docs/docs/rest-api/public/api/v2/events.raml
```diff
@@ -16,11 +16,6 @@ get:
         description:
             Specify subscribed event types.
             You can specify this parameter multiple times with different values.
-    plan-format:
```
timcharper: 🎉

docs/docs/rest-api/public/api/v2/types/deploymentPlan.raml
```diff
@@ -10,18 +10,10 @@ types:
     (pragma.serializeOnly): # Used only for serialization
     description: |
        A deployment is a change in the service setup.
-       Note that the `orignal` and `target` properties could be missing when the plan
```
timcharper: 🎉

docs/docs/rest-api/public/api/v2/types/deploymentPlan.raml
```diff
@@ -10,18 +10,10 @@ types:
...
-       is generated in "lightweight" mode.
     properties:
       id:
         type: string
         description: The UUID of the deployment plan
-      original?:
```
timcharper: 🎉

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -273,26 +255,27 @@ trait EventFormats {
...
   implicit lazy val GroupChangeFailedWrites: Writes[GroupChangeFailed] = Json.writes[GroupChangeFailed]
 
-  implicit lazy val DeploymentSuccessWrites: Writes[DeploymentSuccess] = Json.writes[DeploymentSuccess]
-  implicit lazy val DeploymentFailedWrites: Writes[DeploymentFailed] = Json.writes[DeploymentFailed]
-  implicit lazy val DeploymentStatusWrites: Writes[DeploymentStatus] = Json.writes[DeploymentStatus]
-  implicit lazy val DeploymentStepSuccessWrites: Writes[DeploymentStepSuccess] = Json.writes[DeploymentStepSuccess]
-  implicit lazy val DeploymentStepFailureWrites: Writes[DeploymentStepFailure] = Json.writes[DeploymentStepFailure]
-
-  implicit lazy val LightDeploymentSuccessWrites: Writes[DeploymentSuccess] = Writes { event =>
+  implicit lazy val DeploymentPlanWrites: Writes[DeploymentPlan] = Writes { plan =>
```
timcharper: 🎉

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -273,26 +255,27 @@ trait EventFormats {
...
+      "id" -> plan.id,
+      "steps" -> plan.steps,
+      "version" -> plan.version
+    )
+  }
+  implicit lazy val DeploymentSuccessWrites: Writes[DeploymentSuccess] = Writes { event =>
     Json.obj(
       "id" -> event.id,
-      "plan" -> LightDeploymentPlanWrites.writes(event.plan),
+      "plan" -> DeploymentPlanWrites.writes(event.plan),
```
timcharper: Now that there's just one implicit deployment plan writes mechanism, can we just remove DeploymentPlanWrites.writes and use the implicit?

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -273,26 +255,27 @@ trait EventFormats {
...
-      "timestamp" -> Timestamp.now().toString
+      "timestamp" -> event.timestamp
     )
   }
-  implicit lazy val LightDeploymentFailedWrites: Writes[DeploymentFailed] = Writes { event =>
+  implicit lazy val DeploymentFailedWrites: Writes[DeploymentFailed] = Writes { event =>
     val serializedEvent = Json.obj(
       "id" -> event.id,
-      "plan" -> LightDeploymentPlanWrites.writes(event.plan),
+      "plan" -> DeploymentPlanWrites.writes(event.plan),
```
timcharper: ditto; we should just use the implicit

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -302,28 +285,28 @@ trait EventFormats {
         serializedEvent
     }
   }
-  implicit lazy val LightDeploymentStatusWrites: Writes[DeploymentStatus] = Writes { event =>
+  implicit lazy val DeploymentStatusWrites: Writes[DeploymentStatus] = Writes { event =>
     Json.obj(
-      "plan" -> LightDeploymentPlanWrites.writes(event.plan),
+      "plan" -> DeploymentPlanWrites.writes(event.plan),
```
timcharper: implicit

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -302,28 +285,28 @@ trait EventFormats {
...
       "eventType" -> "deployment_info",
-      "timestamp" -> Timestamp.now().toString
+      "timestamp" -> event.timestamp
     )
   }
-  implicit lazy val LightDeploymentStepSuccessWrites: Writes[DeploymentStepSuccess] = Writes { event =>
+  implicit lazy val DeploymentStepSuccessWrites: Writes[DeploymentStepSuccess] = Writes { event =>
     Json.obj(
-      "plan" -> LightDeploymentPlanWrites.writes(event.plan),
+      "plan" -> DeploymentPlanWrites.writes(event.plan),
```
timcharper: implicit

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -302,28 +285,28 @@ trait EventFormats {
...
       "eventType" -> "deployment_step_success",
-      "timestamp" -> Timestamp.now().toString
+      "timestamp" -> event.timestamp
     )
   }
-  implicit lazy val LightDeploymentStepFailureWrites: Writes[DeploymentStepFailure] = Writes { event =>
+  implicit lazy val DeploymentStepFailureWrites: Writes[DeploymentStepFailure] = Writes { event =>
     Json.obj(
-      "plan" -> LightDeploymentPlanWrites.writes(event.plan),
+      "plan" -> DeploymentPlanWrites.writes(event.plan),
```
timcharper: implicit

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
```diff
@@ -368,7 +351,7 @@ trait EventFormats {
     )
   }
 
-  def eventToJson(event: MarathonEvent, lightweightPlan: Boolean): JsValue = event match {
```
timcharper: 🎉

src/main/scala/mesosphere/marathon/api/v2/json/Formats.scala
takirala: Good catch Tim!

I followed your suggestion and when I was cleaning up, i found that we no longer need the custom writers and thus replaced all the custom writers with the default `Json.writes[T]`. I made sure that all the "keys" in json are identical to the field names in the case classes. But lmk if this sort of implicit assumption (so that nothing breaks if someone decides to change the case class field names) is not ideal and I can revert. I see that we are doing the same thing for other case classes (like `GroupChangeFailed` for example) so I did the same for all other subclasses of `UpgradeStep`.

Also, sorry to make more commits after approval(s).

PR #6680, opened 2018-11-27
Tharper/kill instances via goal

timcharper:
<<Replace this line with your Revision Title>>

Summary:

JIRA issues:


src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -69,7 +79,28 @@ class TaskKiller @Inject() (
     }
   }
 
-  private[this] def expunge(instances: Seq[Instance]): Future[Done] = {
+  private[this] def awaitConsideredTerminal(instanceIds: Set[Instance.Id]): Source[Set[Instance.Id], NotUsed] =
+    instanceTracker.instanceUpdates
+      .flatMapConcat {
+        case (snapshot, updates) =>
+          val pendingInstances: Set[Instance.Id] = snapshot.instances.collect {
```
jeschkies: Hm, this is tricky. What about terminal instances with `Goal.Running`?

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -69,7 +79,28 @@ class TaskKiller @Inject() (
...
+            .filter { change => instanceIds(change.id) }
+            .scan(pendingInstances) {
+              case (remaining, deleted: InstanceDeleted) =>
+                remaining - deleted.id
+              case (remaining, InstanceUpdated(instance, _, _)) if considerTerminal(instance.state.condition) =>
+                remaining - instance.instanceId
+              case (remaining, _) =>
+                remaining
+            }
+            .takeWhile(_.nonEmpty)
```
jeschkies: This is way easier to understand than the kill watcher 🙂.

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
timcharper: That's a great catch; those instances will be restarted and we will need to wait for the goal and the status to be terminal

src/main/scala/mesosphere/marathon/api/TaskKiller.scala
```diff
@@ -43,22 +47,28 @@ class TaskKiller @Inject() (
           val activeInstances = foundInstances.filter(_.isActive)
 
           if (wipe) {
-            val instancesAreTerminal: Future[Done] = killService.watchForKilledInstances(activeInstances)
+            val foundInstanceIds = Set(foundInstances.map(_.instanceId): _*)
```
meln1k: Just wonder, why not 
```suggestion
            val foundInstanceIds = foundInstances.map(_.instanceId).toSet?
```

PR #6677, opened 2018-11-24
Stable instance ids

alenkacz:
Summary:
Until now, Marathon expunged instances once all of their tasks ended up in a terminal state, and eventually launched replacements as a result of those instances being expunged.
This commit adjusts this behavior. Instances will now only be expunged from the state when their goal is to be `Decommissioned` and all their tasks are in a terminal state. If their goal is still `Running`, they will be considered to be scheduled and used to launch replacement tasks.

This means that instance Ids are now stable for as long as an instance shall be kept running. New instances will be created only when replacing unreachable instances, and when replacing instances with new versions. Similar to the way we handle task Ids for stateful services, tasks of stateless services will now also provide an incarnation count, appended to the task Id. The first task created for an instance will be the .1, and subsequent replacements will increment that incarnation counter.
When killing an instance using the `wipe=true` flag, its goal will be set to decommissioned, and it will eventually be expunged when all tasks are terminal. Note that as long as its tasks are e.g. unreachable, it will not be expunged until they are reported terminal (`GONE, GONE_BY_OPERATOR, or UNKNOWN`).

When killing instances without the `wipe=true` or `scale=true` flags, Marathon will only issue kill requests to Mesos, but keep the current goal and will, therefore, launch replacements that are still associated with the existing instance.

JIRA Issues: MARATHON-8373, MARATHON-8140

Co-authored-by: Matthias Eichstedt <matthias.eichstedt@mesosphere.io>
Co-authored-by: Tim Harper <timcharper@gmail.com>
Co-authored-by: Karsten Jeschkies <k@jeschkies.xzy>

PR #6076, opened 2018-03-09
Make instance and task ids case classes.

jeschkies:
Summary:
We have different types of instance and task ids. Yet, all are modeled as strings.
Information is extracted with regular expressions ad-hoc. This patch introduces
different type for different ids and has one method that transforms strings to ids. 

JIRA issues: MARATHON-8139, MARATHON-8142

PR #6681, opened 2018-11-27
Increase incarnation for pod tasks as well.

jeschkies:

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -11,7 +11,6 @@ import mesosphere.marathon.core.launcher.{InstanceOp, InstanceOpFactory, OfferMa
 import mesosphere.marathon.core.plugin.PluginManager
 import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.core.task.Task
-import mesosphere.marathon.core.task.Task.TaskIdWithIncarnation
 import mesosphere.marathon.core.task.state.NetworkInfo
 import mesosphere.marathon.metrics.Metrics
 import mesosphere.marathon.plugin.scheduler.SchedulerPlugin
@@ -89,8 +88,10 @@ class InstanceOpFactoryImpl(
     matchedOffer match {
       case matches: ResourceMatchResponse.Match =>
         val instanceId = scheduledInstance.instanceId
-        val taskIds = pod.containers.map { container =>
-          Task.Id.forInstanceId(instanceId, Some(container))
+        val taskIds = if (scheduledInstance.tasksMap.isEmpty) {
+          pod.containers.map { container => Task.Id.forInstanceId(instanceId, Some(container)) }
+        } else {
+          scheduledInstance.tasksMap.keysIterator.map(Task.Id.forExistingTaskId).to[Seq]
         }
         val (executorInfo, groupInfo, hostPorts) = TaskGroupBuilder.build(pod, offer,
           instanceId, taskIds, builderConfig, runSpecTaskProc, matches.resourceMatch, None)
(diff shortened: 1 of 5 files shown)


src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -367,23 +368,32 @@ object Task {
...
-      * Create a taskId for a pod instance's task. This will create a taskId designating the instance and each
-      * task container's name. It may be used for reservations for persitent pods as well.
+      * Create a taskId for an app or pod instance's reservation. This will create a taskId designating the instance and
+      * each task container's name.
       *
       * @param instanceId the ID of the instance that this task is contained in
       * @param container the name of the task as per the pod container config.
       */
-    def forInstanceId(instanceId: Instance.Id, container: Option[MesosContainer] = None): Id = EphemeralOrReservedTaskId(instanceId, container.map(_.name))
+    def forReservation(instanceId: Instance.Id, container: Option[MesosContainer] = None): Id = EphemeralOrReservedTaskId(instanceId, container.map(_.name))
```
meln1k: The method name says `forReservation` however we return `EphemeralOrReservedTaskId`. Do you think we should make it more concise?

jeschkies: Unfortunately we cannot name the type just `ReservedTaskId` since old ephemeral task uses the same id pattern. We could rename `forReservation` but it should only be used for reservations so that does not make much sense to me either. I'm happy for any suggestion.

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -343,13 +341,13 @@ class InstanceOpFactoryImpl(
         // will be replaced with a new task once we launch on an existing reservation this way, the reservation will be
         // labeled with a taskId that does not relate to a task existing in Mesos (previously, Marathon reused taskIds so
         // there was always a 1:1 correlation from reservation to taskId)
-        val reservationLabels = TaskLabels.labelsForTask(frameworkId, Task.Id.forInstanceId(scheduledInstance.instanceId))
+        val reservationLabels = TaskLabels.labelsForTask(frameworkId, Task.Id.forReservation(scheduledInstance.instanceId))
```
timcharper: Can you explain the motivation for this change?

src/main/scala/mesosphere/marathon/core/task/Task.scala
timcharper: Is there a reason why we need to have reserved / non-reserved semantics in our task ids? Ideally, I think we'd just have task-ids, since both have incarnations.

alenkacz: I ➕ @meln1k confusion, I don't get this part as well :( Maybe we can clean this up a bit more... We can chat tomorrow

jeschkies: Ok, I understand the confusion. The issue is the ids for reservations are the same as the *old* ephemeral task ids. When the ids string of a reservation or a task is parsed there is no way to tell whether it should be `ReservationId` or `EphemerlId`. Hence `EphemeralOrReservationId`. Unfortunately this means we have to use the later when we create reservation ids. I can try a follow up and really separate them but this is already much cleaner than on master since we at least call `forReservation` in the right places. This will make a follow up easier. If you want to know what I went through take a look at #6076.

That aside I'm surprised you all are surprised 🙂. Reservation ids and ephemeral ids will be the same until #6677 lands. With #6677 we simply get a new type for ephemeral task ids. Also, keep in mind reservation ids are *not* resident task ids! Ideally reservation ids would be complete separate but so far we cannot do that.

jeschkies:
> why do we have reservationId on Task.Id at all? Even for ephemeral? I think the only usage is for task labels, we can use idString directly there maybe?

Well, we have the `EphemeralOrReservationId` type for historical reasons. I did not manage to separate them. Ideally we would introduce a `ReservationId` type that is completely indepenent of task ids. However, that is not easy.

> why is fromReservation special? Why can't we use forInstanceId there? It seems to be doing the same thing...

`forReservation` returns the id for a reservation *not* for a resident task. It has nothing to do with ids for launched tasks.

jeschkies:
@timcharper, @alenkacz and @meln1k I created a new `Reservation.Id` type in #6687. I hope that clears up a few things. Please not that #6687 has some outstanding todos. I really would like to merge this pull request before.

PR #6683, opened 2018-11-27
default value of root for --mesos_user in Marathon Universe template

kensipe:
default value of root for --mesos_user in Marathon Universe template

Summary:
update based on universe creator
many changes based on the fact that we haven't used our automated tool in some time.

JIRA issues:  DCOS-43109


kensipe:
There are a number of changes  for instance... `decline-offer-duration` changed from integer to a long
enabled features has "vips and task_killing" removed... I'm thinking it shouldn't.   We need to capture all issues and update the generator.

PR #6685, opened 2018-11-28
Fix the log message for delay from an old runspec

alenkacz:
I saw this in our logs

```
[39mDEBUG[0;39m[16:06:38 AppDeployIntegrationTest-LocalMarathon-32857] [31mWARN [0;39m[16:06:38 TaskLauncherActor] BUG! Received delay update for other run spec RunSpecConfigRef(/app-with-update-test,2018-11-27T16:06:36.421Z) and delay None. Current run spec is RunSpecConfigRef(/app-with-update-test,2018-11-27T16:06:37.498Z)
```

Basically I am in the middle of replacing one version for another versions when the instance of old version failed thus I tried to increase delay. That can happen so it should not log "BUG!"

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
+++ b/src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
@@ -201,7 +201,7 @@ private class TaskLauncherActor(
       logger.debug(s"After delay update $status")
 
     case msg @ RateLimiter.DelayUpdate(ref, delayUntil) if ref != runSpec.configRef =>
-      logger.warn(s"BUG! Received delay update for other run spec ${ref} and delay $delayUntil. Current run spec is ${runSpec.configRef}")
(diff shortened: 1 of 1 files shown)


src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -201,7 +201,7 @@ private class TaskLauncherActor(
       logger.debug(s"After delay update $status")
 
     case msg @ RateLimiter.DelayUpdate(ref, delayUntil) if ref != runSpec.configRef =>
-      logger.warn(s"BUG! Received delay update for other run spec ${ref} and delay $delayUntil. Current run spec is ${runSpec.configRef}")
+      logger.warn(s"Ignoring delay update for old run spec $ref and delay $delayUntil. Current run spec is ${runSpec.configRef}")
```
jeschkies: I thought it *is* a bug 😉. In any case this should go with #6664.

alenkacz: hmm no, when we are in the middle of replace, I think it's expected behavior :)

PR #6687, opened 2018-11-28
Introduce reservation id type.

jeschkies:
Summary:
The new reservation id type makes the difference between a task id,
instance id and reservation id clear. So far the are very similar and it
is trivial to construct one from the other. However, in the near future
we would like to reassign a reservation to a different instance. That
means reservation and instance ids have to be independent. This is the
first step.

Depends on #6681.


src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryHelper.scala
```diff
@@ -18,10 +17,9 @@ class InstanceOpFactoryHelper(
 
   def provision(
     taskInfo: Mesos.TaskInfo,
-    newTask: Task,
     instance: Instance): InstanceOp.LaunchTask = {
 
-    assume(newTask.taskId.mesosTaskId == taskInfo.getTaskId, "marathon task id and mesos task id must be equal")
+    assume(instance.tasksMap.keys.forall(_.mesosTaskId == taskInfo.getTaskId), "marathon task id and Mesos task id must be equal")
```
alenkacz: why not `keysIterator`?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -175,8 +175,9 @@ class InstanceOpFactoryImpl(
 
       // resources are reserved for this role, so we only consider those resources
       val rolesToConsider = config.mesosRole.toOption.toSet
-      val taskId = Task.Id.forInstanceId(volumeMatch.instance.instanceId)
-      val reservationLabels = TaskLabels.labelsForTask(request.frameworkId, taskId).labels
+      //TODO(karsten): This was and is not quite right. We should pass the instance id to the resource matcher instead
```
alenkacz: hmm, not sure I get it. Can you elaborate more?

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
```diff
@@ -25,16 +24,17 @@ object TaskLabels {
...
   }
 
-  def labelsForTask(frameworkId: FrameworkId, taskId: Task.Id): ReservationLabels = {
+  def labelsForTask(frameworkId: FrameworkId, reservationId: Reservation.Id): ReservationLabels = {
     ReservationLabels(Map(
       FRAMEWORK_ID_LABEL -> frameworkId.id,
       // This uses taskId.reservationId to match against the id that was originally used to create the reservation
-      TASK_ID_LABEL -> taskId.reservationId
+      //TODO(karsten): We probably want to call it RESERVATION_ID_LABEL in the future.
+      TASK_ID_LABEL -> reservationId.label
```
alenkacz: well you can at least rename the property 🤷‍♀️ and put this note in there...

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -223,7 +223,7 @@ object Task {
 
     override lazy val instanceId: Instance.Id = Instance.Id(runSpecId, Instance.PrefixMarathon, uuid)
 
-    override val reservationId = idString
+    override val reservationId = Reservation.LegacyId(runSpecId, separator, uuid)
```
alenkacz: I still don't quite understand why reservation id must be a property on task id... It just doesn't make sense to me especially since most of the tasks are not even resident... can we just drop this and call Reservation.Id directly?

src/main/scala/mesosphere/marathon/core/launcher/impl/TaskLabels.scala
jeschkies: I don't think I can just rename the property. It belongs to https://jira.mesosphere.com/browse/MARATHON-8517. I'll note it in the JIRA.

src/main/scala/mesosphere/marathon/core/task/Task.scala
jeschkies: > It just doesn't make sense to me especially since most of the tasks are not even resident

You are correct. This will go away with MARATHON-8517. I did not want to change too much in this diff.

PR #6689, opened 2018-11-28
Use NonEmptyIterable instead of headOption.

jeschkies:
Summary:

JIRA issues:

### DIFF
--- a/src/main/scala/mesosphere/marathon/NonEmptyIterable.scala
+++ b/src/main/scala/mesosphere/marathon/NonEmptyIterable.scala
@@ -1,0 +1,17 @@
+package mesosphere.marathon
+
+object NonEmptyIterable {
+
+  def apply[A](head: A, tail: Iterable[A] = Iterable.empty[A]): NonEmptyIterable[A] = new NonEmptyIterable(head, tail)
+
+  def unapply[A, I <: Iterable[A]](iter: I with Iterable[A]): Option[NonEmptyIterable[A]] = {
+    if (iter.nonEmpty) Some(NonEmptyIterable(iter.head, iter.tail))
+    else None
+  }
+}
+
+case class NonEmptyIterable[A](override val head: A, override val tail: Iterable[A]) extends Iterable[A] {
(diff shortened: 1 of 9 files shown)


src/main/scala/mesosphere/marathon/package.scala
```diff
@@ -29,6 +32,22 @@ package object marathon {
...
+    def unapply[A, I <: Iterable[A]](iter: I with Iterable[A]): Option[NonEmptyIterable[A]] = {
+      if (iter.nonEmpty) Some(NonEmptyIterable(iter.head, iter.tail))
+      else None
+    }
+  }
+
+  class NonEmptyIterable[A](head: A, tail: Iterable[A]) extends Iterable[A] {
+    override def iterator: Iterator[A] = Iterator.apply(head) ++ tail.iterator
+
+    override def headOption: Option[A] = throw new NoSuchMethodException("Please use NonEmptyIterable.head instead.")
```
jeschkies: I'd love this error to happen at compilation time. Otherwise I'll make it a `Some(head)`.

src/main/scala/mesosphere/marathon/NonEmptyIterable.scala
```diff
@@ -0,0 +1,17 @@
...
+  def unapply[A, I <: Iterable[A]](iter: I with Iterable[A]): Option[NonEmptyIterable[A]] = {
+    if (iter.nonEmpty) Some(NonEmptyIterable(iter.head, iter.tail))
+    else None
+  }
+}
+
+case class NonEmptyIterable[A](override val head: A, override val tail: Iterable[A]) extends Iterable[A] {
+  override def iterator: Iterator[A] = Iterator.apply(head) ++ tail.iterator
+
+  override def headOption: Option[A] = Some(head)
```
zen-dog: On the other site, `head` should be used - otherwise users will either endup treating a non-existing match case, or `get`-ing an `Option` and both doesn't make much sense for a `NonEmptyIterable`. On the other side it's simple and you don't have to throw an exception...

PR #6690, opened 2018-11-28
Added support for multiroles

ccaballe:
Summary:
Currently, Marathon can only be registered with one role using the parameter --mesos_role when startup. This PR allows Marathon to register with multiple roles using the new parameter --mesos_roles with a list of roles separated by commas.

JIRA issues: MARATHON-2775


zen-dog:
@ccaballe could you please rebase your branch on _master_? Overall, we're thankful for your contribution and are discussing internally what would be the best way to proceed with this feature, what are the possible edge cases etc. We'll get back to you asap.

ccaballe:
@zen-dog Done. Thanks for your comment. I'm waiting for your feedback to adapt the functionality in the best way.

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -70,8 +70,9 @@ class OfferOperationFactory(
           .setLabels(reservationLabels.mesosLabels)
           .setPrincipal(principal)
 
+        // TODO @ccaballe volume reservation with the first role
```
timcharper: let's get rid of this TODO?

src/main/scala/mesosphere/marathon/core/matcher/base/util/OfferOperationFactory.scala
```diff
@@ -70,8 +70,9 @@ class OfferOperationFactory(
...
         Mesos.Resource.newBuilder(resource)
-          .setRole(role)
+          .setRole(role(0))
```
timcharper: This is fine for now; let's be sure and document this behavior clearly in a changelog entry describing this change (in this PR, please :) ). You can search the changelog for other deprecation notices we've included in the past.

We can follow up a design session and a change to specify the role that should be used for reservation. This will be a mechanism that will also work for reservation refinements.

timcharper:
Also, @ccaballe - can you describe more your use case? Is this because you have multiple quotas that you want your Marathon instance to use?

src/main/scala/mesosphere/marathon/MarathonConf.scala
```diff
@@ -124,15 +124,15 @@ trait MarathonConf
 
   def executor: Executor = Executor.dispatch(defaultExecutor())
 
-  lazy val mesosRole = opt[String](
```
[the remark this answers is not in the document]
timcharper: One thing we like to do is to code in the enforcement of our deprecation policy ahead of time. You might have a custom validator for the mesosRole opt parser which fails if a value is specified and `BuildInfo.version >= SemVer(1,9.0)`

IE see this for an example:

/Users/tim/src/m8e/marathon/src/main/scala/mesosphere/marathon/Features.scala

  32 |     if (BuildInfo.version < SemVer(1, 8, 0))
  33 |       b += MAINTENANCE_MODE -> "(on by default, has no effect) Decline offers from agents undergoing a maintenance window"
  34 | 
  35 |     b.result()
```

ccaballe:
@timcharper Thanks for the comments. Now we can use both `--mesos_role` and `--mesos_roles`. I have added a validator for `--mesos_role` in versions 1.9.x+ and the entry in the Changelog. 

About the use cases, there might be several. From multiple quotas to multi-tenant scenarios using only one Marathon framework registered. As an example, consider a cluster with three different roles: 

> role0, role1, role2

Now you can ensure that your Marathon tasks will use all these resources across the cluster running Marathon with `--mesos_roles role0,role1,role2`. 

Let me know if I have to do more changes.

ccaballe:
@timcharper @zen-dog Hi, I've just rebased my branch on master.

timcharper:
@ccaballe Hi! 👋 Thanks for the update and for explaining your use case. I've been doing a lot of thinking about how this feature fits in to the future of Marathon. Ultimately, our end goal is to support Mesos hierarchical roles feature (and reservation refinements), which will require dynamic role subscriptions.

I've written up a proposal document about how we can proceed forward, fitting multi-role and hierarchical role in to our existing v2 API.

https://docs.google.com/document/d/1A5n2dxEEDALhcZmSBOLIPPYRcMyk-9_UgtVLBCTbF04/edit#heading=h.3w97ms7xrsgx

After thinking deeply about this, I'm having reservations with introducing a new parameter and a default-role-by-convention; it doesn't fit in the long term future of where we want to go. I think the ideal next step we should take to support multi-role is to instead support dynamic role subscription. This is a little bit more difficult to do than what you've proposed to do, but it's better in the long-run as there's less stuff to deprecate and remove, and it introduces overall less cruft to our API.

How willing are you to update this PR to use dynamic roles? Complexity wise, It's doable, and I think it may be just slightly above the threshold which mere mortals would be willing to face. It'd require some pairing, deeper collaboration, and would be subjected to highly-scrutinized review from the Marathon team. And, it'll require some design that requires some intricate knowledge of existing code. I'd be happy to jump on a call and point out some areas where to start, or, write out a list of some steps to guide you.

timcharper:
We have plans to introduce multirole support for marathon soon. Closing this.

timcharper:
@ccaballe - The new multi-role support is working well and is currently in DC/OS 1.14 beta. You can download a build here:

https://downloads.mesosphere.io/marathon/builds/1.9.81-a246a440e/marathon-1.9.81-a246a440e.tgz

timcharper:
Docs here: https://github.com/mesosphere/marathon/blob/master/docs/docs/multirole.md

PR #6691, opened 2018-11-28
Remove JSON Schema Resources from 1.8

meln1k:
Summary: JSON Schema Resources validator and related code is removed.

JIRA issues: MARATHON-8457


src/test/scala/mesosphere/marathon/api/v2/json/AppDefinitionSchemaJSONTest.scala
```diff
@@ -1,225 +0,0 @@
...
-
-/**
-  * Tests that test that the given JSON is rejected by the JSON schema.
-  *
-  * Since the JSON is not representable by an AppDefinition,
-  * JSON is used directly.
-  */
-class AppDefinitionSchemaJSONTest extends UnitTest {
-
-  "AppDefinitionJsonSchema" should {
```
alenkacz: Did we verify that these test cases are covered somewhere else? In our current validation logic... I just don't want us to decrease coverage :)

timcharper: My vote is we just delete these tests. They're testing that our JSON schema validates things properly and have all along been totally divorced from our RAML specification.

timcharper:
@alenkacz it's not removing your rejection

@meln1k we need to fix the CI failures before this can be merged.

PR #6692, opened 2018-11-28
Add documentation for instances

meichstedt:
Add documentation for instances

JIRA issues: MARATHON-8519


docs/articles/instances.md
```diff
@@ -0,0 +1,26 @@
+# Instances
+
+## What is an Instance?
+Instances are the Marathon representation of [⇒Tasks](tasks.md) in Mesos. An Instance can exist even if there is currently no Task associated. Depending on the goal of a given Instance, Marathon will take actions to reach that goal: it might match offers in order to launch a Task or [⇒TaskGroup](taskGroups.md), or it might send kill requests to Mesos and eventually expunge the Instance from its state. See [⇒Instance Life Cycle](instanceLifeCycle.md) for more information.
```
zen-dog: Wdyt about a small paragraph about `Goal`s, which ones exist, what are the allowed transitions etc.?

docs/articles/instances.md
```diff
@@ -0,0 +1,26 @@
...
+
+## How can I use Instances?
+Instances are created based on [⇒Service](services.md) specifications which can be either an [⇒App](apps.md) or a [⇒Pod](pods.md). Instances are implicitly created according to their App or Pod definition, which contains an instances property that designates the required amount of instances.
+
+## Which related concepts should I understand?
+When upgrading a Service during a [⇒Deployment](deployments.md) from one version to another, Marathon can create more instances than specified in either Service definition. It will do so if allowed via the [⇒upgradeStrategy](upgradeStrategy.md) configuration.
+When an instance is running on an agent disconnected from the cluster, it will eventually be reported as unreachable. According to the Service’s [⇒unreachableStrategy](unreachableStrategy.md), Marathon can schedule a replacement, even though the unreachable instance might still be running. When an unreachable instance becomes becomes reachable again, Marathon will scale down accordingly to the required amount of instances. It will decide over which instance shall be killed based on the [⇒killSelection](killSelection.md).
+
+## How should I use Instances?
```
zen-dog: Do we need both sections `How can I use ***` and `How should I use ***`?

docs/articles/instances.md
```diff
@@ -0,0 +1,26 @@
...
+## Are there any limitations or things to consider?
+_(This section could point out known limitations wrt number of instances.)_
+
+## Links
+* [⇒Scaling Applications](apps-scaling.md)
+* [⇒Scaling Pods](pods-scaling.md)
+* [⇒Killing Instances](instance-operations.md#delete-instances)
+
+## Examples
+* [⇒Simple Application](examples/app-simple.json) definition configuration 3 instances
```
zen-dog: Two questions:
- should we keep examples at the end of the article? Naively I'd expect an example in the text, where I introduce the topic
- what are good instances examples? Is it about `instances: 3` filed in the definition? Or is it about some `/v2/pods/{id}::status` or `/apps?embed=tasks` output?

docs/articles/instances.md
zen-dog: Also, we should stay consistent with docs article naming e.g. (1) camel-case `instanceLifeCycle.md` vs (2) kebab-case e.g. `instance-life-cycle.md`. My vote goes to (2)

zen-dog: I think we should spend a few additional words on why this layer is meaningful e.g. something along the lines of:
```There are a few reasons why this additional layer of indirection between actual Mesos tasks and users is meaningful. First and foremost: an Instance can have one task (see [Apps](apps.md)), two or more tasks (see [Pods](pods.md)) or no tasks at all (if the task was killed). 
The connection between Intances and it's tasks is quickly visible in their names. An app with name `/my-service` will have an instance named `my-service.<UUID>` and a task `my-service.<UUID>.3` where `.3` means that this is the third incarnation of this instance (previous two were killed or failed for some reason)```

Comment (meichstedt) [reply docs/articles/instances.md]: 👍 I adjusted the links to what I generated for index.md – which is (2) :)

I'll add a sentence or two wrt goals; I'd then link to _Instance Life Cycle_ as I think goals could be explained in more depth there. WDYT?

File: docs/articles/instances.md (hunk 2 above)
Comment (meichstedt) [reply docs/articles/instances.md]: In the scope doc I had the rhetoric questions

* What is this concept? Why or when would I use it?
* How can I use it?

So more like "you'd use X when you want to do Y" vs "in order to interact with X, you can do a, b, c"

I'll see to change the header

File: docs/articles/instances.md (hunk 3 above)
Comment (meichstedt) [reply docs/articles/instances.md]: > should we keep examples at the end of the article?

I'm all for _linking_ examples in the text above. If we find any sort of collapsible boxes, or a second column etc, let's have examples in the text. The one thing I dislike is arbitrarily long code snippets in the docs, because they make articles very long and hard to parse IMO.

I'd like to distinguish between tutorial/recipe and documentation here: in a tutorial or recipe ("how to to xyz"), let's have all the steps and configs needed. In an article explaining what an instance is, I'd like to keep it short and easy to skim through.

A reason to link examples is also to have them as files we can use for tests.

> what are good instances examples?

I guess a spec that is as simple as possible, and only shows how instances are created through (currently) the instances field of a spec.

**Comment by zen-dog** [comment]
> However I think that we should eventually break out those (surely growing) sections into a lower level article.

I agree we can always split out big chunks later. It's more important now to have them.

File: docs/articles/instances.md [hunk 4]
```diff
@@ -0,0 +1,41 @@
+# Instances
+
+## What is an Instance?
+Instances are the Marathon representation of [⇒Tasks](tasks.md) or [Task Groups](task-groups.md) in Mesos. An instance can exist even if there is currently no Task associated. We introduced this abstraction in order to support Mesos TaskGroups, where an instance can contain more than one Task. Beyond support for Pods, this abstraction also allows us to represent scheduled instances, which do not yet have Mesos Tasks associated with them.
```
zen-dog:
```suggestion
Instances are the Marathon representation of [⇒Tasks](tasks.md) or [Task Groups](task-groups.md) in Mesos. An instance can exist even if there is currently no Task associated. We introduced this abstraction in order to support Mesos TaskGroups (called [⇒Pods](pods.md) in Marathon), where an instance can contain more than one Task. Beyond support for Pods, this abstraction also allows us to represent scheduled instances, which do not yet have Mesos Tasks associated with them. 
```

docs/articles/instances.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Which related concepts should I understand?
+When upgrading a Service during a [⇒Deployment](deployments.md) from one version to another, Marathon can create more instances than specified in either Service definition. It will do so if allowed via the [⇒upgradeStrategy](upgrade-strategy.md) configuration.
+When an instance is running on an agent disconnected from the cluster, it will eventually be reported as unreachable. According to the Service’s [⇒unreachableStrategy](unreachable-strategy.md), Marathon can schedule a replacement, even though the unreachable instance might still be running. When an unreachable instance becomes becomes reachable again, Marathon will scale down accordingly to the required amount of instances. It will decide over which instance shall be killed based on the [⇒killSelection](kill-selection.md).
+
+## Are there any limitations or things to consider?
+_(This section could point out known limitations wrt number of instances.)_
```
zen-dog: Do you want to address this section before the merge?

docs/articles/instances.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Instance Goals
+Depending on the goal of a given instance, Marathon will take actions to reach that goal: it might match offers in order to launch a Task or TaskGroup. In this case, these Tasks will eventually be associated with the instance. If an instance shall be stopped, Marathon will send kill requests to Mesos and eventually expunge the instance from its state.
+
+Currently, instance goals are only used internally, but are reflected in the API. Supported goals are
+* `Running` means the goal is to have a Mesos Task or TaskGroup associated with this instance, which should be running all the time.
+* `Stopped` means, all Mesos Tasks or TaskGroups associated with this instance shall be stopped. The instance and all its metadata will be retained. It is possible to change the goal to `Running` again.
+* `Decommissioned` means the Mesos Task or TaskGroup associated with this instance will be killed, and reservations and volumes will be destroyed. The instance cannot be revived, and once in a reportedly terminal state, it will be expunged from Marathon's state.
```
kensipe: I feel like there is value in adding a link to Goals.scala . https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/core/instance/Goal.scala
and there is a `Terminal` goal

docs/articles/instances.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Instances in the API
+The `v2/apps` and `v2/tasks` API endpoints do not provide any means of direct interaction with instances. For historical reasons, these endpoints list tasks, and only allow to interact with them. The `v2/pods` endpoint on the other hand was build with instances as an abstraction for Mesos tasks, and thus does only allow direct interaction with instances, not with tasks.
+
+### Instance Ids
+The relation between instances and their Tasks is reflected in their names. An app with name `/my-service` will result in instances named `my-service.<UUID>`. The first Task associated with this instance will be named `my-service.instance-a41beca1-fa49-11e8-93c4-269daf75ef3f.$anon.1` where `$anon` designates the container as an anonymous container. The `.1` reveals the Task is the first task incarnation of this instance. Accordingly, `my-service.instance-<UUID>.$anon.42` will tell you that this is task incarnation number 42. This means 41 previous tasks for this instance became terminal, either because they exited normally or abnormally, or were killed by Marathon during a scale or upgrade operation.
```
kensipe: older task id references... I'll update

docs/articles/instances.md
meichstedt: The link makes sense – however `Terminal` is only a trait, not a Goal on its own.

docs/articles/instances.md
```diff
@@ -0,0 +1,38 @@
...
+* `Running` means the goal is to have a Mesos Task or TaskGroup associated with this instance, which should be running in the `TASK_RUNNING` state.
+* `Stopped` means, all Mesos Tasks or TaskGroups associated with this instance shall be stopped. The instance and all its metadata will be retained. It is possible to change the goal to `Running` again.
+* `Decommissioned` means the Mesos Task or TaskGroup associated with this instance will be killed, and reservations and volumes will be destroyed. The instance cannot be revived, and once a terminal state is reported by Mesos, the instance will be expunged from Marathon's state.
+
+See [Instance Life Cycle](instance-lifecycle.md) for more information on instance life cycle and goals.
+
+### Instances in the API
+The `v2/apps` and `v2/tasks` API endpoints do not provide any means of direct interaction with instances. For historical reasons, these endpoints list apps or tasks respectively, and only allow interacts with them (not instances). The `v2/pods` endpoint on the other hand was build with instances as an abstraction for Mesos tasks, and thus only allows direct interaction with instances, not with tasks.
```
meichstedt: `s/interacts/interactions`?

docs/articles/instances.md
```diff
@@ -0,0 +1,38 @@
...
+## How can I use Instances?
+Instances are created based on [Service](services.md) specifications which can be either an [AppDefinition](applications.md) or a [PodDefinition](pods.md). Instances are implicitly created according to their App or Pod definition, which contains an `instances` property that designates the required amount of instances.
+
+## Which related concepts should I understand?
+When upgrading a Service during a [Deployment](deployments.md) from one version to another, Marathon can create more instances than specified in either Service definition. It will do so if allowed via the [upgradeStrategy](upgrade-strategy.md) configuration.
+When an instance is running on an agent which has disconnected from the cluster, it will eventually be reported as unreachable. According to the Service’s [unreachableStrategy](unreachable-strategy.md), Marathon might schedule a replacement, even though the unreachable instance might still be running. When an unreachable instance becomes reachable again, Marathon will scale down accordingly to the required amount of instances. It will decide which instance shall be killed based on the [killSelection](kill-selection.md).
+
+## Links
```
meichstedt: As noted by @zen-dog earlier, both Links and Examples point to resources which don't exist at the moment. We should either drop those links, or add the resources.

kensipe: yeah... I think we should create the docs (links that are missing) with an empty to be documented note... but I as assume that we will discuss this in our 1 on 1.

PR #6694, opened 2018-11-29
Do not put whole instance into provision update op

alenkacz:
Summary:
InstanceUpdateOperation `Provision` contained the whole instance. Instance update operations might be queued so it might happen that once the update is applied, that instance we enqueued is already obsolete and the instance changed (e.g. the goal changed).

With this change we always work with the newest instance while applying provision update.

JIRA issues: MARATHON-8516

### DIFF
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -78,52 +78,19 @@ case class Instance(
   override def region: Option[String] = agentInfo.flatMap(_.region)
 
   /**
-    * Factory method for creating provisioned instance from Scheduled instance for apps
+    * Factory method for creating provisioned instance from Scheduled instance
     * @return new instance in a provisioned state
     */
-  def provisioned(agentInfo: Instance.AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition, now: Timestamp, taskId: Task.Id): Instance = {
-    require(isScheduled, s"Instance '${instanceId}' must not be in state '${state.condition}'. Scheduled instance is required to create provisioned instance.")
+  def provisioned(agentInfo: Instance.AgentInfo, runSpecVersion: Timestamp, tasks: Seq[Task], now: Timestamp): Instance = {
+    require(isScheduled, s"Instance '$instanceId' must not be in state '${state.condition}'. Scheduled instance is required to create provisioned instance.")
 
     this.copy(
       agentInfo = Some(agentInfo),
-      state = Instance.InstanceState(Condition.Provisioned, now, None, None, Goal.Running),
-      tasksMap = Map(taskId -> Task(
-        taskId = taskId,
-        runSpecVersion = app.version,
-        status = Task.Status(
-          stagedAt = now,
-          condition = Condition.Provisioned,
-          networkInfo = networkInfo
-        ))),
-      runSpecVersion = app.version
-    )
-  }
-
-  /**
-    * Factory method for creating provisioned instance from Scheduled instance for pods
-    * @return new instance in a provisioned state
-    */
-  def provisioned(agentInfo: Instance.AgentInfo, hostPorts: Seq[Option[Int]], pod: PodDefinition, taskIds: Seq[Task.Id], now: Timestamp): Instance = {
-    require(isScheduled, s"Instance '${instanceId}' must not be in state '${state.condition}'. Scheduled instance is required to create provisioned instance.")
-
-    val taskNetworkInfos = Instance.podTaskNetworkInfos(pod, agentInfo, taskIds, hostPorts)
-
-    this.copy(
-      agentInfo = Some(agentInfo),
-      state = Instance.InstanceState(Condition.Provisioned, now, None, None, Goal.Running),
-      tasksMap = taskIds.map { taskId =>
-        // the task level host ports are needed for fine-grained status/reporting later on
-        val networkInfo = taskNetworkInfos.getOrElse(
-          taskId,
-          throw new IllegalStateException("failed to retrieve a task network info"))
-        val task = Task(
-          taskId = taskId,
-          runSpecVersion = pod.version,
-          status = Task.Status(stagedAt = now, condition = Condition.Provisioned, networkInfo = networkInfo)
-        )
+      state = Instance.InstanceState(Condition.Provisioned, now, None, None, this.state.goal),
+      tasksMap = tasks.map { task =>
         task.taskId -> task
       }(collection.breakOut),
-      runSpecVersion = pod.version
+      runSpecVersion = runSpecVersion
     )
   }
 }
@@ -171,40 +138,6 @@ object Instance {
     scheduledInstance.copy(
       reservation = Some(reservation),
       agentInfo = Some(agentInfo))
-  }
-
-  private def podTaskNetworkInfos(
-    pod: PodDefinition,
-    agentInfo: Instance.AgentInfo,
-    taskIDs: Seq[Task.Id],
-    hostPorts: Seq[Option[Int]]
-  ): Map[Task.Id, NetworkInfo] = {
-
-    val reqPortsByCTName: Seq[(String, Option[Int])] = pod.containers.flatMap { ct =>
-      ct.endpoints.map { ep =>
-        ct.name -> ep.hostPort
-      }
-    }
-
-    val totalRequestedPorts = reqPortsByCTName.size
-    require(totalRequestedPorts == hostPorts.size, s"expected that number of allocated ports ${hostPorts.size}" +
-      s" would equal the number of requested host ports $totalRequestedPorts")
-
-    require(!hostPorts.flatten.contains(0), "expected that all dynamic host ports have been allocated")
-
-    val allocPortsByCTName: Seq[(String, Int)] = reqPortsByCTName.zip(hostPorts).collect {
-      case ((name, Some(_)), Some(allocatedPort)) => name -> allocatedPort
-    }(collection.breakOut)
-
-    taskIDs.map { taskId =>
-      // the task level host ports are needed for fine-grained status/reporting later on
-      val taskHostPorts: Seq[Int] = taskId.containerName.map { ctName =>
-        allocPortsByCTName.withFilter { case (name, port) => name == ctName }.map(_._2)
-      }.getOrElse(Seq.empty[Int])
-
-      val networkInfo = NetworkInfo(agentInfo.host, taskHostPorts, ipAddresses = Nil)
-      taskId -> networkInfo
-    }(collection.breakOut)
   }
 
   /**
@@ -451,13 +384,6 @@ object Instance {
       attributes = offer.getAttributesList.toIndexedSeq
     )
   }
-
-  /**
-    * Marathon has requested (or will request) that this instance be launched by Mesos.
-    *
-    * @param instance is the thing that Marathon wants to launch
-    */
-  case class LaunchRequest(instance: Instance)
 
   implicit class LegacyInstanceImprovement(val instance: Instance) extends AnyVal {
     /** Convenient access to a legacy instance's only task */
--- a/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/Instance.scala
@@ -7,9 +7,7 @@ import com.fasterxml.uuid.{EthernetAddress, Generators}
 import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.condition.Condition.UnreachableInactive
 import mesosphere.marathon.core.instance.Instance.InstanceState
-import mesosphere.marathon.core.pod.PodDefinition
 import mesosphere.marathon.core.task.Task
-import mesosphere.marathon.core.task.state.NetworkInfo
 import mesosphere.marathon.raml.Raml
 import mesosphere.marathon.state.{MarathonState, PathId, Timestamp, UnreachableDisabled, UnreachableEnabled, UnreachableStrategy, _}
 import mesosphere.marathon.stream.Implicits._
--- a/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
@@ -35,7 +35,7 @@ private[marathon] class InstanceUpdateOpResolver(clock: Clock) extends StrictLog
       case op: Provision =>
         updateExistingInstance(maybeInstance, op.instanceId) { oldInstance =>
           // TODO(karsten): Create events
-          InstanceUpdateEffect.Update(op.instance, oldState = Some(oldInstance), Seq.empty)
+          InstanceUpdateEffect.Update(oldInstance.provisioned(op.agentInfo, op.runSpecVersion, op.tasks, op.now), oldState = Some(oldInstance), Seq.empty)
         }
 
       case RescheduleReserved(instance, runSpecVersion) =>
--- a/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
@@ -3,9 +3,8 @@ package core.instance.update
 
 import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.instance.{Goal, Instance}
-import mesosphere.marathon.core.instance.Instance
 import mesosphere.marathon.core.task.{Task, TaskCondition}
-import mesosphere.marathon.state.{AppDefinition, RunSpec, Timestamp}
+import mesosphere.marathon.state.Timestamp
 import org.apache.mesos
 
 import scala.collection.immutable.Seq
--- a/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
@@ -7,8 +7,6 @@ import mesosphere.marathon.core.task.{Task, TaskCondition}
 import mesosphere.marathon.state.Timestamp
 import org.apache.mesos
 
-import scala.collection.immutable.Seq
-
 sealed trait InstanceUpdateOperation {
   def instanceId: Instance.Id
 
--- a/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
+++ b/src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
@@ -4,9 +4,11 @@ package core.instance.update
 import mesosphere.marathon.core.condition.Condition
 import mesosphere.marathon.core.instance.{Goal, Instance}
 import mesosphere.marathon.core.instance.Instance
-import mesosphere.marathon.core.task.TaskCondition
-import mesosphere.marathon.state.Timestamp
+import mesosphere.marathon.core.task.{Task, TaskCondition}
+import mesosphere.marathon.state.{AppDefinition, RunSpec, Timestamp}
 import org.apache.mesos
+
+import scala.collection.immutable.Seq
 
 sealed trait InstanceUpdateOperation {
   def instanceId: Instance.Id
@@ -48,11 +50,8 @@ object InstanceUpdateOperation {
     * Scheduled instance have no agent info. Provisioned instances have such info. They are created when offer is
     * matched.
     *
-    * @param instance An instance that has been created during an offer match.
     */
-  case class Provision(instance: Instance) extends InstanceUpdateOperation {
-    override def instanceId: Instance.Id = instance.instanceId
-  }
+  case class Provision(instanceId: Instance.Id, agentInfo: Instance.AgentInfo, runSpecVersion: Timestamp, tasks: Seq[Task], now: Timestamp) extends InstanceUpdateOperation
 
   /**
     * Describes an instance update.
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -97,8 +97,14 @@ class InstanceOpFactoryImpl(
         val agentInfo = Instance.AgentInfo(offer)
         val taskIDs: Seq[Task.Id] = groupInfo.getTasksList.map { t => Task.Id(t.getTaskId) }(collection.breakOut)
         val now = clock.now()
-        val instanceOp = taskOperationFactory.provision(executorInfo, groupInfo, scheduledInstance.instanceId, agentInfo, pod.version, Tasks.provisioned(taskIDs, agentInfo, hostPorts, pod, now), now)
-        OfferMatchResult.Match(pod, offer, instanceOp, clock.now())
+        val instanceOp = taskOperationFactory.provision(
+          executorInfo,
+          groupInfo,
+          scheduledInstance.instanceId,
+          agentInfo,
+          pod.version,
+          Tasks.provisioned(taskIDs, agentInfo, hostPorts, pod, now), now)
+        OfferMatchResult.Match(pod, offer, instanceOp, now)
       case matchesNot: ResourceMatchResponse.NoMatch =>
         OfferMatchResult.NoMatch(pod, offer, matchesNot.reasons, clock.now())
     }
@@ -131,9 +137,15 @@ class InstanceOpFactoryImpl(
 
         val agentInfo = AgentInfo(offer)
 
-        val instanceOp = taskOperationFactory.provision(taskInfo, scheduledInstance.instanceId, agentInfo, app.version, Task.provisioned(taskId, networkInfo, app.version, clock.now()), clock.now())
+        val now = clock.now()
+        val instanceOp = taskOperationFactory.provision(
+          taskInfo,
+          scheduledInstance.instanceId,
+          agentInfo,
+          app.version,
+          Task.provisioned(taskId, networkInfo, app.version, now), now)
 
-        OfferMatchResult.Match(app, offer, instanceOp, clock.now())
+        OfferMatchResult.Match(app, offer, instanceOp, now)
       case matchesNot: ResourceMatchResponse.NoMatch => OfferMatchResult.NoMatch(app, offer, matchesNot.reasons, clock.now())
     }
   }
--- a/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
+++ b/src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
@@ -10,7 +10,7 @@ import mesosphere.marathon.core.instance.{Instance, LocalVolume, LocalVolumeId, 
 import mesosphere.marathon.core.launcher.{InstanceOp, InstanceOpFactory, OfferMatchResult}
 import mesosphere.marathon.core.plugin.PluginManager
 import mesosphere.marathon.core.pod.PodDefinition
-import mesosphere.marathon.core.task.Task
+import mesosphere.marathon.core.task.{Task, Tasks}
 import mesosphere.marathon.core.task.state.NetworkInfo
 import mesosphere.marathon.metrics.Metrics
 import mesosphere.marathon.plugin.scheduler.SchedulerPlugin
@@ -96,8 +96,8 @@ class InstanceOpFactoryImpl(
 
         val agentInfo = Instance.AgentInfo(offer)
         val taskIDs: Seq[Task.Id] = groupInfo.getTasksList.map { t => Task.Id(t.getTaskId) }(collection.breakOut)
-        val instance = scheduledInstance.provisioned(agentInfo, hostPorts, pod, taskIDs, clock.now())
-        val instanceOp = taskOperationFactory.provision(executorInfo, groupInfo, Instance.LaunchRequest(instance))
+        val now = clock.now()
+        val instanceOp = taskOperationFactory.provision(executorInfo, groupInfo, scheduledInstance.instanceId, agentInfo, pod.version, Tasks.provisioned(taskIDs, agentInfo, hostPorts, pod, now), now)
         OfferMatchResult.Match(pod, offer, instanceOp, clock.now())
       case matchesNot: ResourceMatchResponse.NoMatch =>
         OfferMatchResult.NoMatch(pod, offer, matchesNot.reasons, clock.now())
@@ -131,8 +131,7 @@ class InstanceOpFactoryImpl(
 
         val agentInfo = AgentInfo(offer)
 
-        val provisionedInstance = scheduledInstance.provisioned(agentInfo, networkInfo, app, clock.now(), taskId)
-        val instanceOp = taskOperationFactory.provision(taskInfo, provisionedInstance.appTask, provisionedInstance)
+        val instanceOp = taskOperationFactory.provision(taskInfo, scheduledInstance.instanceId, agentInfo, app.version, Task.provisioned(taskId, networkInfo, app.version, clock.now()), clock.now())
 
         OfferMatchResult.Match(app, offer, instanceOp, clock.now())
       case matchesNot: ResourceMatchResponse.NoMatch => OfferMatchResult.NoMatch(app, offer, matchesNot.reasons, clock.now())
@@ -269,7 +268,7 @@ class InstanceOpFactoryImpl(
             .build(offer, resourceMatch, Some(volumeMatch))
 
         val now = clock.now()
-        val stateOp = InstanceUpdateOperation.Provision(reservedInstance.provisioned(agentInfo, networkInfo, app, now, newTaskId))
+        val stateOp = InstanceUpdateOperation.Provision(reservedInstance.instanceId, agentInfo, app.version, Seq(Task.provisioned(newTaskId, networkInfo, app.version, now)), now)
 
         taskOperationFactory.launchOnReservation(taskInfo, stateOp, reservedInstance)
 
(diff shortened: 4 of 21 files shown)


src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -173,40 +140,6 @@ object Instance {
       agentInfo = Some(agentInfo))
   }
 
-  private def podTaskNetworkInfos(
```
alenkacz: I moved this to Task class since it's used for creating tasks, not instances

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOperation.scala
```diff
@@ -48,11 +50,8 @@ object InstanceUpdateOperation {
     * Scheduled instance have no agent info. Provisioned instances have such info. They are created when offer is
     * matched.
     *
-    * @param instance An instance that has been created during an offer match.
     */
-  case class Provision(instance: Instance) extends InstanceUpdateOperation {
-    override def instanceId: Instance.Id = instance.instanceId
-  }
+  case class Provision(instanceId: Instance.Id, agentInfo: Instance.AgentInfo, runSpecVersion: Timestamp, tasks: Seq[Task], now: Timestamp) extends InstanceUpdateOperation
```
alenkacz: this is one of the main changes

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -468,4 +471,64 @@ object Task {
   }
 
   implicit val taskFormat: Format[Task] = Json.format[Task]
+
+  def provisioned(taskId: Id, networkInfo: NetworkInfo, appVersion: Timestamp, now: Timestamp): Task = {
```
alenkacz: this was originally part of Instance class, but it's actually creating tasks so I moved it here. I think it's a better fit

src/test/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManagerTest.scala
```diff
@@ -52,27 +51,15 @@ class MarathonHealthCheckManagerTest extends AkkaUnitTest with Eventually {
     )
   }
 
-  def makeRunningTask(appId: PathId, version: Timestamp)(implicit instanceTracker: InstanceTracker): (Instance.Id, Task.Id) = {
```
alenkacz: this was not used

alenkacz:
most of the changes are changes in tests. Maaan, that's a mess. I think I'll remove some of the helpers after we merge this. It's just painful on every change I do...

zen-dog:
> I think I'll remove some of the helpers after we merge this

not only I'm willing to reveiw/help but you'll get kudos for that! 👍

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
```diff
@@ -121,18 +121,17 @@ class LaunchQueueModuleTest extends AkkaUnitTest with OfferMatcherSpec {
...
-    val scheduledInstance = Instance.scheduled(app)
-    val launchTaskInstanceOp = LaunchTask(mesosTask, InstanceUpdateOperation.Provision(scheduledInstance), Some(scheduledInstance), Seq.empty)
+    val provisionOp = InstanceUpdateOperation.Provision(scheduledInstance.instanceId, AgentInfoPlaceholder(), app.version, Seq(Task.provisioned(Task.Id.forInstanceId(scheduledInstance.instanceId), NetworkInfoPlaceholder(), app.version, Timestamp.now())), Timestamp.now())
+    val launchTaskInstanceOp = LaunchTask(mesosTask, provisionOp, Some(scheduledInstance), Seq.empty)
     val instanceChange = TaskStatusUpdateTestHelper(
-      operation = InstanceUpdateOperation.Provision(instance),
-      effect = InstanceUpdateEffect.Update(instance = instance, oldState = None, events = Nil)).wrapped
+      operation = provisionOp,
+      effect = InstanceUpdateEffect.Update(instance = scheduledInstance.provisioned(AgentInfoPlaceholder(), app.version, Seq(Task.provisioned(Task.Id.forInstanceId(scheduledInstance.instanceId), NetworkInfoPlaceholder(), app.version, Timestamp.now())), Timestamp.now()), oldState = None, events = Nil)).wrapped
```
zen-dog: I understand why `Task.provisioned` makes more sence in `Task.scala` but do we have to pass the tasks explicitely when calling `Instance.provisioned`? Shouldn't the later (Instance) just call the former (Task)?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -468,4 +471,64 @@ object Task {
...
+      Task(
+        taskId = taskId,
+        runSpecVersion = pod.version,
+        status = Task.Status(stagedAt = now, condition = Condition.Provisioned, networkInfo = networkInfo)
+      )
+    }
+  }
+
+  private def podTaskNetworkInfos(
+    pod: PodDefinition,
```
jeschkies: What do you think about just passing the container names?

src/main/scala/mesosphere/marathon/core/launcher/impl/InstanceOpFactoryImpl.scala
```diff
@@ -303,7 +302,8 @@ class InstanceOpFactoryImpl(
         val (executorInfo, groupInfo, hostPorts) = TaskGroupBuilder.build(pod, offer,
           instanceId, podContainerTaskIds, builderConfig, runSpecTaskProc, resourceMatch, Some(volumeMatch))
 
-        val stateOp = InstanceUpdateOperation.Provision(reservedInstance.provisioned(agentInfo, hostPorts, pod, podContainerTaskIds, clock.now()))
+        val now = clock.now()
+        val stateOp = InstanceUpdateOperation.Provision(reservedInstance.instanceId, agentInfo, pod.version, Tasks.provisioned(podContainerTaskIds, agentInfo, hostPorts, pod, now), now)
```
jeschkies: Hm, this is kind of mixing two things here. IMHO `InstanceUpdateOperation.Provision` is either an update or and command. If we make it a command I'd sent `Provision(reservedInstance.instanceId, agentInfo, version, taskIds, hostPorts, now)` and call `Tasks.provisioned()` in the update.

That said, I'm afraid the issue is structural, ie they way we update. The question for me is why are we overriding? As I see it we should set the goal and provision in sequence per instance id. This would avoid all these races we keep hunting. So something like

```scala
class Scheduler  {

val queue = Source.queue[Commands]
  .groupBy(16, cmd => Math.abs(cmd.instanceId.idString.hashCode) % 16)
  .mapAsync(1) {
     case ProcessMesosOffer(offer) => 
        val match = OfferMatcher(...)
        instanceTracker.update(match)
     case UpdateInstance(goal) => instanceTracker.update(...)
   }
  .mergeSubstreams

}

def processOffer(offer) = queue.offer(ProcessMesosOffer(offer, ...))

def setGoal(goal) = queue.offer(UpdateInstance(goal, ...))
```

I know it would not be possible right now but it is kind of how I imagined the architecture. Maybe going closer to what @timcharper proposed.

Anyways, coming back to this diff. We change the semantics of `InstanceUpdateOperation.Provision`. I never really liked that pattern and wish it would be a simple update. So what is the real issue here? That we matched something that had its goal changed in the mean time? IMHO that is something the `TaskLauncherActor` or `OfferMatcherManagerActor` should deal with.

jeschkies: Hm, this is kind of mixing two things here. IMHO `InstanceUpdateOperation.Provision` is either an update or and command. If we make it a command I'd sent `Provision(reservedInstance.instanceId, agentInfo, version, taskIds, hostPorts, now)` and call `Tasks.provisioned()` in the update.

That said, I'm afraid the issue is structural, ie they way we update. The question for me is why are we overriding? As I see it we should set the goal and provision in sequence per instance id. This would avoid all these races we keep hunting. So something like

```scala
class Scheduler  {

val queue = Source.queue[Commands]
  .groupBy(16, cmd => Math.abs(cmd.instanceId.idString.hashCode) % 16)
  .mapAsync(1) {
     case ProcessMesosOffer(offer) => 
        val match = OfferMatcher(...)
        instanceTracker.update(match)
     case UpdateInstance(goal) => instanceTracker.update(...)
   }
  .mergeSubstreams

}

def processOffer(offer) = queue.offer(ProcessMesosOffer(offer, ...))

def setGoal(goal) = queue.offer(UpdateInstance(goal, ...))
```

I know it would not be possible right now but it is kind of how I imagined the architecture. Maybe going closer to what @timcharper proposed.

Anyways, coming back to this diff. We change the semantics of `InstanceUpdateOperation.Provision`. I never really liked that pattern and wish it would be a simple update. So what is the real issue here? That we matched something that had its goal changed in the mean time? IMHO that is something the `TaskLauncherActor` or `OfferMatcherManagerActor` should deal with.

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
```diff
@@ -349,9 +349,10 @@ private class TaskLauncherActor(
 
     // Mark instance in internal map as provisioned
     instanceOp.stateOp match {
-      case InstanceUpdateOperation.Provision(instance) =>
-        assert(instanceMap.contains(instance.instanceId), s"Internal task launcher state did not include provisioned instance ${instance.instanceId}")
-        instanceMap += instance.instanceId -> instance
+      case InstanceUpdateOperation.Provision(instanceId, agentInfo, runSpecVersion, tasks, now) =>
+        assert(instanceMap.contains(instanceId), s"Internal task launcher state did not include newly provisioned instance $instanceId")
+        val existingInstance = instanceMap(instanceId)
+        instanceMap += instanceId -> existingInstance.provisioned(agentInfo, runSpecVersion, tasks, now)
```
jeschkies: I'm confused. Why did we call `Tasks.provision` before?

src/main/scala/mesosphere/marathon/core/task/Task.scala
jeschkies: Ah, I was just wondering why it is not part of the task class. Could you add some docs?

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -468,4 +471,64 @@ object Task {
...
+    Task(
+      taskId = taskId,
+      runSpecVersion = appVersion,
+      status = Task.Status(stagedAt = now, condition = Condition.Provisioned, networkInfo = networkInfo)
+    )
+  }
+}
+
+object Tasks {
+  def provisioned(taskIds: Seq[Id], agentInfo: Instance.AgentInfo, hostPorts: Seq[Option[Int]], pod: PodDefinition, now: Timestamp): Seq[Task] = {
```
jeschkies: Hm, It seems strange that the `provisioned` methods are in different objects.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
```diff
@@ -35,7 +35,7 @@ private[marathon] class InstanceUpdateOpResolver(clock: Clock) extends StrictLog
       case op: Provision =>
         updateExistingInstance(maybeInstance, op.instanceId) { oldInstance =>
           // TODO(karsten): Create events
-          InstanceUpdateEffect.Update(op.instance, oldState = Some(oldInstance), Seq.empty)
+          InstanceUpdateEffect.Update(oldInstance.provisioned(op.agentInfo, op.runSpecVersion, op.tasks, op.now), oldState = Some(oldInstance), Seq.empty)
```
jeschkies: Ok, so this is the issue, right? We would override other updates. I think I missed that. However, should we just change two lines. This seems all very overhead. Why not a simple

```scala
val updatedInstance = oldInstance.privisoned(op.agentInfo, op.networkInfo, op.app, op.taskId)
InstanceUpdateEffect.Update(updatedInstance, olderState = Some(oldInstance))
```

Ok, this misses pods but it seems way simpler than all this refactoring.

src/main/scala/mesosphere/marathon/core/task/Task.scala
alenkacz: I just wanted to have a plural in a name so `Tasks.provisioned` since it returns a collection...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -78,52 +76,19 @@ case class Instance(
...
-        // the task level host ports are needed for fine-grained status/reporting later on
-        val networkInfo = taskNetworkInfos.getOrElse(
-          taskId,
-          throw new IllegalStateException("failed to retrieve a task network info"))
-        val task = Task(
-          taskId = taskId,
-          runSpecVersion = pod.version,
-          status = Task.Status(stagedAt = now, condition = Condition.Provisioned, networkInfo = networkInfo)
-        )
+      state = Instance.InstanceState(Condition.Provisioned, now, None, None, this.state.goal),
```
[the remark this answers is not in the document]
jeschkies: That's the actual fix. `.provisioned` should not override the goal.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
```diff
@@ -78,52 +76,19 @@ case class Instance(
   override def region: Option[String] = agentInfo.flatMap(_.region)
 
   /**
-    * Factory method for creating provisioned instance from Scheduled instance for apps
+    * Factory method for creating provisioned instance from Scheduled instance
     * @return new instance in a provisioned state
     */
-  def provisioned(agentInfo: Instance.AgentInfo, networkInfo: core.task.state.NetworkInfo, app: AppDefinition, now: Timestamp, taskId: Task.Id): Instance = {
-    require(isScheduled, s"Instance '${instanceId}' must not be in state '${state.condition}'. Scheduled instance is required to create provisioned instance.")
+  def provisioned(agentInfo: Instance.AgentInfo, runSpecVersion: Timestamp, tasks: Seq[Task], now: Timestamp): Instance = {
```
jeschkies: Hm, how about passing a `Map[Task.Id, Task]` instead of `Seq[Task]`?

src/main/scala/mesosphere/marathon/core/task/Task.scala
jeschkies: As discussed it would be nice to unify the `Task.provisioned` for apps and pods in a follow up.

src/main/scala/mesosphere/marathon/core/task/Task.scala
```diff
@@ -18,8 +18,11 @@ import org.apache.mesos.{Protos => MesosProtos}
 
 import scala.concurrent.duration.FiniteDuration
 import mesosphere.marathon.api.v2.json.Formats._
+import mesosphere.marathon.core.task.Task.Id
 import play.api.libs.json._
 
+import scala.collection.immutable.Seq
```
jeschkies: Please remove.

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
```diff
@@ -121,18 +121,17 @@ class LaunchQueueModuleTest extends AkkaUnitTest with OfferMatcherSpec {
 
   class Fixture extends AutoCloseable {
     val app = MarathonTestHelper.makeBasicApp().copy(id = PathId("/app"))
+    val scheduledInstance = Instance.scheduled(app)
+    val task: Task = Task.provisioned(Task.Id.forInstanceId(scheduledInstance.instanceId), NetworkInfoPlaceholder(), app.version, Timestamp.now())
```
jeschkies: Yay. Production code and not test helper!

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
```diff
@@ -121,18 +121,17 @@ class LaunchQueueModuleTest extends AkkaUnitTest with OfferMatcherSpec {
...
 
     val offer = MarathonTestHelper.makeBasicOffer().build()
-    val runspecId = PathId("/test")
-    val instance = TestInstanceBuilder.newBuilder(runspecId).addTaskWithBuilder().taskRunning().build().getInstance()
-    val task: Task = instance.appTask
 
-    val mesosTask = MarathonTestHelper.makeOneCPUTask(task.taskId).build()
-    val scheduledInstance = Instance.scheduled(app)
-    val launchTaskInstanceOp = LaunchTask(mesosTask, InstanceUpdateOperation.Provision(scheduledInstance), Some(scheduledInstance), Seq.empty)
+    val provisionOp = InstanceUpdateOperation.Provision(scheduledInstance.instanceId, AgentInfoPlaceholder(), app.version, Seq(Task.provisioned(Task.Id.forInstanceId(scheduledInstance.instanceId), NetworkInfoPlaceholder(), app.version, Timestamp.now())), Timestamp.now())
```
jeschkies: Nit: I think this line is too long.  Also, why not reused the `task` from above?

src/test/scala/mesosphere/marathon/tasks/InstanceTrackerImplTest.scala
```diff
@@ -53,8 +53,6 @@ class InstanceTrackerImplTest extends AkkaUnitTest {
       val originalInstance: Instance = Instance.scheduled(AppDefinition(TEST_APP_NAME))
       instanceTracker.process(Schedule(originalInstance)).futureValue
 
-      instanceTracker.process(Provision(originalInstance)).futureValue
```
jeschkies: Why is that gone?

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryImplTest.scala
```diff
@@ -67,10 +67,7 @@ class InstanceOpFactoryImplTest extends UnitTest with Inside {
       )
 
       val expectedState = instance.state.copy(condition = Condition.Provisioned)
-      val expectedInstance = Instance(
-        expectedTaskId.instanceId, Some(expectedAgentInfo), expectedState, Map(expectedTaskId -> expectedTask),
-        runSpecVersion = app.version, app.unreachableStrategy, None)
-      assert(matched.instanceOp.stateOp == InstanceUpdateOperation.Provision(expectedInstance))
+      assert(matched.instanceOp.stateOp == InstanceUpdateOperation.Provision(expectedTaskId.instanceId, expectedAgentInfo, app.version, Seq(expectedTask), expectedState.since))
```
jeschkies: Shall we take the chance to move to `should` matchers here?

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryHelperTest.scala
```diff
@@ -51,10 +51,11 @@ class InstanceOpFactoryHelperTest extends UnitTest {
       val taskInfo = MarathonTestHelper.makeOneCPUTask(task.taskId).build()
 
       When("We create a launch operation")
-      val launch = f.helper.provision(taskInfo, task, instance)
+      val now = Timestamp.now()
+      val launch = f.helper.provision(taskInfo, instance.instanceId, instance.agentInfo.get, instance.runSpecVersion, task, now)
 
       Then("The result is as expected")
-      launch.stateOp shouldEqual InstanceUpdateOperation.Provision(instance)
+      launch.stateOp shouldEqual InstanceUpdateOperation.Provision(instance.instanceId, instance.agentInfo.get, instance.runSpecVersion, Seq(task), now)
```
jeschkies: How about using Scala test's `inside` here?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegateTest.scala
```diff
@@ -52,11 +52,11 @@ class InstanceTrackerDelegateTest extends AkkaUnitTest {
       create.futureValue shouldBe a[InstanceUpdateEffect.Update]
     }
 
-    "provisioning fails but the update stream keeps working" in {
+    "schedule fails but the update stream keeps working" in {
```
jeschkies: Why did this change to schedule?

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegateTest.scala
```diff
@@ -31,12 +31,12 @@ class InstanceTrackerDelegateTest extends AkkaUnitTest {
   }
 
   "InstanceTrackerDelegate" should {
-    "Provision succeeds" in {
+    "Schedule succeeds" in {
```
jeschkies: See below. I don't understand why this test changed.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
timcharper: Maybe we should call the param provisionalTasks?

src/test/scala/mesosphere/marathon/LaunchQueueModuleTest.scala
zen-dog: After talking about it, I don't feel strong about it. We can keep it this way.

src/main/scala/mesosphere/marathon/core/instance/update/InstanceUpdateOpResolver.scala
jeschkies: Pushing the tasks is fine for now.

src/main/scala/mesosphere/marathon/core/task/Task.scala
jeschkies: We can unify them in a follow up.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerActorTest.scala
```diff
@@ -241,25 +241,21 @@ class InstanceTrackerActorTest extends AkkaUnitTest with Eventually {
...
-        val runningOne = TestInstanceBuilder.newBuilder(appId).addTaskRunning().getInstance()
-        val runningTwo = TestInstanceBuilder.newBuilder(appId).addTaskRunning().getInstance()
-        val appDataMap = InstanceTracker.InstancesBySpec.forInstances(staged, runningOne, runningTwo)
+        val scheduled = Instance.scheduled(AppDefinition(appId))
+        val appDataMap = InstanceTracker.InstancesBySpec.forInstances(scheduled)
         f.taskLoader.load() returns Future.successful(appDataMap)
 
-        When("a new staged task gets added")
+        And("repository that returns error for store operation")
+        f.repository.store(any) returns Future.failed(new RuntimeException("fail"))
```
jeschkies: Much cleaner!

src/test/scala/mesosphere/marathon/core/task/bus/TaskStatusUpdateTestHelper.scala
```diff
@@ -22,17 +22,19 @@ class TaskStatusUpdateTestHelper(val operation: InstanceUpdateOperation, val eff
     case InstanceUpdateOperation.MesosUpdate(_, marathonTaskStatus, mesosStatus, _) => mesosStatus
     case _ => throw new scala.RuntimeException("the wrapped stateOp os no MesosUpdate!")
   }
+
+  private[this] def instanceFromOperation: Instance = operation match {
```
jeschkies: I think you can remove it. It has only one use case in `def updatedInstance` and since we checkout only for `MesosUpdate` we can to that in `updatedInstance`.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -146,7 +150,7 @@ class OfferProcessorImplTest extends UnitTest {
           val taskId = Task.Id(taskInfo.getTaskId)
           val launch = f.launchWithNewTask(
             taskInfo,
-            InstanceUpdateOperation.Provision(dummyInstance.provisioned(AgentInfoPlaceholder(), NetworkInfoPlaceholder(), AppDefinition(appId), clock.now(), taskId)),
+            InstanceUpdateOperation.Provision(dummyInstance.instanceId, AgentInfoPlaceholder(), AppDefinition(appId).version, Seq(Task.provisioned(taskId, NetworkInfoPlaceholder(), AppDefinition(appId).version, clock.now())), clock.now()),
```
jeschkies: Why do we have `AppDefinition(appId).version` twice? This is hard to read. `AppDefinition(appId)` is used so often that it probably helps to save it as a `val app`.

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -104,7 +106,9 @@ class OfferProcessorImplTest extends UnitTest {
     "match successful, launch tasks unsuccessful" in new Fixture {
       Given("an offer")
       val dummySource = new DummySource
-      val tasksWithSource = tasks.map(task => InstanceOpWithSource(dummySource, f.provision(task._1, task._2, task._3)))
+      val tasksWithSource = tasks.map {
+        case (taskInfo, task, instance) => InstanceOpWithSource(dummySource, f.provision(taskInfo, instance.instanceId, instance.agentInfo.get, instance.runSpecVersion, task, clock.now()))
```
jeschkies: I think we have a 120 or 140 chars line limit. Also why not
```scala
case (taskInfo, task, Instance(instanceId, Some(agentInfo), runSpecVersion)) => 
  val whateverThisIs =  f.provision(taskInfo, instanceId, agentInfo, runSpecVersion, task, clock.now())
  InstanceOpWithSource(dummySource, whateverThisIs)
```

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
```diff
@@ -70,12 +70,14 @@ class OfferProcessorImplTest extends UnitTest {
...
+      }
 
       And("a cooperative offerMatcher and taskTracker")
       offerMatcher.matchOffer(offer) returns Future.successful(MatchedInstanceOps(offerId, tasksWithSource))
-      for (task <- tasks) {
-        val stateOp = InstanceUpdateOperation.Provision(task._3)
+      for ((_, _, instance) <- tasks) {
```
jeschkies: You could use our new fancy [`Tasks`](https://github.com/mesosphere/marathon/pull/6687/files#diff-b113d7828c8c9e92a7004f07797e2e1aR38) extractor

```scala
tasks.foreach {
  case (_, _ , Instance(instanceId, Tasks(tasks), ... Some(agentInfo), runSpecVersion)) =>
}
```

Overall this seems very messy. Eg 
```scala
val dummySource = new DummySource
val tasksWithSource = tasks.map(...)
```

We really have to revisit this test.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
alenkacz: why Map? the fact that instance needs to have it as a id indexed map is an implementation detail to me. Someone has to do the seq to map transformation, I am fine with the updater knowing the instance internals does it, not every client wanting to submit the update...

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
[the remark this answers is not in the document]
alenkacz: Exactly what Karsten said

src/main/scala/mesosphere/marathon/core/task/Task.scala
alenkacz: MARATHON-8521

src/test/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManagerTest.scala
```diff
@@ -52,27 +51,15 @@ class MarathonHealthCheckManagerTest extends AkkaUnitTest with Eventually {
...
-
   def setupTrackerWithProvisionedInstance(appId: PathId, version: Timestamp, instanceTracker: InstanceTracker): Future[Instance] = async {
     val app = AppDefinition(appId, versionInfo = VersionInfo.OnlyVersion(version))
     val scheduledInstance = Instance.scheduled(app)
     // schedule
     await(instanceTracker.schedule(scheduledInstance))
     // provision
-    val provisionedInstance = scheduledInstance.provisioned(AgentInfoPlaceholder(), NetworkInfoPlaceholder(), app, Timestamp.now(), Task.Id.forInstanceId(scheduledInstance.instanceId))
-    await(instanceTracker.process(InstanceUpdateOperation.Provision(provisionedInstance)))
+    await(instanceTracker.process(InstanceUpdateOperation.Provision(scheduledInstance.instanceId, AgentInfoPlaceholder(), version, Seq(Task.provisioned(Task.Id.forInstanceId(scheduledInstance.instanceId), NetworkInfoPlaceholder(), version, Timestamp.now())), Timestamp.now())))
```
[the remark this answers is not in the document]
alenkacz: for some reason, provision was always called via `.process()`. I would defer refactoring this until we have some idea how we want the scheduler interface to look like so that we don't change it multiple times

I extracted the `Timestamp.now()`

src/test/scala/mesosphere/marathon/core/health/impl/MarathonHealthCheckManagerTest.scala
```diff
@@ -52,27 +51,15 @@ class MarathonHealthCheckManagerTest extends AkkaUnitTest with Eventually {
...
 
-    provisionedInstance
+    await(instanceTracker.get(scheduledInstance.instanceId)).get
```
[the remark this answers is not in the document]
alenkacz: good point, thanks!

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
alenkacz: nice. But it's not merged yet though... Also I think this test has much bigger problems that not using the correct extractor 🙈 :D

src/test/scala/mesosphere/marathon/core/launcher/impl/OfferProcessorImplTest.scala
alenkacz: I don't want to be counting chars in my IDE. We should have a formatter than that will enforce it for me...

And re the extractor: I'll use that, thanks...

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegateTest.scala
alenkacz: we are just testing that an operation is applied, I believe it does not matter which operation. It does not cover all operations right now anyway. We can definitely do provision but we can also do schedule at this point. I think it does not matter. It's just easier to create scheduled instance so I went that way.

If you see what is it testing, it's just verifying that InstanceTracker received `UpdateContext`. The operation itself is pretty irrelevant.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegateTest.scala
alenkacz: See reply in the previous comment https://github.com/mesosphere/marathon/pull/6694/files#r237695644

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryHelperTest.scala
alenkacz: I don't like that too much, I find the errors that are coming out of that cryptic. I think this is pretty readable, I would stick with this implementation

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegateTest.scala
alenkacz: btw. I would like to get rid of `newBuilderWithLaunchedTask` once we merge this one, I think there is no need for that since it's just creating tasks in provisioned state

src/test/scala/mesosphere/marathon/tasks/InstanceTrackerImplTest.scala
alenkacz: It is not adding any value... this test is just testing that if you push an update to instance tracker and then query the instance tracker, you get that instance. I believe the provision was here (as in other places in the tests) because up until Phase 2, provision was the operation that created the instance. So to test this, now schedule is enough... This is not the level on which we would want to test all the operations...

src/main/scala/mesosphere/marathon/core/launchqueue/impl/TaskLauncherActor.scala
jeschkies: Ah, dang. Yeah, that should go away once the `TaskLauncherActor` is stateless.

src/main/scala/mesosphere/marathon/core/instance/Instance.scala
jeschkies: Yup. Fine with me.

src/test/scala/mesosphere/marathon/tasks/InstanceTrackerImplTest.scala
jeschkies: Ah, now I remember. Back then I was not sure which futures are called. Thanks!

src/test/scala/mesosphere/marathon/tasks/InstanceOpFactoryHelperTest.scala
jeschkies: Hehe, that's funny. I actually do not like these long stings dumps of X was not Y here are 20 lines with 120 characters each to check for you 😉. We can implement nicer diffs in the future. That should not block this PR.

src/test/scala/mesosphere/marathon/core/task/tracker/impl/InstanceTrackerDelegateTest.scala
jeschkies: Thanks for explaining. That makes a lot of sense to me now.

PR #6695, opened 2018-11-29
Added new MoM-EE test image `v1.7.181_1.12.1`

zen-dog:
Removed old and unsupported MoM-EE 1.4.x definitions. Additionally updated README with instructions on how to run SI tests locally.


tests/system/README.md
```diff
@@ -4,14 +4,20 @@ This directory contains system integration tests of marathon in a DCOS environme
 
 To run the test you need a DC/OS cluster, Ptyhon 3.5+, dcos-cli 0.5.5 and shakedown 1.4.8 installed.
 
-First authenticate with your cluster with
+Then run a specific test:
```
alenkacz: Then does not make sense since you dropped the first part of the docs

zen-dog:
> It's kind of hard to review

You mean the MoM-EE definitions? Beause I simply removed the old (and pretty outdated) 1.4.x one and added new 1.7.xx (which is basically the same as 1.6 and 1.5)

PR #6696, opened 2018-11-29
Fix handling of the missing dcos certificate in SI tests

zen-dog:
Previously an SI run would fail if `DCOS_SSL_VERIFY` would not be set properly. Additionally removed some unnecessary logging.

### DIFF
--- a/tests/shakedown/shakedown/clients/rpcclient.py
+++ b/tests/shakedown/shakedown/clients/rpcclient.py
@@ -32,7 +32,7 @@ def load_error_json_schema():
 
 
 def get_ca_file():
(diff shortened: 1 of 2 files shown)


tests/shakedown/shakedown/clients/rpcclient.py
```diff
@@ -32,7 +32,7 @@ def load_error_json_schema():
 
 
 def get_ca_file():
-    return Path(environ.get('DCOS_SSL_VERIFY'))
+    return None if 'DCOS_SSL_VERIFY' not in environ else Path(environ.get('DCOS_SSL_VERIFY'))
```
jeschkies: Python is awesome, isn't it? 🙂

